PR #314, opened 2014-08-14
Fixed path in SCL statement

lzap:
Or we can completely remove that. I always hit this when I forget to build into
nonscl tag :-) What you think?


mmoll:
what's the status here? :)

lzap:
I guess I can close

PR #389, opened 2014-10-14
Mock configs and instructions for building packages locally

domcleal:


domcleal:
By the way, the site-defaults examples no longer include ~/.mock as the default setting, so it'd use /var.  I don't see why it shouldn't work (mine does), but I was only doing it as I have partition/space issues here.

domcleal:
Merged in 17d1e25.

PR #426, opened 2014-11-04
Refs #4478 - prebuild localized apipie cache

mbacovsky:
Requires https://github.com/theforeman/foreman/pull/1428 to get merged.
Documentation for resources is built during build time. Docs for Index is build in posttrans phase as we need foreman plugins to be present.

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -449,6 +449,8 @@ cp config/settings.yaml.example config/settings.yaml
 export BUNDLER_EXT_NOSTRICT=1
 export BUNDLER_EXT_GROUPS="default assets"
 %{scl_rake} assets:precompile:all RAILS_ENV=production --trace
+%{scl_rake} db:migrate RAILS_ENV=production --trace
+%{scl_rake} apipie:cache RAILS_ENV=production cache_part=resources OUT=public/apipie-cache/plugin/%{name} --trace
 rm config/database.yml config/settings.yaml
 
 %install
@@ -563,7 +565,9 @@ cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-assets << EOF
 # Generate precompiled assets at gem_instdir/public/assets/gem_name/
 # -r<rake_task>     Overrides rake task of plugin:assets:precompile[plugin_name]
 # -n<plugin_name>   Overrides default of gem_name for precompile step
-%%%{name}_precompile_plugin(r:n:) \\
+# -a                Prebuild apipie cache
+# -s                Precompile assets
+%%%{name}_precompile_plugin(r:n:as) \\
 mkdir -p ./usr/share \\
 cp -r %%{%{name}_dir} ./usr/share || echo 0 \\
 pushd ./usr/share/%{name} \\
@@ -573,10 +577,24 @@ cp %%{buildroot}%%{%{name}_bundlerd_dir}/%%{gem_name}.rb ./bundler.d/%%{gem_name
 unlink tmp \\
 \\
 export BUNDLER_EXT_NOSTRICT=1 \\
-/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=production --trace \\
+%%{?-s:/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=production --trace} \\
+%%{?-a:/usr/bin/%%{?scl:%%{scl}-}rake plugin:apipie:cache[%%{gem_name}] RAILS_ENV=development cache_part=resources OUT=%%{buildroot}%%{gem_instdir}/public/apipie-cache/plugin/%%{gem_name} --trace} \\
 \\
 popd \\
-rm -rf ./usr
+rm -rf ./usr \\
+%%{?-a:mkdir -p %%{buildroot}%%{foreman_dir}/public/apipie-cache/plugin} \\
+%%{?-a:ln -s %%{gem_instdir}/public/apipie-cache/plugin/%%{gem_name} %%{buildroot}%%{foreman_dir}/public/apipie-cache/plugin/%%{gem_name}}
+EOF
+
+cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-apipie << EOF
+# Common locations
+%%%{name}_apipie_cache_plugin %%{gem_instdir}/public/apipie-cache/plugin/%%{gem_name}
+%%%{name}_apipie_cache_foreman %%{foreman_dir}/public/apipie-cache/plugin/%%{gem_name}
+
+%%%{name}_apipie_cache_plugin_post \\
+cp -r %%{foreman_apipie_cache_foreman}/* %%{foreman_dir}/public/apipie-cache/ \\
+chown -R %{name}.%{name} %%{foreman_dir}/public/apipie-cache
+
 EOF
 
 %clean
@@ -621,6 +639,8 @@ rm -rf %{buildroot}
 %config(noreplace) %{_sysconfdir}/logrotate.d/%{name}
 %config %{_sysconfdir}/cron.d/%{name}
 %{_sysconfdir}/rpm/macros.%{name}
+%{_sysconfdir}/rpm/macros.%{name}-apipie
+%{_sysconfdir}/rpm/macros.%{name}-assets
 %attr(-,%{name},%{name}) %{_localstatedir}/lib/%{name}
 %attr(750,%{name},%{name}) %{_localstatedir}/log/%{name}
 %attr(750,%{name},%{name}) %{_localstatedir}/log/%{name}/plugins
@@ -672,6 +692,10 @@ if [ ! -e %{_datadir}/%{name}/config/initializers/encryption_key.rb -a \
   ln -s %{_sysconfdir}/%{name}/encryption_key.rb %{_datadir}/%{name}/config/initializers/
 fi
 
+# copy pregenerated apipie cache files
+cp -r %{_datadir}/%{name}/public/apipie-cache/plugin/%{name}/* %{_datadir}/%{name}/public/apipie-cache/
+chown -R %{name}.%{name} %{_datadir}/%{name}/public/apipie-cache/
+
 /sbin/chkconfig --add %{name} || :
 (/sbin/service foreman status && /sbin/service foreman restart) >/dev/null 2>&1
 exit 0
@@ -681,7 +705,7 @@ exit 0
 # always attempt to reencrypt after update in case new fields can be encrypted
 %{foreman_rake} db:migrate db:compute_resources:encrypt >> %{_localstatedir}/log/%{name}/db_migrate.log 2>&1 || :
 %{foreman_rake} db:seed >> %{_localstatedir}/log/%{name}/db_seed.log 2>&1 || :
-%{foreman_rake} apipie:cache >> %{_localstatedir}/log/%{name}/apipie_cache.log 2>&1 || :
+%{foreman_rake} apipie:cache cache_part=index >> %{_localstatedir}/log/%{name}/apipie_cache.log 2>&1 || :
 (/sbin/service foreman status && /sbin/service foreman restart) >/dev/null 2>&1
 exit 0
 
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -594,6 +594,7 @@ unlink tmp \\
 \\
 export BUNDLER_EXT_NOSTRICT=1 \\
 %%{?-s:/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=production --trace} \\
+%%{?-a:/usr/bin/%%{?scl:%%{scl}-}rake db:migrate RAILS_ENV=development --trace} \\
 %%{?-a:/usr/bin/%%{?scl:%%{scl}-}rake plugin:apipie:cache[%%{gem_name}] RAILS_ENV=development cache_part=resources OUT=%%{buildroot}%%{gem_instdir}/public/apipie-cache/plugin/%%{gem_name} --trace} \\
 \\
 popd \\
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -550,8 +550,6 @@ cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name} << EOF
 # Common locations
 %%%{name}_dir %{_datadir}/%{name}
 %%%{name}_bundlerd_dir %%{%{name}_dir}/bundler.d
-%%%{name}_bundlerd_plugin %%{%{name}_bundlerd_dir}/%%{gem_name}.rb
-%%%{name}_pluginconf_dir %{_sysconfdir}/%{name}/plugins
 %%%{name}_log_dir %{_localstatedir}/log/%{name}
 
 # Common commands
@@ -559,7 +557,9 @@ cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name} << EOF
 %%%{name}_db_migrate   %%{%{name}_rake} db:migrate >> %%{%{name}_log_dir}/db_migrate.log 2>&1 || :
 %%%{name}_db_seed      %%{%{name}_rake} db:seed >> %%{%{name}_log_dir}/db_seed.log 2>&1 || :
 %%%{name}_restart      (/sbin/service %{name} status && /sbin/service %{name} restart) >/dev/null 2>&1
+EOF
 
+cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-plugin << EOF
 # Generate bundler.d file for a plugin
 # -n<plugin_name>   Overrides default of gem_name
 %%%{name}_bundlerd_file(n:) \\
@@ -567,10 +567,11 @@ mkdir -p %%{buildroot}%%{%{name}_bundlerd_dir} \\
 cat <<GEMFILE > %%{buildroot}%%{%{name}_bundlerd_dir}/%%{-n*}%%{!?-n:%%{gem_name}}.rb \\
 gem '%%{-n*}%%{!?-n:%%{gem_name}}' \\
 GEMFILE
-EOF
 
-cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-plugin << EOF
 # Common locations
+%%%{name}_bundlerd_plugin %%{%{name}_bundlerd_dir}/%%{gem_name}.rb
+%%%{name}_pluginconf_dir %{_sysconfdir}/%{name}/plugins
+# Common assets locations
 %%%{name}_assets_plugin %%{gem_instdir}/public/assets/%%{gem_name}
 # Common apipie locations
 %%%{name}_apipie_cache_plugin %%{gem_instdir}/public/apipie-cache/plugin/%%{gem_name}
@@ -584,14 +585,20 @@ cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-plugin << EOF
 # -a                Prebuild apipie cache
 # -s                Precompile assets
 %%%{name}_precompile_plugin(r:n:as) \\
-mkdir -p ./usr/share \\
-cp -r %%{%{name}_dir} ./usr/share || echo 0 \\
-pushd ./usr/share/%{name} \\
+mkdir -p ./%{_datadir} \\
+cp -r %%{%{name}_dir} ./%{_datadir} || echo 0 \\
+mkdir -p ./%{_localstatedir}/lib/%{name} \\
+cp -r %{_localstatedir}/lib/%{name}/db ./%{_localstatedir}/lib/%{name} || echo 0 \\
+unlink ./%{_datadir}/%{name}/db \\
+ln -sv \`pwd\`/%{_localstatedir}/lib/%{name}/db ./%{_datadir}/%{name}/db \\
+pushd ./%%{%{name}_dir} \\
 \\
 export GEM_PATH=%%{gem_dir}:%%{buildroot}%%{gem_dir} \\
 cp %%{buildroot}%%{%{name}_bundlerd_dir}/%%{gem_name}.rb ./bundler.d/%%{gem_name}.rb \\
 unlink tmp \\
 \\
+rm \`pwd\`/config/initializers/encryption_key.rb \\
+/usr/bin/%%{?scl:%%{scl}-}rake security:generate_encryption_key \\
 export BUNDLER_EXT_NOSTRICT=1 \\
 %%{?-s:/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=production --trace} \\
 %%{?-a:/usr/bin/%%{?scl:%%{scl}-}rake db:migrate RAILS_ENV=development --trace} \\
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -354,7 +354,19 @@ Meta package to install asset pipeline support.
 
 %files assets
 %{_datadir}/%{name}/bundler.d/assets.rb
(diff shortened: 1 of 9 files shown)


domcleal:
Just a reminder, this needs finishing off for 1.8 (for both RPMs and debs) else I'm going to have to revert the change to Foreman core - the apipie:cache performance is simply dreadful at the moment out of the box.

mbacovsky:
I rebased the PR. 
I changed the default options in foreman_precompile_plugin macro, I tried to avoid that, but this way it seems cleaner. The affected packages were updated in separate commits. I'll have to add the package names to the commit messages.

Rubygem foreman bootdisk updated and scratch builds of all the related packages are still WIP

lzap:
@ohadlevy you asked on Discovery API docs generation yesterday - once this is merged we should have it.

mbacovsky:
I updated the commit messages and added foreman-bootdisk.

foreman/foreman.spec
```diff
@@ -570,6 +588,8 @@ rm -rf %{buildroot}
 %config(noreplace) %{_sysconfdir}/logrotate.d/%{name}
 %config %{_sysconfdir}/cron.d/%{name}
 %{_sysconfdir}/rpm/macros.%{name}
+%{_sysconfdir}/rpm/macros.%{name}-apipie
+%{_sysconfdir}/rpm/macros.%{name}-assets
```
domcleal: Should be removed from -assets' %files

foreman/foreman.spec
```diff
@@ -540,10 +544,24 @@ cp %%{buildroot}%%{%{name}_bundlerd_dir}/%%{gem_name}.rb ./bundler.d/%%{gem_name
...
+EOF
+
+cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-apipie << EOF
+# Common locations
+%%%{name}_apipie_cache_plugin %%{gem_instdir}/public/apipie-cache/plugin/%%{gem_name}
+%%%{name}_apipie_cache_foreman %%{foreman_dir}/public/apipie-cache/plugin/%%{gem_name}
+
+%%%{name}_apipie_cache_plugin_post \\
++cp -r %%{foreman_apipie_cache_foreman}/* %%{foreman_dir}/public/apipie-cache/ \\
++chown -R %{name}.%{name} %%{foreman_dir}/public/apipie-cache
```
domcleal: additional leading "+" on these lines, that doesn't look right to me?

foreman/foreman.spec
```diff
@@ -540,10 +544,24 @@ cp %%{buildroot}%%{%{name}_bundlerd_dir}/%%{gem_name}.rb ./bundler.d/%%{gem_name
 unlink tmp \\
 \\
 export BUNDLER_EXT_NOSTRICT=1 \\
-/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=production --trace \\
+%%{?-s:/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=production --trace} \\
+%%{?-a:/usr/bin/%%{?scl:%%{scl}-}rake plugin:apipie:cache[%%{gem_name}] RAILS_ENV=development cache_part=resources OUT=%%{buildroot}%%{gem_instdir}/public/apipie-cache/plugin/%%{gem_name} #gem_instdir}/public/apipie-cache/plugin/%%{gem_name} --trace} \\
```
domcleal: is there a comment halfway through this line?

foreman/foreman.spec
```diff
@@ -621,6 +641,10 @@ if [ ! -e %{_datadir}/%{name}/config/initializers/encryption_key.rb -a \
   ln -s %{_sysconfdir}/%{name}/encryption_key.rb %{_datadir}/%{name}/config/initializers/
 fi
 
+# copy pregenerated apipie cache files
```
domcleal: Shouldn't this be in the build?

rubygem-foreman_bootdisk/rubygem-foreman_bootdisk.spec
```diff
@@ -42,6 +42,7 @@ BuildRequires: %{?scl_prefix}ruby(release)
 %else
 BuildRequires: %{?scl_prefix}ruby(abi) >= %{rubyabi}
 %endif
+BuildRequires:   foreman >= 1.8.0
```
domcleal: indentation's off

rubygem-foreman_bootdisk/rubygem-foreman_bootdisk.spec
```diff
@@ -82,16 +83,20 @@ cat <<GEMFILE > %{buildroot}%{foreman_bundlerd_dir}/%{gem_name}.rb
 gem '%{gem_name}'
 GEMFILE
 
+%foreman_precompile_plugin -a
```
domcleal: This is failing, as the rake task's trying to initialise the database, and we don't have foreman-sqlite as a build dep (thankfully?).

<pre>
+ /usr/bin/ruby193-rake 'plugin:apipie:cache[foreman_bootdisk]' RAILS_ENV=development cache_part=resources OUT=/builddir/build/BUILDROOT/ruby193-rubygem-foreman_bootdisk-4.0.2-1.git.188.6c277bf.el7.centos.x86_64/opt/rh/ruby193/root/usr/share/gems/gems/foreman_bootdisk    -4.0.2/public/apipie-cache/plugin/foreman_bootdisk
Rails Error: Unable to access log file. Please ensure that /builddir/build/BUILD/rubygem-foreman_bootdisk-4.0.2/usr/share/foreman/log/development.log exists and is chmod 0666. The log level has been raised to WARN and the output directed to STDERR until the problem is     fixed.
rake aborted!
Please install the sqlite3 adapter: `gem install activerecord-sqlite3-adapter` (Could not find sqlite3 (~> 1.3.5) amongst [actionmailer-3.2.8, actionpack-3.2.8, activemodel-3.2.8, activerecord-3.2.8, activeresource-3.2.8, activesupport-3.2.8, ancestry-2.0.0, apipie-ra    ils-0.2.6, arel-3.0.2, audited-3.0.0, audited-activerecord-3.0.0, bigdecimal-1.1.0, builder-3.0.0, bundler-1.3.5, bundler_ext-0.3.0, coffee-rails-3.2.2, coffee-script-2.2.0, coffee-script-source-1.3.3, deep_cloneable-2.0.2, diff-lcs-1.1.3, erubis-2.7.0, execjs-1.4.0,     fast_gettext-0.9.2, foreigner-1.7.1, foreman_bootdisk-4.0.2, friendly_id-4.0.10.1, gettext_i18n_rails-1.2.0, hike-1.2.1, i18n-0.6.0, io-console-0.3, journey-1.0.4, json-1.5.5, ldap_fluff-0.3.3, mail-2.4.4, mime-types-1.19, minitest-2.5.1, multi_json-1.10.1, net-http-p    ersistent-2.7, net-ldap-0.10.0, netrc-0.7.7, oauth-0.4.7, polyglot-0.3.3, rabl-0.11.4, rack-1.4.1, rack-cache-1.2, rack-ssl-1.3.2, rack-test-0.6.1, rails-3.2.8, railties-3.2.8, rake-0.9.2.2, rdoc-3.9.5, rest-client-1.6.7, ruby2ruby-2.1.3, ruby_parser-3.6.3, safemode-1    .2.2, scoped_search-2.7.1, secure_headers-1.4.1, sexp_processor-4.4.4, sprockets-2.8.2, thor-0.18.1, tilt-1.3.3, treetop-1.4.10, turbolinks-2.5.3, tzinfo-0.3.33, validates_lengths_from_database-0.4.0, will_paginate-3.0.2])
Tasks: TOP => plugin:apipie:cache
(See full trace by running task with --trace)
</pre>

Is there any way we can fix the rake task not to initialise the database connection, or is it necessary?

foreman/foreman.spec
```diff
@@ -540,10 +544,24 @@ cp %%{buildroot}%%{%{name}_bundlerd_dir}/%%{gem_name}.rb ./bundler.d/%%{gem_name
...
+%%{?-a:ln -s %%{gem_instdir}/public/apipie-cache/plugin/%%{gem_name} %%{buildroot}%%{foreman_dir}/public/apipie-cache/plugin/%%{gem_name}}
+EOF
+
+cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-apipie << EOF
+# Common locations
+%%%{name}_apipie_cache_plugin %%{gem_instdir}/public/apipie-cache/plugin/%%{gem_name}
+%%%{name}_apipie_cache_foreman %%{foreman_dir}/public/apipie-cache/plugin/%%{gem_name}
+
+%%%{name}_apipie_cache_plugin_post \\
++cp -r %%{foreman_apipie_cache_foreman}/* %%{foreman_dir}/public/apipie-cache/ \\
```
domcleal: similar query to below about the copy/build stage, why are there two copies of this?  It's also visible in %files in bootdisk for instance, which seems strange to me.

domcleal: If there really must be two copies for some reason, and we can't remove it or symlink them, then we should copy them at build time and have the package ship them both, not copy it in %post.  This will ensure that file ownership, file cleanup, "rpm -V" etc works, and the RPM should compress away the second copy.

mbacovsky: For a plugin there is one copy of the prebuilt cache in `%{gem_instdir}/public/apipie-cache/plugin/%{gem_name}` and it is linked to the `%{foreman_dir}/public/apipie-cache/plugin/%%{gem_name}`.
The plugin can contailn the same resources as the foreman core (e.g. environments in Foreman and Katello) so it is not possible to link it to `%{foreman_dir}/public/apipie-cache/` directly in build time as it would cause conflicts. Therefore we copy it in post and overwriting the duplicites.

I agree it would be better to have just one copy in the package. And I would prefere to drop the copy in the `gem_instdir`. Would that be okay?

foreman/foreman.spec
mbacovsky: Oh, fixed

foreman/foreman.spec
mbacovsky: fixed

foreman/foreman.spec
mbacovsky: This is mainly for case of Foreman update. This code is responsible for copying of the plugin's pre-built caches to the main `apipie-cache` directory. It can not happen during the build time because of conflict (resource with same name in core and plugin).

foreman/foreman.spec
mbacovsky: You mean the %foreman_precompile_plugin macro should be romoved from -assets file? Should I move it to -plugin?

mbacovsky:
Part of the issues was fixed. I'll have to investigate the issue with bootdisk. The rake task needs to load rails environment and that requires DB. For example for Katello I was unable to avoid that. I'll have to discuss it with @iNecas.

iNecas:
Just to share our experience with trying to avoid the database requirement when building the apipie documentation, trying to workaround the Rails database needs lead to big pain every time something changed (either in our code or in Rails). The only other thinkable solution would be to not use the model classes inside the apipie definitions, but that's not very convenient either.

mbacovsky:
@domcleal I tried to rebuild rubygem-foreman_bootdisk with updated foreman and it worked. The result rpm contains the prebuilt cache. The build log is here http://fpaste.org/171401/42167563/ and the Foreman scratch build can be found here http://koji.katello.org/koji/taskinfo?taskID=212128 (both on RHEL 7) 
Foreman_bootdisk scratch: https://mbacovsk.fedorapeople.org/packages/ruby193-rubygem-foreman_bootdisk-4.0.2-1.el6.noarch.rpm, https://mbacovsk.fedorapeople.org/packages/ruby193-rubygem-foreman_bootdisk-4.0.2-1.el6.src.rpm

mbacovsky:
The foreman package has been rebased

foreman/foreman.spec
domcleal: Interesting, so what wins in that case?  It looks like it would be the most recently installed package, rather than anything deterministic - e.g. the latest Foreman version, or the latest plugin version?

foreman/foreman.spec
domcleal: Sorry, I meant that if you're adding the -assets macro file to 'foreman' then it can be removed from 'foreman-assets' so it's not in both (though harmless).

I do kind of like the idea of a -plugin or -build subpackage though, which could depend on foreman-sqlite and ship these macro files.  Or we repurpose foreman-assets for this too?

domcleal:
Regarding the sqlite dep, that's OK, we'll live with it.

I think the build worked for you @mbacovsky because you may have had ruby193-rubygem-sqlite installed on your build host.  It failed for me in a clean mock environment.

mock is very easy to set up - install it, then copy one of the configs from #389 to /etc/mock, e.g. /etc/mock/el7-scl-foreman.cfg (that's how I have mine).  If you then generate SRPMs (`rpmbuild -bs`) you can do a chain build to test the foreman build, followed by the plugin that depends on the macros from the first:

    mockchain -r el7-scl-foreman -m --yum foreman-1.8.0-0.develop.201501141458gitd43379f.el7.src.rpm rubygem-foreman_bootdisk-4.0.2-1.git.188.6c277bf.fc21.src.rpm

As mentioned in the inline comment above, we could perhaps put the foreman-sqlite dependency on foreman-assets and keep macros in that file, or put it all in foreman-plugin/foreman-build and have plugins depend on that instead?

foreman/foreman.spec
domcleal: Another thought, if we want Foreman core to always win, we could have plugins %exclude the conflicting files in their file list.

mbacovsky: The idea was that plugin provides docs for resource only if it overrides the core. As it depednds on the core it should be loaded later and win. It is true that with multiple overriding plugins there no rule (is this situation covered on route level?) There might be also  problem with core update.

We can move the copying to the rake task perhaps. That would solve the problem with update and we have access to Rails there if we have to figure out the plugin ordering. What do you think?

domcleal: I feared you'd say it was that way around :)  Yeah, we need to sort this in core or apipie-rails then, rather than packaging, as we need to be able to handle core-without-plugin updates reliably.  I've no idea about routes, probably depends on how the plugin is written.

One idea before we go the copying-in-rake route.. today, we serve up precompiled assets from multiple plugin directories: https://github.com/theforeman/foreman/blob/1.7.1/config/environments/production.rb#L140-L147

Would it be possible to serve up static apipie files in the same way - if we can order the plugins above core?  Then we could build the index of all the plugins, and even avoid copying them out of the plugin package.

If not, then copying in the rake task that builds the index is the next best thing to do.  We would need to support a directory location outside of the gem installation root due to support Debian packaging (same reason as https://github.com/theforeman/foreman/commit/d43379f323c507d19a2e8f051cd2e767be84125e), i.e. stick with /usr/share/foreman/public/apipie-cache/plugin/%{plugin-name}.

domcleal: I'm just going to test out that idea..

domcleal: That idea doesn't work, as we're actually going through an apipie-rails controller rather than statically serving files, so I think we'll have to copy and combine files in the rake task into the cache dir.

foreman/foreman.spec
mbacovsky: Ahh I see now. I'd prefer to merge assets and apipie macros to macros.foreman-plugin and ship that in plugin subpackage with the sqlite dependency. I'll update the PR that way if there are no objections @domcleal.

domcleal: Yep, I like that too.

domcleal: Oh, don't forget to update comps/ when adding a sub-package, so it's shipped.

mbacovsky:
The PR was rebased and updated. I've added the changes in separate TMP commits and will squash them later if they are found okay. Hopefully it makes the review of the changes easier.

- moved the copying of the cache to the rake task so it requires https://github.com/theforeman/foreman/pull/2104
- moved the macros and sqlite dependency to the foreman-plugin subpacakge and updated the dependences.
- removed the core apipie cache from public/apipie-cache/plugin/foreman as it seems unnecessary with current changes

@domcleal thatnks for the mocchain howto, it helped with the testbuilds significantly. Currently there are problems only with bastion and discovery. I'm investigating that, but it seems to be some local issue 
unrelated to my changes. Everything else builds fine.

mbacovsky:
Updated, discovery was missing new build dep and bastion builds okay after rebase. So all packages addressed with this PR build okay now.

foreman/foreman.spec
```diff
@@ -656,7 +678,7 @@ exit 0
 # always attempt to reencrypt after update in case new fields can be encrypted
 %{foreman_rake} db:migrate db:compute_resources:encrypt >> %{_localstatedir}/log/%{name}/db_migrate.log 2>&1 || :
 %{foreman_rake} db:seed >> %{_localstatedir}/log/%{name}/db_seed.log 2>&1 || :
-%{foreman_rake} apipie:cache >> %{_localstatedir}/log/%{name}/apipie_cache.log 2>&1 || :
+%{foreman_rake} apipie:cache:index >> %{_localstatedir}/log/%{name}/apipie_cache.log 2>&1 || :
```
domcleal: The index call will be needed in the plugin %posttrans too, in case they're installed after Foreman.  Maybe update `%%%{name}_apipie_cache` in foreman.spec to do `:index` and use that?

foreman/foreman.spec
```diff
@@ -325,7 +325,19 @@ Meta package to install asset pipeline support.
 
 %files assets
 %{_datadir}/%{name}/bundler.d/assets.rb
-%{_sysconfdir}/rpm/macros.%{name}-assets
+
+%package plugin
+Summary: Foreman plugin support
+Group: Applications/system
```
domcleal: Nitpick: Development/Libraries perhaps?

rubygem-foreman_bootdisk/rubygem-foreman_bootdisk.spec
```diff
@@ -42,6 +42,8 @@ BuildRequires: %{?scl_prefix}ruby(release)
 %else
 BuildRequires: %{?scl_prefix}ruby(abi) >= %{rubyabi}
 %endif
+BuildRequires: foreman >= 1.8.0
```
domcleal: unnecessary?

rubygem-foreman_discovery/rubygem-foreman_discovery.spec
```diff
@@ -49,6 +49,9 @@ BuildRequires: %{?scl_prefix}ruby(release)
 %else
 BuildRequires: %{?scl_prefix}ruby(abi) >= %{rubyabi}
 %endif
+BuildRequires: foreman >= 1.8.0
```
domcleal: unnecessary?

domcleal:
I like the look of this a lot, good work!  Once the %posttrans bit is fixed, I'll take a look at, and test the built packages.

rubygem-foreman_bootdisk/rubygem-foreman_bootdisk.spec
mbacovsky: Yes, indeed

foreman/foreman.spec
mbacovsky: I'm curious what happens when foreman and 3 plugins all with same :index in posttrans are installed. Will it trigger the :index four times?

domcleal: Yep.. I think that's unavoidable for now.  Apart from the rake/app load speed, it's reasonably fast.

mbacovsky: Okay, I'll fix that.

mbacovsky:
Updated. I moved %foreman_apipie_cache to the foreman-plugin and let it build just the index. Posttrans in bootdisk and discovery were updated. The index is being rebuild multiple times when installing more packages with cache, but one run takes about 26sec on my setup so it could be acceptable.

@domcleal, please let me know when you want me to squash the commits.

mbacovsky:
This patch to foreman_bootdisk https://github.com/theforeman/foreman_bootdisk/pull/6 needs to be merged before the changes of packaging.

foreman/foreman.spec
```diff
@@ -424,6 +436,8 @@ cp config/settings.yaml.example config/settings.yaml
 export BUNDLER_EXT_NOSTRICT=1
 export BUNDLER_EXT_GROUPS="default assets"
 %{scl_rake} assets:precompile:all RAILS_ENV=production --trace
+%{scl_rake} db:migrate RAILS_ENV=production --trace
```
domcleal: I think this also needs to be in the RPM macros so plugins also migrate during build.  The foreman_discovery build fails right now as it can't open the DB while running plugin:apipie:cache.

mbacovsky: @domcleal, I've added the migration to the macro. While building discovery the migration finish successfully, cache is generated, but the build fails due to some other error. Investigating...
```
 error: File not found: /builddir/build/BUILDROOT/ruby193-rubygem-foreman_discovery-2.0.0-1.el7.centos.x86_64/opt/rh/ruby193/root/usr/share/gems/gems/foreman_discovery-2.0.0/public
```

mbacovsky: Just for the record, I was building discovery from 2.0 stable branch and overlooked that apipie initialization patch was removed from the stable branch (remains in the develop)

mbacovsky:
Updated, rebased. 
- migration was added
- in discovery I replaced bundler.d creation with macro

- Does it make sense to run db:seed as well?  I think just Katello plugin needs that though.
- There are some other plugin related macros like %foreman_bundlerd_file left in the macros.foreman. Does it make sense to move them to foreman-plugin?

domcleal:
> Does it make sense to run db:seed as well? I think just Katello plugin needs that though.

Sure.

> There are some other plugin related macros like %foreman_bundlerd_file left in the macros.foreman. Does it make sense to move them to foreman-plugin?

Yeah, good idea.

mbacovsky:
@domcleal, updated

domcleal:
I'm struggling to get the db:migrate to work with foreman_bootdisk (.gem generated from master branch) as ./usr/share/foreman/db is a symlink to /var/lib/foreman/db in the build environment, and it doesn't have permission to create the sqlite database.

I'm trying to copy /var/lib/foreman/db to ./var and then change the symlink, but haven't quite got it to work yet.  I wonder if you encountered the same?

mbacovsky:
@domcleal, not sure if I understand when the problem occures. My chain build is not failing and the db:merge proceeds without complaints. Here is a link to a build log (a couple days old Foreman develop and fresh foreman_bootdisk gem generated from master branch) http://fpaste.org/184611/39405142/

domcleal:
Thanks, I'll dig into it a bit more.  Here's the build log for reference: http://paste.fedoraproject.org/184624/23740802/

mbacovsky:
@domcleal, I've updated again. I was able to build foreman and foreman_bootdisk as non-root with these specs. There are wierd errors in the build log while db:seed runs. The seed seems to finish okay though. Running it for second time does not produce the errors anymore. Investigating...

mbacovsky:
@domcleal, updated. Even with @iNecas's help we didn't found out in reasonable time why the roles/permissions seed is failing when ran as non-root. But it seems that the seeds are not necessary to build the cache so I removed them. It seems I was mistaken and even Katello does not need preseeded database to generate the apipie cache, so it should be safe. 
My test builds are looking good now.

domcleal:
@mbacovsky great, thanks.  Can you just do a git rebase please as there's a merge error?

mbacovsky:
Rebased

domcleal:
Merged as 53c3308..9e446af.  Nice one, thanks @mbacovsky!

mbacovsky:
Thanks a 1e6 for your patience, @domcleal !!!  I owe you some well deserved :beers:

lzap:
Guys, this patch changed dependency from >= 1.7 to >= 1.8 which is not yet true and plugin won't load.

rubygem-foreman_discovery/rubygem-foreman_discovery.spec
```diff
@@ -34,7 +34,7 @@ License:    GPLv3
 URL:        http://github.com/theforeman/foreman_discovery
 Source0:    http://rubygems.org/downloads/%{gem_name}-%{version}%{?prever}.gem
 
-Requires:   foreman >= 1.7.0
+Requires:   foreman >= 1.8.0
```
lzap: I mean for Discovery plugin...

domcleal:
@lzap it is true, as the changes to the spec file require 1.8.0.  We'll simply build a second 1.7 compatible version on rpm/1.7 in the same way I did for foreman_docker.

PR #469, opened 2014-12-15
Add rubygem-hammer_cli_foreman_tasks package

komidore64:
into nightly builds


domcleal:
Why closing?  This should be finished off.

PR #474, opened 2014-12-18
Update staypuft to 0.5.9

bcrochet:


domcleal:
Thanks @bcrochet, sorry for the delay.  Building now.

PR #476, opened 2014-12-20
Update rubygem-bastion to 0.1.12

ehelms:
Updating Bastion for the  Foreman 1.7 repositories.


domcleal:
Thanks, though we do this by retagging - now done.

ehelms:
@domcleal thanks -- is that documented somewhere (updating a plugin for a released version)? If not, if you lay out the steps, I'll open a PR to the appropriate place.

domcleal:
I don't think it is - http://projects.theforeman.org/projects/foreman/wiki/RPM_Packaging would probably be the best place for it.  You can retag by running http://ci.theforeman.org/job/packaging_plugin_retag/ or using koji tag-build manually for each OS.

PR #478, opened 2014-12-30
refs #8809 - foreman_salt must depend on deface (RPM)

mmoll:
RPM counterpart


domcleal:
Thanks @mmoll, building now.

PR #485, opened 2015-01-12
Update rubygem-bastion to 0.1.13

ehelms:


domcleal:
Thanks @ehelms, building now.

PR #497, opened 2015-01-16
fixes #8979 - convert hammer packages to SCL

domcleal:
This probably isn't fun to review, as it's mostly spec2scl with a couple of tweaks for conditionals.

Tests will probably fail the hammer_cli* builds themselves as the Jenkins test doesn't work for "core" packages just yet.

TODO:
* [x] convert plugins
* [x] fix plugin deps on ruby(release)
* [x] import and convert net-ssh* dependencies
* [x] powerbar dependencies
* [x] obsoletes
* [x] move configs + bin to root


domcleal:
This should now be ready, I've updated plugins too via spec2scl, added obsoletes, and imported the dependencies that we also needed to convert to SCL (as separate commits).

https://fedorapeople.org/~domcleal/foreman/rpm_hammerscl/ has test repos for each OS, which I tested by doing a `yum install rubygem-hammer_cli*` and then `yum upgrade \*hammer\*`.

domcleal:
Rebased and refreshed the repo with new builds.

mbacovsky:
Looks very well so far. I've tried to install from your pre-built repo (many thanks for this). Everything installed okay including hammer plugins. `hammer` command works well on the system. From the log it runs on Ruby 1.9.3 and the localization is loaded from the right places. I'll try to go through the specs if anything catches my eye. 

@domcleal thanks for all the effort on this!

domcleal:
Thanks for testing, let me know when you're happy with the specs.

mbacovsky:
I checked the sources and everything looks perfect. 
I assume hammer deps specific for ruby 1.8 like fastercsv can be removed once they are removed from hammer sources, right?
Thanks for making development hammer easier! 
LGTM!

domcleal:
> I assume hammer deps specific for ruby 1.8 like fastercsv can be removed once they are removed from hammer sources, right?

Yep, you can probably start dropping 1.8 deps like fastercsv imminently, and remove 1.8 from the test matrix in Jenkins.

We need to make a similar change to this in the Ubuntu 12.04 packaging as that has 1.8.7 by default too, but should be easier I think.

mbacovsky:
Great, will do. Thanks a lot @domcleal!

domcleal:
@ehelms not sure if you've been following this, but note that I'll probably merge and start building this early next week.  You'll need to spec2scl any packages you have depending on these too so they depend on the right version.

ehelms:
@domcleal Thanks for the heads up, my hope is that we are able to get our katello-packaging conversion done and then go through it and update the hammer packages we maintain by mid week (depending on how long new tito takes to make it to Fedora repos).

PR #498, opened 2015-01-18
package hammer_cli_foreman_salt (DEB)

mmoll:
http://ci.theforeman.org/job/packaging_build_deb_dependency/634/ is :green_heart: and the resulting package works for me on wheezy.


domcleal:
Thanks @mmoll, building now!

PR #288, opened 2014-07-24
Fixes #6764 - removed extra/ directory

lzap:
http://koji.katello.org/koji/taskinfo?taskID=132151

PR #500, opened 2015-01-21
foreman-discovery: release 2.0.0.rc1

GregSutcliffe:
No assets for easy release \o/

### DIFF
--- a/plugins/ruby-foreman-discovery/debian/changelog
+++ b/plugins/ruby-foreman-discovery/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-discovery (2.0.0~rc1-1) stable; urgency=low
+
+  * Release 2.0.0 RC1
+
+ -- Greg Sutcliffe <greg.sutcliffe@gmail.com>  Wed, 21 Jan 2015 09:11:00 +0000
+
 ruby-foreman-discovery (1.4.1-2) stable; urgency=low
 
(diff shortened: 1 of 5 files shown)


plugins/ruby-foreman-discovery/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-foreman-discovery (2.0.0~rc1-1) stable; urgency=low
+
+  * Set HOME to ~foreman as rb-readline requires it
```
domcleal: update this

domcleal:
postinst needs db:migrate, db:seed and apipie:cache rake tasks adding.

plugins/ruby-foreman-discovery/debian/changelog
GregSutcliffe: Whoops, thanks

domcleal:
Is this also for deb/1.7, like the RPMs? (https://github.com/theforeman/foreman-packaging/pull/288#issuecomment-70805889)

GregSutcliffe:
Indeed. Technically according to the engine.rb it's for 1.6 too, but no one has checked if it works ;)

domcleal:
Thanks @GregSutcliffe, merged also to deb/1.7 as 4b8bef8.  I'll build it now.

PR #501, opened 2015-01-21
Staypuft - removed discovery dependency < 1.4.0

lzap:
Assuming this is compatible.


domcleal:
This doesn't help - the gem depends on this version, it needs to be fixed there first.

domcleal:
Sorry, workrave forced me into a rushed reply!  The plugin will fail to load at runtime if it's paired with foreman_discovery 2.0, as the gemspec specifies this version and rubygems will interpret it and raise an error.  This is why I've been pushing for precise versions in spec files to mirror the gemspec, to catch these problems closer to build/publish time.

I can't see much direct use of discovery in staypuft, so hopefully it won't be hard to relax the version specification there.

lzap:
The gem is up. Updated. https://rubygems.org/gems/staypuft

domcleal:
Thanks @lzap!  Building now.

PR #502, opened 2015-01-23
Update staypuft to 0.5.16

sseago:
Scratch builds:
http://koji.katello.org/koji/taskinfo?taskID=214725
http://koji.katello.org/koji/taskinfo?taskID=214727
http://koji.katello.org/koji/taskinfo?taskID=214728


domcleal:
Thanks @sseago, building into nightlies.

PR #504, opened 2015-01-25
package dynflow and foreman-tasks (DEB)

mmoll:
This is nowhere near perfection, but more or less working. @ares wrote a debianized version of the init script, which hangs on my wheezy box when starting, but is looking correct otherwise. Asset precompilation is not yet included.

Discuss! :boom:


domcleal:
Do you think the separate dynflow package is needed, or could it just be one of the deps inside ruby-foreman-foreman-tasks?

plugins/ruby-foreman-foreman-tasks/debian/init.d
```diff
@@ -0,0 +1,274 @@
...
+FOREMAN_TASK_PARAMS=${FOREMAN_TASK_PARAMS:--p foreman}
+FOREMAN_PID_DIR=$FOREMAN_HOME/tmp/pids
+FOREMAN_LOG_DIR=${FOREMAN_LOG_DIR:-/var/log/foreman}
+DAEMON="/usr/bin/foreman-tasks"
+PID_FILE=$FOREMAN_PID_DIR/dynflow_executor_monitor.pid
+export RAILS_ENV=$FOREMAN_ENV
+export FOREMAN_LOGGING=${FOREMAN_LOGGING:-warn}
+export FOREMAN_LOGGING_SQL=${FOREMAN_LOGGING_SQL:-warn}
+export RAILS_RELATIVE_URL_ROOT=$FOREMAN_PREFIX
+export BUNDLER_EXT_GROUPS
```
domcleal: Probably best to remove all references to BUNDLER_EXT, just to ensure it doesn't activate.

mmoll:
We could do that, but IMHO it's good to have it seperated, also because of MIT vs. GPLv3+ licensing.

domcleal:
Okay, that's fine, I don't mind much either way.  We really should drop the second "foreman" from the foreman-tasks package name though.  I think I see why you did it, but other plugins don't do it :smile: 

The package seems to work fine.  The hanging init script is "normal" (FSVO..) if no plugins require dynflow.. I think that should be improved, but it's in foreman-tasks itself.  It eventually times out internally it seems and no longer attempts to start the executor.

Used with foreman_chef 0.1.0 it appears to start up just fine, once a small change is made to the initialiser (noted in that PR).

So perhaps just add the asset precompilation and rename?

mmoll:
Should be fine now, just running into a missing plugin repo when building now:
http://ci.theforeman.org/job/packaging_build_deb_plugin/367/label=debian/console

domcleal:
I think it's failing because ruby-foreman-dynflow isn't there in the repo... I'll merge and build that, then re-examine the foreman-tasks bit of the PR.

domcleal:
Turns out it was along those lines - it didn't use either your scratch plugins repo, nor the official plugin repo during the build, so I've added both to the job.  Seems to be getting further.

plugins/ruby-foreman-tasks/debian/control
```diff
@@ -0,0 +1,13 @@
...
+Section: ruby
+Priority: extra
+Maintainer: Michael Moll <mmoll@mmoll.at>
+Build-Depends: debhelper, foreman (>= 1.7), ruby-foreman-dynflow
+Standards-Version: 3.9.3
+Homepage: https://github.com/Dynflow/dynflow
+
+Package: ruby-foreman-tasks
+Architecture: all
+Depends: ${misc:Depends}, bundler, foreman (>= 1.7), ruby-foreman-dynflow
```
domcleal: dynflow >= 0.7.2

plugins/ruby-foreman-tasks/debian/control
```diff
@@ -0,0 +1,13 @@
+Source: ruby-foreman-tasks
+Section: ruby
+Priority: extra
+Maintainer: Michael Moll <mmoll@mmoll.at>
+Build-Depends: debhelper, foreman (>= 1.7), ruby-foreman-dynflow
```
domcleal: no need for these build deps now?

plugins/ruby-foreman-tasks/bin/foreman-tasks
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env ruby
```
domcleal: This might be a problem on 12.04, should we use the foreman-ruby symlink we installed as the shebang line?

mmoll:
updated.

domcleal:
Great work @mmoll and @ares, thank you!  Merged as cd3708c and 2a6b2f2.

PR #505, opened 2015-01-26
Update staypuft to version 0.5.17

bcrochet:


domcleal:
Thanks @bcrochet, merged and building now.

I've added you to this repo so you can release package updates more easily, thanks for your consistently good PRs.  If it's just a regular update like this, please feel free to commit directly - after running scratch builds to ensure it still builds (or you can open a PR so Jenkins runs them, then merge it yourself).

As you may be aware, once you've committed the change, then run `tito tag --keep-version` to tag it and update %changelog, do the `git push` that it outputs and then run `kkoji release koji-foreman-plugins` to build.

Double check for any changes in dependencies/gemspec in the release, as we should reflect those in the spec file deps.

If you have any questions, just let me know.  Cheers!

PR #506, opened 2015-01-27
Fix hooks version number to reflect packaged gem

domcleal:
I guess adding a new entry is the correct thing to do here, rather than editing historical entries.

For deb/develop & 1.7 I think.


domcleal:
6aa73a4

PR #508, opened 2015-01-28
Update to dynflow 0.7.6

iNecas:


domcleal:
The uuidtools dep has been removed from the gem, could you remove it from the spec file please?

ares:
just out of curiosity, required ruby version changed in gemspec? otherwise seems good

domcleal:
@ares yeah, I did that when replacing uuidtools with SecureRandom (1.9 only)

ares:
I see, thanks

iNecas:
Ah, you're right. Updated

domcleal:
Thanks for the PR @iNecas, building now.

PR #509, opened 2015-01-28
Update foreman_openscap to 0.3.0

isimluk:
I have checked that it builds on my local RHEL6 box. The scratch build in koji is missing though.


rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -26,7 +26,7 @@ Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 Requires: foreman >= 1.5.0
 
 Requires: %{?scl_prefix}rubygem(deface)
-Requires: %{?scl_prefix}rubygem(scaptimony) >= 0.2.0
+Requires: %{?scl_prefix}rubygem(scaptimony) = 0.3.0
```
domcleal: There shouldn't be a version spec here, the gem requires ~> 0

domcleal:
Could you use git annex add to when committing the gem, instead of committing the binary itself?  There are a few instructions in the README which should help you set it up and show the commands, shout if you get stuck.

isimluk:
Updated new changeset. Thanks for review.

domcleal:
[test]

rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -39,9 +39,6 @@ BuildRequires: %{?scl_prefix}ruby(abi) >= %{rubyabi}
 BuildRequires: %{?scl_prefix}rubygems-devel
 BuildRequires: %{?scl_prefix}rubygems
 BuildRequires: foreman-assets >= 1.7.0
-# TODO could be removed later when dependencies are fixed https://github.com/OpenSCAP/foreman_openscap/issues/67
```
domcleal: I don't think this comment was correct actually, they will still be build-deps for asset precompilation.  The tests should confirm!

isimluk:
Yep, you are right. it fails in my copr https://copr.fedoraproject.org/coprs/isimluk/OpenSCAP/build/69599/

domcleal:
Okay, if you could revert those lines then we'll re-test.

isimluk:
Ok, reverted, but patch is still not ready. The assets precompile does not create assets in public/assets/foreman_openscap directory.

```
RPM build errors:
    File not found: /root/rpmbuild/BUILDROOT/ruby193-rubygem-foreman_openscap-0.3.0-1.el6.x86_64/opt/rh/ruby193/root/usr/share/gems/gems/foreman_openscap-0.3.0/public/assets/foreman_openscap
    File not found: /root/rpmbuild/BUILDROOT/ruby193-rubygem-foreman_openscap-0.3.0-1.el6.x86_64/opt/rh/ruby193/root/usr/share/gems/gems/foreman_openscap-0.3.0/public/assets/foreman_openscap
```

Also, I have decided to put SCAPtimony version requirements here. It will not work with other versions anyway. (Fixing in upstream as well OpenSCAP/foreman_openscap#72).

domcleal:
Thanks for updating the version requirements in the gemspec too, it's important they stay in sync otherwise it can cause runtime errors.

It seems to build in mock just fine, so let's see if the PR [test] on Koji works.

domcleal:
Looks fine to me, did you want to investigate more or are you happy for me to merge this?

isimluk:
Please merge as it builds in koji. I will investigate issue with my setup on my own. Thanks!

rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -105,13 +106,14 @@ mkdir -p %{buildroot}%{foreman_bundlerd_dir}
 
 %posttrans
 # We need to run the db:migrate (because of SCAPtimony) after the install transaction
-/usr/sbin/foreman-rake db:migrate  >/dev/null 2>&1 || :
-/usr/sbin/foreman-rake db:seed  >/dev/null 2>&1 || :
```
domcleal: You need to keep db:seed as there's a seed file in the plugin

rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -105,13 +106,14 @@ mkdir -p %{buildroot}%{foreman_bundlerd_dir}
...
-/usr/sbin/foreman-rake apipie:cache  >/dev/null 2>&1 || :
```
domcleal: ditto apipie:cache as there's an API controller

rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -105,13 +106,14 @@ mkdir -p %{buildroot}%{foreman_bundlerd_dir}
...
-(/sbin/service foreman status && /sbin/service foreman restart) >/dev/null 2>&1
+%foreman_db_migrate
```
domcleal: but no DB migrations!

domcleal: Ack, cancel that.. there are in scaptimony aren't there?  So leave this line in, but the others need to stay, or be replaced by macros if you like.

domcleal:
Thanks @isimluk!  Will build it now.

PR #522, opened 2015-02-07
fixes #9156 - package gssapi rubygem

mmoll:
http://ci.theforeman.org/job/packaging_build_deb_dependency/637/ is :green_heart:, the commit ID is different because I amended the commit message to include the issue ID.

For the user who originally had the problem, it was fixed after he ran gem2deb on gssapi and installed it. I'm not sure about the proxy dependency/recommendation, but as long as the proxy modules are not packaged seperately, I feel this is the best way.

Could also be cherry-picked into deb/1.7.


domcleal:
Nice one, thanks @mmoll!

I think you're right about the dep, and that rkerberos is in the wrong place.

PR #525, opened 2015-02-09
Remove angular-rails-templates version dependency on rubygem-bastion

ehelms:


domcleal:
This will cause a runtime error as the gemspec still depends on ~> 0.0.4, you need to release a new gem with that fixed first.

domcleal:
(This is the reason I'm enforcing identical RPM requires to match the gem, because rubygems will evaluate the gemspecs when the gem is loaded and throws an error if they're unsatisfied, so this allows us to catch the errors earlier.)

rubygem-bastion/rubygem-bastion.spec
```diff
@@ -26,16 +26,14 @@ BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 
 Requires: foreman >= 1.6.0
-Requires: %{?scl_prefix}rubygem(angular-rails-templates) >= 0.0.4
```
domcleal: Please keep the version specs matching the new gem (~> 0.1.3)

domcleal:
:+1: otherwise, please go ahead and build once that's fixed and the PR is green.  (I'm off for the day.)

rubygem-bastion/rubygem-bastion.spec
```diff
@@ -26,16 +26,16 @@ BuildArch: noarch
...
-Requires: %{?scl_prefix}rubygem(angular-rails-templates) < 0.1.0
+BuildRequires: %{?scl_prefix}rubygem(angular-rails-templates) >= 0.1.3
```
domcleal: Requires, not BR

PR #528, opened 2015-02-10
plugins:foreman_docker - Release 1.1.0

eLobato:


domcleal:
Which releases is this and #528 to be built for?

domcleal:
Thanks @eLobato, building for nightlies + 1.7.

PR #532, opened 2015-02-11
Update rubygem-angular-rails-templates to 0.1.2

ehelms:
This is required to fix the breakage I was previously attempting to fix. My original update to 0.1.3 while fixing the issue for me didn't fix it on all machines due to plugin load orders differing across environments. After further exploration, 0.1.3 had a change to the initialization process that led to this inconsistent breakage. The 0.1.2 version does not contain that change and has, to my testing, fixed all of the issues we are experiencing. This will require an update to bastion as well that will follow.


domcleal:
I think you should bump the epoch, otherwise this isn't going to upgrade - unless you don't mind about the window in which people might have installed 0.1.3?

ehelms:
Probably doesn't matter due to our breakages, but I went ahead and added it just to be safe.

domcleal:
Sure, okay.  Setting the epoch means you'll need to put `1:` in front of the versions in bastion's spec update too.  Looks fine once you've got the corresponding bastion release.

ehelms:
Corresponding bastion PR - https://github.com/theforeman/foreman-packaging/pull/535

PR #533, opened 2015-02-11
Fixes #9341 - Updating the docker-api version

parthaa:


domcleal:
Please change this in the gemspec, not the RPM spec file.  On the next release, we'll update the spec to match the gemspec.

PR #534, opened 2015-02-12
Prevented from missing base policy in foreman-selinux

lzap:
We've hit issue downstream when base policy was not loaded in the buildroot 
causing some optional_blocks to be ignored. We should error out early when 
base policy is not loaded in the build root to prevent that, otherwise we will
likely find this in user reported bugs (allow rule is missing), because 
SELinux will silently ignore those blocks.


lzap:
As expected, it turned out that the command fails due to missing BuildRequire. Added. Will wait for my scratchbuild.

lzap:
Scratch that. Optional blocks are runtime things. As long as we won't  delete `Requires(post): selinux-policy-targeted` we are fine :-)

PR #536, opened 2015-02-12
refs #8641 - update fog to 1.27, introduce foreman-ec2

domcleal:
Couple of key things to review here:

* introduction of foreman-ec2, which should be installed when upgrading from Foreman 1.7.2 with foreman-compute installed
* usual review of new package, rubygem-fog-aws

The rest should be business as usual.


lzap:
So we are dropping `foreman-compute` completely and only provide particular compute resources as packages.

domcleal:
Not yet, foreman-compute still provides OpenStack and Rackspace support for now.  I think in a few versions time it might be replaced by specific gems and packages, yep.

lzap:
Ooh now I get it. Okay.

domcleal:
Yeah, the "Obsoletes" is a bit strange but it seems to upgrade correctly from 1.7 and install both foreman-compute 1.8 and foreman-ec2 1.8.

lzap:
No issues, shell I initiate the builds?

domcleal:
No, I can.  Thanks @lzap!

PR #541, opened 2015-02-13
Introduced foreman-debug subpackage

lzap:
WIP will try to build few times :-)

### DIFF
--- a/debian/precise/foreman/foreman-debug.dirs
+++ b/debian/precise/foreman/foreman-debug.dirs
@@ -1,0 +1,1 @@
+usr/share/foreman/script/foreman-debug.d/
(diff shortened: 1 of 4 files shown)


debian/precise/foreman/foreman-debug.dirs
```diff
@@ -0,0 +1 @@
+usr/share/foreman/script/foreman-debug.d/
```
GregSutcliffe: Whats this for?

lzap: Katello and plugins use that to drop "extensions" to the foreman-debug to collect more information. Let's move the review to the new PR, had to change my branch name.

PR #542, opened 2015-02-13
Introduced foreman-debug subpackage

lzap:
Replaces https://github.com/theforeman/foreman-packaging/pull/541

### DIFF
--- a/debian/precise/foreman/foreman-debug.dirs
+++ b/debian/precise/foreman/foreman-debug.dirs
@@ -1,0 +1,1 @@
+usr/share/foreman/script/foreman-debug.d/
--- a/debian/precise/foreman-proxy/control
+++ b/debian/precise/foreman-proxy/control
@@ -11,7 +11,7 @@ Vcs-Browser: https://github.com/ohadlevy/smart-proxy
 Package: foreman-proxy
 Architecture: all
 Depends: ruby1.9.1|ruby-interpreter, rake, ruby-sinatra, ruby-rack (>= 1.1.0), ruby-json, ruby-rkerberos (>= 0.1.1), libaugeas-ruby, ruby-bundler-ext
-Recommends: sudo, wget, ruby-gssapi
+Recommends: sudo, wget, ruby-gssapi, foreman-debug
 Description: RESTful proxies for DNS, DHCP, TFTP, and Puppet
  Smart-Proxy is a project which provides a RESTful API to various sub-systems
  like DNS, DHCP, TFTP, and Puppet.
--- a/debian/precise/foreman-proxy/install
+++ b/debian/precise/foreman-proxy/install
@@ -13,3 +13,4 @@ tasks                     usr/share/foreman-proxy
 VERSION                   usr/share/foreman-proxy
 views                     usr/share/foreman-proxy
 foreman-proxy             var/cache
+extra/foreman-debug-proxy usr/share/foreman/script/foreman-debug.d
--- a/debian/precise/foreman/control
+++ b/debian/precise/foreman/control
@@ -16,7 +16,7 @@ Priority: extra
 Depends: ruby1.9.3 (>= 1:1.9.3.545), bundler (>= 1.2), ruby1.9.1-dev (>= 1:1.9.3.545),
          libxml2-dev, libxslt1-dev, facter (>= 1.6.12-1puppetlabs2), build-essential,
          rake (>=0.8.4), ${misc:Depends}
-Recommends: foreman-proxy
+Recommends: foreman-proxy, foreman-debug
 Description: Systems management web interface
  Foreman is aimed to be a single address for all machines life cycle management.
  .
@@ -186,3 +186,13 @@ Description: metapackage providing GCE dependencies for Foreman
  a flexible systems management web application.
  .
  The foreman package should be installed along with this package.
+
+Package: foreman-debug
+Architecture: all
+Section: ruby
+Priority: extra
+Depends: rsync
+Description: provides support utility foreman-debug.
+ .
+ The foreman package should be installed along with this package.
+
--- a/debian/precise/foreman/foreman-debug.install
+++ b/debian/precise/foreman/foreman-debug.install
@@ -1,0 +1,1 @@
+script/foreman-debug                     usr/sbin
--- a/debian/precise/foreman/foreman.install
+++ b/debian/precise/foreman/foreman.install
@@ -2,7 +2,6 @@
 config/email.yaml.example                etc/foreman
 debian/settings.yaml                     etc/foreman
 debian/database.yml                      etc/foreman
-script/foreman-debug                     usr/sbin
 script/foreman-rake                      usr/sbin
 script/foreman-tail                      usr/sbin
 app                                      usr/share/foreman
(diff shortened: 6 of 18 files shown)


debian/precise/foreman/foreman-debug.dirs
```diff
@@ -0,0 +1 @@
+usr/share/foreman/script/foreman-debug.d/
```
GregSutcliffe: (Just to keep the discussion from the old PR) What's this for? :)

lzap: It's an extension point, example plugin: https://github.com/Katello/katello/blob/master/deploy/script/katello-debug.sh

domcleal: Or https://github.com/theforeman/smart-proxy/blob/develop/extra/foreman-debug-proxy

lzap:
Build done, where do I get the deb files? http://ci.theforeman.org/view/Packaging/job/packaging_build_deb_coreproject/1647/

domcleal:
Don't forget to add the foreman{,-proxy} deps on foreman-debug.

domcleal:
(Maybe Recommends?)

domcleal:
And ensure that foreman-proxy's debug.d file is placed in ~foreman's debug.d dir.

debian/precise/foreman/foreman-debug.dirs
GregSutcliffe: This should be part of the main package if the proxy is going to drop files into it.

lzap:
Adding hard dependency, that's better (users will ask where `foreman-debug` actually is otherwise).

Will amend the foreman-proxy change in this patch as well.

domcleal:
Recommends will be installed by default though, but allows users to remove the package to create a minimal installation.

debian/precise/foreman/foreman-debug.dirs
domcleal: In the RPMs, foreman-proxy depends on foreman-debug so the package would exist.  Even if it's a recommends and foreman-debug's not installed, will the proxy/plugin package install the file OK?  (It'd just be unused.)

lzap:
Oh okay, will change that.

lzap:
Amended incl. the proxy change (man that was close to reformatting). Triggering the build.

lzap:
http://ci.theforeman.org/view/Packaging/job/packaging_build_deb_coreproject/1648/

domcleal:
And add foreman-debug to foreman-proxy's recommends?

lzap:
*facepalm*

lzap:
http://ci.theforeman.org/view/Packaging/job/packaging_build_deb_coreproject/1649/

domcleal:
Looks good.  We should test the installation without foreman-debug just to confirm the inline discussion above.

lzap:
Problem A: `apt-get upgrade` did not install recommended package for me.

Problem B: `foreman-debug` package does not contain any files :-)

```
root@fwheezy:~# dpkg -L foreman-debug | grep debug
/usr/share/doc/foreman-debug
/usr/share/doc/foreman-debug/changelog.Debian.gz
/usr/share/doc/foreman-debug/copyright
/usr/share/doc/foreman-debug/changelog.gz
```

lzap:
I left the grep in the pipeline but there is nothing else left.

I don't belive I've installed correct build.

lzap:
Hmmm it really looks like the version is correct. http://ci.theforeman.org/job/packaging_build_deb_coreproject/label=debian,os=wheezy/1650/console

    Setting up foreman-debug (9999-wheezy+scratchbuild+201502131319) ...

Help :-)

domcleal:
foreman-debug.{dirs,install} appear to be outside of the 'foreman' subdirectory for trusty/wheezy, but appear to be in the correct place for precise.

debian/precise/foreman/foreman-debug.dirs
GregSutcliffe: I'm not 100% sure that an `install` entry without a `dirs` entry will work as described, but it's fairly likely. we can test it :)

GregSutcliffe:
Agreed, your dirs are wrong. We'll also need scratchbuilds of foreman-proxy to check the debug dir is written correctly - I only see foreman scratchbuilds at the  moment.

lzap:
http://ci.theforeman.org/view/Packaging/job/packaging_build_deb_coreproject/1660/
http://ci.theforeman.org/view/Packaging/job/packaging_build_deb_coreproject/1661/

domcleal:
http://ci.theforeman.org/view/Packaging/job/packaging_build_deb_coreproject/1664/ is the right build for the proxy, replaces the second link.

lzap:
I think we are fine, it works. Uploaded a tarball actually to investigate if all things got collected. One issue tho (out of scope for this?) - the foreman-debug is installed twice - sbin and then script directories. We should either do a symlink or remove the latter.

```
root@fwheezy:~# dpkg -L foreman | grep debug
/usr/share/foreman/script/foreman-debug
/usr/share/foreman/lib/foreman/gettext/debug.rb
root@fwheezy:~# dpkg -L foreman-debug | grep debug
/usr/share/doc/foreman-debug
/usr/share/doc/foreman-debug/changelog.Debian.gz
/usr/share/doc/foreman-debug/copyright
/usr/share/doc/foreman-debug/changelog.gz
/usr/share/foreman/script/foreman-debug.d
/usr/sbin/foreman-debug
root@fwheezy:~# dpkg -L foreman-proxy | grep debug
/usr/share/foreman/script/foreman-debug.d
/usr/share/foreman/script/foreman-debug.d/foreman-debug-proxy
root@fwheezy:~# ls /usr/sbin/foreman-debug -la
-rwxr-xr-x 1 root root 9103 Feb 17 04:54 /usr/sbin/foreman-debug
root@fwheezy:~# ls /usr/share/foreman/script/foreman-debug -la
-rwxr-xr-x 1 foreman foreman 9103 Feb 17 04:54 /usr/share/foreman/script/foreman-debug
root@fwheezy:~# diff /usr/sbin/foreman-debug /usr/share/foreman/script/foreman-debug
root@fwheezy:~# foreman-debug


 HOSTNAME: fwheezy.zzz.lan
       OS: debian
  RELEASE: 7.8
  FOREMAN: 1.8.0-develop
     RUBY: ruby 1.9.3p194 (2012-04-20 revision 35410) [x86_64-linux]
   PUPPET: 2.7.23


A debug file has been created: /tmp/foreman-debug-QaA48.tar.xz (301484 bytes)

You may want to upload the tarball to our public server via rsync. There is a
write only directory (readable only by Foreman core developers) for that. Note
the rsync transmission is UNENCRYPTED:

  rsync /tmp/foreman-debug-QaA48.tar.xz rsync://theforeman.org/debug-incoming

Do you want to do this now? [y/N] y
Uploading...

The tarball has been uploaded, please contact us on mailing list or IRC.
```

domcleal:
@lzap all tested & works great, except for the proxy part.  Looks like the RPMs have an identical problem I missed when reviewing that.

https://github.com/theforeman/foreman/blob/develop/script/foreman-debug#L265 checks for *.sh extension files, but names in both the deb and RPM don't have a .sh extension.

lzap:
Oh I want to fix that in the script itself and include all the executable files and fix this in foreman-proxy git repo (the file currently does not have proper permissions).

lzap:
Proxy fix: https://github.com/theforeman/smart-proxy/pull/259

lzap:
Foreman core fix: https://github.com/theforeman/foreman/pull/2178

debian/precise/foreman/foreman-debug.dirs
domcleal: Tests fine for me.

domcleal:
Great, thanks for implementing this @lzap!  One comment remains on the core fix, but altogether this works fine.

PR #546, opened 2015-02-22
Adds define ruby(abi) to openscap specs

shlomizadok:


rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
```diff
@@ -1,6 +1,8 @@
 %global gem_name foreman_scap_client
 %global config_dir %{_sysconfdir}/%{gem_name}
 
+%define rubyabi 1.9.1
```
domcleal: This would be 1.8.7 for a non-SCL EL6 package.

rubygem-smart_proxy_openscap/rubygem-smart_proxy_openscap.spec
```diff
@@ -5,6 +5,8 @@
 %global spool_dir %{_var}/spool/foreman-proxy/openscap
 %global proxy_user foreman-proxy
 
+%define rubyabi 1.9.1
```
domcleal: Unused.

rubygem-openscap/rubygem-openscap.spec
```diff
@@ -10,6 +10,8 @@
 
 %global gem_name openscap
 
+%define rubyabi 1.9.1
```
domcleal: Unused.

rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
domcleal: If you're not going to add the definition, then maybe remove the usage of `%{rubyabi}` instead?

domcleal:
re rubygem-foreman_scap_client.spec, if you're not going to add the definition, then maybe remove the usage of %{rubyabi} instead?

shlomizadok:
re rubygem-foreman_scap_client.spec, added the definition as it seems to be the standard

shlomizadok:
squashed. tests passed.

domcleal:
Thanks!

PR #547, opened 2015-02-22
Update fog-core to 1.29.0

eLobato:


domcleal:
Thanks @eLobato!  Building now.

PR #551, opened 2015-02-23
Update package rubygem-bastion to 0.2.7

ehelms:


domcleal:
Thanks @ehelms, building!

PR #552, opened 2015-02-24
plugins:foreman_docker - Release 1.2.0

eLobato:
For 1.7 and nightly

https://rubygems.org/gems/foreman_docker/versions/1.2.0

rubygem-docker-api is already packaged as 1.17 - f459b6e3cf385f8426daa37a63c6fac39d7c7839


domcleal:
I need to have a little think about how to do this in 1.7, as our usual method of retagging back to the older release won't work here due to the API i18n support changes merged the other day (the 1.1.0-2 build).  It actually needs to be branched and rebuilt for 1.7, but with a different NEVRA.  Maybe we can add a little tag to the rpm/1.7 branch release or dist tag...

domcleal:
Thanks @eLobato, I'm going to insert a Foreman version into the release field I think and build it on rpm/1.7 and rpm/develop separately.

I may try adding a %{foreman_dist} universally into our release tags soon, it might give us a way to circumvent this problem automatically.

PR #561, opened 2015-03-02
plugins:foreman_docker - Release 1.2.1

eLobato:


domcleal:
Thanks @eLobato, building for rpm/develop and 1.7.

PR #565, opened 2015-03-03
update foreman_salt to 2.0.1 (DEB)

mmoll:
http://ci.theforeman.org/job/packaging_build_deb_plugin/409/ is :green_heart: and the resulting package works for me.

Should also go into deb/1.8.


domcleal:
Thanks @mmoll!  Also on deb/1.8, building now.

PR #566, opened 2015-03-03
update smart_proxy_salt to 2.0.0 (DEB)

mmoll:
http://ci.theforeman.org/job/packaging_build_deb_proxy_plugin/27/ is :rotating_light:, seems like an error in the Jenkins job, if anybody could check that... :sleepy:

Should also go into deb/1.8

### DIFF
--- a/plugins/smart_proxy_salt/changelog
+++ b/plugins/smart_proxy_salt/changelog
@@ -1,10 +1,16 @@
+ruby-smart-proxy-salt (2.0.0-1) unstable; urgency=low
+
+  * Update to 2.0.0
+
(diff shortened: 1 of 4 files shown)


domcleal:
http://ci.theforeman.org/job/packaging_build_deb_proxy_plugin/29/ is now green.  Seems there was a bug in the job (using `${os}` which for some reason is now defined, and even more strangely, as 'el7').  Also the plugin name needed to be "smart_proxy_salt" to match the dir name.

domcleal:
Merged as 63e16ea, 558d18b and building.  Thanks @mmoll!

PR #567, opened 2015-03-03
update hammer_cli_foreman_salt to 0.0.4 (DEB)

mmoll:
http://ci.theforeman.org/job/packaging_build_deb_dependency/640/ is :green_heart: and the resulting package is working for me on Debian/wheezy.

Should also go into deb/1.8


domcleal:
Thanks @mmoll!  Also on deb/1.8, building now.

PR #568, opened 2015-03-04
refs #4478 - precompile API docs [deb]

domcleal:
Both Foreman and plugins that ship API controllers now have API docs for their resources generated at build time, and use `rake apipie:cache:index` to rebuild the resource indexes post-install.

Previously we used `apipie:cache` which also built individual resource pages, however #4478 changed core to build docs for each locale, which made apipie:cache very slow (minutes) during installation.  This PR moves the hard work to build time, so installation of Foreman or a single plugin is now running a very quick index rake task.

The code's near-identical to asset precompilation (as you'll see in ruby-foreman-docker), albeit with an extra db:migrate step (necessary, sorry).  I included a bootdisk 5.0.0 update as it uses the new plugin specification so we can build its API docs.

---

Builds:
* http://ci.theforeman.org/job/packaging_build_deb_coreproject/1724/ :white_check_mark: 
* http://ci.theforeman.org/job/packaging_build_deb_plugin/413/ bootdisk :white_check_mark: 
* http://ci.theforeman.org/job/packaging_build_deb_plugin/418/ docker :white_check_mark: 
* http://ci.theforeman.org/job/packaging_build_deb_plugin/419/ salt :white_check_mark:

### DIFF
--- a/debian/precise/foreman/foreman.postinst
+++ b/debian/precise/foreman/foreman.postinst
@@ -80,9 +80,9 @@ fi
 
 # Generate apipie cache
 if [ ! -z "${DEBUG}" ]; then
-  /usr/sbin/foreman-rake apipie:cache || true
+  /usr/sbin/foreman-rake apipie:cache:index || true
 else
-  /usr/sbin/foreman-rake apipie:cache >> $LOGFILE 2>&1 || true
+  /usr/sbin/foreman-rake apipie:cache:index >> $LOGFILE 2>&1 || true
 fi
 
 # Generate a static session token for signing cookies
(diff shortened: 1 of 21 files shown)


GregSutcliffe:
Packages installed fine, and i see `/var/lib/foreman/public/apipie-cache/plugin/foreman_bootdisk/apidoc/v2` on the filesystem. Not sure what else to check, so I'll give a tentative :+1:

domcleal:
@GregSutcliffe you should also see /var/lib/foreman/public/apipie-cache/apidoc/v2/disks* (for bootdisk, off the top of my head) which is created when the index is run during install, i.e. it combines the Foreman package's API docs, plus the plugin API docs.  You'll also see them in the web browser at /apidocs in one combined index.

GregSutcliffe:
:+1: then, there's a minor issue in core, but the docs are indeed present at /apidoc.

domcleal:
Thanks for the testing, merged.  I intend to pull this into 1.8 for consistency.

PR #569, opened 2015-03-04
Fix wrong file permission in the rubygem-scaptimony

isimluk:
Addressing:
0664 permssions from upstream repository. Upstream will fix it in next
release.

### DIFF
--- a/rubygem-scaptimony/rubygem-scaptimony.spec
+++ b/rubygem-scaptimony/rubygem-scaptimony.spec
@@ -14,7 +14,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.3.1
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: SCAPtimony is SCAP database and storage server
 Group: Applications/System
 License: GPLv3
@@ -72,7 +72,7 @@ gem install --local --install-dir .%{gem_dir} \
 
 %install
 mkdir -p %{buildroot}%{gem_dir}
-cp -a .%{gem_dir}/* \
+cp -dr --no-preserve=mode .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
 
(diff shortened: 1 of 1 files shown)


domcleal:
Which files do you mean?  Usually files are installed as root:root anyway, so it shouldn't make any difference, and I can't see any non-root owned files here.

0664 and 0644 in gems is a very common problem unfortunately, I regularly see gems flip-flop between modes on each release depending on the maintainer's host's configuration.

isimluk:
Issue is indeed 066 on files (I have got report at rhbz#1198351).

I am glad it is common. I was wondering how that could happen.

isimluk:
Hm, looking at the upstream the permissions are correct in the git.

domcleal:
I think git respects the umask (or simply doesn't store permission bits outside of +x), so if the maintainer has the repo checked out then it may be 0644 or 0664 depending on their umask (locally I use a umask of 0002 and my checkout is 0664).  When the maintainer runs gem build (which is just a fancy tarball), then the permissions get preserved.

mmoll:
what's the status here? :)

isimluk:
@mmoll thanks for checking. 

Given the information provided by Dominic, I think this patch should perhaps land in each specfile in this repository.

Based on the downstream ticket (https://bugzilla.redhat.com/show_bug.cgi?id=1198351#c4) however, this issue has perhaps been fixed elsewhere.

mmoll:
as I don't really get where that downstream ticket is really standing, is there anything really to do here? or can this get closed?

domcleal:
Well, this package has since been removed so the patch itself is obsolete.

> Given the information provided by Dominic, I think this patch should perhaps land in each specfile in this repository.

I'm reluctant to do that unless Fedora did, since the spec files generally follow those guidelines and many are imports - I'd rather not deviate.  It'd be useful to get this added to the upstream packaging guidelines (whether that's a change to `cp`, `chmod` or defattr) if it's indeed a problem that matters.

PR #570, opened 2015-03-04
update clamp to 0.6.4 (DEB)

mmoll:
http://ci.theforeman.org/job/packaging_build_deb_dependency/645/ is :green_heart: and the resulting package is working for me on Debian/wheezy.


domcleal:
Thanks @mmoll!  Building here: http://ci.theforeman.org/job/packaging_build_deb_dependency/646/

PR #573, opened 2015-03-10
package hammer_cli_foreman_discovery (DEB)

mmoll:
http://ci.theforeman.org/job/packaging_build_deb_dependency/647/ is :green_heart: and the resulting package works for me on Debian/wheezy. Also good to go for 1.7 and 1.8, IMHO


domcleal:
Nice one, thanks @mmoll!  Would you mind updating the Discovery docs on theforeman.org, as they only mention RPMs.

PR #583, opened 2015-03-17
Foreman tasks 0.6.13 + dynflow 0.7.7 (deb)

iNecas:


plugins/ruby-foreman-tasks/debian/gem.list
```diff
@@ -1,5 +1,5 @@
 # list of gem urls to download via Jenkins, space separated
-GEMS="https://rubygems.org/downloads/foreman-tasks-0.6.12.gem"
+GEMS="https://rubygems.org/downloads/foreman-tasks-0.6.13.gem"
 GEMS="$GEMS https://rubygems.org/downloads/daemons-1.1.9.gem"
 GEMS="$GEMS https://rubygems.org/downloads/sequel-4.16.0.gem"
```
domcleal: could you update daemons/sequel too?

iNecas: Updated

domcleal:
Thanks @iNecas!  Building now.

PR #584, opened 2015-03-17
Update smart_proxy_chef to 0.1.3

ares:


mmoll:
http://ci.theforeman.org/job/packaging_build_deb_proxy_plugin/33/ is :green_heart: - there's some more logic needed to determine which Jenkins job to use for the PR test...

mmoll:
@ares I suppose this is also OK for 1.8?

domcleal:
[test]

ares:
Yes, it contains few fixes, improves logging and adds more API, should be backward compatible, I built RPM only to nightly as I though, 1.8 is sort of frozen but if you're OK with getting this into 1.8, that would be great.

domcleal:
Merged as cc9bb2b and f3c1fa6 to develop and 1.8, thanks @ares.  PR tester bug fixed too, thanks @mmoll.

The plugin repos can and should be updated async to Foreman releases, so releasing plugin updates into 1.6, 1.7 and 1.8 is entirely possible. I'll retag the RPM into 1.8 so they match.  You can use koji commands for that (three of them) or http://ci.theforeman.org/job/packaging_plugin_retag/.

PR #590, opened 2015-03-18
Update Kafo to 0.6.6

ares:


domcleal:
Merged as 0d9c5ea, thanks @ares.  Will consider for deb/1.8 too if all is good in a day or two.

ares:
awesome, fingers crossed :ok_hand:

PR #592, opened 2015-03-19
Update to discovery 2.0.1 (Foreman 1.7)

lzap:
Do not merge, just a preview what I cherry-picked prior the SPEC change to
2.0.1.


domcleal:
The only one you're missing is the changelog from 895861d.  Otherwise :+1:

lzap:
Yeah I thought I can skip all the similar because tito will regenerate it... will cherry pick it too.

domcleal:
It'll only add a new changelog entry, it won't fill in the 2.0.0 one.

lzap:
Ok pushed thanks, going to bump to 2.0.1 now.

domcleal:
> Ok pushed thanks, going to bump to 2.0.1 now.

@lzap?

lzap:
On my plate... approaching.

lzap:
Done. http://koji.katello.org/koji/taskinfo?taskID=238958

lzap:
Do I do the same (cherry-picking) for debian?

domcleal:
You shouldn't need to cherry pick anything, as we always build from deb/* branches directly - so it should already be at 2.0.0.  No retagging for debs makes it far less confusing.

lzap:
Here we go, this was nicer :-) https://github.com/theforeman/foreman-packaging/pull/606

domcleal:
Thanks @lzap!  Building now.

I've been working on ways that we can build RPMs from branches rather than retagging, but unfortunately we have to deal with the uniqueness of git/tito tags and uniqueness of NEVRAs in Koji.  I'll post something soon...

lzap:
Thanks guys, going to bump hammer discovery cli as well. Develop only (new features).

domcleal:
I opened #627 which implements what I hinted about above.  This should mean that we can always build releases on rpm/1.7, rpm/1.8, rpm/develop etc like deb/*, without the outdated mess in branches that retagging from rpm/develop caused in this PR.

PR #593, opened 2015-03-19
Add rubygem-hammer_cli_foreman_tasks package

ehelms:
Based off https://github.com/theforeman/foreman-packaging/pull/469
See also: https://github.com/theforeman/hammer-cli-foreman-tasks/pull/16

### DIFF
--- a/rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
+++ b/rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
@@ -1,0 +1,70 @@
+%global gem_name hammer_cli_foreman_tasks
+%global confdir hammer
+
+Summary: Foreman CLI plugin for showing task information for resources and users
+Name: rubygem-%{gem_name}
+Version: 0.0.5
+Release: 1%{?dist}
+Group: Applications/Systems
+License: GPLv3+
+URL: http://github.com/theforeman/hammer-cli-foreman-tasks
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+%if 0%{?rhel} == 6
+Requires: ruby(abi)
+%else
+Requires: ruby(release)
+%endif
+
+Requires: ruby(rubygems)
+Requires: rubygem(hammer_cli_foreman) >= 0.1.1
+Requires: rubygem(hammer_cli_foreman) < 0.2.0
+Requires: rubygem(powerbar) >= 1.0.11
+Requires: rubygem(powerbar) < 1.1.0
+BuildRequires: ruby(rubygems)
+BuildRequires: rubygems-devel
+BuildRequires: ruby
+BuildArch: noarch
+Provides: rubygem(%{gem_name}) = %{version}
+
+%description
+Foreman CLI plugin for showing task information for resources and users
+
+%package doc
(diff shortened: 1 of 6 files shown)


rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
```diff
@@ -0,0 +1,62 @@
+%global gem_name hammer_cli_foreman_docker
```
domcleal: tasks

rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
```diff
@@ -0,0 +1,62 @@
...
+%global confdir hammer
+
+Summary: Contains the code for showing of the tasks (results and progress) in the Hammer CLI
```
domcleal: Should be less than 72 chars, just remove the "Contains the code" bit, it's superfluous

rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
```diff
@@ -0,0 +1,62 @@
...
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+%if 0%{?rhel} == 6
+Requires: ruby(abi)
+%else
+Requires: ruby(release)
+%endif
+
+Requires: ruby(rubygems)
+Requires: rubygem(hammer_cli_foreman) >= 0.1.2
```
domcleal: should be >= 0.1.1 and < 0.2.0

rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
```diff
@@ -0,0 +1,62 @@
...
+Requires: rubygem(powerbar)
```
domcleal: >= 1.0.11 and < 1.1.0

rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
domcleal: The gem summary is better actually, although it contains a typo: "Foreman CLI plugin for showing tasks information for resoruces [resources!] and users"

rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
domcleal: Looks like this is the current cause of the build/test failure.

domcleal:
Many thanks for resurrecting this!

domcleal:
The package should install a configuration file in hammer's .d directory to activate the plugin on installation.  For now you can include this as a source, but it should also be added into the gem so it's maintained in one place (see also 1db3327 for discovery).

ehelms:
@domcleal Updated with a proper release of the gem with the config file.

domcleal:
Thanks @ehelms!  Will build into nightly and 1.8 plugin repos.

PR #595, opened 2015-03-20
Update rubyipmi, add to foreman-proxy [deb]

domcleal:


domcleal:
[test] error in the PR test job again.

mmoll:
The very recent IPMI changes in smart-proxy won't go to 1.8, right?

domcleal:
No, I think it's a bit too much change right now and I think there are more core Foreman changes coming to use it anyway.

mmoll:
Not 100% sure, but I think the -proxy package should be bumped to 1.9.0-2, at least for FreeBSD ports we have a rule like that the port revision must get bumped when dependencies change.

domcleal:
I'd usually agree, but 1.9.0-1 hasn't been released yet and we're only producing nightly builds at this stage.  They actually have "9999" version numbers with some timestamped suffix so don't even use the entry in the changelog (we could probably drop it, frankly).

mmoll:
Ah, right... merging then.

PR #597, opened 2015-03-20
Update rubygem-bastion to 0.3.0

ehelms:


domcleal:
Thanks @ehelms, building now.

PR #600, opened 2015-03-23
Update rubygem-apipie-bindings to 0.0.12

mbacovsky:


mbacovsky:
Too late, seems already done.

domcleal:
Ah sorry, it's on my rubygems.org subscription list :)

mbacovsky:
Nice. Does it mean I don't have to care about packaging of this gem in the future? Does it work automatically? And does it include Hammer gems as well?

domcleal:
Yeah, you can leave it to me, I don't mind.  I do the same for all plugins and some dependencies including Hammer, and I'm happy to update them as and when I see them and think they're needed - though help is always appreciated if I haven't gotten to it yet.

It isn't automatic, it's just that I'm subscribed to various gems and they appear in my RSS feed (RSS to e-mail), then I usually work through them each morning.

mbacovsky:
@domcleal, you are unbelievable :clap: .

PR #603, opened 2015-03-24
Fixes #7008: support config file migration for smart-proxy with dns providers

witlessbird:


mmoll:
I think that's also needed for debs then

domcleal:
I think this is going to get horrendously complicated when any more modules are migrated..

witlessbird:
I was planning on working on something similar to AR migrations next. Unless you'd rather have it before dns providers...

domcleal:
I think I would, if that's OK, as we're otherwise going to have to revert this, which could be difficult depending on which release it all lands in.

lzap:
IMHO we should move this into external script (even the migrationscript itself) and keep this clean. Otherwise you need to maintain this twice: RPM and DEB.

domcleal:
Closing, due to #10003 giving us a generic mechanism.  Thanks @witlessbird!

PR #612, opened 2015-04-03
fixes #9990 - update apipie-bindings to 0.0.10

mmoll:
for deb/1.7 only!


mmoll:
http://ci.theforeman.org/job/packaging_build_deb_dependency/679/ is :green_heart:

domcleal:
Thanks @mmoll, building now.  I think it'll go to staging, then I'll push it next time we do a 1.7 update, which'll be soon.

PR #616, opened 2015-04-08
Copy packaging files over for Debian/jessie

mmoll:
with minor tweaks. This works after install puppetlabs-release-testing, as the Debian package "puppetmaster-common" depends on the package ruby-activerecord-deprecated-finders which renders db:migrate into being defective.


mmoll:
I added a Conflicts line for ruby-activerecord-deprecated-finders to the foreman package to at least prevent the worst...

domcleal:
Would you mind just wiping the changelogs?  Just keep the latest entry, I think that's what @GregSutcliffe usually does for new OSes.

We can extend this conditional for now to install PL Puppet in our tests: https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/config/scripts/systest_foreman.sh#L38

mmoll:
Wiped the changelogs for dependency packages.

mmoll:
No idea why precise is failing there... BTW, setting the Conflict line works when having everything in the repository, as opposed to the first try I did only in the local database. So no PuppetLabs repo required. :)

stumyp:
http://ci.theforeman.org/job/packaging_build_deb_coreproject/label=debian,os=precise/1864/console

says:
```
19:22:30  pbuilder-satisfydepends-dummy : Depends: ruby1.9.3 (>= 1:1.9.3.545) but it is not going to be installed.
19:22:30                                  Depends: ruby1.9.1-dev (>= 1:1.9.3.545) but it is not going to be installed.
19:22:30                                  Depends: git but it is not going to be installed. or
19:22:30                                           git-core but it is not going to be installed.
19:22:30                                  Depends: libxml2-dev but it is not going to be installed.
19:22:30                                  Depends: libxslt1-dev but it is not going to be installed.
19:22:30                                  Depends: libmysqlclient-dev but it is not going to be installed.
19:22:30                                  Depends: libpq-dev but it is not going to be installed.
19:22:30                                  Depends: pkg-config but it is not going to be installed.
19:22:30                                  Depends: libvirt-dev but it is not going to be installed.
19:22:30                                  Depends: libsqlite3-dev but it is not going to be installed.
19:22:30                                  Depends: facter but it is not going to be installed.
19:22:30 Unable to resolve dependencies!  Giving up...
```

mmoll:
Yes, I saw that, but no idea why this is happening, the ruby-ng ppa seems to have been added just fine before.

stumyp:
Sorry, I might be stating obvious.
I see also this:
```
19:22:26 Unpacking pbuilder-satisfydepends-dummy (from .../pbuilder-satisfydepends-dummy.deb) ...
19:22:26 dpkg: pbuilder-satisfydepends-dummy: dependency problems, but configuring anyway as you requested:
19:22:26  pbuilder-satisfydepends-dummy depends on debhelper (>= 7); however:
19:22:26   Package debhelper is not installed.
```

Package `debhelper` is from core distribution, and doesn't need extra repos.
So I think it is because of this warning above:

```
19:22:19 W: Duplicate sources.list entry http://ubuntu.osuosl.org/ubuntu/ precise/universe i386 Packages (/var/lib/apt/lists/ubuntu.osuosl.org_ubuntu_dists_precise_universe_binary-i386_Packages)
19:22:19 W: Duplicate sources.list entry http://deb.theforeman.org/ precise/nightly i386 Packages (/var/lib/apt/lists/deb.theforeman.org_dists_precise_nightly_binary-i386_Packages)
19:22:19 W: You may want to run apt-get update to correct these problems
```

mmoll:
I triggered a rebuild, but hit another issue (#10099), however as no Ubuntu/precise files were touched, I'm very sure this is unrelated anyway.

domcleal:
Merged as 4960a64, thanks @mmoll!

We'll figure out the precise failures if they continue once the execjs issue's fixed.

Please feel free to start building into nightlies.

GregSutcliffe:
>  Just keep the latest entry, I think that's what @GregSutcliffe usually does for new OSes

Indeed it is. @mmoll++ :)

PR #619, opened 2015-04-10
Copy 1.8 packaging files over for Debian/jessie

mmoll:
with minor tweaks.

For this to work for 1.8.0RC3
- [x]  theforeman/foreman-bats#75 has to get merged.
- [x] theforeman/puppet-foreman#316 has to get merged and 3.0.2 released to Puppet Forge.
- [x] puppetlabs-postgresql has to work with Debian/jessie.

For the third point, I prepared the ```41_debian_jessie``` branch in mmoll/puppetlabs-postgresql, although pulling in 4.3.0 would be my preferred option.


mmoll:
BTW, I tested with RC2 packages, there's just the "db_migrate/seed is not idempotent" thing, so a systest should succeed with a working postgesql module.

domcleal:
Thanks @mmoll!  All the items above should be resolved - I updated postgresql to 4.3.0 in 1.8-stable as that seems to be OK in nightlies.

Would you mind building the dependencies, then I'll include jessie in the RC3 builds which I think might be today.

PR #620, opened 2015-04-12
remove salt-master from the list of hard dependencies, fixes #10120

bechtoldt:


mmoll:
I'm a bit sceptical about that one.

Saltstack does eben provide own APT repos and personally I'm not very fond of stuff getting installed by bypassing the OS package manager. Maybe that's more of a topic to raise at Salt's side...

I would be more pro if in the future the foreman-proxy could be decoupled from the salt-master by only using its API to communicate so one can install it on a different node and thus the requirement itself would indeed not exist anymore.

stbenjam:
> I would be more pro if in the future the foreman-proxy could be decoupled from the salt-master by only using its API to communicate so one can install it on a different node and thus the requirement itself would indeed not exist anymore.

I think that's a long way off. I've had a look at the API while doing the fileserver stuff, and I don't see any endpoint to manipulate keys or autosign.  

I don't know enough about Debian to comment here, but I don't think I'd accept this for the RPM spec.  Why does the bootstrap script do a source install?

mmoll:
In theory one could change the dependency to a suggestion, but I really see very little use for that in real world setups, where everything should come as OS packages.

mmoll:
After some more thoughts and discussions on IRC, we'll close this one, sorry...

bechtoldt:
ack

PR #622, opened 2015-04-14
Refs #10129 - foreman-ec2 must depend on foreman-compute (fog) (DEB)

mmoll:


mmoll:
reopening with another source branch name to let the tests succeed...

PR #623, opened 2015-04-14
Refs #10129 - foreman-ec2 must depend on foreman-compute (fog) (DEB)

mmoll:


eLobato:
This would solve the ticket, but if I'm not wrong, the whole purpose of the ec2 package is to not have to install the whole bundle for just one compute resource - see https://github.com/theforeman/foreman-packaging/blob/rpm/develop/foreman/foreman.spec#L278 , -compute 'obsoletes' -ec2.

mmoll:
strange, I thought -compute is the fog base package and now that fog is modularized and aws/ec2 has it's own fog extension -ec2 is the package for Amazon/EC2 support for Foreman.

mmoll:
foreman.spec says ```Obsoletes: foreman-ec2 < 1.3.0```, I think that's because the -compute package came with 1.3.0, I'll prepare a rpm PR and we'll see.

mmoll:
see #624

domcleal:
Merged, thanks @mmoll.

I had hoped we could avoid making ec2 dependent on fog, but it needs #8516 and a bit more work in fog_extensions I think.  Maybe in 1.9!

PR #625, opened 2015-04-16
Refs #10154 - added rubygem-locale to non-scl rhel6

mbacovsky:
non-SCL rubygem-locale is missing from RHEL6 repos and Hammer is failing.

So far I'm not sure what changed or why we didn't caught this earlier.

Related Hammer PR: https://github.com/theforeman/hammer-cli/pull/168


domcleal:
Merged as c38b9a0 with a comps/ update too.  Thanks @mbacovsky.

PR #627, opened 2015-04-22
Better branched builds with Foreman version macro [rpm]

domcleal:
Packages (plugins) built on foreman output different binary RPMs due to the
version of Foreman they're built against, as macros change.  The .fm1_9 string
is put into the release field to differentiate the EVRs, since koji and tito
both rely on its uniqueness - building the same EVR on rpm/1.8 and rpm/1.7 to
get both 1.8 and 1.7 builds would otherwise not be possible.

(Consider `rubygem-foreman_docker-1.2.4-1.el7` where %post might have
different behaviour between versions, so `-1.fm1_8.el7` allows both to exist.)

This adds a %foremandist macro to foreman-release, which is used in plugin
release fields to differentiate them.  It will be added to the buildroots in
koji so it can be expanded.  (`Release: 1%{?foremandist}%{?dist}`)

For tito, the tag_suffix configuration is set so the version number is also
included in its git tags for uniqueness across branches.

This is a pre-req for our own SCL, as rebuilding packages on multiple branches
will be common when the SCL configuration differs, instead of retagging.  It
will perhaps allow us to build most plugins on versioned branches easily rather
than retagging develop builds to older versions.  [In the same way that deb/* works.]

All packages using Foreman at build time have the dist tag added to release.

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -225,6 +225,7 @@ for Yum.
 %files release
 %config %{_sysconfdir}/yum.repos.d/*
 /etc/pki/rpm-gpg/*
+%{_sysconfdir}/rpm/macros.%{name}-dist
 
 %package libvirt
 Summary: Foreman libvirt support
@@ -360,6 +361,7 @@ Meta package to install asset pipeline support.
 Summary: Foreman plugin support
 Group: Development/Libraries
 Requires: %{name} = %{version}-%{release}
+Requires: %{name}-release = %{version}-%{release}
 Requires: %{name}-sqlite = %{version}-%{release}
 
 %description plugin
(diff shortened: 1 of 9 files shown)


domcleal:
EL7 test chained build at https://fedorapeople.org/~domcleal/foreman/rpm_pr627/ if you want to check it.  Since it needs to be chained and requires a koji config change, I can't link to a scratch build there.

I hope the explanation's reasonably clear, let me know if you have questions.

ehelms:
Two questions that I think I know the answer to but want to double check:

1) For nightly builds, there won't be a version denoted within the release?
2) This is only for packages that are making use of RPM macros currently? This doesn't affect say hammer packages?

domcleal:
> 1) For nightly builds, there won't be a version denoted within the release?

Nightly builds of what?

> 2) This is only for packages that are making use of RPM macros currently? This doesn't affect say hammer packages?

Yeah, only those that use our macros.  Hammer and other packages are simple RPMs where the binary RPM output won't differ between major versions of Foreman (since they don't build on it in any way) - which is the scenario I'm trying to tackle.

ehelms:
Oh right, none of these have nightly builds. My brain was on the Katello case that uses some of the macros but builds and releases nightly RPMs at present.

The additional release denoting Foreman version does seem like a nice way to quickly see what version of Foreman a package was built for when looking at the RPM NVREA (if it was propagated to all project packages we maintain). Just a thought.

:+1: for me

domcleal:
> Oh right, none of these have nightly builds.

Yeah, indeed, no plugins get nightly builds.  (Nobody's asked, so it simplifies matters, and perhaps encourages people to release early + often rather than relying on nightlies and quarterly releases.)

> The additional release denoting Foreman version does seem like a nice way to quickly see what version of Foreman a package was built for when looking at the RPM NVREA (if it was propagated to all project packages we maintain). Just a thought.

I see what you mean, it might be useful to see, even if the binary RPM doesn't differ based on that version.

domcleal:
Merged, and then partially reverted again in 8d57a9d as it doesn't work properly with koji and tito.  I've sent a patch to the latter which should allow it to generate SRPMs with foremandist defined, which should make it work once the next version's out.

PR #629, opened 2015-04-24
update hammer_cli_foreman_ssh to 0.0.3 (DEB)

mmoll:
- move ruby-sinatra directroy to correct place (it's needed only on Debian/wheezy)
- package/update net-ssh-* gems where needed
- update hamemr_cli_foreman_ssh


domcleal:
Thanks @mmoll!  Please go ahead and build.

PR #631, opened 2015-04-28
plugins:foreman_docker - Release 1.3.0

eLobato:


domcleal:
This is still for 1.7 through 1.9, right?

eLobato:
Yup, 1.7 compatible.

domcleal:
Merged as 0873563 without the changelog (tito does this, makes cherry picking much easier).  Thanks @eLobato.

PR #634, opened 2015-04-28
[digitalocean] Update to version 0.2.0 (RPM)

TJM:
This should be released to develop and 1.8 branches only.


domcleal:
Thanks @TJM!

PR #637, opened 2015-04-30
Scap client is not in SCL

ares:
Package is built without SCL http://koji.katello.org/koji/packageinfo?packageID=806 (which makes sense)


domcleal:
Thanks @ares

PR #638, opened 2015-05-05
refs #10203 - make default scap content available to Foreman

shlomizadok:
foreman_openscap has a [seed](https://github.com/OpenSCAP/foreman_openscap/blob/master/db/seeds.d/openscap_scap_default.rb) and a [bulk uploader](https://github.com/OpenSCAP/foreman_openscap/blob/master/lib/foreman_openscap/bulk_upload.rb#L12), which takes scap content from scap-security-guide and creates default scap_content in Foreman


domcleal:
The changes you link to only seem to be in 'master', and not 0.4.0, which this spec file's for?

PR #640, opened 2015-05-07
Adds foreman_openscap-0.3.4

shlomizadok:
I hope I did this correctly... :)


rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -23,8 +23,9 @@ License: GPLv3
 URL: https://github.com/OpenSCAP/foreman_openscap
 Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 
-Requires: foreman >= 1.5.0
+Requires: foreman >= 1.7.0
 
+Requires: scap-security-guide >= 0.1.22
```
domcleal: This won't resolve on el6 or f19.

rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -23,8 +23,9 @@ License: GPLv3
 URL: https://github.com/OpenSCAP/foreman_openscap
 Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 
-Requires: foreman >= 1.5.0
+Requires: foreman >= 1.7.0
```
domcleal: Be sure to make the same update in rpm/develop on the next release, or sooner.

domcleal:
You also need to update the Version field.

domcleal:
A quick bit of background: we have plugin repos for each Foreman version (1.7, 1.8, nightly which becomes 1.9).

I just realised that foreman_openscap 0.4.0 is already available in the 1.8 repos, so I'm not sure what I can do with this PR, so will suggest closing it and then just build 0.4.1 for 1.8 and nightly from your other PR.

shlomizadok:
I know the background.

PR #643, opened 2015-05-10
Fixes foreman_chef dependency

ares:


mmoll:
Normally it would be deface <1.0 in the gemspec and in the package, but I'll merge as is and open an issue over the the gem.

mmoll:
Also merged into ```deb/1.8``` as 6380a59, thanks @ares!
Building now.

PR #648, opened 2015-05-11
Fog 1.30.0 new packages [rpm]

domcleal:
Two new dependencies that will be needed for Fog 1.30.0 (coming in #10261).

Both based on fog-aws, only weirdness is around using ruby-libvirt on Fedora for fog-libvirt.


domcleal:
Merged, building now.

PR #651, opened 2015-05-12
Fixes #10362 - Update daemons to 1.2.2

iNecas:
There was a bug on daemons 1.2.1 to cause 100% load on a monitor task.


domcleal:
Thanks @iNecas!

PR #652, opened 2015-05-12
update foreman_xen to 0.0.6 (DEB)

mmoll:
And mark incompatible with foreman 1.8, because of the networking changes. I'd like to have this in nightly, 1.8 and 1.7, so people upgrading from 1.7 will at least know, it won't work.


domcleal:
Thanks, building!

PR #653, opened 2015-05-12
update foreman_xen to 0.0.6 (RPM)

mmoll:
And mark incompatible with foreman 1.8, because of the networking changes. I'd like to have this in nightly, 1.8 and 1.7, so people upgrading from 1.7 will at least know, it won't work.


rubygem-foreman_xen/rubygem-foreman_xen.spec
```diff
@@ -26,6 +26,7 @@ URL:        http://github.com/theforeman/foreman-xen
 Source0:    http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 
 Requires:   foreman-compute >= 1.5.0
+Requires:   foreman-compute < 1.8.0-0.1.RC1
```
mmoll: not sure if this line is correct.

domcleal: `< 1.8.0` will suffice

<pre>
$ rpmdev-vercmp 1.8.0 1.8.0-1
1.8.0 < 1.8.0-1
$ rpmdev-vercmp 1.8.0 1.8.0-0.1.RC1
1.8.0 < 1.8.0-0.1.RC1
</pre>

However this'll cause us problems with our repoclosure tests if I was to put it in 1.8 and nightly, since the dep won't resolve.  I could put it in 1.7 only, or leave this out for the time being until a new 1.8-compat release is available.  Preference?

mmoll: Changed. Only to 1.7 then, please.

domcleal:
Ok, will build into nightlies and move into 1.7.  Thanks @mmoll!

PR #657, opened 2015-05-18
Update dynflow to 0.7.8

iNecas:


domcleal:
@iNecas please also add a changelog entry

domcleal:
And update gem.list.

iNecas:
Updated

plugins/ruby-foreman-dynflow/debian/changelog
```diff
@@ -1,3 +1,7 @@
+ruby-foreman-dynflow (0.7.8-1) stable; urgency=low
+
+  * Update to 0.7.8
+
 ruby-foreman-dynflow (0.7.7-1) stable; urgency=low
```
domcleal: Missing a line with your name + date here.  `scripts/changelog.rb` may help you.

iNecas: It helped indeed :)

domcleal:
Merged as 3e80c87 for nightlies, thanks @iNecas!

PR #658, opened 2015-05-18
update foreman-tasks in deb/1.8

mmoll:
I think this has at least a positive effect on http://projects.theforeman.org/issues/10353 as the daemons update fixed this for me.


domcleal:
@iNecas ok with you to bring 0.6.13 into 1.8?  I'll bring the RPMs in too.

iNecas:
yes, it should be compatible

domcleal:
Thanks @mmoll!

PR #659, opened 2015-05-19
Fix foreman_chef deface dependency name

domcleal:
Noted in http://projects.theforeman.org/issues/10353#note-8


mmoll:
totally missed that when merging the originbal PR, thanks @domcleal!

PR #663, opened 2015-05-21
Update rubygem-logging to 2.0.0

ehelms:


ehelms:
Figured I would go ahead and get this open for review. I am unfamiliar with debian packaging, but willing to attempt it to update this there as well.

domcleal:
We can't do this, since our non-SCL version is used for Hammer and so needs 1.8 compatibility.

domcleal:
Oh, and Debian packages for Foreman use bundler so a similar update wouldn't be required for Foreman core, they'd get the best version that bundler lets us use.

ehelms:
Bummer,

PR #664, opened 2015-05-22
[do not merge] refs #10003 - use new config migration script [rpm]

domcleal:
Depends on https://github.com/theforeman/smart-proxy/pull/272, and possibly a couple of bug fixes.  Output also needs quietening, once we know it's OK.

http://koji.katello.org/koji/taskinfo?taskID=265524 is an el7 build.

### DIFF
--- a/foreman-proxy/foreman-proxy.spec
+++ b/foreman-proxy/foreman-proxy.spec
@@ -121,7 +121,6 @@ install -d -m0755 %{buildroot}%{_sysconfdir}/%{name}
 install -d -m0755 %{buildroot}%{_sysconfdir}/%{name}/settings.d
 install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}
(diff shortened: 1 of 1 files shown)


domcleal:
Updated with quiet output going to a log file, and with the latest fixes in the smart proxy PR.

https://fedorapeople.org/~domcleal/foreman/rpm_proxymigrate_10003/ has test builds for all three RPM OSes. (Version numbers might be behind nightly, use rpm --force for those.)

domcleal:
@witlessbird did you get time to try this out?

witlessbird:
Nope, but will give it a try today...

domcleal:
Thanks.  I just tried running it on a 1.5.4 el6 installation and hit a couple of issues noted on the main PR, so it probably only works for el7/f19 right now.

witlessbird:
just tested 1.8.1->1.9.0 on el7, which worked ok. Going to check the issues you found on earlier releases now...

domcleal:
Merged as 85a7dd0.

PR #668, opened 2015-05-28
refs #10003 - use new config migration script [deb]

domcleal:
I've only been able to test this by testing the scripts standalone, which is pretty much copied from #664, as I don't have a quick way to build a package from both the smart proxy + packaging branches.

### DIFF
--- a/debian/jessie/foreman-proxy/postinst
+++ b/debian/jessie/foreman-proxy/postinst
@@ -41,23 +41,22 @@ case "$1" in
 
     chmod 0700 "/var/cache/foreman-proxy"
 
-    if [ -d /var/lib/foreman-proxy ] ; then
-      # Migrate legacy monolithic proxy config
-      # To detect if a migration is needed, we run the script in a tmpdir and
-      # check if the data in the tmpdir is the same as what's already in place
-      cd /var/lib/foreman-proxy
-      if $FOREMAN_HOME/extra/migrate_settings.rb ./settings.yml; then
-        mv settings.yml $CONFIGDIR/settings.yml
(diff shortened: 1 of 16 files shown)


debian/jessie/foreman-proxy/postinst
```diff
@@ -41,22 +41,19 @@ case "$1" in
...
-      cd /tmp && rm -rf /var/lib/foreman-proxy
-      # If we migrated, we also need to fix the settings.d dir
-      if grep -q ^:settings_directory: $CONFIGDIR/settings.yml; then
-        sed -i "s|:settings_directory:.*|:settings_directory: $CONFIGDIR/settings.d|" $CONFIGDIR/settings.yml
-      else
-        echo ":settings_directory: $CONFIGDIR/settings.d" >> $CONFIGDIR/settings.yml
-      fi
+    TEMP=$(mktemp -d)
+    trap "rm -rf $TEMP" EXIT
+    pushd $TEMP >/dev/null
```
mmoll: that's a bashism:
```
/var/lib/dpkg/info/foreman-proxy.postinst: 46: /var/lib/dpkg/info/foreman-proxy.postinst: pushd: not found
```

domcleal: Thanks, switched to a subshell and `cd`.

domcleal:
Once it builds I'll also give it a bit of testing before merging.

domcleal:
Also fixed the missing extra/migrations/ directory and it seems to be working OK now.

mmoll:
Looking good, at least for me on Debian/jessie, thanks @domcleal!

PR #672, opened 2015-06-03
Add rubygem-foreman_admin package

komidore64:


domcleal:
This isn't a core Foreman project so it wouldn't be added to our main release repo as suggested in this PR, and we don't really have a contrib repo for third party tools.

Is this meant to be installed on a Foreman server?  How will that be done - manual installation, package dependencies or via our installer?

komidore64:
Yes, this is meant to be installed on a Foreman (as well as on a Katello) server. I want it to be installed with the installer. That's next on the list of things to do for `foreman-admin`.

mmoll:
Please also think about a Debian package then.

domcleal:
Yeah, in that case I'd suggest submitting it to https://github.com/theforeman/foreman/tree/develop/script so it can be installed as part of Foreman.

komidore64:
I have little knowledge of building DEB packages, help would be greatly appreciated.

I have some high-priority stuff to work on currently, but I will get back on this sometime next week.

domcleal:
Yeah, that's one reason why simply putting it in Foreman saves us the burden of extra packages.

komidore64:
@domcleal placing `foreman-admin` into `foreman`?

domcleal:
Yes.

komidore64:
the Foreman package is where Foreman bits live though. The server lives there. Foreman-Admin isn't really part of the server, so it doesn't make sense to me to lump it in with the server.

domcleal:
If it's being installed on the Foreman server, then it's part of it and it can be maintained in the same way with branches etc.

komidore64:
@domcleal I don't quite follow. can you please explain further? "maintained with branches" ?

domcleal:
> "maintained with branches" ?

Foreman, like the other three top-level projects we ship in this repo is branched for releases.  It's likely that if we adopted this tool from you that we'd have to do the same, and I don't think adding another project is worth it for one command/tool.

If it's installed with the Foreman server then we should ship it as part of it, along with the other scripts and commands we install too.  It also will make it easier for plugins to extend as extensions can be carried in the Foreman plugin itself.

mmoll:
what's the status here? :)

komidore64:
``foreman-admin`` needs some more work before getting distributed with Foreman and/or Katello. I'm going to close this. I'll revisit in the future. @domcleal and @mmoll thank you for all your feedback!

PR #673, opened 2015-06-04
Extend custom fetcher to get source from local checkout

domcleal:
Our custom tito sources fetcher, usually used to pull nightly sources from
Jenkins, can now take a "source_dir" argument pointing to a local checkout.
If it's a rubygem, it'll build the gem and otherwise calls "rake
pkg:generate_source" (defined in all Foreman projects) to generate a
tarball.

This allows easy scratch building of a git branch.  See the README change
for usage.

@iNecas I think this is what you were asking after yesterday.

### DIFF
--- a/rel-eng/custom/custom.py
+++ b/rel-eng/custom/custom.py
@@ -12,11 +12,13 @@
 # in this software or its documentation.
 
 from datetime import datetime
+import glob
 import json
 import os
 import os.path
 import re
 import shutil
+import subprocess
 import urllib
 from zipfile import ZipFile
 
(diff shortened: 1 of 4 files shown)


rel-eng/custom/custom.py
```diff
@@ -120,13 +87,108 @@ def fetch(self):
...
+
+    It first copies source files from git, then copies the source file
+    over the top, so they're merged (patches etc can then be stored
+    in git).
+
+    Takes the following arguments:
+      source_dir: path to repo ("~/foreman")
+    """
+    def _fetch_local(self):
+        source_dir = self.builder.args['source_dir']
```
iNecas: we probably need `os.path.expanduser()` for the `~` to work, otherwise I'm getting ` No such file or directory:` errors

domcleal: Strange, your shell ought to expand that before it's passed into tito, e.g.

<pre>
10:33 $ echo tito release --scratch --arg source_dir=~/code/foreman/foreman koji-foreman-nightly                                       
tito release --scratch --arg source_dir=/home/dcleal/code/foreman/foreman koji-foreman-nightly
</pre>

iNecas: not working on zsh that way

domcleal: okay, I added that.

iNecas:
Works like a charm now! ACK and thanks

domcleal:
Thanks for testing!

PR #674, opened 2015-06-09
update clamp to 1.0.0 (DEB)

mmoll:


domcleal:
Do we know if hammer_cli's i18n support is fully compatible with the patch that got merged, or if we need to make any changes first?  (@tstrachota)

mmoll:
From what I saw in the clamp PR it's fully compatible and it works on FreeBSD (where I already updated clamp to 1.0.0) with German and Russian locale.

domcleal:
Ah nice, hammer_cli tests pass too.

tstrachota:
@domcleal it's fully compatible. We've been already using it as a custom patch that is now merged into clamp.

domcleal:
Thanks @mmoll!  Building.

PR #675, opened 2015-06-09
update clamp to 1.0.0 (RPM)

mmoll:


rubygem-clamp/rubygem-clamp.spec
```diff
@@ -80,6 +74,7 @@ cp -pa .%{gem_dir}/* \
 %doc %{gem_dir}/doc/%{gem_name}-%{version}
 %{gem_instdir}/README.md
 %{gem_instdir}/CHANGES.md
+%{gem_instdir}/LICENSE
```
domcleal: no need for this line, it's in the main package under %files (and -doc depends on the main package)

mmoll: done

domcleal:
Thanks @mmoll!

PR #678, opened 2015-06-15
Update foreman_bootdisk to 6.0.0

domcleal:


mmoll:
seems to work just fine here, building now. thanks @domcleal!

PR #679, opened 2015-06-16
refs #10787 - remove foreman-compute subpackage deps

domcleal:
Only from these three CRs as their bundler groups are now truly independent of the main Fog group (https://github.com/theforeman/foreman/commit/3f73d0054d935e220c74a87dc0bcbbd8731dc5dd).


mmoll:
looking good, thanks @domcleal!

PR #680, opened 2015-06-16
[do not merge] refs #10641 - add select2-rails to assets

domcleal:
Just going to leave this here ready for https://github.com/theforeman/foreman/pull/2415 in case it's merged.


shlomizadok:
[test]

mmoll:
[test]

domcleal:
The PR job doesn't work for Foreman core packages right now, which is why one of the test jobs will fail - also, it can't do chained builds like this would require.

Will merge and build this anyway.

PR #683, opened 2015-06-19
Update dynflow to 0.7.9

iNecas:


plugins/ruby-foreman-tasks/debian/control
```diff
@@ -2,7 +2,7 @@ Source: ruby-foreman-tasks
 Section: ruby
 Priority: extra
 Maintainer: Michael Moll <mmoll@mmoll.at>
-Build-Depends: debhelper, foreman (>= 1.8.0~rc1), foreman-sqlite3 (>= 1.8.0~rc1), ruby-foreman-dynflow (>= 0.7.7)
+Build-Depends: debhelper, foreman (>= 1.8.0~rc1), foreman-sqlite3 (>= 1.8.0~rc1), ruby-foreman-dynflow (>= 0.7.9)
```
ares: I think foreman-tasks should be bumped if you change it's spec, I suppose we need 0.6.13-4 in [changelog](https://github.com/theforeman/foreman-packaging/blob/deb/develop/plugins/ruby-foreman-tasks/debian/changelog)

iNecas: something like what I've updated now?

ares:
LGTM, @mmoll could you please check as well?

plugins/ruby-foreman-tasks/debian/control
mmoll: If that package should really forcefully depend on dynflow 0.7.9 or later, the ruby-foreman-tasks version must indeed be bumped to -4, but as foreman-tasks 0.6.13 works with dynflow 0.7.7 also, I don't really a point in doing that.

iNecas: I will have new foreman-tasks version, that will need this version as well, so I think we can leave it here.

mmoll: As we reflect the dependencies from the gems in the packages, please leave that as ```ruby-foreman-dynflow (>= 0.7.7)``` for now and when a new foreman-tasks gem is relased with ```dynflow >= 0.7.9```, change it in the foreman-tasks packaging PR.

iNecas: Ok, updated

mmoll:
Working fine here, thanks @iNecas!

PR #685, opened 2015-06-22
Update smart_proxy_chef to 0.1.4

ares:


domcleal:
Please also update plugins/smart_proxy_chef/chef.rb, plugins/ruby-foreman-chef/foreman_chef.rb and plugins/ruby-foreman-chef/debian/gem.list.

domcleal:
We ought to write a list of steps in deb/develop's README, as it has some unique constructs for each type of plugin.

ares:
fixed, thanks @domcleal 
+1 to what you suggest if it's too hard to unify all packages

domcleal:
Thanks @ares, building both for nightlies.

I don't think it'd be hard to unify.

PR #693, opened 2015-06-30
Update foreman-tasks to 0.6.14

iNecas:


domcleal:
Thanks @iNecas, will build to nightly plugins.

PR #695, opened 2015-07-01
Rpm packages for dynflow 0.8.1 and its dependencies

iNecas:
The deb packages on its way once I learn how to do it from scratch


domcleal:
Will there also be a foreman-tasks release to go with it, since dynflow's currently pinned to < 0.8.0?

iNecas:
Yes, I'm preparing the foreman-tasks release as well https://github.com/theforeman/foreman-tasks/pull/116, I need this rpms to get in first to get the BuildDependencies in before the foreman-tasks rpm

domcleal:
Sure, that's what I expected - just didn't want to update one and then have deps broken for a long time.  Will review in due course!

domcleal:
Could you add the new package to rel-eng/tito.props and comps/*plugins* please?

rubygem-concurrent-ruby-edge/rubygem-concurrent-ruby-edge.spec
```diff
@@ -0,0 +1,71 @@
...
+%{?scl:scl enable %{scl} "}
+gem install --local --install-dir .%{gem_dir} \
+            -V --force %{SOURCE0}
+%{?scl:"}
+
+%build
+
+%install
+mkdir -p %{buildroot}%{gem_dir}
+mkdir -p %{buildroot}%{gem_extdir}/lib
```
domcleal: Unused

rubygem-concurrent-ruby/rubygem-concurrent-ruby.spec
```diff
@@ -0,0 +1,71 @@
...
+%{?scl:scl enable %{scl} "}
+gem install --local --install-dir .%{gem_dir} \
+            -V --force %{SOURCE0}
+%{?scl:"}
+
+%build
+
+%install
+mkdir -p %{buildroot}%{gem_dir}
+mkdir -p %{buildroot}%{gem_extdir}/lib
```
domcleal: ditto

iNecas:
@updated the comps files

rubygem-concurrent-ruby-edge/rubygem-concurrent-ruby-edge.spec
iNecas: Good catch, updated both

comps/comps-foreman-plugins-fedora19.xml
```diff
@@ -86,6 +88,8 @@
       <packagereq type="default">rubygem-bastion-doc</packagereq>
       <packagereq type="default">rubygem-chef-api-doc</packagereq>
       <packagereq type="default">rubygem-commonjs-doc</packagereq>
+      <packagereq type="default">rubygem-concurrent-ruby</packagereq>
```
domcleal: -docs, ditto on the others.  Re-running ./comps_docs.sh should fix it.

iNecas: You really have scripts for everything! :)

iNecas: Updated

rubygem-concurrent-ruby-edge/rubygem-concurrent-ruby-edge.spec
```diff
@@ -0,0 +1,70 @@
...
+Release: 1%{?dist}
+Group: Development/Languages
+
+License: MIT
+URL: https://github.com/ruby-concurrency/concurrent-ruby
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
+Requires: %{?scl_prefix}ruby(abi) = %{rubyabi}
+Requires: %{?scl_prefix}ruby(rubygems)
+Requires: %{?scl_prefix}ruby
+BuildRequires: %{?scl_prefix}ruby(abi) = %{rubyabi}
```
domcleal: The build failure (on Fedora) is coming from these lines, you need to put in a check for Fedora and use `ruby(release)` (no specific version) and only use `ruby(abi)` on SCL.  Copy the style that's in other spec files, it's in almost all of them.  Be sure to update both BR and Requires.

iNecas: Thanks, updated. I've used https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-pg/rubygem-pg.spec#L21 as a base for the beginning, which we probably don't build for Fedora, so it was not a lucky choice for me.

rubygem-concurrent-ruby-edge/rubygem-concurrent-ruby-edge.spec
```diff
@@ -0,0 +1,76 @@
...
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.1.0.pre3
+Release: 1%{?dist}
+Group: Development/Languages
+
+License: MIT
+URL: https://github.com/ruby-concurrency/concurrent-ruby
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+%if "%{?scl}" == "ruby193" || (0%{?rhel} == 6 && "%{?scl}" == "")
```
domcleal: I'd suggest just using `%if 0%{?fedora} > 18` as we usually do in SCL-enabled spec files - this condition's rather clunky and old :)

iNecas: What is the prototype of a good spec here, I used https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-apipie-bindings/rubygem-apipie-bindings.spec#L22 , but apparently I'm not lucky today :)

domcleal: Ah yes, apipie-bindings is built for non-SCL usage.  I'd tend to pick something new, like fog-libvirt or even dynflow!

Sorry for the confusion, I'll get back to my gem2spec templates soon..

iNecas: Updated

domcleal:
Thanks @iNecas, I'll build them into nightly plugins.

domcleal:
> The deb packages on its way once I learn how to do it from scratch

These shouldn't be too hard if you do it as a PR, as you'd just add the new gems to the dynflow package without splitting it up any further (not sure there's any benefit to doing so, we normally bundle).

PR #696, opened 2015-07-03
Update foreman-tasks to 0.7.0

iNecas:


iNecas:
I will open debian PR after https://github.com/theforeman/foreman_chef/pull/23 is merged, so that I can move all the dependent gems to the new version at once

iNecas:
I've discovered some issue when running katello tests, I will reopen with 0.7.1 version with the fix

PR #697, opened 2015-07-03
Update foreman-tasks to 0.7.1

iNecas:
The deb version will be released after
https://github.com/theforeman/foreman_chef/pull/23 is merged


domcleal:
@iNecas would you or @stbenjam mind making the same version change in foreman_salt?

rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -34,8 +34,8 @@ Requires: %{?scl_prefix}ruby(release)
 Requires: %{?scl_prefix}ruby(abi)
 %endif
 
-Requires: %{?scl_prefix}rubygem(dynflow) >= 0.7.9
-Requires: %{?scl_prefix}rubygem(dynflow) < 0.8.0
+Requires: %{?scl_prefix}rubygem(dynflow) >= 0.8.0
```
domcleal: `>= 0.8.1`, ditto below

domcleal:
foreman_salt's done, thanks @stbenjam.  Regarding the deb stuff, please just go ahead with the PRs for dynflow + foreman-tasks if you're willing, it's much easier to work up the stack bit by bit.

rubygem-foreman-tasks/rubygem-foreman-tasks.spec
iNecas: Updated

iNecas:
@domcleal ad the debian: I was not sure how is it, when we have different version of the foreman-tasks than the foreman_chef actually needs… would the multi-versioned variant actually work?

domcleal:
Depends if the package has version deps matching the gems - ruby-foreman-chef does, so it should be fine, apt should correctly resolve it.  ruby-foreman-salt didn't, but that'll be fixed shortly.  It's only nightlies, so little harm if the versions aren't quite right for a day.

domcleal:
Thanks @iNecas!  Building.

PR #698, opened 2015-07-06
Fixes #10438 - Removed trends:clean from crontab

ShimShtein:
rpm part here: https://github.com/theforeman/foreman-packaging/pull/699

PR #699, opened 2015-07-06
Fixes #10438 - Removed trends:clean from crontab

ShimShtein:
deb part is here: https://github.com/theforeman/foreman-packaging/pull/698


domcleal:
Thanks!

PR #704, opened 2015-07-06
update dynflow and foreman-tasks (DEB)

mmoll:
@iNecas let's get this in like this first and when it's time we'll see how to handle the dynflow requirement on proxies for rex.

Tested on Debian/jessie.


domcleal:
Thanks @mmoll!  Building.

iNecas:
Thanks @mmoll indeed

PR #710, opened 2015-07-24
plugins:foreman_docker - Release 1.4.0

eLobato:


domcleal:
Which releases is this for?

plugins/ruby-foreman-docker/debian/changelog
```diff
@@ -1,3 +1,7 @@
+ruby-foreman-docker (1.4.0-1) stable; urgency=low
+
+  * Update release to 1.4.0
+ -- Daniel Lobato Garcia <elobatocs@gmail.com> Fri, Jul 24 2015 17:39:41 -0400
```
domcleal: Missing line breaks

eLobato:
@domcleal 1.7 through 1.9 included

domcleal:
Thanks @eLobato, building for 1.7-nightly repos - RPMs too.

plugins/ruby-foreman-docker/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-foreman-docker (1.4.0-1) stable; urgency=low
+
+  * Update release to 1.4.0
+
+ -- Daniel Lobato Garcia <elobatocs@gmail.com>  Mon, Jul 27 2015 12:04:44 -0400
```
domcleal: I had to fix an error here, the day and month are the wrong way around.  Please use the provided helper, scripts/changelog.rb or dch.

domcleal: While I'm thinking of it, please don't include the changelog in RPM PRs too, since tito handles that.

PR #712, opened 2015-07-29
Document steps to update different package types

domcleal:


mmoll:
merged as c208e8e, thanks @domcleal!

PR #713, opened 2015-07-29
fixes #8996 - use Ruby 1.9 for hammer on precise

mmoll:
yet untested.

### DIFF
--- a/dependencies/precise/hammer_cli/changelog
+++ b/dependencies/precise/hammer_cli/changelog
@@ -1,3 +1,9 @@
+ruby-hammer-cli (0.2.0-2) stable; urgency=low
+
+  * Use Ruby 1.9 explicitly on Ubuntu/precise
+
+ -- Michael Moll <mmoll@mmoll.at>  Wed, 29 Jul 2015 17:32:35 +0200
+
 ruby-hammer-cli (0.2.0-1) stable; urgency=low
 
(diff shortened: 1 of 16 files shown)


domcleal:
Interesting, I'll try and give it a go soon.  I was wondering if we should rely on the Brightbox version too, but hopefully the p0 in 12.04 will be enough.

mmoll:
[test]

domcleal:
A quick test of upgrading installed the right locale package, but the shebang line in /usr/bin/hammer needs changing to ruby1.9.1 after which point a host list worked fine.

mmoll:
This should be handled by the XS-Ruby-Versions line, need to check this again...

domcleal:
Hm okay, it was just `/usr/bin/env ruby`.

mmoll:
I see the problem now: https://wiki.debian.org/Teams/Ruby/Packaging#Handling_of_shebangs requires ```/usr/bin/ruby```, while hammer is using ```/usr/bin/env ruby```, I'll just replace this after unpacking and see what's happening.

mmoll:
Should all be fine now. By also dropping the dependency onto the rubygems package, it even works without any ruby1.8 components.

domcleal:
Merged, thanks very much @mmoll!  Building now.

PR #717, opened 2015-07-31
Update foreman_setup to 3.0.1

domcleal:
For 1.9 too, if this works.


domcleal:
Hm, PR tests aren't actually running, let me investigate that as I changed it the other day.

domcleal:
[test]

mmoll:
also merged as c2cbd59 for 1.9, building then, thanks @domcleal!

PR #718, opened 2015-07-31
Update foreman-tasks to 0.7.2 (including relevant dependencies)

iNecas:
Update foreman-tasks to 0.7.2
Update dynflow to 0.8.2
Update concurrent-ruby-edge to 0.1.0
Update concurrent-ruby to 0.9.0

The deb PR on its way


domcleal:
foreman-tasks now ships a config file, could you install that into %{foreman_pluginconf_dir} (provided by foreman automatically, aka /etc/foreman/plugins)?  Mark it as a %config, and it's up to you whether you want it to be foreman-tasks.yaml (which will get updated as a %config automatically, if the user doesn't change it) or an .example file.

Do you also want to ship a cron.d file for the cleanup rake task?

iNecas:
@domcleal good point, I've updated the spec. I would prefer not shipping the cron file right away, leaving some time for feedback from users running it manually, the cron file would be added in some of the next versions

rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -155,6 +158,7 @@ exit 0
 %exclude %{gem_cache}
 %{gem_spec}
 %{foreman_bundlerd_plugin}
+%config(noreplace) %{foreman_pluginconf_dir}/%{gem_name}.yaml
```
domcleal: Is noreplace deliberate?  (I know we have this in error in some other packages, so just checking.)

noreplace means all updates go to .rpmnew, while %config alone would mean _unmodified_ files will be updated, but user-modified ones are left and the new copy becomes .rpmnew, IIRC.

rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -103,6 +103,9 @@ cp -a .%{gem_dir}/* \
 %foreman_bundlerd_file
 %foreman_precompile_plugin -a
 
+mv %{buildroot}/%{gem_instdir}/foreman-tasks.yaml.example \
```
domcleal: gem_name?

domcleal: Also missing the config/ subdir.

rubygem-foreman-tasks/rubygem-foreman-tasks.spec
iNecas: Inspired myself with foreman_custom_parameters, I can remove the noreplace

iNecas:
Updated

rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -103,6 +103,9 @@ cp -a .%{gem_dir}/* \
 %foreman_bundlerd_file
 %foreman_precompile_plugin -a
 
+mv %{buildroot}/%{gem_instdir}/config/%{gem_name}.yaml.example \
+  %{buildroot}%{foreman_pluginconf_dir}/%{gem_name}.yaml
```
domcleal: I think you need to `mkdir -p` this first.

iNecas: Right, it could be ok now.

domcleal:
Thanks @iNecas, merged.  Are you happy to tag (with --keep-version) and build these, or shall I?

iNecas:
I can do that, thanks!

domcleal:
Cheers!

PR #719, opened 2015-07-31
Update foreman-tasks to 0.7.2 (including related gems)

iNecas:


domcleal:
The gem.list needs updating for foreman-tasks, which causes the build error.  There are some questions as well on the other PR that we should answer before merging this one.

iNecas:
Thanks, I always forget something… I've answered the questions on #718 , how does that translate to changes on this PR?

domcleal:
We just updated the README to contain some instructions, should help next time!

In this one, you should be able to a) mv/cp the example file in `rules` to remove the extension, then b) add a line to plugins/ruby-foreman-tasks/debian/install to copy config/foreman-tasks.yaml to the final location (see debian/jessie/foreman/{rules,foreman.install} for an example).

iNecas:
Updated 
no-idea

domcleal:
Hm, not sure config/ is going to work actually for this case, since the package isn't using gem2deb.  We don't seem to have any other ruby-foreman-* plugins that do this, only the regular gem2deb type packages (dependencies, ruby-smart-proxy-*).  The config might have to be copied into the packaging, like ruby-foreman-default-hostgroup and then just refer to it as foreman-tasks.yaml in `install`.  Do you agree @mmoll?

mmoll:
Yes, this needs to go into -packaging, as we don't unpack the gems for core plugins.

iNecas:
Thanks for the pointers, lets see how this one goes…

domcleal:
Thanks @iNecas!  Building.

PR #725, opened 2015-08-04
Fixes #11229 (part 3): removed puppet classes cache directory creation

witlessbird:


witlessbird:
hmm, it got confused by the branch path. I'll close this and do it differently...

PR #726, opened 2015-08-04
fixes #11275, #11170 - pass spaces/quotes in args through SCL wrappers

domcleal:
Changes the ruby193/SCL wrappers to create a temporary file containing
all original arguments from the ruby193-* command and then executes it
via "scl enable".  This fixes two quoting-related bugs with these
wrappers.

1. #11170: the "scl enable" command has a bug (RHBZ#1248418) where it
   internally uses temporary shell scripts and doesn't escape double
   quotes, so JSON strings can't be passed through as arguments.
2. #11275: the ruby-wrappers passed through arguments word-by-word,
   causing arguments containing spaces to be split up.  It should have
   used the special quoted "$@" variable.

### DIFF
--- a/ruby-wrapper/ruby-wrapper
+++ b/ruby-wrapper/ruby-wrapper
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-COMMAND="ruby $@"
+COMMAND="ruby"
 SCL=FIXMESCL
 
-scl enable $SCL "$COMMAND"
-
+# Execute a single script with no arguments with 'scl', else it fails to
+# pass spaces/quotes in arguments through correctly (RHBZ#1248418).
+TMP=$(mktemp)
+trap "rm -f $TMP" EXIT
+echo -n $COMMAND > $TMP
+for arg in "$@"; do
+  printf " %q" "$arg" >> $TMP
+done
+scl enable $SCL "bash $TMP"
--- a/ruby-wrapper/ruby-wrapper-rails
+++ b/ruby-wrapper/ruby-wrapper-rails
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-COMMAND="rails $@"
+COMMAND="rails"
 SCL=FIXMESCL
 
-scl enable $SCL "$COMMAND"
-
+# Execute a single script with no arguments with 'scl', else it fails to
+# pass spaces/quotes in arguments through correctly (RHBZ#1248418).
+TMP=$(mktemp)
+trap "rm -f $TMP" EXIT
+echo -n $COMMAND > $TMP
+for arg in "$@"; do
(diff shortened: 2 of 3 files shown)


ekohl:
Would it make sense to rewrite this in a language where you have more control over the array? Otherwise I think this is the only option.

domcleal:
Well, the fix for the second bug is actually very easy in bash.  I'd have changed it to just `scl enable ruby193 -- ruby "$@"` if that was all I was trying to fix.  It'd handle the quoting - the problem was just use of unquoted `$@`.

The first bug is the one that required more changes.  Rewriting this wouldn't have helped I don't think, since the bug is actually in `scl` rather than our wrapper.  It's a C program that writes a temporary shell script that looks like this:

<pre>
# cat /var/tmp/scl1xifnT
eval "SCLS=( ${X_SCLS[*]} )"
/usr/bin/scl_enabled ruby193
if [ $? != 0 ]; then
  SCLS+=(ruby193)
  export X_SCLS=$(printf '%q ' "${SCLS[@]}")
. /opt/rh/ruby193//enable
fi
"ruby" "/bin/hammer" "puppet-class" "list" "--search" "name = ntp" 
</pre>

Unfortunately it's simply broken with regards to quoting, since it just puts each argument into double quotes and hopes for the best.  We could probably push some artificially escaped arguments to it to make it write the correct thing, but this might then break if `scl` was fixed.  (The version of scl-utils in Fedora has some support for newer style SCLs which activate and exec commands entirely in C, with no shell scripts, so doesn't have this bug.)

I don't think using a different language for the wrapper would help much.  In fact, I rely on bash's printf %q support to generate correctly quoted arguments in the generated script.

ekohl:
Fair enough. The whole rewrite to another language is done at another level then :) Maybe it would make sense to have the BZ reference in the comments.

domcleal:
Yep, I think the scl-utils "2014" versions are much better with their rewrite to activate SCLs entirely in C.  I've no idea whether they'll be made available to EL6/7 in the future though, or what we need to do to write an SCL that uses it.  I'll try and find out some more about it.

Thanks for the comment about comments, I've added the ticket number.

EL6 build: http://koji.katello.org/koji/taskinfo?taskID=312670
EL7 build: http://koji.katello.org/koji/taskinfo?taskID=312671

ekohl:
:+1:

ekohl:
> Yep, I think the scl-utils "2014" versions are much better with their rewrite to activate SCLs entirely in C. I've no idea whether they'll be made available to EL6/7 in the future though, or what we need to do to write an SCL that uses it. I'll try and find out some more about it.

Even if they become available on EL6/7, then it will take a long time till you can rely on it being available for everyone.

PR #730, opened 2015-08-04
Increase the epoch number for the concurrent-ruby gems

iNecas:
The previous version numbers (0.9.0.pre3/0.1.0.pre3) were considered higher 
from the yum perspective, than the 0.9.0/0.1.0


domcleal:
Thanks @iNecas, building.

PR #735, opened 2015-08-10
Add rubygem-foreman_remote_execution (RPM)

stbenjam:
Will add the gem once its on rubygems (today or tomorrow), but figured we could get a jumpstart on the spec review dance where I find out my plugin packaging knowledge is outdated and there's now more helpers for boilerplate :)


stbenjam:
@domcleal We'll need to merge these all in order when it comes time:

- [ ] SCL dynflow 0.8.4  https://github.com/theforeman/foreman-packaging/pull/754
- [ ] non-SCL dynflow 0.8.4 TBD
- [ ] foreman-tasks 0.7.3 https://github.com/theforeman/foreman-packaging/pull/750
- [ ] apipie-params 0.0.4 TBD
- [ ] smart_proxy_dynflow 0.0.2 https://github.com/theforeman/foreman-packaging/pull/737
- [ ] smart_proxy_remote_execution_ssh 0.0.1 https://github.com/theforeman/foreman-packaging/pull/737
- [ ] foreman_remote_execution 0.0.1 This PR.

stbenjam:
[test]

dependencies/Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
+
+gem 'gem2deb'
```
domcleal: Wrong branch!

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -0,0 +1,115 @@
...
+Name:       %{?scl_prefix}rubygem-%{gem_name}
+Version:    0.0.1
+Release:    1%{?foremandist}%{?dist}
+Group:      Applications/System
+License:    GPLv3
+URL:        http://github.com/theforeman/foreman_remote_execution
+Source0:    http://rubygems.org/downloads/%{gem_name}-%{version}.gem
+
+Requires:   foreman >= 1.9.0
+Requires:   foreman < 1.10.0
```
domcleal: This should be removed, since it'll be built to nightlies and the plugin itself specifies >= 1.9 only.

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -0,0 +1,115 @@
...
+BuildRequires: %{?scl_prefix}rubygem-foreman-tasks >= 0.7.3
+BuildRequires: %{?scl_prefix}rubygem-foreman-tasks < 0.8.0
+
+BuildArch: noarch
+
+Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
+Provides: foreman-plugin-remote_execution
+
+%description
+A plugin bringing remote execution to the Foreman, completing the config management functionality with remote management functionality
```
domcleal: Should be wrapped to 80 chars

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -0,0 +1,115 @@
...
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
+
+%foreman_bundlerd_file
+%foreman_precompile_plugin -a -s
+
+%posttrans
+# We need to run the db:migrate after the install transaction
+/usr/sbin/foreman-rake db:migrate  >/dev/null 2>&1 || :
```
domcleal: Use %{foreman_db_migrate} etc on all of these lines (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/foreman/foreman.spec#L581-L585).

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -0,0 +1,115 @@
...
+%{gem_instdir}/config
+%{gem_instdir}/db
+%{gem_instdir}/locale
+%{gem_instdir}/lib
+%{gem_instdir}/public
+%exclude %{gem_cache}
+%{gem_spec}
+%doc %{gem_instdir}/LICENSE
+%{foreman_bundlerd_plugin}
+%foreman_apipie_cache_foreman
```
domcleal: Use {} for consistency?

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -0,0 +1,115 @@
...
+
+%files
+%dir %{gem_instdir}
+%{gem_instdir}/app
+%{gem_instdir}/config
+%{gem_instdir}/db
+%{gem_instdir}/locale
+%{gem_instdir}/lib
+%{gem_instdir}/public
+%exclude %{gem_cache}
```
domcleal: already on line 107

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -0,0 +1,115 @@
...
+exit 0
+
+%files
+%dir %{gem_instdir}
+%{gem_instdir}/app
+%{gem_instdir}/config
+%{gem_instdir}/db
+%{gem_instdir}/locale
+%{gem_instdir}/lib
+%{gem_instdir}/public
```
domcleal: Not in the gem itself, and I think the assets line below will cover the specific bit needed for precompiled assets

domcleal:
1) missing a dependency on rubygem(rails), pinned to ~> 3.2.8
2) doc/ looks like it should perhaps be added to the gem and stored in the -doc subpackage

stbenjam:
>  2) doc/ looks like it should perhaps be added to the gem and stored in the -doc subpackage

We're not actually using it, only a handful of methods out of 100+ are documented.  

Also, I don't notice in other foreman rpm's where you're actually doing anything with the %doc directory.  Shouldn't it be compiled with rdoc and installed into `%{_defaultdocdir}` if we're going to really include it?

domcleal:
> Also, I don't notice in other foreman rpm's where you're actually doing anything with the doc/ directory. Shouldn't it be compiled with rdoc and installed into %{_defaultdocdir} if we're going to really include it?

Sure, if it's provided from the gem then we add --rdoc to gem install and it should be installed into place.  I wasn't sure if it was in use - only if it is then it'd be useful to include.  It'd be gem_docdir rather than defaultdocdir, IIRC.

Build's red.

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -0,0 +1,116 @@
...
+%else
+BuildRequires: %{?scl_prefix}ruby(abi) >= %{rubyabi}
+%endif
+BuildRequires: foreman-plugin >= 1.9.0
+BuildRequires: foreman-assets
+BuildRequires: %{?scl_prefix}rubygems-devel
+BuildRequires: %{?scl_prefix}rubygems
+BuildRequires: %{?scl_prefix}rubygem-foreman-tasks >= 0.7.3
+BuildRequires: %{?scl_prefix}rubygem-foreman-tasks < 0.8.0
+BuildRequires: %{?scl_prefix}rubygem-rails >= 3.2.8
```
domcleal: Use rubygem(rails), not rubygem-rails

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -0,0 +1,116 @@
...
+%{foreman_bundlerd_plugin}
+%{foreman_apipie_cache_foreman}
+%{foreman_apipie_cache_plugin}
+%{foreman_assets_plugin}
+%{gem_instdir}/Rakefile
+%exclude %{gem_cache}
+%exclude %{gem_instdir}/test
+
+%files doc
+%doc %{gem_instdir}/doc
```
domcleal: I don't think you want to add that yet, only if/when it's actually provided.

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
stbenjam: Same for tasks?

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
stbenjam: Yea the remote_execution PR's open, I'll update the gem or remove this depending on the outcome

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
domcleal: Yeah, I'd prefer it but it's not as important.  The reason it fails is because the rubygem-rails RPM has an epoch set, so it's important to compare the version of the virtual rubygem(rails) rather than the actual package EVR.

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
stbenjam: Without this:

```
   /opt/rh/ruby193/root/usr/share/gems/gems/foreman_remote_execution-0.0.2/public/assets/template_input-c2a9e2f925179b669abb91f1a3230c43.js
```

Is there a problem with the assets macro?

stbenjam:
In addition to the comment above about `public`, `%{gem_docdir}` doesn't seem to work without also including `%{gem_instdir}/doc` either.

It's green now, though.

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
domcleal: Hm, so the macro is `%%%{name}_assets_plugin %%{gem_instdir}/public/assets/%%{gem_name}`, which means the assets are outside of the foreman_remote_execution namespace - they ought to be moved into a subdirectory inside app/assets/*.

stbenjam: @domcleal I should do that? Not sure what I understand what changes I need to make.

[foreman_salt includes `/public`](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-foreman_salt/rubygem-foreman_salt.spec) and that RPM works fine.

domcleal: I think the assets should move to app/assets/javascripts/foreman_remote_execution/ in the repo/gem instead of being in app/assets/javascripts/ only, which will prevent them perhaps clashing with any names in Foreman itself.

foreman_salt does include that line, though it also has assets_plugin listed so I suspect that it isn't actually doing anything extra because foreman_salt's assets are properly namespaced.

stbenjam: Oh I understand, thanks.  foreman_salt's mostly in its own namespace, but it's intentional that remote_execution isn't, to supposedly make it easier some day to put in core.

Wouldn't it be odd to namespace the assets but not everything else?

domcleal: Yes, namespacing assets would be odd if the rest wasn't - I hadn't seen that this had been omitted.

That sounds like a bad idea, we try to support and encourage namespacing for development of plugins to prevent things clashing.  If plugin authors don't namespace then we can't prevent clashes between core and the plugin or between plugins.  Things can be renamed if it was being merged to core, that's not difficult.

So be it, but I think it's a bad idea to build a new plugin in this way.

stbenjam: ` ¯\_(ツ)_/¯` I just work here.  I asked when I saw the lack of name spacing and that's what they told me.

ares: The reason for not namespacing was that the templates STI and tooling didn't work with namespaced JobTemplate model. Since most of objects starts with Job prefix and we felt we might merge remote execution to core later we decided not to put everything under RemoteExecution namespace.

domcleal: > The reason for not namespacing was that the templates STI and tooling didn't work with namespaced JobTemplate model.

That ought to be fixed then...

You'll also have to fix this in the Debian package I suspect, since it probably isn't copying the assets into place if they're unnamespaced.

rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -0,0 +1,118 @@
...
+%{foreman_apipie_cache_plugin}
+%{foreman_assets_plugin}
+%{gem_instdir}/Rakefile
+%exclude %{gem_cache}
+%exclude %{gem_instdir}/test
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/doc
+%doc %{gem_instdir}/LICENSE
```
domcleal: LICENSE needs to remain in the main package

domcleal:
Thanks @stbenjam, will build it.  Please note the issue noted about the Debian package assets, I'd bet they're not in the final package as they're unnamespaced.

stbenjam:
Thanks, will check

PR #736, opened 2015-08-10
Add smart_proxy_dynflow (RPM)

stbenjam:
Todo:

- [x] Dynflow gem built non-SCL for el7 plugins


domcleal:
As per https://groups.google.com/d/msg/foreman-dev/RUeJtbnGBBA/5ROD7vMRJAAJ, I've made changes to our tag configuration on Koji.  In short, you now have two whitelists in tito.props for plugins, one for SCL and one for non-SCL so a package can be added to and built in both configurations.  Hopefully that'll enable you to build out the dep tree required for this package.

stbenjam:
@domcleal thanks!

stbenjam:
Checklist:

- [ ] algebrick https://github.com/theforeman/foreman-packaging/pull/769
- [ ] concurrent-ruby
- [ ] concurrent-ruby-edge	
- [ ] apipie-params
- [ ] dynflow

stbenjam:
Need to remove the Gemfile from the spec, and add the README.

https://github.com/theforeman/smart_proxy_dynflow/pull/5

domcleal:
It looks like this should be v0.0.3.

stbenjam:
@domcleal And missed some dependencies too. Will it ever end?!

This one's green.

comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -47,8 +47,10 @@
       <packagereq type="default">ruby193-rubygem-puppetdb_foreman</packagereq>
       <packagereq type="default">rubygem-foreman_scap_client</packagereq>
       <packagereq type="default">rubygem-newt</packagereq>
+      <packagereq type="default">rubygem-sequel</packagereq>
```
domcleal: In the lower dependencies section please.

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -0,0 +1,88 @@
...
+
+cp -pa .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+mkdir -p %{buildroot}%{foreman_proxy_bundlerd_dir}
+cp -pa .%{gem_instdir}/bundler.d/dynflow.rb %{buildroot}%{foreman_proxy_bundlerd_dir}
+mkdir -p  %{buildroot}%{foreman_proxy_settingsd_dir}
+cp -pa .%{gem_instdir}/settings.d/dynflow.yml.example %{buildroot}%{foreman_proxy_settingsd_dir}/dynflow.yml
+mkdir -p  %{buildroot}%{dynflow_config_dir}
+mkdir -p %{buildroot}%{_bindir}
```
domcleal: Unused?

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -0,0 +1,88 @@
...
+%prep
+
+%setup -q -c -T
+%gem_install -n %{SOURCE0}
+
+%build
+
+%install
+mkdir -p %{buildroot}%{gem_dir}
+mkdir -p %{buildroot}%{_sbindir}
```
domcleal: Unused?

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -0,0 +1,88 @@
...
+
+cp -pa .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+mkdir -p %{buildroot}%{foreman_proxy_bundlerd_dir}
+cp -pa .%{gem_instdir}/bundler.d/dynflow.rb %{buildroot}%{foreman_proxy_bundlerd_dir}
+mkdir -p  %{buildroot}%{foreman_proxy_settingsd_dir}
+cp -pa .%{gem_instdir}/settings.d/dynflow.yml.example %{buildroot}%{foreman_proxy_settingsd_dir}/dynflow.yml
+mkdir -p  %{buildroot}%{dynflow_config_dir}
```
domcleal: Unused?

comps/comps-foreman-plugins-rhel7.xml
stbenjam: rubygem-newt isn't a core package either, not sure I understand the sections in comps here

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -0,0 +1,88 @@
...
+mkdir -p %{buildroot}%{_bindir}
+
+%files
+%dir %{gem_instdir}
+%{gem_instdir}/lib
+%{gem_instdir}/bundler.d
+%{gem_instdir}/settings.d
+%{foreman_proxy_bundlerd_dir}/dynflow.rb
+%config(noreplace) %{_sysconfdir}/foreman-proxy/settings.d/dynflow.yml
```
domcleal: I'd suggest plain %config, then it'll get updated on upgrade *if* it hasn't been changed.

comps/comps-foreman-plugins-rhel7.xml
stbenjam: Shall I move it lower as well?

domcleal: It's top-level in the sense that we don't have another plugin depending directly on it, so it's just listed here which'll help us remember not to remove it while tidying.  It's used inside the Discovery image.

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -0,0 +1,85 @@
...
+mkdir -p  %{buildroot}%{foreman_proxy_settingsd_dir}
+cp -pa .%{gem_instdir}/settings.d/dynflow.yml.example %{buildroot}%{foreman_proxy_settingsd_dir}/dynflow.yml
+
+%files
+%dir %{gem_instdir}
+%{gem_instdir}/lib
+%{gem_instdir}/bundler.d
+%{gem_instdir}/settings.d
+%{foreman_proxy_bundlerd_dir}/dynflow.rb
+%config %{_sysconfdir}/foreman-proxy/settings.d/dynflow.yml
```
domcleal: Use the macro defined up at the top: foreman_proxy_settingsd_dir

domcleal:
Please also add to the fedora19 comps and whitelists, it'll be fine on that platform.

stbenjam:
@domcleal I would have to go back and add all the dependencies for dynflow and all the way down, and that's a huge time investment for a dead platform.

domcleal:
They should already be there, since Fedora doesn't use SCLs and already has foreman-tasks.

stbenjam:
It's a dead platform, for over a year - why are we still maintaining it?

domcleal:
19, or Fedora generally?  If 19, it's because we technically can't run on anything newer right now.  If Rails 4 ever happens, then we'll be able to jump straight to supporting something newer, such as 22 or 23.

If we removed Fedora entirely from our setup, we'd probably never put the effort in to reinstate it.  While we don't really actively support 19, just keeping the packages flowing means it'll be much easier to support a newer version once the Rails issue's fixed.

stbenjam:
I thought Fedora rails moves too fast, and you'd end up having to do SCL anyway, which means profound changes to move from the 19 comps to some f23(24,25, whenever) setup.

stbenjam:
Anyway, updated

domcleal:
> I thought Fedora rails moves too fast, and you'd end up having to do SCL anyway, which means profound changes to move from the 19 comps to some f23(24,25, whenever) setup.

Yeah, it does move fast - well, faster than we can move the SCL based OSes to new versions.  I'm optimistic in the long term though that either SCL development will accelerate under the CentOS SCL SIG or that SCLs will be maintained for Fedora.  The latter would be a significant change, but it's probably still easier than dropping and then reinstating it later.  Thanks for entertaining the idea.

domcleal:
Also we're stuck on Fedora 19 through our own fault - since SCLs for Rails 4 have been available for a year or two.  If we'd been on the ball, we could have been on 20 and moving to 21/22 with Rails 4.1 now.

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -0,0 +1,85 @@
+%global gem_name smart_proxy_dynflow
+
+%global foreman_proxy_dir /usr/share/foreman-proxy
+%global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
+%global foreman_proxy_settingsd_dir %{_sysconfdir}/foreman-proxy/settings.d
+
+%global dynflow_config_dir %{_sysconfdir}/dynflow
```
domcleal: Unused macro?

domcleal:
Thanks @stbenjam, building.

PR #739, opened 2015-08-11
add rubygem-foreman_noenv RPM spec

joshuabaird:
Add a new RPM spec file for the foreman_noenv plugin.


domcleal:
1) If you can mark your membership as public on this page, then your PR will get tested automatically: https://github.com/orgs/theforeman/people
2) Can you start by adding the new packages to comps/comps-foreman-plugins*.xml and rel-eng/tito.props under the three plugin sections (just copy/paste)?

comps/comps-foreman-plugins-rhel6.xml
```diff
@@ -25,6 +25,7 @@
       <packagereq type="default">ruby193-rubygem-foreman_host_rundeck</packagereq>
       <packagereq type="default">ruby193-rubygem-foreman-mco</packagereq>
       <packagereq type="default">ruby193-rubygem-foreman_memcache</packagereq>
+      <packagereq type="default">rubygem-foreman_noenv</packagereq>
```
domcleal: ruby193

comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -25,6 +25,7 @@
       <packagereq type="default">ruby193-rubygem-foreman_host_rundeck</packagereq>
       <packagereq type="default">ruby193-rubygem-foreman-mco</packagereq>
       <packagereq type="default">ruby193-rubygem-foreman_memcache</packagereq>
+      <packagereq type="default">rubygem-foreman_noenv</packagereq>
```
domcleal: ruby193

domcleal:
The current failure's on missing the gem.  You'd need git-annex to add it to the repo (available in EPEL etc).  Just run: `spectool -g *.spec && git annex add *.gem` and commit it.

rubygem-foreman_noenv/rubygem-foreman_noenv.spec
```diff
@@ -0,0 +1,110 @@
...
+        %{buildroot}%{gem_dir}/
+
+mkdir -p %{buildroot}%{foreman_bundlerd_dir}
+cat <<GEMFILE > %{buildroot}%{foreman_bundlerd_dir}/%{gem_name}.rb
+gem '%{gem_name}'
+GEMFILE
+
+mkdir -p %{buildroot}%{foreman_pluginconf_dir}
+mv %{buildroot}%{gem_instdir}/config/foreman_noenv.yaml \
+   %{buildroot}%{foreman_pluginconf_dir}/%{gem_name}.yaml.example
```
domcleal: I'd suggest calling this just %{gem_name}.yaml so it's in the right location by default.  It'll help users when the plugin automatically activates on installation I think.

rubygem-foreman_noenv/rubygem-foreman_noenv.spec
```diff
@@ -0,0 +1,110 @@
...
+%{gem_instdir}/app
+%{gem_instdir}/lib
+%{gem_instdir}/db/migrate
+%{gem_instdir}/locale
+
+%exclude %{gem_cache}
+%exclude %{gem_instdir}/test
+%{gem_spec}
+%{foreman_bundlerd_dir}/%{gem_name}.rb
+%doc %{foreman_pluginconf_dir}/%{gem_name}.yaml.example
```
domcleal: ... and then here you can use %config rather than %doc so it's treated as a config file.

rubygem-foreman_noenv/rubygem-foreman_noenv.spec
```diff
@@ -0,0 +1,110 @@
...
+%doc %{gem_instdir}/LICENSE
+
+%exclude %{gem_dir}/cache/%{gem_name}-%{version}.gem
```
domcleal: Not needed, this is already on line 94 (%{gem_cache})

domcleal:
Looks good!  Please also add the -doc subpackages to comps/.  You can do this quickly by running `./comps_doc.sh` in the root of the repo, it'll update it automatically.

domcleal:
One last bit, the package needs to run db:migrate after installation else the server fails with a column missing error.  This normally goes between %files doc and %changelog:

<pre>
%posttrans
/usr/sbin/foreman-rake db:migrate  >/dev/null 2>&1 || :
(/sbin/service foreman status && /sbin/service foreman restart) >/dev/null 2>&1
exit 0
</pre>

domcleal:
Merged as c573f880.  Thanks very much @joshuabaird!  I'm building it into nightly and 1.9 repos.

PR #740, opened 2015-08-11
Add smart_proxy_dynflow (DEB)

stbenjam:


plugins/smart_proxy_dynflow/control
```diff
@@ -0,0 +1,15 @@
...
+Maintainer: Stephen Benjamin <stephen@redhat.com>
+Build-Depends: debhelper, gem2deb
+Standards-Version: 3.9.3
+Homepage: https://github.com/theforeman/smart_proxy_dynflow
+XS-Ruby-Versions: all
+
+Package: ruby-smart-proxy-dynflow
+Architecture: all
+XB-Ruby-Versions: ${ruby:Versions}
+Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter, foreman-proxy (>= 1.9.0~rc3), ruby-dynflow
```
domcleal: The ruby-dynflow dependency should match the `~> 0.8.4` dep on this gem.

plugins/smart_proxy_dynflow/copyright
```diff
@@ -0,0 +1,28 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: ruby-smart-proxy-dynflow
+Source: https://github.com/theforeman/smart_proxy_dynflow
+
+Files: *
+Copyright: 2012-2013 Red Hat Inc. <foreman-dev@googlegroups.com>
```
domcleal: Probably 2015?

plugins/smart_proxy_dynflow/control
domcleal: Also seems to be missing the sequel dep.

domcleal: And sqlite3?

plugins/smart_proxy_dynflow/ruby-smart-proxy-dynflow.docs
```diff
@@ -0,0 +1 @@
+README.md
```
domcleal: I don't think this file's in the gem, so it's causing a build error.

domcleal:
Merged as 3f9e014, building now.  Thanks @stbenjam.

PR #742, opened 2015-08-11
Remove dependency on Foreman from dynflow (DEB)

stbenjam:
We need dynflow on the smart proxies.  Dynflow doesn't require foreman, but I have some questions about to handle this:

- Can we just call this ruby-dynflow? It's not a foreman package, but no idea how to "rename" a deb package and obsolete the other one

- What about postinst? It's using foreman ruby, can we make it only use system ruby?

### DIFF
--- a/plugins/ruby-foreman-dynflow/debian/changelog
+++ b/plugins/ruby-foreman-dynflow/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-dynflow (0.8.1-2) stable; urgency=low
+
+  * 0.8.2-2 released
+
+ -- Stephen Benjamin <stephen@bitbin.de>  Tue, 11 Aug 2015 12:27:00 -0500
+
 ruby-foreman-dynflow (0.8.2-1) stable; urgency=low
 
   * 0.8.2-1 released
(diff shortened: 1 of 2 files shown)


plugins/ruby-foreman-dynflow/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-foreman-dynflow (0.8.1-2) stable; urgency=low
```
domcleal: 0.8.2

domcleal:
> Can we just call this ruby-dynflow? It's not a foreman package, but no idea how to "rename" a deb package and obsolete the other one

It is based on Foreman though, since it installs a gem that will only ever work with the Foreman installation.  Check debian/install, it puts a bundler.d entry in and installs into Foreman's bundler cache.

I think if you wanted to use dynflow with the smart proxy, you'd introduce an additional package that's a traditional gem2deb so the smart proxy (which doesn't use bundler, only bundler_ext) can load it.  This package would still be required for Foreman core.

> What about postinst? It's using foreman ruby, can we make it only use system ruby?

foreman-ruby is also required as on Ubuntu 12.04 we can't use the default Ruby 1.8.7 for Foreman's bundler commands.

domcleal:
Just adding @mmoll's unanswered question on -dev:

    100509 < mmoll> regarding ##foreman-packaging/742 - is that dynflow executor from foreman-tasks that has to run in the Rails environment of Foreman ok when just having dynflow availabel as a gem? - basically I think that for -proxy dynflow needs to be there as gem2deb package (like hammer), the only question is if that new dynflow package is sufficient for foreman-tasks

stbenjam:
Thanks, I'll reopen a PR for a dependency package.

PR #743, opened 2015-08-12
Updated foreman_discovery to 4.0.0

lzap:
I pushed 4.0.0 with few harmless locale POX files. Made a patch in discovery repo to prevent this in the future: https://github.com/theforeman/foreman_discovery/pull/207

```
Submitting build with: koji -c ~/.koji/katello-config build --nowait --scratch foreman-plugins-nightly-fedora19 /tmp/tito/rubygem-foreman_discovery-4.0.0-1.fm1_10.git.383.3e95ee3.fc19.src.rpm
Uploading srpm: /tmp/tito/rubygem-foreman_discovery-4.0.0-1.fm1_10.git.383.3e95ee3.fc19.src.rpm
[                                    ]  00% 00:00:00     0.00 B        - B/sec
[====================================] 100% 00:00:01 100.01 KiB  82.71 KiB/sec
Created task: 316911
Task info: http://koji.katello.org/koji/taskinfo?taskID=316911
Wrote: /tmp/tito/ruby193-rubygem-foreman_discovery-4.0.0-1.fm1_10.git.383.3e95ee3.git.383.3e95ee3.el6.src.rpm



Submitting build with: koji -c ~/.koji/katello-config build --nowait --scratch foreman-plugins-nightly-rhel6 /tmp/tito/ruby193-rubygem-foreman_discovery-4.0.0-1.fm1_10.git.383.3e95ee3.git.383.3e95ee3.el6.src.rpm
Uploading srpm: /tmp/tito/ruby193-rubygem-foreman_discovery-4.0.0-1.fm1_10.git.383.3e95ee3.git.383.3e95ee3.el6.src.rpm
[                                    ]  00% 00:00:00     0.00 B        - B/sec
[====================================] 100% 00:00:01 100.09 KiB  89.06 KiB/sec
Created task: 316913
Task info: http://koji.katello.org/koji/taskinfo?taskID=316913
Wrote: /tmp/tito/ruby193-rubygem-foreman_discovery-4.0.0-1.fm1_10.git.383.3e95ee3.git.383.3e95ee3.git.383.3e95ee3.el7.src.rpm



Submitting build with: koji -c ~/.koji/katello-config build --nowait --scratch foreman-plugins-nightly-rhel7 /tmp/tito/ruby193-rubygem-foreman_discovery-4.0.0-1.fm1_10.git.383.3e95ee3.git.383.3e95ee3.git.383.3e95ee3.el7.src.rpm
[...]

PR #744, opened 2015-08-12
Updated foreman_discovery to 4.0.0

lzap:
RPMs: https://github.com/theforeman/foreman-packaging/pull/743

### DIFF
--- a/plugins/ruby-foreman-discovery/debian/changelog
+++ b/plugins/ruby-foreman-discovery/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-discovery (4.0.0) stable; urgency=low
+
+  * 4.0.0 released
+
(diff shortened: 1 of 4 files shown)


mmoll:
working fine here, is ```foreman (>= 1.8.0~rc1)``` still correct?

lzap:
> working fine here, is ```foreman (>= 1.8.0~rc1)``` still correct?

Oh right, not at all. Changed to `1.9.0~rc1`. Will that hit rc2, 3 and
the final release? I think so.

-- 
Later,
 Lukas #lzap Zapletal

domcleal:
It will match rc2, 3 etc just fine.  You ought to fix that in engine.rb then, since the plugin still states that it's compatible with 1.8.

domcleal:
Thanks @lzap, building for develop and 1.9.

PR #745, opened 2015-08-12
Package dynflow as dependency (DEB)

stbenjam:
Checklist:

- [x] apipie-params https://github.com/theforeman/foreman-packaging/pull/764 READY
- [x] algebrick https://github.com/theforeman/foreman-packaging/pull/761
- [x] concurrent-ruby https://github.com/theforeman/foreman-packaging/pull/762
- [x] concurrent-ruby-edge https://github.com/theforeman/foreman-packaging/pull/763
- [X] multi_json ~~https://github.com/theforeman/foreman-packaging/pull/765~~ Already packaged
- [x] dynflow. This PR!


stbenjam:
Do note precise is failing as its intentionally left out, we can't do < 1.9.3.

mmoll:
The dependencies of dynflow also need to get packaged that way.

IMHO we can live with precise not being supported here, otherwise it would be needed to package -proxy and plugin with ruby1.9.3 from the brightbox PPA.

domcleal:
> The dependencies of dynflow also need to get packaged that way.

Which is quite a few, both for Debian and RPMs.  It's probably the largest package anybody's tried to put in the smart proxy...

At least AS is available for Debian/Ubuntu already I suppose.

stbenjam:
How can I easily see what's already packaged for Debian? Is it really that many? It should just be these, right?


```
  s.add_dependency "activesupport"
  s.add_dependency "multi_json"
  s.add_dependency "apipie-params"
  s.add_dependency "algebrick", '~> 0.7.0'
  s.add_dependency "concurrent-ruby", '~> 0.9.0'
  s.add_dependency "concurrent-ruby-edge", '~> 0.1.0'
```

stbenjam:
And for RPM we should have everything already, we just need a non-scl tag to put them in, no?

domcleal:
> How can I easily see what's already packaged for Debian? Is it really that many? It should just be these, right?

Yes, https://rubygems.org/gems/dynflow shows it + click through for further deps.  Use http://packages.debian.org/foo (and ubuntu.com) to search for packages.

> And for RPM we should have everything already, we just need a non-scl tag to put them in, no?

Well, AS comes from the SCL normally so that'd have to be repackaged (which is a bit iffy, frankly as it could interfere with things, something for that PR).

stbenjam:
Alright so for debian dependencies... do I need to go out and package the gems individually? I'll need to do concurrent-ruby, algebrick, and maybe a few others.

Foreman dynflow taks this strategy: https://github.com/theforeman/foreman-packaging/blob/deb/develop/plugins/ruby-foreman-dynflow/debian/gem.list

Can that be done here or is that something special about how Foreman gems work?

domcleal:
The foreman package is a vendored bundler installation, while foreman-proxy is a regular gem2deb style set of individually packaged gems using bundler_ext.  So yeah, every dependency would need packaging separately.

stbenjam:
Checklist:

- [ ] algebrick https://github.com/theforeman/foreman-packaging/pull/761
- [ ] concurrent-ruby https://github.com/theforeman/foreman-packaging/pull/762
- [ ] concurrent-ruby-edge
- [ ] apipie-params
- [ ] multi_json (already packaged)
- [ ] dynflow. This PR!

stbenjam:
:green_apple:

mmoll:
merged as c8317b1, thanks @stbenjam :tada: 
building...

PR #749, opened 2015-08-13
Release foreman_salt 3.0.1 (RPM)

stbenjam:
Nightly + 1.9


domcleal:
Thanks @stbenjam, will build for both.

Nitpick, if you can avoid updating the changelogs it makes it easier to cherry pick across branches when there are changes.  tito'll generate it when tagging.  Thanks!

stbenjam:
Will do

PR #751, opened 2015-08-13
Update foreman-tasks to 0.7.3 (DEB)

stbenjam:


stbenjam:
We'll need the new dynflow first, will hold off until tomorrow as we're due for another release to remove activesupport.

domcleal:
Don't forget to update the version number in both deps on dynflow in the control file.

mmoll:
merged as 84b385d, thanks @stbenjam!

PR #756, opened 2015-08-14
Release apipie-params 0.0.4 (RPM)

stbenjam:


iNecas:
It doesn't require 'activesupport' anymore: the spec should be updated

stbenjam:
Thanks, fixed

iNecas:
ACK

domcleal:
Thanks @stbenjam, building.

PR #768, opened 2015-08-19
Add dynflow to non-SCL tag

stbenjam:


stbenjam:
@domcleal Will this do it or do I need to do something else?

domcleal:
You'll also need to add it to comps/, and ensure the packages actually build.  I don't think our PR test will run unless you actually modified the specs in some way, so you will probably need to run `tito release koji-foreman-plugins --scratch --test` yourself and link the scratch builds here to check.

comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -45,6 +45,11 @@
       <packagereq type="default">ruby193-rubygem-hammer_cli_foreman_tasks</packagereq>
       <packagereq type="default">ruby193-rubygem-ovirt_provision_plugin</packagereq>
       <packagereq type="default">ruby193-rubygem-puppetdb_foreman</packagereq>
+      <packagereq type="default">rubygem-algebrick</packagereq>
```
stbenjam: http://koji.katello.org/koji/taskinfo?taskID=321784

stbenjam:
I'll do one at a time... requires spec changes.

PR #769, opened 2015-08-19
Package algebrick for non-SCL el7

stbenjam:


comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -45,6 +45,7 @@
       <packagereq type="default">ruby193-rubygem-hammer_cli_foreman_tasks</packagereq>
       <packagereq type="default">ruby193-rubygem-ovirt_provision_plugin</packagereq>
       <packagereq type="default">ruby193-rubygem-puppetdb_foreman</packagereq>
+      <packagereq type="default">rubygem-algebrick</packagereq>
```
domcleal: It's not clear, but the upper group is top-level packages (plugins) while the bottom set, where algebrick should be, is for dependencies.

Also run ./comps_doc.sh and it'll add the -doc subpackage.

rubygem-algebrick/rubygem-algebrick.spec
```diff
@@ -15,13 +15,15 @@ Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 Requires: %{?scl_prefix}ruby(rubygems)
 Requires: %{?scl_prefix}ruby
 Requires: %{?scl_prefix}rubygem(activesupport)
```
domcleal: This dependency needs to be removed, seems it went a while ago and we missed it.

stbenjam:
Moved to https://github.com/theforeman/foreman-packaging/pull/772

PR #770, opened 2015-08-19
Package concurrent-ruby for non-SCL el7

stbenjam:


comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -79,6 +79,7 @@
       <packagereq type="default">ruby193-rubygem-opennebula</packagereq>
       <packagereq type="default">ruby193-rubygem-wicked</packagereq>
       <packagereq type="default">ruby193-rubygem-zscheduler</packagereq>
+      <packagereq type="default">rubygem-concurrent-ruby</packagereq>
```
domcleal: after chef-api

domcleal:
Thanks @stbenjam, building.

PR #772, opened 2015-08-19
Package dynflow and dependencies for non-SCL on el7

stbenjam:


domcleal:
:+1: on rebase.

stbenjam:
Moved all here because of we'll need a rebase every time...

stbenjam:
@domcleal Green

domcleal:
Thanks @stbenjam, building.

PR #778, opened 2015-08-21
Update deface to 1.0.1 (deb)

domcleal:
Change core Foreman packages to ship nokogiri 1.6.x as a requirement
for deface 1.x.


domcleal:
Seems OK when doing a non-deface 1.8 installation and upgrading it to this Foreman scratch build, adding column view 0.2.1 and the deface scratch build.  (Sorry, I messed up doing a full upgrade test.)

mmoll:
Working just fine here, thanks @domcleal!
Will build the deface package now and hope the breakage in the timespan without core packages containing the correct nokogiri will be limited.

domcleal:
I've triggered a rebuild of Foreman itself, so that should be fixed today.

PR #780, opened 2015-08-23
libaugeas-ruby -> ruby-augeas for Debian/jessie

mmoll:
As libaugeas-ruby is just a transitional package depending on ruby-augeas this is a noop on Debian/jessie, but makes the -proxy package installable on testing/unstable again.

I'd like to see this also in 1.9, so 1.9.1 packages will get built with this change.

While here, also add the dependency for the trusty package.


domcleal:
Thanks!

domcleal:
Also merged as 0175172 to deb/1.9.

PR #782, opened 2015-08-24
Update remote_execution to 0.0.4 (DEB)

stbenjam:


plugins/ruby-foreman-remote-execution/debian/install
```diff
@@ -1,4 +1,4 @@
 foreman_remote_execution.rb           usr/share/foreman/bundler.d
 cache                                 usr/share/foreman/vendor
-foreman_remote_execution              var/lib/foreman/public/assets
+assets                                var/lib/foreman/public
```
stbenjam: foreman_remote_execution assets aren't namespaced, so we have to do something like this.

Not sure about the implications of this though - will non-plugin assets end up in the DEB too?

stbenjam:
The errors don't seem related, something about foreman not being configured.  Any ideas?

domcleal:
Seems we have multiple package builds now failing, including ruby-foreman-salt and -tasks.  I'm trying to narrow it down and reproduce the issue.

domcleal:
It's my nokogiri related change, missing a runtime dep for the gem to build.  I'll send a PR to fix it tomorrow.

stbenjam:
@domcleal Thanks for figuring it out! :smile:

plugins/ruby-foreman-remote-execution/debian/install
domcleal: It should be OK, but you also need to update the `rules` so it copies the whole assets directory (which is beneath GEM_PATH, so will only be this plugin's assets).  We'll have to check the resulting build contents.

domcleal:
[test]

domcleal:
[test] as the nightly packages hadn't been pushed yet.

plugins/ruby-foreman-remote-execution/debian/install
domcleal: Indeed, failure's probably because a corresponding change is needed to `rules`.

domcleal: Looks okay to me:

<pre>
./var/lib/foreman/public/assets
./var/lib/foreman/public/assets/foreman_remote_execution
./var/lib/foreman/public/assets/foreman_remote_execution/manifest.yml
./var/lib/foreman/public/assets/job_invocations-99630200907501afe5bc9b2593ee2bf7.css
./var/lib/foreman/public/assets/job_invocations-99630200907501afe5bc9b2593ee2bf7.css.gz
./var/lib/foreman/public/assets/job_invocations.css
./var/lib/foreman/public/assets/job_invocations.css.gz
./var/lib/foreman/public/assets/template_input-81fd594267126475ae727e179f34b397.js
./var/lib/foreman/public/assets/template_input-81fd594267126475ae727e179f34b397.js.gz
./var/lib/foreman/public/assets/template_input.js
./var/lib/foreman/public/assets/template_input.js.gz
./var/lib/foreman/public/assets/template_invocation-07fde21d2fcad9abc6bacd5089d8649e.css
./var/lib/foreman/public/assets/template_invocation-07fde21d2fcad9abc6bacd5089d8649e.css.gz
./var/lib/foreman/public/assets/template_invocation-3f60f1b8b6ea70f9814826eb94936af4.js
./var/lib/foreman/public/assets/template_invocation-3f60f1b8b6ea70f9814826eb94936af4.js.gz
./var/lib/foreman/public/assets/template_invocation.css
./var/lib/foreman/public/assets/template_invocation.css.gz
./var/lib/foreman/public/assets/template_invocation.js
./var/lib/foreman/public/assets/template_invocation.js.gz
</pre>

stbenjam: Yup :+1:  Thanks for checking

mmoll:
tested, looking good, merged as 82c4f0a. thanks @stbenjam!

PR #791, opened 2015-08-30
Package smart_proxy_dns_powerdns

ekohl:


ekohl:
I should note that https://packages.debian.org/jessie/ruby-mysql2 is only available on Jessie now.

plugins/smart_proxy_dns_powerdns/watch
```diff
@@ -0,0 +1,2 @@
+version=3
```
mmoll: this file can get deleted

plugins/smart_proxy_dns_powerdns/changelog
```diff
@@ -0,0 +1,5 @@
+ruby-smart-proxy-dns-powerdns (0.1.0-0) stable; urgency=low
```
mmoll: the revision should be -1

mmoll:
:+1: after changing the revision

ekohl:
Updated and also copied ruby-mysql2 from wheezy backports.

mmoll:
The copied ruby-mysql2 could have some files deleted as we don't run the tests after building the gem normally, but as it's an 1:1 copy I'm fine with it. :+1:

plugins/smart_proxy_dns_powerdns/control
```diff
@@ -0,0 +1,18 @@
...
+Homepage: https://github.com/theforeman/smart_proxy_dns_powerdns
+XS-Ruby-Versions: all
+
+Package: ruby-smart-proxy-dns-powerdns
+Architecture: all
+XB-Ruby-Versions: ${ruby:Versions}
+Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter, foreman-proxy (>= 1.1.10~rc1), ruby-mysql2
+Recommends: pdns-server
+Description: PowerDNS DNS provider plugin for Foreman's smart proxy
+ PowerDNS DNS provider plugin for Foreman's smart proxy
```
domcleal: Duplicate line?

plugins/smart_proxy_dns_powerdns/copyright
```diff
@@ -0,0 +1,28 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: smart_proxy_dns_powerdns
+Source: https://github.com/theforeman/smart_proxy_dns_powerdns
+
+Files: *
+Copyright: 2015 Red Hat Inc. <foreman-dev@googlegroups.com>
```
domcleal: Perhaps just yourself, or Foreman developers?  Ditto below.

plugins/smart_proxy_dns_powerdns/control
ekohl: This was generated by gem2deb and the salt plugin has the same thing, so I left it. @mmoll?

plugins/smart_proxy_dns_powerdns/copyright
ekohl: Tbh I didn't really think of a licence and don't know what the policy is. I'm fine with just my name there but I don't care that much as long as the plugin is used and open :)

plugins/smart_proxy_dns_powerdns/control
mmoll: That's OK, the further line(s) with spaces are the extended description and might actually even be required.

ekohl:
Verified it works on Debian Wheezy and Jessie (using debstaging.theforeman.org). Updated the copyright by just putting my name there.

ekohl:
Failures appear to be because it tries to build ruby-mysql2 for jessie and trusty but they don't need it.

mmoll:
merged, bedankt @ekohl!
building now

PR #794, opened 2015-09-01
Release Remote Execution Plugins 0.0.5 (RPM)

stbenjam:
Nightly Only


domcleal:
Thanks @stbenjam, building.

PR #795, opened 2015-09-01
Release Remote Execution Plugins 0.0.5 (DEB)

stbenjam:
Nightly only, please


mmoll:
Looking good, merged, thanks @stbenjam!
Next time I'd prefer one commit per package.

stbenjam:
Sure thing, @domcleal had suggested the opposite for RPM so thought it applied here too

domcleal:
Not sure I meant to advocate it, though I can't remember now what/when I said.  Multiple commits per PR is fine on both - it makes it a lot easier if they need to be selectively cherry picked to other branches.

PR #796, opened 2015-09-01
Remote execution to 1.9 (DEB)

stbenjam:
You need this to make REX available for 1.9, right?


mmoll:
only 0.0.1? :)
At least the dependencies of dynflow are missing.

stbenjam:
I didn't look closely, it was scripted but guess it didn't grab all my commits, will fix

dependencies/trusty/dynflow/changelog
```diff
@@ -0,0 +1,5 @@
+ruby-dynflow (0.8.5-1) stable; urgency=low
```
iNecas: 0.8.7 is the current one

stbenjam:
I think we can just drop this, 1.10 is almost released, and next iteration of REX is 1.10 or higher

PR #797, opened 2015-09-03
add highline to dependencies for trusty

mmoll:
This is an untested 1:1 copy of the wheezy version, which was 1:1 copied from the Debian upstream.


domcleal:
Thanks @mmoll, merged as 76aaf6e and building.

PR #798, opened 2015-09-03
use node.js in assets for trusty and jessie

mmoll:
let's see if this works :)


domcleal:
If you name your branch "deb/develop-nodejs" then the job will download a nightly tarball, it's assuming it's a release or something right now.

PR #799, opened 2015-09-03
use node.js in assets for trusty and jessie

mmoll:
same as #798


domcleal:
[test]

domcleal:
Merged as 7ccc096, thanks @mmoll!

PR #802, opened 2015-09-14
Release foreman_remote_execution 0.0.6 (RPM)

stbenjam:


domcleal:
Thanks @stbenjam, building for both develop and 1.9.

domcleal:
Actually, I've only built it for develop right now.  You asked for 0.0.5 to only be nightly in #794, does the same apply here?

stbenjam:
Yea just nightly for now

PR #803, opened 2015-09-14
Release foreman_remote_execution 0.0.6 (DEB)

stbenjam:

PR #804, opened 2015-09-14
Include therubyracer as the JS engine on wheezy/precise

domcleal:
A continuation of #799, which I think may fix the build error under #803.  I'll try and build another plugin package on top of this to check.


domcleal:
Foreman package built for wheezy from this PR: http://ci.theforeman.org/job/packaging_build_deb_coreproject/label=debian,os=wheezy/2347/console

Built a plugin, which always uses wheezy, on top of that foreman build successfully: http://ci.theforeman.org/job/packaging_build_deb_plugin/711/label=debian/console

mmoll:
merged, thanks @domcleal!
my bad, very sorry :cry:

PR #805, opened 2015-09-14
Update rubygem-bastion to 2.0.1

ehelms:


domcleal:
Thanks @ehelms, building.  I'll also reset Release to 1.

PR #808, opened 2015-09-20
update foreman_xen to 0.1.3 (DEB)

mmoll:
also for 1.9, please


domcleal:
Thanks @mmoll, building for both.

PR #809, opened 2015-09-20
update foreman_xen to 0.1.3 (RPM)

mmoll:
also for 1.9, please


domcleal:
Thanks @mmoll, will build for both.

rubygem-foreman_xen/rubygem-foreman_xen.spec
```diff
@@ -18,7 +18,7 @@
 
 Summary:    Provision and manage XEN Server from Foreman
 Name:       %{?scl_prefix}rubygem-%{gem_name}
-Version:    0.1.2
+Version:    0.1.3
 Release:    2%{?dist}
```
domcleal: I'll reset this to 1.

PR #811, opened 2015-09-21
update hammer_cli(_foreman) to 0.4.0 (DEB)

mmoll:
ideally this could also go to 1.9 for the reduced dependencies to ease life on Debian/testing


domcleal:
Can you add a deb/ prefix please?

I don't think we'll put this in 1.9, as IIRC 0.4.0 would be intended for 1.10.  I'd also like to avoid introducing any bugs into 1.9 with changing deps.

PR #813, opened 2015-09-22
Discovery 4.1.0

lzap:


domcleal:
Thanks @lzap, merged to deb/develop only and building.

PR #819, opened 2015-10-01
Release dynflow 0.8.6 (RPM)

stbenjam:


domcleal:
Thanks @stbenjam, building.

rubygem-dynflow/rubygem-dynflow.spec
```diff
@@ -5,7 +5,7 @@
 
 Summary: DYNamic workFLOW engine
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.8.5
+Version: 0.8.6
 Release: 3%{?dist}
```
domcleal: also reset to 1, I'll do it

stbenjam: Thanks

PR #827, opened 2015-10-01
Release smart_proxy_remote_execution_ssh 0.0.6 (DEB)

stbenjam:


ares:
Seems that something gone wrong here, it could not connect to deb.theforeman.org. I wonder if [test] retriggers the job here as well.

mmoll:
merged as d7ec284

PR #834, opened 2015-10-09
Update dynflow to 0.8.7 (RPM)

iNecas:


domcleal:
This version of dynflow adds a runtime dep on sequel, could you add the requires line too please?  Note that it should be `rubygem-sequel` rather than `rubygem(sequel)` due to BZ#1174138 in Fedora.

iNecas:
Updated, also noting that we would like to get this build into 1.9, as it has important fixes related to foreman_remote_execution

domcleal:
Thanks @iNecas, building.  Would you like the 1.9 version built now, or did you want to wait?  (The other PR you say "eventually", not sure if you're planning a testing period in nightly.)

iNecas:
Tested with 1.9 production: good to go to 1.9! + 1.10 RC1 as well

domcleal:
Updating this gem frequently in older releases is causing me a bit of a headache, just to ensure it's built correctly (under the right SCL etc).  I think I'm going to add %{foremandist} to its release field, which should help.  Are there any others, apart from foreman-tasks and remote execution that you plan on updating?

iNecas:
dynflow, foreman-tasks, smart_proxy_dynflow and smart_proxy_remote_execution_ssh are those that are main subjects of the work, however, we don't expect that much backporting later (the remote execution is just moving on from being compatible with 1.9), this particular release is mainly due to one bug that prevented the users to try out the execution in future.

domcleal:
Okay, thanks.  So adding foremandist to dynflow will let us more easily build it in older versions, and foreman-tasks has it already.  The smart proxy ones ought to be fine as they're non-SCL, so can be safely built in nightly and retagged.

iNecas:
ok

domcleal:
Done in 23f67f7, so we can now build it independently on both rpm/1.10 and rpm/develop without retagging.

iNecas:
Thanks @domcleal !

PR #835, opened 2015-10-09
Update foreman_remote_execution to 0.0.10 (RPM)

iNecas:


domcleal:
Looks like this will fail due to new files being added into the gem.  They'll need to be excluded or added as appropriate.  (Consider an exclude on `%{gem_instdir}/.*` etc.)

iNecas:
Updated (waiting for CI if it helped) but not sure there is a better way to do so. Also, similarly as #834, we will ask for tagging into 1.9 repos eventually

iNecas:
Damn, bad gemspec :) I already have a weekend, again and better (hopefully)

domcleal:
Thanks @iNecas, building for nightlies.  Let me know when you want it in 1.9.  We'll need to update 1.10 as well once RC1 is available - for now we can't build in that tag.

iNecas:
Tested with 1.9 production: good to go to 1.9! + 1.10 RC1 as well

domcleal:
Cherry picked over to rpm/1.9 and building now.  I'll do 1.10 once we've got RC1 available.

PR #837, opened 2015-10-09
Update foreman_remote_execution to 0.0.10 (DEB)

iNecas:


mmoll:
`deb/` prefix for the branch name also missing here for the PR tests

PR #838, opened 2015-10-09
Update dynflow to 0.8.7 (DEB)

iNecas:


mmoll:
ruby-sequel needs to be specified as dependency.

iNecas:
Like this?

mmoll:
Oh, sorry, mixed that up with the system ruby dynflow package... the change to the `control` file must get reverted then, but the addition to `gem.list` is correct.

iNecas:
Can I make the system ruby dynflow package changes in this PR as well?

iNecas:
Ok, updated the old file, as well as the system ruby dynflow

mmoll:
sequel needs to get removed from the -tasks gem.list then:
```
trying to overwrite '/usr/share/foreman/vendor/cache/sequel-4.24.0.gem', which is also in package ruby-foreman-tasks 0.7.6-1
```

mmoll:
BTW, would it make sense to remove sequel as dependency in the gems of foreman-tasks,  smart_proxy_dynflow (and maybe more I'm not aware of)?

iNecas:
We don't have a foreman-tasks dependent on the newest dynflow, so not right now, but once the foreman-tasks increases the dynflow version dependency, it would be the right time to do

mmoll:
ok, so I'm cutting a -2 release of foreman-tasks without sequel for now to get this handled and revisit this PR later. Also I'll open Issues on the other gems about this.

mmoll:
#851 tries to remove sequel from -tasks, but it doesn't work, so I think it's best to close this PR and wait for a foreman-tasks gem release without the sequel dependency.

What IMHO could go in right now would be the updated ruby-dynflow packages (the ones in the dependencies dir).

domcleal:
This can be rebased now to only change dependencies/ (for the smart proxy), as the Foreman plugin has been merged and updated in #852.

mmoll:
rebased and merged as 7145392 for develop and ab4e10c for 1.10

PR #841, opened 2015-10-12
Release foreman_salt 4.0.0 (DEB)

stbenjam:
Nightly, and the 1.10 RC please


mmoll:
Seems to work just fine, also merged as 75f7581 for 1.10 (will be built for 1.10 once we have RC1 packages). Thanks @stbenjam!

mmoll:
also built for 1.10 now

PR #842, opened 2015-10-13
Change dynflow arch to 'all'

domcleal:
There doesn't appear to be any native content in dynflow, so it should be 'all'.  This will prevent us rebuilding it for ARM archs separately to x86.


mmoll:
merged, thanks @domcleal!

PR #843, opened 2015-10-13
Add lsb-release dep for LSB facts (fixes #12134)

eLvErDe:
Some foreman puppet deployment scripts depends on LSB based facts to create proper configurations files. LSB release command is not installed in Debian/Ubuntu system, so add the explicit dependency on foreman-installer package


mmoll:
[test]

mmoll:
also merged as 33f766f for 1.10 and 928ecd7 for 1.9, merci @eLvErDe

PR #844, opened 2015-10-13
ruby-mysql2: disable tests for package building (DEB)

mmoll:
100% copied from Debian upstream, except the new gem2deb version restriction, let's see if it's building with the older gem2deb anyway...


domcleal:
It went red, but at least for a better reason than the current build.  It seems to be failing in the specs, I wonder if it's expecting a different rspec version (e.g. 2 vs 3) to the one provided in precise?

mmoll:
mh... maybe we can just disable the tests like we do for the other packages...

domcleal:
Yes, that would work.  I think we could probably fix the tests on .16, with divergence from Debian - or we remove them on update to .18.

I don't think that it is rspec 2 vs 3 as I guessed, since both .16 and .18 appear to use rspec ~> 2.8.0.  It'd need a bit more investigation to solve.

domcleal:
Oh! dependencies/precise/mysql2/patches/port_to_rspec3.patch is added in this PR.  Try removing it, since precise ships rspec 2?

dependencies/precise/mysql2/patches/series
```diff
@@ -1,5 +1,5 @@
 remove_rpath_compilation_flag.patch
 remove_rubygems_from_examples.patch
 #deactivate_failing_specs.patch
-avoid_openssl_loop.patch
-correct_mysql_init.patch
+cleanup_fork.patch
+port_to_rspec3.patch
```
domcleal: This needs to be removed.

mmoll:
all working on precise, but still no joy on wheezy because of the failing connection to mysql, so I'm going to disable the tests.

mmoll:
that's green now, but as failing tests were the original problem to be addressed here, we don't need the update but rather just disable tests, right?

domcleal:
Indeed, would you prefer to keep it at the current version with tests disabled?

mmoll:
Yes, that's what I did now and it's green. The version was not bumped intentionally to not cause potential problems for dist-upgrades.

domcleal:
Merged as 3078fa4, thanks @mmoll.

mmoll:
also merged as 9d7c308 for 1.10

domcleal:
I triggered a build for it at http://ci.theforeman.org/job/packaging_build_deb_dependency/997/ and for rkerberos at http://ci.theforeman.org/job/packaging_build_deb_dependency/998/.

PR #847, opened 2015-10-15
update foreman_xen to 0.1.5 (RPM)

mmoll:
for 1.9 and 1.8 only, please


domcleal:
Thanks @mmoll, building and tagging down to 1.8.

PR #850, opened 2015-10-15
remove sequel from foreman-tasks gem.list

mmoll:
see #838


mmoll:
this change won't work standalone.

PR #851, opened 2015-10-15
update dynflow to 0.8.7 (DEB)

mmoll:
continuation of #838


mmoll:
doesn't work for -tasks.

PR #852, opened 2015-10-17
roll dynflow into foreman-tasks (DEB)

mmoll:
@iNecas @ares @domcleal does this make any sense? It could help with http://projects.theforeman.org/issues/11714


iNecas:
My Debian packaging knowledge is too limited for me to give a relevant comment here, other than I'm for everything that makes it easier for Debian users to use the Dynflow related stack.

plugins/ruby-foreman-tasks/debian/control
```diff
@@ -2,12 +2,13 @@ Source: ruby-foreman-tasks
...
 Maintainer: Michael Moll <mmoll@mmoll.at>
-Build-Depends: debhelper, foreman (>= 1.8.0~rc1), foreman-sqlite3 (>= 1.8.0~rc1), ruby-foreman-dynflow (>= 0.8.6), ruby-foreman-dynflow (<< 0.9.0)
+Build-Depends: debhelper, foreman (>= 1.8.0~rc1), foreman-sqlite3 (>= 1.8.0~rc1)
 Standards-Version: 3.9.3
 Homepage: https://github.com/Dynflow/dynflow
 
 Package: ruby-foreman-tasks
 Architecture: all
-Depends: ${misc:Depends}, bundler, foreman (>= 1.8.0~rc1), ruby-foreman-dynflow (>= 0.8.6), ruby-foreman-dynflow (<< 0.9.0)
+Conflicts: ruby-foreman-dynflow
```
domcleal: Should it also declare `Replaces: ruby-foreman-dynflow`?

mmoll: indeed, changed.

mmoll:
I did only test this in one install, but I think, this is good.

domcleal:
So I think this OK, and it works well.  I also think it'll work fine as nothing else in the Foreman plugin packages depend on dynflow directly.

What I don't quite understand is why #851 didn't work?

mmoll:
Unfortunately I don't remember the details, but the update from the existing packages didn't work, probably just a case of missing versioned Breaks/Replace/Conflicts lines.

domcleal:
Merged as 9f9c0fe and building, thanks @mmoll.

mmoll:
@domcleal could you (ACK to) merge this into 1.10 when 1.10 RC2 is released to ease updates from 1.9?

domcleal:
Sure.

mmoll:
merged as 0591ebd into 1.10

PR #853, opened 2015-10-18
package smart_proxy_dns_route53 (DEB)

mmoll:


ekohl:
Untested, but looks good.

plugins/smart_proxy_dns_route53/install
```diff
@@ -0,0 +1,2 @@
+debian/dns_route53.rb usr/share/foreman-proxy/bundler.d
+config/dns_route53.yml etc/foreman-proxy/settings.d
```
domcleal: Can we force permissions and ownership to something tighter here?  It's being installed as root:root 0644, but since it usually contains credentials, it should be root:foreman-proxy 0640, like the installer sets up.

mmoll: done via .postinst, it seems that's the only reasonable way with .deb...

domcleal:
Merged as cc2d1ed, 24c39ed to deb/develop, and eca86b9, 7ef2edb to deb/1.10.  Thanks @mmoll, building now.

PR #855, opened 2015-10-19
Updated rubygem-smart_proxy_discovery to 1.0.3 (DEB)

lzap:
1.10 as well please.

### DIFF
--- a/plugins/smart_proxy_discovery/changelog
+++ b/plugins/smart_proxy_discovery/changelog
@@ -1,3 +1,9 @@
+ruby-smart-proxy-discovery (1.0.3) stable; urgency=low
+
+  * 1.0.3 released
+
+ -- Lukas Zapletal <lzap+git@redhat.com>  Mon, 19 Oct 2015 15:42:31 +0200
+
 ruby-smart-proxy-discovery (1.0.2-1) unstable; urgency=low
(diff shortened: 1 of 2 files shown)


plugins/smart_proxy_discovery/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-smart-proxy-discovery (1.0.3) stable; urgency=low
```
mmoll: `1.0.3-1` please

domcleal: I don't know if @lzap used it, but I think the update script I contributed the other week leaves off the `-1`.  I've always been a little unclear about whether to include it or not for the first package of a new upstream version, but on re-reading [this bit of the Debian manual](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version) (actually describing the binary control file), it's convention to include it.

domcleal: #867 fixes this, but @lzap you still need to update this PR.

domcleal:
Fixed and merged in 020dd29, thanks @lzap.

domcleal:
Merged as 7aadef6 to deb/1.10.

PR #858, opened 2015-10-24
ship systemd service file on Debian/jessie

mmoll:
also move the init file, conforming to modern debhelper standards


ekohl:
Wheezy users can install systemd as well. Would it be interesting (and possible) to support that as well?

mmoll:
dh-systemd is only in wheezy-backports, making it impossible the (easy) dh way, so I'm leaning towards only having this for jessie. But thanks for the comment, I thought about adding the service file to foreman-tasks, too, but remember the build instances running wheezy having the same problem (it would still needed to get tested if the resulting plugin package would work on older Debian/Ubuntu versions).

ekohl:
Not limited by knowledge about Debian: are debhelpers build-time or install time? If only build time, you'd think that only our build servers needs wheezy-backports. In any case, I think that shouldn't hold back this PR so :+1:

mmoll:
build-time, but I think if we really think about using backports we might as well think about jessie instances :)

domcleal:
I'd lean towards jessie-only, because trying to put this onto an older OS might lead to complications.  I think we could add wheezy-backports to our pbuilder roots easily, though as I discovered last week when adding precise-updates (breaking building of ruby-mysql2), new packages can break other builds.

When I upgrade to this package, it works and continues running beautifully, but the service unit isn't active so doesn't start on subsequent reboot.  Any idea why that isn't being migrated?

mmoll:
added `--with=systemd` in the rules file, how about now? :)

domcleal:
Yep, that's better, it's active on boot now.

Unfortunately it's failing as /run is tmpfs and there's no mechanism to create the directory.  The old init.d script would mkdir -p it, but for systemd you usually need to install a tmpfiles.d/ config.  https://github.com/theforeman/foreman-packaging/blob/rpm/develop/foreman-proxy/foreman-proxy.tmpfiles is the one from the RPMs, but you won't need the cache dir as it doesn't appear to be tmpfs.

domcleal:
<pre>
● foreman-proxy.service - Foreman Proxy
   Loaded: loaded (/lib/systemd/system/foreman-proxy.service; enabled)
   Active: failed (Result: resources) since Tue 2015-10-27 09:49:34 UTC; 9min ago
  Process: 337 ExecStart=/usr/share/foreman-proxy/bin/smart-proxy (code=exited, status=0/SUCCESS)

Oct 27 09:49:34 foremandeb.localdomain systemd[1]: foreman-proxy.service never wrote its PID file. Failing.
Oct 27 09:49:34 foremandeb.localdomain systemd[1]: Failed to start Foreman Proxy.
Oct 27 09:49:34 foremandeb.localdomain systemd[1]: Unit foreman-proxy.service entered failed state.


E, [2015-10-27T09:49:34.626840 #860] ERROR -- : Error during startup, terminating. Permission denied @ dir_s_mkdir - /var/run/foreman-proxy
D, [2015-10-27T09:49:34.626940 #860] DEBUG -- : Permission denied @ dir_s_mkdir - /var/run/foreman-proxy:/usr/lib/ruby/2.1.0/fileutils.rb:250:in `mkdir'
/usr/lib/ruby/2.1.0/fileutils.rb:250:in `fu_mkdir'
/usr/lib/ruby/2.1.0/fileutils.rb:224:in `block (2 levels) in mkdir_p'
/usr/lib/ruby/2.1.0/fileutils.rb:222:in `reverse_each'
/usr/lib/ruby/2.1.0/fileutils.rb:222:in `block in mkdir_p'
/usr/lib/ruby/2.1.0/fileutils.rb:208:in `each'
/usr/lib/ruby/2.1.0/fileutils.rb:208:in `mkdir_p'
/usr/share/foreman-proxy/lib/launcher.rb:92:in `write_pid'
/usr/share/foreman-proxy/lib/launcher.rb:110:in `launch'
/usr/share/foreman-proxy/bin/smart-proxy:6:in `<main>'
</pre>

mmoll:
done

domcleal:
I don't see the tmpfiles.d config in the binary package?

mmoll:
Found out it's .tmpfile (singular) on Debian :watch:

domcleal:
Nice, thanks @mmoll!  Merged as c79c92e.

PR #859, opened 2015-10-25
Small foreman_salt/smart_proxy_salt update (RPM)

stbenjam:
Nightly + 1.10 please


domcleal:
Thanks @stbenjam, merged as eccc6f4, 0c38980.  I split the commit into one per package, if you can do that in future it'll be much easier as foreman_salt gets built per branch so needs cherry picking, but for proxy plugins we just retag back.

Building now.

PR #860, opened 2015-10-25
Small foreman_salt/smart_proxy_salt update (DEB)

stbenjam:
Nightly + 1.10 please

### DIFF
--- a/plugins/ruby-foreman-salt/debian/changelog
+++ b/plugins/ruby-foreman-salt/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-salt (4.0.1-1) stable; urgency=low
+
+  * 4.0.1 released
+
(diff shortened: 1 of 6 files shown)


mmoll:
see https://github.com/theforeman/foreman-packaging/pull/855#discussion_r42464134 for the -1 suffix I'd like to see (and yes, we should fix the script...).

domcleal:
Also please update the dependency on the Foreman to 1.10 or higher in foreman_salt.

I've fixed the -1 problem in #867, sorry for the hassle.

stbenjam:
Updated

mmoll:
merged as 9d7f4fb & 090f347 for develop and 24fc7c9 & d51500f for 1.10, thanks @stbenjam!

stbenjam:
Thanks!   :snowman:

PR #862, opened 2015-10-25
plugins:puppetdb_foreman - Release 0.1.3 (DEB)

dLobatog:


plugins/ruby-puppetdb-foreman/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-puppetdb_foreman (0.1.3-1) stable; urgency=low
+
+  * Update release to 0.1.3
+
+ -- Daniel Lobato Garcia <elobatocs@gmail.com>  Sun, Oct 25 2015 13:40:36 +0100
```
mmoll: the timestamp has the wrong format, it has to be `25 Oct...`

plugins/ruby-puppetdb-foreman/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-puppetdb_foreman (0.1.3-1) stable; urgency=low
```
mmoll: sorry, didn't catch that in the first version... dash instead of underscore is needed here.

dLobatog: righto - I should fix my script

domcleal: There's one in scripts/ now.

domcleal:
This is for both nightly and 1.10 repos I guess, because of the 1.10 fix?

dLobatog:
@domcleal Yup, it's only relevant to those versions.

domcleal:
Merged as 637a503, a8fe000 and building - thanks @dLobatog.

PR #866, opened 2015-10-26
Updated foreman_discovery to 4.1.2 (DEB)

lzap:
Ditto, 1.10 please, thanks!

### DIFF
--- a/plugins/ruby-foreman-discovery/debian/changelog
+++ b/plugins/ruby-foreman-discovery/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-discovery (4.1.2) stable; urgency=low
+
+  * 4.1.2 released
+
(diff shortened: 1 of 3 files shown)


domcleal:
The failure seems to be a problem on the build servers caused by the bad package name in #862 being cached somewhere, I'm trying to figure out where it's coming from.

domcleal:
No, I was reading the wrong log from Jenkins.  The problem is the branch name - please change "dev/" to "deb/" by opening a new PR.

plugins/ruby-foreman-discovery/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-foreman-discovery (4.1.2) stable; urgency=low
```
domcleal: Note that you'll need to append -1 to the version number as per #867 until it's merged.

mmoll:
merged with the version fixup after testing as 4efefc7 for develop and 26ef673 for 1.10, dík @lzap!

PR #867, opened 2015-10-26
Append -1 to package version if no Debian package version given

domcleal:


mmoll:
merged as dcb31aa, thanks @domcleal!

PR #872, opened 2015-11-01
foreman_openscap 0.5.0

shlomizadok:


rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -23,12 +23,10 @@ License: GPLv3
 URL: https://github.com/OpenSCAP/foreman_openscap
 Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 
-Requires: foreman >= 1.8.0
+Requires: foreman >= 1.11.0
 
 Requires: scap-security-guide
 Requires: %{?scl_prefix}rubygem(deface) < 2.0
-Requires: %{?scl_prefix}rubygem(scaptimony) >= 0.3.2
```
shlomizadok: Do I need to add `Obsoletes:  %{?scl_prefix}rubygem(scaptimony)`? As we do not need it for openscap anymore

domcleal: You could do, but if you wish to remove the package then you should also actually delete the dir, remove it from comps/ and rel-eng/.

rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -15,7 +15,7 @@
 %global foreman_bundlerd_dir %{foreman_dir}/bundler.d
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.4.3
+Version: 0.5.0
 Release: 2%{?foremandist}%{?dist}
```
domcleal: reset to 1

rubygem-foreman_openscap/rubygem-foreman_openscap.spec
domcleal: (Please do this in a separate commit.)

shlomizadok: Will Do. One question: When I `Obsoletes: %{?scl_prefix}rubygem(scaptimony)` I don't need to specify version number(s), right?

domcleal: It's preferable to specify a version number in case it gets reintroduced later.

shlomizadok: Added. Thanks

rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -15,20 +15,19 @@
...
 Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 
-Requires: foreman >= 1.8.0
+Requires: foreman >= 1.11.0
 
 Requires: scap-security-guide
 Requires: %{?scl_prefix}rubygem(deface) < 2.0
-Requires: %{?scl_prefix}rubygem(scaptimony) >= 0.3.2
-Requires: %{?scl_prefix}rubygem(scaptimony) < 0.4.0
+Obsoletes: %{?scl_prefix}rubygem(scaptimony) >= 0.3.2
```
domcleal: It'd be `< 0.3.2-3` in this case, since you're aiming to obsolete all previously shipped versions, not future/newer versions.

(https://fedoraproject.org/wiki/Upgrade_paths_%E2%80%94_renaming_or_splitting_packages)

rubygem-foreman_openscap/rubygem-foreman_openscap.spec
domcleal: Could you also remove the package files (https://github.com/theforeman/foreman-packaging/pull/872/files#r43603837) if it's no longer needed?

shlomizadok:
Removed Scaptimony package files.

domcleal:
Thanks @shlomizadok, building now.

PR #885, opened 2015-11-10
safe_yaml for Fedora 19

xprazak2:
Addition to the tito.props and comps -  foreman_deployments were added to the fedora 19 section, so safe_yaml has to be there as well - sorry I did not notice earlier :cry:

### DIFF
--- a/comps/comps-foreman-plugins-fedora19.xml
+++ b/comps/comps-foreman-plugins-fedora19.xml
@@ -79,6 +79,7 @@
       <packagereq type="default">rubygem-newt</packagereq>
       <packagereq type="default">rubygem-opennebula</packagereq>
       <packagereq type="default">rubygem-openscap</packagereq>
+      <packagereq type="default">rubygem-safe_yaml</packagereq>
       <packagereq type="default">rubygem-satyr</packagereq>
       <packagereq type="default">rubygem-wicked</packagereq>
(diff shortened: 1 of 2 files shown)


domcleal:
Isn't it already in F19?

xprazak2:
@domcleal: I thought it was as well, but it is missing in the [fed19 section](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rel-eng/tito.props#L364-L439).

domcleal:
I mean in Fedora 19 the operating system.  It looks present in Koji?  http://koji.fedoraproject.org/koji/buildinfo?buildID=439085

xprazak2:
@domcleal: Oh, I see. It turns out deployments need at least version 1.0.0, otherwise they (rightly) complain about missing files. Updating it in Fedora would be the best approach I guess?

domcleal:
Oh right, then yes, we'll need to provide an updated version here (can't update it in F19 due to its age).  Please make sure the gemspec and RPM spec specifies that as a minimum version then, I thought it was OK because it said `>= 0`.

xprazak2:
@domcleal: I thought it was OK as well but then I ran into errors. I will make the changes.

domcleal:
Thanks @xprazak2, building now.

PR #886, opened 2015-11-10
Remote execution iteration 3 (RPM)

stbenjam:


stbenjam:
Don't merge yet, found a pretty big bug while testing

domcleal:
The proxy plugin requires should pin net-ssh to `<= 2.10.0` to match the gemspec.  I'd also say that in future the pinning should be done in development deps only (probably in the Gemfile) so packages can be more flexible - inevitably some OS will package 3.x at some point.

stbenjam:
Updated, and will look at how we manage dependencies in the next release, thanks

stbenjam:
This is for nightly and 1.10, btw

domcleal:
Thanks @stbenjam, building for both versions.

PR #887, opened 2015-11-11
Remove bundler, rr, rspec-rails

domcleal:
Small tidyup, mostly because I noticed rspec-rails isn't Rails 4 compatible.  All are unused now.


ehelms:
:+1:

PR #888, opened 2015-11-11
plugins:puppetdb_foreman - Release 0.2.0

dLobatog:


rubygem-puppetdb_foreman/rubygem-puppetdb_foreman.spec
```diff
@@ -17,7 +17,7 @@
 
 Summary:    Foreman plugin to interact with PuppetDB
 Name:       %{?scl_prefix}rubygem-%{gem_name}
-Version:    0.1.3
+Version:   0.2.0
```
domcleal: Nitpick, the indentation's not right here.

domcleal:
Which repos did you want this in?

dLobatog:
@domcleal It should work with anything 1.4+, but I don't think you're building for anything earlier than 1.8, right? 1.8 1.9 1.10 then

domcleal:
Yep, thanks @dLobatog.  Building for those three plus nightly.

PR #892, opened 2015-11-12
Add foreman_pipeline package

xprazak2:


domcleal:
This looks like it has a hard dependency on Katello, but isn't declared in the gemspec.  If so, could you submit it to their repositories instead?  We wouldn't be able to build it in Foreman or resolve its deps.

xprazak2:
@domcleal: OK, I will close this and resubmit to Katello.

domcleal:
Thanks, and the same for the other PRs?

xprazak2:
@domcleal: No problem, I will move them as well.

PR #893, opened 2015-11-12
Remote execution iteration 3 (DEB)

stbenjam:
For 1.10 and nightly


domcleal:
Do you need to set a min version of Foreman of 1.10 on the main plugin?

stbenjam:
Yes, thank you.  Did I get the tilda's right? I don't quite understand why one does the `~rcX` thing.

mmoll:
1.2.3~rc1 is < 1.2.3, while 1.2.3rc1 or 1.2.3.rc1 would be >1.2.3 for the dpkg/apt version comparator.

domcleal:
Thanks @stbenjam, merged as cba1bf5, 9c0209d for develop and f48b5ea, 2283113 for 1.10.  Packages are now building.

PR #894, opened 2015-11-13
bug: Resolve difference between service unit and config

brianredbeard:
In configuring system the foreman-proxy service refuses to start.
Upon investigation this is due to a conflict in the default
settings.yml file and the systemd service unit.  In the
settings.yml file the default behavior is to run in the foreground
while in the systemd service unit the process is specified as
forking, utilizing a pid file. Either the settings file for
systemd based systems needs to be changed to reflect the intended
behavior of the service unit or the service unit needs to be
changed to reflect the desired behavior of settings.yml.

### DIFF
--- a/foreman-proxy/foreman-proxy.service
+++ b/foreman-proxy/foreman-proxy.service
@@ -4,10 +4,9 @@ Wants=basic.target
 After=basic.target network.target
 
 [Service]
-Type=forking
+Type=simple
 User=foreman-proxy
 ExecStart=/usr/share/foreman-proxy/bin/smart-proxy
-PIDFile=/run/foreman-proxy/foreman-proxy.pid
 
 [Install]
 WantedBy=multi-user.target


domcleal:
Thanks for the PR and bug report.  We just addressed in this in the Debian packages actually, but we took the approach of changing settings.yml to have a forking configuration.

The reason for this is that our installer (https://github.com/theforeman/puppet-foreman_proxy) defaults to this mode for all OSes, including non-systemd ones.

Would you mind doing the same here?  Probably applying a patch in the spec against the default settings.yml.example will work.

The equivalent Debian commit is here: 38a2776.  We additionally set it to listen on localhost by default as that's needed to ensure the dpkg postinstall doesn't fail (as the service won't start).  It's up to you if you want to do that here or not.

mmoll:
@brianredbeard what's the status here? :)

ehelms:
This file no longer exists - I think we can close this out?

mmoll:
while this PR stalled, I think the original issue was not fixed.

PR #895, opened 2015-11-16
Update deface to 1.0.2, plus Bundler upgrade fixes [deb]

domcleal:
Only needed in nightly.

### DIFF
--- a/plugins/ruby-foreman-bootdisk/debian/postinst
+++ b/plugins/ruby-foreman-bootdisk/debian/postinst
@@ -25,13 +25,13 @@ BUNDLE=bundle
 export HOME=/usr/share/foreman
 cd /usr/share/foreman
 if [ ! -z "${DEBUG}" ]; then
-  if $BUNDLE show $PLUGIN >/dev/null 2>&1; then
+  if [ -n "$2" ] ; then
     $BUNDLE update $PLUGIN --local
   else
     $BUNDLE install --local --no-prune
   fi
 else
-  if $BUNDLE show $PLUGIN >/dev/null 2>&1; then
+  if [ -n "$2" ] ; then
     $BUNDLE update $PLUGIN --local 2>&1 >> $LOGFILE
   else
     $BUNDLE install --local --no-prune 2>&1 >> $LOGFILE
--- a/plugins/ruby-foreman-chef/debian/postinst
+++ b/plugins/ruby-foreman-chef/debian/postinst
@@ -25,13 +25,13 @@ BUNDLE=bundle
 export HOME=/usr/share/foreman
 cd /usr/share/foreman
 if [ ! -z "${DEBUG}" ]; then
-  if $BUNDLE show $PLUGIN >/dev/null 2>&1; then
+  if [ -n "$2" ] ; then
     $BUNDLE update $PLUGIN --local
   else
     $BUNDLE install --local --no-prune
   fi
 else
-  if $BUNDLE show $PLUGIN >/dev/null 2>&1; then
+  if [ -n "$2" ] ; then
     $BUNDLE update $PLUGIN --local 2>&1 >> $LOGFILE
   else
     $BUNDLE install --local --no-prune 2>&1 >> $LOGFILE
--- a/plugins/ruby-foreman-cockpit/debian/postinst
+++ b/plugins/ruby-foreman-cockpit/debian/postinst
@@ -25,13 +25,13 @@ BUNDLE=bundle
 export HOME=/usr/share/foreman
 cd /usr/share/foreman
 if [ ! -z "${DEBUG}" ]; then
-  if $BUNDLE show $PLUGIN >/dev/null 2>&1; then
+  if [ -n "$2" ] ; then
     $BUNDLE update $PLUGIN --local
   else
     $BUNDLE install --local --no-prune
   fi
 else
-  if $BUNDLE show $PLUGIN >/dev/null 2>&1; then
+  if [ -n "$2" ] ; then
     $BUNDLE update $PLUGIN --local 2>&1 >> $LOGFILE
   else
     $BUNDLE install --local --no-prune 2>&1 >> $LOGFILE
--- a/plugins/ruby-foreman-column-view/debian/postinst
+++ b/plugins/ruby-foreman-column-view/debian/postinst
@@ -25,13 +25,13 @@ BUNDLE=bundle
 export HOME=/usr/share/foreman
 cd /usr/share/foreman
 if [ ! -z "${DEBUG}" ]; then
-  if $BUNDLE show $PLUGIN >/dev/null 2>&1; then
+  if [ -n "$2" ] ; then
     $BUNDLE update $PLUGIN --local
   else
     $BUNDLE install --local --no-prune
   fi
 else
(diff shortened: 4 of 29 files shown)


mmoll:
Looking correct. I did have the following problem when updating:
```
[...]
+ /usr/bin/foreman-ruby /usr/bin/bundle install --local --no-prune
Don't run Bundler as root. Bundler can ask for sudo if it is needed, and
installing your bundle as root will break this application for all non-root
users on this machine.
Resolving dependencies...
Some gems seem to be missing from your vendor/cache directory.
You have requested:
  deface = 1.0.2

The bundle currently has deface locked at 1.0.1.
Try running `bundle update deface`
Bundler::GemNotFound: You have requested:
  deface = 1.0.2

The bundle currently has deface locked at 1.0.1.
Try running `bundle update deface`
[...]
```
It might be specific to my setup, after deleting Gemfile.lock everything went fine.

domcleal:
Do you have the earlier log?  It should be running `bundle show` to determine whether to run bundle install or update.  Running install suggests that it failed.

mmoll:
actually one line above that:
```
+ /usr/bin/foreman-ruby /usr/bin/bundle show deface
```

domcleal:
Yeah, that will fail with something like:

<pre>
# bundle show deface
Could not find deface-1.0.2 in any of the sources
Bundler::GemNotFound: Could not find deface-1.0.2 in any of the sources
/usr/lib/ruby/vendor_ruby/bundler/spec_set.rb:92:in `block in materialize'
/usr/lib/ruby/vendor_ruby/bundler/spec_set.rb:85:in `map!'
/usr/lib/ruby/vendor_ruby/bundler/spec_set.rb:85:in `materialize'
/usr/lib/ruby/vendor_ruby/bundler/definition.rb:132:in `specs'
/usr/lib/ruby/vendor_ruby/bundler/cli/common.rb:15:in `select_spec'
/usr/lib/ruby/vendor_ruby/bundler/cli/show.rb:21:in `run'
/usr/lib/ruby/vendor_ruby/bundler/cli.rb:179:in `show'
/usr/lib/ruby/vendor_ruby/thor/command.rb:27:in `run'
/usr/lib/ruby/vendor_ruby/thor/invocation.rb:126:in `invoke_command'
/usr/lib/ruby/vendor_ruby/thor.rb:359:in `dispatch'
/usr/lib/ruby/vendor_ruby/thor/base.rb:440:in `start'
/usr/lib/ruby/vendor_ruby/bundler/cli.rb:9:in `start'
/usr/bin/bundle:20:in `block in <main>'
/usr/lib/ruby/vendor_ruby/bundler/friendly_errors.rb:5:in `with_friendly_errors'
/usr/bin/bundle:18:in `<main>'
</pre>

as bundler.d/foreman_deface.rb requests 1.0.2.  I guess the intention was for it to run the bundle update, but instead bundle install runs.  What I don't get is why that works for me, but not you:

<pre>
root@foremandeb:/usr/share/foreman# /usr/bin/foreman-ruby /usr/bin/bundle install --local --no-prune                                                                                                                                                                           
Don't run Bundler as root. Bundler can ask for sudo if it is needed, and installing your bundle as root will break this application for all non-root users on this machine.
Resolving dependencies...
...
Installing deface 1.0.2 (was 1.0.1)
0:  deface (1.0.2) from /usr/share/foreman/vendor/ruby/2.1.0/specifications/deface-1.0.2.gemspec
</pre>

The same thing happens if I try an upgrade of ruby-foreman-xen from 0.2.1 to 0.2.2.  It always installs, not updates.  I'm on Jessie with bundler 1.7.4-1.

domcleal:
Oh, just a thought.. I was installing it by itself.  I'll try again later with another plugin installed that actually depends on it, in case Bundler's behaviour depends on whether there's something with a dependency.  That might also explain why dynflow had an issue.

domcleal:
Yeah, that's it.

I think the solution then is to check the second argument passed into our postinst scripts ([doc](https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#s-configdetails)) instead of relying on `bundle show`, which should be the *last* version configured successfully.

On a new installation of the package, we get `+ exec /usr/share/debconf/frontend /var/lib/dpkg/info/ruby-foreman-deface.postinst configure`.

On an upgrade from 1.0.1 we get: `+ exec /usr/share/debconf/frontend /var/lib/dpkg/info/ruby-foreman-deface.postinst configure 1.0.1`.

domcleal:
> My theory was (from tasks -> dynflow) that this could happen with dependent gems, but with foreman_xen this is for sure not the case.

Oh, I didn't mean that I got a bundler/apt-get error - the upgrade worked.  Internally it looks like the `bundle show` fails (because of the bundler.d vs Gemfile.lock mismatch) so the script assumes it's not even installed and runs `bundle install`.

In the case of foreman_xen and other packages that don't have something else depending on them, `bundle install` to update it unexpectedly works.  It's still kind of wrong I think in every plugin's postinst, we ought to fix the `bundle show` conditional and ensure we call `bundle update` in all upgrade cases.

domcleal:
That looks better to me!

<pre>
Unpacking ruby-foreman-deface (9999-plugin+scratchbuild+201511181108) over (1.0.1) ...
Setting up ruby-foreman-deface (9999-plugin+scratchbuild+201511181108) ...
+ . /usr/share/debconf/confmodule
+ [ !  ]
+ PERL_DL_NONLAZY=1
+ export PERL_DL_NONLAZY
+ [  ]
+ exec /usr/share/debconf/frontend /var/lib/dpkg/info/ruby-foreman-deface.postinst configure 1.0.1
+ . /usr/share/debconf/confmodule
+ [ ! 1 ]
+ [ -z  ]
+ exec
+ [  ]
+ exec
+ DEBCONF_REDIR=1
+ export DEBCONF_REDIR
+ LOGFILE=/var/log/foreman-install.log
+ PLUGIN=deface
+ trap db_stop EXIT
+ BUNDLE=bundle
+ [ -h /usr/bin/foreman-ruby ]
+ BUNDLE=/usr/bin/foreman-ruby /usr/bin/bundle
+ export HOME=/usr/share/foreman
+ cd /usr/share/foreman
+ [ ! -z 1 ]
+ [ -n 1.0.1 ]
+ /usr/bin/foreman-ruby /usr/bin/bundle update deface --local
Resolving dependencies...
...
Your bundle is updated!
</pre>

I suggest that I update all postinsts, and we do new builds of both deface and tasks (in case my hunch is right) in deb/develop and also deb/1.10 to ensure upgrades are reliable.

I think the only problem will be when people upgrade only the plugin, as earlier I was seeing that upgrading Foreman at the same time triggered its own `bundle update` (it uses presence of Gemfile.lock) and fixed everything.

domcleal:
Updated all plugins, but only with changelog entries for tasks and deface.

mmoll:
@domcleal once again my hat off to you!
merged and building deface and tasks... Maybe that's also eligible for 1.10 (for tasks)?

domcleal:
Agreed, I'll send a PR to deb/1.10.  I think we're OK with deface there because it'll be upgraded to 1.0.0 at the same time as Foreman, but tasks will probably change.

PR #902, opened 2015-11-20
refs #12146 - include gemspec in foreman-proxy package [deb]

domcleal:


mmoll:
merged, now we're also running into #12546 on .debs, thanks @domcleal!

PR #906, opened 2015-12-02
update foreman_templates to 2.0.1

GregSutcliffe:
Suitable for 1.10 and 1.9 as well.

### DIFF
--- a/plugins/ruby-foreman-templates/debian/changelog
+++ b/plugins/ruby-foreman-templates/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-templates (2.0.1-1) stable; urgency=low
+
+  * 2.0.1 released
+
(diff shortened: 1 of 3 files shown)


plugins/ruby-foreman-templates/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-foreman-templates (2.0.1) stable; urgency=low
```
domcleal: Add a -1 here, or use the scripts to do it automatically.  (The previous entry was a mistake.)

GregSutcliffe: Done.

domcleal:
Thanks @GregSutcliffe, will build for all three.

PR #905, opened 2015-12-01
Remove unused bundler.d files [deb]

domcleal:
Unknown platforms in bundler.d/windows.rb causes a startup failure on
the older Bundler on 12.04, 14.04 and Wheezy.

---

    /usr/share/foreman-proxy/Gemfile.in:6:in `instance_eval': `x64_mingw` is not a valid platform. The available options are: [:ruby, :ruby_18, :ruby_19, :mri, :mri_18, :mri_19, :rbx, :jruby, :mswin, :mingw, :mingw_18, :mingw_19] (Bundler::DslError)

(Sorry, new PR as I got the branch name wrong.)

PR #907, opened 2015-12-03
Fix command to remove bundler.d files

domcleal:
Seems that the expansion I used quietly fails in the rules file, but this does work (by checking the deb contents).

Followup from #905.


mmoll:
hrm, maybe some bash vs. sh thing? anyway, merged, thanks @domcleal!

PR #908, opened 2015-12-04
fixes #12671 - switch Ubuntu 14.04 to ruby2.0 packages [deb]

domcleal:
Seems OK with a small bit of testing, the app starts and Puppet still runs.  I had to change PassengerRuby to either /usr/bin/foreman-ruby or ruby2.0, which I'll send a PR to puppet-foreman for too.


mmoll:
Seems to work fine, let's systest it as 400b5c5. Thanks @domcleal!

PR #911, opened 2015-12-09
smart_proxy_openscap version 0.5.1

shlomizadok:


domcleal:
This (patch) update looks like it's only compatible with 1.11.  Should the minimum version be raised?

shlomizadok:
Update to 1.11, thanks

domcleal:
Thanks, merged to rpm/develop only.  I'd suggest reconsidering the versioning scheme for future updates to prevent users getting the wrong update, or to allow you to update older releases more easily.

shlomizadok:
kk, will do

2015-12-10 11:46 GMT+02:00 Dominic Cleal <notifications@github.com>:

PR #912, opened 2015-12-09
rex_ssh: correct ruby-net-ssh version constraint (DEB)

mmoll:
should also go into 1.10


domcleal:
Thanks @mmoll, also merged as ca71677 to 1.10.

domcleal:
Looks like it also has an epoch that should be in the version number?  http://projects.theforeman.org/issues/12761#change-52956

mmoll:
Indeed, because in upstream Debian/Ubuntu it's set, just a moment...

PR #913, opened 2015-12-10
fixes #12761 - bump epoch of ruby-net-ssh version constraint

mmoll:
also for 1.10, please


domcleal:
Merged, thanks @mmoll.  Also tested the resolution this time in aptitude on trusty.

6681f88 in deb/1.10.

PR #914, opened 2015-12-11
Release hammer_cli_foreman_remote_execution 0.0.2 (RPM)

stbenjam:
Will be for nightly and 1.10, when it's ready.


rubygem-hammer_cli_foreman_remote_execution/rubygem-hammer_cli_foreman_salt.spec
```diff
@@ -0,0 +1,75 @@
...
+
+%if 0%{?fedora} > 18
+Requires: %{?scl_prefix_ruby}ruby(release)
+%else
+Requires: %{?scl_prefix_ruby}ruby(abi)
+%endif
+
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
+Requires: %{?scl_prefix}rubygem(hammer_cli_foreman) >= 0.1.3
+Requires: %{?scl_prefix}rubygem(hammer_cli_foreman) < 0.5.0
```
domcleal: Please avoid pinning this tightly, it causes a test failure and more work each time a new major version of Hammer is shipped.  You should be OK to pin to `< 1.0.0`.  (See also cli-tasks which did this and had to continually update it.)

domcleal:
Tests won't run until rel-eng/ and comps/ are updated for the new package.

comps/comps-foreman-fedora19.xml
```diff
@@ -127,6 +127,7 @@
       <packagereq type="default">rubygem-fog-google-doc</packagereq>
       <packagereq type="default">rubygem-fog-json-doc</packagereq>
       <packagereq type="default">rubygem-fog-libvirt-doc</packagereq>
+      <packagereq type="default">rubygem-fog-xenserver-doc</packagereq>
```
stbenjam: These were added by comps_doc

rubygem-hammer_cli_foreman_remote_execution/rubygem-hammer_cli_foreman_salt.spec
stbenjam: Gem is pinned, and I based it on how the other hammer plugins have it done, but I can fix it when we're ready for 0.0.2, unless you think it's important enough to release just for that. https://github.com/theforeman/hammer_cli_foreman_remote_execution/pull/4

domcleal: It's important enough, since Hammer 0.5.1 is already out, so this won't be installable.

rubygem-hammer_cli_foreman_remote_execution/rubygem-hammer_cli_foreman_salt.spec
```diff
@@ -0,0 +1,81 @@
...
+Summary: Foreman Remote Execution commands for Hammer CLI
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.0.2
+Release: 1%{?dist}
+Group: Applications/System
+License: GPLv3
+URL: http://github.com/theforeman/hammer_cli_foreman_remote_execution
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+%if 0%{?fedora} > 18
```
domcleal: This can be removed now, it's only `ruby(release)`.

comps/comps-foreman-fedora19.xml
domcleal: Thanks, I re-ran it on rpm/develop too to fix that.

domcleal:
[test] as the plugin tags should now work a bit.

domcleal:
Thanks, I'll merge and build this once the first set of plugins are rebuilt in nightly so I don't mess it up.

domcleal:
Cherry picked and building now in rpm/1.10.

PR #922, opened 2015-12-18
update foreman_one to 0.4 (RPM)

mmoll:
All the way down to 1.8, please.


domcleal:
Thanks @mmoll.  Will cherry pick to 1.10, 1.9 and retag to 1.8.

PR #925, opened 2015-12-18
plugins:foreman_docker - Release 2.0.0 [deb]

dLobatog:


domcleal:
The commit message has the wrong version number in it, could you amend please?

domcleal:
It should also increase the min version in the control file I think.

dLobatog:
Updated the control file so that the min version is 1.11, thanks @domcleal

dLobatog:
Well, Jenkins is not going to be able to build the packages with 1.11 (unreleased), I guess this will have to wait until there's a initial version of 1.11 http://ci.theforeman.org/job/packaging_build_deb_plugin/860/arch=x86,label=debian/console

plugins/ruby-foreman-docker/debian/control
```diff
@@ -2,12 +2,12 @@ Source: ruby-foreman-docker
 Section: ruby
 Priority: extra
 Maintainer: Pujan Shah <pujan14@gmail.com>
-Build-Depends: debhelper, foreman-compute (>= 1.8.0~rc1), foreman-assets (>= 1.7.0), foreman-sqlite3 (>= 1.8.0~rc1)
+Build-Depends: debhelper, foreman-compute (>= 1.11.0), foreman-assets (>= 1.11.0)
```
domcleal: a) don't remove foreman-sqlite3, which is causing the build error with a missing sqlite3 DB adapter
b) This needs to be `>= 1.11.0~rc1` as it originally was, because `~rc1` is *lower* than 1.11.0

plugins/ruby-foreman-docker/debian/control
```diff
@@ -2,12 +2,12 @@ Source: ruby-foreman-docker
...
 Standards-Version: 3.9.3
 Homepage: https://github.com/theforeman/foreman-docker
 
 Package: ruby-foreman-docker
 Architecture: all
-Depends: ${misc:Depends}, bundler, foreman-compute (>= 1.7.2)
+Depends: ${misc:Depends}, bundler, foreman-compute (>= 1.11.0)
```
domcleal: also `~rc1`

domcleal:
> Well, Jenkins is not going to be able to build the packages with 1.11 (unreleased), I guess this will have to wait until there's a initial version of 1.11

Nightly Debian builds are actually built with version 9999, so it'll resolve fine - the error's due to a missing sqlite3 adapter.

dLobatog:
Updated - not sure why the sqlite3 adapter is required for this package though

domcleal:
Because the plugin contains apipie documentation, to precompile it during the package build with rake we need a working Rails environment (and therefore a database), so the `rules` file runs db:migrate.  It's only a build requirement, not a runtime requirement.

domcleal:
[test] to rebuild on top of the Rails 4 based packages.

domcleal:
Merged as 1f509a2, thanks @dLobatog.

PR #926, opened 2015-12-19
fixes #12876 - package bundler 1.3 for precise and wheezy

mmoll:
The package sources were taken from Debian and Ubuntu upstreams with slight modifcations to not use packages that aren't available without using backports and allow tests to fail because of too old rspec in the non-backports repo.


debian/jessie/foreman/control
```diff
@@ -3,7 +3,7 @@ Maintainer: Greg Sutcliffe <greg.sutcliffe@gmail.com>
 Section: web
 Priority: extra
 Standards-Version: 3.9.1
-Build-Depends: debhelper (>= 7), cdbs, bundler (>= 1.1), ruby2.1-dev, git | git-core,
+Build-Depends: debhelper (>= 7), cdbs, bundler (>= 1.3), ruby2.1-dev, git | git-core,
```
GregSutcliffe: There's already bundler 1.7 in Jessie, is there a reason not to use it? Apart from consistency across our packages?

mmoll: This is to explicitly state that bundler >= 1.3 is _needed_.

For jessie and trusty, no bundler packages were added in this PR and the ones coming from the stock OK repos were/are already used with success.

mmoll:
[test] again, as there seems to be an older build of the bundler package is still used on wheezy (I deleted bundler_1.3.5-2_all.deb in the freigt and stagingdeb dirs)

debian/jessie/foreman/control
GregSutcliffe: Oh, my mistake, I totally misread `>=` as `==`. Carry on :)

mmoll:
[test]

mmoll:
seems all good now :)

domcleal:
Merged and building, thanks @mmoll!

PR #933, opened 2015-12-21
Package parse-cron gem (RPM)

stbenjam:


rubygem-parse-cron/rubygem-parse-cron.spec
```diff
@@ -0,0 +1,76 @@
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
+%global gem_name parse-cron
+
+Summary: Cron expression parser
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.1.4
+Release: 1%{?dist}
+Epoch: 1
```
domcleal: No epoch required in the first package?

rubygem-parse-cron/rubygem-parse-cron.spec
```diff
@@ -0,0 +1,76 @@
...
+Group: Development/Languages
+
+License: MIT
+URL: https://github.com/siebertm/parse-cron
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
+Requires: %{?scl_prefix_ruby}ruby
+
+%if "%{?scl_ruby}" == "ruby193" || (0%{?el6} && 0%{!?scl:1})
```
domcleal: If this is SCL-only, get rid of this and just require `ruby(release)`

rubygem-parse-cron/rubygem-parse-cron.spec
```diff
@@ -0,0 +1,76 @@
...
+
+Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
+
+%description
+Parses cron expressions and calculates the next occurence after a given date
+
+%package doc
+Summary: Documentation for %{pkg_name}
+Group: Documentation
+Requires: %{?scl_prefix}%{pkg_name} = %{epoch}:%{version}-%{release}
```
domcleal: ditto, no epoch

rubygem-parse-cron/rubygem-parse-cron.spec
```diff
@@ -0,0 +1,76 @@
...
+%exclude %{gem_instdir}/%{gem_name}.gemspec
+%exclude %{gem_instdir}/Gemfile
+%exclude %{gem_instdir}/spec
+%{gem_instdir}/Rakefile
+%{gem_libdir}
+%exclude %{gem_cache}
+%{gem_spec}
+
+%files doc
+%doc %{gem_instdir}/License
```
domcleal: should be in the main %files list instead, so it's always installed

rubygem-parse-cron/rubygem-parse-cron.spec
```diff
@@ -0,0 +1,76 @@
...
+%doc %{gem_instdir}/README.md
+
+%changelog
+* Mon Dec 21 2015 Stephen Benjamin <stephen@redhat.com>
```
domcleal: missing version number

rubygem-parse-cron/rubygem-parse-cron.spec
stbenjam: I could see it maybe being used on the smart proxy at some point in the future, still want it removed for now?

domcleal: Yes - the SCL version is always now ruby(release), and since remote execution's proxy support is EL7+ it wouldn't be needed for that either.

domcleal:
[test] as buildroot is now fixed.

domcleal:
Thanks @stbenjam, building shortly.

PR #934, opened 2015-12-22
remote execution iteration 4 (DEB)

mmoll:


mmoll:
also for 1.10, please

domcleal:
The PR test job is only processing dependency builds and then failing because they failed on precise (expected).  I think I've reconfigured it to continue for all types of package builds despite the failure and then report the whole job as failed instead: [test]

Else you'll have to split it up a bit.

mmoll:
I thin the last package (foreman_remote_execution itself) will still fail because the foreman-tasks requirement can't be satisfied with the scratch version. In that case it would be good to merge all the commits except 6477d28, then build everything and re-run the test again here with the only commit left.

stbenjam:
Thank you @mmoll !

domcleal:
I've merged the first five commits (minus foreman_remote_execution) to deb/develop, will build in a short while once the 1.10.0 build is done, then try to rebuild the last commit.

domcleal:
First five packages are built into nightly/develop, but the last, foreman_remote_execution fails as it has a hard dependency on Rails 3.2.8.

http://ci.theforeman.org/job/packaging_build_deb_plugin/870/arch=x86,label=debian/console

This could be pushed to the deb/1.10 repo, but as it is, it's not compatible at all with develop.

mmoll:
ah, right... rex got locked onto rails 3. I think it's best to merge it into development to ease future merges, but not build it and then cherry-pick everything down to 1.10 and build there.

mmoll:
hrm, and maybe add foreman <<1.11 to the dependencies?

domcleal:
Okay, I'll cherry pick the five commits over to deb/1.10 and build those now.  Do you want to open a new PR against deb/1.10 for the rex plugin, so we can check it works, then I'll merge that to both and build only on 1.10?

domcleal:
> hrm, and maybe add foreman <<1.11 to the dependencies?

Quite possibly.  I didn't allow this in the RPM version because we have repoclosure checks which would have failed in nightly, but there aren't any for the Debian archives at the moment.

mmoll:
sure, #937

PR #935, opened 2015-12-22
Precompile assets in the production Rails environment [deb]

domcleal:
Ensures the correct filenames are generated with digests etc.

---

Currently you'll find JS and other assets being loaded from /javascripts/dashboard.js etc, but it should be with the digest in the filename.  Check that the dashboard page renders properly, and not like this:

screenshot from 2015-12-22 15-45-37


mmoll:
[test] again

mmoll:
wheezy and precise build failed because of the yet missing bundler 1.3 packages, but it succeeded on the other distors and makes perfectly sense, merging...

mmoll:
thanks @domcleal!

PR #937, opened 2015-12-23
update foreman_remote_execution to 0.1.2

mmoll:


mmoll:
that job seems to use the nightly repos, no matter which target branch...

domcleal:
Yeah, just seeing if I can fix that.

domcleal:
http://ci.theforeman.org/job/packaging_build_deb_plugin/873/ is running against 1.10.

I've added a version parameter to the deb plugin job which works in the same way as coreproject and dependencies have, so future PRs should work properly by adding apt sources with that version number.  Note that branch names will also need prefixing with `deb/1.10` in the same way to detect the version when doing PRs.

domcleal:
Merged to deb/1.10 and building at http://ci.theforeman.org/job/packaging_build_deb_plugin/874/.

Cherry picked to deb/develop as 69ade1c and not built.

Thanks @mmoll.

PR #938, opened 2015-12-23
Import kafo, kafo_parsers from current builds [rpm]

domcleal:
Importing what I think are the last two RPMs built outside of this repo so it's easier to update and build them for new OSes (F21 shortly).

OK with you @ares?  I'll send PRs to the projects shortly to remove packaging files there.


ares:
Yeah, kafo is not getting updates frequently any more so I'm fine with having less control over releasing :-) Thanks for taking care of it!

domcleal:
Thanks!

PR #939, opened 2015-12-25
fixes #11152 - add docbook-xsl to installer build-depends (DEB)

mmoll:
xsltproc was not able to resolve the XSL for docbook locally and tried
retrieving it from the internet, leading to build errors in case of
flakey connections.

FreeBSD's packaging CI does offline builds by purpose, exposing the bug reliably. Maybe something similar should be applied to RPMs as well.


domcleal:
Merged, thanks @mmoll!

The RPM builds happen in koji and mock which will be disconnected by default, so that'd be why we wouldn't have seen this.  Also asciidoc there depends on this package.

PR #940, opened 2015-12-29
update foreman_templates to 2.0.3 (DEB)

GregSutcliffe:
Deb changes for templates 2.0.2. Should be suitable for 1.10 / 1.9 but currently untested by me.


mmoll:
that should wait for 2.0.3, because of theforeman/foreman_templates#21

GregSutcliffe:
Apologies, I pushed this to theforeman instead of to my fork when updating. I've not started any builds, just in case there's still issues.

mmoll:
6fac47b for reference

mmoll:
built and tested, seems fine to me, thanks @GregSutcliffe!

PR #941, opened 2015-12-30
use a vendored facter gem for Foreman core (DEB)

mmoll:
Helps in cases where no usable facter is available via packages, i.e. when using AIO packages. See theforeman/puppet-foreman#385


mmoll:
seems like the new logic applied for versioned branches needs more tweaking :)

domcleal:
Try using a branch name with a `deb/develop` prefix.

debian/wheezy/foreman/foreman.install
```diff
@@ -23,6 +23,7 @@ extras/ignored_environments.yml.sample   usr/share/foreman/extras
 extras/noVNC                             usr/share/foreman/extras
 extras/rdoc                              usr/share/foreman/extras
 bundler.d/jsonp.rb                       usr/share/foreman/bundler.d
+bundler.d/facter.rb                      usr/share/foreman/bundler.d
```
domcleal: Nitpick, facter comes before jsonp.

PR #942, opened 2015-12-31
use a vendored facter gem for Foreman core (DEB)

mmoll:
Helps in cases where no usable facter is available via packages, i.e. when using AIO packages. See theforeman/puppet-foreman#385


domcleal:
Merged as 5f32856, thanks @mmoll!

PR #945, opened 2016-01-02
plugins:foreman_ansible - Release 0.2.1 [rpm]

dLobatog:
TODOs before merging this:

* [ ] Move foreman_ansible to theforeman organization

Possibly I want to make it depend on ansible so that it installs Ansible on the Foreman host, thoughts? I'd use it to register hosts in Foreman within it's 'network reach'


rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -0,0 +1,82 @@
...
+            --force %{SOURCE0} --no-rdoc --no-ri
+%{?scl:"}
+
+%build
+
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
+
+%{foreman_bundlerd_file}
```
domcleal: Without a build requires on foreman-plugin, our helper macros won't be available - you'd have to write the file yourself or add the BR.  (Perhaps see foreman_templates for an example of the former.)

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -0,0 +1,82 @@
...
+%foreman_precompile_plugin -a -s
```
domcleal: Ditto, but also no need to precompile anything as far as I can tell - this is only for API docs (`-a`) or assets (`-s`), it has neither.

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -0,0 +1,82 @@
...
+
+%files
+%dir %{gem_instdir}
+%license %{gem_instdir}/LICENSE
```
domcleal: %doc for EL6 compatibility

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -0,0 +1,82 @@
...
+%{gem_instdir}/app
+%{gem_libdir}
+%{gem_instdir}/locale
+%exclude %{gem_cache}
+%{gem_spec}
+%{foreman_bundlerd_dir}/%{gem_name}.rb
+%exclude %{gem_instdir}/test
```
domcleal: This line should be removed as it's in the doc subpackage, or the tests should be removed from the doc package listing.

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -0,0 +1,82 @@
...
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/README.md
+%{gem_instdir}/Rakefile
+%{gem_instdir}/test
+
+%changelog
+* Sat Jan 02 2016 Daniel Lobato - 0.2.1-1
```
domcleal: missing e-mail address

domcleal:
This is missing comps/ and rel-eng/ updates to build and test the package.  Please rebase on rpm/develop before updating them to avoid a conflict.

> Possibly I want to make it depend on ansible so that it installs Ansible on the Foreman host, thoughts? I'd use it to register hosts in Foreman within it's 'network reach'

If the package is available and the plugin has a hard dependency, then yep.  I'd suggest avoiding it if it's a soft dep, some people don't like strictly unnecessary tools being installed.

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -0,0 +1,80 @@
...
+%{gem_instdir}/app
+%{gem_libdir}
+%{gem_instdir}/locale
+%exclude %{gem_cache}
+%{gem_spec}
+%{foreman_bundlerd_dir}/%{gem_name}.rb
+%exclude %{gem_instdir}/test
+
+%files doc
+%doc %{gem_docdir}
```
domcleal: docdir will only exist if rdoc and/or ri docs are enabled, but they're disabled above on the gem install line.  This causes the test failure.

You could leave one/both enabled, remove this line, and/or replace the `gem install` line with a more modern `%gem_install` macro (see some other spec files for examples) which enables them by default.

dLobatog:
[test]

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
dLobatog: Interesting, does this macro work on el6? Tito tasks passed on f21 and el7 but el6 failed with `/var/tmp/rpm-tmp.EdOGhv: line 39: gem: command not found`

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -0,0 +1,76 @@
...
+Group: Documentation
+Requires: %{?scl_prefix}%{pkg_name} = %{version}-%{release}
+BuildArch: noarch
+
+%description doc
+Documentation for %{name}.
+
+%prep
+%setup -q -c -T
+%gem_install -n %{SOURCE0}
```
domcleal: missing an SCL wrapper here, so it's executing in a non-SCL environment I think.  Not sure why it happens to be installed and work on EL7, it might be that some of our packages are causing non-SCL Ruby to be installed too (http://projects.theforeman.org/issues/12915).

Change to:

<pre>
%{?scl:scl enable %{scl} - << \EOF}
%gem_install -n %{SOURCE0} 
%{?scl:EOF}
</pre>

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -0,0 +1,76 @@
...
+Release: 1%{?foremandist}%{?dist}
+Group:   Applications/System
+License: GPLv3
+URL:     https://github.com/theforeman/foreman_ansible
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+Requires: foreman >= 1.6.0
+
+Requires: %{?scl_prefix_ruby}ruby(release)
+Requires: %{?scl_prefix_ruby}rubygems-devel
```
domcleal: Not a runtime dependency, only build time.

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -0,0 +1,76 @@
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
+%global gem_name foreman_ansible
+
+%global foreman_dir /usr/share/foreman
+%global foreman_bundlerd_dir %{foreman_dir}/bundler.d
```
domcleal: These two foreman* lines should be removed as you're using foreman-plugin which provides them.

dLobatog:
Removed foreman_dir and foreman_bundlerd_dir since there's a requirement on foreman-plugin

domcleal:
Could you rebase the PR please?  Seems my comps/ change today conflicted with this.

dLobatog:
Rebased, there was a conflict w/ bootdisk & chef on comps yeah

comps/comps-foreman-plugins-rhel6.xml
```diff
@@ -11,6 +11,12 @@
       <packagereq type="default">tfm-rubygem-bastion</packagereq>
       <packagereq type="default">tfm-rubygem-bastion-devel</packagereq>
       <packagereq type="default">tfm-rubygem-foreman_abrt</packagereq>
+<<<<<<< 7624354cfbbeeed163d68611279c932eaba6e4db
```
domcleal: Oops, I didn't recheck the rebase.  I'll remove this.

domcleal:
Thanks @dLobatog, merged and building now to nightly.

PR #929, opened 2015-12-21
Iteration 4 Main PR - Update foreman_remote_execution to 0.1.2 (RPM)

stbenjam:
This is the main PR.  Have to do the package dance :dancers: again, because there's new dependencies and updates.

Core
- [x] concurrent-ruby https://github.com/theforeman/foreman-packaging/pull/928
- [x] concurrent-ruby-edge https://github.com/theforeman/foreman-packaging/pull/931
- [x] dynflow https://github.com/theforeman/foreman-packaging/pull/930
- [x] parse-cron https://github.com/theforeman/foreman-packaging/pull/933
- [x] foreman-tasks https://github.com/theforeman/foreman-packaging/pull/932
- [ ] foreman_remote_execution (~~this pr~~ https://github.com/theforeman/foreman-packaging/pull/948)

Smart Proxy
- [x] smart_proxy_remote_execution_ssh https://github.com/theforeman/foreman-packaging/pull/936

Hammer
- [x] hammer plugin https://github.com/theforeman/foreman-packaging/pull/914

PR #948, opened 2016-01-04
Update foreman_remote_execution to 0.1.2

stbenjam:


domcleal:
Looks like this needs dynflow >= 0.8.8 and presumably some of the other package updates too?

stbenjam:
I thought you'd cherry-pick the commits mentioned from #929? They'd all have to be built and in 1.10 first, before this one will be green?

domcleal:
Ok, I'll try and get to that soon, it'll be tricky as the various new and updated package dependencies will need %{?foremandist} adding to prevent duplicate NEVRAs in Koji, which isn't permitted.  (They'll need to be rebuilt since develop is incompatible with 1.10.)

stbenjam:
Not sure I really understand, why can't our packages from develop be used in 1.10?

domcleal:
develop's built against a different software collection (rh-ruby22 etc), so the packages have to be rebuilt and not just retagged.

stbenjam:
I guess my question is, what could I have done or do to make this work? Rex will pretty much always go to stable.

domcleal:
Most importantly I'd suggest trying to separate the foreman-tasks and rex updates, it'll make it all more manageable as we won't have a big list of packages to build in order.  Hopefully in future this will happen naturally too.

The foreman-tasks/dynflow updates will be needed in develop even if rex itself isn't, since other projects depend on it.  A PR against rpm/develop and then either asking me or another maintainer to cherry pick it down to a stable branch works, or wait for the rpm/develop merge and then propose another PR against the stable branch with the cherry picks to save us some work.  I'd usually do this sooner, but have been a little busy with all of the other updates.

As for the rex update itself, I think updating develop and then cherry picking is the most feasible model and matches every other package.  The only alternative that springs to mind is to remove it entirely from the rpm/develop branch (and nightlies) and add it like a new package to each stable branch as and when they appear and the plugin becomes compatible with it.

domcleal:
I've cherry picked parse-cron, concurrent-ruby, -edge and dynflow and they're currently building.  I'll do foreman-tasks tomorrow which should then make this PR possible.

stbenjam:
Thanks!!

domcleal:
[test]

domcleal:

PR #944, opened 2015-12-31
Fixes #12970 - Package patternfly-sass

dLobatog:
* [x] Add patternfly-sass
* [x] Update bootstrap-sass to 3.3.6 - https://github.com/theforeman/foreman-packaging/pull/951
* [x] Add autoprefixer-rails 5.2.1.3 - https://github.com/theforeman/foreman-packaging/pull/950
* [x] Update sprockets-rails to 2.3.3 - https://github.com/theforeman/foreman-packaging/pull/949
* [x] Add sprockets 3 (at least). - https://github.com/theforeman/foreman-packaging/pull/954
* [x] Add ejs 1.1 for sprockets 3 - https://github.com/theforeman/foreman-packaging/pull/965
* [x] Add sass-rails 5.0, ror41 max version is 4.0.3 - https://github.com/theforeman/foreman-packaging/pull/961
* [x] Add font-awesome-sass - #943
* [x] Add sass 3.4.19 - https://github.com/theforeman/foreman-packaging/pull/966
* [x] Update foreman.spec to depend on the aforementioned packages
* [x] Build a temporary repository with this 
[patternfly.zip](https://github.com/theforeman/foreman-packaging/files/91820/patternfly.zip)

cc - theforeman/foreman#294

PR #949, opened 2016-01-05
Refs #12970 - sprockets-rails 2.3.3 for Patternfly

dLobatog:
https://github.com/theforeman/foreman-packaging/pull/944#issue-124499878


rubygem-sprockets-rails/rubygem-sprockets-rails.spec
```diff
@@ -5,17 +5,17 @@
...
-Version: 2.2.4
-Release: 2%{?dist}
+Version: 2.3.3
+Release: 1%{?dist}
 Summary: Sprockets Rails integration
 Group: Development/Languages
 License: MIT
 URL: https://github.com/rails/sprockets-rails
-Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
```
ekohl: Why doesn't https work?

dLobatog: No idea, but when Jenkins runs wget it fails for rubygems https - http://ci.theforeman.org/job/packaging_build_rpm/3032/console

domcleal: The Jenins slave that tito's usually run on is EL6 and it doesn't like those certificates, among other problems with HTTPS URLs.  It needs rebuilding.

domcleal:
1) The -tests tarball needs to committed directly as a binary.  We can't use git-annex with a URL remote as there's nowhere to download it from unfortunately (and we're not using any storage with git-annex).

2) Change scl_prefix in the Requires/BuildRequires of the spec file to scl_prefix_ror.  I copied this spec file from the rh-ror41 collection to build only for Fedora so didn't fix it at the time.  Since we're building it in `tfm`, it needs to explicitly require the `ror` variants.

3) The commit message should say sprockets-rails, not sprockets.

4) foreman/foreman.spec will need to depend on `%{?scl_prefix}rubygem(sprockets-rails)` rather than the ror variant.

Hopefully tests will get further with those first two changes.

dLobatog:
@domcleal Updated according to your comments, thanks!

dLobatog:
Looks like tito attempts to `get` the tests tgz from `git-annex` and doesn't run them from the binary file. http://ci.theforeman.org/job/packaging_build_rpm/3050/console

rubygem-sprockets-rails/sprockets-rails-2.3.3-tests.tgz
```diff
@@ -0,0 +1 @@
+../.git/annex/objects/fJ/qZ/SHA256E-s92435--64526351be8f06a44da4c019ae42a89e55a627373d9ffa5caa064bddf4fb6089.tgz/SHA256E-s92435--64526351be8f06a44da4c019ae42a89e55a627373d9ffa5caa064bddf4fb6089.tgz
```
domcleal: This is still an annexed file.

dLobatog:
Duh, nevermind, I *did* git-annex the tests again.

dLobatog:
Now tests are a binary again. I did `git reset HEAD~1` before and when I `git add`ed back the tests.tgz, it was added as an annex again.

foreman/foreman.spec
```diff
@@ -190,6 +190,8 @@ BuildRequires: %{?scl_prefix_ror}rubygem(execjs) >= 1.4.0
 BuildRequires: %{?scl_prefix_ror}rubygem(execjs) < 2.5.0
 BuildRequires: %{?scl_prefix_ror}rubygem(jquery-rails) >= 3.1
 BuildRequires: %{?scl_prefix_ror}rubygem(jquery-rails) < 4.0
+BuildRequires: %{?scl_prefix_ror}rubygem(sprockets-rails) >= 2.3.3
```
domcleal: a) scl_prefix as you're using the one in our collection
b) the older dependency on sprockets-rails needs removing, this replaces it

foreman/foreman.spec
```diff
@@ -368,6 +370,8 @@ Requires: %{?scl_prefix_ror}rubygem(uglifier) >= 1.0.3
 Requires: %{?scl_prefix_ror}rubygem(execjs) >= 1.4.0
 Requires: %{?scl_prefix_ror}rubygem(jquery-rails) >= 3.1
 Requires: %{?scl_prefix_ror}rubygem(jquery-rails) < 4.0
+Requires: %{?scl_prefix_ror}rubygem(sprockets-rails) >= 2.3.3
```
domcleal: ditto above.

Also http://projects.theforeman.org/issues/13074 should be fixed as having this dependency only in foreman-assets could lead to the wrong version being installed via the main foreman package.

rubygem-sprockets-rails/rubygem-sprockets-rails.spec
```diff
@@ -5,31 +5,31 @@
...
+# git checkout v2.3.3
+# tar czvf sprockets-rails-2.3.3-tests.tgz test/
 Source2: sprockets-rails-%{version}-tests.tgz
 Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
-Requires: %{?scl_prefix}rubygem(sprockets) => 2.8
-Requires: %{?scl_prefix}rubygem(sprockets) < 4
-Requires: %{?scl_prefix}rubygem(actionpack) >= 3.0
-Requires: %{?scl_prefix}rubygem(activesupport) >= 3.0
+Requires: %{?scl_prefix_ror}rubygem(sprockets) => 2.8
```
domcleal: If you're packaging sprockets into our SCL too, then you need to change this to scl_prefix - ditto in BRs.

dLobatog:
https://github.com/theforeman/foreman/pull/3047/files

dLobatog:
@domcleal Rebased too

foreman/foreman.spec
```diff
@@ -203,6 +203,8 @@ BuildRequires: %{?scl_prefix}rubygem(autoprefixer-rails) >= 5.2
 BuildRequires: %{?scl_prefix}rubygem(autoprefixer-rails) < 6.0
 BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) >= 2.7
 BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) < 3.0
+BuildRequires: %{?scl_prefix}rubygem(sprockets-rails) >= 2.3.3
```
domcleal: This should be in the Gemfile section (it's already there, IIRC).

foreman/foreman.spec
```diff
@@ -384,6 +386,8 @@ Requires: %{?scl_prefix}rubygem(autoprefixer-rails) >= 5.2
 Requires: %{?scl_prefix}rubygem(autoprefixer-rails) < 6.0
 Requires: %{?scl_prefix}rubygem(patternfly-sass) >= 2.7
 Requires: %{?scl_prefix}rubygem(patternfly-sass) < 3.0
+Requires: %{?scl_prefix}rubygem(sprockets-rails) >= 2.3.3
```
domcleal: This should be in the main 'foreman' package's requires.

dLobatog:
@domcleal Updated - now it's only on the regular Gemfile section & not on the assets package. It was like that because when I originally made this PR we hadn't moved sprockets & sprockets-rails to Gemfile yet.

foreman/foreman.spec
```diff
@@ -186,8 +186,8 @@ BuildRequires: %{?scl_prefix}rubygem(protected_attributes) >= 1.1.1
 BuildRequires: %{?scl_prefix}rubygem(protected_attributes) < 1.2.0
 BuildRequires: %{?scl_prefix}rubygem(sprockets) >= 3
 BuildRequires: %{?scl_prefix}rubygem(sprockets) < 4
-BuildRequires: %{?scl_prefix_ror}rubygem(sprockets-rails) >= 2.2.2
-BuildRequires: %{?scl_prefix_ror}rubygem(sprockets-rails) < 3
+BuildRequires: %{?scl_prefix}rubygem(sprockets-rails) >= 2.2.2
```
domcleal: 2.3.3?

domcleal:
Also the requires on scl_prefix_ror for sprockets-rails in the main foreman spec file should be changed to scl_prefix, and updated to 2.3.3.

dLobatog:
Updated, I realized they were ror and not tfm right after pushing :/

foreman/foreman.spec
```diff
@@ -109,8 +109,8 @@ Requires: %{?scl_prefix}rubygem(protected_attributes) >= 1.1.1
 Requires: %{?scl_prefix}rubygem(protected_attributes) < 1.2.0
 Requires: %{?scl_prefix}rubygem(sprockets) >= 3
 Requires: %{?scl_prefix}rubygem(sprockets) < 4
-Requires: %{?scl_prefix_ror}rubygem(sprockets-rails) >= 2.2.2
-Requires: %{?scl_prefix_ror}rubygem(sprockets-rails) < 3
+Requires: %{?scl_prefix}rubygem(sprockets-rails) >= 2.2.3
```
domcleal: `2.3.3` still :)

dLobatog:
Duh, sorry for the typo, updated

domcleal:
Thanks @dLobatog, building.

PR #952, opened 2016-01-06
Update foreman_setup to 3.1.0 [deb]

domcleal:
Only for deb/develop please.  It's compatible with 1.9+ up, but not needed right now.


mmoll:
seems to work just fine, merged, thanks @domcleal!
building... :construction:

PR #953, opened 2016-01-07
Configure orgs/locations in foreman_precompile_plugin macro

ehelms:
Given some plugins have a requirement around organizations and
locations being enabled, and that the foreman_precompile_plugin
macro instantiates the full stack, enabling organizations and locations
in the settings file allows a broader range of plugins to use this
macro.


domcleal:
Merged, thanks @ehelms.

PR #954, opened 2016-01-07
Refs #12970 - Add sprockets 3.5.2 for Patternfly

dLobatog:
https://github.com/theforeman/foreman-packaging/pull/944


foreman/foreman.spec
```diff
@@ -194,6 +194,8 @@ BuildRequires: %{?scl_prefix_ror}rubygem(execjs) >= 1.4.0
 BuildRequires: %{?scl_prefix_ror}rubygem(execjs) < 2.5.0
 BuildRequires: %{?scl_prefix_ror}rubygem(jquery-rails) >= 3.1
 BuildRequires: %{?scl_prefix_ror}rubygem(jquery-rails) < 4.0
+BuildRequires: %{?scl_prefix_ror}rubygem(sprockets) >= 3
```
domcleal: a) scl_prefix as it's not from the ror collection
b) move it down a little to preserve assets.rb layout

foreman/foreman.spec
```diff
@@ -372,6 +374,8 @@ Requires: %{?scl_prefix_ror}rubygem(uglifier) >= 1.0.3
 Requires: %{?scl_prefix_ror}rubygem(execjs) >= 1.4.0
 Requires: %{?scl_prefix_ror}rubygem(jquery-rails) >= 3.1
 Requires: %{?scl_prefix_ror}rubygem(jquery-rails) < 4.0
+Requires: %{?scl_prefix_ror}rubygem(sprockets) >= 3
```
domcleal: http://projects.theforeman.org/issues/13074 ought to be fixed as this should probably be in the main foreman deps, not assets only.

rubygem-sprockets/rubygem-sprockets.spec
```diff
@@ -0,0 +1,81 @@
...
+Summary: Rack-based asset packaging system
+Group: Development/Languages
+License: MIT
+URL: https://github.com/rails/sprockets
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
+Requires: %{?scl_prefix_ruby}ruby(release)
+Requires: %{?scl_prefix_ror}rubygem(rack) > 1
+Requires: %{?scl_prefix_ror}rubygem(rack) < 3
+Requires: %{?scl_prefix_ror}rubygem(concurrent-ruby) >= 1
```
domcleal: This isn't in the ror collection, but we do happen to have it in the plugins repo.  It'll need moving (update tito props and comps/) and to use scl_prefix here.

domcleal:
The existing RPM from the SCL also has a %check section, so please consider importing and updating that spec file rather than creating a new one from scratch.

The rh-ror41 package is 2.12.3, but I have a newer 3.2.0 (from Rawhide) at https://github.com/domcleal/sclo-ror42/tree/master/rubygem-sprockets which might be a more useful starting point.

comps/comps-foreman-fedora21.xml
```diff
@@ -38,6 +38,8 @@
       <packagereq type="default">rubygem-autoparse</packagereq>
       <packagereq type="default">rubygem-bootstrap-sass</packagereq>
       <packagereq type="default">rubygem-clamp</packagereq>
+      <packagereq type="default">rubygem-concurrent-ruby</packagereq>
+      <packagereq type="default">rubygem-concurrent-ruby-edge</packagereq>
```
domcleal: -edge should stay in plugins I think, it's a separate package and isn't used here.

domcleal:
tito.props also needs updating to remove rubygem-concurrent-ruby from the list of packages built in plugins.

> The existing RPM from the SCL also has a %check section, so please consider importing and updating that spec file rather than creating a new one from scratch.

Unresolved comment?

dLobatog:
Latest update:

* Bases spec off sclo-ror42 spec 
* Adds tests
* Leaves rubygemn-concurrent-edge on plugins

rubygem-sprockets/rubygem-sprockets.spec
```diff
@@ -0,0 +1,163 @@
...
+mv test/test_yui_compressor.rb{,.disabled}
+mv lib/sprockets/autoload/yui.rb{,.disabled}
+sed -i '/:YUI/ s/^/#/' lib/sprockets/autoload.rb
+
+# Required by TestPathUtils#test_find_upwards test.
+touch Gemfile
+
+%{?scl:scl enable %{scl} %{scl_v8} - << \EOF}
+RUBYOPT=-Ilib:test ruby -e 'Dir.glob "./test/**/test_*.rb", &method(:require)' \
+  | grep "763 runs, 3278 assertions, 0 failures, 3 errors, 1 skips"
```
domcleal: This fails on Fedora 21, where it's returning `763 runs, 3280 assertions, 0 failures, 3 errors, 0 skips`.  Do you know what the skip is?

dLobatog: I think I added the wrong tests tgz to this PR - I compared it with the one I have at https://github.com/dLobatog/foreman-packaging/commits/rpm/patternfly-all, rebuilt it and they're different from the one here. I've updated the PR with the right one.

dLobatog: How weird it passed with 0 skips on Fedora, as there's a test that is skipped if the host is not running Windows - `TestURITar#test_"works with windows" [/home/daniel/workspace/sprockets/test/test_uri_tar.rb:41]`

dLobatog: I've been trying to get minitest to run in `--verbose` mode to print the skipped tests here but I couldn't because I can't pass the argument. I tried to run `rake test TESTOPTS='--verbose'` to no avail. 

In any case, the skip locally on f23 is the one I mentioned before about Windows, so I assume it's the same.

dLobatog:
@domcleal Rebased

rubygem-sprockets/rubygem-sprockets.spec
domcleal: This is failing on EL6 and EL7 too now with:

<pre>
  1) Failure:
TestManifest#test_"raises exception when gzip fails" [/builddir/build/BUILD/rubygem-sprockets-3.5.2/opt/theforeman/tfm/root/usr/share/gems/gems/sprockets-3.5.2/test/test_manifest.rb:685]:
kaboom.
[] exception expected, not
Class: <ArgumentError>
Message: <"wrong number of arguments (2 for 0)">
---Backtrace---
/builddir/build/BUILD/rubygem-sprockets-3.5.2/opt/theforeman/tfm/root/usr/share/gems/gems/sprockets-3.5.2/test/test_manifest.rb:684:in `block (2 levels) in <class:TestManifest>'
---------------
777 runs, 3383 assertions, 1 failures, 0 errors, 2 skips
</pre>


> I think I added the wrong tests tgz to this PR

What do you mean?  Are you patching the tests?

dLobatog: No, I initially put the tests tgz for an older version of sprockets by mistake. I realized at the time and updated the PR with the 3.5.2 tests tgz. I'll update the grep

dLobatog:
Update the grep condition to match the 2 skips

rubygem-sprockets/rubygem-sprockets.spec
```diff
@@ -0,0 +1,163 @@
...
+mv test/test_yui_compressor.rb{,.disabled}
+mv lib/sprockets/autoload/yui.rb{,.disabled}
+sed -i '/:YUI/ s/^/#/' lib/sprockets/autoload.rb
+
+# Required by TestPathUtils#test_find_upwards test.
+touch Gemfile
+
+%{?scl:scl enable %{scl} %{scl_v8} - << \EOF}
+RUBYOPT=-Ilib:test ruby -e 'Dir.glob "./test/**/test_*.rb", &method(:require)' \
+  | grep "777 runs, 3383 assertions, 1 failures, 0 errors, 2 skips"
```
domcleal: This is still failing on Fedora 21 with the assertion and skip number differences:

    777 runs, 3385 assertions, 1 failures, 0 errors, 1 skips

dLobatog:
@domcleal sprockets build was green, http://ci.theforeman.org/job/packaging_build_rpm/3203/, foreman is still failing because sprockets-rails is missing.

comps/comps-foreman-rhel6.xml
```diff
@@ -55,6 +55,8 @@
       <packagereq type="default">tfm-rubygem-bootstrap-sass</packagereq>
       <packagereq type="default">tfm-rubygem-bundler_ext</packagereq>
       <packagereq type="default">tfm-rubygem-clamp</packagereq>
+      <packagereq type="default">tfm-rubygem-concurrent-ruby</packagereq>
+      <packagereq type="default">tfm-rubygem-concurrent-ruby-edge</packagereq>
```
domcleal: remove -edge from here, plugins only

comps/comps-foreman-rhel6.xml
```diff
@@ -170,6 +173,8 @@
       <packagereq type="default">rubygem-awesome_print</packagereq>
       <packagereq type="default">rubygem-bundler_ext</packagereq>
       <packagereq type="default">rubygem-clamp</packagereq>
+      <packagereq type="default">rubygem-concurrent-ruby</packagereq>
```
domcleal: Ah, this is a little tricky.. the non-SCL version is still really only needed in plugins.  I think it'll have to stay here (in core), so can you remove it from plugins too please?

comps/comps-foreman-rhel7.xml
```diff
@@ -55,6 +55,8 @@
       <packagereq type="default">tfm-rubygem-bootstrap-sass</packagereq>
       <packagereq type="default">tfm-rubygem-bundler_ext</packagereq>
       <packagereq type="default">tfm-rubygem-clamp</packagereq>
+      <packagereq type="default">tfm-rubygem-concurrent-ruby</packagereq>
+      <packagereq type="default">tfm-rubygem-concurrent-ruby-edge</packagereq>
```
domcleal: ditto, no -edge

foreman/foreman.spec
```diff
@@ -107,8 +107,8 @@ Requires: %{?scl_prefix}rubygem(rails-observers) >= 0.1
 Requires: %{?scl_prefix}rubygem(rails-observers) < 1.0
 Requires: %{?scl_prefix}rubygem(protected_attributes) >= 1.1.1
 Requires: %{?scl_prefix}rubygem(protected_attributes) < 1.2.0
-Requires: %{?scl_prefix_ror}rubygem(sprockets-rails) >= 2.2.2
-Requires: %{?scl_prefix_ror}rubygem(sprockets-rails) < 3
+Requires: %{?scl_prefix}rubygem(sprockets-rails) >= 2.2.2
```
domcleal: `2.3.3`?

foreman/foreman.spec
```diff
@@ -182,8 +182,8 @@ BuildRequires: %{?scl_prefix}rubygem(rails-observers) >= 0.1
 BuildRequires: %{?scl_prefix}rubygem(rails-observers) < 1.0
 BuildRequires: %{?scl_prefix}rubygem(protected_attributes) >= 1.1.1
 BuildRequires: %{?scl_prefix}rubygem(protected_attributes) < 1.2.0
-BuildRequires: %{?scl_prefix_ror}rubygem(sprockets-rails) >= 2.2.2
-BuildRequires: %{?scl_prefix_ror}rubygem(sprockets-rails) < 3
+BuildRequires: %{?scl_prefix}rubygem(sprockets-rails) >= 2.2.2
```
domcleal: 2.3.3?

foreman/foreman.spec
domcleal: Oh hang on, this is in the wrong PR, should be in the sprockets-rails one.

foreman/foreman.spec
```diff
@@ -199,6 +199,8 @@ BuildRequires: %{?scl_prefix}rubygem(autoprefixer-rails) >= 5.2
 BuildRequires: %{?scl_prefix}rubygem(autoprefixer-rails) < 6.0
 BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) >= 2.7
 BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) < 3.0
+BuildRequires: %{?scl_prefix}rubygem(sprockets) >= 3
+BuildRequires: %{?scl_prefix}rubygem(sprockets) < 4
```
domcleal: `~> 3` is equivalent to `>= 3` really.  Is that a mistake in Foreman's Gemfile?  If not, this line should be removed.

dLobatog: I'd call it a mistake on the Gemfile. However, I thought ~> 3 was equivalent to ~> 3.0. If I change it in Foreman to `~> 2` (*) , it installs `2.12.4` instead of `3.5.2`. Are you sure of that?

\* it doesn't actually work with foreman, just for the sake of testing

foreman/foreman.spec
dLobatog: Yup, that one changes it. I think I changed this when I was building the whole mockchain and this package needed to rely on the tfm sprockets-rails.

foreman/foreman.spec
domcleal: Ah yes, thanks!

dLobatog:
@domcleal On the latest update I just changed comps so that:

* plugins contain:
    * `rubygem-concurrent-ruby`
    * `rubygem-concurrent-ruby-edge`
    * `tfm-rubygem-concurrent-ruby-edge`
* fc21, el6, el7 contain:
    * `tfm-rubygem-concurrent-ruby`

Does that look correct? I have not changed foreman.spec to require the sprockets-rails the Gemfile requires as that's on https://github.com/theforeman/foreman-packaging/pull/949/files.

dLobatog:
Actually if using a sprockets-rails that's not built yet is the only problem that's not allowing foreman to build in Jenkins, let me update the foreman.spec to use the ror one and see if that gets both jobs green finally.

dLobatog:
Nope, it still fails to build Foreman because sprockets 3 is clearly not available yet. Anyway https://github.com/theforeman/foreman-packaging/pull/949 changes the foreman.spec correctly to use sprockets-rails from the tfm scl.

comps/comps-foreman-plugins-rhel6.xml
```diff
@@ -78,6 +77,8 @@
       <packagereq type="default">tfm-rubygem-safe_yaml</packagereq>
       <packagereq type="default">tfm-rubygem-wicked</packagereq>
       <packagereq type="default">tfm-rubygem-zscheduler</packagereq>
+      <packagereq type="default">rubygem-concurrent-ruby</packagereq>
```
domcleal: No need to add this to EL6, it isn't used by anything (it's only in EL7/F21 as the smart_proxy_dynflow plugin uses it.)

comps/comps-foreman-plugins-rhel6.xml
```diff
@@ -78,6 +77,8 @@
...
+      <packagereq type="default">rubygem-concurrent-ruby-edge</packagereq>
```
domcleal: ditto

foreman/foreman.spec
```diff
@@ -380,6 +382,8 @@ Requires: %{?scl_prefix}rubygem(autoprefixer-rails) >= 5.2
 Requires: %{?scl_prefix}rubygem(autoprefixer-rails) < 6.0
 Requires: %{?scl_prefix}rubygem(patternfly-sass) >= 2.7
 Requires: %{?scl_prefix}rubygem(patternfly-sass) < 3.0
+Requires: %{?scl_prefix}rubygem(sprockets) >= 3
```
domcleal: This is in the wrong place, it should now be in the main foreman RPM provides, not assets (https://github.com/theforeman/foreman/commit/24f70e14f1da00760b46ce2377db1147d849cf73).

foreman/foreman.spec
```diff
@@ -199,6 +199,8 @@ BuildRequires: %{?scl_prefix}rubygem(autoprefixer-rails) >= 5.2
 BuildRequires: %{?scl_prefix}rubygem(autoprefixer-rails) < 6.0
 BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) >= 2.7
 BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) < 3.0
+BuildRequires: %{?scl_prefix}rubygem(sprockets) >= 3
```
domcleal: These lines should be further up, in the main Gemfile list between protected_attrs and sprockets-rails.

rel-eng/tito.props
```diff
@@ -349,7 +349,6 @@ disttag = .el7
 whitelist = rubygem-algebrick
   rubygem-apipie-params
   rubygem-chef-api
-  rubygem-concurrent-ruby
```
domcleal: Keep this, as it's staying in the non-SCL plugins for the smart proxy.

dLobatog:
Updated according to the comments

domcleal:
Thanks @dLobatog, building now.  I'll let you know when it's in the buildroots before you rebase sprockets-rails.

PR #955, opened 2016-01-07
Release foreman-tasks 0.7.10 (RPM)

stbenjam:
For develop.  1.10 PR here -> https://github.com/theforeman/foreman-packaging/pull/956


domcleal:

domcleal:
I missed something in this PR and also in #957.  The new version has assets in the precompilation list, but neither PR enables asset precompilation, so I'd bet the uglifier/compilation error will still occur with these packages.

Would you mind opening new PRs to enable it?  For RPMs it's essentially adding the `-s` option to the precompile line and adding foreman-assets as a BR, and for the Debian package it's adding precompilation to plugins/ruby-foreman-tasks/debian/rules (see others).

stbenjam:
Oops, sorry, forgot about that.

https://github.com/theforeman/foreman-packaging/pull/958
https://github.com/theforeman/foreman-packaging/pull/959
https://github.com/theforeman/foreman-packaging/pull/960

PR #958, opened 2016-01-08
Handle assets, require foreman 1.9 for foreman-tasks (RPM/develop)

stbenjam:
I also noticed foreman-tasks requires 1.9 or greater.

### DIFF
--- a/rubygem-foreman-tasks/rubygem-foreman-tasks.spec
+++ b/rubygem-foreman-tasks/rubygem-foreman-tasks.spec
@@ -18,12 +18,12 @@
 Summary: Tasks support for Foreman with Dynflow integration
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.7.10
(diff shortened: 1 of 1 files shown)


domcleal:
You may need to override the precompilation rake task name, as the default based on the gem name won't be right (the plugin appears to register itself with an underscore).

    # -r<rake_task>     Overrides rake task of plugin:assets:precompile[plugin_name]

stbenjam:
Green

domcleal:
It looks OK to me - did you want to test the build first to check, or go ahead anyway?  If so, the 1.10 EL7 build is at http://koji.katello.org/koji/taskinfo?taskID=402914

stbenjam:
Ah yes, good idea.  Tested - and all the assets are there and the broken pages work now.  Thanks!

domcleal:
Thanks for checking, building now.

PR #959, opened 2016-01-08
Handle assets, require foreman 1.9 for foreman-tasks (RPM/1.10)

stbenjam:


domcleal:

PR #960, opened 2016-01-08
Handle assets for foreman-tasks, update foreman requirement (deb)

stbenjam:


domcleal:
[test] not sure what's happened with the Jenkins script, the wrong build is associated.

plugins/ruby-foreman-tasks/debian/rules
```diff
@@ -16,12 +16,15 @@ build:
 	cp $(PLUGIN).rb /usr/share/foreman/bundler.d/
 	cd /usr/share/foreman && ( \
 		bundle install --local && \
+		bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production && \
```
domcleal: This may also have to be foreman_tasks rather than $(PLUGIN)

mmoll:
when this is cherry-picked into 1.10, concurrent-ruby should be added to gem.list again

stbenjam:
Hm, I'm a little confused. I updated the rake task to use foreman_tasks, but it's still failing:

```
NoMethodError: undefined method `[]' for nil:NilClass
/usr/share/foreman/lib/tasks/plugin_assets.rake:43:in `initialize'
/usr/share/foreman/lib/tasks/plugin_assets.rake:82:in `new'
/usr/share/foreman/lib/tasks/plugin_assets.rake:82:in `block in <top (required)>'
```

Which indicates it can't find the SETTINGS that say which assets to compile: https://github.com/theforeman/foreman/blob/develop/lib/tasks/plugin_assets.rake#L43

But we do set it: https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/engine.rb#L24

domcleal:
The PR needs rebasing on the current deb/develop - it's based on an older commit which has version 0.7.8 of the gem.

Also you'll need to add a changelog entry and probably add `foreman_tasks var/lib/foreman/public/assets` or similar to the `install` file.

stbenjam:
> Also you'll need to add a changelog entry and probably add foreman_tasks var/lib/foreman/public/assets or similar to the install file.

The javascript assets aren't in a foreman_tasks directory, so I need to include just assets right? Or does the rake task put them in the right place regardless? I don't have a debian box to check

domcleal:
Ah yes, I think that'd be correct.

stbenjam:
Hm, progress I guess, new error.  Sprockets is upset about something.

TypeError: couldn't digest ActiveSupport::StringInquirer

domcleal:
Fixed in http://projects.theforeman.org/issues/13098.

domcleal:
[test]

plugins/ruby-foreman-tasks/debian/changelog
```diff
@@ -1,3 +1,7 @@
+ruby-foreman-tasks (0.7.10-2) stable; urgency=low
+
+  * Fix assets precompilation
+
```
domcleal: Missing a date below this line.

mmoll:
The package is looking good, but with my test system on current nightlies I'm getting `The Dynflow world was not initialized yet. If your plugin uses it, make sure to call ForemanTasks.dynflow.require! in some initializer` when accessing `/foreman_tasks/tasks`. I don't have other plugins requiring -tasks installed at the moment, could that be a problem?

stbenjam:
It sounds like you do need a plugin that requires it, but I thought foreman-tasks included some stuff (like async fact processing) that means it could be used standalone. 

Does the error go away if you install foreman_salt or remote execution which use tasks?

Maybe @iNecas would know more

mmoll:
I think there's no tasks-using plugin available in the package repos that's working with rails 4?

iNecas:
@mmoll yes, this is a known behaviour if no plugin that's using it is installed

mmoll:
Then everything is fine, merged and cherry-picked as 7f3bbac for 1.10, thanks @stbenjam!

PR #961, opened 2016-01-10
Refs #12970 - Package sass-rails 5 for Patternfly

dLobatog:


foreman/foreman.spec
```diff
@@ -187,8 +187,8 @@ BuildRequires: %{?scl_prefix_ror}rubygem(sprockets-rails) < 3
 # assets
 BuildRequires: %{?scl_prefix}rubygem(ace-rails-ap) >= 4.0.0
 BuildRequires: %{?scl_prefix}rubygem(ace-rails-ap) < 4.1.0
-BuildRequires: %{?scl_prefix_ror}rubygem(sass-rails) >= 4.0
-BuildRequires: %{?scl_prefix_ror}rubygem(sass-rails) < 5.0
+BuildRequires: %{?scl_prefix_ror}rubygem(sass-rails) >= 5
```
domcleal: If you're moving it into our collection, `scl_prefix` and not _ror.

rubygem-sass-rails/rubygem-sass-rails.spec
```diff
@@ -0,0 +1,76 @@
...
+Requires: %{?scl_prefix_ruby}ruby
+Requires: %{?scl_prefix_ror}rubygem(sass) >= 3.1
+Requires: %{?scl_prefix_ror}rubygem(sass) < 4
+Requires: %{?scl_prefix_ror}rubygem(railties) >= 4
+Requires: %{?scl_prefix_ror}rubygem(railties) < 5
+Requires: %{?scl_prefix}rubygem(sprockets) >= 2.8
+Requires: %{?scl_prefix}rubygem(sprockets) < 4
+Requires: %{?scl_prefix}rubygem(sprockets-rails) >= 4
+Requires: %{?scl_prefix}rubygem(sprockets-rails) < 2
+Requires: %{?scl_prefix}rubygem(tilt) >= 1.1
```
domcleal: tilt's in ror?

domcleal:
This spec file should be imported from an existing SCL or Fedora, not rewritten if possible, as we lose the tests, history etc.  Here's an SCLed copy of 5.0.4 if it helps - https://github.com/domcleal/sclo-ror42/blob/master/rubygem-sass-rails/rubygem-sass-rails.spec

dLobatog:
@domcleal I based the spec off the link you posted, so now it also has a check section. The only thing I'm not sure of now is whether this package should be 5.0.4-2, as 5.0.4-1 is already listed on the changelog.

domcleal:
Yeah, -2's a good idea as the contents have changed.

domcleal:
Please can you rebase this?  I think the build deps are in now.

dLobatog:
@domcleal Rebased, conflicts were minimal on comps and tito.props

domcleal:
Thanks @dLobatog, building now.

PR #963, opened 2016-01-12
Remove rubygem-less and rubygem-less-rails

ehelms:


domcleal:
Thanks, could you also add these packages to the obsoletes list in tfm/tfm.spec please?  (Seems an OK way to ensure removal of old things from within the tfm collection to me.)

ehelms:
Updated with obsoletes

tfm/tfm.spec
```diff
@@ -62,6 +62,8 @@ Obsoletes: ruby193-ruby-wrapper
 # and to ensure smooth upgrades when old packages deps aren't satisified
 Obsoletes: %{scl_prefix}rubygem-sass < 3.4.9-3
 Obsoletes: %{scl_prefix}rubygem-sprockets < 2.12.3-3
+Obsoletes: %{scl_prefix}less < 2.5.1-4
```
domcleal: -5, as the latest package in nightly is -4

tfm/tfm.spec
```diff
@@ -62,6 +62,8 @@ Obsoletes: ruby193-ruby-wrapper
...
+Obsoletes: %{scl_prefix}less-rails < 2.5.0-3
```
domcleal: -4

ehelms:
Corrected

domcleal:
Thanks @ehelms, merged.

domcleal:
Old packages untagged and tfm 2.0-2 built.

PR #968, opened 2016-01-15
plugins:docker - release 2.0.1 (DEB)

dLobatog:


mmoll:
which releases is this compatible with?

domcleal:
2.x is only compatible with 1.11+, so nightly.

domcleal:
Merged as e6de389, thanks @dLobatog.

PR #971, opened 2016-01-16
update foreman_xen to 0.2.4 (DEB)

mmoll:
also for 1.10


domcleal:
Merged as 8a5afe5, cecf4da and building

PR #975, opened 2016-01-18
package kafo_wizards (DEB)

mmoll:
will be required for the next kafo release

PR #979, opened 2016-01-20
Update Kafo to 0.7.0

ares:
requires https://github.com/theforeman/foreman-packaging/pull/975


domcleal:
Merged as 945ce11 and building, thanks @ares.

PR #984, opened 2016-01-21
Remote Execution - Iteration 5 - Develop (DEB)

stbenjam:
- [x] smart_proxy_remote_execution_ssh
- [x] smart proxy dynflow
- [x] foreman remote execution 0.2.1

### DIFF
--- a/plugins/ruby-foreman-remote-execution/debian/changelog
+++ b/plugins/ruby-foreman-remote-execution/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-remote-execution (0.2.1-1) stable; urgency=low
+
+  * 0.2.1 released
+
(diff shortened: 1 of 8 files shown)


mmoll:
Merging and building, although I know the core plugin won't install on 1.11.
Thanks @stbenjam! :ice_cream:

stbenjam:
it should actually work on 1.11 :)

mmoll:
then `foreman (<< 1.11)` should go from the `Depends` field :)

stbenjam:
Why's that? << is less than, right?

mmoll:
exactly

stbenjam:
So why would I include that? The plugin is compatible with 1.11

domcleal:
It's there at the moment - it can be removed.

stbenjam:
Oh sorry, I was only looking at the Build-Depends line. It doesn't matter, we're about to release 0.3.0  which drops 1.10 support, I'll fix it then

PR #987, opened 2016-01-22
Remote Execution - Iteration 5 - 1.10 (RPM)

stbenjam:


rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
```diff
@@ -8,7 +8,7 @@
 
 Summary: Foreman CLI plugin for showing task information for resources and users
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.0.8
+Version: 0.0.9
 Release: 1%{?dist}
```
domcleal: You'll need to add %{?foremandist} on rpm/develop and also here if you want to rebuild this on rpm/1.10.  Because the NEVR is identical, Koji won't permit the same build again - and since the develop version is incompatible (SCL changes), it needs a different release.

domcleal:
No need to update the two smart proxy plugins on this branch - we can't build them for the same reason as above (identical NEVR), but what we normally do is to retag the builds from nightly/develop because they're fully compatible (smart proxy is non-SCL).  I'll do that now, then you can drop the two smart proxy commits.

stbenjam:
Already, I dropped the smart proxy packages and added `%{?foremandist}` to hammer_cli_foreman_tasks - do you need the same change to develop too?

domcleal:
Yes, otherwise this version will be newer than the develop version:

    $ rpmdev-vercmp 0.0.9-1.fm1_10.el7 0.0.9-1.el7
    0.0.9-1.fm1_10.el7 > 0.0.9-1.el7

domcleal:
Merged, thanks @stbenjam.  Packages are building.

PR #989, opened 2016-01-25
Add foremandist to hammer_cli_foreman_tasks

stbenjam:


rubygem-hammer_cli_foreman_tasks/rubygem-hammer_cli_foreman_tasks.spec
```diff
@@ -9,7 +9,7 @@
 Summary: Foreman CLI plugin for showing task information for resources and users
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.0.9
-Release: 1%{?dist}
+Release: 1%{?foremandist}%{?dist}
```
stbenjam: Should the release get incremented to 2?

domcleal: No, don't worry - it'll happen when I tag the build with tito.

domcleal:

PR #996, opened 2016-01-27
Fixes #10161 - Convert foreman-installer config to scenario

mbacovsky:
- updated dependency constraint on Kafo
- updated scenario paths

### DIFF
--- a/foreman-installer/foreman-installer.spec
+++ b/foreman-installer/foreman-installer.spec
@@ -17,7 +17,7 @@ BuildArch:  noarch
 
 Requires:   curl
 Requires:   %{?scl_prefix}puppet >= 3.0.0
-Requires:   %{?scl_prefix}rubygem-kafo >= 0.6.5
+Requires:   %{?scl_prefix}rubygem-kafo >= 0.7.1
 Requires:   %{?scl_prefix}rubygem-apipie-bindings >= 0.0.6
 Requires:   foreman-selinux
 
@@ -31,7 +31,7 @@ Requires:   %{?scl_prefix}rubygem-highline
 BuildRequires: asciidoc
 BuildRequires: rubygem(rake)
 BuildRequires: %{?scl_prefix}puppet >= 3.0.0
-BuildRequires: %{?scl_prefix}rubygem-kafo
+BuildRequires: %{?scl_prefix}rubygem-kafo >= 0.7.1
 
 %description
 Complete installer for The Foreman life-cycle management system based on Puppet.
@@ -58,11 +58,16 @@ rake install \
   SYSCONFDIR=%{buildroot}%{_sysconfdir} \
   --trace
 
+%post
+foreman-installer --scenario foreman --migrations-only > /dev/null
+
 %files
 %defattr(-,root,root,-)
 %doc README.* LICENSE
-%config %attr(600, root, root) %{_sysconfdir}/foreman/%{name}.yaml
-%config(noreplace) %attr(600, root, root) %{_sysconfdir}/foreman/%{name}-answers.yaml
+%dir %{_sysconfdir}/foreman-installer
+%dir %{_sysconfdir}/foreman-installer/scenarios.d
+%config %attr(600, root, root) %{_sysconfdir}/foreman-installer/scenarios.d/foreman.yaml
+%config(noreplace) %attr(600, root, root) %{_sysconfdir}/foreman-installer/scenarios.d/foreman-answers.yaml
 %{_sbindir}/foreman-installer
 %{_datadir}/%{name}
 %{_mandir}/man8


mbacovsky:
Depends on https://github.com/theforeman/foreman-installer/pull/163

mbacovsky:
The script in `%post` (`foreman-installer --scenario foreman --migrations-only > /dev/null`) is there to apply migrations to Scenario config and answers. Also migration from "old style" configuration is involved. The migrations work similarly to rails db migrations. We start with basic config  to make automatic migration from old style config possible and then scenario migrations are applied. The user ends up with up-to-date config and answer files after installation, can observe and change the values there before running installer.

domcleal:
This PR's not mergeable, please rebase it.

I tried to build it as-is, but hit the following error during parameter export:

<pre>
** Execute /builddir/build/BUILD/foreman-installer-1.11.0/_build/options.asciidoc
/usr/share/gems/bin/kafo-export-params -c config/foreman.yaml -f asciidoc > /builddir/build/BUILD/foreman-installer-1.11.0/_build/options.asciidoc
/usr/share/gems/gems/kafo_parsers-0.0.5/lib/kafo_parsers/puppet_module_parser.rb:32:in `initialize': uninitialized constant KafoParsers::PuppetModuleParser::ModuleName (NameError)
	from /usr/share/gems/gems/kafo_parsers-0.0.5/lib/kafo_parsers/puppet_module_parser.rb:18:in `new'
	from /usr/share/gems/gems/kafo_parsers-0.0.5/lib/kafo_parsers/puppet_module_parser.rb:18:in `parse'
	from /usr/share/gems/gems/kafo-0.7.1/lib/kafo/puppet_module.rb:51:in `parse'
	from /usr/share/gems/gems/kafo-0.7.1/lib/kafo/configuration.rb:87:in `block in modules'
	from /usr/share/gems/gems/kafo-0.7.1/lib/kafo/configuration.rb:87:in `map'
	from /usr/share/gems/gems/kafo-0.7.1/lib/kafo/configuration.rb:87:in `modules'
	from /usr/share/gems/gems/kafo-0.7.1/bin/kafo-export-params:90:in `print_out'
	from /usr/share/gems/gems/kafo-0.7.1/bin/kafo-export-params:40:in `execute'
	from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:68:in `run'
	from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:133:in `run'
	from /usr/share/gems/gems/kafo-0.7.1/bin/kafo-export-params:135:in `<top (required)>'
	from /usr/share/gems/bin/kafo-export-params:23:in `load'
	from /usr/share/gems/bin/kafo-export-params:23:in `<main>'
rake aborted!
Command failed with status (1): [/usr/share/gems/bin/kafo-export-params -c ...]
</pre>

This seems to indicate that a file of some sort is missing, but the SRPM I'm looks OK.  Could you check this builds properly?

mmoll:
the same is happening in the DEB build modeled after this PR.

mbacovsky:
I rebased the PR.
@domcleal, I was able to rebuild with my srpm but the srpm you've provided failed. My build is http://koji.katello.org/koji/taskinfo?taskID=420041. 

While investigating it I found there is a bug in `kafo_prsers` preventing debugging (PR with a fix is https://github.com/theforeman/kafo_parsers/pull/12). kafo_parsers originaly tried to complain about some missing manifest. 
The difference between the two srpms were 
- puppet dependency (>=2.7 vs. >=3.0)
- my foreman-installer tarball contained extra `alternatives`, `extlib`, `inifile`, `puppetdb` and `staging` modules (probably caused by generating it localy on Fedora). Fixed puppet_parser could reveal what manifest is missing. I'll know better tomorrow.

mbacovsky:
I created foreman-installer tarball on clean CentOS7 box and the package built okay with the data: http://koji.katello.org/koji/taskinfo?taskID=420604 
I checked the tarball content and it also contains more packages then the tarball @domcleal used.
Interestingly enough also @mmoll met the same behaviour. Can you provide me your steps to build the foreman-installer tarball?

mmoll:
@mbacovsky I didn't build it from hand but see the Jenkins output in #1015

domcleal:
Thanks, I've been able to build it now.  I had an old _build/ dir in my foreman-installer checkout that was being re-used, so after the kafo_parsers fix I could see that it was missing a manifest that should've been present in more recent modules.

foreman-installer/foreman-installer.spec
```diff
@@ -58,11 +58,14 @@ rake install \
...
 
+%post
+foreman-installer --scenario foreman --migrations-only > /dev/null
+
 %files
 %defattr(-,root,root,-)
 %doc README.* LICENSE
-%config %attr(600, root, root) %{_sysconfdir}/foreman/%{name}.yaml
-%config(noreplace) %attr(600, root, root) %{_sysconfdir}/foreman/%{name}-answers.yaml
+%config %attr(600, root, root) %{_sysconfdir}/foreman/installer-scenarios.d/foreman.yaml
```
domcleal: This package needs to install and own the installer-scenarios.d directory as it isn't owned by anything else.  Ideally perhaps it'd be under /etc/foreman-installer/scenarios.d/ to match the log file layout?

foreman-installer/foreman-installer.spec
```diff
@@ -58,11 +58,14 @@ rake install \
   SYSCONFDIR=%{buildroot}%{_sysconfdir} \
   --trace
 
+%post
+foreman-installer --scenario foreman --migrations-only > /dev/null
```
domcleal: Is specifying the scenario here going to cause any problems if more are added to the package?

mbacovsky: It should be run for each scenario. It is here for the configuration to get updated before users can touch it and is more of cosmetic thing. 
Installer ensures the migrations are run on its start anyway.

foreman-installer/foreman-installer.spec
mbacovsky: Good point, I updated this PR and foreman-installer PR to use `/etc/foreman-installer/scenarios.d/`.

foreman-installer/foreman-installer.spec
```diff
@@ -58,11 +58,16 @@ rake install \
...
+
 %files
 %defattr(-,root,root,-)
 %doc README.* LICENSE
-%config %attr(600, root, root) %{_sysconfdir}/foreman/%{name}.yaml
-%config(noreplace) %attr(600, root, root) %{_sysconfdir}/foreman/%{name}-answers.yaml
+%dir %{_sysconfdir}/foreman-installer
```
domcleal: Nitpick, prefer %{name} rather than repeating foreman-installer here and below.

domcleal:
[test] to verify it builds after the installer change was merged.

domcleal:
[test] fixed tests as I'd temporarily disabled a Jenkins job because of the change.

domcleal:
Merged as c3f00e7, thanks @mbacovsky.

foreman-installer/foreman-installer.spec
domcleal: Changed in 5ea91c1.

PR #999, opened 2016-01-27
Release foreman_salt 5.0.0 (DEB)

stbenjam:
Nightly only, rails 4 compatibility release

### DIFF
--- a/plugins/ruby-foreman-salt/debian/control
+++ b/plugins/ruby-foreman-salt/debian/control
@@ -2,12 +2,12 @@ Source: ruby-foreman-salt
 Section: ruby
 Priority: extra
 Maintainer: Stephen Benjamin <stephen@redhat.com>
-Build-Depends: debhelper (>= 8.0.0), foreman-assets (>= 1.10.0~rc1), foreman (>= 1.10.0~rc1), foreman-sqlite3 (>= 1.10.0~rc1), ruby-foreman-deface, ruby-foreman-tasks (>= 0.7.1), ruby-foreman-tasks (<< 0.8.0)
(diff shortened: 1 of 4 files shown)


plugins/ruby-foreman-salt/debian/control
```diff
@@ -2,12 +2,12 @@ Source: ruby-foreman-salt
 Section: ruby
 Priority: extra
 Maintainer: Stephen Benjamin <stephen@redhat.com>
-Build-Depends: debhelper (>= 8.0.0), foreman-assets (>= 1.10.0~rc1), foreman (>= 1.10.0~rc1), foreman-sqlite3 (>= 1.10.0~rc1), ruby-foreman-deface, ruby-foreman-tasks (>= 0.7.1), ruby-foreman-tasks (<< 0.8.0)
+Build-Depends: debhelper (>= 8.0.0), foreman-assets (>= 1.10.0~rc1), foreman (>= 1.11), foreman-sqlite3 (>= 1.11), ruby-foreman-deface, ruby-foreman-tasks (>= 0.7.1), ruby-foreman-tasks (<< 0.8.0)
```
domcleal: You'll want to keep the whole `.0~rc1` in these strings (and below) because 1.11.0~rc1 is actually *lower* than 1.11.0.  Specifying `>= 1.11` or even `>= 1.11.0` will exclude RCs and only match the final release.

mmoll:
what @domcleal is saying, :+1: otherwise.

stbenjam:
Fixed I think, thanks for the clarification, I didn't realize that's why we always included the ~rc thingy

mmoll:
merged, thanks @stbenjam!
building now...

stbenjam:
Thanks!!

PR #1002, opened 2016-01-28
Update kafo to 0.7.1 (RPM)

ares:


domcleal:
Thanks @ares, merging and building now.

PR #1004, opened 2016-02-01
Require Puppet 3.0 for foreman-installer [deb]

domcleal:
As per https://groups.google.com/forum/#!msg/foreman-users/YQNfxUzG_As/FF7vIfTyDQAJ


mmoll:
merged as 2adb386, thanks @domcleal!

PR #1005, opened 2016-02-02
Add extras to repoclosure

jlsherrill:
allows plugins to pull dependencies from that repository

### DIFF
--- a/repoclosure/yum_el6.conf
+++ b/repoclosure/yum_el6.conf
@@ -23,6 +23,10 @@ enabled=1
 mirrorlist=http://mirrorlist.centos.org/?release=6&arch=x86_64&repo=updates
 failovermethod=priority
 
+[el6-extras]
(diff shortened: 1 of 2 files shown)


repoclosure/yum_el6.conf
```diff
@@ -23,6 +23,11 @@ enabled=1
 mirrorlist=http://mirrorlist.centos.org/?release=6&arch=x86_64&repo=updates
 failovermethod=priority
 
+[el6-extras]
+name=extras
+enabled=1
```
domcleal: Please keep this disabled by default, it's a CentOS-only repository and we shouldn't assume it's available usually.

jlsherrill:
apologies, was following the convention above rather than below, updated!

domcleal:
Merged, thanks @jlsherrill.

PR #1006, opened 2016-02-02
Fixes #13502 - add docker-selinux to build requires

lzap:
We don't use RHEL 7.2 on our builders yet, but downstream this was failing
because we use one interface and type from that file. This fixes it.

In RHEL/CentOS7 this is in extras repo.

### DIFF
--- a/foreman-selinux/foreman-selinux.spec
+++ b/foreman-selinux/foreman-selinux.spec
@@ -48,6 +48,7 @@ Source0:        http://downloads.theforeman.org/%{name}/%{name}-%{version}%{?das
 BuildRequires:  checkpolicy, selinux-policy-devel, hardlink
 BuildRequires:  policycoreutils >= %{selinux_policycoreutils_ver}
 BuildRequires:  /usr/bin/pod2man
+BuildRequires:  docker-selinux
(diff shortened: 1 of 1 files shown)


lzap:
I wonder how to find out it's RHEL 7.2 in SPEC...

lzap:
Hmmm I believe there is no way to do it, we need to wait until we move our builders to RHEL 7.2 and then add this.

lzap:
@domcleal opinon? I could perhaps add a test in this patch and if `docker.if` is not present, we should fail. Otherwise I will definitely forget to add the BuildRequire later on ...

domcleal:
I don't see any way to make this spec file conditional for EL7.0 versus 7.2, but I agree that it should fail hard if docker.if is missing, which means changing the logging in foreman-selinux to a hard failure.

lzap:
Right, added.

domcleal:
So this PR should be closed, since docker-selinux can't be added, right?

lzap:
Oh yeah, sorry of course. In downstream we hit this because we build on/from RHEL 7.2, that's a different case. This will hit us once we upgrade our base OS on koji.

PR #1007, opened 2016-02-04
Release smart_proxy_dynflow 0.0.6 (DEB)

stbenjam:
Nightly + 1.10


plugins/smart_proxy_dynflow/postinst
```diff
@@ -0,0 +1,32 @@
...
+set -e
+
+USERNAME="foreman-proxy"
+GROUPNAME="foreman-proxy"
+DYNFLOW_DB_DIR="/var/lib/foreman-proxy/dynflow"
+
+case "$1" in
+  configure)
+    mkdir -p $DYNFLOW_DB_DIR
+    chown $USERNAME:$GROUPNAME $DYNFLOW_DB_DIR
```
domcleal: Could you also set a restrictive mode for the DB here?  It'll prevent any user reading the DB.

stbenjam: Done below, 750 ok?

domcleal: Yes, I think 750 will be fine.

mmoll:
cherry-picked as 64ebcc0 for develop and as 07adea7 for 1.10, thanks @stbenjam!

PR #1008, opened 2016-02-04
Release smart_proxy_dynflow 0.0.6 (RPM)

stbenjam:
Nightly + 1.10


rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -54,6 +54,9 @@ This package contains documentation for rubygem-%{gem_name}.
 %install
 mkdir -p %{buildroot}%{gem_dir}
 
+mkdir -p %{buildroot}%{_localstatedir}/lib/foreman-proxy/dynflow
+chown foreman-proxy:foreman-proxy %{buildroot}%{_localstatedir}/lib/foreman-proxy/dynflow
```
domcleal: Ownership should go into the %files list using %attr(...), as the foreman-proxy user ID won't be known during build (and may change).

domcleal:
I think the test failure's because the version number in the spec file needs updating to 0.0.6.

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
domcleal: In a similar train of thought, it'd be useful I think to specify a restrictive mode for the directory (e.g. 0700 or 0750) to prevent any user reading the DB.

domcleal:
Merged and building, thanks @stbenjam.  I'll retag it to 1.10.

PR #1011, opened 2016-02-04
Package tasks and rex CLI plugins (DEB)

stbenjam:
These were never packaged for debian.  For 1.10 + nightly.


mmoll:
cherry-picked as 56238de & b7b7931 for develop and af91cef & 05ecf30 for 1.10

PR #1014, opened 2016-02-06
use upstream packaging files for concurrent-ruby (DEB)

mmoll:
As Debian packaged concurrent-ruby in thier repos, but with a different name (missing -ruby) and this already reached Ubuntu/xenial, it would be nice to pull this in also here to have dependent packages unified.


domcleal:
It might be a good idea to rename the directory under dependencies/ too, else we're going to have to pass concurrent-ruby to the Jenkins job, despite the new name.

mmoll:
done.

domcleal:
Oh sorry, it looks like it was right to begin with.  The test is now failing as it can't find a gem called `ruby-concurrent` (http://ci.theforeman.org/job/packaging_build_deb_dependency/1094/arch=armv7,label=debian,os=jessie/console).  I guess it relies entirely on the directory name, not minding what the package itself is called and that we'll have to remember it's different.

Would you mind putting it back?  Hopefully it won't cause any problems that way.  Otherwise tests fine.

mmoll:
sure, changed back.

domcleal:
Merged, thanks @mmoll!

Building:

* http://ci.theforeman.org/job/packaging_build_deb_dependency/1098/
* http://ci.theforeman.org/job/packaging_build_deb_dependency/1099/
* http://ci.theforeman.org/job/packaging_build_deb_dependency/1100

PR #1015, opened 2016-02-07
refs #10369 scenarios support for foreman-installer (DEB)

mmoll:
Completely untested, probably best to look at, when the support in -installer got merged.

Also unify dependency onto apipie-bindings and have kafo as build dependency.


debian/jessie/foreman-installer/postinst
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+# postinst script for smart-proxy
```
mbacovsky: for foreman-installer

mmoll:
[test]

domcleal:
[test] as I've merged the PR to foreman-installer itself.  The config paths changed from /etc/foreman to /etc/foreman-installer, but I don't think that'll affect this build.

domcleal:
I see it's failing to run kafo-export-params during the build, as it can't load the "configuration" file from kafo:

<pre>
16:07:39 /usr/bin/kafo-export-params -c config/foreman.yaml -f asciidoc > /var/lib/workspace/workspace/packaging_build_deb_coreproject/arch/x86/label/debian/os/jessie/debian/jessie/build-foreman-installer/foreman-installer-1.11.0/_build/options.asciidoc
16:07:39 /usr/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- configuration (LoadError)
16:07:39 	from /usr/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
16:07:39 	from /usr/bin/kafo-export-params:11:in `<main>'
16:07:39 rake aborted!
</pre>

I don't see why that's suddenly become an issue, as this doesn't appear to have changed recently.

I'm also unsure why kafo-export-params is changing the LOAD_PATH like it is (trying to add lib/kafo/ as an entry) and loading 'configuration' rather than 'kafo/configuration'.

debian/jessie/foreman-installer/control
```diff
@@ -2,7 +2,7 @@ Source: foreman-installer
 Section: ruby
 Priority: optional
 Maintainer: Greg Sutcliffe <gsutclif@redhat.com>
-Build-Depends: debhelper (>= 7.0.50~), asciidoc, rake, libxml2-utils,
+Build-Depends: debhelper (>= 7.0.50~), asciidoc, rake, libxml2-utils, ruby-kafo (>= 0.7.1),
```
domcleal: Oh, that's why it changed.  The package (well, the gem itself)'s kafo-export-params is broken.  Adding it to the build root means it tries to export parameters for the man page, exposing the problem.

I think you can safely leave this out of build-deps for now and it'll build as it used to do, without the known parameters listed in the man page.  We'll have to fix kafo-export-params and add it again later.

mmoll:
[test] again

domcleal:
The prprocessor (which triggers an immediate Jenkins run instead of the schedule) is failing because the ticket referenced is a tracker - use #10369.

debian/jessie/foreman-installer/postinst
domcleal: also unfixed

mmoll:
[test]

mmoll:
Allright, this built and the migration on my existing foreman installation seems to have gone well.

mmoll:
not sure if this is intentional (dpkg -L):
```
...
/etc/foreman-installer
/etc/foreman-installer/scenarios.d
/etc/foreman-installer/scenarios.d/foreman-answers.yaml
/etc/foreman-installer/scenarios.d/foreman.yaml
/etc/foreman/foreman-installer-answers.yaml
/etc/foreman/foreman-installer.yaml
```

domcleal:
Sorry, I just merged as 43d9927.

There should be some backup files left in /etc/foreman on an upgrade, but not those regular .yaml files.  On a clean installation it should only be in /etc/foreman-installer.  I'll take a closer look myself tomorrow too if they still show up.

domcleal:
Thanks @mmoll!

domcleal:
The newly built packages only contain /etc/foreman-installer/scenarios.d, so perhaps the dpkg -L output after an upgrade is listing previously managed config files too?  Do the files exist, or only the .bak files?

mmoll:
the files do exist, but if it's not in the dpkg -L output, everything should be fine :star2:

PR #1016, opened 2016-02-09
drop foreman-cli dependency onto foreman (RPM)

mmoll:
references theforeman/puppet-foreman#327, Debian packages never had that dependency


domcleal:
Thanks @mmoll!

PR #1017, opened 2016-02-10
plugins:foreman_ansible - Release 0.3

dLobatog:
For 1.9, 1.10 and nightlies.

@domcleal Why does this package now have two 'Provides'?

### DIFF
--- a/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
+++ b/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
@@ -5,21 +5,23 @@
 
 Summary: Ansible integration with Foreman (theforeman.org)
 Name:    %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.2.2
-Release: 2%{?foremandist}%{?dist}
+Version: 0.3
+Release: 1%{?foremandist}%{?dist}
 Group:   Applications/System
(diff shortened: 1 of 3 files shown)


rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -64,6 +66,7 @@ cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
 %{gem_spec}
 %{foreman_bundlerd_dir}/%{gem_name}.rb
 %exclude %{gem_instdir}/test
+%exclude %{gem_instdir}/extras
```
domcleal: This dir isn't in the gem, it shouldn't be listed (breaks EL6).

domcleal:
> @domcleal Why does this package now have two 'Provides'?

It always should have.  The rubygem(%name) one is a Fedora standard, to be preferred over rubygem-name when using dependencies.  The foreman-plugin-%name one is something we have on every plugin package as an easier way to install plugins, but we haven't really used it.

dLobatog:
Updated to not exclude the 'extras' directory which is not in the gem.

domcleal:
Thanks @dLobatog, merged and building for nightly, 1.10 and 1.9.

PR #1021, opened 2016-02-11
Release foreman_remote_execution 0.2.3 (RPM)

stbenjam:
This contains the view fix for rails update.  For 1.10 and nightly.


domcleal:
Thanks @stbenjam, building on both branches.

PR #1024, opened 2016-02-15
Release dynflow 0.8.10 (DEB)

stbenjam:
For nightly + 1.10


domcleal:
Re-running the test under http://ci.theforeman.org/job/packaging_build_deb_dependency/1104/, as there was a misconfig with the GPG key on freightstage after the web server migration this week.

domcleal:
Did you also want to bump the version that's inside the ruby-foreman-tasks package for the Foreman plugin?

stbenjam:
> Did you also want to bump the version that's inside the ruby-foreman-tasks package for the Foreman plugin?

You mean the dynflow dep in tasks? Shouldn't need it, requirements didn't change on tasks.

domcleal:
No, I mean the dynflow bundled inside ruby-foreman-tasks.  The Foreman plugin package contains -tasks and its dependencies, while the dependencies/ package here is only used for the smart proxy.

stbenjam:
Oh, sorry, I forgot about that being there.  Now it makes sense, updated.

domcleal:
Please also update the changelog for ruby-foreman-tasks, there's a helper in scripts/.

stbenjam:
How does this work then? It creates a ginormous tasks package with all the dependencies rolled in?

domcleal:
Yep, exactly.  It installs .gem files into ~foreman/vendor/cache/, the same location as the core Foreman package installs its vendored pile of .gems, then runs `bundle install --local` to activate them in Gemfile.lock.

mmoll:
[test]

mmoll:
Merged into develop and cherry-picked as b011b64 into 1.10, thanks @stbenjam!

PR #1025, opened 2016-02-16
Release foreman-tasks 0.7.14 (RPM)

stbenjam:


domcleal:
[test] as the wrong test results are associated.

I think the %files list is going to need a big update as there are new dot-files and directories.

stbenjam:
Green now, thanks

rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -159,6 +164,7 @@ exit 0
 %{foreman_apipie_cache_foreman}
 %{foreman_apipie_cache_plugin}
 %doc %{gem_instdir}/LICENSE
+%doc %{gem_instdir}/README.md
```
domcleal: This is already in the %doc subpackage, it shouldn't be in the main package.

stbenjam: Sorry, I always forget README doesn't go in the main package

domcleal:
Thanks @stbenjam, building to nightly only.

PR #1031, opened 2016-02-19
Add GitHub PR template to indicate branches for plugin updates

domcleal:
I tested on a repo and the user's own commit message gets appended to the template, after the horizontal rule.


ehelms:
:+1:

domcleal:
Merged to both develop branches.

domcleal:
Seems that this doesn't work, perhaps it's only used on the default branch.

PR #1034, opened 2016-02-21
update smart_proxy_dns_powerdns to 0.2.0

mmoll:
Also for 1.11, please.
@ekohl, please have a look

### DIFF
--- a/plugins/smart_proxy_dns_powerdns/changelog
+++ b/plugins/smart_proxy_dns_powerdns/changelog
@@ -1,3 +1,9 @@
+ruby-smart-proxy-dns-powerdns (0.2.0-1) stable; urgency=low
+
+  * Release 0.2.0
+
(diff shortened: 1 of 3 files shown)


ekohl:
It looks good. I do see that ruby-pg is only in jessie and later so we may need to backport it for wheezy. Same goes for ubuntu which only has it in wily.

mmoll:
Per https://packages.debian.org/search?keywords=ruby-pg and http://packages.ubuntu.com/search?keywords=ruby-pg it's there in all our supported distros?

ekohl:
Oh, I guess I didn't search right. :+1:

domcleal:
Merged, thanks @mmoll and as 086f2bf on deb/1.11.

Building at http://ci.theforeman.org/job/packaging_build_deb_proxy_plugin/150/ and http://ci.theforeman.org/job/packaging_build_deb_proxy_plugin/151/.

PR #1037, opened 2016-02-23
Update to foreman_chef 0.2.2

ares:
1.10 only please


domcleal:
@ares are you sure this is 1.10 compatible?  I see attr_accessible calls were added for instance, that'll change models to be entirely protected which could break a lot.  It looks more like develop + 1.11 only to me!

Note the test failures, the gem file lists have changed.

ares:
oh, I was under impression that this was 1.10 compatible (https://github.com/theforeman/foreman_chef/pull/34#issuecomment-172583701) but I see that there was no 1.10 attr_accessible specified in host model

so I'd have to wrap attr_accessible in rails version condition I guess

domcleal:
> oh, I was under impression that this was 1.10 compatible (theforeman/foreman_chef#34 (comment)) but I see that there was no 1.10 attr_accessible specified in host model

I don't know if the fact name attr_accessible will cause a problem or not because it's STI, but it's the host one in https://github.com/theforeman/foreman_chef/commit/8e26cb1c3795dec1a594856853fcaffa4217ec7b that caught my eye.

> so I'd have to wrap attr_accessible in rails version condition I guess

Yeah, either branch it and have one branch that has the calls and one that doesn't at all, or a conditional on either the Foreman or Rails version.

It should probably also go to rpm/develop first and be cherry picked down rather than 1.10 only?  (It's needed for 1.11/nightly).  The reason I noticed the attr_accessible changes were that I was checking if you had another branch for 1.11 compatibility, but it seems not.

ares:
1.11 branch is waiting for 0.3.0 (https://github.com/theforeman/foreman_chef/pull/41) which was planned to bump foreman dependency, I think I'll remove attr_accessible and release 0.2.2 async and base 0.3.0 on top of 0.2.1. I already have 0.2-stable branch.

> It should probably also go to rpm/develop first and be cherry picked down rather than 1.10 only?

I'll push 0.3.0 to rpm/develop shortly after 0.2.2, there's currently no foreman_chef in nightly because it's incompatible so I think it's fine. I'll work on this tomorrow so if you have any other suggestions I'd appreciate them.

domcleal:
Sounds good!

ares:
@domcleal should be ready now, if you agree, could you please also confirm in #1038?

domcleal:
Thanks @ares, merged and building in 1.10 now.

PR #1039, opened 2016-02-25
templates: correct plugin name in postinst (DEB)

mmoll:
down to 1.9, please


domcleal:
Could you add a changelog entry please, so we can rebuild it?

mmoll:
seems like I was too tired yesterday... :)

domcleal:
Merged to all branches, thanks @mmoll.  Building now.

PR #1044, opened 2016-02-29
Add missing polyglot dependency on 1.11+

domcleal:
On Foreman 1.10 via Rails 3, mail 2.5.x and its treetop dependency,
polyglot was included in the main Foreman gem cache.  On 1.11 with
mail 2.6.x, the treetop dependency was removed and so it has to be
added to the deface package.

Should be cherry picked to deb/1.11 too please.

### DIFF
--- a/plugins/ruby-foreman-deface/debian/changelog
+++ b/plugins/ruby-foreman-deface/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-deface (1.0.2-2) stable; urgency=low
+
+  * Add missing polyglot dependency on 1.11+
+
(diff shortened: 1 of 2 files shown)


domcleal:
[test]

domcleal:
I built Discovery on top of this in my stagingdeb repo successfully: http://ci.theforeman.org/job/packaging_build_deb_plugin/arch=x86,label=debian/994/

mmoll:
good catch, cherry-picked as 88df99d for develop and 89d1de1 for 1.11, thanks @domcleal!

PR #1045, opened 2016-03-01
refs #13947 - remove public, views directories

domcleal:
Should fix http://ci.theforeman.org/job/packaging_build_deb_coreproject/2931/ since merge of https://github.com/theforeman/smart-proxy/commit/2578671415c1a212beb93a9fd6027a172961e38c (develop only).


mmoll:
merged, thanks @domcleal!

PR #1048, opened 2016-03-04
Update foreman_chef to 0.3.0

ares:


ares:
anyone know what's going on? the scratch seems to be built correctly in koji

domcleal:
One of the two packages failed: http://koji.katello.org/koji/getfile?taskID=432963&name=build.log

The test/ dir needs removing from %files as it's no longer shipped in the gem.  You also need to restore the comps/ entries for it so it gets built into the right tag.

ares:
thanks, I reverted comps and rerun comps_doc script which brought few more changes, I hope that's ok

comps/comps-foreman-plugins-fedora21.xml
```diff
@@ -110,6 +110,7 @@
       <packagereq type="default">rubygem-foreman_graphite-doc</packagereq>
       <packagereq type="default">rubygem-foreman_hooks-doc</packagereq>
       <packagereq type="default">rubygem-foreman_host_rundeck-doc</packagereq>
+      <packagereq type="default">rubygem-foreman_chef-doc</packagereq>
```
domcleal: Could you keep the file alphabetised please?  There are a few more below.

ares: ouch, I ran comps_doc.sh but it recognized my locale and considered 'ch' belongs after 'h', you know we have 'ch' as a separate letter... I'll try to memorize to set LC_ALL=C next time

comps/comps-foreman-rhel6.xml
```diff
@@ -270,8 +270,8 @@
       <packagereq type="default">tfm-rubygem-hirb-doc</packagereq>
       <packagereq type="default">tfm-rubygem-hirb-unicode-doc</packagereq>
       <packagereq type="default">tfm-rubygem-hoe-doc</packagereq>
-      <packagereq type="default">tfm-rubygem-i18n-doc</packagereq>
       <packagereq type="default">tfm-rubygem-ipaddress-doc</packagereq>
+      <packagereq type="default">tfm-rubygem-i18n-doc</packagereq>
```
domcleal: I think this is a difference between en_GB and cs_CZ locales/collation... I'd appreciate it if we could keep the order to maintain our sanity.  Perhaps set LC_COLLATE=C in comp_docs.sh?

comps/comps-foreman-fedora21.xml
```diff
@@ -144,17 +144,17 @@
       <packagereq type="default">rubygem-hirb-doc</packagereq>
       <packagereq type="default">rubygem-hirb-unicode-doc</packagereq>
       <packagereq type="default">rubygem-i18n-doc</packagereq>
-      <packagereq type="default">rubygem-jquery_pwstrength_bootstrap_4-doc</packagereq>
       <packagereq type="default">rubygem-jquery-turbolinks-doc</packagereq>
       <packagereq type="default">rubygem-jquery-ui-rails-doc</packagereq>
+      <packagereq type="default">rubygem-jquery_pwstrength_bootstrap_4-doc</packagereq>
```
domcleal: ditto comment about collation settings and locale, I think hyphens versus underscores is also a cz vs. en issue.

comps/comps-foreman-plugins-fedora21.xml
domcleal: Feel free to add LC_COLLATE=C into the script, then it's consistent for all.

comps/comps-foreman-fedora21.xml
ares: no matter what I set as locale I get this order, your env sorts it differently?

domcleal: Ah, that appears to be an en_* thing.  en_GB and en_US sort hyphens before underscores, while C sorts it the other way around.

ares: yeah, I just figured out, please see the script change, it should be fine now

domcleal:
Merged and also to rpm/1.11.  Thanks @ares.

ares:
Thanks you!

PR #1050, opened 2016-03-07
Update foreman_chef to 0.3.1

ares:
I copied changes from other template, just bumped foreman-plugin build dep, I'd appreciate if someone explained why this one is needed. I'll test scratches in production env so please wait with merging a bit.


ares:
seems to work for me, please tag into 1.11 as well (note that it also updates foreman_chef to 0.3.1), should I update commit/pr message

domcleal:
> I copied changes from other template, just bumped foreman-plugin build dep, I'd appreciate if someone explained why this one is needed

The foreman-plugin dependency, or generally those changes?  The foreman-plugin dep provides some build macros, all of those %foreman_precompile type ones for plugins to use at build time.

Generally though these steps are needed to provide a precompiled (compressed, optimised) .js/.css files inside the package for use in the production environment.  This means that at runtime, users don't have to precompile the entire set of application JS+CSS files so it's quick to install and they don't need asset dependencies installed.

The plugin build includes a manifest file (foreman_chef.json) stored in the plugin's installation directory, which lists all of the precompiled assets and their MD5 sums.  This is merged into the main application manifest when Foreman starts up for every registered plugin, so Rails is able to serve up the precompiled plugin assets.

rubygem-foreman_chef/rubygem-foreman_chef.spec
```diff
@@ -96,6 +94,11 @@ exit 0
 %doc %{gem_instdir}/README.md
 %{gem_instdir}/Rakefile
 
+%posttrans
+%foreman_db_migrate
```
domcleal: missing db:seed here (removed from line 79), a %foreman_db_seed macro should be available

domcleal:
> should I update commit/pr message

Can do, it'll improve the changelog if you say it's updated to 0.3.1.

rubygem-foreman_chef/rubygem-foreman_chef.spec
ares: should be fixed now

ares:
> The foreman-plugin dependency, or generally those changes? The foreman-plugin dep provides some build macros, all of those %foreman_precompile type ones for plugins to use at build time.

I was not clear on this one, thanks!

domcleal:
Thanks @ares, merged to both develop/1.11 and building.

PR #1053, opened 2016-03-09
Update safemode to 1.2.3

ares:


rubygem-safemode/rubygem-safemode.spec
```diff
@@ -6,12 +6,12 @@
...
 
-Version: 1.2.2
-Release: 3%{?dist}
+Version: 1.2.3
+Release: 1%{?dist}
 Group: Development/Ruby
 License: MIT
 URL: http://github.com/svenfuchs/safemode
-Source0: https://rubygems.org/downloads/%{gem_name}-%{version}.gem
+Source0: https://www.rubygems.org/downloads/%{gem_name}-%{version}.gem
```
domcleal: Change this to an HTTP URL.

ares: hm I just figured out :-)

ares:
Btw this is mainly required because of https://github.com/theforeman/community-templates/blob/develop/snippets/kickstart_networking_setup.erb#L111 so it would be good to tag it to 1.11 as well.

domcleal:
Thanks @ares, building and will retag for 1.11 to go out in RC2.

PR #1058, opened 2016-03-14
refs #12295 - uses USR1 signal to handle log rotation

witlessbird:


witlessbird:
Pls. see https://github.com/theforeman/smart-proxy/pull/394 for corresponding changes in smart-proxy

mmoll:
will there be also a PR for DEB? :heart:

witlessbird:
I'll take a look.

witlessbird:
pls. see https://github.com/theforeman/foreman-packaging/pull/1060 for debian-specific changes

foreman-proxy/foreman-proxy.init
```diff
@@ -58,6 +58,20 @@ stop() {
     return $RETVAL
 }
 
+logrotate() {
+    echo -n $"Rotating logs for $prog: "
+    if [ -f ${FOREMAN_PROXY_PID} ]; then
+        killproc -p ${FOREMAN_PROXY_PID} -USR1
```
domcleal: This is missing an argument between the pid file and `-USR1`, as currently it's issuing a SIGTERM every time.  Adding -x shows `+ base=-USR1`, but if you instead run:

    killproc -p /var/run/foreman-proxy/foreman-proxy.pid foreman-proxy -USR1

then it logs:

    + killlevel=-USR1
    + base=foreman-proxy

and the process gets rotated.

domcleal: It should be `$prog` rather than plain `foreman-proxy `too.

foreman-proxy/foreman-proxy.init
```diff
@@ -58,6 +58,20 @@ stop() {
...
+        RETVAL=$?
+    else
```
domcleal: I think there should be an extra `echo` after `RETVAL=$?` to ensure the output of the script is formatted correctly.

witlessbird:
How does this look?

foreman-proxy/foreman-proxy.init
```diff
@@ -58,6 +58,21 @@ stop() {
     return $RETVAL
 }
 
+logrotate() {
+    echo -n $"Rotating logs for $prog: "
+    if [ -f ${FOREMAN_PROXY_PID} ]; then
+        killproc -p ${FOREMAN_PROXY_PID} $prog -USR1
+        RETVAL=$?
+        echo -n
```
domcleal: plain `echo` here to get a newline, as `echo -n` won't print anything

witlessbird:
Removed "-n" from ```echo```.

domcleal:
Thanks @witlessbird!

PR #1060, opened 2016-03-15
refs #12295 - uses USR1 signal to handle log rotation (debian version)

witlessbird:
Pls. see theforeman/smart-proxy#394 for the corresponding changes in smart-proxy


domcleal:
Merged as 35922c1, thanks @witlessbird.

PR #1065, opened 2016-03-17
Fixes #14247 - systemd timeout increased

lzap:


domcleal:
I'm unsure if I understand why this is desirable from the ticket description.  If the smart proxy startup is blocked on a name lookup _before_ forking, the service may be stuck perpetually in a state where it's not yet forked without a timeout?

foreman-proxy/foreman-proxy.service
```diff
@@ -8,6 +8,7 @@ Type=forking
 User=foreman-proxy
 ExecStart=/usr/share/foreman-proxy/bin/smart-proxy
 PIDFile=/run/foreman-proxy/foreman-proxy.pid
+TimeoutSec=infinity
```
domcleal: This should probably be `TimeoutStartSec` only, since the bug probably doesn't affect shutdown - disabling the timeout feature on shutdown will prevent systemd killing the service more abruptly when it fails.

domcleal: Also the man page states this should be `0` rather than `infinity`.

lzap: Okay. It was likely removed from documentation: https://www.freedesktop.org/software/systemd/man/systemd.service.html

lzap:
Yes, proxy simply delays for some time without any request made, I have only seen this on a system with wrong DNS resolver setup. I can't tell what causes this, perhaps OpenSSL.

Amended the change as you recommended.

lzap:
Looks like I havent commented your concern about pre-fork stuck - nope, it is responding just fine. It simply needs more time to come up.

ares:
I've tested the patch, works fine and fixes the issue.

domcleal:
> Looks like I havent commented your concern about pre-fork stuck - nope, it is responding just fine. It simply needs more time to come up.

Do you mean that it does eventually fork, or that it's started the WEBrick threads and is timing out between that and forking?  If the former, the timeout shouldn't be infinite, while if the latter then the issue should be filed in the smart proxy project.

lzap:
During the startup delay, it refuses connections. Therefore the former I believe.

```
[root@fdi ~]# time systemctl restart foreman-proxy

real    0m40.554s
user    0m0.003s
sys     0m0.008s
[root@fdi ~]# time ping www.google.com
ping: unknown host www.google.com

real    0m20.005s
user    0m0.002s
sys     0m0.001s
[root@fdi ~]# curl -sk http://localhost:8448/features
["bmc","discovery","discovery_image","facts"]
```

domcleal:
Would you mind picking a higher, but non-infinite timeout then?  Just in case there's a different bug, this'll at least leave the service in a failed state if it times out after say, five minutes.

The commented value of DefaultTimeoutStartSec is apparently 90 seconds on EL7 in /etc/systemd/system.conf, but perhaps that's not the actual value if the time taken above is around 40 seconds.

A few minutes instead?

lzap:
Sure thing, let's say 5 minutes?

domcleal:
OK, that's fine.  Out of curiosity, do you know what the default timeout was?

lzap:
Holy cr*p this is what I set on the discovery image: `DefaultTimeoutStartSec=30s`... sorry.

foreman-proxy/foreman-proxy.service
domcleal: It looks like 0 was changed to infinity in version 229 (the latest), so I guess 0 might be safest for now: http://article.gmane.org/gmane.comp.sysutils.systemd.devel/35665

domcleal:
Heh, that's alright.  Glad we dug into it.

lzap:
Thanks for help. For the record, this fixes it: https://github.com/theforeman/foreman-discovery-image/pull/69

PR #1066, opened 2016-03-17
Fixes #14247 - systemd timeout increased

lzap:
RPM counterpart https://github.com/theforeman/foreman-packaging/pull/1065


domcleal:
The branch names appear to be backwards, this has an `rpm/develop-` prefix, which is why the build's failing.  Probably not worth changing it, one won't break the build.

PR #1069, opened 2016-03-18
fixes #12912 - switch to CentOS SCLo SIG builds [rpm]

domcleal:
Swaps dependencies for softwarecollections.org with the CentOS SCLo SIG
release packages, which are copied into our repo by the mash script when
listed in extras/ so non-CentOS users and those without CentOS Extras
enabled can install it.

The packages are better maintained and will allow us to depend on sclo-*
packages in future.  Maintenance of softwarecollections.org builds seems
to have stopped in favour of the SCLo SIG.

---

Koji's already switched over to using these rebuilds when I was doing the ror41 rebuild as it was easier to mirror the one repo.


ekohl:
I assume we need a similar change in puppet-foreman then.

domcleal:
I don't think so, we changed puppet-foreman to install this foreman-release-scl metapackage.

ekohl:
You're right, we already prepared ourselves for this.

PR #1072, opened 2016-03-23
Fixes #14328 - foreman_openscap should require smart_proxy_openscap

sean797:


domcleal:
Is there a hard dependency in the Foreman plugin for this?  Usually there isn't, because Foreman can be installed on a different host to the smart proxy.

sean797:
The reason I suggest this is everyone will have a integrated proxy on their Foreman, So in order to upload to that they would have to install smart_proxy_openscap.

I guess its not much of a pain to install smart_proxy_openscap separately but it seem unnecessary?

ares:
I don't think we should enforce it as a dependency, there are people that do not run smart proxy on Foreman host. It's likely most do but it should be fairly easy to setup smart proxy plugin with our installer, you just use one extra parameter.

PR #1073, opened 2016-03-29
Organizations and locations should be enabled for apipie docs

ares:
I know we have taxonomies disabled by default but it might be better to generate apipie docs including taxonomies in API (note that hammer works regardless, since we generate index during installer run)

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -508,6 +508,9 @@ make -C locale all-mo
 mv Gemfile Gemfile.in
 cp config/database.yml.example config/database.yml
 cp config/settings.yaml.example config/settings.yaml
+#we need to allow taxonomies so apipie cache renders documentation with them
+sed -i 's/:locations_enabled: false/:locations_enabled: true/' config/settings.yaml
+sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml
 export BUNDLER_EXT_NOSTRICT=1
 export BUNDLER_EXT_GROUPS="default assets"
 %{scl_rake} assets:precompile RAILS_ENV=production --trace


domcleal:
Would this affect the behaviour of say `host#create` if organisations were disabled?  The `organization_id` parameter could now be marked as required, and Hammer would enforce it.

ares:
@mbacovsky correct me if I'm wrong please, but I think Hammer would use the result of `rake apipie:cache:index` which would be generated during installer run respecting whether orgs are allowed or disabled. Therefore this should have only impact on what you see at /apidoc

domcleal:
Sorry, I was thinking of the individual organisation-related parameters (i.e. `organization_id`) to non-organisation resources (i.e. `hosts`), not the organisation resource itself.  I agree that the latter will be generated today and included in the index from the installer or any `apipie:cache:index` run if orgs are later enabled.

The problem that might come up if this is enabled is when we have lines like this: https://github.com/theforeman/foreman/blob/1.11.0-RC3/app/controllers/api/v2/hosts_controller.rb#L47

This will add a required parameter into the hosts resource in all installations, whether or not orgs are enabled at runtime, I think.

ares:
I think these should be also regenerated by `apipie:cache:index`, which would remove these parameters from index. They would only remain in `/apidoc`, marked as required (which I agree is confusing). Or maybe I'm missing what you mean by "hosts resource".

domcleal:
Oh, I thought apipie:cache:index would only regenerate the list of resources (domains, hosts, organisations, subnets etc.) and not change the list of parameters on the hosts resource page.

> They would only remain in /apidoc, marked as required (which I agree is confusing).

It's not the confusion that I'm worried about, it's that apipie-bindings may enforce the requirement for the parameter.

ares:
I think we have apipie-bindings validations disabled and hammer loads it from the index. I can do some tests to verify this, according to @mbacovsky it seems that it should work.

domcleal:
Ah yes, looking at the default in hammer_cli I think you're right.  If you could give it a test, that'd be appreciated, else I'll do so soon before merging.

ares:
Tests were successful. After I disabled organizations, ran `apipie:cache:index`, hammer does not list the option (after reloading cache `-r`) in `hammer host create -h`. 

When trying to create one
```
be bin/hammer host create --name a --managed false
Could not create the host:
  Error: Could not find location, please set one of options --location, --location-id.
```

and a success story (last param is to avoid primary interface bug on older verion)
```
be bin/hammer host create --name a --managed false --location Brno --interface 'primary=true'
Host created
```

domcleal:
Thanks for checking that.  Could you send the same to deb/develop please?

Merged, thanks @ares.

ares:
Thank you, PR opened.

PR #1078, opened 2016-04-06
Update foreman_bootdisk to 7.0.0

domcleal:
For both deb/develop and deb/1.11, please.


mmoll:
seems to work just fine here, also cherry-picked into 1.11 as 7552bfd, thanks @domcleal!

PR #1079, opened 2016-04-06
fixes #14483 - package and use bin/rails to launch standalone server

domcleal:


mmoll:
looking good, thanks @domcleal!

PR #1083, opened 2016-04-10
plugins:foreman-cockpit - Release 2.0.0 (DEB)

dLobatog:


mmoll:
Which Foreman versions is this compatible with?

dLobatog:
I updated this to use 2.0.1 instead, as I fixed one thing recently. This is compatible for everything 1.7+ so whatever we maintain between then and now :smile:

mmoll:
cherry-picked:

into | as
----|----
deb/develop | 0027362
deb/1.11 | 1b56e0d
deb/1.10 | 95de25f
deb/1.9 | 200c98f

building now, gracias @dLobatog!

plugins/ruby-foreman-cockpit/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-foreman-cockpit (2.0.1-1) stable; urgency=low
+
+  * Update release to 2.0.1
+
+ -- Daniel Lobato Garcia <elobatocs@gmail.com>  Sun, Apr 10 2016 12:50:53 +0200
```
domcleal: Note that this date format's invalid and causes build failures. @mmoll has kindly fixed it in be0040a, but be aware for the future.  Use the standard deb tooling or scripts/* please.

PR #1086, opened 2016-04-11
Update kafo_parsers to 0.1.0

ares:

PR #1087, opened 2016-04-11
Update kafo_parsers to 0.1.0

ares:


ares:
deb mirror of #1086

dependencies/jessie/kafo_parsers/control
```diff
@@ -11,6 +11,7 @@ Package: ruby-kafo-parsers
 Architecture: all
 XB-Ruby-Versions: ${ruby:Versions}
 Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter,
-         puppet (<< 4.0.0)
+         ruby-json,
```
domcleal: This shouldn't be needed with Ruby 1.9 or higher as JSON's built into Ruby, so all apart from precise can probably have it removed.

dependencies/jessie/kafo_parsers/control
```diff
@@ -11,6 +11,7 @@ Package: ruby-kafo-parsers
...
+         ruby-rdoc
```
domcleal: Probably the same reason this is missing, it's in Ruby 1.9.

ares:
right, I removed rdoc and json deps from jessie control file

domcleal:
Sorry, I hadn't meant to imply that it was only applicable to jessie.  The same comment should apply to wheezy and trusty too, since they're all 1.9 or higher platforms.  Only precise has a default Ruby version of 1.8.

ares:
ok, fixed there as well, for some reason, they had rdoc before

domcleal:
ruby-json looks correct now, it's now only on precise.

> ok, fixed there as well, for some reason, they had rdoc before

It looks like we ship a ruby-rdoc for precise, trusty and wheezy, so perhaps the version in the base Ruby package is too old.  A quick check of wheezy suggests the default version is 3.9.4 and 4.1.0 on jessie, while the gemspec requires >= 3.9.0 and the package requires >= 3.12.0.

So sorry, I think rdoc should probably stay unchanged from deb/develop - explicit dependencies on ruby-rdoc on precise, trusty and wheezy - but don't add it to jessie.  ruby-json should stay on precise only.

ares:
I hope I have it right now :-)

mmoll:
Works just fine for me on Debian/jessie, cherry-picked as 89dea28 into develop, building now. Díky @ares!

PR #1094, opened 2016-04-15
Update docker-api to 1.28

dLobatog:


dLobatog:
I see some files need to be handled by the spec, fixing...

domcleal:
The archive-tar-minitar dependency is now gone, it should be removed from the spec and as a second step, removed from the repository as it's otherwise completely unused:

1. delete the whole directory
2. remove from comps/
3. remove from rel-eng/tito.props
4. don't delete the rel-eng/packages/ file
5. add to tfm/tfm.spec list of obsoletions
6. remove from tfm/scl_prefix_update.conf

dLobatog:
@domcleal Thanks, good catch - I think the latest update contains all these fixes

tfm/tfm.spec
```diff
@@ -65,6 +65,7 @@ Obsoletes: %{scl_prefix}rubygem-sprockets < 2.12.3-3
 Obsoletes: %{scl_prefix}rubygem-less < 2.5.1-5
 Obsoletes: %{scl_prefix}rubygem-less-rails < 2.5.0-4
 Obsoletes: %{scl_prefix}rubygem-uuidtools < 2.1.3-6
+Obsoletes: %{scl_prefix}rubygem-archive-tar-minitar < 0.5.2-11
```
domcleal: This should be `-12`, as the current release is 11.

domcleal: Also, please keep the list alphabetical.

domcleal:
Merged and will build shortly, thanks @dLobatog.

PR #1098, opened 2016-04-15
plugins:foreman_digitalocean - 1.0.0

dLobatog:
This release needs Fog 1.36 or more to work, so it will not work with
anything less than 1.11


domcleal:
Merged and building for nightly and 1.11, thanks @dLobatog.

PR #1100, opened 2016-04-18
Update kafo to 0.7.3

ares:
rake < 11 was pinned as a development dependency which should not have any impact on the spec I think


domcleal:

PR #1102, opened 2016-04-18
bootdisk: precompile assets to fix missing JS [deb]

domcleal:
For deb/1.11 as well please, but where it will be 7.0.0-2 (I can send a second PR if it's easier).


mmoll:
Also cherry-picked as b1c58e1 into 1.11, building now. Thanks @domcleal!

PR #1103, opened 2016-04-18
foreman_bootdisk: include precompiled assets in rpm

timogoebel:
@domcleal : As discussed on IRC, foreman_bootdisk rpms do not include the precompiled assets. This is the fix for rpm/develop.


domcleal:
The test might confirm, but you'll need to also add `-s` to the end of the `%foreman_precompile_plugin` line.  The -a compiles API docs, and -s compiles assets.

timogoebel:
@domcleal : Fixed. I _almost_ figured that out myself. ;-)

domcleal:
Thanks @timogoebel!  I'll build it now for nightly and 1.11.

PR #1104, opened 2016-04-18
Add kafo to installer build deps to generate man page [deb]

domcleal:
With kafo-export-params available, the installer rake build will
generate an extended man page containing a list of parameters.


mmoll:
Looking good, cherry-picked as 356cabe, thanks @domcleal!

PR #1107, opened 2016-04-23
fixes #7087 - install bash completion file (DEB)

mmoll:
would also be cherry-pickable to 1.11


domcleal:
Merged and building, thanks @mmoll!

PR #1108, opened 2016-04-23
fixes #5539 - remove Gemfile.lock (DEB)

mmoll:


mmoll:
Tests fail because I forgot to add "develop" to the branch name, however, I think this can be reviewed without these test packages having been built.

domcleal:
Merged as d32d6d7, thanks @mmoll!

PR #1109, opened 2016-04-23
use foreman-ruby also for foreman-sqlite3 (DEB)

mmoll:
also for 1.11, I guess...


domcleal:
Merged as f2764d4, thanks @mmoll!

PR #1113, opened 2016-04-27
fixes #14848 - specify localstatedir to use /var [deb]

domcleal:


mmoll:
merged, thanks @domcleal!

PR #1114, opened 2016-04-27
Release foreman_remote_execution 0.3.2 (RPM)

stbenjam:


domcleal:
It looks like 0.3.1 isn't compatible with Rails 4.2, so I'm not sure this is meant to be targeted at rpm/develop.  Should it be on rpm/1.11?

stbenjam:
0.3.2 should work on both

stbenjam:
Good to go for 1.11 and nightly

domcleal:
Thanks @stbenjam, merged and building for both now.

PR #1116, opened 2016-04-28
Add `EnvironmentFile` to the service unit file of foreman-proxy.

tdevelioglu:
The old init script used to source /etc/sysconfig/foreman-proxy for
environment variables and this behavior wasn't carried over to the
systemd unit file.

This makes it possible to set custom environment variables via
`/etc/sysconfig/foreman-proxy` again.


domcleal:
Thanks @tdevelioglu!

mmoll:
@tdevelioglu I know it's asked a bit much, but if you have the time, could you have a look into http://projects.theforeman.org/issues/12463 (comments also in #894)?

PR #1119, opened 2016-04-28
fixes #14734 - use foreman-ruby in all bin/* scripts [deb]

domcleal:


mmoll:
merged, thanks @domcleal!

PR #1121, opened 2016-04-29
Copy dependencies files over for Ubuntu/xenial

mmoll:


domcleal:
Merged, thanks @mmoll!  (Not built.)

PR #1127, opened 2016-05-06
refs #14941 - load Kafo lib from relative path [deb]

domcleal:
Updates the patch to remove the assumption that Puppet and Kafo are in
the same Ruby environment.


mmoll:
please rebase :)

domcleal:
Done!

mmoll:
working fine here, merged, thanks @domcleal!

PR #1128, opened 2016-05-06
refs #14453 - require kafo 0.8.0 for parser cache [deb]

domcleal:
Dependent on https://github.com/theforeman/foreman-installer/pull/172. This means puppet-strings isn't required at runtime when using AIO, and it's faster than parsing on the fly.


mmoll:
merged, thanks @domcleal!
this is just bumping the kafo version requirement

PR #1129, opened 2016-05-06
fixes #14454 - permit and prefer puppet-agent (AIO) [deb]

domcleal:
1. Allows and prefers puppet-agent for installation.
1. apipie-bindings is no longer a dependency as the appropriate package is installed at runtime (this was here for some older 2.7 support as it couldn't load on the fly, but is now unsupported).

Depends on:

1. ~~#1128~~
1. ~~#1127~~
1. ~~Kafo release with https://github.com/theforeman/kafo/pull/150~~


mmoll:
[test] again

mmoll:
One test is failing, because on xenial with the repos used, kafo-parsers is not new enough. The rest is working as expected.

mmoll:
merged, thanks @domcleal!

PR #1132, opened 2016-05-06
Use zpassenger.load for passenger

xprazak2:
This is another attempt to fix [#14687](http://projects.theforeman.org/issues/14687). I do not think any changes regarding this problem will be accepted into puppetlabs-apache, since the module works as intended. The issue is rather a result of interaction between said module and our package.

puppetlabs-apache introduced `zpassenger.load` file that loads passenger module with lower priority as described [here](https://github.com/puppetlabs/puppetlabs-apache/pull/1072). Since we want our package to work when not managed by puppet, I moved the loading of a module from `passenger.conf` to `zpassenger.load`. This should remove the duplicated loading when managed by puppet and will work on its own as well.

### DIFF
--- a/rubygem-passenger/rubygem-passenger.spec
+++ b/rubygem-passenger/rubygem-passenger.spec
@@ -46,6 +46,7 @@ Source1: passenger.logrotate
 Source2: rubygem-passenger.tmpfiles
 Source10: apache-passenger.conf.in
 Source11: locations.ini
+Source12: zpassenger.load.in
 
 # Include sys/types.h for GCC 4.7
 Patch2:         rubygem-passenger-4.0.18-gcc47-include-sys_types.patch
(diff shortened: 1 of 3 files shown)


rubygem-passenger/rubygem-passenger.spec
```diff
@@ -315,15 +316,10 @@ install -pm 0755 buildout/apache2/mod_passenger.so %{buildroot}/%{_httpd_moddir}
...
-%{__sed} -n /^LoadModule/p passenger.conf > 10-passenger.conf
-%{__sed} -i /^LoadModule/d passenger.conf
-touch -r %{SOURCE10} 10-passenger.conf
-install -pm 0644 10-passenger.conf %{buildroot}%{_httpd_modconfdir}/passenger.conf
-%endif
 touch -r %{SOURCE10} passenger.conf
 install -pm 0644 passenger.conf %{buildroot}%{_httpd_confdir}/passenger.conf
-
+touch -r %{SOURCE12} zpassenger.load
+install -pm 0644 zpassenger.load %{buildroot}%{_httpd_modconfdir}/zpassenger.load
```
domcleal: Why install an empty file here, shouldn't it be the SOURCE12?

xprazak2:
It definitely should be SOURCE12, it is not empty now.

rubygem-passenger/rubygem-passenger.spec
```diff
@@ -315,15 +316,11 @@ install -pm 0755 buildout/apache2/mod_passenger.so %{buildroot}/%{_httpd_moddir}
...
-%{__sed} -n /^LoadModule/p passenger.conf > 10-passenger.conf
-%{__sed} -i /^LoadModule/d passenger.conf
-touch -r %{SOURCE10} 10-passenger.conf
-install -pm 0644 10-passenger.conf %{buildroot}%{_httpd_modconfdir}/passenger.conf
-%endif
 touch -r %{SOURCE10} passenger.conf
 install -pm 0644 passenger.conf %{buildroot}%{_httpd_confdir}/passenger.conf
-
+touch -r %{SOURCE12} zpassenger.load
+cat %{SOURCE12} > zpassenger.load
```
domcleal: just install %{SOURCE12} below, I don't see why this extra copy's needed?

xprazak2:
I finally got a green build on Jenkins, could you pls take another look @domcleal ?

xprazak2:
Thanks @domcleal, closing.

PR #1133, opened 2016-05-06
Update kafo to 0.8.1

ares:


domcleal:
Could you also delete the patch fix_require_in_add_progress.diff?  I think it's obsolete with the inclusion of #14941.

ares:
I removed the patch from all versions

domcleal:
Thanks, though I think you also need to remove it from dependencies/*/kafo/patches/series or the build may fail.

ares:
Right, here it is.

mmoll:
working fine here on jessie, merged, díky @ares!

PR #1135, opened 2016-05-09
Remove Debian Wheezy, Ubuntu Precise

domcleal:
Wheezy is now in the LTS phase and superseded by Jessie, while Precise
is superseded by both Trusty and Xenial. This will help drop Ruby 1.9.3
support in Foreman core and allow some dependencies to be updated.


mmoll:
I concur, thanks @domcleal!

PR #1138, opened 2016-05-11
templates: 2.1.0 release

GregSutcliffe:
It's a trivial addition, so while untested it should be fine for 1.11 and 1.10 too.


mmoll:
merged and cherry-pick into 1.11 as 3b3d684, thanks @GregSutcliffe!

PR #1139, opened 2016-05-11
Update foreman_templates to 2.1.0

GregSutcliffe:
I *think* I've done this right - it's the first time I've gotten the rpm side of the repo to work on Arch ;)

As with #1138 should be good for all recent Foreman releases.


domcleal:
Thanks @GregSutcliffe, building for nightly and 1.11.

PR #1144, opened 2016-05-18
Add fast_gettext back to plugin repos

lzap:
This reverts commit d321020ac74551d32053fca6fc5c3e5ea8bcb1d2.

This gem is used in Foreman Discovery Image:
https://github.com/theforeman/foreman-discovery-image/pull/71 I mean it wasn't yet used but I would like to use this gem.

### DIFF
--- a/rel-eng/tito.props
+++ b/rel-eng/tito.props
@@ -536,6 +536,7 @@ whitelist = rubygem-algebrick
   rubygem-concurrent-ruby
   rubygem-concurrent-ruby-edge
   rubygem-dynflow
+  rubygem-fast_gettext
   rubygem-foreman_scap_client
   rubygem-logify
   rubygem-mysql2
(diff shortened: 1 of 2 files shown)


domcleal:
Could you add it to the plugins repo only please?  Is it also needed for both EL6 and 7 or just one?

lzap:
Absolutely, just EL7 (and Fedora).

lzap:
So I guess I only need to whitelist it in `foreman-plugins-nightly-rhel7` and fedora and then add the lines to according plugins XML files, do I?

lzap:
Oh wait it should be `foreman-plugins-nightly-nonscl-rhel7` I guess.

domcleal:
Correct.

lzap:
Done, I hope our transition to COPR also clears this a bit. Not easy to maintain.

rel-eng/tito.props
```diff
@@ -570,6 +571,7 @@ whitelist = rubygem-algebrick
   rubygem-dynflow
   rubygem-newt
   rubygem-openscap
+  rubygem-fast_gettext
```
domcleal: No need for this in Fedora, it's already built in the main repo.

lzap: Removed.

domcleal:
Thanks @lzap, merged and building now.

PR #1150, opened 2016-06-02
Packaging new releases of REX smart proxy plugins

stbenjam:
In the newest releases of our smart proxy plugins, we're supporting RHEL 6.  This accomplished by breaking out the dynflow parts into a separate service called smart_proxy_dynflow_core that runs inside of SCL and uses the smart_proxy_remote_execution_ssh_core.

Very much a WIP.


stbenjam:
Untested so far, once I know these are good I'll open PR's against 1.12 and 1.11 (if we need it, I think we do).

stbenjam:
[test]

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -15,8 +15,6 @@ Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 
 Requires: ruby(rubygems)
 Requires: foreman-proxy >= 1.9.0
-Requires: rubygem(dynflow) >= 0.8.6
-Requires: rubygem(dynflow) < 0.9.0
 Requires: rubygem(sqlite3)
 Requires: rubygem(sequel)
```
adamruzicka: If I recall correctly `smart_proxy_dynflow` shouldn't need `sqlite3` nor `sequel`

rubygem-smart_proxy_remote_execution_ssh_core/rubygem-smart_proxy_remote_execution_ssh_core.spec
```diff
@@ -0,0 +1,76 @@
...
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.1.1
+Release: 1%{?foremandist}%{?dist}
+Group: Development/Libraries
+License: GPLv3
+URL: http://github.com/theforeman/smart_proxy_remote_execution_ssh
+Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
+Requires: foreman-proxy >= 1.11.0
+
+Requires: %{?scl_prefix}rubygem(smart_proxy_dynflow_core) = 0.1.1
```
adamruzicka: Missing dependencies on `net-ssh <= 2.9.4` and `net-scp`

rubygem-smart_proxy_remote_execution_ssh_core/rubygem-smart_proxy_remote_execution_ssh_core.spec
```diff
@@ -0,0 +1,76 @@
...
+
+Summary: Smart Proxy Remote Execution SSH Core
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.1.1
+Release: 1%{?foremandist}%{?dist}
+Group: Development/Libraries
+License: GPLv3
+URL: http://github.com/theforeman/smart_proxy_remote_execution_ssh
+Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
+Requires: foreman-proxy >= 1.11.0
```
adamruzicka: Ditto

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,139 @@
...
+
+Summary: Core Smart Proxy Dynflow Service
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.1.1
+Release: 1%{?foremandist}%{?dist}
+Group: Development/Libraries
+License: GPLv3
+URL: http://github.com/theforeman/smart_proxy_dynflow
+Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
+Requires: foreman-proxy >= 1.11.0
```
stbenjam: It does indeed, the service runs as the foreman-proxy user.

adamruzicka: Right, sorry, I missed that

rubygem-smart_proxy_remote_execution_ssh_core/rubygem-smart_proxy_remote_execution_ssh_core.spec
```diff
@@ -0,0 +1,77 @@
...
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.1.1
+Release: 1%{?foremandist}%{?dist}
+Group: Development/Libraries
+License: GPLv3
+URL: http://github.com/theforeman/smart_proxy_remote_execution_ssh
+Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
+
+Requires: %{?scl_prefix}rubygem(net-scp)
+Requires: %{?scl_prefix}rubygem(net-ssh) <= 2.9.4
```
stbenjam: @adamruzicka Why do we need <= 2.9.4? We don't have this available and conflicts with foreman's version (3.0.x) - I think we need to be on the same release for this to work right

stbenjam: Ah I see, because we used to require older ruby.  Newer net-ssh should be ok in SCL: https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/25

adamruzicka:
@stbenjam yeah, it was because of older rubies in case the core would run as part of the proxy

rubygem-smart_proxy_remote_execution_ssh_core/rubygem-smart_proxy_remote_execution_ssh_core.spec
adamruzicka: @stbenjam yeah, it was because of older rubies in case the core would run as part of the proxy

stbenjam:
Packages are green, and services start on EL7.  I'll begin testing el6 this afternoon, to make sure its init files and such work.  @adamruzicka I seem to be having an issue where I get no SSH feature on the proxy, do you know why? It's enabled in /etc/foreman-proxy/settings.d, I only have a Dynflow service.

I'd greatly appreciate additional reviews of the spec files, this packaging is a bit complicated, especially your feedback @domcleal.

This PR updates smart_proxy_dynflow, and smart_proxy_remote_execution_ssh and adds respective _core parts which run inside SCL.

adamruzicka:
@stbenjam I don't recall ever seeing that, I'll look into it tomorrow morning

adamruzicka:
@stbenjam After tweaking some config files to fit into my setup it works for me on el7.

Also shouldn't `smart_proxy_dynflow_core` rpm provide a path in `/etc` where its config should reside along with either sane defaults or an example file?

> I seem to be having an issue where I get no SSH feature on the proxy

Does the `~foreman-proxy/.ssh/id_rsa_foreman_proxy{,.pub}` exist on your machine? If not, it might be the reason why you're missing the ssh feature on the proxy. Debug log from proxy might help to pinpoint the issue more accurately.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,142 @@
...
+Summary:    Documentation for rubygem-%{gem_name}
+
+%description doc
+This package contains documentation for rubygem-%{gem_name}.
+
+%prep
+%setup -n %{pkg_name}-%{version} -q -c -T
+mkdir -p .%{gem_dir}
+%{?scl:scl enable %{scl} "}
+gem install --local --install-dir .%{gem_dir} \
```
domcleal: prefer `%{gem_install}`

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,142 @@
...
+            --force %{SOURCE0} --no-rdoc --no-ri
+%{?scl:"}
+
+%build
+sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X/usr/bin/%{?scl:%{scl_prefix}}rubyX' %{_builddir}/%{pkg_name}-%{version}/%{gem_instdir}/bin/%{service_name}
```
domcleal: You should already be in the builddir here, so `./%{gem_instdir}`

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,142 @@
...
+
+# switches to bundler_ext instead of bundler and uses bundler.d from /etc/foreman-proxy/dynflow/bundler.d
+mv %{_builddir}/%{pkg_name}-%{version}/%{gem_instdir}/Gemfile %{_builddir}/%{pkg_name}-%{version}/%{gem_instdir}/Gemfile.in
+sed -ri 'sX\#\{File\.dirname\(__FILE__\)\}/bundler\.d/\*\.rbX%{_sysconfdir}/foreman-proxy/dynflow/bundler.d/*.rbX' %{_builddir}/%{pkg_name}-%{version}/%{gem_instdir}/Gemfile.in
+
+%install
+mkdir -p %{buildroot}%{_sysconfdir}/foreman-proxy/dynflow/bundler.d
```
domcleal: Should be in datadir like other bundler.d files?

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,142 @@
...
+        %{buildroot}%{gem_dir}/
+
+#copy init scripts and sysconfigs
+%if 0%{?rhel} == 6
+install -Dp -m0755 %{buildroot}%{gem_instdir}/deploy/%{service_name}.init %{buildroot}%{_root_initddir}/%{service_name}
+%else
+install -Dp -m0644 %{buildroot}%{gem_instdir}/deploy/%{service_name}.service %{buildroot}%{_unitdir}/%{service_name}.service
+%endif
+mkdir -p %{buildroot}%{_scl_root}%{bin_dir}
+ln -sv %{gem_instdir}/bin/%{service_name} %{buildroot}%{_scl_root}%{bin_dir}/%{service_name}
```
domcleal: This should be a binstub, but if you use the gem_install macro then it should install correctly.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,142 @@
...
+
+%files
+%dir %{gem_instdir}
+%dir %{_sysconfdir}/foreman-proxy/dynflow/bundler.d
+%exclude %{gem_instdir}/.*
+%{gem_libdir}
+%{gem_instdir}/bin
+%{gem_instdir}/config
+%{gem_instdir}/Gemfile.in
+%{gem_instdir}/smart_proxy_dynflow_core.gemspec
```
domcleal: -doc

rubygem-smart_proxy_remote_execution_ssh_core/rubygem-smart_proxy_remote_execution_ssh_core.spec
```diff
@@ -0,0 +1,77 @@
...
+Summary:    Documentation for rubygem-%{gem_name}
+
+%description doc
+This package contains documentation for rubygem-%{gem_name}.
+
+%prep
+%setup -n %{pkg_name}-%{version} -q -c -T
+mkdir -p .%{gem_dir}
+%{?scl:scl enable %{scl} "}
+gem install --local --install-dir .%{gem_dir} \
```
domcleal: ditto, prefer %gem_install

domcleal:
The update is missing comps and rel-eng updates for fedora24 tags too.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
adamruzicka: I'm pretty sure this is a path in which plugins for the core should drop their files so they get loaded when bundler requires gems when the core starts.

domcleal: Right, so that should probably be in the datadir (like Foreman and the smart proxy), it's not really a user configuration file.

stbenjam: Right, but foreman and smart proxy are a bit different.  `smart_proxy_dynflow_core`'s  directory is versioned, `/opt/theforeman/tfm/root/usr/share/gems/gems/smart_proxy_dynflow_core-0.1.1`, and we don't want to lock remote_execution_ssh_core and dynflow_core versions together, which is the only way I could think of to ensure we know the path.  That's why we decided to use sysconfdir.  Is there a better suggestion?

stbenjam:
>  Also shouldn't smart_proxy_dynflow_core rpm provide a path in /etc where its config should reside along with either sane defaults or an example file?

@adamruzicka Can you clarify what this file should be and where I should put it? Is there an example in the gem?

>>  I seem to be having an issue where I get no SSH feature on the proxy

> Does the ~foreman-proxy/.ssh/id_rsa_foreman_proxy{,.pub} exist on your machine? If not, it might be the reason why you're missing the ssh feature on the proxy. Debug log from proxy might help to pinpoint the issue more accurately.

Yea, I generated SSH keys.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
stbenjam: Would it make sense to use datadir and only store the bunder.d there but keep the ruby code in the gem dir? Foreman and smart proxy seem to get it all in the same place.

adamruzicka:
@stbenjam There's `settings.yml.example` in config directory in `smart_proxy_dynflow_core` gem. The core looks for it in `/etc/smart_proxy_dynflow_core`, `~/.config/smart_proxy_dynflow_core` and in `$gemdir/config`. I'd put it in the `/etc/` one

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
domcleal: > Would it make sense to use datadir and only store the bunder.d there but keep the ruby code in the gem dir?

Yep, I think that's probably best - e.g. %{_datadir}/%{gem_name}

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
stbenjam: What's -doc mean?

domcleal: Sorry. Please move this file to the -doc subpackage, it isn't required at runtime (only %gem_spec is).

stbenjam: Hmm, I thought it was actually needed, I see the gem_spec is included in the package but it gets deployed to:

/opt/theforeman/tfm/root/usr/share/gems/specifications/smart_proxy_dynflow_core-0.1.1.gemspec

And that didn't work for me with the Gemfile specifying `gemspec`, which is why I put it in the root dir.  Maybe I'm mistaken, I'll look again.

domcleal: Oh, if this relies on the Gemfile with `gemspec` to boot the application then yes, you would need the gemspec here. In that case it's bizarre that it's packaged as a rubygem if that's not how it's invoked.  How is this code loaded, only through a service?

stbenjam: > In that case it's bizarre that it's packaged as a rubygem if that's not how it's invoked. How is this code loaded, only through a service?

Correct, it's a Sinatra application like the smart-proxy (which also seems to actually be a gem too?).

The smart_proxy_dynflow is a gem that gets loaded into the proxy, and _core is the service components. It's a little odd, I guess. I think they did it this way so the code could stay in the same repo? @adamruzicka could probably clarify.

stbenjam: Oh sorry, I know why it's a gem.  If you're using a newer ruby, and smart_proxy_dynflow and smart_proxy_dynflow_core are running in the same ruby environment (say on el7 where the ruby is new enough), then you don't actually need to use the separate service.

We're not *currently* using that feature, but could in the future.

stbenjam:
Should be red until I get in https://github.com/theforeman/smart_proxy_dynflow/pull/18

stbenjam:
@domcleal When I removed the explicit ln, and even with executables set in the gemspec, I don't get a stub in the SCL bin dir:

```
    File not found: /builddir/build/BUILDROOT/tfm-rubygem-smart_proxy_dynflow_core-0.1.1-1.fm1_13.git.0.b3cffb2.el7.noarch/opt/theforeman/tfm/root/usr/bin/smart_proxy_dynflow_core
```

Any idea why?

domcleal:
The files under _bindir probably need copying in %files from the build dir to the install root, like it does with %{gem_dir}. Check other spec files that include binaries or the gem2rpm templates.

stbenjam:
Thanks, got it.  Packaging looks better, ready for another look.  Please don't merge until we solve http://projects.theforeman.org/issues/15424.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,148 @@
...
+
+%build
+sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X/usr/bin/%{?scl:%{scl_prefix}}rubyX' ./%{gem_instdir}/bin/%{service_name}
+
+# switches to bundler_ext instead of bundler
+mv %{_builddir}/%{pkg_name}-%{version}/%{gem_instdir}/Gemfile ./%{gem_instdir}/Gemfile.in
+sed -ri 'sX\#\{File\.dirname\(__FILE__\)\}/bundler\.d/\*\.rbX%{_datadir}/%{gem_name}/bundler.d/*.rbX' ./%{gem_instdir}/Gemfile.in
+
+# switch to SCL directory for config
+sed -ri 'sX/etcX%{_sysconfdir}X' ./%{gem_instdir}/lib/%{gem_name}/*.rb
```
iNecas: I wonder if we could place the configs to some etc directory such as `/etc/foreman-proxy/smart_proxy_dynflow_core` instead of `/opt/theforeman/tfm/root/etc/smart_proxy_dynflow_core`.

iNecas: It was a bit surprising to find the config files somewhere else

stbenjam: Yea, I meant to ask about that. The `%{_sysconfdir}` macro is set for the SCL root.  I can hardcode `/etc` I guess.

stbenjam: @iNecas Do you have any thoughts about the binary location as well?

It's in /opt/theforeman/tfm/root/usr/bin/smart_proxy_dynflow_core.

I was mostly following http://developers.redhat.com/blog/2014/08/26/how-package-initscripts-unit-files-software-collections/, which does keep SCL packages strictly confined to the scl root.

Do you have opinions on this @domcleal?

iNecas: I don't have preferrence about the executable. It's just about the configuration file that seems a bit odd located, but we could probably live with this.

iNecas: Anyway, based on the link you've sent, it seems people do it that way.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,148 @@
...
+
+%prep
+%setup -n %{pkg_name}-%{version} -q -c -T
+mkdir -p .%{gem_dir}
+%{?scl:scl enable %{scl} - <<EOF}
+%gem_install -n %{SOURCE0}
+%{?scl:EOF}
+
+%build
+sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X/usr/bin/%{?scl:%{scl_prefix}}rubyX' ./%{gem_instdir}/bin/%{service_name}
```
domcleal: I don't think this is editing the correct file, it should be the binstub as that's the command that's executed (the entrypoint), and it just loads this file in gem_instdir.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,148 @@
...
+mkdir -p %{buildroot}%{_sysconfdir}/smart_proxy_dynflow_core
+cp -pa .%{gem_instdir}/config/settings.yml.example %{buildroot}%{_sysconfdir}/smart_proxy_dynflow_core/settings.yml
+
+#copy init scripts and sysconfigs
+%if 0%{?rhel} == 6
+install -Dp -m0755 %{buildroot}%{gem_instdir}/deploy/%{service_name}.init %{buildroot}%{_root_initddir}/%{service_name}
+%else
+install -Dp -m0644 %{buildroot}%{gem_instdir}/deploy/%{service_name}.service %{buildroot}%{_unitdir}/%{service_name}.service
+%endif
+mkdir -p %{buildroot}%{_scl_root}%{bin_dir}
```
domcleal: unused?  (Then remove bin_dir too, please.)

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,148 @@
...
+mkdir -p .%{gem_dir}
+%{?scl:scl enable %{scl} - <<EOF}
+%gem_install -n %{SOURCE0}
+%{?scl:EOF}
+
+%build
+sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X/usr/bin/%{?scl:%{scl_prefix}}rubyX' ./%{gem_instdir}/bin/%{service_name}
+
+# switches to bundler_ext instead of bundler
+mv %{_builddir}/%{pkg_name}-%{version}/%{gem_instdir}/Gemfile ./%{gem_instdir}/Gemfile.in
```
domcleal: The first argument should also use `./%{gem_instdir}`

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,148 @@
...
+sed -ri 'sX\#\{File\.dirname\(__FILE__\)\}/bundler\.d/\*\.rbX%{_datadir}/%{gem_name}/bundler.d/*.rbX' ./%{gem_instdir}/Gemfile.in
```
domcleal: Use a patch for this as it'll fail nicely when the source code changes, while the sed will fail quietly.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
domcleal: If these are user-editable files, you could put them in `%{root_sysconfdir}` as we do this already with Hammer gem packages. It's not self-contained into the SCL, but I don't think it's your intention to ensure it is self-contained (so you could have multiple versions of this package installed).  It's probably better to use `/etc`.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,148 @@
...
+mkdir -p %{buildroot}%{_bindir}
+cp -pa .%{_bindir}/smart_proxy_dynflow_core %{buildroot}%{_bindir}/smart_proxy_dynflow_core
+mkdir -p %{buildroot}%{_sysconfdir}/smart_proxy_dynflow_core
+cp -pa .%{gem_instdir}/config/settings.yml.example %{buildroot}%{_sysconfdir}/smart_proxy_dynflow_core/settings.yml
+
+#copy init scripts and sysconfigs
+%if 0%{?rhel} == 6
+install -Dp -m0755 %{buildroot}%{gem_instdir}/deploy/%{service_name}.init %{buildroot}%{_root_initddir}/%{service_name}
+%else
+install -Dp -m0644 %{buildroot}%{gem_instdir}/deploy/%{service_name}.service %{buildroot}%{_unitdir}/%{service_name}.service
```
domcleal: This unit file (and also the init script) contains SCL specifics which won't work on Fedora and shouldn't really be in the source.  I'd prefer:

1. Remove the scl enable from the scripts/unit files
2. Move the `smart_proxy_dynflow_core` binstub from the SCL bindir to the root bindir, like we do with Hammer, so it's installed at /usr/bin/smart_proxy_dynflow_core.  (Like the above discussion about configs, this means it's not contained in the SCL.)
3. Fix the comment at https://github.com/theforeman/foreman-packaging/pull/1150/files#r67669806 so the binstub gets the /usr/bin/tfm-ruby shebang line, then the unit file doesn't need to enable the SCL.

or

1. Put separate SCL/non-SCL unit files into either the gem or the packaging repo (preferably, since everything's OS-specific).

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,148 @@
...
+
+# switch to SCL directory for config
+sed -ri 'sX/etcX%{_sysconfdir}X' ./%{gem_instdir}/lib/%{gem_name}/*.rb
+
+%install
+mkdir -p %{buildroot}%{_datadir}/%{gem_name}/bundler.d
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
+mkdir -p %{buildroot}%{_bindir}
+cp -pa .%{_bindir}/smart_proxy_dynflow_core %{buildroot}%{_bindir}/smart_proxy_dynflow_core
```
domcleal: I'd suggest copying `%{_bindir}/*` and leave %files as it is with a specific filename - then you'll get a build error if new binaries are added to the gem but we've not packaged/expected them.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
domcleal: OK, thanks for the explanation, that clears up my confusion about other parts of the spec.

domcleal:
A couple of more general points:

1) The upgrade path looks unclear to me, how will the new *_core packages be installed for users with the existing ones installed?
2) It may be worth removing the non-SCL rubygem-dynflow at the same time, obsoleting it in the upgrade path.

iNecas:
1) the migration means the user needs to install the *_core package and make sure the service is running. I'm not sure we can handle this in the packaging, we have a PR opened in puppet modules to make the migration easier https://github.com/theforeman/puppet-foreman_proxy/pull/258

stbenjam:
Foreman-only users generally don't run the installer, although we can recommend it in the release announcement I guess.  We could also explicitly require the _core packages on the non-SCL gems so at least they get installed? The core settings.yml would still be problem though, as it needs the SSL certs configuration.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
stbenjam: Done, no longer confined to SCL. User-editable config is /etc (eliminating the need for the patch you mentioned), and the binary is in /usr/bin.

iNecas:
Tested and works well. :+1: from me

stbenjam:
Just a small change, `_core` are required by the non-core parts, as we don't support any other case and this addresses the upgrade case.  Users will still need to run the installer, but we'll talk about this in our docs (and it's already mentioned in the foreman docs to do this on an upgrade).

stbenjam:
This should be good to go, for 1.11 + 1.12 + nightly.

rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -0,0 +1,149 @@
...
+%endif
+
+%exclude %{gem_instdir}/deploy
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/LICENSE
+
+%changelog
+* Thu May 26 2016 Stephen Benjamin <stephen@redhat.com> 0.1.1-1
```
domcleal: 0.1.3, or leave it out for tito

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -6,19 +6,21 @@
...
-Requires: rubygem(dynflow) >= 0.8.6
-Requires: rubygem(dynflow) < 0.9.0
-Requires: rubygem(sqlite3)
-Requires: rubygem(sequel)
+Requires: foreman-proxy >= 1.11.0
+
+%if 0%{?fedora}
+Requires: rubygem-smart_proxy_dynflow_core
+%else
+Requires: tfm-rubygem-smart_proxy_dynflow_core
```
domcleal: It might be worth adding `= %{version}` here so you ensure the two packages are in lock-step with each other in case there are internal API changes, and don't get out of sync - especially considering they have different prefixes, it's possible that somebody updates one without the other.

If so, prefer the `rubygem(..)` form as it doesn't include an epoch, i.e.

    Requires: rubygem(smart_proxy_dynflow_core) = %{version}

rubygem-smart_proxy_remote_execution_ssh/rubygem-smart_proxy_remote_execution_ssh.spec
```diff
@@ -6,21 +6,25 @@
...
 URL: https://github.com/theforeman/smart_proxy_remote_execution_ssh
 Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 
-Requires: foreman-proxy >= 1.9.0
-Requires: rubygem(smart_proxy_dynflow) >= 0.0.3
-Requires: rubygem(smart_proxy_dynflow) < 0.1.0
+%if 0%{?fedora}
+Requires: rubygem-smart_proxy_remote_execution_ssh_core
+%else
+Requires: tfm-rubygem-smart_proxy_remote_execution_ssh_core
```
domcleal: ditto, re using `= %{version}`

domcleal:
Could you rebase the PR please?

> This should be good to go, for 1.11 + 1.12 + nightly.

Are you sure you want to push this to 1.11 and 1.12 immediately, both without installer changes and without testing it on nightly?

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
stbenjam: Updated, but how do I use the form of rubygem(..) for the SCL requirement?  Not sure I understand what you mean by not including the epoch/why implication that has.

stbenjam:
>> This should be good to go, for 1.11 + 1.12 + nightly.
> Are you sure you want to push this to 1.11 and 1.12 immediately, both without installer changes and without testing it on nightly?

We've tested it, but I guess let's just put it in nightly for now.

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
```diff
@@ -6,19 +6,21 @@
...
-Requires: rubygem(dynflow) >= 0.8.6
-Requires: rubygem(dynflow) < 0.9.0
-Requires: rubygem(sqlite3)
-Requires: rubygem(sequel)
+Requires: foreman-proxy >= 1.11.0
+
+%if 0%{?fedora}
+Requires: rubygem(smart_proxy_dynflow_core) = %{version}
+%else
+Requires: tfm-rubygem-smart_proxy_dynflow_core = %{version}
```
domcleal: `tfm-rubygem(smart_proxy_dynflow_core)`

rubygem-smart_proxy_remote_execution_ssh/rubygem-smart_proxy_remote_execution_ssh.spec
```diff
@@ -6,21 +6,25 @@
...
 URL: https://github.com/theforeman/smart_proxy_remote_execution_ssh
 Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
 
-Requires: foreman-proxy >= 1.9.0
-Requires: rubygem(smart_proxy_dynflow) >= 0.0.3
-Requires: rubygem(smart_proxy_dynflow) < 0.1.0
+%if 0%{?fedora}
+Requires: rubygem(smart_proxy_remote_execution_ssh_core) = %{version}
+%else
+Requires: tfm-rubygem-smart_proxy_remote_execution_ssh_core = %{version}
```
domcleal: `tfm-rubygem(smart_proxy_remote_execution_ssh_core)`

rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec
domcleal: Just tfm-rubygem(..), I added inline comments.

The epoch thing is just an implementation difference in the version number of the `rubygem(x)` provide versus the actual `rubygem-x` package version - it's more reliable to require the former style than the latter. If a package adds an epoch (i.e. rubygem-concurrent-ruby) then the rubygem(x) provide doesn't include it, so the requirements from other packages still work.

stbenjam: Oh I see, thanks, updated

domcleal:
Merged and building for nightlies, thanks @stbenjam.

PR #1154, opened 2016-06-03
Require Puppet >= 3.1 to match theforeman/puppet [deb]

domcleal:
For deb/1.12 too please.


mmoll:
also cherry-picked as bc903bd for 1.12, thanks @domcleal!

PR #1157, opened 2016-06-06
Link foreman-ruby to specific ruby version [deb]

domcleal:
When building a plugin that has a build-dep on `foreman`, or just
installing the package, the foreman-ruby symlink used in
foreman.postinst will be non-functional as it points to /usr/bin/ruby on
on jessie/xenial. This only exists when the `ruby` package is installed,
but `foreman` depends on a specific `rubyX.Y` package.

---

Should fix build failures such as http://ci.theforeman.org/job/packaging_build_deb_plugin/1095/arch=x86,label=debian,os=jessie/console

<pre>
13:17:44 Setting up ruby2.1 (2.1.5-2+deb8u2) ...
[..]
13:17:44 Setting up ruby2.1-dev:amd64 (2.1.5-2+deb8u2) ...
13:17:44 Setting up foreman (9999-jessie+scratchbuild+201606031407) ...
13:17:44 /var/lib/dpkg/info/foreman.postinst: 48: /var/lib/dpkg/info/foreman.postinst: /usr/bin/foreman-ruby: not found
</pre>

PR #1158, opened 2016-06-06
Use foreman-ruby to run bundler in plugin build scripts [deb]

domcleal:
Fixes builds of plugins on OSes such as trusty where the default Ruby is
different to the foreman-ruby version, causing `foreman` installation
errors as gems are missing the correct `ruby-dev` package.

---

Should fix build errors such as http://ci.theforeman.org/job/packaging_build_deb_plugin/1096/arch=x86,label=debian,os=trusty/console:

<pre>
13:37:05 cd /usr/share/foreman && ( \
13:37:05 		bundle install --local && \
13:37:05 		bundle exec rake plugin:assets:precompile[foreman_azure] RAILS_ENV=production \
13:37:05 	)

13:37:11 Resolving dependencies...
13:37:11 Installing rake (11.1.2) 
13:37:11 Installing ace-rails-ap (4.0.2) 
13:37:12 Installing i18n (0.7.0) 
13:37:12 Installing json (1.8.3) 
13:37:12 Gem::Installer::ExtensionBuildError: ERROR: Failed to build gem native extension.
13:37:12 
13:37:12         /usr/bin/ruby1.9.1 extconf.rb 
13:37:12 /usr/lib/ruby/1.9.1/rubygems/custom_require.rb:36:in `require': cannot load such file -- mkmf (LoadError)
13:37:12 	from /usr/lib/ruby/1.9.1/rubygems/custom_require.rb:36:in `require'
13:37:12 	from extconf.rb:1:in `<main>'
13:37:12 
13:37:12 
13:37:12 Gem files will remain installed in /usr/share/foreman/vendor/ruby/1.9.1/gems/json-1.8.3 for inspection.
</pre>

### DIFF
--- a/plugins/ruby-foreman-azure/debian/rules
+++ b/plugins/ruby-foreman-azure/debian/rules
@@ -15,10 +15,10 @@ build:
 	cp cache/* /usr/share/foreman/vendor/cache/
 	cp $(PLUGIN).rb /usr/share/foreman/bundler.d/
 	cd /usr/share/foreman && ( \
-		bundle install --local && \
-		bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production \
+		/usr/bin/foreman-ruby /usr/bin/bundle install --local && \
+		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production \
 	)
-	GEM_PATH=$$(cd /usr/share/foreman && bundle show $(PLUGIN)) && \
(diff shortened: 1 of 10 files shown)


mmoll:
also for 1.12, I presume?

domcleal:
I think we might need this on all releases where we actively maintain plugins if we switch the plugin builder to trusty (which seems the best OS, as choosing another would require #1157 in Foreman core on every release), so I'd suggest 1.10/1.11 and 1.12 too.

We don't need to rebuild the plugins now, but next time we build a plugin then we'll get this fix included.

mmoll:
merged, thanks @domcleal!

Also cherry-picked:

| as | into |
| ---- | ---- |
| 5b2021d | 1.12 |
| 9d3cc73 | 1.11 |
| 9bb666e | 1.10 |

PR #1162, opened 2016-06-07
Update kafo to 0.9.0

ares:


mmoll:
working fine here :+1: 
Is this also for 1.12?

ares:
Yes, I think it should get to 1.12. @domcleal wanted hiera patch for some 1.12 issues. I know there's at least [this](http://projects.theforeman.org/issues/14789). We might want to keep it in nightly for few days just to be sure. 0.9 adds hiera patch and one minor fix for settings reloading so it should be quite safe though.

domcleal:
Probably best to build for nightly only for now and then we can add 1.12 as/when needed.

mmoll:
merged, díky @ares!

PR #1164, opened 2016-06-08
Added hammer_cli_foreman_admin (RPM)

lzap:
Do we need Debian patch or is this also automatic?


mmoll:
A Debian PR would be needed also, you can use gem2deb on a Debian system and have a look at the other hammer plugins in deb/devlop to see, what to adjust, or just shout here an I'll do it. ;)

lzap:
Oh the package was missing locales, added and amended. Well if you could do the Debian package for me it would be big help @mmoll !

lzap:
Huh forgot to push the gem, [test]

lzap:
Solved few issues with config file and bumped to the final 0.0.4. Tests should be finally green now.

lzap:
Looks like rubygems.org were slow to update. [test]

```
07:37:28 addurl rubygem-hammer_cli_foreman_admin/hammer_cli_foreman_admin-0.0.4.gem 
07:37:29   failed to verify url: http://rubygems.org/gems/hammer_cli_foreman_admin-0.0.4.gem
```

domcleal:
Merged and building for nightlies, thanks @lzap.

PR #1165, opened 2016-06-09
Add gem2rpm templates (RPM)

domcleal:
Includes templates for common rubygem types:

- SCL-only RPMs
- Non-SCL only RPMs
- Foreman plugins
- Smart proxy plugins
- Hammer CLI plugins

This won't work 100% for _everything_, as there are a handful of assumptions about what's in a gem or how plugins are laid out (particularly what config/bundler files are called in hammer/smart proxy gems), but it ought to cover a good 80% of cases.

The templates are using the Fedora gem2rpm default, which use the modern gem unpack/build/install style, but with additions for SCL macros, our config file locations, dependencies and so on.

### DIFF
--- a/gem2rpm/foreman_plugin.spec.erb
+++ b/gem2rpm/foreman_plugin.spec.erb
@@ -1,0 +1,184 @@
+# FIXME:
+#   1. Change scl_prefix to _ruby or _ror for Requires/BuildRequires that are
+#      in Ruby or Rails software collections (not Foreman's)
+#   2. Edit foreman requirement(s) and specify minimum version
+#   3. Delete these lines
+#
+# Generated from <%= package.spec.file_name %> by gem2rpm -*- rpm-spec -*-
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
+%global gem_name <%= spec.name %>
+%global plugin_name <%= spec.name.sub(/\Aforeman_/, '') %>
+<%
+has_assets = spec.files.any? { |f| f.start_with?('app/assets/') }
+has_apidoc = spec.files.any? { |f| f =~ %r{\Aapp/controllers/?.*/api/} }
+has_db_migrations = spec.files.any? { |f| f.start_with?('db/migrate/') }
+has_db_seeds = spec.files.any? { |f| f.start_with?('db/seeds.d/') }
+
+precompile_flags = ''
+precompile_flags << ' -a' if has_apidoc
+precompile_flags << ' -s' if has_assets
+
+config.rules[:doc] << /\/?CHANGES/
+config.rules[:ignore] << /.tx(\/.*)?/
+-%>
+
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: <%= spec.version %>
+Release: 1%{?foremandist}%{?dist}
+Summary: <%= spec.summary.gsub(/\.$/, "") %>
+Group: Applications/Systems
+License: <%= spec.licenses.join(" and ") %>
+<% if spec.homepage -%>
+URL: <%= spec.homepage %>
+<% end -%>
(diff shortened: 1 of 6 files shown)


gem2rpm/foreman_plugin.spec.erb
```diff
@@ -0,0 +1,184 @@
+# FIXME:
+#   1. Change scl_prefix to _ruby or _ror for Requires/BuildRequires that are
```
dLobatog: Nice, I suppose we can go the extra mile after this PR & fetch the ror/ruby packages list so that these are also set automatically. :smile:

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -0,0 +1,184 @@
...
+Requires: foreman >= FIXME
+Requires: %{?scl_prefix_ruby}ruby(release)
+<% for req in spec.required_ruby_version -%>
+Requires: %{?scl_prefix_ruby}ruby <%= req %>
+<% end -%>
+<% for req in spec.required_rubygems_version -%>
+Requires: %{?scl_prefix_ruby}ruby(rubygems) <%= req %>
+<% end -%>
+<% for d in spec.runtime_dependencies -%>
+<% for req in d.requirement -%>
```
dLobatog: Just a nitpick but maybe you can indent the ruby code, and leave the actual spec code unindented? It gets a bit difficult to read IMO

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -0,0 +1,184 @@
...
+
+%build
+# Create the gem as gem install only works on a gem file
+%{?scl:scl enable %{scl} - << \EOF}
+gem build %{gem_name}.gemspec
+%{?scl:EOF}
+
+# %%gem_install compiles any C extensions and installs the gem into ./%%gem_dir
+# by default, so that we can move it into the buildroot in %%install
+%{?scl:scl enable %{scl} - << \EOF}
```
dLobatog: I never understood why we can't put `gem_install` in the same `scl enable` block as `gem build` (it seems to fail on a couple of gems I tried), is it parsed as just 1 long comand with `\n`?

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -0,0 +1,184 @@
...
+%{foreman_db_migrate}
+<% end -%>
+<% if has_db_seeds -%>
+%{foreman_db_seed}
+<% end -%>
+<% if has_apidoc -%>
+%{foreman_apipie_cache}
+<% end -%>
+%{foreman_restart}
+exit 0
```
dLobatog: Why not leaving the exit-code up to `%{foreman_restart}` ?

gem2rpm/foreman_plugin.spec.erb
domcleal: I think that will add indentation in the output of template, which isn't ideal. I'd also prefer to leave it out so it more closely matches the upstream gem2rpm templates that this is copied from, then it's easier to keep in sync.

gem2rpm/foreman_plugin.spec.erb
domcleal: It should work, in that the `scl enable` command will run both commands, but if the first command failed then the RPM build wouldn't fail so it's better to have them run separately. (The RPM scriptlets will run with set -e so any failed command causes a failure, but scl enable won't fail if one command fails.)

gem2rpm/foreman_plugin.spec.erb
domcleal: %foreman_restart will exit with a non-zero code if the Foreman service isn't running or if it's running under Passenger, so we usually need to ignore it.

I do plan on improving the posttrans snippets so they support Passenger, and would probably fix the error handling at the same time.

PR #1166, opened 2016-06-09
refs #14789 - require kafo 0.9.0 [deb]

domcleal:


domcleal:
The PR test fails because kafo 0.9.0 isn't in today's nightly repo, but it should pass from tomorrow (Friday).

ares:
kafo is now in repos [test]

mmoll:
merged, thanks @domcleal!

PR #1167, opened 2016-06-09
Bump rubygem-awesome_print to 1.7.0

akofink:
This is to fix an issue calling awesome_print on a multipart response
string that contains file upload metadata, which causes an IOError due
to awesome_print trying to read data a closed stream. Upgrading the gem
fixes this issue, which was blocking bats tests.


mmoll:
Is this also needed for Debian/Ubuntu?

rubygem-awesome_print/rubygem-awesome_print.spec
```diff
@@ -5,8 +5,8 @@
 
 Summary: Pretty print Ruby objects with proper indentation and colors
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 1.0.2
-Release: 13%{?dist}
+Version: 1.7.0
+Release: 1%{?dist}
 Group: Development/Languages
 License: MIT
 URL: http://github.com/michaeldv/awesome_print
```
domcleal: This has changed and needs updating.

domcleal:
> Is this also needed for Debian/Ubuntu?

Looks like we only ship a standalone package for 14.04 (version 1.2.0), so we could update it. The gem in the Foreman package should be 1.7.0 already. I can't tell from the description which was affected.

mmoll:
Right, I thought about hammer when I asked. If it's not needed for hammer I'd prefer everything to stay like it is, and rather update the package in upstream Debian first.

akofink:
I noticed the issue in the bats tests, which run on EL6 and EL7. I suspect Debian is also affected. I will try to run bats on Debian to find out.

rubygem-awesome_print/rubygem-awesome_print.spec
akofink: @domcleal Please explain what needs updating. (sorry I'm new at packaging)

mmoll: the URL is https://github.com/awesome-print/awesome_print now :mag:

akofink:
@mmoll awesome_print v1.2.0 and v1.6.1 fix the problem on Katello bats as well as the latest version, v1.7.0.

mmoll:
@akofink thanks, then there's no need for action on Debian/Ubuntu :+1:

domcleal:
Merged and building, thanks @akofink.

PR #1172, opened 2016-06-13
Update kafo to 0.7.4 in 1.11

ares:
Katello requires fixes from 0.7.4. I'm a bit confused by the fact that this branch seems to have [kafo 0.7.1](https://github.com/theforeman/foreman-packaging/blob/rpm/1.11/rubygem-kafo/kafo-0.7.1.gem) while there's already [0.7.2](http://yum.theforeman.org/releases/1.11/el7/x86_64/rubygem-kafo-0.7.2-1.el7.noarch.rpm) in our repos.


ares:
thanks for clarification, makes sense

ehelms:
@domcleal when you do suspect you'll release this to the official 1.11 repos? This can lead to some upgrade issues for some cases and we are holding the Katello 3.0 release for it.

domcleal:
On the next point release, which ought to be within the next week (http://projects.theforeman.org/projects/foreman/wiki/Foreman_111_Schedule).

PR #1174, opened 2016-06-14
Add webpack:compile task to build proccess [rpm]

tbrisker:
This is needed for https://github.com/theforeman/foreman/pull/3433 to
build the webpack managed assets correctly for production.

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -192,6 +192,8 @@ BuildRequires: %{?scl_prefix}rubygem(responders) >= 2.0
 BuildRequires: %{?scl_prefix}rubygem(responders) < 3
 BuildRequires: %{?scl_prefix}rubygem(roadie-rails) >= 1.1
 BuildRequires: %{?scl_prefix}rubygem(roadie-rails) < 2
+BuildRequires: %{?scl_prefix}rubygem(webpack-rails) >= 0.9.7
+BuildRequires: %{?scl_prefix}rubygem(webpack-rails) < 1
 # assets
 %if 0%{?scl:1}
 BuildRequires: %{scl}-runtime-assets >= 3
@@ -204,10 +206,6 @@ BuildRequires: %{?scl_prefix_ror}rubygem(sass-rails) < 6
 BuildRequires: %{?scl_prefix_ror}rubygem(uglifier) >= 1.0.3
 BuildRequires: %{?scl_prefix_ror}rubygem(execjs) >= 1.4.0
 BuildRequires: %{?scl_prefix_ror}rubygem(execjs) < 2.5.0
-%if 0%{?scl:1}
-BuildRequires: %{?scl_prefix}rubygem(jquery-rails) >= 3.1
-%endif
-BuildRequires: %{?scl_prefix}rubygem(jquery-rails) < 4.0
 BuildRequires: %{?scl_prefix}rubygem(jquery-ui-rails) < 5.0.0
 BuildRequires: %{?scl_prefix}rubygem(autoprefixer-rails) >= 5.2
 BuildRequires: %{?scl_prefix}rubygem(autoprefixer-rails) < 6.0
@@ -229,8 +227,6 @@ BuildRequires: %{?scl_prefix}rubygem(jquery_pwstrength_bootstrap_4) < 2.0
 BuildRequires: %{?scl_prefix}rubygem(jquery-turbolinks) >= 2.1
 BuildRequires: %{?scl_prefix}rubygem(jquery-turbolinks) < 3.0
 BuildRequires: %{?scl_prefix}rubygem(select2-rails) = 3.5.10
-BuildRequires: %{?scl_prefix}rubygem(underscore-rails) >= 1.8
-BuildRequires: %{?scl_prefix}rubygem(underscore-rails) < 2.0
 %if %precompile_nodejs
 BuildRequires: %{?scl_prefix_nodejs}nodejs
 %else
@@ -412,10 +408,6 @@ Requires: %{?scl_prefix_ror}rubygem(sass-rails) >= 5
 Requires: %{?scl_prefix_ror}rubygem(sass-rails) < 6
 Requires: %{?scl_prefix_ror}rubygem(uglifier) >= 1.0.3
 Requires: %{?scl_prefix_ror}rubygem(execjs) >= 1.4.0
-%if 0%{?scl:1}
-Requires: %{?scl_prefix}rubygem(jquery-rails) >= 3.1
-%endif
-Requires: %{?scl_prefix}rubygem(jquery-rails) < 4.0
 Requires: %{?scl_prefix}rubygem(jquery-ui-rails) < 5.0.0
 Requires: %{?scl_prefix}rubygem(autoprefixer-rails) >= 5.2
 Requires: %{?scl_prefix}rubygem(autoprefixer-rails) < 6.0
@@ -437,8 +429,6 @@ Requires: %{?scl_prefix}rubygem(jquery_pwstrength_bootstrap_4) < 2.0
 Requires: %{?scl_prefix}rubygem(jquery-turbolinks) >= 2.1
 Requires: %{?scl_prefix}rubygem(jquery-turbolinks) < 3.0
 Requires: %{?scl_prefix}rubygem(select2-rails) = 3.5.10
-Requires: %{?scl_prefix}rubygem(underscore-rails) >= 1.8
-Requires: %{?scl_prefix}rubygem(underscore-rails) < 2.0
 %if %precompile_nodejs
 Requires: %{?scl_prefix_nodejs}nodejs
 %else
@@ -570,6 +560,7 @@ sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/se
 export BUNDLER_EXT_NOSTRICT=1
 export BUNDLER_EXT_GROUPS="default assets"
 %{scl_rake} assets:precompile RAILS_ENV=production --trace
+%{scl_rake} webpack:compile --trace
 %{scl_rake} db:migrate RAILS_ENV=production --trace
 %{scl_rake} apipie:cache RAILS_ENV=production cache_part=resources --trace
 rm config/database.yml config/settings.yaml


tbrisker:
This should be merged only when the related PR on Foreman is merged as it will not work before that.

domcleal:
Doesn't this require webpack-rails to be added as a dependency?  Other Gemfile changes from the same PR are also missing.

domcleal:
Also, is the webpack:compile task required for %foreman_plugin_precompile?

tbrisker:
@domcleal I'm quite new to packaging, I hope I correctly updated the dependencies to reflect the changes in the pr. `webpack:compile` is currently not needed for plugins, they will continue using the asset pipeline for now.

foreman/foreman.spec
```diff
@@ -192,6 +192,8 @@ BuildRequires: %{?scl_prefix}rubygem(responders) >= 2.0
 BuildRequires: %{?scl_prefix}rubygem(responders) < 3
 BuildRequires: %{?scl_prefix}rubygem(roadie-rails) >= 1.1
 BuildRequires: %{?scl_prefix}rubygem(roadie-rails) < 2
+BuildRequires: %{?scl_prefix}rubygem(webpack-rails) >= 0.9.7
```
domcleal: This package doesn't exist, you'd need to add it (see README).

domcleal:
This is also missing a way to install the npm packages, they should be added as BuildRequires and to foreman-assets as runtime deps.

For the SCL-enabled Foreman package, the tfm meta-package will probably first need rebuilding against a nodejs SCL too (i.e. nodejs010 or rh-nodejs4 when it's ready), which could be done independently of this PR before nodejs and other packages can be added as deps.

Any nodejs-* packages that aren't in Fedora or the nodejs SCL could then be added to this repo too.

If you're unfamiliar with RPM packaging and SCL technologies then I'm unsure you want to be doing this.

tbrisker:
@domcleal I am indeed unfamiliar with RPM packaging. I was hoping to get your input and assistance in order to do this correctly, as you are quite the expert on the matter. 
Adding the node modules to packaging is not the path we are currently looking to go down, we have had some discussion and background about this in https://github.com/theforeman/rfcs/pull/3, in case you missed my ping in the mail noise. 
The way we are looking into implementing this is similar to how cockpit does it, by adding the node_modules directory into the source tarball, see @dLobatog's comment on this topic: https://github.com/theforeman/foreman/pull/3433#discussion_r67108713

domcleal:
The asset build dependencies should be provided via BRs to the package build as they are today with every other dep, not distributed through two different mechanisms.

ohadlevy:
I'm in mixed feelings about javascript client side assets, the extra effort for packaging is immense and can potentially block innovation, we also see that this is not a hard requirement by packaging guidelines, will you be open to discuss this further?

domcleal:
Not certain who you're talking to, but if it's in response to https://github.com/theforeman/foreman-packaging/pull/1174#issuecomment-226151372, then:

Client side assets are mostly already packaged. This PR proposes to split external dependencies between both packages and a vendored bundle, which isn't an improvement - dependencies should be supplied from packages or perhaps changed to vendored entirely, but this suggests neither.  Bundling is strongly discouraged by the Fedora guidelines, so continuing to use packages seems preferable.

ohadlevy:
@domcleal you are correct that client side assets ATM are mostly packaged via gems, the move to webpack suggest a change in how assets will be handled in the future. One of the main goal is to migrate foreman core assets away to be under webpack controll (e.g. no / little rails asset pipeline assets in core).

The main motivation points for the change afaiu are the following:
1. move away from globals javascript, webpack enables our code to be refactored in a way we name space it all.
2. move to a better javascript syntax - es6
3. use a much richer javascript eco system (live reload when js code changes, early detection of syntax errors etc).
4. modern packaging system for javascript libraires, atm its non trival to include 3rd party js ilb to not pollute the global name space, and most modern libraries will simply not work (require commonJS or requirejs etc).

if you disagree with any of the points above, I suggest you review the RFC document  at https://github.com/theforeman/rfcs/pull/3.

Overall, this is a much modern stack for working with javascript, and while this deviates from the "rails way" i find it a move in the right direction for the future of our project, this is just the initial commit, a first POC (e.g.  https://github.com/theforeman/foreman/pull/3433) was already done using it and it looks promising as well.

final note about the packaging it self, I personally see little value in maintaining many packages (ATM npm has 100+ in https://github.com/theforeman/foreman/pull/3433) as build dependencies just for client side assets. the ultimate goal would be to reduce all assets related packaging (of core) to a minimum, yet we would still need to maintain precompile support for 3rd party plugins.

vendoring all assets is an option (e.g the output of the webpack process can create one or multiple js/css/img etc output) - yet to me it sounds like it will mostly spam our repository. this PR actually does something very similar, but instead of putting it in git itself, it puts it in the generated tar file to reduce changelog noise.

foreman/foreman.spec
```diff
@@ -204,10 +206,6 @@ BuildRequires: %{?scl_prefix_ror}rubygem(sass-rails) < 6
 BuildRequires: %{?scl_prefix_ror}rubygem(uglifier) >= 1.0.3
 BuildRequires: %{?scl_prefix_ror}rubygem(execjs) >= 1.4.0
 BuildRequires: %{?scl_prefix_ror}rubygem(execjs) < 2.5.0
-%if 0%{?scl:1}
-BuildRequires: %{?scl_prefix}rubygem(jquery-rails) >= 3.1
```
dLobatog: Do jQuery plugins work without this?

dLobatog:
> Client side assets are mostly already packaged. This PR proposes to split external dependencies between both packages and a vendored bundle, which isn't an improvement - dependencies should be supplied from packages or perhaps changed to vendored entirely, but this suggests neither. Bundling is strongly discouraged by the Fedora guidelines, so continuing to use packages seems preferable.

Continuing to use packages for npm dependencies sounds insane IMO. try to run `npm install ` on this PR https://github.com/theforeman/foreman/pull/3433/files then `ls node_modules | wc -l`   to see what I mean. I'd also like to see assets dependencies supplied from package.

If you feel uncomfortable with having two solutions (gemified assets + webpack), maybe we can get to a middle ground where gemified assets are pulled into the tarball too. That way we get rid of all assets packaging, not just `npm` modules. 

What `njs2rpm` does to circumvent the issue, is to create 'randomnpmpackage-bundle' rpms which contain top-level dependencies, but all subdependencies are bundled in these top-level packages. (https://github.com/sfreire/njs2rpm). In our case, that'd mean packaging:

* nodejs-bundle-jquery
* nodejs-bundle-jquery-js
* nodejs-bundle-jquery.cookie
* nodejs-bundle-jstz
* nodejs-bundle-lodash
* nodejs-bundle-select2

Then the `foreman` package would require these dependencies as `bundled-npm(jquery)`, etc... I took a look at `fpm` (another packaging solution) and it does this too. Honestly I don't see the advantages of doing this instead of just bundling everything in the tarball. 

To sum up, @domcleal what do you think of doing either of these:

* Removing gemified assets and bundling them in the tarball like we propose to do with node_modules. Stop serving frontend devps as packages.
* Making top-level dependencies that `Provide bundled-npm(xxxxx)` which contain all subdependencies. Continue serving frontend deps as packages. 

I don't think anyone wants to package 278 npm modules so one of these ought to be good enough? 

PS: Apparently OpenStack Fuel is packaging npm modules automatically as rpm/deb then using these as dependencies.. https://specs.openstack.org/openstack/fuel-specs/specs/7.0/package-javascript.html - still can't see why

tbrisker:
One point to clarify about only splitting part of the assets from the gems - the related core PR only moved part of the assets to npm to prove that it can be done, and to create a PR that is small enough (hopefully) to review and understand. 
The goal is to completely get rid of all vendored and gem-based assets in following PRs and replace them all with npm-based assets, from which we only load the needed parts for our usage using webpack.

domcleal:
> If you feel uncomfortable with having two solutions (gemified assets + webpack), maybe we can get to a middle ground where gemified assets are pulled into the tarball too.
> [..]
> Removing gemified assets and bundling them in the tarball like we propose to do with node_modules. Stop serving frontend devps as packages.

While an issue for `foreman`, not here, I don't think or expect that other project sources should be distributed in the `foreman` tarball.

> Making top-level dependencies that Provide bundled-npm(xxxxx) which contain all subdependencies. Continue serving frontend deps as packages.

Sure, OK.  Will leave this as WoC to implement.

dLobatog:
>     Removing gemified assets and bundling them in the tarball like we propose to do with node_modules. Stop serving frontend devps as packages.

> > 
> While an issue for foreman, not here, I don't think or expect that other project sources should be distributed in the foreman tarball.

@domcleal It doesn't have to be 'sources', we can minify & concatenate the JavaScript when building the tar and package just that

domcleal:
> @domcleal It doesn't have to be 'sources', we can minify & concatenate the JavaScript when building the tar and package just that

Precompiled sources aren't very useful in a source tarball, as we can't build plugins against them and assets can't be rebuilt with different options (e.g. relative path).

dLobatog:
I made a POC of nodejs-jquery here - https://github.com/theforeman/foreman-packaging/pull/1206 other packages will ensue

ohadlevy:
@dLobatog I'm not sure I follow your approach, could you explain why multiple **built time** dependence packages - each one containing a bundle of assets is better then a single one?

Since users will not use those packages directly, any change in any of the bundled libraries will most likely result in a new foreman package regardless, furthermore, I don't think this gains any points for developer happiness either, e.g. having a bundled package per main level library still requires extra effort (that is something that will be most likely non common), vs a single file with all npm modules in it.
IMHO, once we decided we can bundle it in a package, we should use one package for all built time assets.

dLobatog:
@ohadlevy Changes on the bundled libraries should just result on a new `foreman-assets` package at most, or even just updating the package (let's say jQuery) would do if the min/max versions are specified correctly. 

For developer happiness this is not the easiest, I agree. It's easier than the current model of having to package all ruby dependencies though. 

> IMHO, once we decided we can bundle it in a package, we should use one package for all built time assets.

As I said, it's a middle ground between 1. just putting the "uglified" node_modules files we need in the `foreman` package, no dependency packages required. and 2. making a package per npm dependency (insane, 257 packages IIRC) https://github.com/theforeman/foreman-packaging/pull/1174#issuecomment-228024054 . The bundled approach allows us to continue with this PR while following the guidelines for packaging and without making a zillion new nodejs packages.

ohadlevy:
On Mon, Jul 4, 2016 at 11:06 AM, Daniel Lobato García <
notifications@github.com> wrote:

> @ohadlevy <https://github.com/ohadlevy> Changes on the bundled libraries
> should just result on a new foreman-assets package at most, or even just
> updating the package (let's say jQuery) would do if the min/max versions
> are specified correctly.
>
I think this is were I missunderstand you. is that a build time dependency
or a run time?
if its  build time (as that what I thought) then, no matter which build
package is changing, the resulting compliled output will change the package
(afaiu - currently its foreman package it self, maybe in the future we
could change that).

> For developer happiness this is not the easiest, I agree. It's easier than
> the current model of having to package all ruby dependencies though.
>
> IMHO, once we decided we can bundle it in a package, we should use one
> package for all built time assets.
>
> As I said, it's a middle ground between 1. just putting the "uglified"
> node_modules files we need in the foreman package, no dependency packages
> required. and 2. making a package per npm dependency (insane, 257 packages
> IIRC) #1174 (comment)
> <https://github.com/theforeman/foreman-packaging/pull/1174#issuecomment-228024054>
> . The bundled approach allows us to continue with this PR while following
> the guidelines for packaging and without making a zillion new nodejs
> packages.
>
I'm missing option 1.5, which is have one bundle package, with all node
modules in it. this way its one package for all assets, which is only build
dependency of foreman.

dLobatog:
@ohadlevy @domcleal 
I'm trying to understand the whole picture, feel free to edit this comment: This is how I currently see the issue at hand:

Reasons against having 1 bundle package with all npm assets
* gemified assets packages + node_modules package - 2 systems for maintaining frontend assets is more work than just 1
* bundling too many libraries in one package is 'bad'
* less maintainability? (not very sure)

Reasons for having 1 bundle package with all npm assets 
* it's easier to make one package with node_modules sources than 10 packages (probably more in the future) with top-level stuff
* we're still bundling many libraries in one package if we package top-level dependencies
* people can send PRs with frontend dependencies without caring about packaging them
* jenkins can read package.json and update the spec w/o anyone having to touch foreman-packaging

domcleal:
The rubygem-webpack-rails dependency is still missing (comment at https://github.com/theforeman/foreman-packaging/pull/1174/files#r67130109), and BRs on the new packages in PRs (https://github.com/theforeman/foreman-packaging/milestone/1) need adding too. Please chainbuild it all and ensure it works to resolve dependencies, and can build assets.

dLobatog:
Cool, webpack-rails is here https://github.com/theforeman/foreman-packaging/pull/1221 , and I started a branch to chainbuild it & ensure it runs webpack:compile - https://github.com/dlobatog/foreman-packaging/tree/chainbuild-webpack

dLobatog:
Just an update, on the [branch I mentioned above](https://github.com/dlobatog/foreman-packaging/tree/chainbuild-webpack)  I'm able to build all deps & foreman-assets, however `webpack:compile` is failing, I'm trying to fix that:
```
 ** Execute webpack:compile
rake aborted!
Can't find our webpack executable at /builddir/build/BUILD/foreman-1.13.0/node_modules/.bin/webpack - have you run `npm install`?
/opt/theforeman/tfm/root/usr/share/gems/gems/webpack-rails-0.9.8/lib/tasks/webpack.rake:10:in `block (2 levels) in <top (required)>'   /```

File: foreman/foreman.spec [hunk 3]
```diff
@@ -570,6 +560,7 @@ sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/se
 export BUNDLER_EXT_NOSTRICT=1
 export BUNDLER_EXT_GROUPS="default assets"
 %{scl_rake} assets:precompile RAILS_ENV=production --trace
+%{scl_rake} webpack:compile --trace
```
dLobatog: We'll have to install webpack (`npm install') before running this, or put it in the /bin directory of the Foreman source tar. Any preferences? This means we have to add npm as a BuildRequires too, otherwise we'll get a `npm: command not found` error when building

domcleal: webpack isn't part of Foreman, so it wouldn't be distributed there. It'd just be packaged and set as a build dep.

dLobatog: Oh well.. there's more than 1 package needed to run webpack compile, I guess these will be the first 'bundled-npm(xxxx)' packages https://github.com/theforeman/foreman/pull/3433/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R8

dLobatog:
@tbrisker Can you rebase this PR - and actually make it only include these changes to the spec https://github.com/theforeman/foreman-packaging/pull/1255/commits/f5bef88304ffe5a7b9252b24aa16e118cabf1936 ?  (don't include the git annexed tar) That should make it work with the other packages https://github.com/theforeman/foreman-packaging/pull/1255

tbrisker:
Closing in favor of #1255

PR #1175, opened 2016-06-14
Add webpack:compile task to debian build proccess [deb]

tbrisker:
This is needed for theforeman/foreman#3433 to
build the webpack managed assets correctly for production.

### DIFF
--- a/debian/jessie/foreman/rules
+++ b/debian/jessie/foreman/rules
@@ -17,8 +17,10 @@ build/foreman::
 	/bin/cp config/database.yml.example config/database.yml
 	/bin/cp config/email.yaml.example config/email.yaml
 	/usr/bin/gem install -f thor -v '0.19.1'
+	/usr/bin/npm install
 	/usr/bin/bundle package
 	/usr/bin/bundle exec rake locale:pack assets:precompile RAILS_ENV=production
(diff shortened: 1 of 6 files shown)


mmoll:
I didn't test this, but do I get it correctly that, as the result in a package different to now, there will be a public/webpack folder where the assets live in?

debian/jessie/foreman/rules
```diff
@@ -19,6 +19,7 @@ build/foreman::
 	/usr/bin/gem install -f thor -v '0.19.1'
 	/usr/bin/bundle package
```
domcleal: Presumably an `npm install` is also required here?

tbrisker: Yes, thanks, I will update this.

mmoll:
seems to do its job :)

mmoll:
[test]

mmoll:
Could you please add `nodejs-legacy` to the Builddeps? /usr/bin/node should be linked to /usr/bin/nodejs then and webpack should work.

tbrisker:
@mmoll added.

domcleal:
Merged as 6c4f1de, thanks @tbrisker.

PR #1176, opened 2016-06-14
Build *.mo files for plugins

ehelms:


domcleal:
Same comment as https://github.com/theforeman/foreman-packaging/pull/1177#issuecomment-226102994, the gems should already ship mo files.

PR #1177, opened 2016-06-14
Refs #15417: Build mo files during RPM build

ehelms:
Will require https://github.com/theforeman/foreman_openscap/pull/172 and a new release that is built including those locale files, but I wanted to open this up so as not to forget about it.

### DIFF
--- a/rubygem-foreman_openscap/rubygem-foreman_openscap.spec
+++ b/rubygem-foreman_openscap/rubygem-foreman_openscap.spec
@@ -69,6 +69,8 @@ gem install --local --install-dir .%{gem_dir} \
 %{?scl:"}
 
 %build
+#build locale files
+make -C locale all-mo
 
 %install
 mkdir -p %{buildroot}%{gem_dir}


domcleal:
This should be fixed in the plugin, not RPM packaging.  Plugins should ship .mo files within the .gem because they are still required on source installations.

http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Pulling-translations-from-Transifex describes the requirements.

ehelms:
I don't see how these are different than asset compilation in production given the mo files are only required in production and are binary files. Or at least adding this extra line to ensure they get properly compiled when an RPM is built. I saw the Foreman RPM does this so should it be removed from there?

domcleal:
They're shipped in plugin gems for source installations where the MO files are required for the translations to work. Foreman doesn't have an equivalent to `assets:precompile` or `apipie:cache` to build MO files for plugins - perhaps if it did, then the gems could contain PO files only and users would build PO files as an additional step.

PR #1179, opened 2016-06-15
nonSCL route53 0.3.2

dLobatog:


rubygem-route53/rubygem-route53.spec
```diff
@@ -0,0 +1,93 @@
...
+        %{buildroot}%{gem_dir}/
+
+mkdir -p %{buildroot}%{_bindir}
+cp -pa .%{_bindir}/* \
+        %{buildroot}%{_bindir}/
+find %{buildroot}%{gem_instdir}/bin -type f | xargs chmod a+x
+
+%files
+%dir %{gem_instdir}
+%{_bindir}/route53
```
dLobatog: This is supposedly wrong by https://fedoraproject.org/wiki/Packaging:Guidelines#No_inclusion_of_pre-built_binaries_or_libraries - not sure if the proxy uses the binary directly, so maybe it's best to exclude it?

rubygem-route53/rubygem-route53.spec
```diff
@@ -0,0 +1,93 @@
...
+%global gem_name route53
+
+Name: rubygem-%{gem_name}
+Version: 0.3.2
+Release: 1%{?dist}
+Summary: Library for Amazon's Route 53 service
+Group: Development/Languages
+License: GPLv3
+URL: http://github.com/pcorliss/ruby_route_53
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
```
domcleal: Needs to be http for our tests (looks like both non-SCL templates need fixing).

comps/comps-foreman-plugins-fedora24.xml
```diff
@@ -81,6 +81,7 @@
       <packagereq type="default">rubygem-logify</packagereq>
       <packagereq type="default">rubygem-newt</packagereq>
       <packagereq type="default">rubygem-satyr</packagereq>
+      <packagereq type="default">rubygem-route53</packagereq>
```
domcleal: Not alphabetical, ditto on the others.

rubygem-route53/rubygem-route53.spec
dLobatog: I'll do that here too if you dont mind

dLobatog:
Updated, including the gem2rpm templates

rubygem-route53/rubygem-route53.spec
domcleal: It isn't a pre-built binary, it's a bin stub script that's generated by `gem install`.  The guideline is forbidding binaries/libraries that are already built in the source tarball/gem.  It's fine as-is.

rubygem-route53/rubygem-route53.spec
```diff
@@ -0,0 +1,93 @@
...
+        %{buildroot}%{_bindir}/
+find %{buildroot}%{gem_instdir}/bin -type f | xargs chmod a+x
+
+%files
+%dir %{gem_instdir}
+%{_bindir}/route53
+%exclude %{gem_instdir}/.gitignore
+%{gem_instdir}/.rspec
+%{gem_instdir}/.ruby-gemset
+%{gem_instdir}/.ruby-version
```
domcleal: All of these dot-files can be excluded, either .* or individually

dLobatog:
Updated to exclude dotfiles from the package

domcleal:
Could you rebase this PR please?

domcleal:

PR #1180, opened 2016-06-15
nonSCL rubygem- builder 3.2.2 - EL6 only

dLobatog:


comps/comps-foreman-plugins-rhel6.xml
```diff
@@ -67,6 +67,7 @@
       <packagereq type="default">tfm-rubygem-azure-core</packagereq>
       <packagereq type="default">tfm-rubygem-angular-rails-templates</packagereq>
       <packagereq type="default">tfm-rubygem-apipie-params</packagereq>
+      <packagereq type="default">tfm-rubygem-builder</packagereq>
```
domcleal: No tfm- prefix for non-SCL packages.

rubygem-builder/rubygem-builder.spec
```diff
@@ -0,0 +1,82 @@
...
+BuildRequires: ruby(release)
+%endif
+BuildRequires: ruby
+BuildRequires: rubygems-devel
+BuildArch: noarch
+Provides: rubygem(%{gem_name}) = %{version}
+
+%description
+Builder provides a number of builder objects that make creating structured
+data
```
domcleal: odd wrapping here

rubygem-builder/rubygem-builder.spec
```diff
@@ -0,0 +1,82 @@
...
+%gem_install
+
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+%files
+%dir %{gem_instdir}
+%{gem_instdir}/CHANGES
```
domcleal: move to -doc

dLobatog:
Updated the spec to fix description wrapping & moved CHANGES to doc. Also the comps file lists it as rubygem-builder instead of tfm-rubygem-

domcleal:

PR #1181, opened 2016-06-15
nonSCL ruby-hmac 0.4.0 - EL6/7 only

dLobatog:


domcleal:
Merged and building, thanks @dLobatog!

PR #1182, opened 2016-06-15
nonSCL smart_proxy_dns_route53 2.0.0

dLobatog:
Depends on

- [x] https://github.com/theforeman/foreman-packaging/pull/1181
- [x] https://github.com/theforeman/foreman-packaging/pull/1180
- [x] https://github.com/theforeman/foreman-packaging/pull/1179


rubygem-smart_proxy_dns_route53/rubygem-smart_proxy_dns_route53.spec
```diff
@@ -0,0 +1,77 @@
+# Generated from smart_proxy_dns_route53-2.0.0.gem by gem2rpm -*- rpm-spec -*-
```
domcleal: Looks like this is using the wrong template, it should use the smart proxy plugin template which installs bundler.d files etc.

dLobatog:
My bad, I was a bit hasty making this, the latest spec is made from the gem2rpm template for smart proxy plugins.

rubygem-smart_proxy_dns_route53/rubygem-smart_proxy_dns_route53.spec
```diff
@@ -0,0 +1,96 @@
...
+        %{buildroot}%{gem_dir}/
+
+# bundler file
+mkdir -p %{buildroot}%{foreman_proxy_bundlerd_dir}
+mv %{buildroot}%{gem_instdir}/bundler.d/%{plugin_name}.rb \
+   %{buildroot}%{foreman_proxy_bundlerd_dir}
+
+# sample config
+mkdir -p %{buildroot}%{foreman_proxy_settingsd_dir}
+mv %{buildroot}%{gem_instdir}/settings.d/%{plugin_name}.yml.example \
```
domcleal: The build appears to fail as this is config/%{plugin_name}.yml for this plugin, not settings.d and .example.

It seems smart proxy plugins are pretty inconsistent about this, so it'd be useful if you could fix the gem2rpm ERB template to check `spec.files` to find out the exact path of the file the spec should be copying (similar to how the foreman_plugin ERB template checks for presence of paths to find DB migrations etc).

dLobatog: What I think I'm doing to do here is to copy any .yml  or .yml.example on `/config`to `foreman_proxy_settingsd_dir`. That sounds better than having both config/ and settings.d/ directories storing the same kind of files.

dLobatog:
https://github.com/theforeman/foreman-packaging/pull/1199 handles `config/` in addition to `settings.d/` on the gem2rpm templates now

domcleal:
Thanks, could you please run ./comps_doc.sh to add the -doc subpackages to the comps files?

dLobatog:
-doc subpackages added :smile:

domcleal:
Merged and building for nightly, thanks @dLobatog!

PR #1183, opened 2016-06-16
Use default ruby instead of specific version

sathieu:
ruby-interpreter was used when Ruby packages were using alternatives, now the 'ruby' packages depends on the default version.

### DIFF
--- a/debian/jessie/foreman-proxy/control
+++ b/debian/jessie/foreman-proxy/control
@@ -10,7 +10,7 @@ Vcs-Browser: https://github.com/theforeman/smart-proxy
 
 Package: foreman-proxy
 Architecture: all
-Depends: ruby2.1|ruby-interpreter, rake, ruby-sinatra (>= 1.3.3), ruby-rack (>= 1.1.0), ruby-json, ruby-augeas, ruby-bundler-ext
+Depends: ruby | ruby-interpreter, rake, ruby-sinatra (>= 1.3.3), ruby-rack (>= 1.1.0), ruby-json, ruby-augeas, ruby-bundler-ext
 Recommends: sudo, wget, ping, ruby-gssapi, ruby-rubyipmi (>= 0.9.2), ruby-rkerberos (>= 0.1.1), foreman-debug
 Description: RESTful proxies for DNS, DHCP, TFTP, and Puppet
  Smart-Proxy is a project which provides a RESTful API to various sub-systems
--- a/debian/trusty/foreman-proxy/control
+++ b/debian/trusty/foreman-proxy/control
@@ -10,7 +10,7 @@ Vcs-Browser: https://github.com/theforeman/smart-proxy
 
 Package: foreman-proxy
 Architecture: all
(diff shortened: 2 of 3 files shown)


sathieu:
@mmol FYI

domcleal:
What's this fixing exactly?  The foreman-ruby symlink still points to the specific version, so I can't see what installing `ruby` versus `ruby2.1` would help with.

sathieu:
@domcleal  /usr/share/foreman-proxy/bin/smart-proxy has this:
```
#!/usr/bin/env ruby
```
And:
```
$ ls -l /usr/bin/ruby
lrwxrwxrwx 1 root root 7 avril 14 09:46 /usr/bin/ruby -> ruby2.3
```
i don't see a foreman-ruby symlink

domcleal:
Ah sorry, I thought the patch was against `foreman`, not `foreman-proxy`.

domcleal:
If the smart proxy script relies on /usr/bin/ruby, then shouldn't the dependency _only_ be on `ruby` in order to get the symlink?  Ditto for all OSes, not just jessie.

domcleal:
> If the smart proxy script relies on /usr/bin/ruby, then shouldn't the dependency only be on ruby in order to get the symlink?

foreman-installer does this, it depends on `ruby | ruby-interpreter` and has a shebang line of `/usr/bin/env ruby` - no dependency on a specific version.

sathieu:
@domcleal done ```ruby | ruby-interpreter```

domcleal:
That looks correct to me, but the same change should be made to debian/trusty/foreman-proxy/ and xenial too, please.

sathieu:
Note that ruby-interpreter is gone in Debian >= stretch.

domcleal:
[test]

domcleal:
:+1: seems OK and tests fine, though it's cosmetic because other dependencies still depend on `ruby`.

(Packages are in my stagingdeb repo as the branch name here is missing the deb/ prefix.)

mmoll:
This looks good to me, maybe also remove the `| ruby-interpreter` part when it's superflous/non-working with all the Debian/ubuntu versions we support in develop.

sathieu:
it's not only cosmetic, it allowed me to install foreman-proxy on stretch while removing ruby2.1 (I still have some jessie packages because of ruby-rkerberos, which should be an optional dep IMO).

mmoll:
rkerberos is already only recommended in current nightlies.

sathieu:
> rkerberos is already only recommended in current nightlies.
Could this be done in 1.12 too?

mmoll:
merged, merci @sathieu!
I cherry-picked 00c4a7e as 65d7cdd for 1.12 RC2. This PR (65d7cdd) is targeted for 1.12.1. What I'm doing on Debian/testing is to use the Ubuntu/xenial packages, as these should work without modifications on current rolling Debians.

sathieu:
Danke @mmol ;-)

PR #1184, opened 2016-06-16
Use default ruby instead of specific version

sathieu:
ruby-interpreter was used when Ruby packages were using alternatives, now the 'ruby' packages depends on the default version.

### DIFF
--- a/debian/jessie/foreman-proxy/control
+++ b/debian/jessie/foreman-proxy/control
@@ -10,7 +10,7 @@ Vcs-Browser: https://github.com/theforeman/smart-proxy
 
 Package: foreman-proxy
 Architecture: all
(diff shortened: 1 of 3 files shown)


sathieu:
Same as #1183

domcleal:
I think we'd probably only merge this to deb/develop, as I don't think it fixes any particular bug needed for 1.12.

mmoll:
I don't see the need for 1.12 here, @sathieu which particular problem does this fix for you?

sathieu:
I need it to use jessie packages on stretch (as there is no stretch package, and I don't know how to build them myself).

mmoll:
Get a tarball from http://downloads.theforeman.org/foreman-proxy/ and move debian/jessie/foreman-proxy from this repo here to /path/to/extraced/proxy/source/debian, then build the package as usual (iirc).

mmoll:
see comment in #1183

PR #1188, opened 2016-06-21
Update kafo to 0.7.5

ares:
into 1.11 please

### DIFF
--- a/dependencies/jessie/kafo/changelog
+++ b/dependencies/jessie/kafo/changelog
@@ -1,3 +1,9 @@
+ruby-kafo (0.7.5-1) stable; urgency=low
+
+  * Update kafo to 0.7.5
+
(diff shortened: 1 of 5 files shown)


domcleal:
The fix in this update hasn't been released to nightly or 1.12 yet, are you going to be doing that too (first)?

ares:
yes, I'll likely do 0.9.1 today, I suppose 0.7.5 would not get into repos before new 1.11.x release so I though we have some time

domcleal:
I'm part way through a 1.11.3 release right now, so it's probably best to wait until at least tomorrow (I expect) before merging this and #1189, so we can get the change into the later releases before pushing straight into 1.11.

ares:
Ok, I opened https://github.com/theforeman/foreman-packaging/pull/1192 with 0.9.1 release.

domcleal:
Merged as 5b4cbdb, thanks @ares, and building at http://ci.theforeman.org/job/packaging_build_deb_dependency/1287/.

PR #1192, opened 2016-06-21
Update kafo to 0.9.1

ares:
This should go to nightlies and 1.12


domcleal:
Thanks @ares, building now and will retag to 1.12 shortly.

PR #1193, opened 2016-06-21
Update kafo to 0.9.1

ares:
This should go into nightlies and 1.12 (mirrors #1192)


domcleal:
Merged and also as fb6f428 to deb/1.12, thanks @ares.

Building at http://ci.theforeman.org/job/packaging_build_deb_dependency/1285/ and 1286.

PR #1194, opened 2016-06-21
Update dynflow to 0.8.11 in 1.12 (RPM)

iNecas:


domcleal:
Merged and building now, thanks @iNecas.

If you'd like a plugin update available in older releases in future, just ask when opening the PR against */develop, it's trivial for whoever's merging to do it there and then. Else since you have commit access you can also cherry pick the update commit, `tito tag --keep-version` (that argument is important), and release it.

iNecas:
Thanks for info, I will do that the next time I need to get something into older release

PR #1195, opened 2016-06-21
Update dynflow to 0.8.11 and foreman-tasks to 0.7.8 to Foreman 1.11 (RPM)

iNecas:


iNecas:
I'm getting "Error: No Package found for tfm-rubygem(dynflow) >= 0.8.8" in koji scratch build for foreman-tasks, while 0.8.10 should already be present in the repos: am I missing something obvious again?

domcleal:
Don't worry about it, the PR test job isn't working against rpm/1.11 properly. It's trying to test both the koji-foreman (core) and plugin releasers, and obviously failing on the former as the dep's only in the plugins tag. The PR test works properly on 1.12 and newer as the whitelist there prevents koji-foreman being used to build plugins.

domcleal:
Merged and building, thanks @iNecas.

PR #1198, opened 2016-06-22
update chef-api to 0.6.0

mmoll:
for nightly only, please. As Ubuntu/trusty uses Ruby 1.9.3 for foreman-proxy, no changes here.


domcleal:
Merged and building for both OSes, thanks @mmoll.

PR #1199, opened 2016-06-23
gem2rpm smart_proxy_plugin settings file handling

dLobatog:
This commit helps gem2rpm deal with inconsistent smart_proxy plugins
which use config/ to store sample config files instead of settings.d/

It also bundles any .yml files found in these folders. In case the
settings file name doesn't match with the plugin name, this will fix it.

### DIFF
--- a/gem2rpm/smart_proxy_plugin.spec.erb
+++ b/gem2rpm/smart_proxy_plugin.spec.erb
@@ -112,8 +112,13 @@ mv %{buildroot}%{gem_instdir}/bundler.d/%{plugin_name}.rb \
 
 # sample config
 mkdir -p %{buildroot}%{foreman_proxy_settingsd_dir}
-mv %{buildroot}%{gem_instdir}/settings.d/%{plugin_name}.yml.example \
-   %{buildroot}%{foreman_proxy_settingsd_dir}/%{plugin_name}.yml
+<% config_file_paths = spec.files.select { |f| f.start_with?('config/') } +
+                       spec.files.select { |f| f.start_with?('settings.d/') } -%>
+<% config_file_paths.each do |config_file_path| -%>
+<% config_file = config_file_path.split('/').last -%>
+mv %{buildroot}%{gem_instdir}/<%= config_file_path %> \
+   %{buildroot}%{foreman_proxy_settingsd_dir}/<%= config_file.chomp('.example') %>
+<% end -%>
 
 %files
 %dir %{gem_instdir}
@@ -125,13 +130,16 @@ mv %{buildroot}%{gem_instdir}/settings.d/%{plugin_name}.yml.example \
 <% unless spec.extensions.empty? -%>
 %{gem_extdir_mri}
 <% end -%>
+<% config_file_paths = config_file_paths.map { |path| path.split('/').last.chomp('.example') } -%>
+<% config_file_paths.each do |config_file_path| -%>
+%config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/<%= config_file_path %>
+<% end -%>
 <%= main_file_entries(spec).gsub(/^%license/, '%doc') %>
 <% unless doc_subpackage -%>
 %doc %{gem_docdir}
 <%= doc_file_entries(spec) -%>
 <% end -%>
 %{foreman_proxy_bundlerd_dir}/%{plugin_name}.rb
-%config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/%{plugin_name}.yml
 %exclude %{gem_cache}
 %{gem_spec}
 


gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -112,8 +112,12 @@ mv %{buildroot}%{gem_instdir}/bundler.d/%{plugin_name}.rb \
 
 # sample config
 mkdir -p %{buildroot}%{foreman_proxy_settingsd_dir}
-mv %{buildroot}%{gem_instdir}/settings.d/%{plugin_name}.yml.example \
+<% config_files = spec.files.select { |f| f.start_with?('config/') } +
+                  spec.files.select { |f| f.start_with?('settings.d/') } %>
+<% config_files.each do |config_file| %>
+mv %{buildroot}%{gem_instdir}/<%= config_file %> \
    %{buildroot}%{foreman_proxy_settingsd_dir}/%{plugin_name}.yml
```
domcleal: > It also bundles any .yml files found in these folders. In case the settings file name doesn't match with the plugin name, this will fix it.

It's only going to keep one file and not multiple, because they're all being moved to the same destination filename. I think it might be best to keep the loop for every file, but chomp any .example suffixes from the filename and copy it exactly into place.

It's probably best to preserve the original filename instead of using %{plugin_name} for two reasons:

1. It supports multiple files, if a plugin had them.
2. The smart proxy doesn't just load every file, it loads specific filenames based on the internal plugin name or the plugin configuration. The plugin author knows best, so %{plugin_name} may be incorrect.

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -112,8 +112,14 @@ mv %{buildroot}%{gem_instdir}/bundler.d/%{plugin_name}.rb \
 
 # sample config
 mkdir -p %{buildroot}%{foreman_proxy_settingsd_dir}
-mv %{buildroot}%{gem_instdir}/settings.d/%{plugin_name}.yml.example \
-   %{buildroot}%{foreman_proxy_settingsd_dir}/%{plugin_name}.yml
+<% config_file_paths = spec.files.select { |f| f.start_with?('config/') } +
+                       spec.files.select { |f| f.start_with?('settings.d/') } %>
+<% config_file_paths.map! { |config_file| config_file.chomp('.example') } %>
```
dLobatog: This is now checking all `config/` and `settings.d/` files, chomping `.example` then putting it in `settingsd_dir` with the original name. Notice `split('/').last` works even if there were subdirectories like `settings.d/dns/route53.yml`

dLobatog:
Updated

gem2rpm/smart_proxy_plugin.spec.erb
domcleal: This is chomping .example off of the source path as well as the destination, so for example:

<pre>
$ gem2rpm -t smart_proxy_plugin.spec.erb --fetch smart_proxy_remote_execution_ssh | grep yml
mv %{buildroot}%{gem_instdir}/settings.d/remote_execution_ssh.yml \
   %{buildroot}%{foreman_proxy_settingsd_dir}/remote_execution_ssh.yml
mv %{buildroot}%{gem_instdir}/settings.d/smart_proxy_remote_execution_ssh.yml \
   %{buildroot}%{foreman_proxy_settingsd_dir}/smart_proxy_remote_execution_ssh.yml
</pre>

It should be copying `remote_execution_ssh.yml.example` to `remote_execution_ssh.yml`.

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -112,8 +112,14 @@ mv %{buildroot}%{gem_instdir}/bundler.d/%{plugin_name}.rb \
 
 # sample config
 mkdir -p %{buildroot}%{foreman_proxy_settingsd_dir}
-mv %{buildroot}%{gem_instdir}/settings.d/%{plugin_name}.yml.example \
-   %{buildroot}%{foreman_proxy_settingsd_dir}/%{plugin_name}.yml
+<% config_file_paths = spec.files.select { |f| f.start_with?('config/') } +
+                       spec.files.select { |f| f.start_with?('settings.d/') } %>
```
domcleal: Nitpick, change some end ERB markers to `-%>` to prevent lots of blank lines appearing in the output.

dLobatog:
Updated to close ERB tags with `-%>`, I was not aware that fixed the empty lines problem

domcleal:
I think the issue about .example being chomped is unfixed still, see one of the hidden diff conversations (https://github.com/theforeman/foreman-packaging/pull/1199#discussion_r68231625).

dLobatog:
Updated so that now the source in `mv` still contains `.example`. Notice if a plugin contains two files with the same name under `settings.d` or `config`, say `dns.yml.example` and `dns.yml` they'd both be copied to the same destination (plugins shouldn't have this kind of structure though)

domcleal:
Makes sense, and that part looks fine. Can you also update the %files list please? It currently only lists `%{plugin_name}.yml` as a known settings file.

dLobatog:
Updated to list the config files in %files

domcleal:
Merged, thanks @dLobatog.

PR #1200, opened 2016-06-24
re-add ovirt_provision_plugin (RPM)

mmoll:
This should be built into:
* [x] Nightly
* [x] 1.12
* [x] 1.11
* [ ] 1.10

see https://github.com/theforeman/ovirt_provision_plugin/pull/8


domcleal:
This also needs adding to comps/.

mmoll:
should all be fixed now

rubygem-ovirt_provision_plugin/rubygem-ovirt_provision_plugin.spec
```diff
@@ -0,0 +1,111 @@
...
+Name:       %{?scl_prefix}rubygem-%{gem_name}
+Version:    1.0.2
+Release:    1%{?foremandist}%{?dist}
+Group:      Applications/System
+License:    GPLv3
+URL:        http://github.com/theforeman/ovirt_provision_plugin
+Source0:    http://rubygems.org/downloads/%{gem_name}-%{version}.gem
+
+Requires:   foreman >= 1.11.0
+Requires:   %{?scl_prefix}rubygem(deface)
```
dLobatog: Not sure if this should be an open dependency, I recall there were breaking changes pre 1.0 - https://github.com/spree/deface#upgrading-from-09-to-10

mmoll: this only mirrors the open dependency in the gem

domcleal:
Thanks @mmoll, will build for all three releases.

PR #1206, opened 2016-06-29
nodejs-jquery 1.11.3

dLobatog:
First version of nodejs-jquery 1.11.3, as the future changes in Foreman
frontend pipeline using Webpack will need it. This package has no dependencies so it doesn't need to "Provide" anything else or be named 'nodejs-bundle-jquery'


domcleal:
This is failing to build, setting to WoC until fixed.

dLobatog:
It builds fine now, however it'd be set to 'After core merge' I guess. It doesn't make much sense to have the package in our repos for no reason.

nodejs-jquery/nodejs-jquery.spec
```diff
@@ -0,0 +1,57 @@
...
+
+Name: %{?scl_prefix}nodejs-%{npm_name}
+Version: 1.11.3
+Release: 1%{?dist}
+Summary: JavaScript library for DOM operations
+License: MIT
+Group: Development/Libraries
+URL: https://github.com/jquery/jquery.git
+Source0: http://registry.npmjs.org/jquery/-/jquery-1.11.3.tgz
+BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
```
domcleal: Obsolete (EL5 era), can be removed.

nodejs-jquery/nodejs-jquery.spec
```diff
@@ -0,0 +1,57 @@
...
+BuildArch: noarch
+
+%description doc
+This package contains documentation for nodejs-%{npm_name}.
+
+%prep
+%setup -q -n package
+
+%install
+rm -rf %{buildroot}
```
domcleal: Obsolete (EL5 era), can be removed.

comps/comps-foreman-fedora24.xml
```diff
@@ -31,6 +31,7 @@
       <packagereq type="default">foreman-vmware</packagereq>
 
       <packagereq type="default">foreman-installer</packagereq>
+      <packagereq type="default">nodejs-jquery</packagereq>
```
domcleal: Move into a new section below like you've done with others, this bit contains installer deps.

nodejs-jquery/nodejs-jquery.spec
```diff
@@ -0,0 +1,57 @@
...
+mkdir -p %{buildroot}%{nodejs_sitelib}/jquery
+cp -pfr dist src  *.json *.md *.js *.txt node_modules %{buildroot}%{nodejs_sitelib}/jquery > /dev/null 2>&1 || :
```
domcleal: Remove `|| :` here, or it'll hide errors.

nodejs-jquery/nodejs-jquery.spec
```diff
@@ -0,0 +1,57 @@
...
+Name: %{?scl_prefix}nodejs-%{npm_name}
+Version: 1.11.3
+Release: 1%{?dist}
+Summary: JavaScript library for DOM operations
+License: MIT
+Group: Development/Libraries
+URL: https://github.com/jquery/jquery.git
+Source0: http://registry.npmjs.org/jquery/-/jquery-1.11.3.tgz
+BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+Requires: nodejs(engine)
```
domcleal: The package name is using SCL macros, but these aren't - did you want to include %{?scl_nodejs} here so they switch to the SCL collection when present (tfm can be rebuilt with it)?  Otherwise I'm not sure the SCL prefixes are necessary.

nodejs-jquery/nodejs-jquery.spec
```diff
@@ -0,0 +1,57 @@
...
+mkdir -p %{buildroot}%{nodejs_sitelib}/jquery
+cp -pfr dist src  *.json *.md *.js *.txt node_modules %{buildroot}%{nodejs_sitelib}/jquery > /dev/null 2>&1 || :
+
+%build
+
+%check
+%nodejs_symlink_deps --check
+
+%files
+%defattr(-,foreman,foreman,-)
```
domcleal: These shouldn't be owned by `foreman` as the package doesn't depend on foreman for the account, nor does foreman need write access. Perhaps leave this out?

nodejs-jquery/nodejs-jquery.spec
domcleal: Actually, the SCL prefixes should probably be removed unless this is built with a NodeJS SCL because the files will not be in the SCL root.

dLobatog: The name uses SCL macros so it's served as part of our SCL (tfm-nodejs-jquery), other packages like tfm-nodejs-jquery-ujs can have a `Requires:  %{?scl_prefix}npm(jquery)`. 

nodejs(engine), nodejs-devel and nodejs-packaging are served already in all of our distributions, I don't think we should put these in the SCL, we can require them wherever. 

That way we know which assets are served as part of Foreman (tfm-nodejs) vs distribution packaged assets.

domcleal: > The name uses SCL macros so it's served as part of our SCL (tfm-nodejs-jquery), other packages like tfm-nodejs-jquery-ujs can have a Requires: %{?scl_prefix}npm(jquery).

The files are not packaged in the SCL root though, so there's no point in just having the prefix. When it's using the non-SCL NodeJS build/devel packages and paths, the SCL prefix should be removed to match.

> nodejs(engine), nodejs-devel and nodejs-packaging are served already in all of our distributions, I don't think we should put these in the SCL, we can require them wherever.

Yeah, that's fine, it doesn't make much difference using the SCL or non-SCL version as they both appear to be maintained.

dLobatog: Sorry that last bit confused me:

As I understood your comment, there's no hard requirement on requiring the SCL nodejs engine/devel/packaging  instead of the non-SCL ones, but if I don't the files won't be in the SCL root. These packages are not in our SCL, so the only way I have to package it is on the 'global space' as nodejs-jquery. 

Maybe there's a way of putting %files in the SCL root manually on %install? My understanding is that if I package this as `nodejs-jquery` and Fedora/CentOS/RHEL decide to package it for the distro, the package from the distro would prevail and wipe this one out if it's newer.

domcleal: Yes, but that sounds like correct behaviour to prefer the newer package if it exists.  Adding the SCL prefix to avoid that doesn't help on Fedora, nor does it prevent file conflicts if the package installs into the system root, not the SCL root.

If you want to build the package with an SCL prefix and into the SCL root then the `tfm` SCL should depend on a NodeJS SCL so the various %nodejs* macros get redefined to the right location (https://github.com/theforeman/foreman-packaging/compare/rpm/develop...domcleal:rpm/develop-execjs-nodejs should do this).

comps/comps-foreman-rhel7.xml
```diff
@@ -42,7 +41,26 @@
 
       <packagereq type="default">tfm-mod_passenger</packagereq>
 
+      <!-- Assets dependencies -->
```
dLobatog: I started doing this to add `nodejs-jquery` here, then it's when I realized it'd be only non-scl so it had to be in other section. Anyway I kept these sections just in case.

nodejs-jquery/nodejs-jquery.spec
```diff
@@ -0,0 +1,51 @@
...
+
+%description doc
+This package contains documentation for nodejs-%{npm_name}.
+
+%prep
+%setup -q -n package
+
+%install
+mkdir -p %{buildroot}%{nodejs_sitelib}/%{npm_name}
+cp -pfr dist src *.json *.md *.txt %{buildroot}%{nodejs_sitelib}/%{npm_name}
```
dLobatog: Notice this doesn't hide errors anymore. It used to hide an error where a folder that didn't exist was supposed to be copied.

dLobatog:
Updated according to all comments, and taken out from the SCL

dLobatog:
Reminder to self @dlobatog - document how you're creating these packages with njs2rpm and upload the template

nodejs-jquery/nodejs-jquery.spec
dLobatog: BTW dist/ includes the source, and the minified version. I guess the package has to contain both?

dLobatog:
Rebased

domcleal:
Merged as 3eb83cb, thanks @dLobatog.

PR #1207, opened 2016-06-29
Add tfm-rubygem-ovirt_provision_plugin-doc on el7 comps

dLobatog:

* [x] Nightly
* [x] 1.12
* [x] 1.11
* [x] 1.10


domcleal:
It's already listed, but it's in the wrong section. Can you remove the existing entry?

domcleal:
Thanks, merged.

PR #1212, opened 2016-07-06
fixes #15594 - Use Ruby 2.0 on trusty for hammer

mmoll:


mmoll:
tested the resulting hammer_cli and hammer_cli_foreman package on a trusty system and it seems to work

domcleal:
Merged, thanks @mmoll.  Building all of the packages now.

PR #1213, opened 2016-07-07
Update foreman_expire_hosts to v2.0.2

timogoebel:
Please build for

* [x] Nightly
* [x] 1.12
* [x] 1.11


domcleal:
Merged and building now, thanks @timogoebel.

As an aside, if you can leave the changelog out of future commits then it helps with cherry picking, since tito generates it.

timogoebel:
> As an aside, if you can leave the changelog out of future commits then it helps with cherry picking, since tito generates it.

Will do, even less work for me 👍 - thanks @domcleal !

PR #1220, opened 2016-07-11
fixes #15641 - create migration_state file in correct dir [deb]

domcleal:
The file's entirely missing from the package (/etc/foreman-proxy/migration_state) at the moment, since 38a2776.


mmoll:
merged, thanks @domcleal! please cherry-pick at will to 1.12 (and probably also to 1.11?)

PR #1223, opened 2016-07-14
Webpack devDependencies

dLobatog:
For the moment I only packaged webpack as a bundle, the rest will come up soon. The source tgz for webpack has been packaged with all bundled dependencies.

https://github.com/dlobatog/njs2rpm contains the fixes to njs2rpm to do this, `./njs2rpm webpack 1.13.1 1 bundle spec` should get you 95% of the spec completed + the source tar created.

### DIFF
--- a/nodejs-bundle-webpack/nodejs-bundle-webpack.spec
+++ b/nodejs-bundle-webpack/nodejs-bundle-webpack.spec
@@ -1,0 +1,191 @@
+%global npm_name webpack
+
+Name: nodejs-%{npm_name}
+Version: 1.13.1
+Release: 1%{?dist}
+Summary: Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jade, coffee, css, less, ... and your custom stuff.
+License: MIT
+Group: Development/Libraries
+URL: https://github.com/webpack/webpack.git
+Source0: http://registry.npmjs.org/webpack/-/webpack-1.13.1.tgz
+Requires: nodejs(engine)
+BuildRequires: nodejs-devel
+BuildRequires: nodejs-packaging
+BuildArch: noarch
+ExclusiveArch: %{nodejs_arches} noarch
+Provides: npm(%{npm_name}) = %{version}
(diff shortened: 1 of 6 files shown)


nodejs-bundle-webpack/nodejs-bundle-webpack.spec
```diff
@@ -0,0 +1,191 @@
+%global npm_name webpack
+
+Name: nodejs-%{npm_name}
+Version: 1.13.1
+Release: 1%{?dist}
+Summary: Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jade, coffee, css, less, ... and your custom stuff.
+License: MIT
+Group: Development/Libraries
+URL: https://github.com/webpack/webpack.git
+Source0: http://registry.npmjs.org/webpack/-/webpack-1.13.1.tgz
```
domcleal: All other bundled packages inside this should be listed here as sources too, then they can all be installed in %prep, %install etc.

dLobatog: The way I was doing it https://github.com/dLobatog/njs2rpm/commit/084b5187c3cfd2b54e9db17144c82c68007ec8a0 seems easier but it won't match the source. I think I can use the same list of packages I put in Provides and convert them to URLs here.

domcleal:
In response to your comments in #theforeman-dev, I gather the issue with this PR is that the sources have to be unpacked into their nested dependency structure.

It looks like nodejs-symlink-deps is insufficient here as it assumes there will be a single major version of a single dependency in %nodejs_sitelib, and the file that defines whether there are multiple versions is part of the nodejs-packaging RPM itself.

Here's an alternative approach, amending this commit that uses `npm` offline to perform the installation from the same source tarballs, plus a cache of the registry information: https://github.com/domcleal/foreman-packaging/commit/149b21195a27eafe8dc36e55454ca1c31c6a2355#diff-2fbc5d49e1f68f9a391153fadade5155, more on the branch at https://github.com/domcleal/foreman-packaging/commits/rpm/develop-12364-webpack.

domcleal:
Obsoleted by #1242.

PR #1228, opened 2016-07-15
core package: recommend python-minimal (DEB)

mmoll:
otherwise opening a VM console will fail, as the web socket opening is python based:
```
root@xxx:~# cat /usr/share/foreman/extras/noVNC/websockify.py
#!/usr/bin/python

import websockify

websockify.websocketproxy.websockify_init()
```
I'd like to see this also in 1.12.


mmoll:
[test]

domcleal:
Merged, thanks @mmoll!

PR #1229, opened 2016-07-15
add ruby-libvirt 0.6.0 (DEB)

mmoll:


mmoll:
This is to solve http://projects.theforeman.org/issues/15462 for Debian based OSes. The packaging files were taken unmodified from Debian upstream (except changing back libvirt-daemon to libvirt-bin for trusty).

I'd like to see this also in 1.12.

domcleal:
Merged and building for both develop and 1.12, thanks @mmoll.

PR #1233, opened 2016-07-21
adding qpid copr repo for el6 repoclosure

jlsherrill:
we are now able to use qpid from upstream in epel for el7
and from this copr repo for el6, so we are switching over to it


jlsherrill:
Please ignore, not only is this a copr lookaside, but katello uses its own repoclosure script now.

PR #1234, opened 2016-07-21
Update apipie-params to 0.0.5 (RPM)

iNecas:


domcleal:
Please also remove the dependency on `rake`.

iNecas:
Right, I have not realized it was added to the spec

domcleal:
Yeah, that's why I was removing it again!

Thanks @iNecas, building now.

PR #1236, opened 2016-07-21
Update kafo to 0.9.2 (RPM)

ares:

* [x] Nightly
* [x] 1.12
* [ ] 1.11
* [ ] 1.10

---


dLobatog:
@ares The highline Requires is missing '>= 1.6.21' as per the Kafo gemspec, otherwise :+1: !https://github.com/theforeman/kafo/blob/master/kafo.gemspec

ares:
hm, that hasn't changed for a while, I wonder if there was some reason... anyway added, good catch

domcleal:
Merged, thanks @ares.

PR #1238, opened 2016-07-21
rubygem-x-editable-rails 1.5.5

dLobatog:
Just for nightly, needed after https://github.com/theforeman/foreman/pull/3523/files


domcleal:
Merged and building, thanks @dLobatog!

PR #1240, opened 2016-07-22
foreman-node_modules - proof of concept [wip]

dLobatog:
This is a proof of concept where we have a BuildRequirement in Foreman on a package that provides all `node_modules`. 

This would be the implementation of point 3 here: https://groups.google.com/forum/?fromgroups=#!topic/foreman-dev/x37DmiGT0q4 - 

I thought about writing a script to generate all of the 'Provides' for the top level stuff contained in node_modules, and 'Provides: bundled(npm-xxx)' for the subdependencies, but that's quite wrong as other packages in the system shouldn't build against these.

The way foreman-node_modules works is by dumping node_modules in /opt/foreman/node_modules, then copying that sources to the Foreman root to ensure webpack:compile can run. Aside from that the Foreman package itself now has a BuildDependency on node itself (& libuv) to run webpack.


dLobatog:
It builds now properly, runs webpack:compile and everything. I suppose we should 'mv' instead of 'cp' the stuff from foreman-node_modules but anyways it's working!

domcleal:
> PS: It's not building here on Jenkins because there's no Source0 with URL for foreman-node_modules,

The other PRs use pre-existing sources so will work much more reliably - please update #1223.

dLobatog:
Closing as the `nodejs-bundled-xxx` packages option looks more likely now, this was just a POC after all

PR #1242, opened 2016-07-27
nodejs-bundle-webpack 1.13.1

dLobatog:


dLobatog:
Looks like it fails to fetch webpack-1.13.1-registry.npmjs.org.tgz

nodejs-bundle-webpack/webpack-1.13.1-registry.npmjs.org.tgz
```diff
@@ -0,0 +1 @@
+../.git/annex/objects/F9/MG/SHA256E-s462617--9134da88a58c0a92e72da2e7d2c6f06b187f5c06c90683f5baba383bf3873ad1.org.tgz/SHA256E-s462617--9134da88a58c0a92e72da2e7d2c6f06b187f5c06c90683f5baba383bf3873ad1.org.tgz
```
domcleal: Add this to git directly, no git-annex.

dLobatog:
Not sure what's exactly wrong with this one, it's having trouble installing inherits@2.0.1 which should be available, I see it in the cache too. I'll investigate

dLobatog:
Looks like it's working now, at least locally on el7 I've been able to build it offline. I'm not sure why but there seems to be a bug where the order of addition to the npm_cache matters. I moved `inherits` to be added as soon as possible to the cache (right after webpack) and it seemed to fix the problem. Still not sure why though.

dLobatog:
Success - tbh no idea what was the problem with the old package, I just recreated the spec, downloaded all sources again with npm2rpm and generated the tgz and it's good now.

nodejs-webpack/nodejs-webpack.spec
```diff
@@ -0,0 +1,533 @@
...
+Provides: bundled-npm(number-is-nan) = 1.0.0
+Provides: bundled-npm(supports-color) = 2.0.0
+Provides: bundled-npm(is-property) = 1.0.2
+Provides: bundled-npm(qs) = 6.2.1
+AutoReq: no
+AutoProv: no
+
+
+%description
+Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jade, coffee, css, less, ... and your custom stuff.
```
domcleal: Line wrapping's needed here, usually 72 characters.

domcleal:
Looks fine apart from inline comments and missing -doc from comps.

nodejs-webpack/nodejs-webpack.spec
```diff
@@ -0,0 +1,533 @@
...
+%prep
+mkdir npm_cache
+for tgz in %{sources}; do
+  echo $tgz | grep -q registry.npmjs.org || npm cache add --cache ./npm_cache $tgz
+done
+
+%setup -T -q -a 234 -D -n npm_cache
+
+%build
+npm install webpack@1.13.1 --cache-min Infinity --cache .  --verbose
```
domcleal: Prefer `%{npm_name}@%{version}` if possible, it'll reduce the changes required to update or between files.

dLobatog:
Repushed - @domcleal as far as I can see nonscl packages don't have -doc packages, should I add them still for these?

domcleal:
@dLobatog there are many, circa line 210 in comps/comps-foreman-rhel7.xml - non-SCL or not makes no difference.

dLobatog:
OK, I guess I was misguided by comps_doc which specifically checks for rubygem, also since mod_passenger & puppet-agent-oauth didn't have them I thought that was the standard. I'll add them to the rest as none of the nodejs packages have -doc in comps

domcleal:
Ah, the rubygem check in comps_doc.sh is interesting, that _probably_ shouldn't be there. We ought to ship doc packages for every spec file if they exist (and they do for all of these nodejs-* packages).

dLobatog:
Pushed with -doc comps, I updated the script in https://github.com/theforeman/foreman-packaging/pull/1243

domcleal:
Merged as d0646a5, thanks @dLobatog.

PR #1243, opened 2016-07-27
nodejs-stats-webpack-plugin

dLobatog:


nodejs-stats-webpack-plugin/nodejs-stats-webpack-plugin.spec
```diff
@@ -0,0 +1,52 @@
+%global npm_name stats-webpack-plugin
+
+Name: nodejs-%{npm_name}
+Version: 0.2.2
+Release: 1%{?dist}
+Summary: Write the stats of a build to a file
+License: undefined
```
domcleal: MIT

nodejs-stats-webpack-plugin/nodejs-stats-webpack-plugin.spec
```diff
@@ -0,0 +1,52 @@
...
+Group: Development/Libraries
+URL: undefined
```
domcleal: https://github.com/unindented/stats-webpack-plugin

nodejs-stats-webpack-plugin/nodejs-stats-webpack-plugin.spec
```diff
@@ -0,0 +1,52 @@
+%global npm_name stats-webpack-plugin
+
+Name: nodejs-%{npm_name}
+Version: 0.2.2
```
domcleal: Ideally up to 0.4.0, the latest (`^0.2.1`).

dLobatog:
Updated - this PR also contains an update to comps_doc to handle the nodejs packages

dLobatog:
Ah my bad, it failed because it still has the old cache. I'll update that

dLobatog:
Looks like the author didn't pack README.md in the tar... http://koji.katello.org/koji/getfile?taskID=508648&name=build.log - I've changed the %doc section now, it's an empty -doc package but in case in the future the tar includes docs, we can put them there.

domcleal:
Merged as 6a77248, thanks @dLobatog.

PR #1244, opened 2016-07-27
nodejs-bundle-babel-loader

dLobatog:


dLobatog:
Hm... this one is interesting because it fails since it has peerdependencies on babel-core and webpack. I'll see if I can make these 2 a dependency of this package and build them with mockchain

nodejs-bundle-babel-loader/nodejs-bundle-babel-loader.spec
```diff
@@ -0,0 +1,79 @@
...
+
+%prep
+mkdir npm_cache
+for tgz in %{sources}; do
+  echo $tgz | grep -q registry.npmjs.org || npm cache add --cache ./npm_cache $tgz
+done
+
+%setup -T -q -a 8 -D -n npm_cache
+
+%build
```
dLobatog: A better solution I found after asking around in #npm , this package should have a dependency on `nodejs-bundle-webpack`, and `nodejs-bundle-babel-core`. 

After these are installed, before running `npm install` we should run `npm link webpack` and `npm link babel-core`. That will link /usr/lib/node_modules/webpack into the current node_modules folder, hopefully allowing us to run `npm install` without installing any peerDependencies

dLobatog:
I made a few small changes now, it requires `npm(bundle-webpack)` and `npm(bundle-babel-core)`. It also runs `npm link webpack` and `npm link babel-core` so these are available.

nodejs-babel-loader/nodejs-babel-loader.spec
```diff
@@ -0,0 +1,84 @@
...
+
+%setup -T -q -a 8 -D -n npm_cache
+
+%build
+mkdir node_modules/
+# Notice I make regular symlinks instead of running npm link, as the latter will ask for
+# root permissions
+ln -s %{nodejs_sitelib}/webpack node_modules/webpack
+ln -s %{nodejs_sitelib}/babel-core node_modules/babel-core
+npm install babel-loader@6.2.4 --cache-min Infinity --cache .
```
domcleal: ditto, prefer `%{npm_name}@%{version}`

nodejs-babel-loader/nodejs-babel-loader.spec
```diff
@@ -0,0 +1,84 @@
...
+Source5: http://registry.npmjs.org/emojis-list/-/emojis-list-2.0.1.tgz
+Source6: http://registry.npmjs.org/minimist/-/minimist-0.0.8.tgz
+Source7: http://registry.npmjs.org/big.js/-/big.js-3.1.3.tgz
+Source8: babel-loader-6.2.4-registry.npmjs.org.tgz
+Requires: nodejs(engine)
+BuildRequires: nodejs-devel
+BuildRequires: nodejs-packaging
+BuildRequires: npm
+BuildRequires: npm(webpack)
+BuildRequires: npm(babel-core)
```
domcleal: Should these _also_ be `Requires` for runtime deps?

nodejs-babel-loader/nodejs-babel-loader.spec
```diff
@@ -0,0 +1,84 @@
...
+done
+
+%setup -T -q -a 8 -D -n npm_cache
+
+%build
+mkdir node_modules/
+# Notice I make regular symlinks instead of running npm link, as the latter will ask for
+# root permissions
+ln -s %{nodejs_sitelib}/webpack node_modules/webpack
+ln -s %{nodejs_sitelib}/babel-core node_modules/babel-core
```
domcleal: It may help to symlink each of `%{nodejs_sitelib}/*` which can always go into the template then.

dLobatog: You mean on all packages? It'll have no effect in most cases I guess (as long as there are no build deps on nodejs-xxxx packages...)

domcleal: Yep, that's what I mean - if it has no effect then you could include it in the template by default, if you wanted.

nodejs-babel-loader/nodejs-babel-loader.spec
dLobatog: I don't think so, they're never used in runtime. These are all `devDependencies` just needed to run `webpack compile`, which runs in the foreman.spec %build section. https://github.com/theforeman/foreman/pull/3433/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R7

domcleal: Not runtime of foreman (that's not relevant here), but runtime of this package I think they're required - they're peer dependencies of babel-loader.

dLobatog:
Updated - I did everything but adding the runtime Requires, which I don't think are needed

nodejs-babel-loader/nodejs-babel-loader.spec
dLobatog: I'll put them - but still it's unclear to me why is it necessary. babel-loader & this package will never run outside of the context of being a 'BuildDependency' of foreman - then why should they be regular Requires?

domcleal: It helps if we mirror the dependencies of what we're packaging, same with gems or any other package.. saves copying every dep between dependencies into the top-level `foreman` package, specifying here makes sure all dependencies are installed when this is.

domcleal: Perhaps to be clear - if nodejs-babel-loader is a BuildRequire of `foreman`, that doesn't mean that `BuildRequires` of nodejs-babel-loader are installed, it installs the _runtime_ `Requires`.

dLobatog:
Repushed with -doc comps

nodejs-babel-loader/nodejs-babel-loader.spec
dLobatog: I was under the wrong impression then, thanks. I thought BuildRequires for Foreman would pull their dependent BuildRequires just in case they are needed

domcleal: OK, could you add Requires then please?

dLobatog:
Pushed again - the only change is that I added npm webpack and npm babel-core as runtime requires

domcleal:
Merged as b01ea73, thanks @dLobatog.

PR #1245, opened 2016-07-27
nodejs-bundle-babel-core

dLobatog:


dLobatog:
This won't build in jenkins because the last source tgz with the npm cache isn't online

dLobatog:
Repushed with the npm cache file checked in as binary + changes to setup_sources

setup_sources.sh
```diff
@@ -8,6 +8,7 @@ for spec in $dir/*.spec; do
   for source in $(spectool --list-files $spec | awk '{print $2}'); do
     sourcebase=$(basename "$source")
     [ -h $d/$sourcebase ] || continue
+    if [[ $sourcebase =~ ^.*registry\.npmjs\.org\.tgz$ ]]; then continue; fi
```
domcleal: You probably won't need this, as the `-h` check above will skip over non-symlinks.

dLobatog:
I think we'll have to tweak npm a bit so it's truly online, the build failed because npm failed to `getaddrinfo EAI_AGAIN registry.npmjs.org:443` - http://koji.katello.org/koji/getfile?taskID=505830&name=build.log

dLobatog:
Just to be clear, I tried a few workarounds but none of them worked:

1 - pointing to localhost - failed
2 - running `node -e "var http = require('http'); var fs = require('fs'); http.createServer(function (req, res) { res.writeHead(304, {'Content-Type': 'text/plain'}); res.end(); }).listen(9615);"` and then pointing to `--registry http://localhost:9615` , same result

dLobatog:
This took me a great deal of time to figure out by putting breakpoints on npm itself - but I think I finally cracked it, npm is expecting a `npm_cache/$registryname/` dir with the `.cache.json` files, it's not OK if we just dump the .cache.json files in the general `npm_cache/packagename/.cache.json`

domcleal:
I replied on the other thread: https://github.com/theforeman/foreman-packaging/pull/1246#issuecomment-235877288 has more information, likely version differences.

nodejs-babel-core/nodejs-babel-core.spec
```diff
@@ -0,0 +1,181 @@
...
+%prep
+mkdir npm_cache
+for tgz in %{sources}; do
+  echo $tgz | grep -q registry.npmjs.org || npm cache add --cache ./npm_cache $tgz
+done
+
+%setup -T -q -a 61 -D -n npm_cache
+
+%build
+npm install babel-core@6.7.7 --cache-min 9999999 --cache .
```
domcleal: ditto, prefer npm_name/version

domcleal: Also, other packages use Infinity here - might want to standardise based on whatever's in the final template?

dLobatog:
Repushed with %{npm_name}@{version} and --cache-min Infinity

domcleal:
Merged as 8be6e4e, thanks @dLobatog.

PR #1246, opened 2016-07-27
nodejs-bundle-babel-preset-es2015

dLobatog:


dLobatog:
This won't build in Jenkins because  the source with the npm cache isn't online yet

nodejs-bundle-babel-preset-es2015/babel-preset-es2015-6.6.0-registry.npmjs.org.tgz
```diff
@@ -0,0 +1 @@
+../.git/annex/objects/97/70/SHA256E-s314319--50075c5fa4399737aff808d3297e184a59e1dfd3422c05c4286ef6aa4d0e2f6d.org.tgz/SHA256E-s314319--50075c5fa4399737aff808d3297e184a59e1dfd3422c05c4286ef6aa4d0e2f6d.org.tgz
```
domcleal: If the file's not too big, add this as a regular file instead of via git-annex.

dLobatog: They're usually under 1 megabyte

domcleal: I think we can live with that for now, given all of the other sources are coming from the NPM site, it's a good reduction. I'll have a think to see if it can be improved in future.

dLobatog:
Pushed this with the cached npm tar as a binary, let's see if it builds in Jenkins

dLobatog:
Looks like ./setup_sources.sh needs a change so that this works

domcleal:
The file is still a symlink. It's a little unreliable to change it back - I usually do a git rm, commit, replace the file, git add, then add a second commit _rather than amending_. You can then squash the commits after.

dLobatog:
I'll do that, I've already changed setup_sources.sh though, it will try to 'git annex addurl' for an URL that doesn't exist

dLobatog:
Pushed w/o changes to setup_sources, the file isn't a symlink indeed

dLobatog:
Failed for the same reason as babel-core, npm install with cache isn't truly offline https://github.com/theforeman/foreman-packaging/pull/1245#issuecomment-235825962

dLobatog:
Let me try to do `npm install` with `--registry localhost`? I don't see why would it need to connect to npmjs itself, but at least that doesn't seem to throw errors here when offline.

dLobatog:
No dice, for some reason it still tries to connect and fails - `npm ERR! fetch failed http://registry.npmjs.org/babel-preset-es2015/-/babel-preset-es2015-6.9.0.tgz`

domcleal:
I've made some further changes in https://github.com/domcleal/foreman-packaging/commits/rpm/nodejs-bundle-babel-preset-es2015 that appear to work (http://koji.katello.org/koji/taskinfo?taskID=505849).

1. Regenerated the tarball on EL7's npm 0.10.x, seems to be forward-compatible if not backwards-compatible. (So generate on the oldest version.)  This required me to change generate_npm_tarball.sh back to searching all of `$wd/cache` for .cache files.
1. Removed the `--registry` argument (unsure if this is needed, might be embedded in the cache?).
1. Added the version number to the npm install command (`@%{version}`) so it installs the specific cached version instead of trying to check for the latest.
1. Regenerated the tarball again, but this time specifying `babel-preset-es2015@6.6.0` specifically, so the the version matches what the spec installs.
1. Updated the freshly released `babel-traverse` package to match the newly regenerated registry data.

dLobatog:
Interesting - I tried this and it also seemed to work, no changes. The issue is that npm looked for cache in a different place than where it used to be extracted

```
#!/bin/bash

mkdir npm_cache
for tgz in $(ls *.tgz); do
  echo $tgz | grep -q registry.npmjs.org || npm cache add --cache ./npm_cache $tgz
done

cd npm_cache
mkdir registry.npmjs.org
cd registry.npmjs.org
gzip -dc ../babel-preset-es2015-6.6.0-registry.npmjs.org.tgz | tar -xof -
cd ../
npm install babel-preset-es2015 --verbose --cache-min Infinity --cache .
```

dLobatog:
Alright, now it builds fine on f24 - http://koji.katello.org/koji/getfile?taskID=505868&name=build.log 
el7 doesn't work, most likely because npm 1.x works different to 2.x (f24) - http://koji.katello.org/koji/getfile?taskID=505867&name=build.log 

I'll check if creating the registry cache tarball with npm 1.3.6 makes a difference

dLobatog:
@dominic Are you able to generate the tarball for this repo on el7s NPM (1.3.6)? I get this error https://gist.github.com/7a968ae71a61ef6ed63ce270dbdab7a0

domcleal:
@dLobatog yes, it works fine for me, it resolves `babel-plugin-transform-es2015-duplicate-keys` correctly.  Perhaps it was an issue at the registry or the packages were being updated at the time.  There's also a tarball I did yesterday on the branch mentioned above.

dLobatog:
@domcleal Nevermind, since I wrote that comment I tried on a CentOS 7 host and it works there. 

I used `npm install npm@1.3.6` locally (f24), then `node_modules/.bin/npm` in the script to generate the tarball and I got that error.

It's annoying having to ssh & scp everytime I do this though, so I'll figure out what's the difference - also so we can put this in a script that does the whole thing (generate the spec + registry tarball + sources)

dLobatog:
Repushed with the registry tarball generated on CentOS 7, let's see how that goes in koji. If it succeeds I'll continue with the other bundle packages.

dLobatog:
Oh that's annoying... now f24 fails and el7 succeeds:

el7: http://koji.katello.org/koji/taskinfo?taskID=506123
f24: http://koji.katello.org/koji/taskinfo?taskID=506124

It's again because npm in f24 (2.15.8) expects /registry.npmjs.org to contain the cached registry. In order to avoid checking in 2 files depending on the OS, I'll just put both caches in the same tgz.

dLobatog:
Success :smile: I'll do the same with the other bundle packages

nodejs-babel-preset-es2015/nodejs-babel-preset-es2015.spec
```diff
@@ -0,0 +1,250 @@
...
+%prep
+mkdir npm_cache
+for tgz in %{sources}; do
+  echo $tgz | grep -q registry.npmjs.org || npm cache add --cache ./npm_cache $tgz
+done
+
+%setup -T -q -a 93 -D -n npm_cache
+
+%build
+npm install babel-preset-es2015@6.6.0 --cache-min Infinity --cache . --verbose
```
domcleal: ditto, npm_name/version?

dLobatog:
Repushed with npm_name@version - I named the packages originally 'nodejs-bundle-xxxxx' hence I had to use the actual npm names instead of the variable, but that's not necessary anymore.

domcleal:
Merged as b68edb8, thanks @dLobatog.

PR #1247, opened 2016-07-27
nodejs-expose-loader 0.6.0

dLobatog:
Aaaand this seems to be the last dependency, with all of the current PRs webpack:compile runs fine.

next up - PR to foreman.spec with the appropriate BuildRequires & ensure it works with mockchain


nodejs-expose-loader/nodejs-expose-loader.spec
```diff
@@ -0,0 +1,51 @@
+%global npm_name expose-loader
+
+Name: nodejs-%{npm_name}
+Version: 0.6.0
+Release: 1%{?dist}
+Summary: expose loader module for webpack
+License: undefined
```
domcleal: MIT

dLobatog:
Pushed with the fix for License, my bad, I guess npm2rpm has a bug

domcleal:
Merged as 56cd54a, thanks @dLobatog.

PR #1250, opened 2016-07-29
refs #2687 - recommend rb-inotify for DHCP ISC module [deb]

domcleal:


mmoll:
merged, thanks @domcleal!

PR #1261, opened 2016-08-04
Webpack changes in Foreman

dLobatog:
This commit removes unnecessary libraries with webpack, namely,
jquery-rails and underscore-rails. It also adds the necessary
BuildRequires to run webpack compile on the Foreman %build section.

This won't build in Jenkins until https://github.com/theforeman/foreman/pull/3433 is merged in core, so please build your own tar to test it.


foreman/foreman.spec
```diff
@@ -196,22 +198,45 @@ BuildRequires: %{?scl_prefix}rubygem(roadie-rails) >= 1.1
...
+BuildRequires: npm(jquery-ujs) < 1.3.0
+BuildRequires: npm(jquery.cookie) >= 1.4.1
+BuildRequires: npm(jquery.cookie) < 1.5.0
+BuildRequires: npm(jstz) >= 1.0.7
+BuildRequires: npm(jstz) < 1.1.0
+BuildRequires: npm(lodash) >= 2.4.1
+BuildRequires: npm(lodash) < 2.5.0
+BuildRequires: npm(select2) >= 3.5.2
+BuildRequires: npm(select2) < 3.6.0
+BuildRequires: libuv
```
domcleal: What's libuv required for?

foreman/foreman.spec
```diff
@@ -577,6 +594,8 @@ sed -i 's/:locations_enabled: false/:locations_enabled: true/' config/settings.y
 sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml
 export BUNDLER_EXT_NOSTRICT=1
 export BUNDLER_EXT_GROUPS="default assets"
+ln -s /usr/lib/node_modules/ .
```
domcleal: exporting NODE_PATH=%{node_sitedir} (and add node-packaging) on the line below might be preferable?

domcleal:
1. The two package removals should also be added to tfm/tfm.spec in the Obsoletes section so they're cleaned up on upgrade.
2. The new npm(..) packages should probably be added to the Requires list of foreman-assets too.

foreman/foreman.spec
dLobatog: Yeah, I wasn't sure if we should add node-packaging hence I went with that

foreman/foreman.spec
dLobatog: It's necessary in order to run nodejs without requiring nodejs-packaging - shouldn't be after I add it as per your other comment

dLobatog:
Repushed with Obsoletes in tfm.spec, nodejs-packaging dep on foreman.spec, it also now exports NODE_PATH and NODE_ENV prior to run webpack - I have not tested this thoroughly but we'll see if it passes once the rest of the packages are merged anyway

domcleal:
The new npm(..) packages should probably be added to the Requires list of foreman-assets too?

dLobatog:
Ah sure, only the runtime ones, dev dependencies (like webpack stats plugin, babel...) that are only required to run webpack shouldn't be regular requires, right?

domcleal:
I'd have thought all of them - enough to permit `webpack:compile` to be run or for other packages to use webpack?

domcleal:
Could you restore the dependency on libuv please? It seems to be necessary due to a broken nodejs010-libuv package in the SCL repositories - the wrong libuv is being installed. Please just add a comment pointing to https://bugs.centos.org/view.php?id=10606.

foreman/foreman.spec
```diff
@@ -577,6 +618,9 @@ sed -i 's/:locations_enabled: false/:locations_enabled: true/' config/settings.y
 sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml
 export BUNDLER_EXT_NOSTRICT=1
 export BUNDLER_EXT_GROUPS="default assets"
+export NODE_PATH=%{nodejs_sitelib}
```
domcleal: Sorry, this doesn't seem to work, webpack is unable to find modules:

<pre>
+ export NODE_PATH=/usr/lib/node_modules
+ NODE_PATH=/usr/lib/node_modules
+ export NODE_ENV=production
+ NODE_ENV=production
+ webpack.js --bail --config config/webpack.config.js
EntryModuleNotFoundError: Entry module not found: Error: Cannot resolve module 'babel-loader' in /builddir/build/BUILD/foreman-1.13.0
</pre>

Going back to nearly the original code, this does work:

<pre>
ln -s %{nodejs_sitelib} node_modules
</pre>

foreman/foreman.spec
```diff
@@ -196,22 +198,45 @@ BuildRequires: %{?scl_prefix}rubygem(x-editable-rails) >= 1.5.5
...
+BuildRequires: npm(webpack) < 2.0.0
+BuildRequires: npm(jquery) >= 1.11.0
+BuildRequires: npm(jquery) < 1.12.0
+BuildRequires: npm(jquery-ujs) >= 1.2.0
+BuildRequires: npm(jquery-ujs) < 1.3.0
+BuildRequires: npm(jquery.cookie) >= 1.4.1
+BuildRequires: npm(jquery.cookie) < 1.5.0
+BuildRequires: npm(jstz) >= 1.0.7
+BuildRequires: npm(jstz) < 1.1.0
+BuildRequires: npm(lodash) >= 2.4.1
```
domcleal: This _minimum_ requirement should be wrapped in a `%if 0%{scl:1}` conditional, and ditto on the assets Require, for use on Fedora 24.

F24 contains npm(lodash) 3.x, but since this ships an old version then removing the minimum dependency will ensure the correct version gets installed (dnf/yum will try to prefer 3.x as it matches >= 2.4.1).

<pre>
%if 0%{?scl:1}
BuildRequires: npm(lodash) >= 2.4.1
%endif
BuildRequires: npm(lodash) < 2.5.0
</pre>

dLobatog: I suppose I'll have to remove the package from f24 props/comps too?

domcleal: No, you can leave the old package in - just that removing this min. dependency on F24 will ensure it selects the older one we supply. lodash should be updated in Foreman though to 3.x.

dLobatog:
Updated and changed NODE_PATH by `ln -s` and also put in the `if scl` condition for lodash

foreman/foreman.spec
```diff
@@ -412,16 +440,41 @@ Requires: %{name} = %{version}-%{release}
...
+Requires: npm(jquery) < 1.12.0
+Requires: npm(jquery-ujs) >= 1.2.0
+Requires: npm(jquery-ujs) < 1.3.0
+Requires: npm(jquery.cookie) >= 1.4.1
+Requires: npm(jquery.cookie) < 1.5.0
+Requires: npm(jstz) >= 1.0.7
+Requires: npm(jstz) < 1.1.0
+%if 0%{?scl:1}
+Requires: npm(lodash) >= 2.4.1
+%%endif
```
domcleal: Double % is probably causing the test failure.

foreman/foreman.spec
```diff
@@ -196,22 +198,50 @@ BuildRequires: %{?scl_prefix}rubygem(x-editable-rails) >= 1.5.5
...
+BuildRequires: npm(jquery) < 1.12.0
+BuildRequires: npm(jquery-ujs) >= 1.2.0
+BuildRequires: npm(jquery-ujs) < 1.3.0
+BuildRequires: npm(jquery.cookie) >= 1.4.1
+BuildRequires: npm(jquery.cookie) < 1.5.0
+BuildRequires: npm(jstz) >= 1.0.7
+BuildRequires: npm(jstz) < 1.1.0
+%if 0%{?scl:1}
+BuildRequires: npm(lodash) >= 2.4.1
+%%endif
```
domcleal: ditto, double %

dLobatog: Sorry, vim's trying to be smarter than it really can...

PR #1265, opened 2016-08-15
Update foreman-tasks to 0.8.0 (RPM)

daviddavis:

* [x] Nightly
* [ ] 1.12
* [ ] 1.11
* [ ] 1.10

---


domcleal:
a) rubygem-foreman_salt, rubygem-foreman_chef and rubygem-foreman_remote_execution all have dependencies on < 0.8.0, which cannot be satisfied if this is updated. They will need to be updated first.
b) Missing change in `dynflow from: ["~> 0.8.8"] to: ["~> 0.8.13"]`

stbenjam:
> a) rubygem-foreman_salt, rubygem-foreman_chef and rubygem-foreman_remote_execution all have dependencies on < 0.8.0, which cannot be satisfied if this is updated. They will need to be updated first. (That is, the plugins need to have gemspecs updated and released, not spec file changes.)

Can you clarify why? bundle install would just pull the older version of the gem until they're updated to support the new tasks.

domcleal:
The repository tests (repoclosure) will fail  if 0.8.0 is added to the repo, as the package dependencies (`< 0.8.0`) won't be satisfied.  `bundle install` isn't used to install packages, only used on source installations.

stbenjam:
Oh I see, Koji nightly repo only hosts 1 version, there's no way to have multiple copies?  Anyway, why do only new gems need releasing and not updating the rpm spec? Your comment here isn't clear to me:

> (That is, the plugins need to have gemspecs updated and released, not spec file changes.)

domcleal:
> Oh I see, Koji nightly repo only hosts 1 version, there's no way to have multiple?

No, only the latest will ever be kept in the mashed repo. There are older versions on the web server, but that's further down the line, it'll still break tests and package builds.

> Anyway, why do only new gems need releasing and not updating the rpm spec? Your comment here isn't clear to me:
> 
>     (That is, the plugins need to have gemspecs updated and released, not spec file changes.)

If a version of a package is installed that the gemspec says is not permitted, then there will usually be a runtime error as the dep can't be satisfied.

By keeping the RPM dependencies matching the gemspec then these errors are caught when building the repo (or during PRs like this) instead of by users.

stbenjam:
I still don't get it, you need the gem requirement bumped as well as the RPM spec file, why would you not change the rpm spec?

domcleal:
> I still don't get it, you need the gem requirement bumped as well as the RPM spec file, why would you not change the rpm spec?

I don't think I understand the question. The RPM spec gets changed to match the gem dependencies when the package/gem is updated to the new version. I'm saying that the RPM dependencies wouldn't be changed without also updating the gem.

stbenjam:
You said this:

> (That is, the plugins need to have gemspecs updated and released, not spec file changes.)

That to mean means "update the gem but don't bother with the spec files"

But that's not what you meant.

domcleal:
No, I meant don't _only_ change the foreman-tasks dependencies in spec files.

stbenjam:
I see.  Is there a reason not to set the nightly repo on fedorapeople as a lookaside? It seems odd our build infrastructure works this way.

domcleal:
Koji works the same way when it creates buildroots, so it's of limited use doing that.

stbenjam:
I released foreman_salt 7 for use with foreman 1.13(DEB: https://github.com/theforeman/foreman-packaging/pull/1269, RPM: https://github.com/theforeman/foreman-packaging/pull/1268)

iNecas:
rex packages for RPM https://github.com/theforeman/foreman-packaging/pull/1270

stbenjam:
chef https://github.com/theforeman/foreman-packaging/pull/1274

daviddavis:
[test]

domcleal:
Merged and building, thanks @daviddavis.

PR #1266, opened 2016-08-18
 Update foreman-tasks to 0.8.0 (DEB)

mmoll:
this should be built into:
* [x] Nightly


domcleal:
Per #1265, I think this depends on having new foreman_salt, remote_execution and chef releases first.

mmoll:
@domcleal I also thought that. However, as of now, these plugins don't work with current Foreman nightlies anyway and it's easier to test build their respective DEB PRs, once updated and there are no repoclosure checks on the apt repository anyway, so I think it's better to have a working foreman-tasks package in the nightly repo now and if anybody is installing any of the mentioned plugins, the apt/deb install process would even fail "correctly".

domcleal:
Merged, thanks @mmoll.

Building at http://ci.theforeman.org/job/packaging_build_deb_plugin/1159/

PR #1283, opened 2016-08-22
Revert el6

ehelms:

* [x] Nightly
* [ ] 1.12
* [ ] 1.11
* [ ] 1.10

---


ehelms:
@domcleal I know you aren't a big fan of this revert, and that this, given our current workflows, would mean it falls on you to maintain and build out the pieces that are now missing from EL6. However, I am willing to take on the work that needs doing (and if you prefer) the maintenance there after if thats what it takes to get one more release with EL6 support for our users.

dLobatog:
@ehelms This would require:

* likely updating the registry caches for the nodejs packages - https://github.com/dLobatog/npm2rpm/blob/master/bin/generate_npm_tarball.sh
* adding back any 'if' conditions that were there for el6 
will update this list when I can think of more stuff

domcleal:
> given our current workflows, would mean it falls on you to maintain and build out the pieces that are now missing from EL6

No it doesn't, any maintainer could. I certainly won't be building/maintaining packages for EL6 that I've already removed.

ehelms:
@domcleal Does that mean that if I push this and update specs for EL6 (aka maintain the EL6 side of things), that when you do Foreman and plugins releases you'll release them to the EL6 tags etc. as part of the workflow? I'm trying to find a compromise here that allows us to reduce overhead and keep EL6 for one release for our users.

domcleal:
If it's in the main branch then I'd release to it, yes.

domcleal:
Just a reminder that if you're serious about this, then there's very little time before 1.13.0-RC1 is ready to build and test all of the outstanding packages in nightly before duplicating it to 1.13.

ehelms:
@domcleal appreciate the reminder, @dLobatog and myself are working to build all the packages and test them. So far, so good. What artifacts or resources would you need to feel comfortable having this merged?

ehelms:
I ran a full build with mockchain and was able to successfully install EL6 from my builds. I added a commit that updates rubygem-rb-inotify spec file for EL6.

dLobatog:
@ehelms These two packages (rubygem-rkerberos and rubygem-builder) were also removed with EL6 - f4338d0d03fcdc96b9c0938b21ab33f3985bdd02 and they're missing here

ehelms:
@dLobatog those packages have been added back to the commit stack here via a revert (in my testing, I just used previously built versions in Koji given they hadn't changed)

ehelms:
Clicking around and playing with the UI I saw no issues either.

ehelms:
Can I get a little help understanding what the next steps in y'alls eyes to progress this are? I'd like to get it situated and merged as soon as possible due to the RC1 deadline but without feedback to help me I don't know what is expected. Unless y'all are satisfied and I can merge and start re-updating packages.

comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -28,6 +28,7 @@
       <packagereq type="default">tfm-rubygem-foreman_hooks</packagereq>
       <packagereq type="default">tfm-rubygem-foreman_host_extra_validator</packagereq>
       <packagereq type="default">tfm-rubygem-foreman_host_rundeck</packagereq>
+      <packagereq type="default">tfm-rubygem-foreman-mco</packagereq>
```
dLobatog: Unrelated?

dLobatog:
@ehelms Here are some (all) todos I found:

* All references to rubygem-select2 , nodejs-select2 substitutes it, add it for rhel6
* All references to rubygem-ipaddrjs-rails , nodejs-ipaddr.js substitutes it, add it for rhel6
* foreman-mco from comps, it's removed and unmaintained
* Add foreman_monitoring and smart_proxy_monitoring to rhel6
* Add nodejs-es6-promise to rhel6
* Add nodejs-extract-text-webpack 
* Add nodejs-style-loader
* Add nodejs-css-loader
* Add nodejs-file-loader
* Add nodejs-url-loader
* Add rubygem-infoblox 
* Add rubygem-multipart-post to comps/comps-foreman-plugins-rhel6.xml (keep it in regular foreman-rhel6 comps too, like el7). Add it to nonscl in rhel6 tito.props
* Add a nonscl entry for rubygem-faraday and farady_middleware to comps/comps-foreman-plugins-rhel6.xml (check out el7, it's the same) Add it to nonscl in foreman-plugins-nightly-nonscl-rhel6 in tito.props.
* Add nodejs-diff
* Add rubygem-useragent

After all of that, it'll build everywhere but apparently Koji doesn't have the foreman-nonscl-rhel6 tag so if you can add it back, it'd be helpful. Should be ready to merge at that point - I'll build it locally but if you can add the tag back that'd help so we can see it green here (or @domcleal ).

dLobatog:
From what I see in Jenkins, and in http://ci.theforeman.org/view/Release%20pipeline/ , all release jobs preserved el6 and it was not removed? 

http://ci.theforeman.org/view/Release%20pipeline/job/release_push_rpm/configure preserves el6, I'll check the others but it seems like el6 was only removed from testing nightly https://github.com/theforeman/foreman-infra/commit/d392dadc212641be872b819d0407e1f2cf696af2 which I guess we don't want to recover

ehelms:
Think I should re-open this against `rpm/1.13` so that we are only adding it in for 1.13 and keeping El6 out of 1.14 and beyond?

dLobatog:
@ehelms To be honest, I don't know, but it seems like for that (making el6 only for 1.13)  we would need a serious effort to create a tool to backup and migrate to el7 in ~3 months that as far as I'm aware nobody is doing or even planning to do it

ehelms:
@dLobatog I think that is the goal and the resolution we arrived at was to support EL6 for one last release at the start of the release (1.13) and start communicating to users and building any tools or docs to help users migrate to EL7 for 1.14 and beyond.

ehelms:
I went ahead and opened https://github.com/theforeman/foreman-packaging/pull/1344 just in case. My thinking is that if we add all this to nightly, we then have to port it to 1.13 branch anyway and then we have to go and remove all this from nightly so we might as well just make it work and build it for 1.13.

ehelms:
[test]

ehelms:
[test]

ehelms:
@dLobatog I've updated this per our discussion, and built nearly everything into the EL6 tags. The only thing unbuilt are a few newer packages that required spec updates and will thus require a new release:

 * rubygem-useragent
 * rubygem-rb-inotify

There is a test repo here -- http://koji.katello.org/releases/yum/foreman-nightly/RHEL/6/x86_64/ but will fail on the installer run due to the above packages missing.

dLobatog:
Good to merge now IMO but some of the packages here will never be in stable repos so really I don't see the point vs merging just the other into rpm/1.13 - I'm fine with both PRs, this one will give us nightlies at least and we can cp things to 1.13 I suppose which is the normal route

domcleal:
The foreman package hasn't been building successfully since this was merged, and there are other packages built (e.g. foreman-proxy) that are incompatible with EL6. No tests have been running either.

Please fix these issues or revert the changes here before rpm/1.14 is branched.

PR #1288, opened 2016-08-26
require puppet >= 3.6.0 for the installer (DEB)

mmoll:
this should be built into:
* [x] Nightly

following theforeman/puppet-foreman@2ad07e7


mmoll:
[test]

domcleal:
The branch name has a typo, it should be "deb/", not "dev/". Could you re-open this please, just so we can check the test's OK?

PR #1289, opened 2016-08-26
require puppet >= 3.6.0 for the installer (RPM)

mmoll:
this should be built into:

* [x] Nightly

following theforeman/puppet-foreman@2ad07e7


domcleal:
Thanks @mmoll!

PR #1303, opened 2016-09-01
require kafo >= 0.9.5 for the installer (DEB)

mmoll:
following theforeman/foreman-installer@c5b504c

this should be built into:
* [x] Nightly


domcleal:
Merged, thanks.  (Tests only failing as the new release isn't in the main repos yet.)

PR #1306, opened 2016-09-02
Update lodash to 4.15.0

dLobatog:
Needed for https://github.com/theforeman/foreman/pull/3796

I removed the 'if scl's on the Foreman because the nodejs-lodash available in Fedora is not 4.15


nodejs-lodash/nodejs-lodash.spec
```diff
@@ -29,22 +29,18 @@ This package contains documentation for nodejs-%{npm_name}
...
-
 rm -rf node_modules
 
 %build
 #nothing to do
 
 %install
 mkdir -p %{buildroot}%{nodejs_sitelib}/%{npm_name}
-cp -pfr dist  *.json *.md *.js *.txt %{buildroot}%{nodejs_sitelib}/%{npm_name}
+cp -pfr *.json *.md *.js %{buildroot}%{nodejs_sitelib}/%{npm_name}
```
domcleal: missing `fp` and `LICENSE`?

nodejs-lodash/nodejs-lodash.spec
```diff
@@ -29,22 +29,18 @@ This package contains documentation for nodejs-%{npm_name}
...
 %nodejs_symlink_deps
 
-%check
-
 %files
 %{nodejs_sitelib}/%{npm_name}
-%doc LICENSE.txt
```
domcleal: `LICENSE` is still present I think, just the extension's removed

dLobatog:
Updated to include LICENSE and fp (npm2rpm spec generated a weird spec that's why I missed them)

domcleal:

PR #1309, opened 2016-09-05
add_npm_package script

dLobatog:
A convenience to run when creating npm packages. It does everything but
changing the comps (the current format makes it hard to add a new line
alphabetically).

I think that comps, a better spec template, and creating a proper cache for both el7/f24 through npm2rpm (not sure how to tackle that one yet...) is all that's left to make adding packages nearly a 1-step thing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,8 @@ You'll also need an alias `kojikat` to point to:
 1. Ensure the source file (e.g. the .gem) is in the spec directory and run
    `git annex add foo.gem`
 1. Update rel-eng/tito.props and add to the appropriate whitelists
-1. Update comps/comps-foreman-\*.xml
-2. Run ./comps_doc.sh to automatically add docs
+1. Update comps/comps-foreman-\*.xml by running `./add_to_comps.rb comps/compsfile.xml rubygem-example`
+2. Run `./comps_doc.sh` to automatically add doc packages
 1. Commit the changes
   1. `git add -A`
   1. `git commit -m "Add NAME package"`
@@ -143,29 +143,17 @@ Otherwise we will bundle the dependencies. You can read an explanation of how we
 
 In both cases:
 
-1. Install npm2rpm: `npm install -g npm2rpm`
-1. Checkout a branch 'rpm/example-version' or similar (the rpm/ part is important)
-  * `mkdir nodejs-example`
-  * `cd nodejs-example`
 1. Generate the spec and download the sources automatically with npm2rpm.
-  * For packages without dependencies - `npm2rpm -n example -v version -s single`
-  * For packages that bundle dependencies - `npm2rpm -n example -v version -s bundle`
-1. This should have created two directories npm2rpm/SPECS and npm2rpm/SOURCES.
-   Move the content of them both to your nodejs-example directory and remove the npm2rpm folder.
-  * mv npm2rpm/SPECS/\* .
-  * mv npm2rpm/SOURCES/\* .
-  * rm -r npm2rpm
-1. If building a package with bundled dependencies, add the binary with the npm cache directly to git.
-  * `git add example-version-registry.npmjs.org.tgz`
-1. `git-annex` the rest of the sources
-  * `git annex add *.tgz`
-1. Update rel-eng/tito.props and add to the whitelists (nonscl and Fedora)
-1. Update comps/comps-foreman-\*.xml
+  * For packages without dependencies - `./add_npm_package.sh example version single`
+  * For packages that bundle dependencies - `./add_npm_package.sh example version bundle`
+1. This should have created a nodejs-example directory with the packages needed, the spec,
+and the cache if it's a bundled package. It should have modified tito.props, comps and
+added everything to git.
+1. Update the spec if needed (e.g: for peer dependencies)
 1. Commit the changes
-  1. `git add -A`
   1. `git commit -m "Add NAME package"`
-1. Follow the "test a package" section above until it builds for all
-   targeted platforms and required non-SCL modes.
+1. Follow the "test a package" section above until it builds for all targeted platforms and
+required non-SCL modes.
 1. Submit a pull request against `rpm/develop`
 
 ## HOWTO: update a package
--- a/add_npm_package.sh
+++ b/add_npm_package.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+display_usage() {
+  echo "This script adds a new npm package based on the module found on npmjs.org"
+  echo -e "\nUsage:\n$0 NPM_MODULE_NAME VERSION STRATEGY \n"
+  exit 1
+}
+
+if [ $# -ne 3 ]; then
+  display_usage
+fi
+
+program_exists() {
+  which "$@" &> /dev/null;
+}
+
+if !(program_exists crudini); then
+  echo "crudini is not installed - you can install it with"
+  echo "sudo yum install crudini"
+  exit 1
+fi
+
+if !(program_exists npm2rpm); then
+  echo "npm2rpm is not installed - you can install it with:"
+  echo "sudo npm install npm2rpm"
+  exit 1
+fi
+
+echo -n "Making directory..."
+mkdir nodejs-$1
+echo "FINISHED"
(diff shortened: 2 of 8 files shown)


dLobatog:
Updated the README. With the latest build of npm2rpm, I think all that's needed to create a package is to run this and pdate the comps (& probably the spec, but that will be less and less necessary as npm2rpm improves)

add_npm_package.sh
```diff
@@ -0,0 +1,54 @@
...
+echo "FINISHED"
+echo -n "Creating specs and downloading sources..."
+npm2rpm -n $1 -v $2 -s $3
+echo "FINISHED"
+echo -n "Copying specs..."
+cp npm2rpm/SPECS/* nodejs-$1
+echo "FINISHED"
+echo -n "Copying sources..."
+cp npm2rpm/SOURCES/* nodejs-$1
+echo "FINISHED"
```
dLobatog: It should remove npm2rpm/ at this point

dLobatog:
Updated to remove npm2rpm after the sources and specs are copied

add_npm_package.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+# Run as ./add_npm_package.sh name version strategy
+program_exists() {
+  which "$@" &> /dev/null;
+}
+
+if !(program_exists crudini); then
+  echo "crudini is not installed - installing..."
+  sudo yum install crudini
```
domcleal: Could you just print the command out instead of running sudo and yum? (Note that Fedora uses `dnf` by default nowadays too.)

add_npm_package.sh
```diff
@@ -0,0 +1,55 @@
...
+  echo "crudini installed..."
+fi
+
+if !(program_exists npm2rpm); then
+  echo "npm2rpm is not installed - installing..."
+  sudo npm install -g npm2rpm
```
domcleal: ditto, or preferably use a non-global installation and add node_modules to .gitignore

add_npm_package.sh
```diff
@@ -0,0 +1,55 @@
...
+cp npm2rpm/SPECS/* nodejs-$1
+echo "FINISHED"
+echo -n "Copying sources..."
+cp npm2rpm/SOURCES/* nodejs-$1
+echo "FINISHED"
+rm -r npm2rpm
+echo -n "Setting tito props..."
+# Get tito.props whitelists and add node package
+el7whitelist=$(crudini --get rel-eng/tito.props foreman-nightly-nonscl-rhel7 whitelist)
+el7whitelist=$(echo "$el7whitelist nodejs-$1" | tr " " "\n" | sort)
```
domcleal: set LC_COLLATE when sorting as different people's locales will change the list order (e.g. comps_doc.sh uses en_GB)

add_npm_package.sh
```diff
@@ -0,0 +1,55 @@
...
+  echo "FINISHED"
+fi
+echo -e "Annexing sources... - "
+git annex add *.tgz
+echo "FINISHED"
+echo -e "Adding other files to git... - "
+git add *.spec
+git add ../rel-eng/tito.props
+echo "FINISHED"
+# TODO: add to comps (do it through a ruby script?)
```
domcleal: `xmlstarlet` could probably be used for this

add_npm_package.sh
```diff
@@ -0,0 +1,55 @@
...
+  echo "npm2rpm is not installed - installing..."
+  sudo npm install -g npm2rpm
+  echo "npm2rpm installed..."
+fi
+
+echo -n "Making directory..."
+mkdir nodejs-$1
+echo "FINISHED"
+echo -n "Creating specs and downloading sources..."
+npm2rpm -n $1 -v $2 -s $3
```
domcleal: Add a check at the top of the script for usage, `if [ $# -ne 3 ]; then` echo usage out.

add_npm_package.sh
dLobatog: You mean print the command to install it & exit? It already prints out the output of yum

domcleal: Yes, print & exit.

README.md
```diff
@@ -143,29 +143,21 @@ Otherwise we will bundle the dependencies. You can read an explanation of how we
 
 In both cases:
 
-1. Install npm2rpm: `npm install -g npm2rpm`
 1. Checkout a branch 'rpm/example-version' or similar (the rpm/ part is important)
```
tbrisker: Is this step still needed? It seems the script already creates the directory

dLobatog: Actually not, thanks :smile:

add_npm_package.sh
dLobatog: fixed

add_npm_package.sh
dLobatog: fixed

dLobatog:
Updated - now `add_to_comps.rb` adds comps too. IMO this automates the entire thing, my plan is to add better templates support to npm2rpm so there are less fixes to correct. 

cc @domcleal @tbrisker

README.md
```diff
@@ -123,8 +123,8 @@ You'll also need an alias `kojikat` to point to:
 1. Ensure the source file (e.g. the .gem) is in the spec directory and run
    `git annex add foo.gem`
 1. Update rel-eng/tito.props and add to the appropriate whitelists
-1. Update comps/comps-foreman-\*.xml
-2. Run ./comps_doc.sh to automatically add docs
+1. Update comps/comps-foreman-\*.xml by running `./add_to_comps.rb rubygem-example comps/compsfile.xml`
```
domcleal: Wrong argument order? It takes comps file, then package.

add_npm_package.sh
```diff
@@ -0,0 +1,73 @@
...
+echo -e "Annexing sources... - "
+git annex add *.tgz
+echo "FINISHED"
+echo -e "Adding spec to git... - "
+git add *.spec
+echo "FINISHED"
+echo -e "Updating comps... - "
+cd ..
+./add_to_comps.rb nodejs-$1 comps/comps-foreman-fedora24.xml
+./add_to_comps.rb nodejs-$1 comps/comps-foreman-rhel7.xml nonscl
```
domcleal: Wrong argument order? Filename then package name.

add_to_comps.rb
```diff
@@ -0,0 +1,65 @@
...
+  nonscl_comment
+end
+
+def prepend_packagereqs(packagelist, packagereqs)
+  packagereqs.each do |packagereq|
+    packagelist.children.before(packagereq)
+  end
+end
+
+comps = Nokogiri::XML(File.open(ARGV[0])) { |config| config.noblanks }
```
domcleal: missing ARGV checks + usage if fewer than two args are given

add_to_comps.rb
```diff
@@ -0,0 +1,65 @@
...
+exit(1) unless find_packagereq(comps, ARGV[1]) == []
+packagelist = comps.xpath("//packagelist").first
+new_packagereq = Nokogiri::XML(
+  "<packagereq type='default'>#{ARGV[1]}</packagereq>"
+).children.first
+if ARGV[2] == 'nonscl'
+  scl_packagereqs = get_packagereqs(packagelist)
+  scl_packagereqs.sort! { |x,y| y.text <=> x.text }
```
domcleal: This seems to re-sort all of the existing files drastically and loses the groupings, do you intend to change them?

dLobatog: Yeah, it orders everything alphabetically within a 'scl' or 'nonscl' group. That sounds sensible to me, particularly I don't think it's going to be easy to tell the script how things are grouped and whether X or Y package belong in one group or another (+ the groups have no 'boundaries' right now,)

domcleal: That's fine then, just add another commit re-sorting the files so it doesn't fall to the first person to add a new package.

dLobatog:
Updated so that:
* the comps scripts handle wrong number of ARGV or other problems with args
* there's a commit with sorted comps
* docs reflect the right order of arguments when calling add_to_comps

domcleal:
Merged, thanks @dLobatog.

PR #1311, opened 2016-09-06
update hammer to 0.8.0 (DEB)

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.13 (after systest ran successfully)


mmoll:
please do review now :)

domcleal:
Are the new fast_gettext versions necessary? The one for trusty isn't I don't think, the 0.7 version ought to be just fine and saves us carrying a package update.

Hammer still works with any version as far as I know, it's just faster when used with 1.2.0. I'm happy to keep 1.2.0 on the newer OSes if you'd prefer to get the speed improvement.

mmoll:
For jessie and xenial I'd like to get 1.2.0 in, see https://groups.google.com/forum/#!topic/foreman-dev/kDPl1hgP9lQ

For trusty it would not be necessary, but that commit would bring 1.0.0 from Debian upstream in to oust the homegrown package, we have now.

domcleal:
Merged, thanks @mmoll.

Building for nightly only at http://ci.theforeman.org/job/packaging_build_deb_dependency/1362/ to 1364, will apply to deb/1.13 if it passes later.

domcleal:
Cherry-picked and built on deb/1.13.

PR #1325, opened 2016-09-11
packaging for puppet-foreman_scap_client [rpm]

shlomizadok:

* [x] Nightly
* [x] 1.13
* [x] 1.12
* [ ] 1.11

---


ares:
This is weird, when I try to download the module directly, I'm getting the following output with no redirection

```
*   Trying 52.10.130.237...
* Connected to forgeapi.puppetlabs.com (52.10.130.237) port 443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* ALPN/NPN, server did not agree to a protocol
* SSL connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* Server certificate:
*       subject: CN=*.puppetlabs.com
*       start date: čen 24 00:00:00 2016 GMT
*       expire date: srp 29 23:59:59 2019 GMT
*       common name: *.puppetlabs.com
*       issuer: CN=RapidSSL SHA256 CA,O=GeoTrust Inc.,C=US
> GET /v3/files/theforeman-foreman_scap_client-0.3.11.tar.gz HTTP/1.1
> Host: forgeapi.puppetlabs.com
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Server: nginx
< Date: Mon, 12 Sep 2016 07:25:47 GMT
< Content-Type: application/octet-stream
< Transfer-Encoding: chunked
< Status: 200 OK
< Cache-Control: public
< content-disposition: attachment; filename=theforeman-foreman_scap_client-0.3.11.tar.gz
< ETag: "664dddcc163b39ceffea70bfc2208ce1"
< X-Node: forgenext-app06-prod
< X-Revision: 66c9f4d
< Vary: Origin
< X-Content-Type-Options: nosniff
< X-App-Server: forgeapi/forgeapi-i-75aa79ae_forgenext-app06-prod.ops.puppetlabs.net
< X-Lb-Server: forgenext-lb02-prod.ops.puppetl
< X-UUID: D5AF250A:2129_0AE00633:01BB_57D6587A_44310FE:4768
```

the tarball has 12K, could we make an exception and maybe put this file into packaging repo?

ares:
what's even more weird, following jenkins steps locally lead to [working build in koji](http://koji.katello.org/koji/taskinfo?taskID=527238), I just had to modify tito.props

ares:
btw Shlomi, you might want to add this to this PR

```diff
diff --git a/rel-eng/tito.props b/rel-eng/tito.props
index ba5bf1e..f76f2e5 100644
--- a/rel-eng/tito.props
+++ b/rel-eng/tito.props
@@ -362,6 +362,7 @@ whitelist = rubygem-algebrick
 [foreman-plugins-nightly-nonscl-rhel7]
 disttag = .el7
 whitelist = rubygem-algebrick
+  puppet-foreman_scap_client
   rubygem-apipie-params
   rubygem-chef-api
   rubygem-concurrent-ruby
@@ -397,6 +398,7 @@ whitelist = rubygem-algebrick
 [foreman-plugins-nightly-fedora24]
 disttag = .fc24
 whitelist = rubygem-algebrick
+  puppet-foreman_scap_client
   rubygem-angular-rails-templates
   rubygem-apipie-params
   rubygem-azure
```

shlomizadok:
@ares - thanks, I have added the changes to `tito.props`
Got no clue why this is failing :/

ares:
could someone from @theforeman/packaging please take a look? would it be possible to avoid using annex for this package? or does anyone have some idea why it reports redirects?

domcleal:
git-annex 3.x doesn't support HTTPS sources, only HTTP sources, but Puppet Forge is only accessible over HTTPS. We now have enough EL7 slaves that I can switch these tests over to run on EL7 with 5.x which should fix the problem. I'll re-test it when I've updated Jenkins.

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -0,0 +1,37 @@
...
+Name:       puppet-%{puppet_module}
+Version:    0.3.11
+Release:    1%{?dist}
+Summary:    Puppet module to configure foreman_scap_client
+License:    GPLv2
+URL:        https://github.com/OpenSCAP/%{name}
+Source0:    https://forgeapi.puppetlabs.com/v3/files/%{puppet_full_name}-%{version}.tar.gz
+BuildArch:  noarch
+Requires:   puppet >= 2.7.0
+Requires:   puppetlabs-stdlib >= 4.2.0
```
domcleal: What provides this?

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -0,0 +1,37 @@
...
+%global puppet_full_name theforeman-%{puppet_module}
+
+%global puppet_modules_dir %{_datadir}/puppet/modules
+
+Name:       puppet-%{puppet_module}
+Version:    0.3.11
+Release:    1%{?dist}
+Summary:    Puppet module to configure foreman_scap_client
+License:    GPLv2
+URL:        https://github.com/OpenSCAP/%{name}
```
domcleal: Org is theforeman.

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -0,0 +1,37 @@
...
+
+%prep
+%setup -qn %{puppet_full_name}-%{version}
+
+%build
+
+%install
+mkdir -p %{buildroot}/%{puppet_modules_dir}/%{puppet_module}
+cp -p metadata.json %{buildroot}/%{puppet_modules_dir}/%{puppet_module}/
+cp -rp manifests/ %{buildroot}/%{puppet_modules_dir}/%{puppet_module}/manifests
```
domcleal: Other directories and files seem to be missing here. It would be better to copy everything and list subdirs in %files. This way rpmbuild will catch dirs and files not listed in %files, so they can be excluded or included explicitly.

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -0,0 +1,37 @@
...
+cp -rp templates/ %{buildroot}/%{puppet_modules_dir}/%{puppet_module}/templates
+
+
+%files
+%doc COPYING README.md
+%{puppet_modules_dir}/%{puppet_module}
```
domcleal: Prefer a list of dirs/files beneath here, then own this dir only with %dir.

domcleal:
[test] - packaging PR test slave updated to EL7, ought to fix git-annex HTTPS handling.

shlomizadok:
@domcleal - I kinda got lost here... am I doing this right?

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -0,0 +1,44 @@
...
+Name:       puppet-%{puppet_module}
+Version:    0.3.11
+Release:    1%{?dist}
+Summary:    Puppet module to configure foreman_scap_client
+License:    GPLv2
+URL:        https://github.com/theforeman/%{name}
+Source0:    https://forgeapi.puppetlabs.com/v3/files/%{puppet_full_name}-%{version}.tar.gz
+BuildArch:  noarch
+Requires:   puppet >= 2.7.0
+Requires:   puppetlabs-stdlib >= 4.2.0
```
domcleal: What provides this?

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -0,0 +1,44 @@
...
+cp -rp %{puppet_modules_dir}/%{puppet_module} %{buildroot}/%{puppet_modules_dir}/%{puppet_module}/
+
+
+%files
+%doc NEWS COPYING README.md
+%{puppet_modules_dir}/%{puppet_module}/metadata.json
+%{puppet_modules_dir}/%{puppet_module}/checksums.json
+%{puppet_modules_dir}/%{puppet_module}/Gemfile
+%{puppet_modules_dir}/%{puppet_module}/Rakefile
+%{puppet_modules_dir}/%{puppet_module}/lib/facter/rh_certificates.rb
```
domcleal: Just list `%{puppet_modules_dir}/%{puppet_module}/lib` so it includes everything below `lib/`. Ditto with manifests, templates.

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -0,0 +1,44 @@
...
+%{puppet_modules_dir}/%{puppet_module}/manifests/init.pp
+%{puppet_modules_dir}/%{puppet_module}/manifests/params.pp
+%{puppet_modules_dir}/%{puppet_module}/templates/config.yaml.erb
+%{puppet_modules_dir}/%{puppet_module}/templates/cron.erb
+%dir %{puppet_modules_dir}/%{puppet_module}
```
domcleal: Move this between the %doc and files within this dir, it should be first.

rel-eng/tito.props
```diff
@@ -362,6 +362,7 @@ whitelist = rubygem-algebrick
 [foreman-plugins-nightly-nonscl-rhel7]
 disttag = .el7
 whitelist = rubygem-algebrick
+  puppet-foreman_scap_client
```
domcleal: Move this up a line, it's not alphabetical.

rel-eng/tito.props
```diff
@@ -397,6 +398,7 @@ whitelist = rubygem-algebrick
 [foreman-plugins-nightly-fedora24]
 disttag = .fc24
 whitelist = rubygem-algebrick
+  puppet-foreman_scap_client
```
domcleal: ditto

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -0,0 +1,44 @@
...
+%build
+
+%install
+mkdir -p %{buildroot}/%{puppet_modules_dir}/%{puppet_module}
+cp -rp %{puppet_modules_dir}/%{puppet_module} %{buildroot}/%{puppet_modules_dir}/%{puppet_module}/
+
+
+%files
+%doc NEWS COPYING README.md
+%{puppet_modules_dir}/%{puppet_module}/metadata.json
```
domcleal: Consider defining %{puppet_module_dir} to make this less repetitive.

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -0,0 +1,44 @@
...
+to run scans and upload results to foreman proxy.
+
+%prep
+%setup -qn %{puppet_full_name}-%{version}
+
+%build
+
+%install
+mkdir -p %{buildroot}/%{puppet_modules_dir}/%{puppet_module}
+cp -rp %{puppet_modules_dir}/%{puppet_module} %{buildroot}/%{puppet_modules_dir}/%{puppet_module}/
```
domcleal: This is failing, check the source exists (probably `cp -rp . %{buildroot}/%{puppet_modules_dir}/%{puppet_module}` or similar), you need to be copying from the unpacked source, not the installation dir.

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
shlomizadok: How do I define this?

domcleal: Exactly the same way you already have: https://github.com/theforeman/foreman-packaging/pull/1325/files#diff-b80b253e976748958e41ab342c0c2d30R4

shlomizadok:
@domcleal, thanks for the help.
I have moved the "Requires puppetlabs-stdlib >= 4.2.0" as I am assume we do not provide this (and it could be installed via puppet forge)

domcleal:
I don't think removing the dependency is correct, it's specified as a dep within the metadata.json - so to use this package shouldn't require the user to install a package locally (else, why not just do it with this module too?). The package is available in Fedora, but not EPEL7.

shlomizadok:
Got it. Will return it. How do I make the package available in EPEL7?

domcleal:
I suggest filing a bug against the package first, else request the branch (https://fedoraproject.org/wiki/EPEL/FAQ#How_do_I_request_a_EPEL_branch_for_an_existing_Fedora_package.3F).

shlomizadok:
@domcleal - thanks, filed a bug and requested for EPEL7 branch.

ares:
meanwhile we could package stdlib too I guess, it does not have any other deps

shlomizadok:
@domcleal - EPEL7 has been approved:
https://admin.fedoraproject.org/pkgdb/package/rpms/puppetlabs-stdlib/

domcleal:
Yes, only the branch though - the build hasn't even moved into epel-testing yet: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2016-0c2c4e4800

puppet-foreman_scap_client/puppet-foreman_scap_client.spec
ares: seems it got into epel 7

https://dl.fedoraproject.org/pub/epel/7/x86_64/p/puppetlabs-stdlib-4.5.1-2.20150121git7a91f20.el7.noarch.rpm

rel-eng/tito.props
```diff
@@ -364,7 +364,8 @@ scl = tfm
 
 [foreman-nightly-fedora24]
 disttag = .fc24
-whitelist = foreman
+whitelist = puppet-foreman_scap_client
```
domcleal: Please remove it from here, it should only be in the plugin sections.

shlomizadok: Added by mistake. Removed.

domcleal:
Merged and will build, thanks @shlomizadok.

PR #1330, opened 2016-09-13
Update smart_proxy_remote_execution (DEB)

iNecas:

* [x] Nightly
* [x] 1.13
* [ ] 1.12
* [ ] 1.11

---


plugins/smart_proxy_remote_execution_ssh/remote_execution_ssh.rb
```diff
@@ -1 +1,2 @@
-gem 'smart_proxy_remote_execution_ssh', '0.1.3'
+gem 'smart_proxy_remote_execution_ssh', '0.1.4'
+gem 'foreman_remote_execution_core'
```
mmoll: I guess smart_proxy_remote_execution_ssh is loading foreman_remote_execution_core, so I hope it can get ommited here...

dependencies/jessie/foreman-tasks-core/changelog
```diff
@@ -0,0 +1,5 @@
+ruby-foreman-tasks-core (0.1.1-1) stable; urgency=low
```
mmoll: I don't like the name of the package, but given the current naming conventions, we don't have a choice :(

mmoll:
one note, :+1: otherwise

plugins/smart_proxy_remote_execution_ssh/remote_execution_ssh.rb
iNecas: Not really: the problem is with Ruby 1.8.7, which we need to support on proxy side for some distros ,while the core package is 1.9.3+. Therefore, we can't put it into deps of smart_proxy_remote_execution_ssh directly (on RHEL6, we delegate the execution to another process running in SCL)

dependencies/jessie/foreman-tasks-core/changelog
iNecas: What name do you think would better fit here? I'm open to suggestions, although I would prefer not to change that right away.

mmoll: As said, there's nothing we can do about this here... per convention "ruby-foreman-*" are Foreman core plugins, but using the convention for "standard" Debian Ruby packages (what we put under "dependencies" also leads to such a name now).

plugins/smart_proxy_remote_execution_ssh/remote_execution_ssh.rb
mmoll: I see, then it has to stay in.

mmoll:
I'll merge this for nightlies in just a bit and see how it behaves and in a few days to 1.13

mmoll:
merged, díky @iNecas!

PR #1326, opened 2016-09-12
Introduce foreman_tasks_core (RPM)

iNecas:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [x] 1.13
* [ ] 1.12
* [ ] 1.11

See Foreman's [plugin maintainer documentation](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

Since https://github.com/theforeman/foreman-packaging/pull/1321
contains multiple related packages, that one depends on others,
I'm splitting it to PRs to be handled on by one. This packages
are the first to get in.

PR #1335, opened 2016-09-14
Update foreman-ansible to 1.2 [rpm]

dLobatog:
Requires foreman_ansible_core to work - https://github.com/theforeman/foreman-packaging/pull/1336 (will update this comment to link to the actual PR)

This is for 1.13, 1.12 and nightlies


domcleal:
This release looks like it ships API controllers for the first time, so please add precompilation of the API docs to the build process, to %posttrans and the %files list (it might be easiest to compare against/merge the gem2rpm output).

The new -core build dependency should be available within 30 mins or so.

dLobatog:
The ansible plugin does not have the apipie engine so apparently it's not possible to generate those apipie docs. I think the best course of action is to release 1.2 with that, and remove the apipie stuff from this spec in this PR so we can have 1.1 published in the repos (albeit without apipie docs & caching)

dLobatog:
I've updated the spec to *not* include the apipie cache stuff as it makes it fail since the plugin is missing the initializer

dLobatog:
https://github.com/theforeman/foreman_ansible/pull/51 is merged, maybe wait a moment until I update the spec and release the gem, no reason to not release 1.2 directly at this point

dLobatog:
@domcleal Updated to 1.2

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -2,10 +2,11 @@
 %{!?scl:%global pkg_name %{name}}
 
 %global gem_name foreman_ansible
+%global plugin ansible
```
domcleal: Should be `plugin_name` to match the Provides on line 40.

rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -72,14 +78,23 @@ exit 0
 %{gem_instdir}/locale
 %exclude %{gem_cache}
 %{gem_spec}
-%{foreman_bundlerd_dir}/%{gem_name}.rb
+%{foreman_bundlerd_plugin}
+%{foreman_apipie_cache_foreman}
```
domcleal: There should also be foreman_apipie_cache_plugin here (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/gem2rpm/foreman_plugin.spec.erb#L163)

dLobatog: I think that macro might be wrong. If you see the `ln -s` part here: https://gist.github.com/dLobatog/f9f23eabda26b9356d41b7433a62fa5b

The second argument to `ln` is *NOT* the output directory of `plugin:apipie:cache`:
````
/builddir/build/BUILDROOT/tfm-rubygem-foreman_ansible-1.2-1.fm1_14.git.366.e7c240e.el7.centos.x86_64/opt/theforeman/tfm/root/usr/share/gems/gems/foreman_ansible-1.2/public/apipie-cache/plugin/foreman_ansible
```
but instead it's 
```
/builddir/build/BUILDROOT/tfm-rubygem-foreman_ansible-1.2-1.fm1_14.git.366.e7c240e.el7.centos.x86_64/usr/share/foreman/public/apipie-cache/plugin/foreman_ansible
```

Should I change the macro? I don't think it's linking to the right place (or outputting apipie:cache to the right place)

domcleal: > The second argument to ln is NOT the output directory of plugin:apipie:cache:

The _first_ argument is the output directory of the rake task though - and the first argument is the target of the symlink.

The `ln -s` should create a symlink from /usr/share/foreman/public/apipie-cache/plugin/foreman_ansible to the directory containing the rake task output in the gems/foreman_ansible-1.2/ directory.

You then need `foreman_apipie_cache_foreman` listed to include the symlink, _and_ `foerman_apipie_cache_plugin` to include the rake output directory.

dLobatog:
@domcleal Updated - the problem was in the plugin itself, the apipie initializer in engine.rb was wrong

domcleal:
Merged and building for nightly, thanks @dLobatog.

Currently this isn't buildable for 1.13 as #1336 was only for rpm/develop - did you want that cherry picked to rpm/1.13 too?

dLobatog:
Yeah, sorry for removing the comment - both should be available for 1.12, 1.13 and nightly

domcleal:
Are you sure about 1.12? The ansible-core RPM depends on tasks-core, but that _wasn't_ requested for 1.13 (#1326).

domcleal:
For the record, I built this for nightly and 1.13 only, not 1.12 (yet) as the chain of deps are missing.

PR #1338, opened 2016-09-14
Update foreman ansible to 1.2 [deb]

dLobatog:

* [x] Nightly
* [x] 1.13
* [ ] 1.12
* [ ] 1.11

---

Requires a few dependencies that are not there yet

### DIFF
--- a/plugins/ruby-foreman-ansible/debian/control
+++ b/plugins/ruby-foreman-ansible/debian/control
@@ -2,12 +2,16 @@ Source: ruby-foreman-ansible
 Section: ruby
 Priority: extra
 Maintainer: Daniel Lobato Garcia <elobatocs@gmail.com>
-Build-Depends: debhelper, foreman (>= 1.12.0~rc1), ruby-foreman-deface (<<2.0.0)
+Build-Depends: debhelper, foreman (>= 1.12.0~rc1), ruby-foreman-deface (<<2.0.0),
+  ruby-foreman-tasks (>= 0.8.2-2), ruby-foreman-tasks (<<0.9.0),
+  foreman-sqlite3 (>= 1.12.0~rc1)
 Standards-Version: 3.9.3
 Homepage: https://github.com/theforeman/foreman_ansible
 
 Package: ruby-foreman-ansible
 Architecture: all
-Depends: ${misc:Depends}, bundler, foreman (>= 1.12.0~rc1), ruby-foreman-deface (<<2.0.0)
+Depends: ${misc:Depends}, bundler, foreman (>= 1.12.0~rc1),
+  ruby-foreman-deface (<<2.0.0), ruby-foreman-tasks (>= 0.8.2-2),
+  ruby-foreman-tasks (<<0.9.0)
 Description: Foreman Ansible plugin
  Integration with Ansible in Foreman.
--- a/plugins/ruby-foreman-ansible/debian/rules
+++ b/plugins/ruby-foreman-ansible/debian/rules
@@ -9,5 +9,20 @@
 # Uncomment this to turn on verbose mode.
 #export DH_VERBOSE=1
 
+PLUGIN = foreman_ansible
+
+build:
+	cp cache/* /usr/share/foreman/vendor/cache/
+	cp $(PLUGIN).rb /usr/share/foreman/bundler.d/
+	cd /usr/share/foreman && ( \
+		/usr/bin/foreman-ruby /usr/bin/bundle install --local && \
+		/usr/bin/foreman-ruby /usr/bin/bundle exec rake db:migrate RAILS_ENV=development && \
+		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
+		OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
+		)
+	GEM_PATH=$$(cd /usr/share/foreman && /usr/bin/foreman-ruby /usr/bin/bundle show $(PLUGIN))
+	[ -e apipie-cache ] || mkdir apipie-cache/
+	cp -rp /var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) ./apipie-cache/
+
 %:
 	dh $@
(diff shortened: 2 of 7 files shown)


plugins/ruby-foreman-ansible/debian/control
```diff
@@ -8,6 +8,9 @@ Homepage: https://github.com/theforeman/foreman_ansible
 
 Package: ruby-foreman-ansible
 Architecture: all
-Depends: ${misc:Depends}, bundler, foreman (>= 1.12.0~rc1), ruby-foreman-deface (<<2.0.0)
+Depends: ${misc:Depends}, bundler, foreman (>= 1.12.0~rc1),
+  ruby-foreman-deface (<<2.0.0), ruby-foreman-tasks (>= 0.8.1),
+  ruby-foreman-tasks (<<0.9.0), ruby-dynflow (>= 0.8.14),
```
mmoll: AFAICT, foreman-tasks in the 1.12 repos is  0.7.x, so that'll be nightly and 1.12. I presume?

mmoll:
does this require foreman_ansible_core?

plugins/ruby-foreman-ansible/debian/control
dLobatog: Yeah - my bad :smile:

dLobatog:
@mmoll Yup, both the proxy plugin and the Foreman plugin need it

mmoll:
@dLobatog Then, as it is in the latest foreman-tasks plugin for Foreman itself, it needs to get added to get gem.list and removed as dependency here, as Foreman on Debian uses its own gemset. Dynflow and friends are also already getting pulled in by the -tasks Foreman plugin package.

For the proxy plugin, these are still needed, as the proxy does use system gems.

dLobatog:
@mmoll Looks good now? As far as I understood I can leave the tasks package in `control`, which will fetch dynflow, and leave `foreman_ansible_core` only in gem.list

mmoll:
This looks good, but it doesn't work with foreman_remote_execution also installed. The same problem was the reason for pulling in dynflow directly into the foreman-tasks package. Would it be possible for foreman_ansible and foreman_remote_execution not to depend on dynflow directly, but only on foreman-tasks in the gemspec? /cc @iNecas

iNecas:
The reason for this was the plugins require some change that was introduced in newer version of dynflow, that the tasks itself have as dependency. The workaround would be to increase the dependency in tasks, and than bump the required dependency here from here to foreman_tasks, but it's suboptimal, ale the tasks itself don't have hard requirement on that dynflow version. Is there any change to fix this issue in packaging rather than in the gemspecs?

mmoll:
I'm pretty sure we're hitting an issue here that has the same root cause as http://projects.theforeman.org/issues/11714, which led to rolling dynflow into the foreman-tasks package as no other solution was available back then.

mmoll:
this needs a newer dynfloe version first anyway.

iNecas:
Well, I think after that change, the puppet code should actually be compatible with the version of packages that are in  debian 1.12

mmoll:
With the new -tasks package containing the newer dynflow, my installation problems also went away. I'll chrry-pick this with the -tasks version requirement amended. However, we should be watching if problems like the old issue are going to be seen again on updates of these components.

mmoll:
hang on a sec... if the new version of the ansible plugin is extending the API, we'll need to do the plugin:apipie:cache dance for this plugin here, also.

dLobatog:
@mmoll Updated so that it creates the apipie cache now

dLobatog:
Rebased & pushed again because the job was stuck after 18h and Jenkins was restarted

dLobatog:
Seems like dynflow was not pulled from foreman-tasks , which was not a builddep of this package.. I've added sqlite and tasks as build deps of foreman ansible now

mmoll:
dynflow is being pulled in, but no idea what this means:
```
09:47:05 I, [2016-09-23T09:46:28.017639 #7613]  INFO -- /default_dead_letter_handler: got dead letter #<Concurrent::Actor::Envelope:0x000000079a4528 @message=:periodic_check_inbox, @future=nil, @sender=#<Dynflow::ClockReference:0x46d6358 /clock (Dynflow::Clock)>, @address=#<Concurrent::Actor::Reference:0x7af06c0 /connector-database-core (Dynflow::Connectors::Database::Core)>>
```

Also, please raise the needed ruby-foreman-tasks version to 0.8.2-2, as this is the only version containing the newer, required dynflow version (0.8.15 at this time).

plugins/ruby-foreman-ansible/debian/rules
```diff
@@ -9,5 +9,22 @@
...
 #export DH_VERBOSE=1
 
+PLUGIN = foreman_ansible
+
+build:
+	cp cache/* /usr/share/foreman/vendor/cache/
+	cp $(PLUGIN).rb /usr/share/foreman/bundler.d/
+	cd /usr/share/foreman && ( \
+		/usr/bin/foreman-ruby /usr/bin/bundle install --local && \
+		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production && \
```
mmoll: there are assets, AFACT, so this fails.

mmoll:
we're getting there...

mmoll:
Per IRC, it seems that there's no apipie initializer. so no APi docs get generated. So either we'd need a new foreman_siable gem release with that in, or we scrap that part for now, which means that the whole apipie dance in the packaging needs to get removed again.

dLobatog:
Correct - the initializer is here https://github.com/theforeman/foreman_ansible/pull/51 so I think I might as well scrap this and directly release 1.2 for debs

dLobatog:
Updated (including the initializer) to 1.2 - this should build the apipie cache

mmoll:
it seems there still problems with the apidocs
```
11:51:27 cp -rp /var/lib/foreman/public/apipie-cache/plugin/foreman_ansible ./apipie-cache/
11:51:27 cp: cannot stat '/var/lib/foreman/public/apipie-cache/plugin/foreman_ansible': No such file or directory
11:51:27 make: *** [build] Error 1
11:51:27 dpkg-buildpackage: error: debian/rules build gave error exit status 2
```

dLobatog:
@mmoll Updated - the problem was in the plugin itself, the apipie initializer in engine.rb was wrong

mmoll:
cherry-picked as bb91005 for develop and 478875e for 1.13, gracias @dLobatog!

PR #1343, opened 2016-09-14
fixes #16553 use --local for gem install

mmoll:


domcleal:
Could you add a changelog entry too please, so we can rebuild it?

mmoll:
done.

domcleal:
Merged, thanks @mmoll.  Building at http://ci.theforeman.org/job/packaging_build_deb_plugin/1215/.

PR #1351, opened 2016-09-19
update dynflow to 0.8.15 (DEB)

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.13


domcleal:
Merged and building for both releases, thanks @mmoll.

PR #1355, opened 2016-09-21
refs #16599 - update npm to latest release [deb]

domcleal:
Required to read a partial shrinkwrap file.

--

develop only for now, please. Should be harmless by itself, may even speed up installation.


domcleal:
I removed the change from xenial as it already has npm 3 available, should only be needed on the older OSes with 1.x.

domcleal:
[test] as I think it tested against an older version of develop.

mmoll:
I did restart the build after the core PR went in as http://ci.theforeman.org/job/packaging_build_deb_coreproject/3657/ and it came back green :green_heart:

mmoll:
merged, thanks @domcleal!

PR #1356, opened 2016-09-23
Add nodejs-multiselect 0.9.12

dLobatog:

* [x] Nightly
* [ ] 1.13
* [ ] 1.12
* [ ] 1.11

---

### DIFF
--- a/nodejs-multiselect/nodejs-multiselect.spec
+++ b/nodejs-multiselect/nodejs-multiselect.spec
@@ -1,0 +1,58 @@
+%global npm_name multiselect
+%global enable_tests 1
+
+%{?nodejs_find_provides_and_requires}
+
+Name: nodejs-%{npm_name}
+Version: 0.9.12
+Release: 1%{?dist}
+Summary: A replacement for the standard select with multiple attributes
+License: MIT
+URL: http://loudev.com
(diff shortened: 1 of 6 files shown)


dLobatog:
Wrong dep version on foreman - updated so it's **>=**

nodejs-multiselect/nodejs-multiselect.spec
```diff
@@ -0,0 +1,58 @@
+%global npm_name multiselect
+%global enable_tests 1
+
+%{?nodejs_find_provides_and_requires}
+
+Name: nodejs-%{npm_name}
+Version: 0.9.12
+Release: 1%{?dist}
+Summary: A user-friendlier drop-in replacement for the standard select with multiple attribute activated
```
domcleal: Should be shorter, IIRC it's a 72 character limit

nodejs-multiselect/nodejs-multiselect.spec
```diff
@@ -0,0 +1,58 @@
...
+License: WTFPL
```
domcleal: The licence text shipped inside the package is MIT, I don't think the metadata is correct.

dLobatog:
Updated the license (apparently npmjs was wrong) and shortened the summary, thanks for taking a look @domcleal

domcleal:
Thanks @dLobatog, building.

PR #1368, opened 2016-10-03
Update foreman_azure 1.0.2 [deb]

dLobatog:
- [x] Nightly
- [x] 1.13
- [ ] 1.12
- [ ] 1.11

---

### DIFF
--- a/plugins/ruby-foreman-azure/debian/changelog
+++ b/plugins/ruby-foreman-azure/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-azure (1.0.2-1) stable; urgency=low
+
+  * 1.0.2 released
+
+ -- Daniel Lobato Garcia <me@daniellobato.me>  Mon, 03 Oct 2016 16:26:42 +0200
+
 ruby-foreman-azure (1.0.1-1) stable; urgency=low
 
   * Initial release.
--- a/plugins/ruby-foreman-azure/debian/gem.list
+++ b/plugins/ruby-foreman-azure/debian/gem.list
@@ -1,5 +1,5 @@
 # list of gem urls to download via Jenkins, space separated
-GEMS="https://rubygems.org/downloads/foreman_azure-1.0.1.gem"
+GEMS="https://rubygems.org/downloads/foreman_azure-1.0.2.gem"
 GEMS="$GEMS https://rubygems.org/downloads/fog-azure-0.0.2.gem"
 GEMS="$GEMS https://rubygems.org/downloads/azure-0.7.5.gem"
 GEMS="$GEMS https://rubygems.org/downloads/azure-core-0.1.1.gem"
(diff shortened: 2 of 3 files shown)


mmoll:
```
15:24:15 Bundler could not find compatible versions for gem "mime-types":
15:24:15   In Gemfile:
15:24:15     foreman_azure (= 1.0.2) ruby depends on
15:24:15       mime-types (< 3.0, >= 1) ruby
15:24:15 
15:24:15     rest-client (< 3, >= 1.8.0) ruby depends on
15:24:15       mime-types (3.1)
```

domcleal:
azure_core should be changed to permit a 3.x version of mime-types, else you may need to consider pinning mime-types in Foreman itself (well, the Debian packaging) to a 2.x version. Currently 3.1's being installed as it's the latest compatible available when `foreman` is built.

dLobatog:
Interesting... I don't see that in my Gemfile.lock, in fact I don't see anything that requests that version of mime-types, any idea @mmoll ? The rpm PR seemed to work without any mime-type changes https://github.com/theforeman/foreman-packaging/pull/1369/files

```
azure (0.7.6)
      addressable (~> 2.3)
      azure-core (~> 0.1)
      faraday (~> 0.9)
      faraday_middleware (~> 0.10)
      json (~> 1.8)
      mime-types (>= 1, < 3.0)
      nokogiri (~> 1.6)
      systemu (~> 2.6)
      thor (~> 0.19)
rest-client (2.0.0)
      http-cookie (>= 1.0.2, < 2.0)
      mime-types (>= 1.16, < 4.0)
      netrc (~> 0.8)
```

domcleal:
> rest-client (2.0.0)
>    mime-types (>= 1.16, < 4.0)

This is allowing 3.1 in your Gemfile.lock, but presumably because you're loading both Foreman and foreman_azure in one bundle, it resolves to a compatible version.

Here Foreman is built without foreman_azure, so it's built with the latest compatible version (3.1) and so that will fail when the plugin's added later with a stricter requirement.

Regarding the RPMs, we don't use Bundler there so don't automatically get the latest version. It will become an issue in the very near future as new OSes (such as Fedora 25) ship with 3.1. This needs fixing in the azure gem, really.

dLobatog:
Yeah - I've opened an issue in azure and fog-azure to support it. If they are responsive I can make the next release of foreman_azure use these new versions.

domcleal:
The pins are kept in debian/*/foreman/debian.rb in case you change your mind!

dLobatog:
I think I'll just make foreman_azure 1.0.3 pin mime-types and release it for debian only. - no point on doing so for rpm

domcleal:
Adding a pin to foreman_azure will not help, it's the same as having the pin in `azure`. The version in `foreman` is too new for foreman_azure/azure, it needs to be downgraded in the `foreman` package.

domcleal:
Note that azure-0.7.7 is now available which permits mime-types 3.x.

domcleal:
Looks like #1412 replaces this?

dLobatog:
Yeah, closing this one

PR #1376, opened 2016-10-04
column_view: 0.3.0 release (RPM)

GregSutcliffe:
Small code updates only, should be compatible with all supported versions. However, the main thing is the UI fix which will only be in 1.13.1+ so there's not much point building for older versions.

* [x] Nightly
* [x] 1.13
* [ ] 1.12
* [ ] 1.11


domcleal:
Merged and will build, thanks @GregSutcliffe.

PR #1383, opened 2016-10-13
fixes #16898 - delete concurrent-ruby-edge for xenial

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.13
* [x] 1.12

While for Ubuntu/trusty and Debian/jessie the ruby-concurrent package is coming from our repos, Ubuntu/xenial does provide this package in the Ubuntu repo. Our builds download the gems from rubygems.org, but the native Ubuntu package is built from a gbp repository of the concurrent-ruby gem, leading to the situation that the speration between the main and the -edge gem is not given anymore.

I have to say I find it now a quite horrid hack from @ruby-concurrency to have these gems all in one rpeo. :)


domcleal:
The dynflow gem still depends on concurrent-edge, I don't see how you can delete the dependency without it causing issues? (requires not working, references to Concurrent::Edge failing)

mmoll:
hm, that's correct... I guess we would have to provide a ruby-concurrent pacage on xenian also then.

domcleal:
Oh sorry, I think I understand what you're saying now - the `ruby-concurrent` package actually contains the contents of concurrent-edge as well, because the tarball's built from the repo. It looks like the package contains everything, so requiring `concurrent-ruby-edge` should work correctly and the separate Edge namespace is preserved too.

I think this ought to work, I'll test the pull request. Sorry about that.

mmoll:
I have to admit that i also didn't test this on a real xenial VM, yet

domcleal:
Merged and will build, thanks @mmoll. I'll remove the edge package from nightly as that's easiest, but build the dependency package for the other releases too.

Apologies - somehow I used a merge commit after selecting "squash", possibly by using Ctrl+Enter instead of clicking the button.

PR #1386, opened 2016-10-16
set NODE_ENV=production for npm install (DEB)

mmoll:
I'm unsure about the consequences in the application itself (@dLobatog? @tbrisker?), but this seems to lead to a significant speedup when building the package.


mmoll:
turns out, "devDependencies" are actually build dependencies.

PR #1387, opened 2016-10-16
remove test and dev gems for building (DEB)

mmoll:
These should not be needed for production build packages. therubyracer got removed in the meanwhile...

This PR is for now mainly to see, where the builds are going, maybe there are hidden prod or build dependencies.


mmoll:
ther's actually a foreman-test pakcage and the speedup at build time is neglectable

PR #1389, opened 2016-10-18
Fixes #16981 - require ruby version 2.3.1 for xenial.

witlessbird:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [ ] 1.13
* [ ] 1.12
* [ ] 1.11

See Foreman's [plugin maintainer documentation](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

PR #1380, opened 2016-10-07
refs #15834 - rebuild bundler on Ubuntu/trusty

mmoll:
The Ubuntu package seems to be built for Ruby 1.8  and 1.9, while 1.9 and 2.0 is needed.

PR #1390, opened 2016-10-18
use ruby2.0 for foreman-proxy on trusty

mmoll:
this should be built into:

* [x] Nightly

contrary to #1389, the ruby interpreter used is changed here. Depends on #1380 (already merged now), to really get bundler correctly for bundler_ext.

This is for future-proofing Ubuntu/trusty as platform, as foreman-proxy might (need to) drop support for Ruby 1.9.3 in the future. @iNecas is there anything to be aware of regarding the smart_proxy*-core proxy plugins on Debian?

### DIFF
--- a/debian/trusty/foreman-proxy/control
+++ b/debian/trusty/foreman-proxy/control
@@ -10,7 +10,7 @@ Vcs-Browser: https://github.com/theforeman/smart-proxy
 
 Package: foreman-proxy
 Architecture: all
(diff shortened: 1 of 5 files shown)


iNecas:
Since we're using in-process dynflow executor on Debian, the -core
gems should not be affected by this change as long as they are
loadable by the ordinary `require`.

mmoll:
Seeing the diff again, I guess it would be better to explicitely use `/usr/bin/ruby2.0` here.

mmoll:
[test]

domcleal:
Can you add a changelog entry to the bundler_ext package please?

mmoll:
done.

domcleal:

PR #1395, opened 2016-10-21
Update foreman_expire_hosts to 2.1.1 [rpm]

timogoebel:
Please build for:

* [x] Nightly
* [x] 1.13
* [ ] 1.12
* [ ] 1.11


domcleal:
The 2.1.1 release doesn't appear to be on rubygems.org, can you upload it?

timogoebel:
@domcleal : Done, sorry. I think, I need to automate plugin releases. Do you have some scripts I could take as a template?

domcleal:
[test] with the gem pushed.

> I think, I need to automate plugin releases. Do you have some scripts I could take as a template?

No, but there are lots of choices out there. Bundler itself provides `bundler/gem_tasks` (require it) which I think includes tagging and pushing, and there's also the `rubygems-tasks` gem, `hoe` and others. You'd add those gems to the development deps, but ensure any require from lib/tasks/ fails quietly for production use.

domcleal:
Merged and building, thanks @timogoebel.

PR #1403, opened 2016-11-02
Added debug hook script

lzap:
Merge only after the next release is ready.

### DIFF
--- a/rubygem-foreman_hooks/rubygem-foreman_hooks.spec
+++ b/rubygem-foreman_hooks/rubygem-foreman_hooks.spec
@@ -74,6 +74,11 @@ GEMFILE
 mkdir -p %{buildroot}%{foreman_dir}/config/hooks
 ln -s %{gem_instdir} %{buildroot}%{foreman_dir}/%{gem_name}
 
+# debug script
+chmod +x %{buildroot}%{gem_instdir}/extra/foreman-debug.sh
+%{__mkdir_p} %{buildroot}%{foreman_dir}/script/foreman-debug.d
+ln -s %{gem_instdir}/extra/foreman-debug.sh %{buildroot}%{foreman_dir}/script/foreman-debug.d/50-foreman_hooks
+
 %files
 %dir %{gem_instdir}
 %{gem_libdir}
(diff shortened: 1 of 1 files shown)


lzap:
@mmoll could you guide me through the same Debian change please? The relevant bit is at https://github.com/theforeman/foreman_hooks/pull/40/files - I only need to create a symlink there.

mmoll:
@lzap I guess it's more complicated on Debian, as we use `bundle install --local` for plugins there. I need to think about this, as it would be needed to build the path to link to from the ruby version and the foreman_hooks version and it probably happens best in the postinst script.

mmoll:
The easiest solution for DEB would be just include the script it with the package source, as -hooks is only shipped as .gem in the package, but it's not nice...

lzap:
Can't we simply extract the file from GEM in the "build" section? I don't exactly how DEB works, but there must be some kind of build/compile step.

mmoll:
Core plugins are very different on Debian. An an example, here's what's coming in a .deb:
```
mmoll@sledge:~$ dpkg -L ruby-foreman-xen
/.
/usr
/usr/share
/usr/share/doc
/usr/share/doc/ruby-foreman-xen
/usr/share/doc/ruby-foreman-xen/copyright
/usr/share/doc/ruby-foreman-xen/changelog.Debian.gz
/usr/share/foreman
/usr/share/foreman/vendor
/usr/share/foreman/vendor/cache
/usr/share/foreman/vendor/cache/foreman_xen-0.4.0.gem
/usr/share/foreman/bundler.d
/usr/share/foreman/bundler.d/foreman_xen.rb
```
In the package postinst the .gem gets installed via `bundle install --local` into foreman's gem environemnt (which is different from the system gem environment). I guess there can be ways to construct the path of the installed gem and thus the foreman-debug script to symlink to in the postinst, but I'd need to explore that further.

rubygem-foreman_hooks/rubygem-foreman_hooks.spec
```diff
@@ -74,6 +74,9 @@ GEMFILE
 mkdir -p %{buildroot}%{foreman_dir}/config/hooks
 ln -s %{gem_instdir} %{buildroot}%{foreman_dir}/%{gem_name}
 
+# debug script
+ln -s %{gem_instdir}/extra/foreman-debug.sh %{foreman_dir}/script/foreman-debug.d/50-foreman_hooks
```
domcleal: 1. Parent directory doesn't seem to be created
1. Symlink would have to be in the buildroot, not in `/`

lzap:
Now it should work!

domcleal:
Merged, thanks @lzap.

PR #1405, opened 2016-11-02
Deb/1.13 dependency

brechsteiner:
fix package dependency


mmoll:
@brechsteiner Whatever your tried to pull request here did get a bit out of control ;)

Please use `deb/develop` as branch to branch from and pull request to for everything. We'll cherry-pick accordingly, where needed.

Regarding the change of ruby-rb-inotify from Recommends to Depends: When using the proxy without dhcp, inotify is not used, so it's no hard dependency.

PR #1369, opened 2016-10-03
Update foreman_azure to 1.0.2 [rpm]

dLobatog:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [x] 1.13
* [ ] 1.12
* [ ] 1.11

See Foreman's [plugin maintainer documentation](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

PR #1411, opened 2016-11-15
Update foreman_azure to 1.1.1 [rpm]

dLobatog:

* [x] Nightly
* [x] 1.13
* [x] 1.12
* [x] 1.11

---

Thanks!


dLobatog:
[test] - the gem wasn't in rubygems yet when I pushed this

domcleal:
Are you sure this is compatible with 1.12/1.11? The last update (#1369) was only for 1.13+ and it looks like this one also contains 1.13-only changes such as https://github.com/theforeman/foreman_azure/commit/24cc09c812cef7617a5b3015b9f95a8170094e83.

domcleal:
https://github.com/theforeman/foreman_azure/blob/master/lib/foreman_azure/engine.rb#L9 states 1.8, but `parameter_filter` is 1.13+ only.

dLobatog:
Hmm - that's right. I'll try to make a patch release with a couple of 'Foreman version' conditionals for these things and test with 1.11/1.12, if it works we'll release for these 2 versions too, if it's more than that it'll be just 1.13. Thanks @domcleal !

dLobatog:
Updated - https://github.com/theforeman/foreman_azure/commit/7ef8e73d5743fba2839741f45df4ba307182c8f3 I had to make a few adjustments for it to work, 1.1.1 has them and it provisions & SSHs the finish template into a host just fine

domcleal:
Thanks, can you update the requires on Foreman to be >= 1.11, matching engine.rb?

dLobatog:
Updated!

domcleal:
Thanks @dLobatog, merged and building.

PR #1429, opened 2016-11-23
execute build tasks only once

mmoll:
this should be built into:

* [x] Nightly

It seems that the build tasks when building all the packages are executed at least twice and this change fixes this. I tested the resulting Ubuntu/xenial packages and a quick all-in-one-install seemed to work well.

@GregSutcliffe do you remember any background on that?


domcleal:
Merged, thanks @mmoll.

GregSutcliffe:
For the record, I think that's a leftover from the original pre-1.0 packaging that Joschi did. I had always meant to figure out what was causing the double builds - nice catch!

PR #1436, opened 2016-11-25
Refs #15779 - add dependency to foreman-tasks (rpm)

iNecas:
To reflect changes in https://github.com/theforeman/foreman/pull/3670


foreman/foreman.spec
```diff
@@ -158,6 +160,8 @@ BuildRequires: %{?scl_prefix}rubygem(deep_cloneable) >= 2.2.2
 BuildRequires: %{?scl_prefix}rubygem(deep_cloneable) < 3.0
 BuildRequires: %{?scl_prefix}rubygem(validates_lengths_from_database) >= 0.5
 BuildRequires: %{?scl_prefix}rubygem(validates_lengths_from_database) < 1.0
+BuildRequires: %{?scl_prefix}rubygem(foreman-tasks) >= 0.8.5
```
domcleal: 1. This creates a circular dependency between the packages, preventing new builds.
1. Plugin packages aren't available in the main build roots (plugins must not be dependencies of core packages).

foreman/foreman.spec
```diff
@@ -77,6 +77,8 @@ Requires: %{?scl_prefix}rubygem(deep_cloneable) >= 2.2.2
 Requires: %{?scl_prefix}rubygem(deep_cloneable) < 3.0
 Requires: %{?scl_prefix}rubygem(validates_lengths_from_database) >= 0.5
 Requires: %{?scl_prefix}rubygem(validates_lengths_from_database) < 1.0
+Requires: %{?scl_prefix}rubygem(foreman-tasks) >= 0.8.5
+Requires: %{?scl_prefix}rubygem(foreman-tasks) < 0.9.0
```
domcleal: The bundler.d file doesn't specify this upper bound.

foreman/foreman.spec
```diff
@@ -848,6 +852,7 @@ rm -rf %{buildroot}
 %{_datadir}/%{name}/bundler.d/jsonp.rb
 %exclude %{_datadir}/%{name}/bundler.d/openid.rb
 %exclude %{_datadir}/%{name}/bundler.d/test.rb
+%exclude %{_datadir}/%{name}/bundler.d/foreman_tasks.rb
```
domcleal: 1. Keep this list alphabetical.
1. The file in the foreman SRPM seems more precise than the one from `rubygem(foreman-tasks)`, shouldn't it be preferred?

foreman/foreman.spec
iNecas: updated

foreman/foreman.spec
iNecas: I don't think we need the tasks in the build phase, the only thing I can think of si the class loading of dynflow actions, which we could workaround if needed. The other solution might be removing foreman from foreman-tasks dependencies, as the foremen-tasks usually get installed only as dependency of something else, so the `foreman` it's guaranted to be around.

I'm trying now without the build dependency specified to see if there are any issues there.

foreman/foreman.spec
iNecas: For the rpm packaging, the source of dependency specification is in the spec itself, so having more precision in the bundler file doesn't seem to be too much a benefit, I went for skpping it here and relying on the foreman-tasks one, as otherwise, the foreman-tasks plugin would not get it's depednency there for older foreman versions. On the other hand, as I mentioned above, the tasks is barely used on it's own, without other plugins or core, so I guess we could go both ways if needed.

foreman/foreman.spec
ares: I think foreman should no longer be foreman-tasks dependency, it's like with fog, rails or other foreman dependencies.

iNecas:
I still see the builds being run against 2775293 foreman commit, so it gives warning about the bundler.d/foreman_tasks.rb file not being present, I guess we need to wait for something to happen before we can see the real results.

iNecas:
Ok, it seems we don't need foreman-tasks at the build time after the last commit

domcleal:
1. There's still a circular dependency with a require on `rubygem-foreman-tasks`, as that package requires `foreman` to build - it will not be possible to build `rubygem-foreman-tasks` from scratch.
1. The dependency on a package in the plugins repo is bad form, it makes `foreman` uninstallable from both the main buildroot and repo.

iNecas:
Could we define a package `foreman-core` that would take (all the requirements - the plugins) and the foreman-plugin package, that the plugins have as BuildDependency, to depend on foreman-core, not the foreman itself. And the foreman would depend on foreman-core + plugins? Anyone has some simpler solution to this?

iNecas:
Putting on need re-review, as I need some answers before continuing

ares:
After going through other alternatives with @iNecas I think splitting the `foreman` package to `foreman-core` and `foreman` would make most sense. If I get it right, `foreman-core` would have the same deps as it has today. New `foreman` would require `foreman-core` and all other code that needs `foreman-core`, atm only `foreman-tasks`. `foreman-tasks` would require `foreman-core` only.


```
+------------------------+
|   foreman core         | <-------------------+
+------------+-----------+                     |
             ^                    +------------+----------+
             |                    |    foreman tasks      |
             |                    +------------+----------+
+------------+-----------+                     ^
|   foreman              +---------------------+
+------------------------+

arrow means require
```

domcleal:
> Could we define a package foreman-core that would take (all the requirements - the plugins)

No, I don't see how, because foreman-tasks is defined as a hard dependency in Foreman. If it were optional then it'd be in a separate bundler group which can be excluded, and the app would work without it.

iNecas:
I've updated the spec to introduce the foreman-core subpackage, that is used for buildling the plugins without requiring the plugins needed by foreman runtime. This way, both foreman and foreman-tasks can be build without the cyclic dependency (I was able to rebuild even existing foreman-tasks srpm with this locally, without need for foreman-tasks to be installed)

As a side-effect, this also avoids some unnecessary steps from %post and %posttrans to be run on plugin build time.

iNecas:
The thing is we don't need foreman-tasks present neither for building foreman nor foreman-tasks, Foreman-tasks is required for runtime, not build time, that's why I fail to see your concerns here.

ares:
Maybe we could change the description of foreman-core package to something that says it's not full Foreman but can be used to build Foreman plugins.

domcleal:
> The thing is we don't need foreman-tasks present neither for building foreman nor foreman-tasks, Foreman-tasks is required for runtime, not build time

OK, perhaps making it an optional dependency via a new Bundler group in Foreman would be best then if you claim it's unneeded. Currently it's in the default group which suggests it's a mandatory dependency.

The PR is currently ignoring that, which I think is problematic. It's likely to break when Foreman uses the gem in its startup or from any of the classes that are used by the "build" tasks.

If declared optional in a separate bundler group then it can be treated like any of the other optional subpackages (e.g. `foreman-ec2`), and it will be left out of the build environment automatically.

iNecas:
Ok, I've opened https://github.com/theforeman/foreman/pull/4075 to move the tasks dependency outside of default group and updated the spec here to use exclude instead of rm.

iNecas:
[test] https://github.com/theforeman/foreman/pull/4075 got merged

foreman/foreman.spec
```diff
@@ -31,6 +31,14 @@ Source7: %{name}-plugins.repo
 Source8: %{name}.gpg
 BuildArch:  noarch
 
+Requires: %{name}-core = %{version}-%{release}
+Requires: %{name}-debug
+Requires: %{?scl_prefix}rubygem(foreman-tasks) >= 0.8.5
```
domcleal: 1. Should this be in a `%package tasks` like other optional functionality, e.g. -ec2?
1. Dependency won't resolve, packages in the main repo can't depend on the contents of the plugin repo.

iNecas: My intention is to make sure tasks are installed by default for foreman runtime, which is something else, than installing foreman for building purposes, which is what this changes try to solve and from the comments it seemed like the issue

domcleal: > Where do we depend on foreman in mail repo? Isn't the solution to put the tasks into the main repo?

There is no dependency on `foreman` in the main repo that I can recall, but repoclosure would fail.

Moving the dependency may be valid, but that's not part of the PR?

iNecas: I put a commit to move tasks from plugins repo to the core ones

foreman/foreman.spec
```diff
@@ -625,7 +691,7 @@ Summary: Foreman sqlite support
 Group:  Applications/System
 Requires: %{?scl_prefix_ror}rubygem(sqlite3) >= 1.3.6
 Requires: %{?scl_prefix_ror}rubygem(sqlite3) < 1.4.0
-Requires: %{name} = %{version}-%{release}
+Requires: %{name}-core = %{version}-%{release}
```
domcleal: All of these database subpackages ought to depend on `%{name}` as installing foreman-sqlite should work to install the entire application. Certainly the installer does this and the manual may suggest it too.

This'll conflict with foreman-plugin though which depends on a database and is the place you're trying to break the circular dependency...

foreman/foreman.spec
```diff
@@ -301,6 +306,68 @@ BuildRequires: %{?scl_prefix}rubygem(facter)
...
+%files core
+%defattr(-,root,root,0755)
+%doc CHANGELOG Contributors LICENSE README.md VERSION
+%dir %{_datadir}/%{name}
+%{_datadir}/%{name}/app
+%exclude %{_datadir}/%{name}/app/assets
+%exclude %{_datadir}/%{name}/script/%{name}-debug.d
+%dir %{_datadir}/%{name}/bundler.d
+%exclude %{_datadir}/%{name}/bundler.d/development.rb
+%exclude %{_datadir}/%{name}/bundler.d/foreman_tasks.rb
```
domcleal: This repo is branched per release, including plugins - changes will not affect older releases.

rel-eng/tito.props
```diff
@@ -124,6 +126,7 @@ whitelist = foreman
   rubygem-gridster-rails
   rubygem-hammer_cli
   rubygem-hammer_cli_foreman
+  rubygem-hammer_cli_foreman_tasks
```
iNecas: I meant to move the hammer to core as well, but missed the comps. Good catch!

foreman/foreman.spec
```diff
@@ -49,9 +57,6 @@ Requires(preun): chkconfig
 Requires(preun): initscripts
 Requires(postun): initscripts
```
iNecas: Updated

foreman/foreman.spec
iNecas: I meant more the fact, that the spec for foreman-tasks would need to look differently for for <= 1.14 releases, unless I miss something?

foreman/foreman.spec
```diff
@@ -835,62 +900,6 @@ EOF
...
-%else
-%ghost %{_datadir}/%{name}/config/initializers/encryption_key.rb
-%endif
-%ghost %attr(0640,root,%{name}) %config(noreplace) %{_datadir}/%{name}/config/initializers/local_secret_token.rb
-# Only need tmpfiles on systemd (F17 and up)
-%if 0%{?rhel} > 6 || 0%{?fedora} > 16
-%{_prefix}/lib/tmpfiles.d/%{name}.conf
-%endif
 
 %pre
```
iNecas: they are meant to be in foreman, not core, moved the pre/post/preun depednencies

comps/comps-foreman-rhel7.xml
```diff
@@ -71,6 +71,8 @@
       <packagereq type="default">tfm-rubygem-fog-xenserver</packagereq>
       <packagereq type="default">tfm-rubygem-fog-xml</packagereq>
       <packagereq type="default">tfm-rubygem-font-awesome-sass</packagereq>
+      <packagereq type="default">tfm-rubygem-foreman-tasks</packagereq>
```
iNecas: Your right, added the other deps

foreman/foreman.spec
```diff
@@ -301,6 +306,68 @@ BuildRequires: %{?scl_prefix}rubygem(facter)
 BuildRequires: facter
```
iNecas: updated

foreman/foreman.spec
iNecas: I've updated the plugins to not depend on the foreman-sqlite, to be independent of the foreman. All the other packages depend on foreman now

iNecas:
I think at this point, we need presence from more people from @theforeman/core, I tried to address all the concerns raised in the comments, but this doesn't seem to lead anywhere

ares:
I'm afraid I got lost. @domcleal keeps mentioning circular dependency. @iNecas did you understand what he means? Is that build or runtime dependency? The only other option to avoid this packaging change that I can see is moving the foreman-tasks code into core. I'd still prefer this one.

iNecas:
There is no circular dependency that I'm aware of. The last technical issue was with the repo closure, that should be fixed by putting the packages into the foreman repos.

iNecas:
Based on IRC discussions with other people, it seems the will is to have the foreman-tasks codebase moved to be part of the core, which we aim for doing for foreman 1.15.  I will write more details description into foreman-dev list. However, this will take some time till done, and I would like to see the nightlies building.

In the mean time, would be viable solution to ship the foreman-tasks gem as part of the foreman core or keeping this change as temporary?

ehelms:
I've read through this, and looked at the code. I agree that all of the foreman-tasks packages should live in the main repository; however, I am not seeing the need for the introduction of a foreman-core package alongside the Foreman package. Especially if this is temporary until foreman-tasks codebase is merged into core proper. 

Wouldn't removing the requires on foreman from foreman-tasks, rebuilding it simplify the solution here to get builds moving again? I realize foreman-tasks technically requires Foreman to run, but I think it's safe to assume that a new requires on Foreman for foreman-tasks will be sufficiently safe.

mmoll:
@ehelms see #1437 for further discussion on that topic. While I think, temporary workarounds are acceptable, introducing a foreman-core package does IMHO not solve any problems.

iNecas:
@ehelms if we don't build the assets in the foreman-tasks package (because we would not have the build dependency on foreman), where would we do it and ship?

I'm right now thinking about putting the foreman-task gem as another source of the foreman package, and building and distributing it as part for foreman, but I need a feedback on this before moving forward.

ehelms:
I believe having the build dependency on the assets and plugin packages is
fine and shouldn't cause issues. The issue is more so the runtime requires
on Foreman. This is the scenario I was trying to mock build.

iNecas:
@ehelms the foreman-tasks having a runtime dependency on foreman was not an issue from the beginning. The issues why we are stuck here so long is the original cyclic dependency in build time (foreman build-depends on tasks and tasks build-depend on foreman), so you could not build without having a least one version of previous package. The foreman-core did break the cyclic dependency, but was not accepted.

The only outcome from this was the tasks need to get into foreman. And I'm proposing, to fix the builds to include the foreman-tasks gem as part of the foreman rpm for now, similarly as we have in the deb part.

ehelms:
@iNecas I think I am missing something as I am not seeing the part "tasks build depend on foreman". As far as I can tell, tasks build depends on foreman-assets and foreman-plugin, but runtime depends on Foreman which would cause the cyclic dependency once Foreman also requires foreman-tasks (since foreman-tasks would then try to require Foreman when the Foreman spec requires foreman-tasks).

iNecas:
If you look at the code before this PR, the foreman-plugin was depending on foreman, I tried to break the cycle using the foreman-core approach, but this turned out to be a no-go, therefore the issue is still there. I'm working not on getting the tasks shipped with the Foreman RPM, similarly as debian change is proposed

ehelms:
@iNecas I think ou are correct, and it would at least unify the approaches. Please ping me once you pushed an update with that strategy.

iNecas:
So far, I've hit two issues with applying the approach with putting the tasks gem inside the foreman srpm:

   * the combination of annexed and fetched sources in nightlies is not supported by tito
   * we don't have the dependencies of tasks available in core rpms, so it's failing right now

Although, we could overcome this issues, BTH the more I tried that, the more unnatural it felt, and we were actually getting the worst of both worlds: complicated release vs. having the source code in separate repositories.

Therefore, at this moment (which might be influence by the hour I'm writing this:), I lean towards making the foreman-tasks usage in the core optional, perhaps by using a unified interface (ActiveJob?), so that it works in both modes and we could work under a bit less pressure on getting the tasks into the core the proper way, without hacks and blockings from their work.

ehelms:
A couple of other thoughts on how we could unstick this for now, given its blocking a lot of other processes and thats a bigger win than getting this "right" the first time:

 1) simply don't have the Foreman RPM require foreman-tasks, cover this by ensuring installer installs foreman-tasks given running the installer is 99% of users (especially for nightly) as a temporary nightly only work around (block release on this getting fixed properly)
 2) extract "plugin" related code to a library so that plugins can require it instead of the entire Foreman stack during RPM build (I already prototyped this for the plugin assets rake task and got it working as a POC)
 3) move Foreman code to use old non-async path if ForemanTasks is not available for now and use new code paths if it is available

iNecas:
Notes from IRC communication: the 2) is probably not feasible, as the api documentation for plugins needs to have the foreman code around (including the runtime). This would also mean all the plugins in the ecosystem would need to get updated for this change and I'm no sure we could justify it for this reason.

iNecas:
Thanks all involved in the discussion around this, I've just opened a revert of making tasks as core dependency https://github.com/theforeman/foreman/pull/4217. Although it doesn't look like it at the first sight, I'm happy we had this conversation that I believe will lead to better results and less bumps both on foreman-core, as well as plugins that already rely on the foreman-tasks and we want to support until the tasks are not in core.

PR #1437, opened 2016-11-25
Refs #15779 - handle foreman-tasks in core (DEB)

mmoll:
this should be built into:

* [x] Nightly

@iNecas @ares have a look, please.

### DIFF
--- a/debian/jessie/foreman/control
+++ b/debian/jessie/foreman/control
@@ -14,7 +14,7 @@ Section: web
 Priority: extra
 Depends: ruby2.1, bundler (>= 1.3), zlib1g-dev,
          ruby2.1-dev, build-essential,
-         rake (>=0.8.4), ${misc:Depends}
+         rake (>=0.8.4), ruby-foreman-tasks (>= 0.8.5), ${misc:Depends}
 Recommends: foreman-proxy, foreman-debug, python-minimal
 Conflicts: ruby-activerecord-deprecated-finders
 Description: Systems management web interface
--- a/debian/jessie/foreman/rules
+++ b/debian/jessie/foreman/rules
@@ -11,7 +11,7 @@ build:
 	#
 	# Build the gem cache
 	#
-	/bin/rm -f bundler.d/therubyracer.rb
+	/bin/rm -f bundler.d/foreman_tasks.rb
 	/bin/cp config/settings.yaml.example config/settings.yaml
 	/bin/cp config/database.yml.example config/database.yml
 	/bin/cp config/email.yaml.example config/email.yaml
--- a/debian/trusty/foreman/control
+++ b/debian/trusty/foreman/control
@@ -13,7 +13,7 @@ Architecture: any
 Section: web
 Priority: extra
 Depends: ruby2.0, bundler (>= 1.3.5-2ubuntu2), ruby2.0-dev,
-         zlib1g-dev, build-essential,
+         zlib1g-dev, build-essential, ruby-foreman-tasks (>= 0.8.5),
          rake (>=0.8.4), ${misc:Depends}
 Recommends: foreman-proxy, foreman-debug, python-minimal
 Description: Systems management web interface
--- a/debian/trusty/foreman/rules
+++ b/debian/trusty/foreman/rules
@@ -11,13 +11,13 @@ build:
 	#
 	# Build the gem cache
 	#
-	/bin/rm -f bundler.d/therubyracer.rb
+	/bin/rm -f bundler.d/foreman_tasks.rb
 	/bin/cp config/settings.yaml.example config/settings.yaml
 	/bin/cp config/database.yml.example config/database.yml
 	/bin/cp config/email.yaml.example config/email.yaml
 	/usr/bin/npm install npm
 	/usr/bin/nodejs node_modules/.bin/npm install --no-optional
-	/usr/bin/ruby2.0 /usr/bin/bundle pack --all
+	/usr/bin/ruby2.0 /usr/bin/bundle package
 	/usr/bin/ruby2.0 /usr/bin/bundle exec rake locale:pack assets:precompile RAILS_ENV=production
 	/usr/bin/ruby2.0 /usr/bin/bundle exec rake webpack:compile
 	/usr/bin/ruby2.0 /usr/bin/bundle exec rake db:migrate --trace RAILS_ENV=production
--- a/debian/xenial/foreman/control
+++ b/debian/xenial/foreman/control
@@ -15,7 +15,7 @@ Section: web
 Priority: extra
 Depends: ruby2.3, bundler (>= 1.3), zlib1g-dev,
          ruby2.3-dev, build-essential,
-         rake (>=0.8.4), ${misc:Depends}
+         rake (>=0.8.4), ruby-foreman-tasks (>= 0.8.5), ${misc:Depends}
 Recommends: foreman-proxy, foreman-debug, python-minimal
 Conflicts: ruby-activerecord-deprecated-finders
 Description: Systems management web interface
--- a/debian/xenial/foreman/rules
+++ b/debian/xenial/foreman/rules
@@ -11,7 +11,7 @@ build:
 	#
 	# Build the gem cache
 	#
-	/bin/rm -f bundler.d/therubyracer.rb
+	/bin/rm -f bundler.d/foreman_tasks.rb
 	/bin/cp config/settings.yaml.example config/settings.yaml
 	/bin/cp config/database.yml.example config/database.yml
 	/bin/cp config/email.yaml.example config/email.yaml


debian/jessie/foreman/control
```diff
@@ -14,7 +14,7 @@ Section: web
 Priority: extra
 Depends: ruby2.1, bundler (>= 1.3), zlib1g-dev,
          ruby2.1-dev, build-essential,
-         rake (>=0.8.4), ${misc:Depends}
+         rake (>=0.8.4), ruby-foreman-tasks (>= 0.8.5), ${misc:Depends}
```
domcleal: Not sure this should be a separate dependency - it's likely to cause problems adding a circular dependency. The bundler.d file is already provided so it should just be vendored like other dependencies.

mmoll:
We need to check for the circular dependency, but i hope that dpkg will just install both and configure both in one run... Regading to leaving it in core, I guess the -tasks fellows won't like that we would bring a specific version of -tasks with the core package that hard/impossible to update by a -tasks plugin package.

ares:
Right, we'd have to wait for Foreman update to update foreman-tasks. I think it would make most sense to remove Foreman from foreman-tasks dependencies to avoid circular deps.

mmoll:
the circular dependency isn't a problem, but the assets: `trying to overwrite '/var/lib/foreman/public/assets/trigger_form-9434ffad72fa0fc76bfd09f66a0fc4710a24b243aa57d3cd954ada6852552162.js.gz', which is also in package foreman 9999-xenial+scratchbuild+201611251751
`

ares:
In the worst case, we could possibly remove assets in the same way just after the compilation.

mmoll:
I'm trying with the initial idea of removing foreman_tasks.rb at the beginning now.

mmoll:
seems to to work for me.

domcleal:
> We need to check for the circular dependency, but i hope that dpkg will just install both and configure both in one run...

It's no longer possible to build foreman and ruby-foreman-tasks from scratch, it creates a bootstrapping issue.

> Regading to leaving it in core, I guess the -tasks fellows won't like that we would bring a specific version of -tasks with the core package that's hard/impossible to update by a -tasks plugin package.

I don't think that's an issue, if it's added to Foreman then it has to be on Foreman's release cycle now.

ares:
> I don't think that's an issue, if it's added to Foreman then it has to be on Foreman's release cycle now.

I don't get it. Why foreman-tasks is considered as a part of Foreman? The only difference from any other dependency is that foreman-tasks needs Foreman in build phase correct? If that somehow implied foreman-tasks being part of Foreman, would it make more sense to copy the code to Foreman project?

domcleal:
> > I don't think that's an issue, if it's added to Foreman then it has to be on Foreman's release cycle now.
> 
> I don't get it. Why foreman-tasks is considered as a part of Foreman? The only difference from any other dependency is that foreman-tasks needs Foreman in build phase correct?

Right, and every other dependency is only ever updated when the 'foreman' package is rebuilt as they're all vendored, so it is the same. This is my point, which is that I don't see why it's in a separate package if it's a dependency.

> would it make more sense to copy the code to Foreman project?

Or remove it, probably.

ares:
> Right, and every other dependency is only ever updated when the 'foreman' package is rebuilt as they're all vendored, so it is the same. 

I thought this was a workaround until we package things properly. We started with plugins, now we have first dependency properly built so we are searching for a way to both vendor some packages and properly package some Foreman deps. Normally we don't pin specific dependency versions, this is just an outcome of the workaround..

mmoll:
Additionally moved the foreman package in ruby-foreman-tasks from "Depends" to "Recommends". It's not pretty, but I think with all these changes together (and as theforeman/foreman#4075 got merged) everything should be OK.

domcleal:
The circular dependency still exists because it's on the _build dependency_ of ruby-foreman-tasks as well as the runtime dep ("fixed" via Recommends). `ruby-foreman-tasks` requires `foreman` to build, `foreman` requires `ruby-foreman-tasks`.

Per #1436, the dependency of ruby-foreman-tasks should probably be on a subpackage rather than `foreman`, or vendor it like other deps. (Though I remain very sceptical about the necessity of this dependency.)

mmoll:
another try: the only change here would be to add ruby-foreman-tasks as recommendation to the foreman package.

Sure, we can create a foreman-tasks subpackage and add the init script for the dynflow executor, but as described above updating foreman-tasks is not possible then. Also I guess the versioning will be problematic (foreman-tasks vs. foreman version).

mmoll:
The Recommendation could be demoted further to "Suggests", which would be kind of in-line to some things only needed by specific foreman-proxy features, which still are included with the foreman-proxy package code-wise.

As said, the situation isn't pretty, however I think the current diff is the least invasive possibility, which also doesn't need changes to the whole installer machinery.

The other possibility (which I originially had in mind, when it was about "adding -tasks to core") would be to deprecate -tasks as plugin and really integrate it into Foreman core sources. However, this seems not to be the desired way the go for the -tasks people.

ares:
I'm happy to copy the code from plugin, only thing I'm afraid of is neverending review of such PR rejected at the end. I still don't fully understand why new foreman-core package discussed in rpm PR does not solve the issue.

mmoll:
I didn't have a real look onto the RPM side, but to tell (very condensed) the problem for the DEB side:

* The foreman-tasks package does need the foreman package for building itself, because the whole foreman core is needed, e.g. for the whole `bundle package` dance, asset precompilation, apipie docs generation and so on...
* As you can see in this diff here, I'm deleting the foreman-tasks gem from the foreman core package to have the same situation as before -tasks came into core.
  * This is a horrid hack.
  * This only can work as long as nothing needed in the foreman core package building process is using anything provided by -tasks or anything brought in by -tasks (dynflow and friends).
  * As soon as that would change (one can not see this in any core PR test or similar!), building the foreman core package that way will fail.
   * Pulling in -tasks into core by not deleting it before building the core package makes it impossible to have a separate -tasks plugin package.
* A foreman-core (or similar) package does not help, because this is all about package **build** time dependencies and packages must still be **bootstrapable** on a clean system:
  * If I have only a .tar.gz of the foreman source and the debian directory from this repo here, I have to be able to reproduce the build, after having installed all the Debian packages listed as build-dependencies.
  * Same for the foreman-tasks plugin package, if there is one, which requires the foreman core package for building. If I need an already packaged foreman-tasks DEB to build foreman, there's an unsolvable circular dependency there.

IMHO the only clean way would be to integrate the -tasks source into Foreman core in the 1.15 release cycle, but I could live with the workaround done in this PR, despite all the dirtyness... at least until the foreman core DEB build gets broken by changes in core, leading to a -tasks presence requirement in the package building process.

ares:
First of all, thanks for you patience and detailed explanation! I understand the potential issue if foreman-tasks would be removed from rules before all the tasks. I agree that it would be hard to catch in core PR that someone uses something from foreman-tasks e.g. on assets level.

But I still don't get the circular dependency problem with foreman-core. Imagine there is foreman-core which is the same package as foreman today, without any dependency on foreman-tasks. Then we'd have foreman-tasks package which would have build dep on foreman-core so its assets could be compiled. Since foreman-core would not depend on tasks, that should be fine. This foreman-core package would not require foreman-tasks for it's build phase but would still have the problem of "what if someone relies on tasks assets". A new foreman package would only need foreman and foreman-tasks as runtime deps and would contain only bundler.d/foreman-tasks.rb, so no need for build deps. The only difference is that we wouldn't delete the foreman-tasks from bundler.d which I agree is not too big.

So I start lean towards moving the code into core codebase. I think I'll start a thread on foreman-dev so I get opinions before I invest too much effort into it. If we get positive feedback I hope this could be merged to get nightly builds back until it's part of core.

mmoll:
So, foreman-core is the same as foreman today content-wise, but with an unlisted dependency? :)

mmoll:
BTW, another solution (for DEB only, I guess) would be not to delete `bundler.d/foreman_tasks.rb`, so it's getting into the core package and define a foreman-tasks subpackage which includes this .rb and an init script for the executor and then delete the ruby-foreman-tasks plugin package. This also means -tasks can not get updated independently.

ares:
> So, foreman-core is the same as foreman today content-wise, but with an unlisted dependency? :)

Normally it would be runtime dependency but that unfortunately becomes build dependency too since rake tasks loads rails environment. That's why we'd workaround it with second package. The other way would be to move tasks specific code to the new foreman package but that would become hard to maintain I'm afraid.

ares:
> This also means -tasks can not get updated independently.

I think at this point it does not matter any more. If we incline to move the code into core, it would be the same.

mmoll:
> That's why we'd workaround it with second package.

I don't think that's better (or worse) than the hacky workaround done here. It even has the same implications ("...but would still have the problem of "what if someone relies on tasks assets"") but introduces another package, and after the whole mess of having also to maintain {tasks, rex, ansible, ...}-core I'm not eager to have another package around.

mmoll:
>>    This also means -tasks can not get updated independently.

> I think at this point it does not matter any more. If we incline to move the code into core, it would be the same.

Allright, then it would be nice to know if the code move will happen, otherwise I'd need to start working on the foreman-tasks subpackage for DEB and puppet-module changes. However, I'm still unsure if/how that could be mapped into RPM packaging.

mmoll:
#1493 should resolve the outstanding problems and superseds this PR.

iNecas:
https://github.com/theforeman/foreman/pull/4217 reverts the changes in core as they were done originally, which means the tasks are not getting dependency of core. The changes that will follow in core will start adding the tasks functionality into the core piece by piece. After the revert is merged, there should be no need for this anymore.

@mmoll I know you did a lot of work that could enable the original approach and I really appreciate 
it. I hope we will meet each other soon at some more optimistic PRs :)

mmoll:
allright :)

PR #1438, opened 2016-11-28
Refs #17491 - use nodejs PPA on ARMv8

mmoll:
This should be built into:

* [x] Nightly

after theforeman/foreman-infra#260


domcleal:
Thanks @mmoll.

PR #1440, opened 2016-12-02
Update foreman-ansible to 1.3.0 [rpm]

dLobatog:

* [x] Nightly
* [x] 1.13
* [x] 1.12
* [ ] 1.11

---

Not sure about 1.12, because foreman_ansible_core wouldn't resolve the tasks-core dependency there. If it can't be built, then just build it for 1.13 and develop. @iNecas is there any problem with having tasks-core in 1.12?


domcleal:
Merged and building for 1.13 and nightly. If you want to do 1.12, please sort out all of the new packages in another PR (if that's a sane thing to do at this point in time..).

PR #1442, opened 2016-12-06
Update hammer_cli_foreman_admin to 0.0.6

lzap:
And no work for Debian as this is a dependency.


domcleal:
Please indicate which releases this should be built for, it should have been in the PR template.

> And no work for Debian as this is a dependency.

It isn't packaged for Debian, I don't think it's a dependency of anything.

lzap:
I don't mind 1.13 but nightly is ok. I don't use github for filing PRs, sorry haven't noticed that.

domcleal:
Merged, thanks @lzap - building for nightly only. Not sure what you mean by don't mind, if you want it in 1.13 or 1.14, please say so.

lzap:
Thanks, not important updates in this one.

PR #1443, opened 2016-12-07
Add puppet-agent-puppet-strings, yard [deb]

domcleal:
New packages to add puppet-strings gem into AIO environment, so we can later switch foreman-installer to build on these packages.


mmoll:
merged as fcc1211 and d01657e, thanks @domcleal!

PR #1450, opened 2016-12-12
update apipie-bindings to 0.0.19 (DEB)

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.14


domcleal:
Merged and building, thanks @mmoll. I'm only doing this for nightly for now, will consider for 1.14.0-RC2 after it's been in nightly first, assuming there may be a new Hammer CLI release requiring it.

PR #1452, opened 2016-12-15
Support Ruby 2.3 in production for RPM based systems

ehelms:

* [x] Nightly
* [ ] 1.14
* [ ] 1.13
* [ ] 1.12

---

Opening this early in 1.15 lifecycle for consideration. This change aims to make three changes:

 * use Ruby 2.3 in production through rh-ruby23
 * switch to rh-ror42 since it depends on rh-ruby23 by default
 * move from EPEL nodejs to rh-nodejs4 SCL which is included by default by rh-ruby23 and provides a newer nodejs to rely on

The testing of the foreman build is currently blocked by the foreman-tasks packaging resolution; however the following can be tested currently:

  * the nodejs packaging can be assessed and tested
  * the dependent rubygems can be tested for rebuild against new SCLs
  * the Foreman RPM itself can be "test" built noting that it errors at the end the same way current foreman develop errors


foreman/foreman.spec
```diff
@@ -671,9 +669,12 @@ cp config/settings.yaml.example config/settings.yaml
 sed -i 's/:locations_enabled: false/:locations_enabled: true/' config/settings.yaml
 sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml
 export BUNDLER_EXT_GROUPS="default assets"
-ln -s %{nodejs_sitelib} node_modules
+ls -la %{?_scl_prefix}/tfm/root/%{nodejs_sitelib}/
```
ehelms: Artifact from trying to verify this path

foreman/foreman.spec
```diff
@@ -671,9 +669,12 @@ cp config/settings.yaml.example config/settings.yaml
...
+ln -s %{?_scl_prefix}/tfm/root/%{nodejs_sitelib} node_modules
```
ehelms: This is kind of ugly since it does not rely on a singular, pre-defined macro for this location. I reached a lack of knowledge point with this change in terms of how to address it. Ideas that crossed my mind:

 * define new macro in the `tfm` package to represent this value
 * require `tfm-build` and use `%{?_scl_root}`
 * other

ehelms:
Breaking this down into smaller changes, starting with switching to rh-ror42 and rh-ruby23 -- https://github.com/theforeman/foreman-packaging/pull/1528

PR #1459, opened 2016-12-20
Remove autoprefixer-rails

dLobatog:
After https://github.com/theforeman/foreman/pull/4086 ,
autoprefixer-rails is no longer needed

* [x] Nightly
* [ ] 1.14
* [ ] 1.13
* [ ] 1.12

---


domcleal:
> After theforeman/foreman#4086 ,
> autoprefixer-rails is no longer needed

It's a dependency of bootstrap-sass. The only thing that appears to have changed is that the explicit dependency in Foreman has been removed, not the gem.

dLobatog:
Ah darn, my mistake. Closing

PR #1461, opened 2016-12-21
Fixes #17781 - update fog-aws to 0.13.0

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.14
* [x] 1.13

See http://projects.theforeman.org/issues/17781 for details.


domcleal:
Merged, thanks @mmoll.

PR #1477, opened 2017-01-03
Refs #13455 - Require genisoimage for -libvirt (RPM)

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.14
* [x] 1.13 (if there will be a 1.13.4 release)


domcleal:
Merged, thanks @mmoll.

PR #1483, opened 2017-01-06
Discovery 8.0.0 (DEB)

lzap:
* [x] Nightly
* [x] 1.14

Thanks.


domcleal:
1. Please use a branch with a `deb/` prefix, you will need to re-open the PR.
1. `deface` has been removed as a dependency, it needs removing from the package dependencies.

domcleal:
The minimum Foreman version should also be increased to 1.14.0, per engine.rb.

lzap:
What a disaster again, replaced with new PR.

PR #1485, opened 2017-01-09
drop special gem installation steps (DEB)

mmoll:
this should be built into:

* [x] Nightly


domcleal:
Merged, thanks @mmoll.

PR #1491, opened 2017-01-16
Only use Puppet 3.x to build foreman-installer [deb/1.14]

domcleal:
Depending on puppet-agent is insufficient as it also needs
puppet-strings to build, so prefer Puppet 3.x, so builds will fail when
puppet-agent becomes available in our pbuilder configuration.

(applicable to deb/1.14, deb/1.13, should be merged before https://github.com/theforeman/foreman-infra/pull/270)


mmoll:
merged, thanks @domcleal!
also cherry-picked as 69ca837c for deb/1.13

PR #1492, opened 2017-01-16
Use puppet-agent-puppet-strings package to build installer [deb]

domcleal:
Switches to Puppet 4 (puppet-agent) for building, which supports
YARD-based documentation but drops parsing of validation functions
from manifests.

Depends on https://github.com/theforeman/foreman-infra/pull/270.


domcleal:
[test] with pbuilder modifications.

mmoll:
As the nightly build failed (for whatever reason puppet-agent instead of puppet was installed), this gets things going again... merged, thanks @domcleal!

PR #1493, opened 2017-01-16
Refs #15779 - handle foreman-tasks in core (DEB)

mmoll:
This should be built into:

* [x] Nightly

This obsoletes the ruby-foreman-tasks plugin DEB package, which will need to get removed after this (and plugins will need their dependencies updated).

This PR should resolve the bootstrapping problem of the plugin approach.


mmoll:
@domcleal @iNecas @ares @ehelms I think this is the cleanest way to handle foreman-tasks in core with the current situation. However, I can't think of a way to apply this to RPM.

iNecas:
Let me try achieving similar thing for RPM today.

PR #1507, opened 2017-01-27
Refresh settings.yml patch for bind_host default change [deb]

domcleal:


mmoll:
merged, thanks @domcleal!

PR #1509, opened 2017-01-27
Make foreman-selinux depend on container-selinux

dLobatog:
foreman-selinux used to define a type docker_port_t because there was no
official policy that defined this type.

container-selinux has this type too, and both policies cannot be
installed together. Since container-selinux is the default policy on EL
systems for container stuff, foreman-selinux shouldn't stop it from
being installed, and depend on it instead for the types it needs.

https://github.com/projectatomic/container-selinux/blob/master/container.te#L77 is where it's being defined. I see that at least version 2.0 is needed to get it, and I can't find RPMs earlier than that. 

https://github.com/theforeman/foreman-selinux/pull/66 removes the policy on foreman-selinux so it depends on it.

@rhatdan are we OK making the dependency on 2.0 here or there's 1.0 RPMs with this type too?

-----------------------------------------------

* [x] Nightly
* [x] 1.14
* [x] 1.13
* [x] 1.12

---

### DIFF
--- a/foreman-selinux/foreman-selinux.spec
+++ b/foreman-selinux/foreman-selinux.spec
@@ -51,6 +51,7 @@ BuildRequires:  /usr/bin/pod2man
 BuildArch:      noarch
 
 Requires:           selinux-policy >= %{selinux_policy_ver}
+Requires:           container-selinux >= 2.0
 Requires(post):     /usr/sbin/semodule, /sbin/restorecon, /usr/sbin/setsebool, /usr/sbin/selinuxenabled, /usr/sbin/semanage
 Requires(post):     policycoreutils-python
 Requires(post):     selinux-policy-targeted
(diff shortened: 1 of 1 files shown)


foreman-selinux/foreman-selinux.spec
```diff
@@ -130,6 +131,7 @@ Summary: SELinux policy module for Foreman Proxy
 Group:   System Environment/Base
 
 Requires:           selinux-policy >= %{selinux_policy_ver}
+Requires:           container-selinux >= 2.0
```
domcleal: Why does this subpackage depend on container-selinux? It doesn't use any of it.

foreman-selinux/foreman-selinux.spec
```diff
@@ -51,6 +51,7 @@ BuildRequires:  /usr/bin/pod2man
 BuildArch:      noarch
 
 Requires:           selinux-policy >= %{selinux_policy_ver}
+Requires:           container-selinux >= 2.0
```
domcleal: 1. Not sure this is necessary if the policy is optional.
1. container-selinux is not present in the base EL7 OS, so it probably shouldn't be a dependency.

foreman-selinux/foreman-selinux.spec
rhatdan: No reason I can think of for requiring the package, please explain why you think you need it?

ehelms:
@dLobatog What should we do with this?

PR #1518, opened 2017-02-03
templates: 4.0.1 release (DEB)

GregSutcliffe:
Just for latest stable please:

* [x] Nightly
* [x] 1.14

Thanks!

### DIFF
--- a/plugins/ruby-foreman-templates/debian/control
+++ b/plugins/ruby-foreman-templates/debian/control
@@ -8,7 +8,7 @@ Homepage: https://github.com/theforeman/foreman_templates
 
 Package: ruby-foreman-templates
 Architecture: all
-Depends: ${misc:Depends}, bundler, foreman (>= 1.12.0), git
+Depends: ${misc:Depends}, bundler, foreman (>= 1.14.1), git
 Description: Foreman Templates Plugin
  This plugin will sync the contents of the Foreman Community Templates
  repository (or a git repo of your choice) to your local Foreman instance.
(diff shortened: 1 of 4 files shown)


mmoll:
If this is compatible with older versions, why not build for these, too?

GregSutcliffe:
@mmoll good catch, ty. Min requirement updated in the plugin (sadly requiring a version bump) and deb files should match.

plugins/ruby-foreman-templates/debian/control
```diff
@@ -8,7 +8,7 @@ Homepage: https://github.com/theforeman/foreman_templates
 
 Package: ruby-foreman-templates
 Architecture: all
-Depends: ${misc:Depends}, bundler, foreman (>= 1.12.0), git
+Depends: ${misc:Depends}, bundler, foreman (>= 1.14.1), git
```
mmoll: should that be 1.14.0?

GregSutcliffe: No, there was a minor fix to exporting capabilities on the Foreman core side that we make use of.

mmoll: ok, thanks

ares: it would work with 1.14 but it would be missing "model" in metadata, but I guess that's minor thing (literally :-) since 1.14.1 is out anyway

mmoll:
> I guess that's minor thing (literally :-)

I See...

mmoll:
merged, thanks @GregSutcliffe!

PR #1528, opened 2017-02-13
Switch to using rh-ruby23

ehelms:

* [x] Nightly
* [ ] 1.14
* [ ] 1.13
* [ ] 1.12

---

### DIFF
--- a/tfm/tfm.spec
+++ b/tfm/tfm.spec
@@ -5,10 +5,9 @@
 %{!?scl_vendor_in_name: %global scl_vendor_in_name 0}
 %{?scl_package:%scl_package %scl}
 
-# Fallback to sclo-ror42 etc. when scldevel's not in the buildroot
 %{!?scl_ror:%global scl_ror sclo-ror42}
 %{!?scl_prefix_ror:%global scl_prefix_ror %{scl_ror}-}
-%{!?scl_ruby:%global scl_ruby rh-ruby22}
+%{!?scl_ruby:%global scl_ruby rh-ruby23}
 %{!?scl_prefix_ruby:%global scl_prefix_ruby %{scl_ruby}-}
 
 # Do not produce empty debuginfo package.


ehelms:
Given this switches away from the current CentOS SCLO for Rails 4.2, are there any concerns? I'd like to get this into nightly for testing prior to the 1.15 release.

dLobatog:
It appears that sclo-ror42 was meant to be a way for us to run that stack before the rh-ror42 SCL was available ([source](https://www.redhat.com/archives/sclorg/2015-November/msg00021.html)). If that's the case, and all packages are available in the rh-ror42, I'm 👍 to this change.

domcleal:
I don't think a dependency on rh-ror42 is an improvement over sclo-ror42. Although it may depend on a newer Ruby version, it has a closed development process and is missing fixes that sclo-ror42 ships. I think Foreman should prefer SCLo collections over RHSCL collections wherever possible.

ehelms:
This has been updated to start with just updating the Ruby dependency to rh-ror23

domcleal:
That doesn't make much sense, there will be dependencies on _two_ different Ruby collections.

ehelms:
I'm confused. If you are suggesting that sclo-ror42 is what needs to update the Ruby dependency it has then please just say that.

domcleal:
Well, I wouldn't go so far as to suggest that, it'd be a disruptive and incompatible change to the SCL. I'm only saying that the Ruby collection that `tfm` is built against (from the buildroot or these macros) must match the dependency on the RoR collection (sclo-ror42).

Changing this one line will cause tfm-built packages to depend on rh-ruby23, while sclo-ror42 packages depend on rh-ruby22, and so half of the libraries won't be available at runtime because they're in the other Ruby paths. If I'm misunderstanding this and you say it really does work, I'll re-review.

ehelms:
Hrmm, I see what you mean now in terms of what the sclo-ror42 collection is built against currently. In order to get a collection that updates the underlying Ruby dependency, would you suggest a new collection or to treat the collection like a project and bump the major version of the collection and rebuild?

domcleal:
Changing the major version of the metapackage will only affect dependencies on the metapackage itself (not that the collection was ever documented as having major version changes like that), so the issue will remain everywhere there are dependencies on specific packages within sclo-ror42. It would probably have to be a new collection.

PR #1531, opened 2017-02-16
Update foreman_hooks to 0.3.13 [deb]

domcleal:

* [x] Nightly
* [x] 1.14
* [x] 1.13


mmoll:
`/usr/share/foreman/script/foreman-debug.d` seems to belong to the `foreman-debug` package, so maybe a mkdir when it doesn't exist could catch errors when `foreman-debug` is not installed (although I didn't test this).

domcleal:
You're correct, it would cause an error processing postrm. I've added a `mkdir` into the postrm now.

mmoll:
merged, thanks @domcleal!

PR #1532, opened 2017-02-19
Use enumerate rather than doing it manually

ekohl:


domcleal:
Merged, thanks @ekohl.

PR #1541, opened 2017-02-22
Add katello packaging

ehelms:
This is a decently large addition but does not alter the behavior of this repository. Instead, the aim is to begin introducing and centralizing the Katello plugin builds with Foreman given that Katello is a plugin and part of the community. This changed would:

 * keep the katello tags, comps and generated Koji repositories separate from Foreman and Plugins
 * the repositories starting with nightly and 3.4 would be changed to land on yum.theforeman.org
   -- this would centralize repositories for users and allow statistics gathering for @GregSutcliffe 
 * the repositories, per version that would be introduced would continue to follow their current structure s they are (e.g. https://fedorapeople.org/groups/katello/releases/yum/3.3/)
    -- this would mean a katello, pulp, candlepin and set of client repositories


domcleal:
Please keep this in the Katello packaging repos, the project is large enough to warrant it.

ehelms:
Are you against the repositories living on `yum.theforeman.org` or just the merging the source code repositories?

domcleal:
Both.

ohadlevy:
@domcleal while I understand the scope of katello packaging is large, I do want to find a way to host it on yum.tf.org. is your objection technical ? (e.g. concerned about resources - disk space, network etc)

ohadlevy:
@domcleal bump?

domcleal:
No, I don't think it has the resources - but that's not really the subject of this PR.

ehelms:
I think both are the subject. The perceived "burden" that adding in Katello's 20 some odd packages to the repository has and whether yum.theforeman.org can support it resource wise. The plugin repository is large in scope than Katello, so if it's a question of size then we ought to move plugins off to their own.

If the alternative was to keep everything in katello-packaging but deploy to yum.theforeman.org would there be any protest?

ares:
I'm sorry but I don't understand the "burden" argument. I would think that people who worked on katello-packaging so far could get permissions so they would continue working on same packages but at foreman-packaging. And after a while, if the process is the same, I don't see a reason why people wouldn't work on other packages as well. I think that would be beneficial for everybody.

johnpmitsch:
+1 from me to moving Katello packaging to this repository and further centralizing foreman/katello resources. I am curious to hear the arguments against doing so elaborated.

dLobatog:
I'm also in favor of moving them over, unless there are any good arguments against it.

iNecas:
:+1: on the direction. If anyone is aware of the issues we will start facing after doing this, and are not around at the moment, let's articulate them and try to find solutions. Keeping the current state forever is not a solution for me.

ekohl:
I would like to note that we should maintain the quality standards the foreman has on this repo. That means there should be a thorough review of all packages being merged.

ehelms:
@ekohl thats fair, and would mean someone needs to step up to do the review since I authored and/or reviewed a large chunk of this PR originally from the katello-packaging repo.

ekohl:
@ehelms I think currently only @domcleal really has enough knowledge to properly review it.

ehelms:
@ekohl Then I don't think this is going to get in.

ares:
> @ekohl Then I don't think this is going to get in. 

Given the number of :+1: in this PR that would make no sense IMHO.

ehelms:
@ekohl I think it is worth noting that when we created katello-packaging we created it to follow the principles and workflows that foreman-packaging had already established. From that standpoint, it works and feel the same as packages in foreman-packaging do.

If there is concern that we'd be merging this in and then expecting the maintainers of the repository (of which I am one) to take on new burden I am can say that is 100% not the goal. At a minimum, myself and @jlsherril have personally reviewed and maintained katello-packaging and will continue to shepherd the packages added by katello to this repository.

komidore64:
bump

PR #1544, opened 2017-02-28
fix build errors seens with with node-sass (DEB)

mmoll:
this should be built into:

* [x] Nightly


mmoll:
This fixes the initial node-sass related build error on Debian/jessie. The others point to lacking isolation of the i386 build from the x86_64 one, but maybe we do want to drop i386 builds for 1.15 anyway?

domcleal:
I don't think dropping support is necessary, the node_modules dir can either be deleted or node-sass reinstalled before `npm install` is run again.

mmoll:
Allright, let's see, where this build is going now.

mmoll:
💚 again. Compiling the C++ stuff on ARM adds about tens minutes to the build, but I guess we don't have any alternative until node-sass provides prebuilt binaries also for these architectures.

domcleal:
Merged, thanks @mmoll.

PR #1546, opened 2017-03-02
update foreman_xen to 0.5.0 (RPM)

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.14

While I think the plugin is still compatible with 1.13, it's better to only retag it into 1.14 support-wise.


rubygem-foreman_xen/rubygem-foreman_xen.spec
```diff
@@ -55,25 +58,22 @@ Summary:    Documentation for rubygem-%{gem_name}
...
-cp -a .%{gem_dir}/* \
-        %{buildroot}%{gem_dir}/
+cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
 
-mkdir -p %{buildroot}%{foreman_bundlerd_dir}
-cat <<GEMFILE > %{buildroot}%{foreman_bundlerd_dir}/%{gem_name}.rb
-gem '%{gem_name}'
-GEMFILE
+%foreman_bundlerd_file
+%foreman_precompile_plugin -a -s
```
domcleal: `-a` (API docs) is unnecessary

rubygem-foreman_xen/rubygem-foreman_xen.spec
```diff
@@ -82,10 +82,10 @@ GEMFILE
 %exclude %{gem_instdir}/Rakefile
 %exclude %{gem_instdir}/test
 %{gem_spec}
-%{foreman_bundlerd_dir}/%{gem_name}.rb
-%doc %{gem_instdir}/LICENSE
+%{foreman_bundlerd_plugin}
 
 %exclude %{gem_cache}
+%{foreman_assets_plugin}
```
domcleal: Nitpick: please move this up into the main list of files, above the excludes etc.

mmoll:
assets seems to get generated wrong somehow?

domcleal:
The JS files included under app/assets/ ought to be in a plugin subdirectory (app/assets/foreman_xen/) but they are not. That's probably unsupported by Foreman though for compute resources.

You may have to expand %foreman_assets_plugin into `%{gem_instdir}/public/assets`, or preferably more specifically `%{gem_instdir}/public/assets/compute_resources/xenserver`.

mmoll:
Should be fixed now, `foreman_assets_plugin` is still needed for the JSON file. I'll have a look into proper namespacing for one of the next releases, as all compute resource plugins would have that problem anyway.

domcleal:
[test] with rebuilt Foreman RPM on F24.

domcleal:

PR #1547, opened 2017-03-02
update foreman_xen to 0.5.1 (DEB)

mmoll:

* [x] Nightly
* [x] 1.14

While I think the plugin is still compatible with 1.13, it's better to only cherry-pick it into 1.14 support-wise.

### DIFF
--- a/plugins/ruby-foreman-xen/debian/control
+++ b/plugins/ruby-foreman-xen/debian/control
@@ -2,7 +2,7 @@ Source: ruby-foreman-xen
 Section: ruby
 Priority: extra
 Maintainer: Michael Moll <mmoll@mmoll.at>
-Build-Depends: debhelper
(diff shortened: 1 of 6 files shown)


plugins/ruby-foreman-xen/debian/control
```diff
@@ -2,7 +2,7 @@ Source: ruby-foreman-xen
 Section: ruby
 Priority: extra
 Maintainer: Michael Moll <mmoll@mmoll.at>
-Build-Depends: debhelper
+Build-Depends: debhelper, foreman-assets(>= 1.13.0~rc1)
```
domcleal: Also needs foreman-sqlite3 for the DB migration.

mmoll:
done.

domcleal:
Merged and building for both releases, thanks @mmoll.

PR #1549, opened 2017-03-02
Deploy dynflow-executor service via Foreman package

dLobatog:
This moves the deployment of the foreman-tasks service, which starts a
Dynflow executor, to the Foreman package.

This should be merged after - https://github.com/theforeman/foreman-tasks/pull/235 
When that PR is merged, I'll revisit this, I have not tried to build it yet. 

* [x] Nightly


foreman/foreman.spec
```diff
@@ -724,6 +738,12 @@ install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}/tmp/pids
 install -d -m0755 %{buildroot}%{_localstatedir}/run/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}/plugins
+#Copy init scripts and sysconfigs
+install -Dp -m0644 %{buildroot}%{gem_instdir}/%{deploy}/%{tasks_service_name}.sysconfig %{buildroot}%{_sysconfdir}/%{tasks_service_name}
```
ohadlevy: these files needs to move from foreman tasks(https://github.com/theforeman/foreman-tasks/tree/master/deploy) to foreman packaging.

dLobatog: Correct - that's done on https://github.com/theforeman/foreman/pull/4316/files now

dLobatog:
Last update ships the 'deploy/' files (service files) that were in foreman-tasks through this package directly

foreman/foreman-tasks.service
```diff
@@ -0,0 +1,17 @@
+
+Unit]
```
ohadlevy: bad copy / paste ?

foreman/foreman-tasks.service
```diff
@@ -0,0 +1,17 @@
...
+Description=Foreman jobs daemon
+Documentation=https://github.com/theforeman/foreman
```
ohadlevy: theforeman.org ?

foreman/foreman-tasks.service
```diff
@@ -0,0 +1,17 @@
+
+Unit]
+Description=Foreman jobs daemon
```
ohadlevy: shouldnt we call the whole thing that way? not just the description?

dLobatog:
https://drive.google.com/file/d/0Bz4WMhz8WLjnMm5pR0toNFVOM00/view?usp=sharing - dynflow 0.21
https://drive.google.com/file/d/0Bz4WMhz8WLjnNGFlSlJVdFl4d28/view?usp=sharing - foreman 1.15 with the change
https://drive.google.com/file/d/0Bz4WMhz8WLjnaU1PaVVuQURPZUU/view?usp=sharing - foreman-tasks with the change

foreman/foreman.spec
```diff
@@ -983,19 +1014,31 @@ exit 0
...
 (/sbin/service foreman status && /sbin/service foreman restart) >/dev/null 2>&1
+(/sbin/service %{tasks_service_name} status && /sbin/service %{tasks_service_name} restart) >/dev/null 2>&1
 exit 0
 
 %preun
 if [ $1 -eq 0 ] ; then
 /sbin/service %{name} stop >/dev/null 2>&1
 /sbin/chkconfig --del %{name} || :
 fi
+%if 0%{?rhel} == 6
```
ehelms: Technically there is no support for EL6 anymore. Which, the `BuildRequires: systemd` would break that anyway.

ehelms:
I'm giving this a test installation. What were the plans/desire for this with respect to 1.15?

ehelms:
I encountered the following on install test:

```
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: Apipie cache enabled but not present yet. Run apipie:cache rake task to speed up API calls.
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: rake aborted!
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: NoMethodError: undefined method `disable_active_record_actions' for #<ForemanTasks::Dynflow::Configuration:0x0000000960bd38>
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.8.6/app/models/foreman_tasks/concerns/action_triggering.rb:103:in `dynflow_task_wrap'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.8.6/app/models/foreman_tasks/concerns/action_triggering.rb:25:in `save_with_dynflow_task_wrap'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/rh/sclo-ror42/root/usr/share/gems/gems/activerecord-4.2.5.1/lib/active_record/persistence.rb:34:in `create'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/rh/sclo-ror42/root/usr/share/gems/gems/activerecord-4.2.5.1/lib/active_record/relation.rb:143:in `block in create'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/rh/sclo-ror42/root/usr/share/gems/gems/activerecord-4.2.5.1/lib/active_record/relation.rb:302:in `scoping'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/rh/sclo-ror42/root/usr/share/gems/gems/activerecord-4.2.5.1/lib/active_record/relation.rb:143:in `create'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/rh/sclo-ror42/root/usr/share/gems/gems/activerecord-4.2.5.1/lib/active_record/relation.rb:155:in `first_or_create'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /usr/share/foreman/db/seeds.d/05-taxonomies.rb:5:in `block in <top (required)>'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/theforeman/tfm/root/usr/share/gems/gems/audited-4.3.0/lib/audited/auditor.rb:279:in `without_auditing'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /usr/share/foreman/db/seeds.d/05-taxonomies.rb:3:in `<top (required)>'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/rh/sclo-ror42/root/usr/share/gems/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:268:in `load'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/rh/sclo-ror42/root/usr/share/gems/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:268:in `block in load'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/rh/sclo-ror42/root/usr/share/gems/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:240:in `load_dependency'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /opt/rh/sclo-ror42/root/usr/share/gems/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:268:in `load'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /usr/share/foreman/db/seeds.rb:81:in `block in <top (required)>'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /usr/share/foreman/db/seeds.rb:79:in `each'
[ WARN 2017-03-27 13:38:56 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: /usr/share/foreman/db/seeds.rb:79:in `<top (required)>'
```

dLobatog:
Updated to remove EL6 support & rebase it, but I still need to update dynflow/tasks to adapt to the newest changes and test it.

rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -1,19 +1,9 @@
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
-%if 0%{?scl:1}
```
dLobatog: All this spec depends on https://github.com/theforeman/foreman-tasks/pull/235 being merged though.

dLobatog:
https://github.com/theforeman/foreman-packaging/pull/1745 required

rubygem-foreman-tasks/rubygem-foreman-tasks.spec
dLobatog: It's merged.

foreman/foreman.spec
```diff
@@ -1004,19 +1012,22 @@ exit 0
 %{foreman_rake} db:seed >> %{_localstatedir}/log/%{name}/db_seed.log 2>&1 || :
 %{foreman_rake} apipie:cache:index >> %{_localstatedir}/log/%{name}/apipie_cache.log 2>&1 || :
 (/sbin/service foreman status && /sbin/service foreman restart) >/dev/null 2>&1
+(/sbin/service %{executor_service_name} status && /sbin/service %{executor_service_name} restart) >/dev/null 2>&1
```
ekohl: Isn't this redundant and already covered in the ```%postun``` systemd macro? It makes me think the other sysvinit-style commands can also be replaced with systemd macros.

foreman/dynflow-executor.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
```
ehelms: Is it standard to name the service after the bin script? I'd think `dynflow` would be a simpler service to remember.

foreman/foreman.spec
ehelms: Also, are these service start and restarts the smart way to go? I feel like our application is so complex, and so much is baked into our installer that doing these operations outside of it can have unintended consequences.

ehelms:
Speaking of installer, we will want to add things to puppet-foreman to manage the new service?

ekohl:
> Speaking of installer, we will want to add things to puppet-foreman to manage the new service?

https://github.com/theforeman/puppet-foreman/pull/530

foreman/foreman.spec
```diff
@@ -46,8 +50,11 @@ Requires: wget
 Requires: /etc/cron.d
 Requires(pre):  shadow-utils
 Requires(post): chkconfig
+Requires(post): systemd-sysv
```
ekohl: I recall there being a macro for the requirements as well

dLobatog: Sorry, what macro/where? I'm missing something here

ekohl: https://github.com/theforeman/foreman-packaging/blob/dc21dfd9cc15499785e4da30cdad67131d99e52a/foreman/foreman.spec#L132

iNecas:
Time to revive this PR, as new tasks with the required change are being released https://github.com/theforeman/foreman-packaging/pull/1821

foreman/dynflow-executor.service
```diff
@@ -0,0 +1,16 @@
...
+Type=forking
+User=foreman
+TimeoutSec=600
+WorkingDirectory=/usr/share/foreman
+ExecStart=/usr/bin/dynflow-executor start
+ExecStop=/usr/bin/dynflow-executor stop
+EnvironmentFile=-/etc/sysconfig/dynflow-executor
+
+[Install]
+WantedBy=multi-user.target
```
iNecas: What about `Alias=foreman-tasks`, at least until we have the other tooling (such as katello-service) updated to the new name?

dLobatog: Sounds good to me, will do

foreman/foreman.spec
dLobatog: That macro adds systemd requires, not systemd-sysv or systemd-units (https://github.com/systemd/systemd/blob/master/src/core/macros.systemd.in)

foreman/dynflow-executor.sysconfig
```diff
@@ -0,0 +1,26 @@
+FOREMAN_USER=foreman
+BUNDLER_EXT_HOME=/usr/share/foreman
+RAILS_RELATIVE_URL_ROOT=$FOREMAN_PREFIX
```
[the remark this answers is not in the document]
dLobatog: Nope, a leftover from the old service

foreman/foreman.spec
```diff
@@ -947,6 +961,11 @@ rm -rf %{buildroot}
 %ghost %attr(0640,root,%{name}) %config(noreplace) %{_datadir}/%{name}/config/initializers/local_secret_token.rb
 %{_prefix}%{_tmpfilesdir}/%{name}.conf
 
+# Service
+%{bin_dir}/%{executor_service_name}
```
ekohl: Why don't you use the ```%{_bindir}``` macro here?

foreman/foreman.spec
```diff
@@ -758,6 +767,11 @@ install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}/tmp/pids
 install -d -m0755 %{buildroot}%{_localstatedir}/run/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}/plugins
+#Copy init scripts and sysconfigs
+install -Dp -m0644 %{SOURCE10} %{buildroot}%{_sysconfdir}/%{executor_service_name}
+install -Dp -m0644 %{SOURCE11} %{buildroot}%{_unitdir}/%{executor_service_name}.service
+mkdir -p %{buildroot}%{bin_dir}
```
ekohl: Why don't you use the ```%{_bindir}``` macro here?

foreman/foreman.spec
ekohl: Why are those needed? I can't even find systemd-units on my Fedora or on a CentOS 7 machine.

foreman/dynflow-executor.service
ekohl: (done)

foreman/foreman.spec
ekohl: To answer my own question: the ```systemd``` RPM has a provides ```systemd-units``` so I'd prefer to use the requires macro. We should also look at the ```chkconfig``` and ```initscripts``` requirements but that might be out of scope for this PR.

dLobatog:
@iNecas I think I will have to include your commits here - foreman-tasks need to be updated, if I use 0.10.0, koji returns the following error:

```
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.10.0/bin/dynflow-executor
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.10.0/bin/foreman-tasks
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.10.0/deploy/foreman-tasks.init
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.10.0/deploy/foreman-tasks.service
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.10.0/deploy/foreman-tasks.sysconfig
```

At 0.10.0 time the dynflow PR was not merged yet, it has to be 0.10.1 at least so I prefer to include your PR which has the newest version.

dLobatog:
@ekohl Now it's using _bindir instead of bin_dir, not sure why I made that change...

iNecas:
I'm ok with including the update here. Btw. this is going to be in 1.16, right?

ehelms:
Is there a reason to not just name the service "dynflow" or "dynflowd" vs. the more complex "dynflow-executor" ?

iNecas:
dynflowd sound good :)

ekohl:
👍 on dynflowd but now I want a client named dynflows ;)

iNecas:
What about dynflowctl? Just kidding :)

dLobatog:
I called it dynflow-executor because the original script in foreman-tasks was called that: https://github.com/theforeman/foreman-tasks/blob/v0.10.0/bin/dynflow-executor

Happy to change it to dynflowd or whatever

foreman/foreman.spec
dLobatog: systemd-sysv is meant for automatic compatibility with SysV

ekohl: Ok, I'd prefer to follow the Fedora guidelines here and this is fine then.

dLobatog:
Repushed as changing the name to dynflowd caused a build to fail as the spec needs to look for 'foreman/bin/dynflow-executor' not 'foreman/bin/dynflowd'

dLobatog:
Back to `ModuleNotFoundError: Module not found: Error: Can't resolve 'object-assign' in '/usr/lib/node_modules/react/lib'`, this was working before renaming dynflow-executor 😕

foreman/foreman.spec
```diff
@@ -758,6 +766,11 @@ install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}/tmp/pids
 install -d -m0755 %{buildroot}%{_localstatedir}/run/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}/plugins
+#Copy init scripts and sysconfigs
+install -Dp -m0644 %{SOURCE10} %{buildroot}%{_sysconfdir}/%{executor_service_name}
+install -Dp -m0644 %{SOURCE11} %{buildroot}%{_unitdir}/%{executor_service_name}.service
+mkdir -p %{buildroot}%{_bindir}
+ln -sv %{gem_instdir}/bin/dynflow-executor %{buildroot}%{_bindir}/%{executor_service_name}
```
ekohl: I wonder why this symlink needed because the unit file uses ```/usr/bin/dynflow-executor```. Could that be the issue?

ehelms:
[test]

ares:
[test]

ehelms:
[test]

ehelms:
@ohadlevy @ekohl @iNecas This is passing now that nightlies are working, if we could get re-reviews on it.

ekohl:
Should there be a ```Conflicts: foreman-tasks <= 0.10.3``` in ```foreman.spec```?

ekohl:
I was waiting for nightlies to stabilize again to test this.

foreman/foreman.spec
```diff
@@ -737,6 +745,8 @@ webpack.js --bail --config config/webpack.config.js
 %{scl_rake} apipie:cache RAILS_ENV=production cache_part=resources --trace
 rm config/database.yml config/settings.yaml
 
+sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X/usr/bin/%{?scl:%{scl_prefix}}rubyX' bin/dynflow-executor
```
ekohl: I think this line is redundant because of https://github.com/theforeman/foreman-packaging/blob/22efa6c9e1b22f82aff3f62b6960bd99d2ce89fd/foreman/foreman.spec#L816-L825

dLobatog: Correct, will remove it

foreman/foreman.spec
```diff
@@ -1063,6 +1076,11 @@ rm -rf %{buildroot}
 %ghost %attr(0640,root,%{name}) %config(noreplace) %{_datadir}/%{name}/config/initializers/local_secret_token.rb
 %{_tmpfilesdir}/%{name}.conf
 
+# Service
+%{_bindir}/%{executor_service_name}
+%config %{_sysconfdir}/%{executor_service_name}
```
ekohl: This gets installed as ```/etc/dynflowd``` which means the service won't start. It should be ```%config(noreplace) %{_sysconfdir}/sysconfig/%{executor_service_name}```

foreman/dynflowd.service
```diff
@@ -0,0 +1,17 @@
...
+Description=Foreman jobs daemon
+Documentation=https://theforeman.org
+After=network.target remote-fs.target nss-lookup.target
+
+[Service]
+Type=forking
+User=foreman
+TimeoutSec=600
+WorkingDirectory=/usr/share/foreman
+ExecStart=/usr/bin/dynflow-executor start
```
ekohl: This gets installed as ```/usr/bin/dynflowd``` now. Same for the line below.

foreman/foreman.spec
ekohl: The gem_instdir doesn't expand so the symlink is broken:

```
[root@pipeline-katello-nightly-centos7 sysconfig]# ls /usr/bin/dynflowd -l
lrwxrwxrwx. 1 root root 35 Nov 13 20:58 /usr/bin/dynflowd -> %{gem_instdir}/bin/dynflow-executor
```

foreman/dynflowd.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Foreman jobs daemon
+Documentation=https://theforeman.org
+After=network.target remote-fs.target nss-lookup.target
+
+[Service]
+Type=forking
+User=foreman
+TimeoutSec=600
```
ekohl: We reduced this startup time in passenger to 90 seconds (same as httpd). I'd argue that we should do the same here since 10 minutes if a VERY long time to start up.

dLobatog: Yeah, it isn't really necessary for `dynflowd`. Unfortunately for Foreman +  Katello I regularly have to increase the startup time to 200/300

dLobatog:
Just pushed it again because `%files` was looking for dynflowd in the wrong path (bin instead of sbin)

foreman/dynflowd.service
```diff
@@ -0,0 +1,17 @@
...
+Description=Foreman jobs daemon
+Documentation=https://theforeman.org
+After=network.target remote-fs.target nss-lookup.target
+
+[Service]
+Type=forking
+User=foreman
+TimeoutSec=90
+WorkingDirectory=/usr/share/foreman
+ExecStart=/usr/bin/dynflowd start
```
ekohl: It's now installed to ```/usr/sbin``` so it fails to start

dLobatog:
@ekohl Cool, I will squash these 2 commits then.

foreman/foreman.spec
```diff
@@ -874,6 +882,11 @@ install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}/tmp/pids
 install -d -m0755 %{buildroot}%{_localstatedir}/run/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}/plugins
+#Copy init scripts and sysconfigs
+install -Dp -m0644 %{SOURCE10} %{buildroot}%{_sysconfdir}/sysconfig/%{executor_service_name}
+install -Dp -m0644 %{SOURCE11} %{buildroot}%{_unitdir}/%{executor_service_name}.service
+mkdir -p %{buildroot}%{_sbindir}
+install -Dp -m0755 bin/dynflow-executor %{buildroot}%{_sbindir}/%{executor_service_name}
```
ekohl: This was moved to ```script/dynflowd```

foreman/foreman.spec
```diff
@@ -874,6 +882,11 @@ install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}/tmp/pids
 install -d -m0755 %{buildroot}%{_localstatedir}/run/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}/plugins
+#Copy init scripts and sysconfigs
+install -Dp -m0644 %{SOURCE10} %{buildroot}%{_sysconfdir}/sysconfig/%{executor_service_name}
+install -Dp -m0644 %{SOURCE11} %{buildroot}%{_unitdir}/%{executor_service_name}.service
+mkdir -p %{buildroot}%{_sbindir}
```
ekohl: This isn't needed because ```install -D``` creates it

mmoll:
The plugin updates could go in independently, I presume?

dLobatog:
@mmoll I don't think so, they require dynflowd to be deployed by core to work

dLobatog:
@ekohl This is green again now that you removed fbjs from the spec :smile:

ares:
@ekohl could we get this merged to avoid unnecessary rebasing? I think @dLobatog actively works on deb part (hopefully should be ready soon)

ekohl:
@ares I think if we merge this now we need installer support as well and we either need to modify that to not break debian or finish debian. My preference is the latter.

dLobatog:
@ekohl Don't we have installer support by merging https://github.com/theforeman/puppet-foreman/pull/602 ?

ekohl:
@dLobatog yes, but that currently assumes the debian side is also using the service named ```dynflowd```.

iNecas:
I lost track on what's the status of the issues at the moment?

Without this, we can't release a new version of foreman-tasks, and we already have remote execution broken due to old tasks version in 1.16. http://projects.theforeman.org/issues/21889#change-98449

I suggest reverting https://github.com/theforeman/foreman-tasks/commit/ecb8c5b39f6281c9c29ff3ea677b3d43002bbea2 from tasks 0.10.x so that we can get it to 1.16 soon.

Also, debian seems to already have tasks 0.10.6 in 1.16 https://github.com/theforeman/foreman-packaging/pull/1908

dLobatog:
@iNecas We're waiting on https://github.com/theforeman/foreman-packaging/pull/1959 to be merged. Would it be possible to pin REX to tasks 0.10.1 or it'd be broken then too?

foreman/dynflowd.service
```diff
@@ -0,0 +1,17 @@
...
+User=foreman
+TimeoutSec=90
+WorkingDirectory=/usr/share/foreman
+ExecStart=/usr/sbin/dynflowd start
+ExecStop=/usr/sbin/dynflowd stop
+EnvironmentFile=-/etc/sysconfig/dynflowd
+
+[Install]
+WantedBy=multi-user.target
+Alias=foreman-tasks
```
ekohl: As suggested by @evgeni in the deb PR: this should be ```foreman-tasks.service```

ekohl:
https://github.com/theforeman/foreman-packaging/pull/1992 should unblock master builds

ekohl:
[test]

ekohl:
Since core was merged and the Debian side too, I think we can merge this even though it's broken. That would allow us to merge the installer side of it as well. On the other hand, due to the alias in the unit file I think it currently works on Debian as well.

ehelms:
Nightly CI pipeline is also now passing, would be great to get this in before we branch this repository.

ehelms:
This has an older foreman-tasks commit in it. The main branch is now up to 0.11.0 of foreman-tasks.

dLobatog:
Updated the tasks spec as this was failing with 0.11 because

```
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.11.0/bin/dynflow-executor
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.11.0/bin/foreman-tasks
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.11.0/deploy/foreman-tasks.service
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman-tasks-0.11.0/deploy/foreman-tasks.sysconfig
```

foreman/foreman.spec
```diff
@@ -30,8 +31,12 @@ Source6: %{name}.repo
 Source7: %{name}-plugins.repo
 Source8: %{name}.gpg
 Source9: message_encryptor_extensions.rb
+Source10: %{executor_service_name}.sysconfig
+Source11: %{executor_service_name}.service
 BuildArch:  noarch
 
+Conflicts: foreman-tasks < 0.10.3
```
ekohl: This needs to be 0.11.0-2 now and foreman-tasks' release needs to be bumped.

dLobatog:
@ehelms @ekohl - thanks for the review, updated the  conflicts version

ehelms:
I have one thought for consideration, and I'd accept doing it as a follow on, but might make sense to just get right now.

This current setup means that to run dynflow as a service I have to install the Foreman package and all of it's dependencies which makes easy sense on an all-in-one box. If I wanted to run dynflow on it's own box or a container, I couldn't do that without installing all of Foreman? Should we consider a sub-package that can be installed independently and in a light weight fashion?

ekohl:
@ehelms I recall having talked about that already. I think that the dynflow executor now requires foreman code so you need that package anyway. Let me know if I forgot any cases:

* AIO: install all packages
* Only install web app: core package
* Only run dynflow: core package + dynflow

Only in the second you wouldn't need the dynflow things. I don't think we can actually separate the code so shipping all files and only enabling the deamon if you want to is our best option.

ehelms:
Fair enough for now -- ACK

I am good to merge this.

iNecas:
Dynflow itself doesn't need Rails, but for Foreman usage, we need the Rails env.

ekohl:
@iNecas so https://github.com/theforeman/foreman-packaging/pull/1549#issuecomment-357238300 was correct? Only if you don't wish to run the dynflow daemon you have additional files and splitting that off into a separate package is quite some effort for little gain.

iNecas:
Yes, we would need to do better job in Foreman to extract the core parts first (the async jobs dont need anything from the controllers or views), in order for this to make sense to do.

ekohl:
In that case I'm going to merge this so we can finally get it in. Thanks everyone!

PR #1550, opened 2017-03-08
Added rubygem-foreman_maintain package

gnurag:
This is the first iteration of packaging foreman_maintain for Foreman project.


ehelms:
[test]

iNecas:
@gnurag  It seems the Jenkins because of git annex. perhaps added the gem that was not exectly the same file that on the rubygems?

PR #1542, opened 2017-02-28
Updated foreman_discovery to 8.0.1 (RPM)

lzap:
For plugin updates, please indicate which repos this should be built into:

* [ ] Nightly
* [x] 1.14

See Foreman's [plugin maintainer documentation](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

PR #1561, opened 2017-03-16
Migrate db immediatelly after loading schema

adamruzicka:
Running `db:migrate` in a separate process makes it load the environment. This might fail when a plugin's initialization relies on permissions from foreman to be present and no migrations are pending.

This affects remote execution and (maybe) discovery

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -862,8 +862,7 @@ rm \`pwd\`/config/initializers/encryption_key.rb \\
 /usr/bin/%%{?scl:%%{scl}-}rake security:generate_encryption_key \\
 export BUNDLER_EXT_NOSTRICT=1 \\
 %%{?-s:/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=production --trace} \\
-%%{?-a:/usr/bin/%%{?scl:%%{scl}-}rake db:create db:schema:load SCHEMA=%{_datadir}/%{name}/schema_plugin.rb RAILS_ENV=development --trace} \\
-%%{?-a:/usr/bin/%%{?scl:%%{scl}-}rake db:migrate RAILS_ENV=development --trace} \\
+%%{?-a:/usr/bin/%%{?scl:%%{scl}-}rake db:create db:schema:load db:migrate SCHEMA=%{_datadir}/%{name}/schema_plugin.rb RAILS_ENV=development --trace} \\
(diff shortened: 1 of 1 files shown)


domcleal:
> Running db:migrate in a separate process makes it load the environment. This might fail when a plugin's initialization relies on permissions from foreman to be present and no migrations are pending.

The plugin would be at fault here, it shouldn't fail to load if `db:seed` hasn't run - the environment and all of these `db:*` rake commands should still be usable. The proposed change looks like a workaround or hack for a bug elsewhere?

adamruzicka:
> The proposed change looks like a workaround or hack for a bug elsewhere?

Kind of. The builds were working just fine until a recent change in the build macros brought this issue to light. Still it would be nice to have the builds working (even with a workaround) until we manage to fix this in all the other places that need it.

domcleal:
Is the issue still present in the plugin's master branch? If not, then I'm more willing to merge this workaround, but not if the plugin issue is unfixed.

iNecas:
As @lzap already pointed out in https://github.com/theforeman/foreman-packaging/pull/1542#issuecomment-283368790, the issue in packaging was introduced in 248db50. Unfortunately, I can't find a PR that introduced this change (a link welcome), that I would expect a discussion around potential consequences: I believe this is not a plugin's bug, but the bug in the core's plugin API, and the 248db50 should not have been pushed there until the issue is resolved in the core.

ares:
@domcleal the problem is in Foreman core, the `permission` method can't handle this circumstances. I don't see why this patch that unblocks plugins shouldn't be merged. I agree that fix should follow in core, would reporting an issue help until someone gets to fixing it? Anyway, the answer to your question is "no", it's not present in plugin master branch.

ehelms:
[test]

ehelms:
Testing this, without reverting the whole change, looks like it can lead to breakages in plugin building:

```
** Invoke db:_dump (first_time)
** Execute db:_dump
** Invoke db:schema:dump (first_time)
** Invoke environment
** Invoke db:load_config
** Execute db:schema:dump
rake aborted!
Errno::EACCES: Permission denied @ rb_sysopen - /usr/share/foreman/schema_plugin.rb
/opt/rh/rh-ror42/root/usr/share/gems/gems/activerecord-4.2.6/lib/active_record/railties/databases.rake:239:in `initialize'
/opt/rh/rh-ror42/root/usr/share/gems/gems/activerecord-4.2.6/lib/active_record/railties/databases.rake:239:in `open'
/opt/rh/rh-ror42/root/usr/share/gems/gems/activerecord-4.2.6/lib/active_record/railties/databases.rake:239:in `block (3 levels) in <top (required)>'
```

iNecas:
Ok, I take this as definite reason why we should revert the original commit for now https://github.com/theforeman/foreman-packaging/pull/1573

adamruzicka:
Closing in favor of https://github.com/theforeman/foreman/pull/4396

iNecas:
Just to make it clear: the issue with plugin packages is still not resolved due to issue described https://github.com/theforeman/foreman-packaging/pull/1561#issuecomment-288560211 and revert request from https://github.com/theforeman/foreman-packaging/pull/1573 is still wanted

PR #1564, opened 2017-03-20
Update deface to 1.2.0 [deb]

domcleal:

* [x] Nightly

---

`colorize` was replaced by `rainbow`, which is already in the `foreman` bundle cache.


mmoll:
merged, thanks @domcleal!
The relaxed nokogori version constraint also helps...

PR #1582, opened 2017-03-27
Fixes 19041 - include sbin in -proxy package (DEB)

mmoll:


domcleal:
Merged, thanks @mmoll.

PR #1581, opened 2017-03-27
[DEB] Update foreman-tasks to 0.9.1 and foreman-tasks-core to 0.1.2

adamruzicka:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [ ] 1.14
* [ ] 1.13
* [ ] 1.12

See Foreman's [plugin maintainer documentation](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

PR #1586, opened 2017-03-28
pin concurrent-ruby to 1.0.3

mmoll:
Dynflow commit 5fc0ba81 locked concurrent-ruby-edge to 0.2.3, which is
locked to concurrent-ruby 1.0.3. Foreman-tasks uses Dynflow and as other
gems pull in concurrent-ruby also in the Foreman core package, it needs to
get locked here also.

This should be built into:

* [x] Nightly

### DIFF
--- a/debian/jessie/foreman/debian.rb
+++ b/debian/jessie/foreman/debian.rb
@@ -1,0 +1,1 @@
+gem 'concurrent-ruby', '= 1.0.3'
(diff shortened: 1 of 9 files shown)


mmoll:
see https://github.com/theforeman/foreman-packaging/pull/1581#discussion_r108441884

domcleal:
Merged, thanks @mmoll.

PR #1587, opened 2017-03-28
update hammer to 0.10.0 (DEB)

mmoll:
this should be built into:

* [x] Nightly


domcleal:
Merged and will build, thanks @mmoll.

PR #1593, opened 2017-03-29
Update foreman_salt (RPM)

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.15
* [x] 1.14


domcleal:
Merged, thanks @mmoll. foreman_salt is built only for 1.14, and smart_proxy_salt for nightly and 1.14, as the build system's in a mess, will build them once it's back to normal.

PR #1605, opened 2017-04-06
[deb] Core projects: Release 1.15.0-RC1

dLobatog:
Changelogs updated for RC1 on debs.

### DIFF
--- a/debian/jessie/foreman-installer/changelog
+++ b/debian/jessie/foreman-installer/changelog
@@ -1,3 +1,15 @@
+foreman-installer (1.15.0~rc1-1) stable; urgency=low
+
+  * 1.15.0-RC1 released
+
+ -- Daniel Lobato Garcia <me@daniellobato.me>  Thu, 06 Apr 2017 13:50:56 +0200
+
+foreman-installer (1.14.3-1) stable; urgency=low
+
+  * 1.14.3 released
+
+ -- Dominic Cleal <dominic@cleal.org>  Mon, 03 Apr 2017 12:05:04 +0100
+
 foreman-installer (1.14.2-1) stable; urgency=low
 
   * 1.14.2 released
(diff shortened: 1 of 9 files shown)


debian/jessie/foreman-installer/changelog
```diff
@@ -1,3 +1,9 @@
+foreman-installer (1.15.0~rc1-1) stable; urgency=low
+
+  * 1.15.0-RC1 released
+
+ -- Daniel Lobato Garcia <me@daniellobato.me>  Thu, 06 Apr 2017 13:50:56 +0200
+
 foreman-installer (1.15.0-1) stable; urgency=low
```
domcleal: This entry shouldn't be present, and the 1.14 entries are missing. I think 602c743 is missing from the branch, it should have been branched _after_ that commit.

dLobatog:
That's true, I'll see how can I amend the changelogs in this PR, thanks for noticing

domcleal:
Include the commit I referenced.

dLobatog:
Updated, I've cherry-picked https://github.com/theforeman/foreman-packaging/commit/602c743c9696a5d5f168c3763bf8da3ae99f006a , fixed the conflicts, and added 1.14.3 which was missing as it was more recent.

domcleal:
Merged, thanks @dLobatog.

domcleal:
I notice the 1.15 jobs were _already queued_ before this was reviewed and merged. I've made sure this was merged just in time, but you really should not build the packages before the version number is updated.

dLobatog:
I was building them so that at least I could sign the RPMs before this was done, does that cause any problem?

domcleal:
I don't know, but I don't think running builds for 1.15.0 (which was the current version in deb/1.15) would have been sensible.

dLobatog:
Yeah it was caused problems in the end - http://ci.theforeman.org/view/Release%20pipeline/job/release_packages/77/ - some builds now have to be manually removed from koji before I can resubmit the job, otherwise it fails when it finds the builds are already there.

domcleal:
The cause appears to have been the cancellation of http://ci.theforeman.org/job/release_packages/75/, most of the jobs had already completed. I queued http://ci.theforeman.org/job/packaging_build_deb_coreproject/4262/ for the last of the three packages to be built, after which `release_mash` may be run to resume the process.

PR #1606, opened 2017-04-06
Release 1.15.0-0.2.RC1 

dLobatog:
Since removing the old builds was not possible & [release_packages](http://ci.theforeman.org/view/Release%20pipeline/) is not green to allow release_mash to proceed. 

To continue the pipeline release_packages needs to be green - which it will not be as the builds exist in Koji already and the job stops. debs are built just fine too https://github.com/theforeman/foreman-packaging/pull/1605#issuecomment-292201828

### DIFF
--- a/foreman-installer/foreman-installer.spec
+++ b/foreman-installer/foreman-installer.spec
@@ -6,7 +6,7 @@
 Name:       foreman-installer
 Epoch:      1
 Version:    1.15.0
-Release:    0.1%{?dotalphatag}%{?dist}
+Release:    0.2%{?dotalphatag}%{?dist}
 Summary:    Puppet-based installer for The Foreman
 Group:      Applications/System
 License:    GPLv3+ and ASL 2.0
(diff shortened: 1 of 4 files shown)


dLobatog:
[test] (19:13:34 ServerOffline: database outage)

domcleal:
If the 0.1 builds were successful (which they appear to be), I don't see the need to rebuild these packages.

dLobatog:
How do you continue with the release pipeline when release_packages is red due to "builds already exist"?

domcleal:
Build the next job directly.

dLobatog:
Alright - I'll keep it open just in case, right now I'm uploading the package signatures before running the job. I was confused since there is no button to trigger the next job when the previous one is red, and I thought that just triggering it would not make it show up in the pipeline.
screenshot from 2017-04-07 10-48-48

domcleal:
Closing, as the existing builds were released.

PR #1609, opened 2017-04-10
Depend on Puppet 4.6.1 or equivalent puppet-agent [deb]

domcleal:


mmoll:
merged, thanks @domcleal!

PR #1610, opened 2017-04-11
Fixes #19245 - Logrotate specific proxy logs

adamruzicka:

* [x] Nightly
* [x] 1.15
* [ ] 1.14
* [ ] 1.13

---

### DIFF
--- a/foreman-proxy/logrotate.conf
+++ b/foreman-proxy/logrotate.conf
@@ -1,4 +1,7 @@
-/var/log/foreman-proxy/*log {
+/var/log/foreman-proxy/cron.log
+/var/log/foreman-proxy/migrate_settings.log
+/var/log/foreman-proxy/proxy.log
+{
   missingok
   notifempty
   create 0644 foreman-proxy foreman-proxy
--- a/foreman-proxy/logrotate.conf
+++ b/foreman-proxy/logrotate.conf
@@ -1,4 +1,3 @@
-/var/log/foreman-proxy/cron.log
 /var/log/foreman-proxy/migrate_settings.log
 /var/log/foreman-proxy/proxy.log
 {


mmoll:
I guess the dynflow-logs need a logratote entry in the smart_proxy_dynflow package then? Also, please open a PR to deb/develop for `debian/*/foreman-proxy/logrotate` once an agreed version went into rpm/develop.

adamruzicka:
@mmoll Yes, logrotate entry in smart_proxy_dynflow is needed, I'll open PRs for it once we release a version of smart_proxy_dynflow which can reopen its logs on a signal.

I'll open the debian PR once this approach gets agreed on.

domcleal:
This is still waiting on contributor per https://github.com/theforeman/foreman-packaging/pull/1610#issuecomment-293245828.

adamruzicka:
@domcleal theforeman-bot is happy and the build is green

foreman-proxy/logrotate.conf
```diff
@@ -1,4 +1,7 @@
-/var/log/foreman-proxy/*log {
+/var/log/foreman-proxy/cron.log
```
domcleal: Where is this file created?

adamruzicka: I'm not sure, but foreman-debug-proxy is set to collect it [here](https://github.com/theforeman/smart-proxy/blob/2cc33adc5e63fb0acf35f714ec3372c5ee555d3d/extra/foreman-debug-proxy#L11) so I thought it would be good to include it as well.

domcleal: It should be removed then unless you can cite the source, I don't think it does anything.

adamruzicka: Ok, removed it

domcleal:
Merged, thanks @adamruzicka. Please make the same change to deb/develop.

adamruzicka:
Created deb PR https://github.com/theforeman/foreman-packaging/pull/1612

Klaas-:
This did not make it into 1.15 yet @adamruzicka / @dLobatog -- maybe put it in 1.15.4?

PR #1613, opened 2017-04-13
Update foreman_monitoring to 0.1.0

dgoetz:

* [x] Nightly
* [x] 1.15
* [ ] 1.14
* [ ] 1.13

---


domcleal:
Merged and building, thanks @dgoetz.

Please also leave off the changelog entry in future, it's added during build.

PR #1615, opened 2017-04-18
Remove email.yaml example file [deb]

domcleal:
Support removed in theforeman/foreman@aae9e73.


mmoll:
merged, thanks @domcleal!

PR #1623, opened 2017-04-21
add Debian/stretch packaging

mmoll:
This should be built into:

* [x] Nightly

This adds the yard and puppet-strings gems as native ruby-* DEBs and uses the native puppet package with these to build the installer.


debian/stretch/foreman-installer/changelog
```diff
@@ -0,0 +1,107 @@
+foreman-installer (1.16.0-1) stable; urgency=low
+
+  * Bump changelog to 1.16.0 to match VERSION
+
+ -- Michael Moll <mmoll@mmoll.at>  Thu, 30 Mar 2017 10:54:21 +0200
+
+foreman-installer (1.14.2-1) stable; urgency=low
```
domcleal: Probably also delete these changelog entries for stretch as they weren't shipped, just keep 1.16.0.

mmoll:
should all be fixed

domcleal:
Merged (not built), thanks @mmoll.

PR #1627, opened 2017-04-25
Import rest-client 1.8.0 from stretch to trusty + jessie

domcleal:
Newer rest-client will be required for hammer_cli_foreman. netrc 0.7+
and http-cookie 1.0.2 are also required for the newer rest-client.

Tests disabled as newer gem2deb required to run them, and unsatisfied
webmock build dep removed from trusty's rest-client as a result.


domcleal:
Added http-cookie 1.0.2 to trusty to match the slightly newer rest-client dependency.

mmoll:
This might interfere with native vagrant packages installed on the same host, but I guess we don't have another choice...

mmoll:
merged, thanks @domcleal!

PR #1630, opened 2017-04-25
[DEB] Update foreman_templates to 5.0.0

ares:

* [x] Nightly
* [x] 1.15
* [ ] 1.14
* [ ] 1.13

---

The 5.x restores the compatibility with Foreman 1.15


ares:
Sorry it took me a while. I left migration in postinst commented so later when some migrations are added it's easier to add it. I can obviously remove comments if requested.

I also needed to add git and foreman as build deps, other plugin use foreman-assets but we don't compile assets here so I though direct foreman dependency would make more sense. That also required foreman-sqlite3 build dependency.

domcleal:

PR #1631, opened 2017-04-25
[RPM] Update foreman_templates to 5.0.0

ares:

* [x] Nightly
* [x] 1.15
* [ ] 1.14
* [ ] 1.13

---

The 5.x restores the compatibility with Foreman 1.15


rubygem-foreman_templates/rubygem-foreman_templates.spec
```diff
@@ -89,6 +94,13 @@ GEMFILE
 %doc %{gem_instdir}/LICENSE
 %doc %{gem_instdir}/README.md
 
+%posttrans
+%{foreman_db_migrate}
```
ares: not related to apipie but other packages run the migration and seed here too so I added it to be consistent

rubygem-foreman_templates/rubygem-foreman_templates.spec
```diff
@@ -35,6 +35,8 @@ Requires: %{?scl_prefix}rubygem(git)
 BuildRequires: %{?scl_prefix_ruby}ruby(release)
 BuildRequires: %{?scl_prefix_ruby}rubygems-devel
 BuildRequires: %{?scl_prefix_ruby}rubygems
+BuildRequires: foreman-plugin >= 1.15
```
ares: this is not related to apipie but according to gem2rpm template it seems it should be here unconditionally

rubygem-foreman_templates/rubygem-foreman_templates.spec
domcleal: Other plugins with this have DB migrations and seeds to run, this plugin doesn't so they should be removed to improve installation speed.

rubygem-foreman_templates/rubygem-foreman_templates.spec
domcleal: It's required for the precompilation-related build macros.

rubygem-foreman_templates/rubygem-foreman_templates.spec
ares: not now but once the plugin will have, we'll need to add this... but I agree with the speed argument, I'll remove it

rubygem-foreman_templates/rubygem-foreman_templates.spec
ares: so it's not needed now but if it does not do any harm and the template would add it anyway, I think we can keep it?

domcleal: It is needed now, don't remove it. The apipie docs (pre)compilation requires it, I was just saying that it is related to apipie docs.

ares: oh I though it was assets precompilation, now I see the the same macro is used with precompile flags

ares:
Now it fails on diffy not being found, I guess all runtime dependencies needs to become build deps as well so the rake task can be run?

rubygem-foreman_templates/rubygem-foreman_templates.spec
```diff
@@ -35,6 +35,10 @@ Requires: %{?scl_prefix}rubygem(git)
 BuildRequires: %{?scl_prefix_ruby}ruby(release)
 BuildRequires: %{?scl_prefix_ruby}rubygems-devel
 BuildRequires: %{?scl_prefix_ruby}rubygems
+BuildRequires: %{?scl_prefix}rubygem(diffy)
+BuildRequires: %{?scl_prefix}rubygem(git)
```
domcleal: `git` also appears to be a build time dependency, as it's a hard dependency of this gem.

    Git::GitExecuteError: git version   2>&1:sh: git: command not found
    /usr/share/gems/gems/git-1.2.5/lib/git/lib.rb:700:in `command'
    /usr/share/gems/gems/git-1.2.5/lib/git/lib.rb:649:in `current_command_version'
    /usr/share/gems/gems/git-1.2.5/lib/git/lib.rb:659:in `meets_required_version?'
    /usr/share/gems/gems/git-1.2.5/lib/git.rb:30:in `<top (required)>'

ares: that should probably be listed at https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-git/rubygem-git.spec#L15 as a runtime dependency too, should I send a separate patch for it?

domcleal: I'd be inclined to agree, but only if the same change was made to the Fedora package too (first).

domcleal:
[test] with apipie-rails 0.5.0 removed.

rubygem-foreman_templates/rubygem-foreman_templates.spec
ares: After I thought more about this, it's actually a bug in foreman_templates. It's usable without git. I'll add this as a dependency for now but I opened http://projects.theforeman.org/issues/19381 to ensure the git is not required and then we can remove it from here. If the issue is in git gem loading itself then I'll look if it can be fixed there.

ares:
[test] I'm not sure what happened, tasks finished successfully http://koji.katello.org/koji/taskinfo?taskID=629767 http://koji.katello.org/koji/taskinfo?taskID=629768 maybe the koji was busy so the job time outed

domcleal:

PR #1632, opened 2017-04-25
update hammer to 0.10.1 (DEB)

mmoll:
this should be built into:

* [x] Nightly


mbacovsky:
@mmoll thanks for packaging. 0.10.1 should also get to 1.15

mmoll:
that would need at least an RC3 for the whole BATS dance, as #1627 is quite intrusive

domcleal:

PR #1634, opened 2017-04-26
Updated ruby-puppetdb-foreman to 3.0.1 [deb]

timogoebel:
Please build for:

* [x] Nightly
* [x] 1.15
* [x] 1.14
* [x] 1.13

⚠️ Fair warning: I have no clue about debian packaging. The changes for apipie cache are just copy-pasted.

### DIFF
--- a/plugins/ruby-puppetdb-foreman/debian/changelog
+++ b/plugins/ruby-puppetdb-foreman/debian/changelog
@@ -1,3 +1,9 @@
+ruby-puppetdb-foreman (3.0.1-1) stable; urgency=low
+
+  * 3.0.1 released
+
+ -- Timo Goebel <mail@timogoebel.name>  Wed, 26 Apr 2017 10:57:40 +0200
+
 ruby-puppetdb-foreman (2.0.0-1) stable; urgency=low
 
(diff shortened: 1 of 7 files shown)


domcleal:
The `rules` file is missing some lines to install the plugin and copy out the results - it should match https://github.com/theforeman/foreman-packaging/blob/deb/develop/plugins/ruby-foreman-discovery/debian/rules#L14-L24 exactly. Also ensure that hard tabs are used, as it's a Makefile.

timogoebel:
@domcleal : Thanks. That did the trick.

Now we're finally at

```
NoMethodError: undefined method `reloader' for #<Foreman::Application:0x00000004379ae8>
```

domcleal:
I've re-triggered the nightly build job to rebuild Foreman with apipie-rails 0.5.1, please try testing it again tomorrow after 08:30 UTC (once the updated packages are available).

timogoebel:
@domcleal : Thanks for your invaluable help with this. Tests pass now. I've found an issue in 3.0.1 I want to fix before packaging this. I'll open a separate PR for this and close this for the time being.

PR #1641, opened 2017-05-08
Update hammer to 0.10.2 (for 1.15, DEB)

mmoll:
This should be built into:

* [x] 1.15

@dLobatog @domcleal @mbacovsky @tstrachota please have a look. I think this should be part of 1.15 RC3.


domcleal:
Merged, thanks @mmoll. Builds 1622-1627 of deb_dependency are running to end up in staging for 1.15.1 (as requested on foreman-dev).

PR #1648, opened 2017-05-11
Replace unit file with shipped example [deb]

domcleal:
Provided since http://projects.theforeman.org/issues/19480.


mmoll:
merged, thanks @domcleal!

PR #1649, opened 2017-05-11
Require matching Foreman subpackage versions [deb]

domcleal:


mmoll:
seems to work like intended 👍

mmoll:
merged, thanks @domcleal!

PR #1664, opened 2017-05-31
[DEB] Add net-ssh-krb

adamruzicka:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

---


mmoll:
This will be a dependency and not a smart proxy plugin, correct? Then I have a few nits, but would like to wait until the rex-proxy PR itself went in.

adamruzicka:
@mmoll It is a package which, if installed, enables certain feature in the remote execution plugin, but rex can work without it. So optional dependency would probably be the most accurate description

adamruzicka:
@mmoll All the code PRs went in, could you please give me some feedback on this?

mmoll:
I'll try to redo this (for inclusion into upstream Debian) and open a PR to your repo then in the next days.

mmoll:
I added (slightly modified for older versions) files from the stuff I'll submit for upstream. Please move back everything you want to change (Maintainer and so on), and ping back.

As this is a dependency of a proxy plugin rather than a proxy plugin itself, I think it shouldn't get loaded directly via bundler-ext?

adamruzicka:
@mmoll ping, so I set myself as maintainer for the package and also put myself into the changelog if that's ok with you. Could you please take a look at it again?

> As this is a dependency of a proxy plugin rather than a proxy plugin itself, I think it shouldn't get loaded directly via bundler-ext?

Sounds about right

mmoll:
all fine then, merged, díky @adamruzicka!

PR #1667, opened 2017-05-31
Pin npm to < 5.0.0 (DEB)

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.15
* [x] 1.14
* [x] 1.13

While here, unify the whole npm thing also for Ubuntu/xenial.

### DIFF
--- a/debian/xenial/foreman/rules
+++ b/debian/xenial/foreman/rules
@@ -11,7 +11,8 @@ build:
 	/bin/cp config/settings.yaml.example config/settings.yaml
 	/bin/cp config/database.yml.example config/database.yml
 	/bin/rm -rf node_modules
-	/usr/bin/npm install --no-optional
+	/usr/bin/npm install npm@'<5.0.0'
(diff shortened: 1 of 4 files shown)


debian/xenial/foreman/rules
```diff
@@ -11,7 +11,8 @@ build:
 	/bin/cp config/settings.yaml.example config/settings.yaml
 	/bin/cp config/database.yml.example config/database.yml
 	/bin/rm -rf node_modules
-	/usr/bin/npm install --no-optional
+	/usr/bin/npm install npm@'<5.0.0'
+	/usr/bin/nodejs node_modules/.bin/npm install --no-optional
```
dLobatog: Why did xenial use the system version of npm to install node_modules before?

mmoll: Because it was new enough :)

domcleal:
Merged, thanks @mmoll.

PR #1670, opened 2017-06-01
Set proper download URLs for rubygems

komidore64:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

---

### DIFF
--- a/rubygem-algebrick/rubygem-algebrick.spec
+++ b/rubygem-algebrick/rubygem-algebrick.spec
@@ -10,7 +10,7 @@ Release: 4%{?dist}
 Group: Development/Languages
 License: GPLv3
 URL: http://github.com/pitr-ch/algebrick
-Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
+Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
 
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
 Requires: %{?scl_prefix_ruby}ruby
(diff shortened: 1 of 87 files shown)


domcleal:
Er, why?

komidore64:
To have unified URLs across rubygems. I agree that it's small change, but It still has merit in terms of shooting for unified and similar style specs in the long run.

dLobatog:
I'd 👎  this change, it's going to be irrelevant unless the packages are rebuilt for a new version. I appreciate the concern but maybe a suitable alternative that actually impacts the RPMs is to ensure we use the `/gems` path on new gems.

[gem2rpm](https://github.com/fedora-ruby/gem2rpm/blob/master/lib/gem2rpm.rb#L59) defaults to the `/gems` path. This is the tool most people would be using to construct these specs, so I'd rather standardize over that (on new gems or rebuilds, not old specs).

komidore64:
@dLobatog ah that's good to know!

I could switch all URLs to `/gems` if that's more desired. There are a few specs that contain `/downloads`.

```
$ git grep -l 'Source[0-9]\+:\s\+.*/downloads/.*\.gem' -- '*.spec' | wc -l
87
```

ehelms:
@komidore64 planning to make the updated change to use `/gems` consistently?

komidore64:
Yes, apologies. Got side tracked with other tasks. Should hopefully have this PR updated within a week's time.

komidore64:
updated

ehelms:
@komidore64 needs a rebase please

komidore64:
rebased

PR #1675, opened 2017-06-06
Refs #19923 - force nokogiri lower version limit

mmoll:
This should be built into:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

Because of whatever reason (bundler bug?) nokogiri 1.6.6.4 is chosen else.

### DIFF
--- a/debian/jessie/foreman/debian.rb
+++ b/debian/jessie/foreman/debian.rb
@@ -1,1 +1,2 @@
 gem 'concurrent-ruby', '= 1.0.3'
+gem 'nokogiri', '~> 1.7'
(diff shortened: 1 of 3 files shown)


mmoll:
Trusty uses Ruby 2.0 and weill die very soon, see theforeman/foreman-infra#297

domcleal:
Should the minimum be raised to 1.7.x? Some build combinations are using it already, I don't think there's a reason to prefer 1.6.x either.

mmoll:
Yeah, we can do that, let's see, where the builds are going... 🏗

mmoll:
looking good 💚

domcleal:
Thanks @mmoll.

PR #1681, opened 2017-06-09
Core projects: Release 1.15.1 [deb]

dLobatog:
Notice I'm sending this to deb/1.15 only.

### DIFF
--- a/debian/jessie/foreman-installer/changelog
+++ b/debian/jessie/foreman-installer/changelog
@@ -1,3 +1,9 @@
+foreman-installer (1.15.1-1) stable; urgency=low
+
+  * 1.15.1 released
+
+ -- Daniel Lobato Garcia <me@daniellobato.me>  Fri, 09 Jun 2017 09:33:17 +0200
+
 foreman-installer (1.15.0-1) stable; urgency=low
 
   * 1.15.0 released
--- a/debian/jessie/foreman-proxy/changelog
+++ b/debian/jessie/foreman-proxy/changelog
@@ -1,3 +1,9 @@
+foreman-proxy (1.15.1-1) stable; urgency=low
+
+  * 1.15.1 released
+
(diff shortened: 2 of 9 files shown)


domcleal:
[test] as the tarballs and signatures have now been uploaded.

domcleal:
The build appears to be failing, can you look please? At a guess, the release is missing the fix from http://projects.theforeman.org/issues/19524 but it is set to 1.15.1. Is it not in the release? Are you sure the release is installable?

domcleal:
If Foreman was released without the fix then you should perhaps add it as a patch series in the package files here so it's added at build time, or _perhaps_ in the `debian.rb` bundler.d file.

dLobatog:
@domcleal I added it to 1.15.1 on Redmine when I noticed this one was failing, then cherry-picked the changes to smart-proxy and Foreman. The pipeline wasn't kicked off though, so there is no release at the moment - http://ci.theforeman.org/view/Release%20pipeline/ 

To fix it I've already cp-ed the fixes to the right tags, made new tarballs (they are signed already too) and made a new package in https://github.com/theforeman/foreman-packaging/pull/1683

As for debs - I think after repushing this, it'll pass tests now. If it doesn't I will update the `debian.rb` file for Jessie.

domcleal:
Wait, you've modified the already-released tags and tarballs? Shouldn't the version number be changed?

dLobatog:
@domcleal I modified the tags and tarballs - wouldn't say "already-released" as they were not released (no announcement, no rpm/debs published). I would call it `1.15.1-2` but `1.15.1-1` never existed as a deb.

dLobatog:
Repushed, I was missing a comma between 'roadie-rails' and '= 1.1.1'

dLobatog:
Trying without debian.rb

mmoll:
💚, merged, gracias @dLobatog!

PR #1705, opened 2017-06-28
Refs #17581 - switch package builds npm -> yarn

mmoll:
The resulting package seems to work just fine.

This should be built into:

* [x] Nightly

I agree that excessive rm -f is a bit overcautions, but it's not expensive and better safe then sorry. 🤕


ohadlevy:
@mmoll any idea why arm build failed?

mmoll:
`bad CRC 2b18c9ec` at unzipping of the foreman source - I'm seeing this from time to time... but IMHO nothing that points into the direction that this change would break that build.

ohadlevy:
thanks @mmoll

PR #1707, opened 2017-06-28
[RPM] Update kafo to 2.0.2

ares:

* [x] Nightly
* [x] 1.15
* [ ] 1.14
* [ ] 1.13

---


ehelms:
@ares Just to double check, the bug fixes in 2.0.2 are not worth 1.15 z-stream?

ares:
I think it would be good but in past, we were only able to build plugin into released versions. If you and @dLobatog are ok with merging into 1.15 then let's do it. There are 4 fixes since last 1.15 version which is 2.0.0, all should be safe, the bigger 2 were tested in nightlies for a while. I'll update DEB PR too once it's confirmed.

ares:
The 19215 fixes issues when parsing the typed params. I haven't seen the problem myself but the reproduced illustrated it pretty well. The second one is very safe, I haven't seen the issue in foreman-installer but katello-installer would probably appreciate that as it seems they define the type which would be registered twice.

Long story short: it should be safe even without huge impact on foreman-installer. I tested 2.0.2 with 1.15 installer and it worked fine.

mmoll:
what's the status here?

ehelms:
@dLobatog your call then here on whether I build this into 1.15. I think it has fixes that affect folks and is fairly safe myself.

ehelms:
@dLobatog ping

mmoll:
OPs comment in #1708 ;)

ares:
ok, OP sets 1.15, he thought it was against rules which seems no longer the case :-)

ares:
any chance this gets in? debs have been made available for some time

mmoll:
merged, díky @ares!

PR #1711, opened 2017-06-30
Update ldap_fluff to 0.4.7

ares:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

---


ares:
I'm not aware of anything, just that we usually don't backport libraries back to older versions of Foreman, only plugins were usually built for older versions.

There's 0.4.6 in 1.15 so it should be safe to put this there. 1.14 has 0.4.4 which means bringing two more fixes, both seems safe enough. If you're OK with it, let's build it into older repos too. Feel free to update the checkboxes according to your decision.

ares:
Sorry, to answer the question, the PR you mentioned is the only use of the new 0.4.7 functionality as far as I know and that is the only new thing in 0.4.7 when compared to 0.4.6.

ehelms:
@dLobatog based on the discussion, do you want this update in 1.15?

PR #1713, opened 2017-07-07
Release 1.15.2 [rpm]

dLobatog:
Only for 1.15.2  - remind me when the tags are pushed so the pipeline can continue please (http://projects.theforeman.org/projects/foreman/wiki/Release_Process#Packaging-a-release)


foreman-installer/foreman-installer.spec
```diff
@@ -1,12 +1,12 @@
 # set and uncomment all three to set alpha tag
-#global alphatag RC2
-#global dotalphatag .%{alphatag}
-#global dashalphatag -%{alphatag}
+#%global alphatag
+#%global dotalphatag .%{alphatag}
+#%global dashalphatag -%{alphatag}
```
ehelms: Ah ha! This is the problem -- you can't comment out a definition directly. RPM spec will still process this. Thats why before it had dropped the `%` and I believe is the cause of the issue.

dLobatog:
[test]

mmoll:
merged (not built/tagged, I guess you will do this with Jenkins), gracias @dLobatog!

dLobatog:
@mmoll It needs to be tagged, Jenkins builds from tags http://ci.theforeman.org/view/Release%20pipeline/job/release_packages/85/

mmoll:
I gues only @ehelms can help for now then... I do have a tito VM, but not with me at the moment and didn't boot since ages...

PR #1712, opened 2017-07-03
Update smart_proxy_dhcp_infoblox to 0.0.6

witlessbird:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [x] 1.15
* [ ] 1.14
* [ ] 1.13

See Foreman's [plugin maintainer documentation](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

PR #1723, opened 2017-07-12
Update infoblox to 2.0.4

ehelms:

* [x] Nightly
* [x] 1.15
* [ ] 1.14
* [ ] 1.13

---


ekohl:
I'm not too familiar with infoblox but given this is a major bump (due to the default WAPI version changing), how will this impact existing users?

ehelms:
@ekohl I honestly have no idea. But https://github.com/theforeman/foreman-packaging/pull/1712 release requires this update to work.

ekohl:
Shouldn't that also be reflected in the plugin's dependencies then? Currently the Gemfile states ```>= 0.4.1```. If a user has ```foreman_proxy::plugin::dns::infoblox``` with ```ensure => latest``` then it results in a broken setup and since it also affects 1.15 that's not very nice.

ehelms:
@ekohl Touche, the Gemfile is what specifies it -- https://github.com/theforeman/smart_proxy_dhcp_infoblox/blob/master/Gemfile#L4

Now as to why I don't know, since I don't maintain that plugin. Maybe @witlessb can help us out.

ekohl:
@witlessbird could you have a look?

witlessbird:
@ekohl: not sure what the question is? The latest release of infoblox provider requires infoblox gem v. 2.0.4. IIRC there were no internal api changes and this release should work with older versions of smart-proxy (the plugin states the minimal smart-proxy version is 1.13).

ekohl:
@witlessbird if the infoblox provider requires version >= 2.0.4 then I'd expect to see that in the Gemfile.

witlessbird:
@ekohl: I think the confusion is due to dhcp_infoblox provider requiring infoblox gem version >= 2.0.4, while dns_infoblox provider specifying infoblox gem version >= 0.4.1. I'm not sure if there were any changes in dns api in infoblox gem v.2.0.4. @ace13 might know?

ace13:
The infoblox gem update seems to only add additional types - and attributes for the types - it exposes, so except for the WAPI change there shouldn't be any noticeable differences for existing code.

ehelms:
I see it in the Gemfile here -- https://github.com/theforeman/smart_proxy_dhcp_infoblox/blob/master/Gemfile#L4 but I thought the conversation would be around whether this should be required in the gemspec so that bundler respects the requirement.

witlessbird:
I don't think that version specification is that useful -- on install/reinstall bundler will use latest available versions of gems anyway.

ekohl:
@witlessbird you're right, I was looking at the dns_infoblox. You're right that the Gemfiles are correct but in that case https://github.com/theforeman/foreman-packaging/blob/e010d978c4b5c7fb4626d8beb49dd092eaca8901/rubygem-smart_proxy_dhcp_infoblox/rubygem-smart_proxy_dhcp_infoblox.spec#L25 needs a ```>= 2.0.4```.

Klaas-:
can someone verify dns_infoblox still works? then this should be good to go :)

ace13:
Our instance runs with both dns_infoblox and dhcp_infoblox, which means the updated infoblox gem. Of course, they're not clean installs due to the fact that this hasn't been released yet - and due to being manually patched during the development of a few of my patches, so I'd prefer not being the only person to verify it.

witlessbird:
Confirmed that dns_infoblox v0.0.6 works with infoblox gem v.2.0.4. Also: released dns_infoblox v0.0.6 gem and submitted a PR to foreman-packaging with corresponding rpm package changes.

Klaas-:
@ehelms this one is good to go I'd say :)

Klaas-:
thank you @ehelms

PR #1727, opened 2017-07-16
Require ruby in Foreman due to webpack plugin script

ehelms:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

---

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -286,8 +286,8 @@ BuildRequires: npm(seamless-immutable) >= 7.0.1
 BuildRequires: npm(seamless-immutable) < 8.0.0
 BuildRequires: npm(select2) >= 3.5.2
 BuildRequires: npm(select2) < 3.6.0
-BuildRequires: npm(stats-webpack-plugin) < 1.0.0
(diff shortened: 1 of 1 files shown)


ehelms:
Due to the addition of the script [1] for webpack compile.


[1] https://github.com/theforeman/foreman/blob/develop/script/plugin_webpack_directories.rb

foreman/foreman.spec
```diff
@@ -719,7 +719,9 @@ sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/se
 export BUNDLER_EXT_GROUPS="default assets"
 ln -s %{nodejs_sitelib} node_modules
 export NODE_ENV=production
+%{?scl:scl enable %{scl} "}
 webpack.js --bail --config config/webpack.config.js
```
mmoll: in the DEBs we're using `bundle exec rake webpack:compile`, maybe using that with scl_rake is more concise?

ohadlevy: @mmoll the script itself does not run under rails/rake it simply need access to bundler gem in order to find all plugin paths.

PR #1734, opened 2017-07-19
Update foreman-tasks to 0.10.0 (DEB)

iNecas:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

---


plugins/ruby-foreman-tasks/debian/gem.list
```diff
@@ -1,6 +1,6 @@
 # list of gem urls to download via Jenkins, space separated
-GEMS="https://rubygems.org/downloads/foreman-tasks-0.9.3.gem"
-GEMS="$GEMS https://rubygems.org/downloads/foreman-tasks-core-0.1.3.gem"
+GEMS="https://rubygems.org/downloads/foreman-tasks-0.10.0.gem"
+GEMS="$GEMS https://rubygems.org/downloads/foreman-tasks-core-0.2.0.gem"
 GEMS="$GEMS https://rubygems.org/downloads/daemons-1.2.4.gem"
```
mmoll: this seems to be in Foreman core with recent nightlies now and needs to get removed here

mmoll:
hm, I guess the dynflow plugin is <1.16 only now, so the packaging needs to change and there will be a split in 1.15 and nightly PRs?

iNecas:
You mean dynflow needs to get out of the gem, right? This is ok, as we are not targeting 0.10.0 for older releases (we will release 0.9.x if needed in 1.15)

plugins/ruby-foreman-tasks/debian/gem.list
iNecas: Good point, updated.

mmoll:
[test]

mmoll:
merged, díky @iNecas!

PR #1740, opened 2017-07-25
Updated rubygem-foreman_maintain to 0.0.8

gnurag:
Updated rubygem-foreman_maintain to 0.0.6 that adds support for consistent config, lib and log directory locations.


ehelms:
[test]

ehelms:
@gnurag if you join the theforeman github organization and set yourself as a public memeber this will auto-test in the future for you.

gnurag:
[test]

rubygem-foreman_maintain/rubygem-foreman_maintain.spec
```diff
@@ -2,14 +2,14 @@
...
+%global package_name foreman-maintain
 
 %{!?_root_bindir:%global _root_bindir %{_bindir}}
 
 Summary: The Foreman/Satellite maintenance tool
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.0.3
-Release: 1%{?dist}
+Version: 0.0.6
+Release: 2%{?dist}
```
ehelms: No need to increment the release here since its a new version.

gnurag: @ehelms What happened is, the above [test] comment didn't trigger auto tests, so I figured bumping something may trigger it. I'll set it back to 1.

iNecas: While updating the Release to 1, bump the version number to 0.0.7 as well please, as I've just pushed the new version to rubygems: it includes the the full upgrade path work.

gnurag: Thanks @iNecas - Bumped Version up to 0.0.7 and Release down to 1.

gnurag:
[test]

rubygem-foreman_maintain/rubygem-foreman_maintain.spec
```diff
@@ -2,13 +2,13 @@
 %{!?scl:%global pkg_name %{name}}
 
 %global gem_name foreman_maintain
-%global confdir foreman_maintain
+%global package_name foreman-maintain
```
ehelms: Curious why the package_name differs from the gem_name ? That seems like it can lead to confusion.

ekohl: Especially when the gem does follow the ```foreman_<name>``` convention. Using the same convention in the package name makes it very easy to install in puppet-foreman.

iNecas: I don't think the `package_name` actually means what it's used for. Mainly, we wanted to match the executable `foreman-maintain` with the directories in `/etc` and `/var/lib`, but I could probably live with `/etc/foreman_maintain`, unless other ideas how to approach this.

gnurag: Yes, that was the idea behind switching to a different macro. I wanted to keep the convention of using hyphens in `foreman-*` directory names in etc and var. But, perhaps `package_name` was a poorly chosen macro name. I'm good with `/etc/foreman_maintain` too if there's no opposition.

ehelms: Oh, I do see your point @iNecas . And I think you logic makes sense. @ekohl any thoughts?

ekohl: That makes sense to me. Would naming it ```directory_name``` prevent future confusion?

iNecas: 👍

gnurag: Thanks @ekohl for the suggestion. I've renamed `package_name` to `directory_name`.

rubygem-foreman_maintain/rubygem-foreman_maintain.spec
```diff
@@ -2,13 +2,13 @@
...
 %global gem_name foreman_maintain
-%global confdir foreman_maintain
+%global directory_name foreman-maintain
 
 %{!?_root_bindir:%global _root_bindir %{_bindir}}
 
 Summary: The Foreman/Satellite maintenance tool
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.0.3
+Version: 0.0.7
```
iNecas: We released 0.0.8 in the meantime: worth updating to safe some review time :)

ehelms:
Changes look good, I'll merge it once its updated to 0.0.8 @gnurag

ehelms:
@gnurag Please update so I can throw this into 1.15 and recently branched 1.16.

gnurag:
@ehelms @iNecas Pushed updated spec for 0.0.8.

iNecas:
The version is updated. Merging, Thanks @gnurag

PR #1751, opened 2017-08-04
Foreman with node-modules tar

ehelms:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

---


ehelms:
This is an idea of what https://github.com/theforeman/foreman/pull/4230 would look like from a RPM build perspective. This includes the binaries which we would not include of course. I wanted the contrast to be shown if we go this route.

ekohl:
In its current form this is not mergeable and I'm going to close this until we decide to actually go this route to keep the open PRs limited.

PR #1755, opened 2017-08-08
Change asset precompilation to use development Env for plugins

ehelms:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

---

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -888,7 +888,7 @@ unlink tmp \\
 rm \`pwd\`/config/initializers/encryption_key.rb \\
 /usr/bin/%%{?scl:%%{scl}-}rake security:generate_encryption_key \\
 export BUNDLER_EXT_NOSTRICT=1 \\
-%%{?-s:/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=production --trace} \\
+%%{?-s:/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=development --trace} \\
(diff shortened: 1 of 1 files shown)


ehelms:
This step fails when using SCSS files (http://koji.katello.org/kojifiles/work/tasks/8233/18233/build.log) but does not if run with development RAILS_ENV. Given the plugin asset task configures assets without regard for environment, we can safely set this when ew run this step.

ehelms:
@dLobatog can you take a look at this for me? From my local testing, this should unblock Katello.

foreman/foreman.spec
```diff
@@ -888,7 +888,7 @@ unlink tmp \\
 rm \`pwd\`/config/initializers/encryption_key.rb \\
 /usr/bin/%%{?scl:%%{scl}-}rake security:generate_encryption_key \\
 export BUNDLER_EXT_NOSTRICT=1 \\
-%%{?-s:/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=production --trace} \\
+%%{?-s:/usr/bin/%%{?scl:%%{scl}-}rake %%{-r*}%%{!?-r:plugin:assets:precompile[%%{-n*}%%{!?-n:%%{gem_name}}]} RAILS_ENV=development --trace} \\
```
ohadlevy: why is this required? (/me just wondering).

ehelms: See my comment above, this fixes plugin asset compilation after the sprockets-rails updates.

mmoll: I have no ideas of RPMs, but with DEBs it still seems to work for plugins with RAILS_ENV=production

ehelms: THis doesn't fail for all plugins, just, as far as I can tell, those that have SCSS files.

dLobatog:
https://github.com/Katello/katello/pull/6901 
and 
https://github.com/theforeman/foreman/pull/4734

should allow Katello assets to compile again.

PR #1760, opened 2017-08-11
Fixes #20557 - Clear tmp session/cache on upgrade

dLobatog:
As per requested in http://projects.theforeman.org/issues/20557


ehelms:
Is this something that only affects the RPM installations? Should the installer be doing this? @mmoll

mmoll:
I guess we should do that in DEBs also, the packages are the right place IMHO. The rake jobs in puppet-foreman in `manifests/database.pp` and it could also get added there.

I'm wondering if some cache clearing would also be needed after plugin installs?

ehelms:
Oh wait, wouldn't this be only done in the buildroot? Oh, sorry. Man, I feel like our RPM does a lot that is risky on install given the complexity and what things the installer is responsible for.

mmoll:
`%postrans` -> "Runs after all RPM package transactions." This should be fine, no?

PR #1761, opened 2017-08-11
add plugin foreman_snapshot_management [rpm]

timogoebel:
Please build for

* [x] Nightly
* [x] 1.15
* [x] 1.14
* [ ] 1.13


timogoebel:
[test]

timogoebel:
test failures look unrelated:
```diff
Error: Package: foreman-1.16.0-0.develop.201705291400git8f7c12b.el7.noarch (build)
           Requires: tfm-rubygem(fast_gettext) < 1.2.0
           Available: tfm-rubygem-fast_gettext-1.4.1-1.el7.noarch (build)
               tfm-rubygem(fast_gettext) = 1.4.1
Error: Package: foreman-1.16.0-0.develop.201705291400git8f7c12b.el7.noarch (build)
           Requires: tfm-rubygem(fast_gettext) < 1.2.0
           Installing: tfm-rubygem-fast_gettext-1.4.1-1.el7.noarch (build)
               tfm-rubygem(fast_gettext) = 1.4.1
```

@mmoll / @ehelms: Do you have any insights? ^^

mmoll:
[test]

rubygem-foreman_snapshot_management/rubygem-foreman_snapshot_management.spec
```diff
@@ -0,0 +1,90 @@
...
+%{foreman_bundlerd_plugin}
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/README.md
+%{gem_instdir}/Rakefile
+%{gem_instdir}/test
+
+%posttrans
+%{foreman_restart}
```
ehelms: Is this a common behavior? Seems dangerous that a yum update will force restart the application.

ehelms:
[test]

Klaas-:
@ehelms yeah, I think that's normal. You'd want the plugin to work after you have installed the rpm and that means restarting foreman. I think most other foreman plugins to the same, I checked a few:
* https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-foreman-tasks/rubygem-foreman-tasks.spec#L127
* https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-foreman_ansible/rubygem-foreman_ansible.spec#L99
* https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-foreman_azure/rubygem-foreman_azure.spec#L97
* https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-foreman_bootdisk/rubygem-foreman_bootdisk.spec#L100
* https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-foreman_chef/rubygem-foreman_chef.spec#L97

rubygem-foreman_snapshot_management/rubygem-foreman_snapshot_management.spec
timogoebel: Although I'm not fond of it, it's defined in the gem2rpm templates. https://github.com/theforeman/foreman-packaging/blob/rpm/develop/gem2rpm/foreman_plugin.spec.erb#L185

ehelms:
OK -- so its expected that a yum update would potentially take down the application? I realize this is systematic, but I am more on the curious side that this is something most users anticipate? @Klaas- @timogoebel

timogoebel:
@ehelms : To be honest: I _believe_, this doesn't actually work. I always restart `httpd` manually after a plugin install / update. What really annoys me is the seeding, migrating and apipie-cache generating for _every_ single plugin. This makes updates extremely slow. And it happens in silence without the rpm package telling me what it currently does.
A better experience would be to do it once per "yum run" and not for every plugin. I hope, I'm making myself clear.

ehelms:
@timogoebel we are getting into some discussions probably better done on a broader forum. Not that I want to stop them, I think these are really good things to talk about. Maybe I should move this to foreman-dev email list and quote our current discussion?

timogoebel:
@ehelms: Yes, please go ahead. Regardless of this: We should not block this PR before we reach a conclusion on the general topics. ;-)

Klaas-:
just for reference: I tried the rpm koji built and it seems to work, a manual systemctl restart httpd was still needed though :)

Klaas-:
it's defined here: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/foreman/foreman.spec#L837 and it doesn't work (here at least) because:
```
# /sbin/service foreman status
Foreman is running under passenger                         [PASSED]
# echo $?
1
```
I'll create a PR to fix that :)

PR #1768, opened 2017-08-16
Update foreman.spec to restart if in passenger mode [rpm]

Klaas-:
Previously this was not actually restarting foreman. I broke the rhel6 compatibility.

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

---

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -834,7 +834,7 @@ cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name} << EOF
 %%%{name}_rake         %{foreman_rake}
 %%%{name}_db_migrate   %%{%{name}_rake} db:migrate >> %%{%{name}_log_dir}/db_migrate.log 2>&1 || :
 %%%{name}_db_seed      %%{%{name}_rake} db:seed >> %%{%{name}_log_dir}/db_seed.log 2>&1 || :
-%%%{name}_restart      (/sbin/service %{name} status && /sbin/service %{name} restart) >/dev/null 2>&1
+%%%{name}_restart      (/sbin/service %{name} status && /sbin/service %{name} restart || /bin/touch %%{%{name}_dir}/tmp/restart.txt) >/dev/null 2>&1
 EOF
 
 # Keep a copy of the schema for quick initialisation of plugin builds


Klaas-:
@ehelms I think this could be the appropriate PR to discuss if you really want to restart httpd on plugin install :)

Klaas-:
also a question if you want to keep restarting foreman: is a ```touch ~foreman/tmp/restart.txt``` enough on plugin install?

ehelms:
I feel like I should know this but, what are the other modes with RPM install? 

To your question, I wouldn't think that would cause an Apache restart to just touch that file.

Klaas-:
I'm not 100% sure an apache restart is actually needed, I know the restart.txt should reload the foreman application; that could be enough after a plugin install but someone with more rails knowledge should answer that; and other modes I think this is from waay back when you still started it with ```rails s``` via /etc/init.d/foreman

timogoebel:
How about restarting `foreman-tasks` service as well? It can cause strange errors if you forget to do that.

ekohl:
So at least in [foreman.init](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/foreman/foreman.init) if ```FOREMAN_USE_PASSENGER``` is set to ```1``` it will restart passenger using ```restart.txt``` which is sufficient to restart the app. puppet-foreman sets that value for you so in a properly configured foreman install this shouldn't be needed.

Klaas-:
@ekohl the service restart is not reached atm because /sbin/service foreman status doesn't exit with 0; so I thought we should update it to use systemctl or just create the restart.txt directly if that is sufficient after a plugin install

Klaas-:
I tested it and ```touch ~foreman/tmp/restart.txt``` is enough after a plugin install, I'll replace the service restart with a call to that directly. I think we should also consider cleaning up that old init script :) This doesn't break rhel6 anymore if that is relevant downstream

Klaas-:
okay, this should work I'd say :) if there is consensus about adding foreman-tasks I can do that aswell

ekohl:
I don't have an install at hand right now, but does ```systemctl try-restart foreman``` work when running in Passenger mode or does it suffer from the same problem? At home I have a WIP patch to remove all use of ```service``` in this file and use the systemd macro's instead.

ehelms:
Given foreman-tasks is optional, and the service is moving to dynflow I'd say not to include that.

Klaas-:
@ekohl ```systemctl try-restart foreman``` doesn't work because foreman is not actually started by systemd. So try-restart does nothing :) I'd say the direct touch of the restart.txt is the best option as of now so this PR should be good to test @ehelms :)

ekohl:
@Klaas- In that case I actually think we should do both: in case a user runs it as a (local) service or through passenger.

Klaas-:
@ekohl okay, this should work then :)

ehelms:
[test]

Klaas-:
http://koji.katello.org/kojifiles/work/tasks/2170/22170/root.log 
`Error: No Package found for tfm-rubygem(fog-core) = 1.44.2`
is there something wrong with the foreman repositories?

ekohl:
@Klaas- you're right that nightlies are still not building. It's being worked on.

Klaas-:
this is not needed anymore with https://github.com/theforeman/foreman-packaging/commit/114dd478c6e4ea55ee579b85032a158cc4f76842

PR #1772, opened 2017-08-23
Update node-gyp requirements

dLobatog:
* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

Meant to fix the errors seen in nightly http://ci.theforeman.org/job/packaging_repoclosure/36504/console

```
07:29:08 Num Packages in Repos: 30464
07:29:08 package: nodejs-node-gyp-3.3.1-1.el7.noarch from undertest-yum_el7-4203183943-68
07:29:08   unresolved deps: 
07:29:08      http-parser-devel
```

### DIFF
--- a/nodejs-node-gyp/nodejs-node-gyp.spec
+++ b/nodejs-node-gyp/nodejs-node-gyp.spec
@@ -132,10 +132,14 @@ Source118: addon-rpm.gypi
 Patch1: node-gyp-addon-gypi.patch
 Requires: nodejs(engine)
 # this is the standard set of headers expected to build any node native module
-Requires: nodejs-devel libuv-devel http-parser-devel
+Requires: nodejs-devel
+Requires: libuv-devel
+Requires: http-parser-devel
 # we also need a C++ compiler to actually build stuff ;-)
 Requires: gcc-c++
 BuildRequires: nodejs-devel
+BuildRequires: libuv-devel
+BuildRequires: http-parser-devel
 BuildRequires: nodejs-packaging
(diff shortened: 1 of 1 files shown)


mmoll:
doesn't that mean that there's no package provding `http-parser-devel`?

mmoll:
allright, then I guess the `Release:` field needs a bump for a rebuild.

ehelms:
So there is a bit of misnomer here because of how Koji works. So http-parser-devel was pulled from EPEL because it was added to RHEL 7.4 proper. This means that in centos land, this is broken until CentOS 7.4 is released containing http-parser-devel.

Koji has local copies of various repositories (for speed) and thus passes here because it has an older EPEL with the http-parser-devel package available meanwhile real deployments do not.

ehelms:
You can further information here -- https://bugs.centos.org/view.php?id=13669

This puts us in a position of having to either wait for CentOS 7.4, or to, for a brief(ish) period, carry this package in our repositories.

dLobatog:
I think carrying the packages http-parser and http-parser-devel during the 'gap time' is probably the right solution. We already carry 'centos-release-scl' as an extra package (http://projects.theforeman.org/projects/foreman/wiki/RPM_Packaging#Signing-extra-centos-release-scl--RPMs) , I have to check how exactly these https://github.com/theforeman/foreman-packaging/blob/rpm/develop/extras/extras-foreman-rhel7-x86_64 are put into our repo, and do the same for http-parser and http-parser-devel

PR #1777, opened 2017-08-24
Removed f24 from copy-tags-command script

dLobatog:
Seems useless to have this script handle this, keeping the NONSCL variable though in case we decide to later on add any nonscl distribution.

### DIFF
--- a/koji/copy-tags-commands.sh
+++ b/koji/copy-tags-commands.sh
@@ -10,11 +10,11 @@ fi
 OLD=$1
 VERSION=$2
 
-NONSCL_SYSTEMS="fedora24"
+NONSCL_SYSTEMS=""
 SCL_SYSTEMS="rhel7"
 
 clone() {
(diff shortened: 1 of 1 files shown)


koji/copy-tags-commands.sh
```diff
@@ -10,11 +10,11 @@ fi
 OLD=$1
 VERSION=$2
 
-NONSCL_SYSTEMS="fedora24"
+NONSCL_SYSTEMS=""
 SCL_SYSTEMS="rhel7"
 
 clone() {
-  echo kkoji clone-tag $PRODUCT-$OLD-$SYSTEM $PRODUCT-$VERSION-$SYSTEM
+  echo kkoji clone-tag --config $PRODUCT-$OLD-$SYSTEM $PRODUCT-$VERSION-$SYSTEM
```
dLobatog: Newer versions of koji require --config to be passed so that the config is copied from the old to the new tag

PR #1779, opened 2017-08-28
Add http-parser to extras

dLobatog:
* [x] Nightly
* [x] 1.15
* [x] 1.14

The extras/ directory files contains links to rpms that the mash script in Koji downloads and puts in our repos.

As explained in https://github.com/theforeman/foreman-packaging/pull/1772, we have to carry http-parser and http-parser-devel until EL 7.4 is out. A possible solution to do so is to just download the RPM and sign it instead of also building it ourselves.

At the moment nightlies are not passing tests consistently as they cannot be built when the http-parser-devel is not found on Koji (apparently some Koji caches still return it), this ought to fix that problem. See [packaging_repoclosure](http://ci.theforeman.org/job/packaging_repoclosure/) logs of recent failures for an example of that.

No need for signing as (http://projects.theforeman.org/projects/foreman/wiki/RPM_Packaging#Signing-extra-centos-release-scl--RPMs) says, if this doesn't get to a final release.

### DIFF
--- a/comps/comps-foreman-rhel7.xml
+++ b/comps/comps-foreman-rhel7.xml
@@ -9,6 +9,8 @@
     <packagelist>
       <packagereq type="default">centos-release-scl</packagereq>
       <packagereq type="default">centos-release-scl-rh</packagereq>
+      <packagereq type="default">http-parser</packagereq>
+      <packagereq type="default">http-parser-devel</packagereq>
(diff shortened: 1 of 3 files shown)


Klaas-:
https://lists.centos.org/pipermail/centos-announce/2017-September/022532.html I think this can be reverted now I'd guess

PR #1780, opened 2017-08-29
Add rubygem-fog-internet-archive

ehelms:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13

---


ekohl:
From http://ci.theforeman.org/job/systest_foreman/13488/console:

```
# [ WARN 2017-08-29 07:42:36 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: rake aborted!
# [ WARN 2017-08-29 07:42:36 verbose]  /Stage[main]/Foreman::Database/Foreman::Rake[db:seed]/Exec[foreman-rake-db:seed]/returns: Gem::LoadError: Could not find 'fog-internet-archive' (>= 0) among 141 total gem(s)
```

dLobatog:
I see, looks like this has to be looked into then. It's probably going to cause another error after fog-internet-archive is added. Maybe it's because the gem added as fog 1.41 when unpacked doesn't match the patched version in https://github.com/theforeman/fog - I thought the spec did it

ekohl:
I'm guessing https://github.com/theforeman/fog/blob/0406be96cbf38deaa618b7b5838bfdcc141890e5/fog.gemspec#L66 needs to be removed as well.

ehelms:
OK - so sounds like this is not the actual fix and that this has to do with our fog fork?

ekohl:
I think I've spotted the problem: https://github.com/theforeman/foreman-packaging/blob/cde2d83d98bcb7a5f6f1f35c7834c7e72e843878/rubygem-fog/rubygem-fog.spec#L147

That doesn't match ```fog-internet-archive``` (note _ vs -). Submitting a PR.

PR #1792, opened 2017-09-04
Updated foreman_discovery to 9.1.2 (RPM)

lzap:
Only for 1.15.


dLobatog:
Building in http://koji.katello.org/koji/taskinfo?taskID=25528

PR #1794, opened 2017-09-04
foreman: Use systemd macros + fix groups

ekohl:
This is untested because I don't know how to do a local build.


ekohl:
@dLobatog that I couldn't find the tarbal to build with. Maybe ```setup_sources.sh``` should mention this?

PR #1795, opened 2017-09-05
release-package script

dLobatog:
This contains a script originally from katello-packaging, but improved
to automatically:
  - download the gem source
  - pass that source as source_dir to the custom tito releaser

This allows to release a package like this (after tagging):
./release_package.rb -p rubygem-foreman_discovery -r koji-foreman-plugins

--

Goal is to combine this script with another one like `./merge_tag_release --versions 1.15 1.16 develop rubygem-foreman_discovery`, and the script would cherry-pick/tag/release in all branches.

### DIFF
--- a/rel-eng/custom/custom.py
+++ b/rel-eng/custom/custom.py
@@ -55,7 +55,7 @@ class ForemanSourceStrategy(SourceStrategy):
                 self.builder.project_name),
             self.spec_file)
         for s in os.listdir(self.builder.start_dir):
-            if os.path.exists(os.path.join(self.builder.start_dir, s)):
+            if os.path.exists(os.path.join(self.builder.start_dir, s)) and os.path.isfile(s):
                 shutil.copyfile(
                     os.path.join(self.builder.start_dir, s),
                     os.path.join(self.builder.rpmbuild_sourcedir, os.path.basename(s)))
--- a/release_package.rb
+++ b/release_package.rb
@@ -1,0 +1,120 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+
+options = {
+  :jenkins_job_id => 'lastSuccessfulBuild'
+}
+
+OptionParser.new do |opts|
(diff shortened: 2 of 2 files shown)


rel-eng/custom/custom.py
```diff
@@ -55,7 +55,7 @@ def fetch(self):
                 self.builder.project_name),
             self.spec_file)
         for s in os.listdir(self.builder.start_dir):
-            if os.path.exists(os.path.join(self.builder.start_dir, s)):
+            if os.path.exists(os.path.join(self.builder.start_dir, s)) and os.path.isfile(s):
```
ekohl: In this case ```s``` is still relative so you need to join, but it's better to just replace the ```exists()``` with ```isfile()``` since ```isfile()``` implies ```exists()```.

release_package.rb
```diff
@@ -0,0 +1,120 @@
...
+def log(msg)
+  puts
+  puts "== #{msg} =="
+end
+
+log("Initializing git annex")
+puts `git-annex init`
+
+log("Setting up source")
+if options[:project] == 'rubygem-katello' || options[:project] == 'katello-installer'
```
ekohl: These packages don't exist in this repo so I'd leave it out. If they are indeed needed then I think it should be fixed in ```setup_sources.sh``` since that's also used directly by users.

release_package.rb
```diff
@@ -0,0 +1,120 @@
...
+  `./setup_sources.sh #{options[:project]} --relaxed`
+else
+  `./setup_sources.sh #{options[:project]}`
+end
+
+Dir.mkdir("rel-eng/build") if !File.exist?('rel-eng/build')
```
ekohl: Why not ```mkdir_p```?

ekohl:
I'm going to closing this. We want to make https://github.com/evgeni/obal the generic tool that can handle it. It also think it shouldn't live in this repository since it operates on files in the repository itself. From experience I know that we do need the tooling though :)

PR #1797, opened 2017-09-05
merge_tag_release script introduced

dLobatog:
This commit takes a number of branches, e.g: '1.15, 1.16', a commit, and
it puts the commit in these branches, then it tags and releases the
package.

Useful for maintainers only

### DIFF
--- a/merge_tag_release.rb
+++ b/merge_tag_release.rb
@@ -1,0 +1,51 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+
+options = {}
+
+OptionParser.new do |opts|
+  opts.banner = "Usage: merge_tag_release.rb [options]"
+
+  opts.on("-c", "--commit [COMMIT]", "SHA hash of the commit to be cherry-picked") do |commit|
+    options[:commit] = commit
+  end
+
+  opts.on("-b", "--branches [BRANCHES]", "Branches to merge commit to, separated by commas. e.g: '1.16-stable, 1.15-stable' ") do |branches|
+    options[:branches] = branches
+  end
+
+  opts.on("-p", "--project [PROJECT]", "Name of the project e.g: rubygem-foreman_discovery") do |project|
+    options[:project] = project
+  end
+end.parse!
+
+raise OptionParser::MissingArgument, "--branches" if options[:branches].nil?
+raise OptionParser::MissingArgument, "--project" if options[:project].nil?
+raise OptionParser::MissingArgument, "--commit" if options[:commit].nil?
+CORE_PROJECTS = %q(foreman foreman-selinux foreman-proxy foreman-installer)
+
+branches = options[:branches].split(',').map!(&:strip)
+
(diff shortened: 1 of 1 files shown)


merge_tag_release.rb
```diff
@@ -0,0 +1,43 @@
...
+CORE_PROJECTS = %q(foreman foreman-selinux foreman-proxy foreman-installer)
+
+branches = options[:branches].split(',').map!(&:strip)
+
+branches.each do |branch|
+  `git checkout rpm/#{branch}`
+  `git cherry-pick #{options[:commit]}`
+  # Cherry-pick commit to branch
+  `tito tag --keep-version`
+  `git push origin rpm/#{branch} --tags`
```
ekohl: As I recall ```--tags``` only pushes tags, not the actual branches. I also don't have an ```origin``` remote but always use ```ekohl``` and ```upstream```. Another thing is that I personally don't push changes to my personal remotes, just the branches for PRs.

ehelms:
I like where this is headed and making it more one stop shopping with scripts to speed things up.

From some past experiences, of having a script that changes branches actively within the repository the script lives I have experienced some oddnesses that can occur between what's in memory and what git has swapped to.

Not that it affects the script, but there will be times this can fail. As some packages won't build on branches because of the way koji works.

dLobatog:
A problem I've found with this is that releasers are not available. e.g: 

Running `ruby merge_tag_release.rb --branches '1.14,1.15,1.16' -c 34058501b0de4e0faf760303fcaeacc3d2ec10bb --project rubygem-foreman_omaha` from rpm/develop works in develop because koji-foreman-plugins-nightly is available there, but not koji-foreman-plugins-1.14. Not sure yet how to address this other than maintaining a bigger `rel-eng/releasers.conf` (not a problem IMO).

merge_tag_release.rb
```diff
@@ -0,0 +1,51 @@
...
+  `git push upstream rpm/#{branch} --tags`
+  # Release on the releasers for the branch you're in. e.g: koji-foreman-1.15
+  # Plugins are automatically detected.
+  branch = 'nightly' if branch == 'develop'
+  releaser = if CORE_PROJECTS.include? options[:project]
+               "koji-foreman-#{branch}"
+             else
+               "koji-foreman-plugins-#{branch}"
+             end
+  `ruby release_package.rb -p #{options[:project]} -r #{releaser}`
```
dLobatog: Looks like the releaser here should always be `koji-foreman-plugins` or `koji-foreman`. Using the `$RELEASER-$VERSION` format creates builds with strange release names, like: http://koji.katello.org/koji/buildinfo?buildID=16694

ekohl:
I'm going to closing this. We want to make https://github.com/evgeni/obal the generic tool that can handle it. It also think it shouldn't live in this repository since it operates on files in the repository itself. From experience I know that we do need the tooling though :)

PR #1799, opened 2017-09-06
installer: require Puppet 4.9 (RPM)

mmoll:
this should be built into:

* [x] Nightly

This is required since https://github.com/theforeman/puppet-tftp/commit/3a97422f61b37fd07e0583b9ad7eba2e4b66de0c and https://github.com/theforeman/puppet-git/commit/bb41057b854ca3338de276a3527d2b2c69c200e3

### DIFF
--- a/foreman-installer/foreman-installer.spec
+++ b/foreman-installer/foreman-installer.spec
@@ -16,7 +16,7 @@ Source0:    http://downloads.theforeman.org/%{name}/%{name}-%{version}%{?dashalp
 BuildArch:  noarch
 
 Requires:   curl
-Requires:   %{?scl_prefix}puppet-agent >= 1.6.1
+Requires:   %{?scl_prefix}puppet-agent >= 1.9.0
 Requires:   %{?scl_prefix}rubygem-kafo >= 1.0.5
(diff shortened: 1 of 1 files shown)


foreman-installer/foreman-installer.spec
```diff
@@ -16,15 +16,15 @@ Source0:    http://downloads.theforeman.org/%{name}/%{name}-%{version}%{?dashalp
 BuildArch:  noarch
 
 Requires:   curl
-Requires:   %{?scl_prefix}puppet-agent >= 1.6.1
+Requires:   %{?scl_prefix}puppet-agent >= 1.9.0
```
ekohl: Would puppet-agent be in the SCL? I know we build this without SCL so it's probably fine for now.

mmoll: no idea.

ehelms: It's not in the SCL and with the way its built I wouldn't ever forsee it being in so we could drop that.

mmoll:
Could this get in for now? And do the de-SCLing later?

PR #1825, opened 2017-09-22
Add rubygem-unicode

ekohl:
This is a copy of the spec from EPEL7 with spec2scl applied and then
some manual changes to build inside SCL.


rubygem-unicode/license.txt
```diff
@@ -0,0 +1 @@
+../.git/annex/objects/q6/66/SHA256E-s2504--5572bf03c3c8c4770e8153038f5f54faad9c3db9900864a7c82a19de4434d74b.txt/SHA256E-s2504--5572bf03c3c8c4770e8153038f5f54faad9c3db9900864a7c82a19de4434d74b.txt
```
ehelms: I don't think setup_sources will pickup Source1 but this is such a small, non-binary file that you can include it directly.

rubygem-unicode/rubygem-unicode.spec
```diff
@@ -0,0 +1,86 @@
...
+%if 0%{?rhel}
+ruby -I.%{gem_instdir}/lib:.%{gem_extdir_mri} test/test.rb
+%else
+ruby-mri -I.%{gem_instdir}/lib:.%{gem_extdir_mri} test/test.rb
+%endif
+%{?scl:EOF}
+
+
+%files
+%doc README license.txt
```
ehelms: I'd split this into:

```
%doc README
%license license.txt
```

ehelms:
You'll want to add this to the appropriate tag listings in rel-eng/tito.props as well

ekohl:
@ehelms updated

ekohl:
Running this on EL7 does cause hammer not to die but it shows:
```
Ignoring unicode-0.4.4.1 because its extensions are not built.  Try: gem pristine unicode --version 0.4.4.1
```

ekohl:
It does open ```/opt/theforeman/tfm/root/usr/lib64/gems/ruby/unicode-0.4.4.1/unicode/unicode_native.so``` so I don't understand why it doesn't work.

ekohl:
Updated, looks like the ```gem.build_complete``` file needs to exist.

PR #1709, opened 2017-06-29
Provide Copr releaser and fix building of nodejs-brace

xsuchy:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [ ] 1.15
* [ ] 1.14
* [ ] 1.13
---

PR #1827, opened 2017-09-26
Provide Copr releaser and fix building of nodejs-brace 

xsuchy:
This replaces https://github.com/theforeman/foreman-packaging/pull/1709
Regarding the README. This information is already in the main README.  But it is hidden in a bunch of general comments. It really does not hurt to repeat it. As it is really non-standard way how to build a package.

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---

### DIFF
--- a/foreman-installer/README
+++ b/foreman-installer/README
@@ -1,0 +1,2 @@
+how to build this package:
+tito release copr-nightly --arg jenkins_job=packaging_trigger_installer_develop
--- a/rel-eng/releasers.conf
+++ b/rel-eng/releasers.conf
@@ -18,3 +18,9 @@ releaser = tito.release.KojiReleaser
 autobuild_tags = foreman-plugins-nightly-nonscl-rhel7 foreman-plugins-nightly-rhel7
 builder = tito.builder.FetchBuilder
 builder.rpmbuild_options = --define "foremandist .fm1_17"
+
+[copr-nightly]
+releaser = tito.release.CoprReleaser
+project_name = @theforeman/foreman-nightly
+builder.jenkins_url = http://ci.theforeman.org
+builder = tito.builder.FetchBuilder
--- a/foreman-selinux/README
+++ b/foreman-selinux/README
@@ -1,0 +1,2 @@
+how to build this package:
+tito release copr-nightly --arg jenkins_job=packaging_trigger_selinux_develop
--- a/foreman/README
+++ b/foreman/README
@@ -1,0 +1,2 @@
+this package can be built using:
+tito release copr-nightly --arg jenkins_job=test_develop
--- a/nodejs-brace/nodejs-brace.spec
+++ b/nodejs-brace/nodejs-brace.spec
@@ -22,8 +22,8 @@ ExclusiveArch: %{nodejs_arches} noarch
 ExclusiveArch: %{ix86} x86_64 %{arm} noarch
 %endif
 Provides: npm(%{npm_name}) = %{version}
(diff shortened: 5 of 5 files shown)


mmoll:
[test]

mmoll:
[test]

mmoll:
💚

foreman-installer/README
```diff
@@ -0,0 +1,2 @@
+how to build this package:
```
ekohl: I'm not that familiar with the build process and to me this line implies a scratch build while the next implies a release (including a tag). What does this exactly achieve?

ehelms:
I'm sorry but I find the READMEs just add clutter and as pointed out, can be confusing when considering releases vs. nightliies (which are run by a CI system and not by hand). Further, the READMEs are Copr specific and currently that is not the main build system.

rel-eng/releasers.conf
```diff
@@ -18,3 +18,9 @@ releaser = tito.release.KojiReleaser
 autobuild_tags = foreman-plugins-nightly-nonscl-rhel7 foreman-plugins-nightly-rhel7
 builder = tito.builder.FetchBuilder
 builder.rpmbuild_options = --define "foremandist .fm1_17"
+
+[copr-nightly]
+releaser = tito.release.CoprReleaser
+project_name = @theforeman/foreman-nightly
```
ehelms: Is a project the equivalent of a repository? Or can a single project have multiple repositories?

rel-eng/releasers.conf
```diff
@@ -18,3 +18,9 @@ releaser = tito.release.KojiReleaser
 autobuild_tags = foreman-plugins-nightly-nonscl-rhel7 foreman-plugins-nightly-rhel7
 builder = tito.builder.FetchBuilder
 builder.rpmbuild_options = --define "foremandist .fm1_17"
+
+[copr-nightly]
```
ehelms: Thinking we should rename this copr-foreman-nightly to distinguish it from plugins?

ekohl:
I do think that the READMEs could add something. Since those packages don't have tarballs in git they are different. A short note about them is worth something because I'd like to know how they're built. I now know that the jenkins job essentially does a rake task to build a tarball and you commit it to git if you want to do a local build but that's not obvious.

lzap:
I was wondering if we could have a script that would send all changed/new packages to copr. I could run it perhaps from cron until we have better CI integration.

ekohl:
Yesterday I had a chat during lunch with @xsuchy and discussed some details. I believe it should be possible and should provide some benefits over the current process.

After talking to @ehelms we agreed that currently it's not a high priority. We already have a large backlog, 1.16 has been delayed for a long time. For 1.17 we'll first focus on rails 4.2 -> 5.1 which includes updating to a newer ruby SCL. That's blocking our developers where the current koji workflow does what it needs to.

In the short term we should merge the bundled() commit in this PR since that's a fix for an existing issue. After the rails migration we can look at building a parallel workflow next to koji to eventually replace it.

While things could move in parallel, it's best to avoid moving multiple big pieces at once since that makes it harder to know in which piece something broke.

ekohl:
For now I'm going to close this. For the bundled-npm() to bundled(npm()) I've opened https://github.com/theforeman/foreman-packaging/pull/2116 and we intend to use the automation used in tfm-ror51. That already uses copr to do builds. We'll likely need to revisit the building of nightly RPMs but that's something for later.

PR #1833, opened 2017-09-28
add rubygem-multi_json to tito.props

mmoll:
tfm-rubygem(multi_json) is missing currently, I hope that's all needed to get it.


mmoll:
what's left to to do to get this into the yum repo?

PR #1834, opened 2017-09-28
Add bump_rpm.sh script

ekohl:
This is an initial version that only works for rubygems and even takes some shortcuts there. I do believe it automates the most tedious work.

The nodejs situation is much more complex because of the bundled dependencies so I'm not touching that.


ekohl:
Updated

PR #1835, opened 2017-09-28
Get rubygem-ffi from the ror SCL

ekohl:


iNecas:
I can confirm the ffi is in ror, not ruby. Thanks @ekohl

PR #1839, opened 2017-10-01
Revert "Add Ansible dependency for Ansible proxy plugin"

mmoll:
This reverts commit a1171d31a145b5b6bc6c3e9232302b44aadb7705.

this should be built into:

* [x] Nightly
* [x] 1.16
* [x] 1.15

This breaks repoclosure, see i.e. http://ci.theforeman.org/job/packaging_repoclosure/36940/console

AFAICT, the rpm coming with EL7 doesn't support something like "Recommends:" or "Suggests:", yet. And it seems the ansible package isn't part of EPEL right now.

### DIFF
--- a/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
+++ b/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
@@ -24,8 +24,6 @@ Requires: rubygem(smart_proxy_dynflow) < 1.0
 Requires: ruby
 Requires: ruby(rubygems)
 
-Requires: ansible >= 2.2
-
 %if 0%{?rhel} == 6
 Requires: ruby(abi)
 BuildRequires: ruby(abi)


ekohl:
I'm surprised this breaks now because that commit is months old. Ansible 2.3 used to be in EPEL but I suppose they pulled it.

ekohl:
I think enabling the centos7-extras repo is a better solution. I'm not sure if it's sufficient to just do this in the Jenkins jobs or if we need a koji change as well.

mmoll:
As it's only a runtime dependency, the jenkins job should be the only place.

ehelms:
Does the plugin not actually require Ansible? Or is this just to fix a repoclosure issue?

mmoll:
the later.

ehelms:
If it requires Ansible to function, I don't think we should break with that. The repoclosure already has extras defined (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/repoclosure/yum_el7.conf#L26) so we just need to add it into the job -- https://github.com/theforeman/foreman-infra/blob/334c3b2aead604ea2d5d1ed5d8ac79abfe10b7e3/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/packaging_repoclosure.sh#L68

mmoll:
https://github.com/theforeman/foreman-infra/pull/334

PR #1840, opened 2017-10-02
Update React to 15.6.2

dLobatog:
* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

After https://github.com/theforeman/foreman/pull/4877 was merged, this should be updated here too.


ohadlevy:
should this pr also delete unused packages ? or do they get removed some how?

comps/comps-foreman-rhel7.xml
```diff
@@ -394,7 +394,6 @@
       <packagereq type="default">tfm-rubygem-trollop-doc</packagereq>
       <packagereq type="default">tfm-rubygem-unf-doc</packagereq>
       <packagereq type="default">tfm-rubygem-unf_ext-doc</packagereq>
-      <packagereq type="default">tfm-rubygem-unicode-doc</packagereq>
```
mmoll: looks unrelated?

dLobatog: Yes, it was created automatically when running `comps_doc.sh` - the package doesn't exist so it may as well be removed from comps

dLobatog:
Updated to use npm_cache (hardcoded name) instead of /tmp/npm_cache_%{name}-%{version}-%{release} , which seems to cause issues.

PR #1842, opened 2017-10-02
Update prop-types to 15.5.9

dLobatog:
* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

After theforeman/foreman#4877 was merged, this should be updated here too.


dLobatog:
Updated to use just `npm_cache` instead of `/tmp/npm_cache_%{name}-%{version}-%{release}`, it seems that it's not installing properly from cache using that.

PR #1866, opened 2017-10-18
Add nodejs-object-assign package

ekohl:


nodejs-object-assign/nodejs-object-assign.spec
```diff
@@ -0,0 +1,50 @@
...
+
+%prep
+%setup -q -n package
+
+%build
+$BUILD
+
+%install
+mkdir -p %{buildroot}%{nodejs_sitelib}/%{npm_name}
+cp -pfr index.js license package.json readme.md %{buildroot}%{nodejs_sitelib}/%{npm_name}
```
ekohl: I just noticed this: since license and readme.md are also in %files they end up twice in the RPM. Not a big deal but should we fix this? I can submit a PR to ```npm2rpm``` to do so.

PR #1880, opened 2017-10-22
Fixes #21418 - run bundle before npm

mmoll:
this should be built into:
* [x] Nightly


mmoll:
still:
```
15:43:13 npm WARN lifecycle TheForemanDevDeps@1.17.0~postinstall: cannot run in wd %s %s (wd=%s) TheForemanDevDeps@1.17.0 node ./script/npm_install_plugins.js /var/lib/workspace/workspace/packaging_build_deb_coreproject/arch/x86/label/debian/os/xenial/debian/xenial/build-foreman/foreman-1.17.0
```

mmoll:
@xprazak2 any idea here?

mmoll:
the fix is to add `--unsafe-perm`, as the DEB build runs inside its chroot with UID 0

PR #1893, opened 2017-10-25
Update foreman dependencies

ekohl:


comps/comps-foreman-rhel7.xml
```diff
@@ -36,6 +36,7 @@
       <packagereq type="default">http-parser</packagereq>
       <packagereq type="default">http-parser-devel</packagereq>
       <packagereq type="default">rubygem-foreman_maintain</packagereq>
+      <packagereq type="default">rubygem-where-or</packagereq>
```
ehelms: Prefix with tfm-

comps/comps-foreman-rhel7.xml
```diff
@@ -287,6 +288,7 @@
       <packagereq type="default">rubygem-rb-inotify-doc</packagereq>
       <packagereq type="default">rubygem-rsec-doc</packagereq>
       <packagereq type="default">rubygem-rubyipmi-doc</packagereq>
+      <packagereq type="default">rubygem-where-or-doc</packagereq>
```
ehelms: Prefix with tfm-

ekohl:
Updated with SCL-prefix and a README update

ekohl:
Looks like the build fails because the package doesn't exist. Created https://github.com/theforeman/foreman-packaging/pull/1894 which only packages where-or.

ohadlevy:
just worth noting that once we upgrade to rails 5+ its not required.

ekohl:
I noticed that, but for now we're stuck on 4.2 until we update the SCL. We don't want broken nightlies that long.

ehelms:
[test]

ekohl:
Currently fails because https://github.com/theforeman/foreman/commit/b0f7db88d181beecf9b1bfc9c2c551de4d561960 switched from babel-preset-es2015 to babel-preset-env which isn't packaged.

ekohl:
https://github.com/theforeman/foreman-packaging/pull/1897 should fix that

ekohl:
[test]

ehelms:
Testing locally, https://github.com/theforeman/foreman-packaging/pull/1904 will be needed to get past this issue.

ekohl:
[test]

ekohl:
Updated to add / update the nodejs packages.

ekohl:
Note I included some helper scripts I used, but can leave those out if desired.

ehelms:
Moving the helper scripts to a separate PR would be useful for reviewing them individually given the breadth of this PR and number of changes.

ehelms:
Looks like this needs an update now? Or was this replaced?

ekohl:
Needs an update, will do so when the packages are done building in koji. I'll also split off the helpers.

ekohl:
Additional updates in https://github.com/theforeman/foreman-packaging/pull/1931. These need to be added & tagged before the updated foreman RPM will build.

foreman/foreman.spec
```diff
@@ -218,97 +220,150 @@ BuildRequires: http-parser
...
-BuildRequires: npm(babel-core) >= 6.7.2
-BuildRequires: npm(babel-loader) < 7.2.0
+BuildRequires: nodejs-packaging
+
+# package.json devDependencies
+BuildRequires: npm(babel-cli) >= 6.10.1
+BuildRequires: npm(babel-cli) < 7.0.0
+BuildRequires: npm(babel-core) >= 6.26.0
+BuildRequires: npm(babel-core) < 7.0.0
+#BuildRequires: npm(babel-eslint) >= 6.1.2
```
ekohl: I commented some dependencies which are in packages.json but I believe should not be needed to build and are test-only deps.

ekohl:
💚

ehelms:
Looks good! Whats the long term plan for those commented out requires?

ekohl:
No real idea yet. I'm not aware of any ```testDependencies``` or similar. I am considering building some scripting to replace these dependencies automatically and maybe I'll just build a blacklist of packages that are only needed for testing.

PR #1899, opened 2017-10-30
Update dynflow/tasks/rex packages (DEB)

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.16


mmoll:
@iNecas foreman-tasks-core 0.2.x should be good here, correct?

iNecas:
That's right

PR #1906, opened 2017-11-01
Update add_npm_package

ekohl:
It now defaults to single builds of the latest version if no arguments
are given. We also use named variables rather than $1 $2 $3. By using
bash -e we should error out when something fails rather than continuing.

### DIFF
--- a/add_npm_package.sh
+++ b/add_npm_package.sh
@@ -1,12 +1,13 @@
-#!/bin/bash
-display_usage() {
+#!/bin/bash -e
+
+NPM_MODULE_NAME=$1
+VERSION=${2:-auto}
+STRATEGY=$3
+
+if [[ -z $NPM_MODULE_NAME ]]; then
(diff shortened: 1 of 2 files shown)


add_npm_package.sh
```diff
@@ -25,31 +26,47 @@ if !(program_exists npm2rpm); then
...
+    exit 1
+  fi
+
+  if !(program_exists jq) ;  then
+    echo "jq is not installed - you can install it with:"
+    echo "sudo yum install jq"
+    exit 1
+  fi
+
+  VERSION=$(curl -s https://api.npms.io/v2/package/$NPM_MODULE_NAME | jq -r .collected.metadata.version)
```
ehelms: Would be cool if there was a way to default to the latest version allowed by package.json, but I am fine with this for now as a default.

ekohl: I figured that if a dependency is added it's usually the latest so this is good enough for now.

ekohl:
Anything holding this back?

PR #1907, opened 2017-11-01
Update NPM playbook

ehelms:
This is ultimately, not where I want this (or any tooling) to live, but I do not want to lose it. Further, this may spark some good conversation. This is lacking a few other niceties like an ansible.cfg but I wanted to share it and the directions I am thinking of going. This will be part of a proposal to foreman-dev.

For now, this can be used as:

```
ansible-playbook update-npm.yml -e name=babel-core -e version=6.26.0 -e strategy=bundle
```

### DIFF
--- a/update-npm.yml
+++ b/update-npm.yml
@@ -1,0 +1,52 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: 'Check if git branch exists'
+      command: "git rev-parse --quiet --verify {{ name }}-{{ version }}"
+      register: branch
+
+    - name: 'Create update branch'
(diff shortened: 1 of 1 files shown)


ekohl:
We should teach ```npm2rpm``` about changelogs. Currently it generates a dummy but if it's a command line argument then ```add_npm_package.sh``` could just feed it an empty one where this bumping could feed it the old one.

ekohl:
This is now also in the ```add_npm_package.sh``` script. Should we close this for now and revisit when we look at automating more via ansible?

ekohl:
Given the current ```add_npm_package.sh``` script can do this and we'll do the ansible thing in obal I'm going to close this.

PR #1908, opened 2017-11-01
update foreman-tasks to 0.10.6 (DEB)

mmoll:
Also, now that DEB builds are using Rails 5, use Rails 5 compatible
dependencies: sinatra and rack-protection, rack itself comes with the
foreman package, as it's a dependency of Rails.
this should be built into:

* [x] Nightly
* [x] 1.16 - at least I guess it's still 1.16 compatible... @iNecas?


mmoll:
currently blocked until nightly DEBs with Rails 5 are functional again

mmoll:
[test]

PR #1910, opened 2017-11-02
Updated smart_proxy_discovery_image to 1.0.9

lzap:

* [x] Nightly

This is only RPM, no Debian package for this one (it is for FDI).


dLobatog:
nightly - http://koji.katello.org/koji/buildinfo?buildID=18922

PR #1912, opened 2017-11-03
drop unused rubygem-protected_attributes

evgeni:
it is not required anymore since 87526d1eca18d3f06c9bb2234a44a58db46020d5 (see http://projects.theforeman.org/issues/3917), but the packaging was never dropped.


evgeni:
@dLobatog good catch, obsoletes added.

dLobatog:
Thanks @evgeni !

PR #1924, opened 2017-11-04
Add foreman plugins from katello-packaging

ehelms:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---


katello/rubygem-foreman_pipeline/rubygem-foreman_pipeline.spec
```diff
@@ -0,0 +1,114 @@
+# This package contains macros that provide functionality relating to
```
ekohl: How well maintained is this plugin? Last update to the code was more than a year ago.

katello/rubygem-foreman_pipeline/rubygem-foreman_pipeline.spec
```diff
@@ -0,0 +1,114 @@
...
+# Specifically, the "scl" macro must not be defined in official Fedora
+# builds. For more information, see:
+# http://docs.fedoraproject.org/en-US/Fedora_Contributor_Documentation
+# /1/html/Software_Collections_Guide/index.html
+
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
+%global gem_name foreman_pipeline
+%global foreman_dir /usr/share/foreman
```
ekohl: Is this macro even used? I think foreman.spec also provides it in ```rpm/macros.foreman```.

katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
```diff
@@ -0,0 +1,103 @@
...
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/README.md
+%{gem_instdir}/Rakefile
+%{gem_instdir}/test
+
+%posttrans
+%{foreman_db_migrate}
+%{foreman_restart}
+/usr/bin/systemctl restart foreman-tasks
```
ekohl: I think this means there should be a Requires on systemd.

katello/rubygem-foreman_pipeline/rubygem-foreman_pipeline.spec
ehelms: I don't know as the Katello team was not maintaining it but since it requires Katello it had to live in our repositories. Do we have a policy for dropping plugins?

katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
ehelms: That would also be true of line 82, but its just in a macro. All plugins do this today without an explicit requires on systemd. I need to finish the broader discussion on posttrans database actions. I'd suggest we leave this similar to other plugins for now.

ekohl: Fair enough

katello/rubygem-foreman_pipeline/rubygem-foreman_pipeline.spec
ekohl: Not sure about a policy. @xprazak2 could you comment on the status of this plugin?

xprazak2: It is unmaintained since 1.13. I have no objections against droping it.

ehelms:
I removed the pipeline plugin

PR #1925, opened 2017-11-04
Add katello client packages from katello-packaging

ehelms:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---


katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -0,0 +1,376 @@
...
+# crontab
+mkdir -p %{buildroot}%{_sysconfdir}/cron.d/
+cp extra/katello-agent-send.cron %{buildroot}%{_sysconfdir}/cron.d/%{name}
+%endif
+
+%clean
+rm -rf %{buildroot}
+
+%post -n katello-agent
+chkconfig goferd on
```
ekohl: This should be in the EL <= 6 section and ```systemctl enable goferd``` in the systemd section.

katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -0,0 +1,376 @@
...
+%postun -n katello-agent
+touch /tmp/katello-agent-restart
+exit 0
+
+%files -n katello-agent
+%defattr(-,root,root,-)
+%config %{_sysconfdir}/gofer/plugins/katelloplugin.conf
+%{_prefix}/lib/gofer/plugins/katelloplugin.*
+
+%doc LICENSE
```
ekohl: Should this be ```%license``` or will that break on older versions?

katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -0,0 +1,376 @@
+Name: katello-host-tools
```
ekohl: I assume all the code to support older versions like EL5 is still relevant in this case.

ehelms: Yup.

katello/katello-host-tools/katello-host-tools.spec
ehelms: I think it would break RHEL6 and RHEL5.

ehelms:
Updated

PR #1926, opened 2017-11-04
Add katello packages from katello-packaging

ehelms:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---


katello/katello-selinux/katello-selinux.spec
```diff
@@ -0,0 +1,148 @@
...
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+%define selinux_variants targeted
+%define selinux_modules katello
+
+%if 0%{?rhel} == 5
```
ekohl: Is EL5 still relevant because of client tools?

katello/katello/backup.rb
```diff
@@ -0,0 +1,445 @@
+require 'optparse'
```
ekohl: Should we accept these big files now or is this a good time to move them to a repository where there is checking on the ruby code?

katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -0,0 +1,1788 @@
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
+%global gem_name katello
+# %%global prever .rc1
+%global mainver 3.6.0
+%global release 1.nightly
+
+%global foreman_dir /usr/share/foreman
```
ekohl: These are both also in the foreman RPM macros. Do we need to duplicate them?

katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -0,0 +1,1788 @@
...
+BuildRequires: %{?scl_prefix}rubygem(deface) >= 1.0.0
+BuildRequires: %{?scl_prefix}rubygem(deface) < 2.0.0
+BuildRequires: %{?scl_prefix}rubygem(qpid_messaging)
+BuildRequires: %{?scl_prefix_ror}rubygem(rails)
+BuildRequires: %{?scl_prefix_ruby}rubygem(json)
+BuildRequires: %{?scl_prefix_ruby}rubygems-devel
+BuildRequires: %{?scl_prefix_ruby}ruby(rubygems)
+BuildRequires: %{?scl_prefix_ruby}ruby(release)
+
+#old dependencies that are no longer needed, remove in katello 2.6
```
ekohl: Is this still relevant? In foreman-packaging these are typically added to the ```tfm``` package so would that be a better place?

katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -0,0 +1,1788 @@
...
+%{foreman_bundlerd_plugin}
+%{foreman_apipie_cache_foreman}
+%{foreman_apipie_cache_plugin}
+%{foreman_assets_plugin}
+%{gem_instdir}/public/assets/bastion_katello
+
+%exclude %{gem_cache}
+%exclude %{gem_instdir}/lib/katello/repository_types/ostree.rb
+
+%doc %{gem_instdir}/LICENSE.txt
```
ekohl: %license?

katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -0,0 +1,1788 @@
...
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/README.md
+
+%files -n %{katello_ostree}
```
ekohl: If this package is just a single file, why is it split from the main package?

katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -0,0 +1,1788 @@
...
+%{gem_instdir}/lib/katello/repository_types/ostree.rb
+
+%changelog
+* Wed Jul 29 2015 Eric D. Helms <ericdhelms@gmail.com> 2.4.0-1.nightly
```
ekohl: Looks like this wasn't updated in a while

katello/katello-installer/katello-installer-base.spec
```diff
@@ -0,0 +1,644 @@
...
+Group:	   Applications/System
+Requires:  %{name} = %{version}-%{release}
+Requires:  foreman-installer-katello
+
+%description -n foreman-installer-katello-devel
+A set of tools for installation of a Katello development environment using
+Katello and Foreman from git.
+
+%posttrans -n foreman-installer-katello-devel
+mv /etc/katello-devel-installer/katello-devel-installer.yaml.rpmsave /etc/katello-devel-installer/katello-devel-installer.yaml >/dev/null 2>&1 || :
```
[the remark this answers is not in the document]
ehelms: I think this had to do with the file names and the previous content. Granted this was much older and I think in general can be removed due to not upgrading between devels.

katello/katello-repos/katello-repos.spec
```diff
@@ -0,0 +1,199 @@
...
+Summary:  Definition of yum repositories for Katello clients
+Group:    Applications/System
+
+%description -n katello-client-repos
+Defines yum repositories for Katello clients.
+
+%files -n katello-client-repos
+%defattr(-, root, root)
+%config %{_sysconfdir}/yum.repos.d/katello-client.repo
+%if 0%{?rhel} == 6
```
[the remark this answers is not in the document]
ehelms: For our client repositories, yes. Note line 30 above.

katello/katello-selinux/katello-selinux.spec
ehelms: Correct.

katello/katello/backup.rb
ehelms: I would prefer not to go down that path or else this will be blocked for a while. Status quo for the win. It's uglier, I know but as some of us discussed, moving this temporarily to then move it again seems silly when we should just move it to foreman_maintain.

katello/rubygem-katello/rubygem-katello.spec
ehelms: I think they are not relevant. However, I would not put them in `tfm` if they were as it should be up to the plugin that requires additional stack items to deprecate IMO.

katello/rubygem-katello/rubygem-katello.spec
ehelms: Because that single files "turns on" ostree support which is configurable. Until the mainline code is refactored into a better way to handle this, this is how we solved the problem. You could make the argument ostree should not be optional now that we are pure EL7 but that's not entirely up to packaging's per view.

katello/rubygem-katello/rubygem-katello.spec
ehelms: That's cause it builds only for nightlies and release branches. I'm not sure there is value in updating it given that.

ehelms:
Updated with the code changes, will continue to iterate on the open comments.

katello/katello/backup.rb
ekohl: It was sort implied that it should move to foreman_maintain but I don't know how difficult that would be.

katello/rubygem-katello/rubygem-katello.spec
ekohl: It might make sense to import some changes from released branches. I think right now 3.5 will not have the 3.4 changelog either.

katello/rubygem-katello/rubygem-katello.spec
ekohl: So mainline code has no configuration switch for it so it's solved in packaging. Is there an issue tracking this?

ehelms: To be fair, it's a switch in both places. This RPM lays down the configuration. The true answer lies in a repository re-factoring that will be undertaken in the coming months from what I understand.

katello/rubygem-katello/rubygem-katello.spec
ehelms: That's fairly true, at a minimum this could follow the Foreman procedure which just adds a line for each major version bump. Not adding a lot of value, but I suppose then you'd at least see this go up over time. These nightly packages are special to the point of even if we had the changelog from each release branch, it would not truly capture the changes to the spec over time. Perhaps the answer is everytime we modify the spec, we need to bump and "release" it to koji as a tagged and built version.

ekohl: Interesting idea. I do wonder how many people actually look at it vs the git log and with the branching that's going on ... Would it make sense to just drop the changelog?

ehelms: I think its good RPM practice, and on a release branch it adds value for those who know RPM. I'd recommend we just set this every time we branch. I'll update it for this PR.

katello/katello-installer/katello-installer-base.spec
```diff
@@ -0,0 +1,641 @@
...
+Summary: Puppet-based installer for the Katello and Katello Capsule
+Group:   Applications/System
+License: GPLv3+ and ASL 2.0
+URL:     http://katello.org
+Source0: https://fedorapeople.org/groups/katello/releases/source/tarball/katello-installer-%{version}%{?prever}.tar.gz
+
+BuildArch: noarch
+Obsoletes: katello-installer < 2.1.0
+
+Requires: foreman-installer >= 1.11.0
```
ekohl: Should we bump this to 1.17.0 too?

katello/katello-installer/katello-installer-base.spec
```diff
@@ -0,0 +1,641 @@
...
+
+# %%global prever .rc1
+
+Name:    katello-installer-base
+Version: 3.6.0
+Release: 1.nightly%{?prever}%{?dist}
+Summary: Puppet-based installer for the Katello and Katello Capsule
+Group:   Applications/System
+License: GPLv3+ and ASL 2.0
+URL:     http://katello.org
```
ekohl: katello.org is broken right now, github link as well?

katello/katello-repos/katello-repos.spec
```diff
@@ -0,0 +1,199 @@
+Name:           katello-repos
+Version:        3.6.0
+Release:        1.nightly%{?dist}
+Summary:        Definition of yum repositories for Katello
+
+Group:          Applications/Internet
+License:        GPLv2
+URL:            http://www.katello.org
```
ekohl: Also broken right now.

ehelms:
Updated -- I also remove the rubygem-katello changelog. When this is merged, I'll start it fresh given how old and stale it was.

ehelms:
Cleaned up the katello-installer and the katello metapackage.

katello/katello-selinux/katello-selinux.spec
```diff
@@ -0,0 +1,148 @@
...
+URL:            http://www.katello.org
+Source0:        https://codeload.github.com/Katello/%{name}/tar.gz/%{version}
+
+BuildRequires:  checkpolicy, selinux-policy-devel, hardlink
+BuildRequires:  policycoreutils >= %{selinux_policycoreutils_ver}
+BuildRequires:  /usr/bin/pod2man
+BuildArch:      noarch
+
+Requires:           selinux-policy >= %{selinux_policy_ver}
+Requires:           foreman-selinux >= 1.8
```
ekohl: Maybe this should be 1.17.0 too

katello/katello/katello-remove-orphans
```diff
@@ -0,0 +1,4 @@
+#!/bin/env bash
+# Script for intiating removal of orphaned content
```
ekohl: initiating

katello/katello/katello.spec
```diff
@@ -0,0 +1,212 @@
...
+
+Name:       katello
+Version:    3.6.0
+Release:    2.nightly%{?dist}
+Summary:    A package for managing application life-cycle for Linux systems
+BuildArch:  noarch
+
+Group:      Applications/Internet
+License:    GPLv2
+URL:        http://www.katello.org
```
ekohl: Same comment as other katello.org comments

katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -0,0 +1,141 @@
...
+%define katello_ostree %{?scl_prefix}rubygem-%{gem_name}_ostree
+
+Name:    %{?scl_prefix}rubygem-%{gem_name}
+Summary: Katello
+
+Version: %{mainver}
+Release: %{?prever:0.}%{release}%{?prever}%{?dist}
+Group:   Development/Ruby
+License: Distributable
+URL:     http://www.katello.org
```
ekohl: Same here

katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -0,0 +1,141 @@
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
+%global gem_name katello
+# %%global prever .rc1
+%global mainver 3.6.0
+%global release 1.nightly
+
+%global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
```
ekohl: I wonder if this can't inherit from foreman macros.

ehelms: One way to find out.

ehelms:
Will require -- https://github.com/theforeman/foreman-infra/pull/375

evgeni:
[test]

ekohl:
Failure is due to a broken develop. Going to submit patches for that.

```
DEBUG util.py:439:  Getting requirements for tfm-rubygem-katello-3.6.0-.201711070824gitbc3c5d91.nightly.el7.src
DEBUG util.py:439:   --> foreman-assets-1.17.0-0.develop.201710191928git5878c99.el7.noarch
DEBUG util.py:439:   --> foreman-plugin-1.17.0-0.develop.201710191928git5878c99.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-foreman_docker-3.2.1-1.fm1_16.el7.noarch
DEBUG util.py:439:   --> 1:tfm-rubygem-angular-rails-templates-1.0.2-1.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-bastion-6.0.0-1.fm1_17.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-bastion-6.0.0-1.fm1_17.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-foreman-tasks-0.10.0-2.fm1_16.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-foreman-tasks-0.10.0-2.fm1_16.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-gettext_i18n_rails-1.2.1-3.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-apipie-rails-0.5.4-1.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-runcible-2.3.0-1.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-anemone-0.7.2-14.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-deface-1.2.0-1.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-deface-1.2.0-1.el7.noarch
DEBUG util.py:439:   --> tfm-rubygem-qpid_messaging-0.34.1-1.el7.x86_64
DEBUG util.py:439:   --> 1:sclo-ror42-rubygem-rails-4.2.5.1-1.el7.noarch
DEBUG util.py:439:   --> Already installed : rh-ruby22-rubygem-json-1.8.1-16.el7.x86_64
DEBUG util.py:439:   --> rh-ruby22-rubygems-devel-2.4.5-16.el7.noarch
DEBUG util.py:439:   --> Already installed : rh-ruby22-rubygems-2.4.5-16.el7.x86_64
DEBUG util.py:439:   --> Already installed : rh-ruby22-ruby-libs-2.2.2-16.el7.x86_64
DEBUG util.py:439:  Error: Package: foreman-assets-1.17.0-0.develop.201710191928git5878c99.el7.noarch (build)
DEBUG util.py:439:             Requires: npm(babel-core) < 6.8.0
DEBUG util.py:439:             Installing: nodejs-babel-core-6.26.0-1.el7.noarch (build)
DEBUG util.py:439:                 npm(babel-core) = 6.26.0
DEBUG util.py:439:  Error: Package: foreman-assets-1.17.0-0.develop.201710191928git5878c99.el7.noarch (build)
DEBUG util.py:439:             Requires: npm(babel-core) < 6.8.0
DEBUG util.py:439:             Available: nodejs-babel-core-6.26.0-1.el7.noarch (build)
DEBUG util.py:439:                 npm(babel-core) = 6.26.0
```

ehelms:
Yep -- this is updated (along with the packaging PR job) and things are testing properly now. As you pointed out, this won't pass on rubygem-katello due to broken Foreman develop. There are nightly test jobs that handle that. So up to you if you want to hold on merging this till it passes or let nightly handle this.

ehelms:
[test]

ehelms:
@ekohl This now passes!

PR #1934, opened 2017-11-08
Fixes #20550 - support for remote DB in backup and restore

mbacovsky:
Backup and restore scripts were changed to access postgress DB
using the configuration from Foreman and Candlepin and not directly
as DB superuser.

In online backup we skip dump of global objects as usually there are no
privileges to restore them on the remote DB.

In offline backup we do ofline backup as before only the remote DBs
are backed up using pg_dump as in online backup because we do not have
direct access to /var/lib/pgsql/data. We assume the remote DB is always
up.

The restore was updated accordingly and can restore offline backup
including dumps of the remote dbs.

katello-service was fixed to not touch postgresql service when the DB is
remote.

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---

### DIFF
--- a/katello/katello/db_config.rb
+++ b/katello/katello/db_config.rb
@@ -1,0 +1,66 @@
+require 'yaml'
+
+module KatelloUtilities
+  class DBConfig
+    attr_reader :foreman, :candlepin
+    INSTALLER_CONFIG = '/etc/foreman-installer/scenarios.d/last_scenario.yaml'
+
+    def initialize(scenario = INSTALLER_CONFIG)
+      config = YAML.load(File.read(scenario))
+      answers = YAML.load(File.read(config[:answer_file]))
+      @foreman = {
+          :username => answers['foreman']['db_username'],
+          :password => answers['foreman']['db_password'],
+          :database => answers['foreman']['db_database'] || 'foreman',
+          :host => answers['foreman']['db_host'] || 'localhost',
+          :port => answers['foreman']['db_port'] || '5432'
+      }
+      @candlepin = {
+          :username => answers['katello']['candlepin_db_user'],
+          :password => answers['katello']['candlepin_db_password'],
+          :database => answers['katello']['candlepin_db_name'] || 'candlepin',
+          :host => answers['katello']['candlepin_db_host'] || 'localhost',
+          :port => answers['katello']['candlepin_db_port'] || '5432',
+      }
+    end
+
+    def remote_db?(config)
+      !['localhost', '127.0.0.1', `hostname`.strip].include? config[:host]
+    end
+
+    def any_local_db?
+      !remote_db?(foreman) || !remote_db?(candlepin)
+    end
+
+    def any_remote_db?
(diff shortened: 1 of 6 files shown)


mbacovsky:
The original PR in `katello-packaging` was https://github.com/Katello/katello-packaging/pull/499. No changes were made after application of the original patch.

mbacovsky:
Tested the scartch builds initiated by Jenkins [1] along with patch [2] that is needed for this to work. Do not merge before [2] is in.

[1] http://koji.katello.org/koji/taskinfo?taskID=49748
[2] https://github.com/Katello/katello-installer/pull/550

katello/katello/db_config.rb
```diff
@@ -0,0 +1,66 @@
+require 'yaml'
+
+module KatelloUtilities
+  class DBConfig
+    attr_reader :foreman, :candlepin
+    INSTALLER_CONFIG = '/etc/foreman-installer/scenarios.d/last_scenario.yaml'
+
+    def initialize(scenario = INSTALLER_CONFIG)
```
ekohl: This function could use some error handling. What if the scenario file doesn't exist? What if it's on a proxy where foreman or katello is disabled?

katello/katello/db_config.rb
```diff
@@ -0,0 +1,66 @@
...
+      config = YAML.load(File.read(scenario))
+      answers = YAML.load(File.read(config[:answer_file]))
+      @foreman = {
+          :username => answers['foreman']['db_username'],
```
ekohl: We default to foreman

katello/katello/db_config.rb
```diff
@@ -0,0 +1,66 @@
...
+          :password => answers['foreman']['db_password'],
+          :database => answers['foreman']['db_database'] || 'foreman',
+          :host => answers['foreman']['db_host'] || 'localhost',
+          :port => answers['foreman']['db_port'] || '5432'
+      }
+      @candlepin = {
+          :username => answers['katello']['candlepin_db_user'],
```
ekohl: We default to candlepin

katello/katello/katello-service
```diff
@@ -2,6 +2,14 @@
 
 require 'optparse'
 
+# check if production install
```
ekohl: This feels very hacky and fragile. If we're going to use it as a library, I'm strongly leaning to making it a properly installed library.

katello/katello/db_config.rb
```diff
@@ -0,0 +1,66 @@
+require 'yaml'
```
ekohl: Since this so heavily ties into katello-installer, should this be part of the katello-installer repository?

ekohl:
Given we intend to do this in foreman_maintain, should we close this PR?

mbacovsky:
@ekohl I wanted to close this when there is new PR in foreman-maintain so that I can reference it form here. As the implementation in F-M seems to be viable now and as it is possible to add refs later I'm closing this PR now.

PR #1938, opened 2017-11-08
Correct nodejs spec filenames

ekohl:


nodejs-brace/nodejs-brace.spec
```diff
@@ -22,8 +22,8 @@ ExclusiveArch: %{nodejs_arches} noarch
 ExclusiveArch: %{ix86} x86_64 %{arm} noarch
 %endif
 Provides: npm(%{npm_name}) = %{version}
-Provides: bundled-npm(brace) = 0.10.0
-Provides: bundled-npm(w3c-blob) = 0.0.1
+Provides: bundled(npm(brace)) = 0.10.0
```
dLobatog: Why is this change needed here and not in other specs?

ekohl: @xsuchy you added this in your PR. Can you elaborate? Is this something we should be doing in all our specs?

xsuchy: @ekohl  Yes. It should be everywhere. See https://fedoraproject.org/wiki/Packaging:Guidelines#Bundling_and_Duplication_of_system_libraries

ekohl: In that case I'll rewrite this PR to only rename the spec files to remove the bundled and so the bulk refactor in another PR.

PR #1939, opened 2017-11-10
Upgrade to runcible 2.4.0

jlsherrill:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---


ehelms:
[test]

jlsherrill:
They aren't actually needed at run time as far as I know, but the move to
foreman packaging broke this spec because it fails with with 'unpackaged
files'.  These files either have to be deleted during build or included
with some package.  Previously they were in the doc sub package, but the
couple rpms in this repo I checked had them in the main rubygem rpm.

ekohl:
I'd be fine with ignoring them altogether as well.

jlsherrill:
@ekohl looking around it looks like some gems do not include those files and some do.  If a gem does, the packaging usually includes it (and it seems split between the regular rubygem rpm and the docs rpm).  So i think we shouldn't ignore those here if the gem is including them, but i'd be open to dropping them from the rubygem.

ekohl:
@ehelms any objections or shall I merge it?

ehelms:
@jlsherrill now you do!

PR #1947, opened 2017-11-20
gem2rpm: handle nil download path

timogoebel:
If a gem has not been published at rubygems, `download_path` is `nil` and the `sub('https:', 'http:')` fails. This commit defaults the download path to `nil` to avoid the issue.

### DIFF
--- a/gem2rpm/foreman_plugin.spec.erb
+++ b/gem2rpm/foreman_plugin.spec.erb
@@ -34,7 +34,7 @@ License: <%= spec.licenses.join(" and ") %>
 <% if spec.homepage -%>
 URL: <%= spec.homepage %>
 <% end -%>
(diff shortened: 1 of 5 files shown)


ekohl:
I wonder why we even need the https -> http substitution. Especially if we in foreman-packaging only use git-annex.

ekohl:
@dLobatog but which tools are unable to handle https?

timogoebel:
@ekohl: Are you proposing to change the templates to always use the `https` url?

ekohl:
@timogoebel I'm wondering if we can just drop the ```.sub('https', 'http')``` part of it.

timogoebel:
@ekohl: Sure, like this? :-)

ekohl:
Going to merge this now and revisit if it turns out this is a problem. Thanks!

PR #1952, opened 2017-11-21
Unbundle nodejs-prop-types

ekohl:
All three dependencies are available in our repository with correct versions. There's no need to bundle this.


ekohl:
Indeed, smaller RPM and less duplication.

PR #1953, opened 2017-11-22
Refs #21736 - Bump mail gem to 2.7[rpm]

tbrisker:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -115,7 +115,7 @@ Requires: %{?scl_prefix}rubygem(deacon) >= 1.0
 Requires: %{?scl_prefix}rubygem(deacon) < 2.0
 Requires: %{?scl_prefix}rubygem(webpack-rails) >= 0.9.7
 Requires: %{?scl_prefix}rubygem(webpack-rails) < 1.0.0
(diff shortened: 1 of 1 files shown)


ekohl:
This is just the dependency rather than the actual package bump

tbrisker:
hmph, looks like we were using the scl version of the gem, I guess we will now need to build our own?

ekohl:
Related core PR: https://github.com/theforeman/foreman/pull/5023

And we indeed need to import it into our packaging.

tbrisker:
@ekohl since this is only targeted at 1.17, the correct version of the mail gem should already be in tfm-ror51 - https://github.com/theforeman/tfm-ror51-packaging/blob/master/packages/tfm-ror51-rubygem-mail/tfm-ror51-rubygem-mail.spec
What should we do if any in the foreman-packaging?

ekohl:
I think it's fine to wait for bumping the SCL to tfm-ror51

tbrisker:
Makes sense, we'll get the new gem with the new scl anyways, no need for a separate pr for it.

PR #1954, opened 2017-11-23
Deploy dynflowd [deb]

dLobatog:
WIP

TODO:

- [x] Deploy dynflowd to /usr/sbin
- [x] Deploy systemd dynflowd.service in jessie, stretch, xenial to lib/systemd/system
- [x] Deploy systemd dynflowd.default in jessie, stretch, xenial to /etc/default/dynflowd
- [x] Deploy init.d script 
- [x] Remove foreman-tasks from plugins/ruby-foreman-tasks

### DIFF
--- a/debian/jessie/foreman/dynflowd.service
+++ b/debian/jessie/foreman/dynflowd.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Foreman jobs daemon
+Documentation=https://theforeman.org
+After=network.target remote-fs.target nss-lookup.target
+
+[Service]
+Type=forking
+User=foreman
+TimeoutSec=90
+WorkingDirectory=/usr/share/foreman
+ExecStart=/usr/sbin/dynflowd start
+ExecStop=/usr/sbin/dynflowd stop
+EnvironmentFile=-/etc/default/dynflowd
+
+[Install]
+WantedBy=multi-user.target
+Alias=ruby-foreman-tasks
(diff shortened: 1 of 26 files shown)


dLobatog:
@mmoll Mind to take a look?

mmoll:
the *.service thing works for all versions

mmoll:
/etc/sysconfig doesn't exist on Debian based distros by default, but we can just drop our file there, if needed. To my understanding, /etc/default would be the place by Debian conventions (but if it makes things harder for us, just ignore this).

debian/jessie/foreman/rules
```diff
@@ -26,3 +26,6 @@ build:
 	/bin/sed -ri '1sXenv rubyXenv foreman-rubyX' bin/*
 	/bin/sed -ri 's~^BUNDLER_CMD=""~BUNDLER_CMD="/usr/bin/foreman-ruby /usr/bin/bundle exec"~' script/foreman-rake
 	/bin/mkdir -p .ssh
+	# Ensure backwards compatiblity by having dynflowd aliased to foreman-tasks
+	/bin/mv /usr/sbin/dynflow-executor /usr/sbin/dynflowd
+	/bin/cp /usr/sbin/dynflowd /usr/sbin/foreman-tasks
```
ekohl: Wouldn't a symlink be cleaner?

debian/jessie/foreman/dynflowd.service
```diff
@@ -0,0 +1,17 @@
...
+User=foreman
+TimeoutSec=90
+WorkingDirectory=/usr/share/foreman
+ExecStart=/usr/sbin/dynflowd start
+ExecStop=/usr/sbin/dynflowd stop
+EnvironmentFile=-/etc/sysconfig/dynflowd
+
+[Install]
+WantedBy=multi-user.target
+Alias=foreman-tasks
```
ekohl: Wasn't this called ruby-foreman-tasks on Debian?

ekohl:
I wonder if we still need to ship the sysvinit files for distros that have sytemd (all but trusty). Other than that the build is failing and I'm not sure why.

dLobatog:
It's failing because http://downloads.theforeman.org/foreman/foreman-1.17.0.tar.bz2 isn't available but the debs are trying to download it before building it

PR #1957, opened 2017-11-23
Deploy dynflowd [deb]

ekohl:
Replaces https://github.com/theforeman/foreman-packaging/pull/1954 because this branch name should correctly build.

### DIFF
--- a/debian/jessie/foreman/rules
+++ b/debian/jessie/foreman/rules
@@ -26,9 +26,8 @@ build:
 	/bin/sed -ri '1sXenv rubyXenv foreman-rubyX' bin/*
 	/bin/sed -ri 's~^BUNDLER_CMD=""~BUNDLER_CMD="/usr/bin/foreman-ruby /usr/bin/bundle exec"~' script/foreman-rake
 	/bin/mkdir -p .ssh
-	# Ensure backwards compatiblity by having dynflowd aliased to foreman-tasks
-	/bin/mv /usr/sbin/dynflow-executor /usr/sbin/dynflowd
-	/bin/ln -s /usr/sbin/dynflowd /usr/sbin/ruby-foreman-tasks
+	# https://projects.theforeman.org/issues/21757
+	/bin/mv bin/dynflow-executor script/dynflowd
 
 %:
 	dh $@ --with=systemd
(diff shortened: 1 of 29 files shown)


ekohl:
[test]

debian/jessie/foreman/rules
```diff
@@ -26,3 +26,9 @@ build:
 	/bin/sed -ri '1sXenv rubyXenv foreman-rubyX' bin/*
 	/bin/sed -ri 's~^BUNDLER_CMD=""~BUNDLER_CMD="/usr/bin/foreman-ruby /usr/bin/bundle exec"~' script/foreman-rake
 	/bin/mkdir -p .ssh
+	# Ensure backwards compatiblity by having dynflowd aliased to foreman-tasks
+	/bin/mv /usr/sbin/dynflow-executor /usr/sbin/dynflowd
```
mmoll: I guess this should be handled in the rules file?

debian/jessie/foreman/rules
```diff
@@ -26,3 +26,9 @@ build:
...
+	/bin/ln -s /usr/sbin/dynflowd /usr/sbin/ruby-foreman-tasks
```
mmoll: there's a links file, see foreman-proxy

mmoll:
There won't be a trusty release fo 1.16, so these files can just get ommited.

ekohl:
> There won't be a trusty release fo 1.16, so these files can just get ommited.

And Jessie has been dropped from nightly too. Should they still be included?

debian/jessie/foreman/rules
ekohl: It also breaks the build:

```
17:16:10 /bin/mv /usr/sbin/dynflow-executor /usr/sbin/dynflowd
17:16:10 /bin/mv: cannot stat '/usr/sbin/dynflow-executor': No such file or directory
```

mmoll:
> And Jessie has been dropped from nightly too. Should they still be included?

Yes, because this is making cherry-picking to 1.16 easy. ;)

ekohl:
And should we still ship the ```dynflowd.init``` file or can we rely on systemd now?

mmoll:
> And should we still ship the dynflowd.init file or can we rely on systemd now?

This is kind of disputed, I guess... Ppersonally I'd just 🚢 only systemd files, but at least jessie can also be ran without systemd. If it's cheap for us to just :shipit: the sysvinit files, why not?

ekohl:
Fair enough.

mmoll:
I think as the foreman package is cdbs based, the dh-stuff can't be used (but didn't check).

PR #1959, opened 2017-11-24
Deploy dynflowd [deb]

dLobatog:
Replaces https://github.com/theforeman/foreman-packaging/pull/1954 and https://github.com/theforeman/foreman-packaging/pull/1957 - I opened it myself in case I need to update the branch.

Note - only review the `stretch` files, I will update the rest if these build.


dLobatog:
According to https://wiki.debian.org/Teams/pkg-systemd/Packaging (cdbs > 0.4.123) should be enough, we use 0.4.130.

debian/stretch/foreman/rules
```diff
@@ -25,3 +25,9 @@ build:
 	/bin/sed -ri '1sXenv rubyXenv foreman-rubyX' bin/*
 	/bin/sed -ri 's~^BUNDLER_CMD=""~BUNDLER_CMD="/usr/bin/foreman-ruby /usr/bin/bundle exec"~' script/foreman-rake
 	/bin/mkdir -p .ssh
+	# Ensure backwards compatiblity by having dynflowd aliased to foreman-tasks
+	/bin/cp bin/dynflow-executor /usr/sbin/dynflowd
+	/bin/ln -s /usr/sbin/dynflowd /usr/sbin/ruby-foreman-tasks
+
+%:
+	dh --with=systemd
```
dLobatog: This or `dh $@ --with=systemd` are failing with:

```
07:44:13 dh --with=systemd
07:44:13 dh: Compatibility levels before 9 are deprecated (level 7 in use)
07:44:13 dh: specify a sequence to run
07:44:13 debian/rules:33: recipe for target 'common-install-prehook-indep' failed
```

Any idea @mmoll ? @GregSutcliffe ?

debian/jessie/foreman/rules
```diff
@@ -26,3 +26,9 @@ build:
 	/bin/sed -ri '1sXenv rubyXenv foreman-rubyX' bin/*
 	/bin/sed -ri 's~^BUNDLER_CMD=""~BUNDLER_CMD="/usr/bin/foreman-ruby /usr/bin/bundle exec"~' script/foreman-rake
 	/bin/mkdir -p .ssh
+	# Ensure backwards compatiblity by having dynflowd aliased to foreman-tasks
+	/bin/mv /usr/sbin/dynflow-executor /usr/sbin/dynflowd
```
ekohl: See my fixes in https://github.com/theforeman/foreman-packaging/pull/1957. This is the build step, not the install step. I fixed it with ```mv bin/dynflow-executor script/dynflowd``` and set the right properties in foreman.install. The symlink is solved via the links file.

dLobatog: I realize that, but that's not the reason this PR is failing. I tried to make some changes in `stretch` (https://github.com/theforeman/foreman-packaging/pull/1959#discussion_r152921779) and when stretch builds fine, I will propagate the changes to the rest.

dLobatog:
I find few rederences to the error we're getting - one of them is [here](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=808969)  so I did the same, comment the line for cdbs

dLobatog:
http://ci.theforeman.org/job/packaging_build_deb_coreproject/4908/arch=x86,label=debian,os=stretch/console

let's see - it looks like it makes a difference and allows `dh --with=systemd` to continue. i'll update the rest of the versions removing cdbs too

dLobatog:
http://ci.theforeman.org/job/packaging_build_deb_coreproject/4909 was able to build (Foreman itself), -tasks wasn't, might be because the version remained the same.

mmoll:
I think there are some files missing from the built package:
```
# dpkg -L foreman|grep dynflow
/usr/share/foreman/vendor/cache/dynflow-0.8.32.gem
/usr/share/foreman/script/dynflowd
/usr/share/foreman/lib/foreman/dynflow
/usr/share/foreman/lib/foreman/dynflow/configuration.rb
/usr/share/foreman/lib/foreman/dynflow.rb
/usr/share/foreman/lib/tasks/dynflow.rake
/var/lib/foreman/db/dynflow-production.sqlite3
/usr/sbin/dynflowd
```
I think the whole *.init and *,default works only automatically for files that are named like a package, and as there's no "dynflowd" package, these files never get deployed. So I guess it would be better to just add the files to the foreman package in its the .install file.

mmoll:
Some more things are missing:
```
# ls -l /usr/bin/foreman-ruby
ls: cannot access '/usr/bin/foreman-ruby': No such file or directory
```
I think mixing cdbs packages with the new dh stuff isn't  a good idea...

dLobatog:
Missing things currently:

- [ ] /etc/default/dynflowd
- [ ] `systemctl status dynflowd` `systemctl status dynflowd`  does not work
- [ ] `foreman-ruby` isn't in /usr/bin/foreman-ruby

(feel free to update my comment with anything else missing)

@mmoll How would you suggest to add the missing files, just using `.install` & not the `--with=systemd` stuff?

dLobatog:
The package was able to build http://ci.theforeman.org/job/packaging_build_deb_coreproject/4926/ , the failure on ARM v7 was a problem in the Jenkins slave.

mmoll:
the files still have thier suffix in the package:
/etc/init.d/dynflowd.init
/lib/systemd/system/dynflowd.service
otherwise it seems good 👍

dLobatog:
@mmoll Oh you're right - the `init.d` one contains the suffix, that's a mistake. The `systemd` one though, that's how it's supposed to be, if you `ls /lib/systemd/system` you'll see other services end with `.service`

debian/stretch/foreman/control
```diff
@@ -6,7 +6,7 @@ Standards-Version: 3.9.1
 Build-Depends: debhelper (>= 7), cdbs, bundler (>= 1.3), ruby2.3-dev, git | git-core,
```
evgeni: You are bumping compat to 9, but not bumping the debhelper dependency to 9.

debian/stretch/foreman/control
```diff
@@ -6,7 +6,7 @@ Standards-Version: 3.9.1
...
                zlib1g-dev, default-libmysqlclient-dev, libpq-dev, pkg-config, libvirt-dev,
                libsqlite3-dev, build-essential, curl, nodejs (>= 8.9),
-               python
+               python, dh-systemd (>= 1.5)
```
evgeni: `dh_systemd` is part of `debhelper` since 9.20160709 (stretch has 10.2.5), so you can safely avoid that dependency if you bump the `debhelper` one accodingly.

debian/stretch/foreman/dynflowd.init
```diff
@@ -0,0 +1,267 @@
...
+
+RETVAL=0
+FOREMAN_USER=${FOREMAN_USER:-foreman}
+FOREMAN_HOME=${FOREMAN_HOME:-/usr/share/foreman}
+FOREMAN_DATA_DIR=${FOREMAN_DATA_DIR:-/var/lib/foreman}
+FOREMAN_ENV=${FOREMAN_ENV:-production}
+FOREMAN_TASK_PARAMS=${FOREMAN_TASK_PARAMS:--p foreman}
+FOREMAN_PID_DIR=$FOREMAN_HOME/tmp/pids
+FOREMAN_LOG_DIR=${FOREMAN_LOG_DIR:-/var/log/foreman}
+DAEMON="/usr/sbin/dynflowd"
```
evgeni: line 18 already sets `DAEMON` to the almost right value

debian/stretch/foreman/dynflowd.init
```diff
@@ -0,0 +1,267 @@
+#!/bin/bash
+#! /bin/sh
```
evgeni: bash is the one you want here, right?

debian/stretch/foreman/foreman.install
```diff
@@ -5,6 +5,7 @@ config/settings.yaml                     etc/foreman
 config/database.yml                      etc/foreman
 script/foreman-rake                      usr/sbin
 script/foreman-tail                      usr/sbin
+script/dynflowd                          usr/sbin
```
evgeni: why sbin? we're running the daemon as `foreman` user, so it's clearly not root only :)

debian/stretch/foreman/foreman.install
```diff
@@ -26,3 +27,6 @@ bundler.d/facter.rb                      usr/share/foreman/bundler.d
 bundler.d/jsonp.rb                       usr/share/foreman/bundler.d
 public                                   var/lib/foreman
 db                                       var/lib/foreman
+debian/dynflowd.default                  etc/default
```
evgeni: this should be handled by `dh_installinit`

debian/stretch/foreman/foreman.install
```diff
@@ -26,3 +27,6 @@ bundler.d/facter.rb                      usr/share/foreman/bundler.d
...
+debian/dynflowd.init                     etc/init.d
```
evgeni: this should be handled by `dh_installinit`

debian/stretch/foreman/foreman.install
```diff
@@ -26,3 +27,6 @@ bundler.d/facter.rb                      usr/share/foreman/bundler.d
...
+debian/dynflowd.service                  lib/systemd/system
```
evgeni: this should be handled by `dh_installinit`

debian/stretch/foreman/links
```diff
@@ -8,3 +8,5 @@ var/lib/foreman/db		usr/share/foreman/db
 var/log/foreman			usr/share/foreman/log
 var/lib/foreman/public		usr/share/foreman/public
 usr/bin/ruby2.3		usr/bin/foreman-ruby
+usr/sbin/dynflowd usr/share/foreman/script/dynflowd
+usr/sbin/ruby-foreman-tasks usr/share/foreman/script/dynflowd
```
evgeni: this should be a different script, I guess?

debian/stretch/foreman/rules
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 # -*- makefile -*-
 
-include /usr/share/cdbs/1/rules/debhelper.mk
+# include /usr/share/cdbs/1/rules/debhelper.mk
```
evgeni: why?

dLobatog:
Updated to rename the .init and .default files

debian/stretch/foreman/rules
dLobatog: This helper conflicts with `dh` , as I commented above, it's not a problem to not use it if we use `dh`, `bitcoin` did the same thing https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=808969

debian/stretch/foreman/links
dLobatog: Why a different script? It's supposed to do the same

debian/stretch/foreman/foreman.install
dLobatog: Sorry, not an expert, could you explain how? Overriding it in `rules`?

debian/stretch/foreman/dynflowd.init
dLobatog: I think so I took the one from tasks & made nearly no changes https://github.com/theforeman/foreman-packaging/blob/deb/develop/plugins/ruby-foreman-tasks/debian/init.d

debian/stretch/foreman/control
evgeni: Jessie only has 9.20150101, so we'd need jessie-backports in our buildroots for that to work. But if that's the case, :+1: on doing `debhelper (>= 9.20160709)`

debian/stretch/foreman/links
evgeni: if I am reading the two lines correctly, they should
1. create a link `usr/share/foreman/script/dynflowd` pointing to `usr/sbin/dynflowd`
2. create a link `usr/share/foreman/script/dynflowd` pointing to `usr/sbin/ruby-foreman-tasks`

and that clashes, as you end up with the same link name

debian/stretch/foreman/foreman.postinst
```diff
@@ -101,6 +101,10 @@ fi
 # Own all the core files
 chown -Rf foreman:foreman '/usr/share/foreman'
 
+# During dh_install we cannot rename files, so we do so here
```
evgeni: nope, please never rename files in postinst, esp not the ones in /etc as those are handled as conffiles

debian/stretch/foreman/rules
evgeni: right, so we don't use cdbs at all anymore, correct? let's nuke it from build-depends then too?

debian/stretch/foreman/dynflowd.init
evgeni: k, just found both lines confusing :)

debian/stretch/foreman/links
ekohl: I think you're reading it the other way around: it's ```<target> <file>``` since I doubt we're making ```/usr/bin/ruby2.3``` point to ```/usr/bin/foreman-ruby```.

debian/stretch/foreman/foreman.install
evgeni: a bit, yeah. see `dh_installinit(1)`:

       --name=name
           Install the init script (and default file) as well as upstart job file using the filename name instead of the default filename, which is the package name. When this parameter is used, dh_installinit looks for and installs
           files named debian/package.name.init, debian/package.name.default and debian/package.name.upstart instead of the usual debian/package.init, debian/package.default and debian/package.upstart.

so calling these files `foreman.dynflow.*` and adding the following to `debian/rules` *should* work:

    override_dh_installinit:
        dh_installinit --name=dynflow

(that is, if we only ship one init script)

debian/stretch/foreman/links
evgeni: That's what I said, it's `<target> <linkname>`, and we define the same linkname for two targets. :)

debian/stretch/foreman/foreman.install
dLobatog: I agree but since other tools (foreman-rake, foreman-ruby, etc) were deployed there I kept the pattern rather than changing it all over. I'm happy to change it all if you 👍 it.

debian/stretch/foreman/foreman.install
dLobatog: Is there an equivalent thing for systemd? This is just for init.d/default

debian/stretch/foreman/links
ekohl: I think you're right that these should be the other way around:
```
usr/share/foreman/script/dynflowd usr/sbin/dynflowd
usr/share/foreman/script/dynflowd usr/sbin/ruby-foreman-tasks
```

evgeni:
`dh_installinit` will install the `dynflowd.service` just fine in compat levels <= 10, which we use. in compat 11 there is `dh_installsystemd`, but that we don't have to care about just yet (as it's yet at experimental level, see changelog of `debhelper` 10.9.2).

and indeed, running `dpkg-deb -c foreman.deb`, I see `./lib/systemd/system/dynflowd.service` so we're good on that topic for now.

debian/stretch/foreman/foreman.dynflowd.init
```diff
@@ -0,0 +1,265 @@
...
+}
+
+case "$1" in
+    start)
+        start
+        ;;
+    stop)
+        stop
+        ;;
+    restart)
```
evgeni: While probably not the topic of this PR, but Debian requires a `force-reload` target, that reloads the config *or* falls back to `restart`: https://www.debian.org/doc/debian-policy/#writing-the-scripts

evgeni:
I've also added this to d/rules:

```
override_dh_systemd_enable:
        dh_systemd_enable
        dh_systemd_enable --name=dynflowd
```

debian/stretch/foreman/control
```diff
@@ -17,7 +17,7 @@ Depends: ruby2.3, bundler (>= 1.3), zlib1g-dev,
          ruby2.3-dev, build-essential,
          rake (>=0.8.4), ${misc:Depends}
 Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal
-Conflicts: ruby-activerecord-deprecated-finders
+Conflicts: ruby-activerecord-deprecated-finders, ruby-foreman-tasks (< 0.10.1)
```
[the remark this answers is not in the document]
dLobatog: It should be `<<`, [0.10.1](https://github.com/theforeman/foreman-tasks/commits/v0.10.1) removes is the first version without the `foreman-tasks` service, we should not ship this foreman package with that foreman-tasks version as we would be shipping 2 equivalent daemons

dLobatog:
@evgeni Just did it, here are the outcomes:

```

root@essie-joaquin:~# systemctl status dynflowd.service
● dynflowd.service - Foreman jobs daemon
   Loaded: loaded (/lib/systemd/system/dynflowd.service; enabled; vendor preset: enabled)
   Active: active (running) since Thu 2017-12-07 12:15:08 UTC; 8min ago
     Docs: https://theforeman.org
 Main PID: 22587 (ruby)
    Tasks: 2 (limit: 4915)
   CGroup: /system.slice/dynflowd.service
           └─22587 dynflow_executor_monitor

Dec 07 12:15:04 essie-joaquin systemd[1]: Starting Foreman jobs daemon...
Dec 07 12:15:06 essie-joaquin dynflowd[22573]: DEPRECATION WARNING: `config.serve_static_files` is deprecated and will be removed in Rai
Dec 07 12:15:06 essie-joaquin dynflowd[22573]: Please use `config.public_file_server.enabled` instead.
Dec 07 12:15:06 essie-joaquin dynflowd[22573]:  (called from <class:Application> at /usr/share/foreman/config/application.rb:218)
Dec 07 12:15:06 essie-joaquin dynflowd[22573]: Dynflow Executor: start in progress
Dec 07 12:15:07 essie-joaquin dynflowd[22573]: dynflow_executor: process with pid 22583 started.
Dec 07 12:15:08 essie-joaquin systemd[1]: Started Foreman jobs daemon.
root@essie-joaquin:~# sudo systemctl restart dynflowd
root@essie-joaquin:~# systemctl status dynflowd.service -l
● dynflowd.service - Foreman jobs daemon
   Loaded: loaded (/lib/systemd/system/dynflowd.service; enabled; vendor preset: enabled)
   Active: active (running) since Thu 2017-12-07 12:23:45 UTC; 14s ago
     Docs: https://theforeman.org
  Process: 22868 ExecStop=/usr/sbin/dynflowd stop (code=exited, status=0/SUCCESS)
  Process: 22871 ExecStart=/usr/sbin/dynflowd start (code=exited, status=0/SUCCESS)
 Main PID: 22587 (code=exited, status=1/FAILURE)
    Tasks: 2 (limit: 4915)
   CGroup: /system.slice/dynflowd.service
           └─22887 dynflow_executor_monitor

Dec 07 12:23:42 essie-joaquin systemd[1]: Stopped Foreman jobs daemon.
Dec 07 12:23:42 essie-joaquin systemd[1]: dynflowd.service: Unit entered failed state.
Dec 07 12:23:42 essie-joaquin systemd[1]: dynflowd.service: Failed with result 'exit-code'.
Dec 07 12:23:42 essie-joaquin systemd[1]: Starting Foreman jobs daemon...
Dec 07 12:23:44 essie-joaquin dynflowd[22871]: DEPRECATION WARNING: `config.serve_static_files` is deprecated and will be removed in Rai
Dec 07 12:23:44 essie-joaquin dynflowd[22871]: Please use `config.public_file_server.enabled` instead.
Dec 07 12:23:44 essie-joaquin dynflowd[22871]:  (called from <class:Application> at /usr/share/foreman/config/application.rb:218)
Dec 07 12:23:44 essie-joaquin dynflowd[22871]: Dynflow Executor: start in progress
Dec 07 12:23:45 essie-joaquin dynflowd[22871]: dynflow_executor: process with pid 22883 started.
Dec 07 12:23:45 essie-joaquin systemd[1]: Started Foreman jobs daemon.
root@essie-joaquin:~# which foreman-ruby
/usr/bin/foreman-ruby
root@essie-joaquin:~# foreman-
foreman-debug  foreman-rake   foreman-ruby   foreman-tail
root@essie-joaquin:~# ls /lib/systemd/system
system/                       systemd-hostnamed             systemd-remount-fs            systemd-timesyncd
systemd                       systemd-initctl               systemd-reply-password        systemd-udevd
systemd-ac-power              systemd-journald              systemd-resolved              systemd-update-utmp
systemd-backlight             systemd-localed               systemd-rfkill                systemd-user-sessions
systemd-binfmt                systemd-logind                systemd-shutdown              system-generators/
systemd-cgroups-agent         systemd-modules-load          systemd-sleep                 system-preset/
systemd-cryptsetup            systemd-networkd              systemd-socket-proxyd         system-shutdown/
systemd-fsck                  systemd-networkd-wait-online  systemd-sysctl                system-sleep/
systemd-fsckd                 systemd-quotacheck            systemd-sysv-install
systemd-hibernate-resume      systemd-random-seed           systemd-timedated
root@essie-joaquin:~# ls /lib/systemd/system/dynflowd.service
/lib/systemd/system/dynflowd.service
root@essie-joaquin:~# systemctl status ruby-foreman-tasks.service
● dynflowd.service - Foreman jobs daemon
   Loaded: loaded (/lib/systemd/system/dynflowd.service; enabled; vendor preset: enabled)
   Active: active (running) since Thu 2017-12-07 12:23:45 UTC; 1min 34s ago
     Docs: https://theforeman.org
  Process: 22868 ExecStop=/usr/sbin/dynflowd stop (code=exited, status=0/SUCCESS)
  Process: 22871 ExecStart=/usr/sbin/dynflowd start (code=exited, status=0/SUCCESS)
 Main PID: 22587 (code=exited, status=1/FAILURE)
    Tasks: 2 (limit: 4915)
   CGroup: /system.slice/dynflowd.service
           └─22887 dynflow_executor_monitor

Dec 07 12:23:42 essie-joaquin systemd[1]: Stopped Foreman jobs daemon.
Dec 07 12:23:42 essie-joaquin systemd[1]: dynflowd.service: Unit entered failed state.
Dec 07 12:23:42 essie-joaquin systemd[1]: dynflowd.service: Failed with result 'exit-code'.
Dec 07 12:23:42 essie-joaquin systemd[1]: Starting Foreman jobs daemon...
Dec 07 12:23:44 essie-joaquin dynflowd[22871]: DEPRECATION WARNING: `config.serve_static_files` is deprecated and will be removed in Rai
Dec 07 12:23:44 essie-joaquin dynflowd[22871]: Please use `config.public_file_server.enabled` instead.
Dec 07 12:23:44 essie-joaquin dynflowd[22871]:  (called from <class:Application> at /usr/share/foreman/config/application.rb:218)
Dec 07 12:23:44 essie-joaquin dynflowd[22871]: Dynflow Executor: start in progress
Dec 07 12:23:45 essie-joaquin dynflowd[22871]: dynflow_executor: process with pid 22883 started.
Dec 07 12:23:45 essie-joaquin systemd[1]: Started Foreman jobs daemon.

root@essie-joaquin:~# /etc/init.d/dynflowd
Usage: {start|stop|restart|condrestart|condstop|status}
root@essie-joaquin:~# /etc/init.d/dynflowd start
Starting dynflowd (via systemctl): dynflowd.service.
root@essie-joaquin:~# ls -l /etc/init.d/
total 76
-rwxr-xr-x 1 root root 1232 Apr  7  2017 console-setup.sh
-rwxr-xr-x 1 root root 3049 May  3  2015 cron
-rwxr-xr-x 1 root root 2813 Jul 30 10:23 dbus
-rwxr-xr-x 1 root root 6687 Dec  7 08:28 dynflowd
-rwxr-xr-x 1 root root 3107 Dec  7 08:28 foreman
-rwxr-xr-x 1 root root 3809 Mar 22  2017 hwclock.sh
-rwxr-xr-x 1 root root 1479 May 19  2016 keyboard-setup.sh
-rwxr-xr-x 1 root root 2044 Dec 26  2016 kmod
-rwxr-xr-x 1 root root 4597 Sep 16  2016 networking
-rwxr-xr-x 1 root root 1191 Nov 22  2016 procps
-rwxr-xr-x 1 root root 4355 Jul 10  2014 rsync
-rwxr-xr-x 1 root root 2868 Jan 18  2017 rsyslog
-rwxr-xr-x 1 root root 4033 Jun 18 00:08 ssh
-rwxr-xr-x 1 root root  731 Jun  5  2017 sudo
-rwxr-xr-x 1 root root 6087 Jul  5 20:31 udev

root@essie-joaquin:~# ls -l /etc/default/dynflowd
-rw-r--r-- 1 root root 778 Dec  7 08:28 /etc/default/dynflowd

```

mmoll:
debhelper 9.20160709 is too new for Ubuntu/xenial (16.04)...

dLobatog:
@evgeni @ekohl Looks like xenial is now passing too, anything else missing? http://ci.theforeman.org/job/packaging_build_deb_coreproject/4971

mmoll:
tested on xenial and it seems to work

debian/jessie/foreman/control
```diff
@@ -3,10 +3,10 @@ Maintainer: Greg Sutcliffe <greg.sutcliffe@gmail.com>
 Section: web
 Priority: extra
 Standards-Version: 3.9.1
-Build-Depends: debhelper (>= 7), cdbs, bundler (>= 1.3), ruby2.1-dev, git | git-core,
+Build-Depends: debhelper (>= 9.20160709), bundler (>= 1.3), ruby2.1-dev, git | git-core,
```
evgeni: `debhelper` in jessie is `9.20150101+deb8u2` so this will not work, unless we also have `jessie-backports` enabled during build-time.

we need either: `debhelper (>= 9), dh-systemd (>= 1.5)` *or* `debhelper (>= 9.20160709)`. the jessie builds on jenkins were skipped, so I can't check if the current version did build correctly

debian/jessie/foreman/rules
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 # -*- makefile -*-
 
-include /usr/share/cdbs/1/rules/debhelper.mk
+#include /usr/share/cdbs/1/rules/debhelper.mk
```
evgeni: let's just nuke that line, shall we?

debian/stretch/foreman/foreman.install
evgeni: `foreman-ruby` is in `bin` and so was `foreman-tasks` ;)

`f-rake` and `f-tail` are in sbin, as those really are `root` tools

debian/trusty/foreman/control
```diff
@@ -3,10 +3,10 @@ Maintainer: Greg Sutcliffe <greg.sutcliffe@gmail.com>
 Section: web
 Priority: extra
 Standards-Version: 3.9.1
-Build-Depends: debhelper (>= 7), cdbs, bundler (>= 1.3.5-2ubuntu2), ruby2.0, ruby2.0-dev, git | git-core,
+Build-Depends: debhelper (>= 9.20160709), bundler (>= 1.3.5-2ubuntu2), ruby2.0, ruby2.0-dev, git | git-core,
```
evgeni: Trusty only has `9.20131227ubuntu1`, and I have no idea about trusty backports

debian/xenial/foreman/control
```diff
@@ -3,10 +3,10 @@ Maintainer: Greg Sutcliffe <greg.sutcliffe@gmail.com>
 Section: web
 Priority: extra
 Standards-Version: 3.9.1
-Build-Depends: debhelper (>= 7), cdbs, bundler (>= 1.3), ruby2.3-dev, git | git-core,
+Build-Depends: debhelper (>= 9.20160115ubuntu3), bundler (>= 1.3), ruby2.3-dev, git | git-core,
```
evgeni: `>= 9` should be sufficient as long we require `dh-systemd`

debian/xenial/foreman/rules
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 # -*- makefile -*-
 
-include /usr/share/cdbs/1/rules/debhelper.mk
+#include /usr/share/cdbs/1/rules/debhelper.mk
```
evgeni: nuke :)

debian/trusty/foreman/rules
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 # -*- makefile -*-
 
-include /usr/share/cdbs/1/rules/debhelper.mk
+#include /usr/share/cdbs/1/rules/debhelper.mk
```
evgeni: nuke :)

debian/trusty/foreman/control
mmoll: Foreman 1.15 is the last version to support trusty, so this can be ommited alltogether.

mmoll:
JFTR, I did keep the trusty and jessie files around in deb/develop to ease cherry-picks. As there will probably be no more 1.15 releases(?) for jessie these could get deleted in another commit. I'd prefer to keep the jessie files still around as otherwise any cherry-pick to deb/1.16 will need manual work to apply the changes to jessie by the cherry-picker.

dLobatog:
Updated once again, thanks so much for the review @mmoll @evgeni !

debian/stretch/foreman/control
```diff
@@ -3,7 +3,7 @@ Maintainer: Greg Sutcliffe <greg.sutcliffe@gmail.com>
 Section: web
 Priority: extra
 Standards-Version: 3.9.1
-Build-Depends: debhelper (>= 7), cdbs, bundler (>= 1.3), ruby2.3-dev, git | git-core,
+Build-Depends: debhelper (>= 9), bundler (>= 1.3), ruby2.3-dev, git | git-core,
```
evgeni: Now you edited too much ;)

On stretch we need `debhelper (>= 9.20160709)` as we don't need `dh-systemd` anymore ;)

evgeni:
@mmoll right, so these changes are 1.17+ only, which is not supported on jessie and trusty anyways. I wonder if we should touch the jessie/trusty packaging at all then or just let it "rot" until after 1.17/1.18 is released and then remove it for good?

@dLobatog @ekohl what do you think?

ekohl:
I agree we're very unlikely to do any more 1.15 releases so I think we can safely remove the foreman trusty packages. Keeping around Jessie for 1.16 cherry picks might be useful but since we're not releasing 1.17 there, I'd be fine with not changing those.

debian/stretch/foreman/control
dLobatog: oops!

dLobatog:
I killed the job because there are OOMs when compiling webpack.

http://ci.theforeman.org/job/packaging_build_deb_coreproject/4977/arch=x86,label=debian,os=stretch/console

Likely caused by https://github.com/theforeman/foreman/pull/4909 - I will kick in the job after the situation is resolved, as I'm afraid we'll hit another OOM if I try to run tests again.

ekohl:
> about jessie, I'd prefer to remove packages in a separate PR, just to not block merging this changes

Absolutely, doesn't belong in this PR.

debian/trusty/foreman/compat
```diff
@@ -1 +1 @@
-7
```
ekohl: Maybe all changes to trusty files should be reverted so we don't imply they do work?

dLobatog:
[test]

dLobatog:
Still failing to build due to react-intl (http://ci.theforeman.org/job/packaging_build_deb_coreproject/lastBuild/arch=x86,label=debian,os=stretch/console), it was building before that was merged to develop.

evgeni:
my ACK remains, given develop is buildable at all ;)

dLobatog:
therefore, [test]

dLobatog:
[test]

dLobatog:
[test]

dLobatog:
[test]

mmoll:
This is looking good and it's time to bring this forward now.
Gracias @dLobatog!
And also thanks to @ekohl and @evgeni - this is making our DEB packaging of Foreman core much better!

dLobatog:
🙇

PR #1974, opened 2017-12-06
fixes #21899 - permit split of pulp_data.tar

jcpunk:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---
Patch for katello-backup

### DIFF
--- a/packages/katello/katello/backup.rb
+++ b/packages/katello/katello/backup.rb
@@ -324,7 +324,17 @@ module KatelloUtilities
     end
 
     def create_pulp_data_tar
-      run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pulp_data.tar')} --exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *")
+      archive_size = @options[:split_tar]
+      tar_command = "tar --selinux --create --file=#{File.join(@dir, 'pulp_data.tar')} "
+
+      if archive_size
+        split_tar_script = default_split_tar_script
+        tar_command += "--tape-length=#{archive_size} --new-volume-script=#{split_tar_script} "
+      end
+
+      tar_command += "--exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *"
+
+      run_cmd(tar_command)
     end
 
     def backup_config_files
@@ -362,6 +372,11 @@ module KatelloUtilities
           else
             opts.abort("Previous backup directory does not exist: #{dir_path}")
           end
+        end
+
+        opts.on("--split-pulp-tar SIZE", "Split pulp data into files of a specified size, i.e. (100M, 50G). See '--tape-length' in 'info tar' for all sizes") do |size|
+          opts.abort("Please specify size according to 'tar --tape-length' format.") unless size
+          @options[:split_tar] = size
         end
 
         opts.on("--online-backup", "Keep services online during backup") do |online|
--- a/packages/katello/katello/katello-backup-rotate-tar.sh
+++ b/packages/katello/katello/katello-backup-rotate-tar.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+## For this script it's advisable to use a shell, such as Bash,
+## that supports a TAR_FD value greater than 9.
+
+# This script taken from the tar info doc
+
+name=`expr ${TAR_ARCHIVE} : '\(.*\)\..*'`
+volnum=`printf "%04d" ${TAR_VOLUME}`
+filename="${name:-$TAR_ARCHIVE}.part${volnum:-$TAR_VOLUME}"
+case ${TAR_SUBCOMMAND} in
+-c)       ;;
+-d|-x|-t) test -r ${filename} || exit 1
+          ;;
+*)        exit 1
+esac
+
+echo ${filename} >&${TAR_FD}
+echo " Switched to multi-volume ${filename}"
--- a/packages/katello/katello/restore.rb
+++ b/packages/katello/katello/restore.rb
@@ -144,7 +144,17 @@ module KatelloUtilities
       puts "Logical backup detected, using the standard backup files to restore" if valid_logical_backup
       if @pulp_data_exists
         puts "Restoring Pulp data"
-        run_cmd("tar --selinux --overwrite --listed-incremental=/dev/null -xf pulp_data.tar -C /")
+
+        tar_command = "tar --selinux --overwrite --listed-incremental=/dev/null "
+
+        if File.exist?('pulp_data.part0002')
+          split_tar_script = default_split_tar_script
+          tar_command += "-M --new-volume-script=#{split_tar_script} "
+        end
+
+        tar_command += "-xf pulp_data.tar -C /"
+
+        run_cmd(tar_command)
       end
       if @mongo_data_exists
         run_cmd("tar --selinux --overwrite --listed-incremental=/dev/null -xzf mongo_data.tar.gz -C /")
--- a/packages/katello/katello/helper.rb
+++ b/packages/katello/katello/helper.rb
@@ -24,6 +24,24 @@ module KatelloUtilities
       "This utility can't run on a non-katello system."
     end
 
+    def default_split_tar_script
+      # check if production install
+      if __FILE__.include?('sbin')
+        utils_path='/usr/share/katello'
+      else
+        utils_path=File.expand_path('..', __FILE__)
+      end
+
+      split_tar_script = File.join(utils_path, 'katello-backup-rotate-tar.sh')
+      if !File.executable?(split_tar_script)
+        puts "**** ERROR: split_tar_script not executable ****"
+        puts split_tar_script
+        exit(-1)
+      end
+
+      split_tar_script
+    end
+
     def foreman_rpm_installed?
       system("rpm -q foreman > /dev/null")
     end
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -31,6 +31,7 @@ Source14:   restore.rb
 Source15:   backup.rb
 Source16:   hostname-change.rb
 Source17:   helper.rb
+Source18:   katello-backup-rotate-tar.sh
 
 BuildRequires: asciidoc
 BuildRequires: util-linux
@@ -98,6 +99,7 @@ install -m 644 %{SOURCE14} %{buildroot}%{_datarootdir}/katello/restore.rb
 install -m 644 %{SOURCE15} %{buildroot}%{_datarootdir}/katello/backup.rb
 install -m 644 %{SOURCE16} %{buildroot}%{_datarootdir}/katello/hostname-change.rb
 install -m 644 %{SOURCE17} %{buildroot}%{_datarootdir}/katello/helper.rb
+install -m 755 %{SOURCE18} %{buildroot}%{_datarootdir}/katello/katello-backup-rotate-tar.sh
 
 # install important scripts
 mkdir -p %{buildroot}%{_bindir}
@@ -152,6 +154,7 @@ Common runtime components of %{name}
 %{_datarootdir}/katello/backup.rb
 %{_datarootdir}/katello/hostname-change.rb
 %{_datarootdir}/katello/helper.rb
+%{_datarootdir}/katello/katello-backup-rotate-tar.sh
 %config(missingok) %{_sysconfdir}/cron.weekly/katello-clean-empty-puppet-environments
 %config(missingok) %{_sysconfdir}/cron.weekly/katello-remove-orphans
 %config(missingok) %{_sysconfdir}/cron.daily/katello-repository-publish-check
@@ -189,6 +192,7 @@ Provides a federation of katello services
 %{_datarootdir}/katello/backup.rb
 %{_datarootdir}/katello/hostname-change.rb
 %{_datarootdir}/katello/helper.rb
+%{_datarootdir}/katello/katello-backup-rotate-tar.sh
 
 # ------ Service ----------------
 %package service


mmoll:
general question, why is this file in the packaging repo and not in the main or an utility one?

ekohl:
@mmoll historical reasons. It used to be in katello-packaging. The intent is to move this to foreman_maintain but due to time it was just moved over so we could continue phasing out katello-packaging.

katello/katello/backup.rb
```diff
@@ -299,7 +299,13 @@ def generate_metadata
     end
 
     def create_pulp_data_tar
+      @split_size = @options[:split_pulp_tar] || "0"
       run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pulp_data.tar')} --exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *")
```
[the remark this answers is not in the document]
jcpunk: Yeah, that is better, I'll push a new version with that

jcpunk:
Pushed updated version

jcpunk:
I've fixed up katello-restore to also properly use the split up archives

ekohl:
@jlsherrill could you have a look as well?

katello/katello/backup.rb
```diff
@@ -336,6 +343,14 @@ def setup_opt_parser
           end
         end
 
+        opts.on("--split-pulp-tar SIZE", "Use split on pulp_data.tar to make smaller files") do |size|
+          @options[:split_pulp_tar] = size
+        end
+
+        opts.on("--split-pulp-tar-rotate-script /PATH/SCRIPT", "Use this script to facilitate splitting pulp_data.tar") do |script|
+          @options[:split_pulp_tar_script] = script
+        end
```
johnpmitsch: This looks like its optional - if that is correct, could you make a note saying so? i.e.
`Use this script to facilitate splitting pulp_data.tar. This is optional, a default script is provided`

katello/katello/katello-backup-rotate-tar.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
+## For this script it's advisable to use a shell, such as Bash,
+## that supports a TAR_FD value greater than 9.
+
+# echo Preparing volume $TAR_VOLUME of $TAR_ARCHIVE.
```
johnpmitsch: is this line meant to be here?

katello/katello/backup.rb
```diff
@@ -299,7 +299,14 @@ def generate_metadata
     end
 
     def create_pulp_data_tar
-      run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pulp_data.tar')} --exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *")
+      @archive_size = @options[:split_pulp_tar] || "0"
+      @split_pulp_tar_script = @options[:split_pulp_tar_script] || '/usr/share/katello/katello-backup-rotate-tar.sh'
```
johnpmitsch: having this hardcoded to `/usr/share/katello` does make it hard to test things out in development. You could do something like we have [here](https://git.io/vNFwU) maybe in a method that is added to helper.rb

katello/katello/restore.rb
```diff
@@ -19,6 +19,7 @@ def initialize(foreman_proxy_content, program, accepted_scenarios=nil)
       @program = program
 
       @confirmed = false
+      @split_pulp_tar_script = '/usr/share/katello/katello-backup-rotate-tar.sh'
```
johnpmitsch: since this is used in both scripts, it seems like a good candidate for helper.rb. (but also keeping in mind the comment about `/usr/share/katello` directory)

katello/katello/katello-backup-rotate-tar.sh
jcpunk: I just did a quick copy from the tar source, I'll get that fixed up

katello/katello/backup.rb
jcpunk: I can re-write that bit.

katello/katello/backup.rb
jcpunk: Can do

katello/katello/restore.rb
jcpunk: I'll take a stab at this

jcpunk:
New version pushed

jcpunk:
Just checking back in

johnpmitsch:
@jcpunk thanks for the update, overall this looks okay, but it will take a bit of manual testing to confirm, which I wll do soon. 

One general question: why provide the `--split-pulp-tar-rotate-script` option? Is there a reason a user would want to use a custom script vs. the one provided in this PR?

jcpunk:
There isn't a reason I can think of why a user would want to change the provided rotate script.  But they might be doing something super weird.

johnpmitsch:
@jcpunk I don't see a reason to provide `--split-pulp-tar-rotate-script` if you can already specify size of the split files. It adds complexity and opens up the script to user error. What do you think of removing that option?

jcpunk:
rebased off master

packages/katello/katello/backup.rb
```diff
@@ -324,7 +324,14 @@ def generate_metadata
...
 
     def create_pulp_data_tar
-      run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pulp_data.tar')} --exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *")
+      @archive_size = @options[:split_pulp_tar] || "0"
+      @split_pulp_tar_script = default_split_pulp_tar_script
+
+      if @archive_size == "0"
+        run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pulp_data.tar')} --exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *")
+      else
+        run_cmd("tar --selinux --create --tape-length=#{@split_size} --new-volume-script=#{@split_pulp_tar_script} --file=#{File.join(@dir, 'pulp_data.tar')} --exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *")
```
johnpmitsch: this and line 331 repeats a lot of the tar command. You could build the tar command dynamically to DRY it up i.e.
```ruby
tar_command = "tar --selinux --create "
tar_command += "--tape-length=#{archive_size} --new-volume-script=#{split_pulp_tar_script} " if archive_size
tar_command += "--exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *"
run_cmd(tar_command)
```

packages/katello/katello/backup.rb
```diff
@@ -324,7 +324,14 @@ def generate_metadata
     end
 
     def create_pulp_data_tar
-      run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pulp_data.tar')} --exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *")
+      @archive_size = @options[:split_pulp_tar] || "0"
+      @split_pulp_tar_script = default_split_pulp_tar_script
```
johnpmitsch: unless I'm missing something, I don't see a reason to make these instance variables with `@`

packages/katello/katello/backup.rb
```diff
@@ -324,7 +324,14 @@ def generate_metadata
     end
 
     def create_pulp_data_tar
-      run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pulp_data.tar')} --exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *")
+      @archive_size = @options[:split_pulp_tar] || "0"
```
johnpmitsch: The `|| 0` isn't needed, `@options[:split_pulp_tar]` will resolve to nil if nothing was passed, and it can be used in conditionals

packages/katello/katello/backup.rb
johnpmitsch: Also note I had to change `@split_size` to `archive_size` as I don't see that defined anywhere

johnpmitsch:
Just got a chance to test this out, sorrry about the delay. I found some issues:

I was getting an error with the `UTILS_PATH`:
```
/root/foreman-packaging/packages/katello/katello/backup.rb:8:in `require_relative': /root/foreman-packaging/packages/katello/katello/helper.rb:30: dynamic constant assignment (SyntaxError)
        UTILS_PATH='/usr/share/katello'
```
which is easily fixed by not using constants in the method
```
diff --git a/packages/katello/katello/helper.rb b/packages/katello/katello/helper.rb
index a5f2bc1..5d50b8b 100644
--- a/packages/katello/katello/helper.rb
+++ b/packages/katello/katello/helper.rb
@@ -27,12 +27,12 @@ module KatelloUtilities
     def default_split_pulp_tar_script
       # check if production install
       if __FILE__.include?('sbin')
-        UTILS_PATH='/usr/share/katello'
+        utils_path='/usr/share/katello'
       else
-        UTILS_PATH=File.expand_path('..', __FILE__)
+        utils_path=File.expand_path('..', __FILE__)
       end
 
-      File.join(UTILS_PATH, 'katello-backup-rotate-tar.sh'
+      File.join(utils_path, 'katello-backup-rotate-tar.sh')
     end
 
     def foreman_rpm_installed?
```

For the flag description, one nitpicky suggestion on verbage: `Use split on pulp_data.tar to make smaller files` could just be `Split pulp data into files of a specified size`

I don't see any specification of what to enter as a parameter for `--split-pulp-tar`, also incorrect input isn't handled, there should be some validation here.
```
[root@35-hnc katello]# ./katello-backup --split-pulp-tar 10 /backup
...snipped...

Failed 'tar --selinux --create --tape-length= --new-volume-script=/root/foreman-packaging/packages/katello/katello/katello-backup-rotate-tar.sh --file=/backup/katello-backup-20180223180551/pulp_data.tar --exclude=var/lib/pulp/katello-export --listed-incremental=/backup/katello-backup-20180223180551/.pulp.snar --transform 's,^,var/lib/pulp/,S' -S *' with exit code 2
Cleaning up backup folder and starting any stopped services... 
```

Looks like permissions need to be changed on the katello-backup-rotate-tar file
```
/bin/sh: /root/foreman-packaging/packages/katello/katello/katello-backup-rotate-tar.sh: Permission denied
tar: `/root/foreman-packaging/packages/katello/katello/katello-backup-rotate-tar.sh' command failed
tar: Error is not recoverable: exiting now
```

johnpmitsch:
It doesn't look like the pulp data file was actually split? 

```
[root@35-hnc katello]# ./katello-backup --split-pulp-tar 100M /backup
**** BACKUP Complete, contents can be found in: /backup/katello-backup-20180223183845 ****

[root@35-hnc katello]# ls -alh /backup/katello-backup-20180223183845
total 2.5G
drwxrwx---. 2 root postgres  203 Feb 23 18:53 .
drwxr-xr-x. 4 root root      200 Feb 23 18:48 ..
-rw-r--r--. 1 root root     1.5M Feb 23 18:38 config_files.tar.gz
-rw-r--r--. 1 root root      55K Feb 23 18:38 .config.snar
-rw-r--r--. 1 root root      32K Feb 23 18:38 metadata.yml
-rw-r--r--. 1 root root     2.3G Feb 23 18:49 mongo_data.tar.gz
-rw-r--r--. 1 root root      120 Feb 23 18:49 .mongo.snar
-rw-r--r--. 1 root root      57M Feb 23 18:48 pgsql_data.tar.gz
-rw-r--r--. 1 root root      21K Feb 23 18:48 .postgres.snar
-rw-r--r--. 1 root root     100M Feb 23 18:48 pulp_data.tar
-rw-r--r--. 1 root root     3.7M Feb 23 18:48 .pulp.snar
```

Let me know if I am using it wrong, the size I specified was using the syntax from the tar man pages and should have created 100 MB files

packages/katello/katello/backup.rb
jcpunk: Good call, I'll get that fixed

packages/katello/katello/backup.rb
jcpunk: I fear I'm not much of a ruby programmer.  I've not yet gotten the hang of when `@` makes sense.  I'll drop it from there.

packages/katello/katello/backup.rb
jcpunk: Neat!  I'll switch to that.

jcpunk:
I think I've got the last bit of feedback sorted out.  I've made the rotate script a bit more chatty to help review its behavior.

packages/katello/katello/backup.rb
```diff
@@ -324,7 +324,18 @@ def generate_metadata
     end
 
     def create_pulp_data_tar
-      run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pulp_data.tar')} --exclude=var/lib/pulp/katello-export --listed-incremental=#{File.join(@dir, '.pulp.snar')} --transform 's,^,var/lib/pulp/,S' -S *")
+      archive_size = @options[:split_tar]
+
```
johnpmitsch: this is nitpicky (sorry!) but you can do without as much whitespace here and below. I think just a space around the top and bottom of the if statement block is ok, same for the restore part as well

packages/katello/katello/backup.rb
```diff
@@ -364,6 +375,11 @@ def setup_opt_parser
           end
         end
 
+        opts.on("--split-pulp-tar SIZE", "Split pulp data into files of a specified size, see 'tar --tape-length' for syntax") do |size|
+          opts.abort("Please specify size according to 'tar --tape-length' format.") unless size
```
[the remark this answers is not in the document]
johnpmitsch: and is impossible to find the actual size command table! looks like `info tar` and some scrolling will show tape-length sizes. 

How about this:
"Split pulp data into files of a specified size, i.e. (100M, 50G). See '--tape-length' in 'info tar' for all sizes"
This should save most users from digging around tar docs

Also what happens if I pass `100XYZ` in there? This will get passed to the tar command and it will error. I think there should be a validation method here, let me know if you need help writing it.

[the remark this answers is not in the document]
jcpunk: > let me know if you need help writing it.

That would be very helpful.  I don't actually know ruby and am just programing by pattern....

jcpunk:
forklift looks neat!

It looks like run_command isn't setting TAR_FD.  I think I've seen this when running under /bin/sh rather than /bin/bash.

When I wrap the tar-command with bash -c it works as expected.  Is there a good way to do this in ruby?

johnpmitsch:
Looks like this isn't creating more than one file:

running `./packages/katello/katello/katello-backup --online-backup --split-pulp-tar 100k /backup/`, I get one 100k pulp_data.tar file:
```
[root@35-hnc foreman-packaging]# ls -alh /backup/katello-backup-20180313203418
total 13M
drwxrwx---. 3 root     postgres  195 Mar 13 20:35 .
drwxr-xr-x. 3 root     root      115 Mar 13 20:35 ..
-rw-r--r--. 1 postgres postgres 462K Mar 13 20:35 candlepin.dump
-rw-r--r--. 1 root     root     1.5M Mar 13 20:34 config_files.tar.gz
-rw-r--r--. 1 root     root      55K Mar 13 20:34 .config.snar
-rw-r--r--. 1 postgres postgres  11M Mar 13 20:35 foreman.dump
-rw-r--r--. 1 root     root      32K Mar 13 20:34 metadata.yml
drwxr-xr-x. 5 root     root       52 Mar 13 20:35 mongo_dump
-rw-r--r--. 1 postgres postgres  650 Mar 13 20:35 pg_globals.dump
-rw-r--r--. 1 root     root     100K Mar 13 20:35 pulp_data.tar
-rw-r--r--. 1 root     root     4.2K Mar 13 20:35 .pulp.snar
```

On a backup w/o `--split-pulp-data-tar`, running `./packages/katello/katello/katello-backup --online-backup  /backup/`, I am getting  a 340K pulp_data.tar file: 

```
[root@35-hnc foreman-packaging]# ls -alh /backup/katello-backup-20180313203548
total 13M
drwxrwx---. 3 root     postgres  195 Mar 13 20:36 .
drwxr-xr-x. 4 root     root      152 Mar 13 20:35 ..
-rw-r--r--. 1 postgres postgres 462K Mar 13 20:36 candlepin.dump
-rw-r--r--. 1 root     root     1.5M Mar 13 20:35 config_files.tar.gz
-rw-r--r--. 1 root     root      55K Mar 13 20:35 .config.snar
-rw-r--r--. 1 postgres postgres  11M Mar 13 20:36 foreman.dump
-rw-r--r--. 1 root     root      32K Mar 13 20:35 metadata.yml
drwxr-xr-x. 5 root     root       52 Mar 13 20:36 mongo_dump
-rw-r--r--. 1 postgres postgres  650 Mar 13 20:36 pg_globals.dump
-rw-r--r--. 1 root     root     340K Mar 13 20:36 pulp_data.tar
-rw-r--r--. 1 root     root     4.2K Mar 13 20:36 .pulp.snar
```

With the 100k parameter passed in the first example, shouldn't I expect 4 files to be created? (Additional note, I did try 100K as well).

I can check on the `bash -c` part and help with a validation method, but lets get this functional first :)

jcpunk:
As near as I can tell, wrapping the tar bits in `bash -c` are required for it to work.

I've had it echo the command out and run it by hand.  It works as expected there.

johnpmitsch:
@jcpunk ok, let me look into what we would need for that in ruby

johnpmitsch:
@jcpunk taking a look at this again, it looks like the tar script actually is working, even when I remove the `bash -c` part. Its just putting the files in the parent directory.
```
[root@35-hnc foreman-packaging]# ls -alh /backup/
total 252K
drwxr-xr-x.  3 root root      115 Mar 13 21:20 .
dr-xr-xr-x. 19 root root      253 Mar 13 21:14 ..
-rw-r--r--.  1 root root     100K Mar 13 21:20 katello-backup-2
drwxrwx---.  3 root postgres  195 Mar 13 21:20 katello-backup-20180313212033
-rw-r--r--.  1 root root     100K Mar 13 21:20 katello-backup-3
-rw-r--r--.  1 root root      50K Mar 13 21:20 katello-backup-4
```

```
[root@35-hnc foreman-packaging]# ls -alh /backup/katello-backup-20180313212033/
total 13M
drwxrwx---. 3 root     postgres  195 Mar 13 21:20 .
drwxr-xr-x. 3 root     root      115 Mar 13 21:20 ..
-rw-r--r--. 1 postgres postgres 462K Mar 13 21:20 candlepin.dump
-rw-r--r--. 1 root     root     1.5M Mar 13 21:20 config_files.tar.gz
-rw-r--r--. 1 root     root      55K Mar 13 21:20 .config.snar
-rw-r--r--. 1 postgres postgres  11M Mar 13 21:20 foreman.dump
-rw-r--r--. 1 root     root      32K Mar 13 21:20 metadata.yml
drwxr-xr-x. 5 root     root       52 Mar 13 21:20 mongo_dump
-rw-r--r--. 1 postgres postgres  650 Mar 13 21:20 pg_globals.dump
-rw-r--r--. 1 root     root     100K Mar 13 21:20 pulp_data.tar
-rw-r--r--. 1 root     root     4.2K Mar 13 21:20 .pulp.snar
```

Those files should be in the backup directory and I would think they would be called pulp-data-2.tar or similar instead of katello-backup-2

jcpunk:
I think I got it sorted

jcpunk:
I believe I've got this rigged up right now.  My forklift tests look like they should.

jcpunk:
Just checking back in

packages/katello/katello/katello-backup-rotate-tar.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+## For this script it's advisable to use a shell, such as Bash,
+## that supports a TAR_FD value greater than 9.
+
+# This script taken from the tar info doc
+
+name=`expr ${TAR_ARCHIVE} : '\(.*\)\..*'`
+volnum=`printf "%04d" ${TAR_VOLUME}`
```
johnpmitsch: Is there a way we can avoid the zero-padding and the "part"? i.e. `pulp_data.2` instead of  `pulp_data.part0002`

Also is this still a tar file? Should it have a `.tar` extension?

packages/katello/katello/katello-backup-rotate-tar.sh
```diff
@@ -0,0 +1,17 @@
...
+case ${TAR_SUBCOMMAND} in
+-c)       ;;
+-d|-x|-t) test -r ${name:-$TAR_ARCHIVE}.part${volnum:-$TAR_VOLUME} || exit 1
+          ;;
+*)        exit 1
+esac
+
+echo ${name:-$TAR_ARCHIVE}.part${volnum:-$TAR_VOLUME} >&${TAR_FD}
+echo " Switched to multi-volume ${name:-$TAR_ARCHIVE}.part${volnum:-$TAR_VOLUME}"
```
johnpmitsch: I'm not a bash expert, but I would think you would want `${name:-$TAR_ARCHIVE}.part${volnum:-$TAR_VOLUME}` as a variable since I see it used multiple places.

packages/katello/katello/katello-backup-rotate-tar.sh
johnpmitsch: I think `pulp_data.2.tar` would be fine unless I'm missing out on any convention around this

packages/katello/katello/restore.rb
```diff
@@ -144,7 +144,17 @@ def restore_databases
       puts "Logical backup detected, using the standard backup files to restore" if valid_logical_backup
       if @pulp_data_exists
         puts "Restoring Pulp data"
-        run_cmd("tar --selinux --overwrite --listed-incremental=/dev/null -xf pulp_data.tar -C /")
+
+        tar_command = "tar --selinux --overwrite --listed-incremental=/dev/null "
+
+        if File.exist?('pulp_data.tar-2')
```
johnpmitsch: The file currently is not called this, looks like this will need updating.

packages/katello/katello/katello-backup-rotate-tar.sh
jcpunk: I'd prefer to have some leading zeros so that they sort nicely for an `ls`.

The files themselves are tar like, but depending on where the rotate kicks off they may not be extractable via `tar` without the leading archive

packages/katello/katello/katello-backup-rotate-tar.sh
jcpunk: done

packages/katello/katello/restore.rb
jcpunk: fixed

packages/katello/katello/katello-backup-rotate-tar.sh
```diff
@@ -0,0 +1,18 @@
...
+## that supports a TAR_FD value greater than 9.
+
+# This script taken from the tar info doc
+
+name=`expr ${TAR_ARCHIVE} : '\(.*\)\..*'`
+volnum=`printf "%03d" ${TAR_VOLUME}`
+filename="${name:-$TAR_ARCHIVE}.part${volnum:-$TAR_VOLUME}"
+case ${TAR_SUBCOMMAND} in
+-c)       ;;
+-d|-x|-t) test -r ${name:-$TAR_ARCHIVE}.part${volnum:-$TAR_VOLUME} || exit 1
```
johnpmitsch: This can be replaced with `filename` too right?

jcpunk: good call, fixed

packages/katello/katello/katello-backup-rotate-tar.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+## For this script it's advisable to use a shell, such as Bash,
+## that supports a TAR_FD value greater than 9.
+
+# This script taken from the tar info doc
+
+name=`expr ${TAR_ARCHIVE} : '\(.*\)\..*'`
+volnum=`printf "%03d" ${TAR_VOLUME}`
```
johnpmitsch: The zero padding works, good point on the sorting. I wonder if we should go to 4 places like you originally had it? Users have databases in the TBs, so breaking into 1000+ files is not inconceivable

packages/katello/katello/restore.rb
```diff
@@ -144,7 +144,17 @@ def restore_databases
       puts "Logical backup detected, using the standard backup files to restore" if valid_logical_backup
       if @pulp_data_exists
         puts "Restoring Pulp data"
-        run_cmd("tar --selinux --overwrite --listed-incremental=/dev/null -xf pulp_data.tar -C /")
+
+        tar_command = "tar --selinux --overwrite --listed-incremental=/dev/null "
+
+        if File.exist?('pulp_data.part002')
```
johnpmitsch: Don't forget an extra zero here as well :)

johnpmitsch:
@jcpunk this still needs some sort of validator for the tape-length sizes, but if I'm going to write it, I think that is best done as a separate issue and I can make a PR for it. Still doing some more testing, so far so good! :+1:

jcpunk:
At this point I'd recommend someone other than me write the validator.  My ruby is just not up to par.

jlsherrill:
[test default]

johnpmitsch:
[test rpm]

jcpunk:
hmmmm, test default doesn't seem to have run....

johnpmitsch:
@jcpunk the tests changed so its only the rpm test that needs to run (which has and passed), Everything looks good from my end, @jlsherrill said he will take a glance over and I am going to double check the rpm build works ok since there were spec file changes and we can merge this after those two things :+1:

johnpmitsch:
rpm build appears to be working fine :100:

johnpmitsch:
Thanks @jcpunk this is great addition to katello backup!

PR #1980, opened 2017-12-14
Ref #19496 - Install passenger-recycler

swapab:
Updated the release version to 0.1.1.**2** this would install passenger-recycler to `/usr/bin/passenger-recycler`

/cc: @lzap


ekohl:
I'd prefer a new foreman_maintain which adds the passenger-recycler here: https://github.com/theforeman/foreman_maintain/blob/baae1bce258ad6b3d6cfa745e27c5a3d3a682f71/foreman_maintain.gemspec#L22

swapab:
@ekohl Thanks for the suggestion. I was not knowing that you can multiple executables to a gem.
https://github.com/theforeman/foreman_maintain/pull/133

ekohl:
Could you update the gem as well? Currently it can't it so it can't build. It's also easier to explicitly tell if you've updated the PR because we don't get a notification if you edit the original message hence why I was waiting with reviewing it.

swapab:
@ekohl I think I have updated gem to 0.1.2 https://rubygems.org/gems/foreman_maintain/versions/0.1.2
Is that what you meant? Or it's something else.

ekohl:
@swapab when bumping a package you remove the old gem file and (git annex) add the new one. See https://github.com/theforeman/foreman-packaging/blob/29a9278054815c829a0c686ee4a3931d0b768e72/bump_rpm.sh#L32-L36

swapab:
@ekohl Got it! Bumped version along with updated gem file. This should make sense now.
Thanks!

rubygem-foreman_maintain/rubygem-foreman_maintain.spec
```diff
@@ -85,6 +86,9 @@ install -D -m0640 %{buildroot}%{gem_instdir}/config/foreman_maintain.yml.packagi
 %doc %{gem_instdir}/README.md
 
 %changelog
+* Tue Jan 17 2018 Swapnil Abnave <sabnave@redhat.com> 0.1.2-1
```
ekohl: The changelog is generated when we tag so there should be no need to update it.

rubygem-foreman_maintain/rubygem-foreman_maintain.spec
```diff
@@ -120,4 +124,3 @@ install -D -m0640 %{buildroot}%{gem_instdir}/config/foreman_maintain.yml.packagi
 
 * Mon Feb 27 2017 Anurag Patel <apatel@redhat.com> 0.0.1-1
 - Package foreman_maintain into RPM (#3, apatel@redhat.com)
-
```
ekohl: This line is added back by tito when we tag it so could you keep it in to reduce the amount of changes?

swapab: Ah! I had configured emacs to delete-trailing-white-space. Fixed now.

ekohl: Totally makes sense and normally I'd do exactly the same. For now this is what tito does though :(

ekohl:
Thanks!

PR #1984, opened 2017-12-15
Update foreman_snapshot_management to 1.3.0 [rpm]

timogoebel:
Please build for:

* [x] Nightly
* [x] 1.16
* [x] 1.15
* [x] 1.14


ekohl:
Looks like deface was added as a dependency. Mind adding that as a Requires? Also wondering if we should backport anything into 1.14.

timogoebel:
@ekohl: Thanks. I noticed the changed dependency but forgot to add it.
The plugin support 1.14. But I don't care as long as we build it for 1.15+.

ekohl:
Looks like the fedora 24 build on 1.15 failed because it was never whitelisted. Unless you object I'm not going to mess with that since the EL7 build did pass.

PR #1985, opened 2017-12-16
Rpm/foreman tasks 0.10.8

iNecas:

* [x] Nightly
* [x] 1.16
* [ ] 1.15
* [ ] 1.14

---

We put back the init scripts for now, as we need this at least for 1.16 to fix rex there. Even for nightlies, I would rather keep this around, until https://github.com/theforeman/foreman-packaging/pull/1549/files resolved that once and for all.


iNecas:
It seems we are blocked here on npm as well:(

iNecas:
Replaced by https://github.com/theforeman/foreman-packaging/pull/2060

PR #1987, opened 2017-12-18
Bump nodejs-patternfly to 3.32.0

ekohl:


ekohl:
I'm testing a new npm2rpm template so there are some failures due to that.

ekohl:
Somehow it does need networking, likely because it's missing caches. I can't figure out why it doesn't store those in the metadata tarball but does store them for older versions.

dLobatog:
[test]

ekohl:
@dLobatog I spent a few hours yesterday trying to understand why - I simply can't make it produce the cache for that particular package version. I think I could produce a correct cache for older versions. Also tried it in a CentOS container rather than my Fedora. Still stuck. If I can't figure it out I'll just try to unbundle this and see if we can make that work.

ekohl:
I couldn't figure it out so I decided to unbundle it. It has become a massive PR because of this. Also opened https://github.com/patternfly/patternfly/issues/892 to check if the downgrade is actually needed.

ekohl:
@dLobatog could you re-review? Given you were unable to create a cache for this version as well I feel like this is currently the best approach.

nodejs-bootstrap-datepicker/nodejs-bootstrap-datepicker.spec
```diff
@@ -0,0 +1,45 @@
+%global npm_name bootstrap-datepicker
+%global enable_tests 1
+
+Name: nodejs-%{npm_name}
+Version: 1.7.1
```
ehelms: In doing some Rails 5.1 testing (cause I need this patch) I ran into patternflly has a requirement of ~1.6.4 which translates into < 1.7.0 within the patternfly RPM spec so I think we need to lower this version included?

ehelms:
```
DEBUG util.py:439:  Error: Package: nodejs-eonasdan-bootstrap-datetimepicker-4.17.47-1.git.4467.62b2c13.el7.centos.noarch (copr_base)
DEBUG util.py:439:             Requires: npm(moment-timezone) < 0.5
DEBUG util.py:439:             Installing: nodejs-moment-timezone-0.5.14-1.git.4466.62b2c13.el7.centos.noarch (copr_base)
DEBUG util.py:439:                 npm(moment-timezone) = 0.5.14
DEBUG util.py:439:  Error: Package: nodejs-eonasdan-bootstrap-datetimepicker-4.17.47-1.git.4467.62b2c13.el7.centos.noarch (copr_base)
DEBUG util.py:439:             Requires: npm(moment) < 2.11
DEBUG util.py:439:             Installing: nodejs-moment-2.14.1-1.git.846.62b2c13.el7.centos.noarch (copr_base)
DEBUG util.py:439:                 npm(moment) = 2.14.1
DEBUG util.py:439:             Available: nodejs-moment-2.17.1-1.git.0.09c913a.el7.centos.noarch (copr_base)
DEBUG util.py:439:                 npm(moment) = 2.17.1
```

ehelms:
That last comment looks like its from an old build I had disregard.

nodejs-bootstrap-datepicker/nodejs-bootstrap-datepicker.spec
ekohl: I assume you found that with a repoclosure check. It'd be good to have that here as well.

ehelms:
Turns out that is not an old build but an issue we need to figure out. It appears that requires are being auto-generated for that package, and following its interpretation of semver, it's generating restraints that break. You can see from this output generated by `rpm -qp --requires`:

```
nodejs(engine)
npm(bootstrap) >= 3.3
npm(bootstrap) >= 3.3
npm(bootstrap) < 3.4
npm(jquery)
npm(jquery) >= 1.8.3
npm(jquery) < 4.0.0
npm(moment) >= 2.10
npm(moment) >= 2.10
npm(moment) < 2.11
npm(moment-timezone) >= 0.4.0
npm(moment-timezone) >= 0.4.0
npm(moment-timezone) < 0.5
npm(moment-timezone) < 1.0.0
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
```

That there is a `npm(moment-timezone) < 0.5` being added because its interpreting the `package.json` entry which says `^0.4.0` to mean variability only within the Z digit.

ehelms:
Even with this change, I am still hitting:

```
webpack --bail --config config/webpack.config.js
'
ModuleNotFoundError: Module not found: Error: Can't resolve 'isomorphic-fetch' in '/builddir/build/BUILD/foreman-1.17.0/webpack/assets/javascripts/react_app/redux/actions/common'
```

ekohl:
I still prefer unbundled dependencies (and think we could still do a lot better). While doing this I opened upstream PRs and issues to fix licensing metadata and requested looser version pinning on dependencies leading to less duplication. While it may be painful and a lot of work right now, we can reduce that pain a lot by better tooling. In the process I improved npm2rpm (yet to be merged).

That said, right now it might not be viable to do so but I did learn a lot about npm in the process.

PR #1990, opened 2017-12-18
Fixes #22014 - Katello-restore with inc backup

johnpmitsch:
Katello-restore currently doesn't work with incremental
backups because the config_files.tar.gz file doesn't contain
the httpd.conf file with the hostname in it. This change
skips that check for incremental backups.

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---

### DIFF
--- a/katello/katello/restore.rb
+++ b/katello/katello/restore.rb
@@ -1,3 +1,4 @@
+require 'rubygems/package'
 require 'socket'
 require 'optparse'
 require 'fileutils'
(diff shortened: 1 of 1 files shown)


katello/katello/restore.rb
```diff
@@ -198,11 +199,31 @@ def display_backup_options
...
-      backup_hostname == hostname
+      config_tarball = File.join(@dir, 'config_files.tar.gz')
+      config_files = tarball_file_list(config_tarball)
+
+      # Incremental backups sometimes don't include httpd.conf. Since a "base" backup is restored before an incremental, we can
+      # assume that the hostname is checked during the base backup restore
+      if config_files.include?("etc/httpd/conf/httpd.conf")
+        backup_hostname = run_cmd("tar zxf #{config_tarball} etc/httpd/conf/httpd.conf --to-stdout | grep ServerName | awk {'print $2'} | tr -d '\"'").chomp
+        hostname = Socket.gethostname.chomp
+        backup_hostname == hostname
```
cfouant: After testing, this will fail without the `else true` line on an incremental restore. Other than that, works well, ACK pending requested changes! Thanks @johnpmitsch

PR #1992, opened 2017-12-19
Fix the nodejs-react-fontawesome spec

ekohl:


ekohl:
Updated to increase the release field.

PR #1993, opened 2017-12-20
Update nodejs-patternfly to 3.32.0

dLobatog:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---

Let's see if this works - basically it'd be the same as https://github.com/theforeman/foreman-packaging/pull/1987


ehelms:
I assume this is an attempt to keep it bundled and avoid https://github.com/theforeman/foreman-packaging/pull/1987 ?

ekohl:
It also fails on network access so I'm guessing that somehow NPM can't generate a cache for this version that's complete.

dLobatog:
Another shot in the dark. Check out what version of npm koji is using:

```
npm ERR! argv "/usr/bin/node" "/usr/bin/npm" "install" "--cache-min" "Infinity" "--cache" "." "--no-optional" "--global-style" "true" "react-bootstrap@0.31.5"
npm ERR! node v6.11.3
npm ERR! npm  v3.10.10
```

yet npm2rpm is not creating the cache for that version. This has been *important* in the past, caches vary slightly between 1.x and 2.x npm, enough to make them incompatible. In the past what I had done in npm2rpm is to simply put 1.x and 2.x caches together, maybe this can be done with 3.x too.
```
https://github.com/dLobatog/npm2rpm/blob/master/bin/generate_npm_tarball.sh
```

I tried to change the version to 3.32 and compare the only cache file it creates but they seem the same:

```
diff .cache.json /tmp/mozilla_dlobatog0/patternfly-3.32.0-registry.npmjs.org/registry.npmjs.org/patternfly/.cache.json
```

ehelms:
[test]

ehelms:
Managed to get this re-tested and it's still failing.

dLobatog:
On all other PRs I've recently submitted, the cache problems went away after creating the cache with 3.10.10 - in this one it still fails to install indeed. I'll keep on investigating..

ekohl:
@dLobatog I had the same issue: tried with NPM 1, 2 and 3. They could all create a cache for older patternfly versions but not this one.

nodejs-patternfly/nodejs-patternfly.spec
```diff
@@ -2,69 +2,76 @@
 %global enable_tests 0
 
 Name: nodejs-%{npm_name}
-Version: 3.29.13
+Version: 3.32.1
 Release: 1%{?dist}
-Summary: A reference implementation of PatternFly based on Bootstrap v3
+Summary: This reference implementation of PatternFly is based on [Bootstrap v3](http://getbootstrap
```
ekohl: Could you change this back? I'm not sure yet how we should deal with this better in npm2rpm yet.

PR #1995, opened 2017-12-21
fixes #22060 - clean up if postgres lacks write access

cfouant:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---

### DIFF
--- a/katello/katello/backup.rb
+++ b/katello/katello/backup.rb
@@ -20,6 +20,7 @@ module KatelloUtilities
       @databases = @databases.dup
       @accepted_scenarios = accepted_scenarios
       @last_scenario = self.last_scenario
+      @services_online = true
 
       # keep as variables for easy backporting
       @foreman_proxy_content = foreman_proxy_content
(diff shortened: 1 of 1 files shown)


johnpmitsch:
@cfouant I like the approach of having a flag (`services_online`) to keep track of whether services have been started. However, I don't think changing that flag manually in the logic is the best approach. It may work fine now, but I could see it easily getting out of sync as the script gets updated. 

What if we used a wrapper method for starting and stopping services services. i.e.
```ruby
def start_services
  run_cmd("katello-service start #{@excluded}") if !@services_online
  @services_online = true
end

def stop_services
  run_cmd("katello-service stop #{@excluded}") if @services_online
  @services_online = false
end
```
Then replacing the `katello-service start/stop` calls with these methods. This way the `@services_online` variable is going to stay in sync with the starting and stopping of services, and we don't unnecessarily start or stop services

The only exception is the part inside the cleanup method, where we can't use `run_cmd` as you already know, but we could change it to this:
`katello-service start #{@excluded}` unless @services_online

cfouant:
@johnpmitsch :+1: Love that suggestion! Added, let me know if you have additional thoughts!

johnpmitsch:
@cfouant great, looks fine, I'll test functionality soon

katello/katello/backup.rb
```diff
@@ -429,11 +438,11 @@ def run
           end
         else
           confirm unless @options[:confirm]
-          run_cmd("katello-service stop #{@excluded}")
+	  stop_services
```
ekohl: Mixing of tabs and spaces

katello/katello/backup.rb
```diff
@@ -429,11 +438,11 @@ def run
...
           @databases.each do |database|
             offline_backup(database)
           end
-          run_cmd("katello-service start #{@excluded}")
+	  start_services
```
ekohl: Mixing of tabs and spaces

cfouant:
@ekohl thanks for the fine-toothed comb! changes made, please re-review when you have the opportunity, thanks!

johnpmitsch:
@cfouant looks like the issue needs to be moved to the packaging project and then we can merge!

PR #2002, opened 2017-12-26
Update nodejs-react-bootstrap to 0.31.5

dLobatog:
* [x] Nightly

Needed to build foreman nightly since theforeman/foreman@3e1be3a#diff-b9cfc7f2cdf78a7f4b91a753d10865a2


dLobatog:
Apparently cache isn't working http://koji.katello.org/kojifiles/work/tasks/3581/63581/build.log

nodejs-react-bootstrap/nodejs-react-bootstrap.spec
```diff
@@ -9,68 +9,66 @@ License: MIT
...
+Source22: http://registry.npmjs.org/whatwg-fetch/-/whatwg-fetch-2.0.3.tgz
+Source23: http://registry.npmjs.org/node-fetch/-/node-fetch-1.7.3.tgz
 Source24: http://registry.npmjs.org/is-stream/-/is-stream-1.1.0.tgz
 Source25: http://registry.npmjs.org/encoding/-/encoding-0.1.12.tgz
 Source26: http://registry.npmjs.org/iconv-lite/-/iconv-lite-0.4.19.tgz
-Source27: http://registry.npmjs.org/setimmediate/-/setimmediate-1.0.5.tgz
+Source27: http://registry.npmjs.org/ua-parser-js/-/ua-parser-js-0.7.17.tgz
 Source28: react-bootstrap-0.31.5-registry.npmjs.org.tgz
 BuildRequires: nodejs-packaging
-BuildRequires: npm(react) >= 15.3.0
```
ekohl: These are here since they're peer dependencies. Not sure if that's relevant but I kept them in when I updated to be safe.

dLobatog: I don't think that's needed anymore, but I can bring them back

nodejs-react-bootstrap/nodejs-react-bootstrap.spec
```diff
@@ -87,9 +87,7 @@ done
 %setup -T -q -a 28 -D -n %{npm_cache_dir}
 
 %build
-mkdir node_modules
```
ekohl: Should this also be kept in because of the peerDependencies?

nodejs-react-bootstrap/nodejs-react-bootstrap.spec
```diff
@@ -124,4 +122,3 @@ rm -rf %{buildroot} %{npm_cache_dir}
 
 * Fri Oct 14 2016 Dominic Cleal <dominic@cleal.org> 0.30.5-1
 - new package built with tito
-
```
ekohl: Should we fix this in the template/update script so we don't go back and forth all the time?

nodejs-react-bootstrap/nodejs-react-bootstrap.spec
```diff
@@ -87,9 +87,7 @@ done
 %setup -T -q -a 28 -D -n %{npm_cache_dir}
 
 %build
-mkdir node_modules
-ln -s %{nodejs_sitelib}/{react,react-dom} node_modules/
-npm install --cache-min Infinity --cache %{npm_cache_dir} --no-optional --global-style true %{npm_name}@%{version}
+npm install --cache-min Infinity --cache %{npm_cache_dir}  --no-optional --global-style true %{npm_name}@%{version}
```
ekohl: There's an added space here

nodejs-react-bootstrap/nodejs-react-bootstrap.spec
dLobatog: Definitely, I removed this by hand, it isn't very relevant though

ekohl: tito adds it again so you'll just end up with noise in the git log.

ekohl:
Oops, just noticed the actual version bump was also reverted.

ekohl:
This was already version 0.31.5 so I'm not going to tag the change.

PR #2013, opened 2018-01-02
Handle nodejs package update by script

dLobatog:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---

### DIFF
--- a/add_npm_package.sh
+++ b/add_npm_package.sh
@@ -72,6 +72,12 @@ if [[ -z $STRATEGY ]] ; then
 fi
 
 echo -n "Making directory..."
+UPDATE=false
+if [ -f "$PACKAGE_NAME/*.spec" ]; then
+  UPDATE=true
+  sed -n '/%changelog/,$p' $PACKAGE_NAME/*.spec > OLD_CHANGELOG
+  git rm -r $PACKAGE_NAME
+fi
 mkdir $PACKAGE_NAME
 echo "FINISHED"
 echo -n "Creating specs and downloading sources..."
@@ -79,6 +85,12 @@ npm2rpm -n $NPM_MODULE_NAME -v $VERSION -s $STRATEGY
 echo "FINISHED"
 echo -n "Copying specs..."
 cp npm2rpm/SPECS/* $PACKAGE_NAME
+if [ "$UPDATE" = true ]; then
+  echo "Restoring changelogs..."
+  cat OLD_CHANGELOG >> $PACKAGE_NAME/*.spec
+  sed -i '/^%changelog/,/^%changelog/{0,//!d}' $PACKAGE_NAME/*.spec
+  rm OLD_CHANGELOG
+fi
 echo "FINISHED"
 echo -n "Copying sources..."
 cp npm2rpm/SOURCES/* $PACKAGE_NAME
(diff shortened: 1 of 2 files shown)


ekohl:
I'd first like to get https://github.com/theforeman/foreman-packaging/pull/2014 merged and this rebased. It introduces some nice variables that make it a bit more readable.

add_npm_package.sh
```diff
@@ -70,13 +70,25 @@ if [[ -z $STRATEGY ]] ; then
 fi
 
 echo -n "Making directory..."
+UPDATE=false
+if [ -d "nodejs-$NPM_MODULE_NAME" ]; then
+  UPDATE=true
+  sed -n '/%changelog/,$p' nodejs-$NPM_MODULE_NAME/*.spec > OLD_CHANGELOG
+  rm -r nodejs-$NPM_MODULE_NAME;
```
ekohl: the semi-colon isn't needed

add_npm_package.sh
```diff
@@ -70,13 +70,25 @@ if [[ -z $STRATEGY ]] ; then
...
+fi
 mkdir nodejs-$NPM_MODULE_NAME
 echo "FINISHED"
 echo -n "Creating specs and downloading sources..."
 npm2rpm -n $NPM_MODULE_NAME -v $VERSION -s $STRATEGY
 echo "FINISHED"
 echo -n "Copying specs..."
 cp npm2rpm/SPECS/* nodejs-$NPM_MODULE_NAME
+echo "Restoring changelogs..."
```
ekohl: This should be in the if statement

add_npm_package.sh
```diff
@@ -70,13 +70,25 @@ if [[ -z $STRATEGY ]] ; then
...
+if [ "$UPDATE" = true ]; then
+  cat OLD_CHANGELOG >> nodejs-$NPM_MODULE_NAME/*.spec
+  sed '/changelog/,/changelog/{0,//!d}' nodejs-webpack/nodejs-webpack.spec
```
ekohl: Should this be ```%changelog``` just to be sure?

add_npm_package.sh
```diff
@@ -70,13 +70,25 @@ if [[ -z $STRATEGY ]] ; then
 fi
 
 echo -n "Making directory..."
+UPDATE=false
+if [ -d "nodejs-$NPM_MODULE_NAME" ]; then
+  UPDATE=true
+  sed -n '/%changelog/,$p' nodejs-$NPM_MODULE_NAME/*.spec > OLD_CHANGELOG
```
ekohl: Would ```^%changelog``` be even more specific and better?

add_npm_package.sh
```diff
@@ -72,13 +72,25 @@ if [[ -z $STRATEGY ]] ; then
 fi
 
 echo -n "Making directory..."
+UPDATE=false
+if [ -d "$PACKAGE_NAME" ]; then
```
ekohl: One thing I've run into a few times is that if the script fails the directory is created but there are no files in it. I'd prefer to check if the spec file actually exists. Especially since you're using it with sed later.

add_npm_package.sh
```diff
@@ -72,13 +72,25 @@ if [[ -z $STRATEGY ]] ; then
 fi
 
 echo -n "Making directory..."
+UPDATE=false
+if [ -d "$PACKAGE_NAME/*.spec" ]; then
```
ekohl: Should be ```-e``` or ```-f``` since that should be a file.

dLobatog:
@ekohl Updated

PR #2021, opened 2018-01-03
Update tfm to use Rails 5.1 and Ruby 2.4

ehelms:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---

### DIFF
--- a/tfm/tfm.spec
+++ b/tfm/tfm.spec
@@ -6,10 +6,10 @@
 %{?scl_package:%scl_package %scl}
 
 # Fallback to sclo-ror42 etc. when scldevel's not in the buildroot
-%{!?scl_ror:%global scl_ror sclo-ror42}
-%{!?scl_prefix_ror:%global scl_prefix_ror %{scl_ror}-}
-%{!?scl_ruby:%global scl_ruby rh-ruby22}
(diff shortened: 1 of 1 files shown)


tfm/tfm.spec
```diff
@@ -6,10 +6,10 @@
...
 
 # Fallback to sclo-ror42 etc. when scldevel's not in the buildroot
-%{!?scl_ror:%global scl_ror sclo-ror42}
-%{!?scl_prefix_ror:%global scl_prefix_ror %{scl_ror}-}
-%{!?scl_ruby:%global scl_ruby rh-ruby22}
-%{!?scl_prefix_ruby:%global scl_prefix_ruby %{scl_ruby}-}
+%global scl_ror tfm-ror51
+%global scl_prefix_ror %{scl_ror}-
+%global scl_ruby rh-ruby24
+%global scl_prefix_ruby %{scl_ruby}-
```
ehelms: I modified this to make it explicit. I was not a big fan of the fact that these would get magically set by the buildroot meaning that the buildroot could be different than the defaults specified here. This in turn caused this to explicitly fail since it forced it to require tfm-ror51 and rh-ruby24 in a previous run. Now that I have modified the buildroot, this should pass accurately.

ehelms:
@ekohl Can you double check my modifications?

PR #2039, opened 2018-01-05
Update fog-google gem

ekohl:
This was needed for me because otherwise git-annex refused to get the file.


ekohl:
Due to this fog-google has been tagged but not rebuilt.

PR #2049, opened 2018-01-07
Add package.json dependency updater

ekohl:
This allows automatic updating from package.json to the relevant RPM requirements.


ekohl:
Since we're looking at creating a separate tooling repository I'm wondering if this is the right repository to house this script, but we can always remove it later. I'll at least have a look at the README to document it.

PR #2057, opened 2018-01-08
Updated apipie-bindings to 0.2.1

mbacovsky:

* [x] Nightly
* [x] 1.17
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---


mmoll:
Having the constraint on oauth is highly problematic. As we have only Ruby 2.x left, that should be removed from the gemspec.

mbacovsky:
@mmoll, thanks for the patch. I'll release updated bindings and send new PR with that

PR #2059, opened 2018-01-09
Fixes #22181 - Update katello-change-hostname to work with custom certs

johnpmitsch:

* [x] Nightly
* [ ] 1.16
* [ ] 1.15
* [ ] 1.14

---

### DIFF
--- a/katello/katello/helper.rb
+++ b/katello/katello/helper.rb
@@ -36,6 +36,15 @@ module KatelloUtilities
       exit(false)
     end
 
+    def fail_if_file_not_found(files)
+      files.reject! { |file| File.exist?(file) }
+      if files.any?
+        multiple_files = files.count > 1
+        fail_with_message("Error: #{multiple_files ? "Files" : "File"} #{files.join(", ")} " \
+                          "#{multiple_files ? "do" : "does"} not exist! Please check the file path and try again.")
+      end
+    end
+
     def run_cmd(command, exit_codes=[0], message=nil)
       result = `#{command}`
       unless exit_codes.include?($?.exitstatus)
--- a/katello/katello/hostname-change.rb
+++ b/katello/katello/hostname-change.rb
@@ -47,11 +47,7 @@ module KatelloUtilities
     def check_for_certs_tar
       STDOUT.puts "Checking for certs tarball"
       if @options[:certs_tar]
-        if File.file?(@options[:certs_tar])
-          true
-        else
-          self.fail_with_message("#{@options[:certs_tar]} does not exist! Please check the file path and try again")
-        end
+        fail_if_file_not_found([@options[:certs_tar]])
       else
         self.fail_with_message("You must specify --certs-tar argument when on a #{@proxy}." \
                                " These can be generated on the #{@default_program} server using " \
@@ -107,8 +103,6 @@ module KatelloUtilities
     def successful_hostname_change_message
     # the following multi-line string isn't indented because the indents are taken literally.
     successful_message = %(
-  If you want to use custom certificates, re-run the #{@options[:program]}-installer with custom certificate options.
-
   You will have to install the new bootstrap rpm and reregister all clients and #{@plural_proxy} with subscription-manager
   (update organization and environment arguments appropriately):
 
@@ -161,6 +155,16 @@ module KatelloUtilities
       proxy["Id"]
     end
 
+    def using_custom_certs?(scenario_answers)
+      scenario_answers["certs"]["server_cert"] &&
+      scenario_answers["certs"]["server_key"] &&
+      scenario_answers["certs"]["server_cert_req"]
+    end
+
+    def all_custom_cert_options_present?(scenario_answers)
+      @options[:custom_cert] && @options[:custom_key] && @options[:custom_cert_req]
+    end
+
     def setup_opt_parser
       @opt_parser = OptionParser.new do |opt|
         opt.banner = "usage: #{@command_prefix}-change-hostname hostname [options]"
@@ -194,7 +198,19 @@ module KatelloUtilities
           opt.on("-c",
                  "--certs-tar certs_tar",
                  "the path to the certs tar generated on the #{@default_program} server with the new hostname (required for #{@plural_proxy})") do |certs_tar|
-            @options[:certs_tar] = certs_tar
+            @options[:certs_tar] = File.expand_path(certs_tar)
+          end
+        else
+          opt.on("-c","--custom-cert CERT","If you are using custom certificates, please provide a server cert with the new hostname") do |custom_cert|
+            @options[:custom_cert] = File.expand_path(custom_cert)
+          end
+
+          opt.on("-k","--custom-key KEY","If you are using custom certificates, please provide a server key with the new hostname") do |custom_key|
+            @options[:custom_key] = File.expand_path(custom_key)
+          end
+
+          opt.on("-r","--custom-cert-req CERTREQ","If you are using custom certificates, please provide a certificate request with the new hostname") do |custom_cert_req|
+            @options[:custom_cert_req] = File.expand_path(custom_cert_req)
           end
         end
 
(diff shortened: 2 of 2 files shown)


katello/katello/hostname-change.rb
```diff
@@ -161,6 +161,16 @@ def get_default_proxy_id
       proxy["Id"]
     end
 
+    def custom_cert_options_check(scenario_answers)
+      if scenario_answers["certs"]["server_cert"] &&
+         scenario_answers["certs"]["server_key"] &&
+         scenario_answers["certs"]["server_cert_req"] &&
```
ekohl: The trailing ```&&``` is odd here. Technically it'll work but I'd rather get rid of it.

johnpmitsch: +1 had something after it that I removed and I forgot to delete the "&&"

johnpmitsch:
From the commit message:
katello-change-hostname fails with custom certs currently. Currently
there are no arguments to pass new certs that match the new hostname
to the installer. This commit adds those options and some checks    
to see if the user is using custom certs. If they are using custom
certs, katello-certs-check will run before any changes are made.
On proxies w/ content, a certs tar must be made with the new
custom certs. This is passed into the existing --certs-tar option.

johnpmitsch:
To review:
Please test the following scenarios plus any edge cases you can think of.

- [ ] change hostname on an install with custom certs is successful
- [ ] change hostname on a proxy w/ content with custom certs is successful
- [ ] change hostname on a normal katello install is successful
- [ ] change hostname on a normal proxy w/ content install is successful

I'm happy to walk through any or all of the scenarios with a reviewer

johnpmitsch:
1. set up a blank centos7 box
1. Follow instructions on setting up custom certs (I will send you an article).
2. Once the custom certs are created, you will need to install the katello repositories and yum install katello as specified in [our docs](https://theforeman.org/plugins/katello/3.5/installation/index.html). I used 3.5 to test, you can also use nightly, but I wouldn't test on any older version as the code will be too far out of sync.
3. When you get to the installer step, pass in the custom certs as arguments. There should be an example custom cert article that is roughly the same.
4. Generate new certs with a new hostname using the same 'ownca' tool mentioned in the article. You won't need to generate a new CA here, just using `./generate-crt.sh <HOSTNAME>`. We will now use them in the hostname change.
5. Run katello-change-hostname with the new certs. i.e.
`katello-change-hostname -u admin -p changeme -c /cert/changed.example.com.crt -k /cert/changed.example.com.key -r /cert/changed.example.com.crt.req -b /cert/cacert.crt  -y changed.example.com`
6. The hostname of the katello box should be changed and katello should function properly

johnpmitsch:
For a foreman proxy w/ content (this is fine to do after changing the hostname on the original katello box) :
1. spin up another centos7 box, we will install the foreman-proxy here
2. on the katello box, generate new certs that match the hostname of the new box. (use ./generate-crt.sh again)
3. on the katello box, run capsule-certs-generate, but pass in the new certs generated for the proxy: i.e.
`foreman-proxy-certs-generate --server-ca-cert /certs/cacert.crt --server-cert /certs/customcertscapsule.john.example.com.crt --server-cert-req /certs/customcertscapsule.john.example.com.crt.req --server-key /certs/customcertscapsule.john.example.com.key --certs-tar /root/customcertscapsule.tar.gz --foreman-proxy-fqdn "customcertscapsule.john.example.com"`
3. Use the output to install the capsule as normal. This will give you a proxy installed with custom certs.
4. To run katello-change-hostname on the proxy, repeat steps 2 and 3 with a new hostname (can be any valid fqdn).
5. Ignore the output from foreman-proxy-certs-generate, just copy the certs tar that was generated to the capsule.
6. Run katello-change-hostname on the proxy with the new certs tar: i.e.
`./katello-change-hostname -u admin -p changeme -c /root/changedcapsule.tar.gz -y changedcapsule.example.com`

johnpmitsch:
installer fails w/o full path being passed in:
```
[ERROR 2018-01-10 20:44:13 verbose] Parameter certs-server-cert invalid: newname.cfoo.example.com.crt is not one of regexes matching /^(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+))/ or regexes matching /^\/([^\/\0]+\/*)+$/
[ERROR 2018-01-10 20:44:13 verbose] Parameter certs-server-key invalid: newname.cfoo.example.com.key is not one of regexes matching /^(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+))/ or regexes matching /^\/([^\/\0]+\/*)+$/
[ERROR 2018-01-10 20:44:13 verbose] Parameter certs-server-cert-req invalid: newname.cfoo.example.com.crt.req is not one of regexes matching /^(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+))/ or regexes matching /^\/([^\/\0]+\/*)+$/
```
We can check if the file is valid and expand the path for them, then use that to pass into the installer

cfouant:
@johnpmitsch ACK - these work great! thank you! @ekohl we need you to approve the requested changes made before this can merge.

katello/katello/helper.rb
```diff
@@ -36,6 +38,12 @@ def fail_with_message(message, opt_parser=nil)
       exit(false)
     end
 
+    def check_if_file_exists(file)
+      unless File.exist?(file) && (Pathname.new file).absolute?
```
ekohl: Nit: I prefer ```Pathname.new(file).absolute?```

katello/katello/hostname-change.rb
```diff
@@ -211,6 +233,22 @@ def run
...
       self.precheck
 
+      scenario_answers = load_scenario_answers(@options[:scenario])
+
+      unless custom_cert_options_check(scenario_answers)
+        fail_with_message("You are currently using custom certificates but not all custom certificate arguments are present. To change the hostname, " \
+                          "you need to provide certificates with the new hostname. Please review the custom certificate arguments required with --help.")
+      end
+
+      if @using_custom_certs && !@foreman_proxy_content
```
ekohl: The interaction here a bit hard to follow. The above check can fail if the options aren't provided but here you exclude if for proxies. Does that mean we should also exclude it above?

katello/katello/hostname-change.rb
```diff
@@ -161,6 +161,16 @@ def get_default_proxy_id
       proxy["Id"]
     end
 
+    def custom_cert_options_check(scenario_answers)
+      if scenario_answers["certs"]["server_cert"] &&
+         scenario_answers["certs"]["server_key"] &&
+         scenario_answers["certs"]["server_cert_req"]
+        @using_custom_certs = true
```
ekohl: I don't like this side effect here. Maybe it should be ```custom_cert_options_check!```. It could also raise an exception if it's invalid.

katello/katello/helper.rb
```diff
@@ -36,6 +38,12 @@ def fail_with_message(message, opt_parser=nil)
       exit(false)
     end
 
+    def check_if_file_exists(file)
```
ekohl: This has a side effect so should this be ```check_if_file_exists!``` for clarity?

katello/katello/hostname-change.rb
```diff
@@ -211,6 +233,22 @@ def run
...
+
+      unless custom_cert_options_check(scenario_answers)
+        fail_with_message("You are currently using custom certificates but not all custom certificate arguments are present. To change the hostname, " \
+                          "you need to provide certificates with the new hostname. Please review the custom certificate arguments required with --help.")
+      end
+
+      if @using_custom_certs && !@foreman_proxy_content
+        STDOUT.puts "\nChecking custom certificates"
+        [@options[:custom_cert], @options[:custom_key], @options[:custom_cert_req]].each do |file|
+          check_if_file_exists(file)
```
ekohl: Would it be nice if we checked all files and failed with one message that some files don't exist?

johnpmitsch:
@ekohl @cfouant thanks for the suggestions, I updated a few things to improve the usability.
The certs check will now accept relative paths in addition to absolute paths. It also now tells all the files that are missing (rather than one at a time) here are some examples:

File name is correct, but I'm in the wrong directory to use absolute paths:
```
[root@customcertsupstream katello]# ./katello-change-hostname -u admin -p changeme -c /certs/changedupstream.example.com.crt -k /certs/changedupstream.example.com.key -r changedupstream.example.com.crt.req changed.example.com

Checking custom certificates
Error: File  /root/foreman-packaging/katello/katello/changedupstream.example.com.crt.req does not exist! Please check the file path and try again.
```

Same situation as above, but now two files:
 ```
[root@customcertsupstream katello]# ./katello-change-hostname -u admin -p changeme -c /certs/changedupstream.example.com.crt -k changedupstream.example.com.key -r changedupstream.example.com.crt.req changed.example.com

Checking custom certificates
Error: Files  /root/foreman-packaging/katello/katello/changedupstream.example.com.key, /root/foreman-packaging/katello/katello/changedupstream.example.com.crt.req do not exist! Please check the file path and try again.
```

Changed to the correct directory and now using relative paths (successful run):
```
[root@customcertsupstream certs]# ~/foreman-packaging/katello/katello/katello-change-hostname -u admin -p changeme -c changedupstream.example.com.crt -k changedupstream.example.com.key -r changedupstream.example.com.crt.req changed.example.com

Checking custom certificates

Checking hostname validity
```

File: katello/katello/hostname-change.rb [hunk 7]
```diff
@@ -209,6 +231,24 @@ def setup_opt_parser
...
+      if using_custom_certs?(scenario_answers)
+        unless all_custom_cert_options_present?(scenario_answers)
+          fail_with_message("You are currently using custom certificates but not all custom " \
+                            "certificate arguments are present. To change the hostname, " \
+                            "you need to provide certificates with the new hostname. Please " \
+                            "review the custom certificate arguments required with --help.")
+        end
+      end
+
+      if using_custom_certs?(scenario_answers) && !@foreman_proxy_content
```
ekohl: Can you elaborate why the certs only need to be checked in non-proxy context?

johnpmitsch: Yes, the foreman proxy hostname change requires a new certs-tarball whether the user is using custom certs or not. This tarball is passed into k-c-h as a single argument and custom certs are included in this tarball. 

Unfortunately there isn't a good way to check the certs in the tarball right now, we should make katello-certs-check work for proxies and check these sorts of things in the future.

cfouant:
ACK

PR #2077, opened 2018-01-12
Add rubygem-thor package

ekohl:


ehelms:
Why are we adding Thor here? https://github.com/theforeman/tfm-ror51-packaging/tree/master/packages/tfm-ror51-rubygem-thor

PR #2083, opened 2018-01-15
Update puppetdb_foreman to 4.0.0 [rpm]

timogoebel:
Please build for:

* [x] Nightly
* [x] 1.17


ekohl:
I'm not sure yet about the 1.17 branch - I think we need to do quite a few backports of bumps so for now I'm just going to include it in nightly. @ehelms can you tell me if we should treat 1.17 like we'd treat 1.16?

timogoebel:
@ekohl: Done.

PR #2087, opened 2018-01-15
Update foreman dependencies

ekohl:
This matches the dependencies to those in package.json and the various gems.

There is still one discrepancy in ruby2ruby. https://github.com/theforeman/foreman/commit/0c9747f85331c00b926de546937f0f3a6463b63a changed it to exactly 2.3.2 where we ship 2.1.3. The commit message implies 2.4.0 breaks it so I've included < 2.4.0 instead but this should be fixed. I'm surprised bundler even accepts this in the first place.

It includes https://github.com/theforeman/foreman-packaging/pull/2049 because that changes the markers and was used to update the package.json part of it.

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -64,23 +64,23 @@ Requires: %{name}-debug
 # Gemfile
 Requires: %{?scl_prefix_ror}rubygem(rails) >= 5.1.4
 Requires: %{?scl_prefix_ror}rubygem(rails) < 5.2
-Requires: %{?scl_prefix}rubygem(rest-client) >= 1.8
+Requires: %{?scl_prefix}rubygem(rest-client) >= 2.0
 Requires: %{?scl_prefix}rubygem(rest-client) < 3.0
-Requires: %{?scl_prefix}rubygem(audited) >= 4.3
+Requires: %{?scl_prefix}rubygem(audited) >= 4.6
 Requires: %{?scl_prefix}rubygem(audited) < 5.0
 Requires: %{?scl_prefix}rubygem(will_paginate) >= 3.0
 Requires: %{?scl_prefix}rubygem(will_paginate) < 4.0
 Requires: %{?scl_prefix}rubygem(ancestry) >= 2.0
 Requires: %{?scl_prefix}rubygem(ancestry) < 4.0
-Requires: %{?scl_prefix}rubygem(scoped_search) >= 4.0
+Requires: %{?scl_prefix}rubygem(scoped_search) >= 4.1.2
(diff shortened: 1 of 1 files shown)


ekohl:
Needs https://github.com/theforeman/foreman-packaging/pull/2097

foreman/foreman.spec
```diff
@@ -121,17 +125,17 @@ Requires: %{?scl_prefix}rubygem(x-editable-rails) >= 1.5.5
...
-Requires: %{?scl_prefix}rubygem(webpack-rails) >= 0.9.7
+Requires: %{?scl_prefix}rubygem(webpack-rails) >= 0.9.8
 Requires: %{?scl_prefix}rubygem(webpack-rails) < 1.0.0
-Requires: %{?scl_prefix_ror}rubygem(mail) >= 2.6
+Requires: %{?scl_prefix_ror}rubygem(mail) >= 2.7
 Requires: %{?scl_prefix_ror}rubygem(mail) < 3.0
 Requires: %{?scl_prefix}rubygem(sshkey) >= 1.9
 Requires: %{?scl_prefix}rubygem(sshkey) < 2.0
-Requires: %{?scl_prefix}rubygem(dynflow) >= 0.8.25
+Requires: %{?scl_prefix}rubygem(ruby2ruby) < 2.4.0
```
ekohl: https://github.com/theforeman/foreman/pull/5179 would make this match the Gemfile

foreman/foreman.spec
```diff
@@ -390,31 +401,29 @@ BuildRequires: npm(urijs) >= 1.18.10
...
+BuildRequires: %{?scl_prefix}rubygem(jquery-turbolinks) < 3.0
 BuildRequires: %{?scl_prefix}rubygem(jquery-ui-rails) < 5.0.0
-BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) >= 3.23.0
-BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) < 3.24.0
-BuildRequires: %{?scl_prefix}rubygem(spice-html5-rails) >= 0.1.5
-BuildRequires: %{?scl_prefix}rubygem(spice-html5-rails) < 0.2.0
-BuildRequires: %{?scl_prefix}rubygem(gettext_i18n_rails_js) >= 1.0
-BuildRequires: %{?scl_prefix}rubygem(gettext_i18n_rails_js) < 2.0
+BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) >= 3.32.1
+BuildRequires: %{?scl_prefix}rubygem(patternfly-sass) < 3.38.0
```
ekohl: @sharvit confirmed we can use 3.37.0 and will send a matching PR to core

PR #2097, opened 2018-01-16
Bump dependencies to more recent versions

ekohl:


ekohl:
Looks like foreman actually needs patternfly-sass < 3.33.0 so I bumped to too new a version.

PR #2107, opened 2018-01-17
Refs #22294 - Set foreman-root for dynflowd and replace shebang with SCL

ehelms:

* [x] Nightly
* [x] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/foreman/foreman.spec
+++ b/foreman/foreman.spec
@@ -875,7 +875,7 @@ plugins required for Foreman to work.
 #replace shebangs and binaries in scripts for SCL
 %if %{?scl:1}%{!?scl:0}
   # shebangs
-  for f in bin/* script/performance/profiler script/performance/benchmarker script/foreman-config ; do
+  for f in bin/* script/performance/profiler script/performance/benchmarker script/foreman-config script/dynflowd ; do
     sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X%{scl_ruby_bin}X' $f
   done
   sed -ri '1,$sX/usr/bin/rubyX%{scl_ruby_bin}X' %{SOURCE1}


ehelms:
Also needs a code change -- https://github.com/theforeman/foreman/pull/5180

ekohl:
I expected this not to be needed because of the WorkingDirectory statement. Also makes me wonder what of the sysconfig file is actually used. Current best practices for systemd is not to use them if you can avoid it IIRC.

ekohl:
[test]

mmoll:
is a similar change for DEBs needed?

ehelms:
@ekohl If you look at the script, [1] this sets the rails_root to `Dir.pwd` which is passed along to Dynflow. If this is not set properly to the location of Foreman it breaks. The working directory statement is to prevent [2] the requires from failing by not being at the location of Foreman's root.



[1] https://github.com/theforeman/foreman/pull/5180/files#diff-4ad43f7794b9a87ae3c0c3a40782c555R9
[2] https://github.com/theforeman/foreman/pull/5180/files#diff-4ad43f7794b9a87ae3c0c3a40782c555R69

ehelms:
@mmoll if the conclusion is this is correct in this PR, I'd think Debian also needs it. We'd know because the Deb pipeline should be failing.

ekohl:
> @ekohl If you look at the script, [1] this sets the rails_root to Dir.pwd which is passed along to Dynflow. If this is not set properly to the location of Foreman it breaks. The working directory statement is to prevent [2] the requires from failing by not being at the location of Foreman's root.

I thought that since the CWD is already set to the foreman home it shouldn't be an isuse:
https://github.com/theforeman/foreman-packaging/blob/e34daec4af994810e56c112218cc478264f2d1d1/foreman/dynflowd.service#L10

When I tested the dynflow PR it could start up the service for me so I'm not sure why this is actually needed. Maybe some regression sneaked in between my testing and the actual merge?

ehelms:
The change to the service file was not actually needed, I crossed my wires testing the script and the service. The change to the script itself that went into Foreman, and this change to the specfile to ensure it uses the SCL ruby is all thats needed.

@mmoll since Debian does not use the SCL, it should not needed this change now

PR #2110, opened 2018-01-18
Fixes #21517 - Backup last_scenario.yaml on hostname change

jturel:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

This makes sure we can run katello-change-hostname consecutively especially in cases where something goes wrong.

### DIFF
--- a/katello/katello/helper.rb
+++ b/katello/katello/helper.rb
@@ -1,12 +1,19 @@
 module KatelloUtilities
   module Helper
     def last_scenario
-      last_scenario_yaml = '/etc/foreman-installer/scenarios.d/last_scenario.yaml'
-      if File.exists?(last_scenario_yaml)
+      if File.exist?(last_scenario_yaml)
         File.basename(File.readlink(last_scenario_yaml)).split(".")[0]
       else
         'katello'
       end
+    end
+
+    def scenarios_path
+      '/etc/foreman-installer/scenarios.d'
+    end
+
+    def last_scenario_yaml
+      "#{scenarios_path}/last_scenario.yaml"
     end
 
     def accepted_scenarios
@@ -31,7 +38,7 @@ module KatelloUtilities
     end
 
     def fail_with_message(message, opt_parser=nil)
-      STDOUT.puts message
+      STDOUT.puts message.red
       puts opt_parser if opt_parser
       exit(false)
     end
@@ -45,23 +52,49 @@ module KatelloUtilities
       end
     end
 
-    def run_cmd(command, exit_codes=[0], message=nil)
+    def run_cmd(command, exit_codes=[0], message=nil, &block)
       result = `#{command}`
-      unless exit_codes.include?($?.exitstatus)
+      exit_code = $?.exitstatus
+      success = $?.success?
+
+      yield(result, success) if block_given?
+
+      unless exit_codes.include?(exit_code)
         STDOUT.puts result
-        STDOUT.puts message if message
-        failed_command = "Failed '#{command}' with exit code #{$?.exitstatus}"
+        STDOUT.puts message.red if message
+        failed_command = "Failed '#{command}' with exit code #{exit_code}"
         if self.respond_to? :cleanup
           STDOUT.puts failed_command
-          cleanup($?.exitstatus)
+          cleanup(exit_code)
         end
         fail_with_message(failed_command)
       end
+
       result
     end
 
     def timestamp
       DateTime.now.strftime('%Y%m%d%H%M%S')
+    end
+
+    class ::String
+      def red
+        colorize(31)
+      end
+
+      def green
+        colorize(32)
+      end
+
+      def gray
+        colorize(37)
+      end
+
+      private
+
+      def colorize(color_code)
+        "\e[#{color_code}m#{self}\e[0m"
+      end
     end
   end
 end
--- a/katello/katello/hostname-change.rb
+++ b/katello/katello/hostname-change.rb
@@ -7,11 +7,14 @@ require "json"
 require "uri"
 require "fileutils"
 require "highline/import"
+require "tempfile"
 require_relative "helper.rb"
 
 module KatelloUtilities
   class HostnameChange
     include ::KatelloUtilities::Helper
+
+    attr_accessor :temp_last_scenario_yaml
 
     def initialize(init_options)
       @default_program = self.get_default_program
@@ -102,9 +105,12 @@ module KatelloUtilities
       end
     end
 
-    def successful_hostname_change_message
-    # the following multi-line string isn't indented because the indents are taken literally.
-    successful_message = %(
+    def next_steps_message
+      if @foreman_proxy_content
+        "You will have to update the Name and URL of the Smart Proxy in #{@options[:program].capitalize} to the new hostname.\n"
+      else
+        # the following multi-line string isn't indented because the indents are taken literally.
+%(
   You will have to install the new bootstrap rpm and reregister all clients and #{@plural_proxy} with subscription-manager
   (update organization and environment arguments appropriately):
 
@@ -125,15 +131,8 @@ module KatelloUtilities
                                   --foreman-proxy-trusted-hosts #{@new_hostname}
 
   Short hostnames have not been updated, please update those manually.\n
-    )
-      STDOUT.puts "**** Hostname change complete! **** \nIMPORTANT:"
-      if @foreman_proxy_content
-        STDOUT.print "You will have to update the Name and URL of the Smart Proxy in #{@options[:program].capitalize} to the new hostname.\n"
-      else
-        STDOUT.print successful_message
+)
       end
-      STDOUT.print ""
-      exit(true)
     end
 
     def warning_message
@@ -245,7 +244,6 @@ module KatelloUtilities
     def run
       raise 'Must run as root' unless Process.uid == 0
 
-
       if using_custom_certs?(@scenario_answers)
         custom_certs_check(@scenario_answers)
       end
@@ -343,14 +341,24 @@ module KatelloUtilities
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' /etc/hosts")
 
       STDOUT.puts "updating hostname in foreman installer scenarios"
-      self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' /etc/foreman-installer/scenarios.d/*.yaml")
+      self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' #{scenarios_path}/*.yaml")
 
-      STDOUT.puts "removing last_scenario.yml file"
-      self.run_cmd("rm -rf /etc/foreman-installer/scenarios.d/last_scenario.yaml")
+      if File.exist?(last_scenario_yaml)
+        STDOUT.puts 'backing up last_scenario.yaml'
+        temp_last_scenario_yaml = Tempfile.new('last_scenario')
+        begin
+          temp_last_scenario_yaml << File.read(last_scenario_yaml)
+        ensure
+          temp_last_scenario_yaml.close
+        end
+
+        STDOUT.puts 'removing last_scenario.yaml'
+        File.unlink(last_scenario_yaml)
+      end
 
       STDOUT.puts "re-running the installer"
 
-      installer = "#{@options[:program]}-installer --scenario #{@options[:scenario]} -v"
+      installer = default_installer
       if @foreman_proxy_content
         installer << fpc_installer_args
       else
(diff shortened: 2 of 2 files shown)


katello/katello/hostname-change.rb
```diff
@@ -366,6 +369,10 @@ def run
       if installer_success
         self.successful_hostname_change_message
       else
+
```
johnpmitsch: nitpick, I dont think you need this whitespace here

katello/katello/hostname-change.rb
```diff
@@ -331,8 +331,11 @@ def run
       STDOUT.puts "updating hostname in foreman installer scenarios"
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' /etc/foreman-installer/scenarios.d/*.yaml")
 
-      STDOUT.puts "removing last_scenario.yml file"
-      self.run_cmd("rm -rf /etc/foreman-installer/scenarios.d/last_scenario.yaml")
+      STDOUT.puts "backing up last_scenario.yaml"
+      self.run_cmd("cp #{last_scenario_yaml} /var/tmp")
```
ekohl: I'd prefer some ```mktemp``` based solution rather than static ```/var/tmp```. Also a check if the last_scenario file actually exists might be a good one, even if the chance is of issues is low.

katello/katello/hostname-change.rb
```diff
@@ -369,18 +377,26 @@ def run
...
       STDOUT.puts installer
-      installer_output = self.run_cmd("#{installer}")
-      installer_success = $?.success?
-      STDOUT.puts installer_output
-
-      STDOUT.puts "Restarting puppet services"
-      self.run_cmd("/sbin/service puppet restart")
-      self.run_cmd("katello-service restart --only puppetserver")
+      default_installer << ' --disable-system-checks' if disable_system_check_option?
+      failure_message = "Something went wrong with the #{@options[:scenario].capitalize} installer. Please check the above output and the corresponding logs. Once fixed you may complete the hostname change by running \"#{default_installer}\""
```
johnpmitsch: plz breakup this line (rubocop is coming soon, I swear!)

katello/katello/helper.rb
```diff
@@ -45,18 +48,24 @@ def fail_if_file_not_found(files)
       end
     end
 
-    def run_cmd(command, exit_codes=[0], message=nil)
+    def run_cmd(command, exit_codes=[0], message=nil, &block)
```
johnpmitsch: `&block` == genius idea :nerd_face:

katello/katello/hostname-change.rb
johnpmitsch: Can we change "Once fixed" to " Once the issue is resolved"?

johnpmitsch:
the functionality works well!

katello/katello/hostname-change.rb
```diff
@@ -345,12 +346,19 @@ def run
       STDOUT.puts "updating hostname in foreman installer scenarios"
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' /etc/foreman-installer/scenarios.d/*.yaml")
 
-      STDOUT.puts "removing last_scenario.yml file"
-      self.run_cmd("rm -rf /etc/foreman-installer/scenarios.d/last_scenario.yaml")
+      if File.exist?(last_scenario_yaml)
+        temp_dir = Dir.mktmpdir
+        STDOUT.puts 'backing up last_scenario.yaml'
+        run_cmd("cp #{last_scenario_yaml} #{temp_dir}")
```
ekohl: Isn't cp + rm equivalent to mv?

katello/katello/hostname-change.rb
```diff
@@ -345,12 +346,19 @@ def run
       STDOUT.puts "updating hostname in foreman installer scenarios"
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' /etc/foreman-installer/scenarios.d/*.yaml")
 
-      STDOUT.puts "removing last_scenario.yml file"
-      self.run_cmd("rm -rf /etc/foreman-installer/scenarios.d/last_scenario.yaml")
+      if File.exist?(last_scenario_yaml)
+        temp_dir = Dir.mktmpdir
```
ekohl: Why a temp dir rather than a temp file? Another thing to be careful about is that the answers file contains passwords and other secrets. It should not become world readable.

katello/katello/hostname-change.rb
```diff
@@ -369,18 +377,26 @@ def run
...
+        end
 
-      if installer_success
-        self.successful_hostname_change_message
-      else
-        self.fail_with_message("Something went wrong with the #{@options[:scenario].capitalize} installer! Please check the above output and the corresponding logs")
+        if success
+          STDOUT.puts result
+          STDOUT.puts 'Restarting puppet services'
+          run_cmd('/sbin/service puppet restart')
```
ekohl: service is old style. systemctl is the current way to go. Does EL7 already support the ```try-reload-or-restart``` since I think that's what we want here: prefer reload but otherwise restart only if it's running.

katello/katello/hostname-change.rb
```diff
@@ -369,18 +377,26 @@ def run
...
+          run_cmd('katello-service restart --only puppetserver')
```
ekohl: How does katello-service work if the service is not running? Would plain ```systemctl try-reload-or-restart``` be preferred?

katello/katello/hostname-change.rb
johnpmitsch: ya we can do systemctl here, `/sbin/service` is leftover from RHEL6 support. I'm fine with just using restart, but don't know much about this `try-reload-or-restart` option (sound great if its supported)

katello/katello/hostname-change.rb
jturel: last_scenario_yaml is a symlink and it seems mv breaks it

katello/katello/hostname-change.rb
jturel: katello-service is not actually a registered service but a script in /usr/sbin

@johnpmitsch what if we did `systemctl try-reload-or-restart puppetserver` instead of using katello-service?

katello/katello/hostname-change.rb
jturel: Can use a tempfile & the permissions look good using that method

katello/katello/hostname-change.rb
```diff
@@ -369,18 +381,30 @@ def run
...
 
-      if installer_success
-        self.successful_hostname_change_message
-      else
-        self.fail_with_message("Something went wrong with the #{@options[:scenario].capitalize} installer! Please check the above output and the corresponding logs")
+        if success
+          STDOUT.puts result
+          STDOUT.puts 'Restarting puppet services'
+          run_cmd('systemctl reload-or-try-restart puppet')
+          run_cmd('katello-service restart --only puppetserver')
```
jturel: Maybe we can do systemctl reload-or-try-restart puppetserver here instead of using katello-service..

katello/katello/helper.rb
```diff
@@ -1,14 +1,17 @@
 module KatelloUtilities
   module Helper
     def last_scenario
-      last_scenario_yaml = '/etc/foreman-installer/scenarios.d/last_scenario.yaml'
-      if File.exists?(last_scenario_yaml)
+      if File.exist?(last_scenario_yaml)
```
ekohl: It looks like we mostly use ```File.exists``` and I think that's also easier to read

katello/katello/hostname-change.rb
```diff
@@ -345,12 +346,23 @@ def run
       STDOUT.puts "updating hostname in foreman installer scenarios"
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' /etc/foreman-installer/scenarios.d/*.yaml")
 
-      STDOUT.puts "removing last_scenario.yml file"
-      self.run_cmd("rm -rf /etc/foreman-installer/scenarios.d/last_scenario.yaml")
+      if File.exist?(last_scenario_yaml)
```
ekohl: I'd also prefer ```File.exists``` here.

katello/katello/hostname-change.rb
```diff
@@ -369,18 +381,30 @@ def run
...
-      self.run_cmd("katello-service restart --only puppetserver")
+      default_installer << ' --disable-system-checks' if disable_system_check_option?
+      failure_message = """
+Something went wrong with the #{@options[:scenario].capitalize} installer.
+Please check the above output and the corresponding logs. Once the issue is resolved
+you may complete the hostname change by running #{default_installer}
+      """
+
+      run_cmd(installer, [0], failure_message) do |result, success|
+        if temp_last_scenario_yaml && temp_last_scenario_yaml.path
```
ekohl: Wouldn't this variable be undefined if the file didn't exist?

katello/katello/hostname-change.rb
```diff
@@ -369,18 +381,30 @@ def run
...
+        end
 
-      if installer_success
-        self.successful_hostname_change_message
-      else
-        self.fail_with_message("Something went wrong with the #{@options[:scenario].capitalize} installer! Please check the above output and the corresponding logs")
+        if success
+          STDOUT.puts result
+          STDOUT.puts 'Restarting puppet services'
+          run_cmd('systemctl reload-or-try-restart puppet')
```
ekohl: This should be try-reload-or-restart:
```
  try-reload-or-restart NAME...   If active, reload one or more units,
                                  if supported, otherwise restart
```

If you want to restart both puppet and puppetserver then you can even do it at once with systemctl (and in parallel):
```systemctl try-reload-or-restart puppet puppetserver```

File: katello/katello/helper.rb (hunk 10 above)
Comment (jturel) [reply katello/katello/helper.rb]: I think so too - but this is what rubocop prefers. But if that's more consistent (we are overriding that cop elsewhere?) then I will change it

File: katello/katello/hostname-change.rb (hunk 12 above)
Comment (jturel) [reply katello/katello/hostname-change.rb]: Correct. When a tempfile is issued `.unlink` all traces are removed and it will return nil which is what I am checking for here. If it responds to path (which we need) then we know it hasn't been unlinked.

File: katello/katello/hostname-change.rb (hunk 13 above)
Comment (jturel) [reply katello/katello/hostname-change.rb]: Should it? I do not see that option on my EL7 nor my Centos box. It also looks like they are equivalent methods in any case:

https://github.com/systemd/systemd/blob/3e52ec43011f0b7f72c5bf43ce60fa855d1a0ee2/src/systemctl/systemctl.c#L2784-L2785

File: katello/katello/helper.rb (hunk 10 above)
Comment (jturel) [reply katello/katello/helper.rb]: Actually it looks like foreman proper uses `File.exist?` which is rubocop preferred as I mentioned. I'm leaning toward leaving it as-is.

File: katello/katello/hostname-change.rb (hunk 13 above)
Comment (ekohl) [reply katello/katello/hostname-change.rb]: That does appear to be a difference between Fedora and CentOS. If this works on EL7 then let's go with this.

File: katello/katello/helper.rb (hunk 10 above)
Comment (johnpmitsch) [reply katello/katello/helper.rb]: yes, whatever gets the script more prepared for rubocop has my vote

File: katello/katello/hostname-change.rb (hunk 9 above)
Comment (johnpmitsch) [reply katello/katello/hostname-change.rb]: I was told to use katello-service for any services supported by it, since that will restart them in the way we specify. Is there something puppet-specific that would benefit from `reload-or-try-restart`?

File: katello/katello/helper.rb (hunk 10 above)
Comment (ekohl) [reply katello/katello/helper.rb]: I prefer as little change as possible since we should really get rid of all these files from the packaging repository. They were accepted on the premise that there would be an effort to move them to another repository. I'd also be against huge changes like https://github.com/theforeman/foreman-packaging/pull/2115

Comment (jturel) [reply katello/katello/helper.rb]: You mentioned preferring as little change as possible, yet in another comment are asking me to change service restart commands which were simply copy/paste as I refactored the code - so I'm getting a mixed message. `File.exist?`seems most appropriate in terms of consistency and rubocop's needs. Let's keep it!

**Comment by jturel** [comment]
I think the issues have been addressed. Please have a look now

File: katello/katello/helper.rb (hunk 10 above)
Comment (ekohl) [reply katello/katello/helper.rb]: You're right, I didn't realize that the restart code was already there and thought it was a new line. In that case I'm fine with keeping the current restart code.

**Comment by johnpmitsch** [comment]
Discussed offline, looks like `last_scenario.yml` needs to be moved back as a symlink to be able to run the installer again. Currently fails with
```
[ INFO 2018-01-22 18:46:18 verbose] Executing hooks in group pre_migrations
[ INFO 2018-01-22 18:46:18 verbose] All hooks in group pre_migrations finished
[ERROR 2018-01-22 18:46:18 verbose] You are trying to replace existing installation with different scenario. This may lead to unpredictable states. Use --force to override. You can use --compare-scenarios to see the differences
[ INFO 2018-01-22 18:46:18 verbose] Log was be written to /var/log/foreman-installer/katello.log
[ INFO 2018-01-22 18:46:18 verbose] Logs flushed
```

**Comment by jturel** [comment]
@johnpmitsch the symlink will now be preserved if the installer fails early - give it a whirl

File: katello/katello/hostname-change.rb [hunk 14]
```diff
@@ -102,9 +103,12 @@ def precheck
       end
     end
 
-    def successful_hostname_change_message
-    # the following multi-line string isn't indented because the indents are taken literally.
-    successful_message = %(
+    def next_steps_message
+      if @foreman_proxy_content
+        "You will have to update the Name and URL of the Smart Proxy in #{@options[:program].capitalize} to the new hostname.\n"
```
ekohl: Idea for future improvement: do this via the API

katello/katello/hostname-change.rb
```diff
@@ -343,14 +339,24 @@ def run
...
+        begin
+          temp_last_scenario_yaml << File.read(last_scenario_yaml)
+        ensure
+          temp_last_scenario_yaml.close
+        end
 
-      STDOUT.puts "removing last_scenario.yml file"
-      self.run_cmd("rm -rf /etc/foreman-installer/scenarios.d/last_scenario.yaml")
+        STDOUT.puts 'removing last_scenario.yaml'
+        run_cmd("rm -f #{last_scenario_yaml}")
```
ekohl: Isn't this equal to ```File.unlink```?

katello/katello/hostname-change.rb
```diff
@@ -363,25 +369,63 @@ def run
...
       STDOUT.puts installer
-      installer_output = self.run_cmd("#{installer}")
-      installer_success = $?.success?
-      STDOUT.puts installer_output
-
-      STDOUT.puts "Restarting puppet services"
-      self.run_cmd("/sbin/service puppet restart")
-      self.run_cmd("katello-service restart --only puppetserver")
+      run_cmd(installer, [0], installer_failure_message) do |result, success|
+        if temp_last_scenario_yaml && temp_last_scenario_yaml.path
```
ekohl: Won't this throw a ```NameError``` in the case where the file didn't exist? temp_last_scenario_yaml is only assigned inside the if block.

katello/katello/hostname-change.rb
jturel: If last_scenario is always a symlink then I'll definitely change this!

katello/katello/hostname-change.rb
jturel: `temp_last_scenario_yaml` will be nil and the `temp_last_scenario_yaml.path` will not be executed if I am understanding your question so it should be good in that scenario

ekohl: I'm talking about:
https://github.com/jturel/foreman-packaging/blob/741cdd39f53af493dc9e99cf88ec9071ceb1f8bc/katello/katello/hostname-change.rb#L344-L355

If false, then the variable isn't nil but actually unset I think.

jturel: Ah, yes. I'll adjust that!

katello/katello/hostname-change.rb
```diff
@@ -7,12 +7,15 @@
...
 require "fileutils"
 require "highline/import"
+require "tempfile"
 require_relative "helper.rb"
 
 module KatelloUtilities
   class HostnameChange
     include ::KatelloUtilities::Helper
 
+    attr_accessor :temp_last_scenario_yaml
```
jturel: @ekohl Now we won't get a NameError

PR #2112, opened 2018-01-18
Updated dhcp_remote_isc to 0.0.4

witlessbird:

* [x] Nightly
* [x] 1.17
* [x] 1.16
* [ ] 1.15

---


witlessbird:
> It also looks like you added rsec to the requirements but only as development dependencies. The code implies it should be a runtime dependency. Is that correct? How about concurrent-ruby? Is that also supposed to be a runtime dependency?

Both rsec and concurrent-ruby are smart-roxy dependencies and provided by it. These only need to be installed to be able to run tests.

ekohl:
> Both rsec and concurrent-ruby are smart-proxy dependencies, provided by smart-proxy. These only need to be installed to be able to run tests.

Ok. If you just bump the foreman-proxy requirement to 1.16 then it should be good to go.

witlessbird:
> Ok. If you just bump the foreman-proxy requirement to 1.16 then it should be good to go.

done.

ekohl:
Thanks!

PR #2116, opened 2018-01-22
Replace bundled-npm() with bundled(npm())

ekohl:
Based on a suggestion in https://github.com/theforeman/foreman-packaging/pull/1827 by @xsuchy


ekohl:
Not sure why that nodejs-node-sass failed on git annex get.

PR #2117, opened 2018-01-22
Chery pick plugin updates to 1.17

ekohl:
I've tried to carefully select everything that's in rpm/develop and left out what's not needed because they don't fit foreman 1.17 dependencies. The katello changes are all left out because I wasn't sure about the status of them.


ekohl:
Looks like it can't build foreman 1.17 (since there's no tag yet) and then everything that relies on 1.17+ fails. Not sure how to fix this. Maybe a manual 1.17-nightly build?

ekohl:
Split off the first batch in https://github.com/theforeman/foreman-packaging/pull/2125

PR #2123, opened 2018-01-23
Fixes #22376 - Use TMPDIR instead own temporary directory

lzap:
@iNecas counterpart for https://github.com/theforeman/foreman/pull/5040

* [x] Nightly

---

### DIFF
--- a/katello/katello/katello-debug.sh
+++ b/katello/katello/katello-debug.sh
@@ -13,6 +13,8 @@ then
   exit 1
 fi
 
+TEMP_DIR=$(mktemp -d) # no trap - foreman-debug cleans automatically
+
 # foreman-debug will truncate any file beyond fixed size limit,
 # for larger files we need to copy the entire file.
 copy_files() {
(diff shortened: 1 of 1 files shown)


ekohl:
For me $AUXTMP is unset. I've never heard of that var either. How will this work?

lzap:
@ekohl have you tried with the patch I reference from the description?

https://github.com/theforeman/foreman/pull/5040/files#diff-530ce7cdbd9b2169eb6a5d364c87b049R264

ekohl:
Wouldn't it be better to just set ```TMPDIR``` since that's the convention? Tools like mktemp should also honor that if the script was ever expanded.

iNecas:
I wasn't aware of the `$TMPDIR` behaviour, but looking at that, it really looks like the most suitable approach.

lzap:
After discussion I am pushing new version that works via TMPDIR. I haven't tested on real setup, but I can confirm it should work.

```
[root@zzzap ~]# mktemp -d
/tmp/tmp.c8UjxXHR8U
[root@zzzap ~]# export TMPDIR=/tmp/test
[root@zzzap ~]# mktemp -d
/tmp/test/tmp.NJn24Ky5rt
```

PR #2125, opened 2018-01-23
Chery pick changes to 1.17 part 1

ekohl:
This leaves out the plugins that currently still fail to build because there is no 1.17 build yet.


ekohl:
The following commits were tagged but not actually released yet:
6a3792104dd12b9f6988e2760e2ad01bc8b91358 
f3b83bbeedb1d3e39071a1aa247c2519332d837a
aa8a1130cc17998a0c2fee37db830f3d50dc3697
8337d913b6c36106c30dd49e89c58ed1a19885b4
c0d9b8947f521d09c37207c1e28fe31156b0f999
0f5d75a371095c26d4adda71e8b492a7de931d99

These are all foreman or smart proxy plugins and fail because of a tag mismatch. Possibly because there is no foreman 1.17 build.

PR #2143, opened 2018-01-30
symlink local_secret_token.rb to /etc/foreman

sean797:
Currently we have:

```
[root@centos7-foreman ~]# ls -l /etc/foreman
total 20
-rw-r-----. 1 root foreman  684 Jan 16 16:27 database.yml
-rw-r-----. 1 root foreman  450 Jan 15 22:22 encryption_key.rb
-rw-r--r--. 1 root root     691 Nov 29 07:12 foreman-debug.conf
-rw-r--r--. 1 root root    2564 Nov 29 07:12 logging.yaml
drwxr-xr-x. 2 root root      72 Jan 16 16:28 plugins
-rw-r-----. 1 root foreman 1215 Jan 17 13:04 settings.yaml
```

It'd would be nice to also have `local_secret_token.rb`

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -1151,6 +1151,7 @@ rm -rf %{buildroot}
 %{_mandir}/man8
 %config(noreplace) %{_sysconfdir}/%{name}
 %ghost %attr(0640,root,%{name}) %config(noreplace) %{_sysconfdir}/%{name}/encryption_key.rb
+%ghost %attr(0640,root,%{name}) %config(noreplace) %{_sysconfdir}/%{name}/local_secret_token.rb
 %config(noreplace) %{_sysconfdir}/sysconfig/%{name}
 %config(noreplace) %{_sysconfdir}/logrotate.d/%{name}
 %config %{_sysconfdir}/cron.d/%{name}
@@ -1162,7 +1163,7 @@ rm -rf %{buildroot}
 %attr(-,%{name},root) %{_datadir}/%{name}/config.ru
 %attr(-,%{name},root) %{_datadir}/%{name}/config/environment.rb
 %ghost %{_datadir}/%{name}/config/initializers/encryption_key.rb
-%ghost %attr(0640,root,%{name}) %config(noreplace) %{_datadir}/%{name}/config/initializers/local_secret_token.rb
+%ghost %{_datadir}/%{name}/config/initializers/local_secret_token.rb
 %{_tmpfilesdir}/%{name}.conf
 
 # Service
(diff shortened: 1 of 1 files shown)


sean797:
Thanks @ekohl, I've added a migration 

> Should the current ghost in the datadir be removed?

I don't think so, `encryption_key.rb` still has ghost

sean797:
rebased

packages/foreman/foreman/foreman.spec
```diff
@@ -1144,6 +1145,17 @@ if [ ! -f %{_datadir}/%{name}/config/initializers/local_secret_token.rb ]; then
   chgrp foreman %{_datadir}/%{name}/config/initializers/local_secret_token.rb
   %{foreman_rake} security:generate_token >/dev/null 2>&1 || :
   chmod 0640 %{_datadir}/%{name}/config/initializers/local_secret_token.rb
+  mv %{_datadir}/%{name}/config/initializers/local_secret_token.rb %{_sysconfdir}/%{name}/
```
ekohl: If we ensure the symlink is in place before running the token generation, would that also ensure it's generated in the right place? Ideally that means we can ship a symlink inside the package. We'd need to ensure the current token isn't overwritten though.

sean797:
hows that @ekohl ?

ekohl:
Apologies for the long delay. In https://github.com/theforeman/foreman-packaging/commit/b0bc45ff818e4da6ca6f5ee4517cd526f51ccbe1 I implemented a similar change. By going the `%pre` route, you can assume the state is correct in `%install` and it greatly simplifies things. On the other hand, I don't know if the rake task will follow the symlink. Probably not.

Below is my explicit thought process so you can check if I've missed anything.

* The `ln -s` in the first if block should be removed because the rake task would overwrite it (assumption, but safe and simplifies things for later).
* A separate if block that moves it to /etc. This would handle the existing installations as well as just after the generation.
* Another if block that creates the symlink.

So:
```sh
if [ ! -e %{_datadir}/%{name}/config/initializers/local_secret_token.rb ] ; then
  touch %{_datadir}/%{name}/config/initializers/local_secret_token.rb
  chmod 0660 %{_datadir}/%{name}/config/initializers/local_secret_token.rb
  chgrp foreman %{_datadir}/%{name}/config/initializers/local_secret_token.rb
  %{foreman_rake} security:generate_token >/dev/null 2>&1 || :
  chmod 0640 %{_datadir}/%{name}/config/initializers/local_secret_token.rb
fi
if [ -f %{_datadir}/%{name}/config/initializers/local_secret_token.rb ] ; then
  mv %{_datadir}/%{name}/config/initializers/local_secret_token.rb %{_sysconfdir}/%{name}/
fi
if [ ! -e %{_datadir}/%{name}/config/initializers/local_secret_token.rb -a \
    -e %{_sysconfdir}/%{name}/local_secret_token.rb ]; then
  ln -s %{_sysconfdir}/%{name}/local_secret_token.rb %{_datadir}/%{name}/config/initializers/
fi
```

ekohl:
https://projects.theforeman.org/issues/16508 is also interesting to consider in this context but I'd be happy with just implementing this now.

sean797:
Yup -  thanks does make sense, thanks @ekohl :+1:

packages/foreman/foreman/foreman.spec
```diff
@@ -1169,13 +1170,20 @@ exit 0
...
 # secret token used for cookie signing etc.
-if [ ! -f %{_datadir}/%{name}/config/initializers/local_secret_token.rb ]; then
+if [ ! -e %{_datadir}/%{name}/config/initializers/local_secret_token.rb ] ; then
   touch %{_datadir}/%{name}/config/initializers/local_secret_token.rb
   chmod 0660 %{_datadir}/%{name}/config/initializers/local_secret_token.rb
   chgrp foreman %{_datadir}/%{name}/config/initializers/local_secret_token.rb
   %{foreman_rake} security:generate_token >/dev/null 2>&1 || :
   chmod 0640 %{_datadir}/%{name}/config/initializers/local_secret_token.rb
 fi
+if [ -f %{_datadir}/%{name}/config/initializers/local_secret_token.rb ] ; then
```
ekohl: In 796febe4203eb87f03f24a4d6a223093a1f2f786 I fixed an oversight I had: `-f` dereferences symlinks.

sean797:
Thanks for the reminder, updated..

evgeni:
So I think the only thing I don't like here is that the token is still generated in datadir and then moved to sysconfdir, but such is life.

ACK

ekohl:
Thanks!

ekohl:
Looks like this caused an issue with building. From http://koji.katello.org/kojifiles/work/tasks/8811/98811/build.log:
```
LoadError: cannot load such file -- /builddir/build/BUILD/katello-3.7.0/usr/share/foreman/config/initializers/local_secret_token.rb
```

PR #2160, opened 2018-02-01
fixes #21795 - limits www backup content to certs and rpms

cfouant:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/katello/katello/backup.rb
+++ b/katello/katello/backup.rb
@@ -125,6 +125,13 @@ module KatelloUtilities
       backup_configs + feature_configs
     end
 
+    def excluded_configs
+      excluded_configs = []
+      excluded_configs.push("--exclude /var/www/html/pub/isos*") if Dir.exist?('/var/www/html/pub/isos')
+      excluded_configs.push("--exclude /var/www/html/pub/exports*") if Dir.exist?('/var/www/html/pub/exports')
+      excluded_configs
+    end
+
     def confirm
       unless agree("WARNING: This script will stop your services. Do you want to proceed(y/n)? ")
         puts "**** cancelled ****"
@@ -329,7 +336,7 @@ module KatelloUtilities
 
     def backup_config_files
       puts "Backing up config files... "
-      run_cmd("tar --selinux --create --gzip --file=#{File.join(@dir, 'config_files.tar.gz')} --listed-incremental=#{File.join(@dir, '.config.snar')} #{configure_configs.join(' ')} 2>/dev/null", [0,2])
+      run_cmd("tar --selinux --create --gzip --file=#{File.join(@dir, 'config_files.tar.gz')} --listed-incremental=#{File.join(@dir, '.config.snar')} #{configure_configs.join(' ')} #{excluded_configs.join(' ')} 2>/dev/null", [0,2])
       puts "Done."
     end
 


johnpmitsch:
Looks fine to me, ACK from me pending on response of this question:
@ehelms is there anything else we need to backup in `/var/www/html/pub/` besides `.crt` and `.rpm` files?

cfouant:
[test default]

cfouant:
[test]

sean797:
`bootstrap.py` & `katello-rhsm-consumer` are needed also.

I personally think we should backup everything but `exports` & `isos`. Keeping this white-list up-to-date is unlikely to happen in the future.

chris1984:
@sean797 for this PR I think we should move forward with how it is, as I have been a victim to an 8 GB config file when in support. I am not opposed to a refactor down the road but for now, this will fix the issue at hand.

johnpmitsch:
@sean797 that is a good point. I am in favor or blacklisting excluded directories (as suggested) instead of whitelisting files. This could be done with the `--exclude` flag for tar. With a whitelist, It is too easy to forget to update the list of files if we do decide to add more to that directory. It seems like excluding `exports` and `isos` would cover most cases of unnecessary large backups

cfouant:
@johnpmitsch - I think I've done this correctly, but if it needs more tweaking, you'll need to take it back since I'm going on leave. If I recall correctly, `--exclude` works on files, but not directories, with the exception of wildcards. So if `exports` and `isos` are simply directories we want to exclude, adding `--exclude /var/www/html/pub/exports* --exclude /var/www/html/pub/isos*` should work. I'm simply unsure If the `exports` and `isos` directories are in `/var/www/html/pub/` or nested further down in the path. If the latter, you'll need to finish the fix, thanks!

katello/katello/backup.rb
```diff
@@ -329,7 +336,7 @@ def create_pulp_data_tar
 
     def backup_config_files
       puts "Backing up config files... "
-      run_cmd("tar --selinux --create --gzip --file=#{File.join(@dir, 'config_files.tar.gz')} --listed-incremental=#{File.join(@dir, '.config.snar')} #{configure_configs.join(' ')} 2>/dev/null", [0,2])
+      run_cmd("tar --selinux --create --gzip --file=#{File.join(@dir, 'config_files.tar.gz')} --listed-incremental=#{File.join(@dir, '.config.snar')} #{configure_configs.join(' ')} #{excluded_configs.join(' ')} 2>/dev/null", [0,2])
```
sean797: This won't work, `--exclude` needs to be positioned before listing the files or directories you want to backup.

katello/katello/backup.rb
```diff
@@ -125,6 +125,13 @@ def configure_configs
       backup_configs + feature_configs
     end
 
+    def excluded_configs
+      excluded_configs = []
+      excluded_configs.push("--exclude /var/www/html/pub/isos*") if Dir.exist?('/var/www/html/pub/isos')
+      excluded_configs.push("--exclude /var/www/html/pub/exports*") if Dir.exist?('/var/www/html/pub/exports')
```
sean797: These `if Dir.exist?` are completely pointless. `--exclude` matches a pattern, if a pattern isn't found it doesn't matter.

chris1984:
@johnpmitsch do you want me to pick this up from Christine since she is on leave?

johnpmitsch:
@chris1984 go for it!

johnpmitsch:
Closing this out since @chris1984 will be taking it over in a new PR as discussed above

johnpmitsch:
New [PR opened](https://github.com/theforeman/foreman-packaging/pull/2192/files)

PR #2172, opened 2018-02-05
Add the foreman_datacenter plugin dependencies

ekohl:


ekohl:
I'm unsure about the licenses.

ekohl:
[test packaging]

ekohl:
Currently fails because ```rubygem-foreman_datacenter``` has a build time dependency on rqrcode which is included in this PR. I'd first like to get some clarity on the licenses before we proceed.

zdenekjanda:
@ekohl do you need anything from us regards licensing ? We are including various libraries with various licenses here (MIT, for example), all this pdf/ttf/qr stuff is maily for labeling functionality, but i havent seen any contradictions so far for packaging.

ekohl:
@zdenekjanda I'm a bit unsure about the pdf/ttf/qr licenses and I'm not a lawyer so I don't know if that's a problem.

dgoetz:
@ekohl I typically stick to https://fedoraproject.org/wiki/Packaging:LicensingGuidelines and https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#SoftwareLicenses

According to this I would say License tag should be:
pdf-core: GPLv2 or GPLv3
prawn: GPLv2 or GPLv3
ttfunk: GPLv2 or GPLv3
chunky_png: MIT
rqrcode: MIT

So the first three allow to choose the license for distribution, using this license tag would redistribute under GPLv2 or GPLv3, the additional custom one would need some checking if it should be added but being an OR-Condition I simply would leave it without. If you want the custom license checked I can ask a colleague to do so.

ekohl:
@dgoetz thanks! I've looked into this with your advice and also looked at Fedora packages themselves. Also added the packages to the manifest (including a sorting PR). Removed the plugin itself so the build passes.

PR #2194, opened 2018-02-12
Fixes #22551 - use runuser instead of sudo in k-backup

evgeni:

### DIFF
--- a/packages/katello/katello/backup.rb
+++ b/packages/katello/katello/backup.rb
@@ -339,7 +339,7 @@ module KatelloUtilities
 
     def validate_directory
       return true unless @databases.include? "pgsql"
-      unless system("sudo -u postgres test -w #{@dir}")
+      unless system("runuser - postgres -c 'test -w #{@dir}'")
         puts "****cancelled****"
         puts "Postgres user needs write access to the backup directory"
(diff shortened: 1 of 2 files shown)


sean797:
I suspect not, but do we care about ?
https://github.com/theforeman/foreman-packaging/blob/3c16b215cd7ce77d14e3cb6ae460291fed9dbe26/katello/katello/restore.rb#L101

evgeni:
not really, unless you run restore from cron? ;)

but I think I'd still align this to match the backup code

evgeni:
Right. I think I'll just use `sudo` for the restore then.

Or I could do as the installer and use `su - apache -s /bin/bash -c pulp-manage-db`

Or I pretend to port the `su` call to `runuser` and use `runuser - apache -s /bin/bash pulp-manage-db`

@johnpmitsch @sean797 what would you prefer?

sean797:
+1 to `su - apache -s /bin/bash -c pulp-manage-db`. Lets do what the installer does

evgeni:
@johnpmitsch @sean797 updated

ekohl:
@johnpmitsch could you have another look?

ekohl:
Thanks!

PR #2195, opened 2018-02-12
[rpm] Update smart_proxy_ansible to 2.0.1

dLobatog:

* [x] Nightly
* [x] 1.17


ekohl:
gem-compare tells me the foreman_ansible_core dependency was dropped. Is that correct? In that case it should also be dropped from packaging.

```
  DIFFERENT runtime dependencies:
    2.0.0->2.0.1:
      * Deleted:
            foreman_ansible_core ["< 3.0", ">= 2.0.2"] (runtime)
```

ares:
what's the status here? smart_proxy_ansible can't be installed with nightlies until this is resolved

PR #2199, opened 2018-02-13
Fixes #22566 - only tar configs that are present on the machine

evgeni:

### DIFF
--- a/packages/katello/katello/backup.rb
+++ b/packages/katello/katello/backup.rb
@@ -123,7 +123,8 @@ module KatelloUtilities
       feature_configs += ["/var/lib/dhcpd", "/etc/dhcp"] if feature_included?("dhcp")
       feature_configs.push("/usr/share/xml/scap") if feature_included?("openscap")
 
-      backup_configs + feature_configs
+      all_configs = backup_configs + feature_configs
+      all_configs.compact.select { |path| Dir.glob(path).any? }
     end
 
     def confirm
@@ -332,8 +333,8 @@ module KatelloUtilities
       puts "Backing up config files... "
       flags = "--exclude=\"/var/www/html/pub/isos*\" --exclude=\"/var/www/html/pub/exports*\" --selinux --create --gzip"
       flags += " --file=#{File.join(@dir, 'config_files.tar.gz')} --listed-incremental=#{File.join(@dir, '.config.snar')}"
-      tar_command = "tar #{flags} #{configure_configs.join(' ')} 2>/dev/null"
-      run_cmd(tar_command, [0,2])
+      tar_command = "tar #{flags} #{configure_configs.join(' ')}"
+      run_cmd(tar_command)
       puts "Done."
     end
 


packages/katello/katello/backup.rb
```diff
@@ -122,7 +122,9 @@ def configure_configs
       feature_configs += ["/var/lib/dhcpd", "/etc/dhcp"] if feature_included?("dhcp")
       feature_configs.push("/usr/share/xml/scap") if feature_included?("openscap")
 
-      backup_configs + feature_configs
+      all_configs = backup_configs + feature_configs
+      all_configs = all_configs.compact
+      all_configs.collect { |path| Dir.glob(path) }.flatten
```
ekohl: Would it make sense to run a uniq on this?

packages/katello/katello/backup.rb
```diff
@@ -331,8 +333,8 @@ def backup_config_files
       puts "Backing up config files... "
       flags = "--exclude=\"/var/www/html/pub/isos*\" --exclude=\"/var/www/html/pub/exports*\" --selinux --create --gzip"
       flags += " --file=#{File.join(@dir, 'config_files.tar.gz')} --listed-incremental=#{File.join(@dir, '.config.snar')}"
-      tar_command = "tar #{flags} #{configure_configs.join(' ')} 2>/dev/null"
-      run_cmd(tar_command, [0,2])
+      tar_command = "tar #{flags} #{configure_configs.join(' ')}"
```
ekohl: Is there a risk that this will now exceed the maximum command length? What about arguments with spaces? Does ```run_cmd``` handle an array as argument?

packages/katello/katello/backup.rb
evgeni: I'd not expect duplicates here, but it also won't hurt.

packages/katello/katello/backup.rb
evgeni: Not more or less than before. The `Dir.glob()` above should only add a few more entries to the already super long line.

And no, `run_cmd` does not handle shell-specialchars properly.

ekohl: Previously it let the shell expand and I think that handles at least spaces. Moving it to Ruby moves the responsibility.

evgeni: We can do `all_configs.select { |path| Dir.glob(path).any? }` above, so just using `Dir.glob` for testing, not for actual resolution, this would keep the escaping of these files at shell level

ekohl: That's neat and I like it.

evgeni:
Updated!

packages/katello/katello/backup.rb
```diff
@@ -123,7 +123,9 @@ def configure_configs
       feature_configs += ["/var/lib/dhcpd", "/etc/dhcp"] if feature_included?("dhcp")
       feature_configs.push("/usr/share/xml/scap") if feature_included?("openscap")
 
-      backup_configs + feature_configs
+      all_configs = backup_configs + feature_configs
+      all_configs = all_configs.compact
+      all_configs.select { |path| Dir.glob(path).any? }
```
ekohl: nit: I'd merge this line with the previous one

evgeni: fair, updated

ekohl:
@johnpmitsch want to take a last look at it before merging?

johnpmitsch:
LGTM

ekohl:
Does this now need a changelog entry + version bump?

ekohl:
In the past tito tag would catch this, but I'm not sure how to deal with this in a tagless world and we're likely to forget it on the next release.

PR #2201, opened 2018-02-15
Add Package Manifest for obal

pcreech:
Add Package Manifest for obal

Obal utilizes a package_manifest.yaml inventory file in the parent directory.  This PR is to add one so obal can be used with foreman-packaging

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---


package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    releasers:
+    - koji-foreman-nightly
+  hosts:
+    foreman: {}
+    rubygem-activerecord-session_store: {}
+    rubygem-addressable: {}
+    rubygem-algebrick:
+      releasers:
+      - koji-foreman-nightly
+      - koji-foreman-plugins-nightly
```
ehelms: This is a good red-flag to find! This should not exist in both of these releasers anymore. The `koji-foreman-plugins-nightly` should be dropped since this is in Foreman core repositories now.

package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    rubygem-ancestry: {}
+    rubygem-apipie-bindings: {}
+    rubygem-apipie-params:
+      releasers:
+      - koji-foreman-nightly
+      - koji-foreman-plugins-nightly
```
ehelms: This is a good red-flag to find! This should not exist in both of these releasers anymore. The `koji-foreman-plugins-nightly` should be dropped since this is in Foreman core repositories now.

package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    rubygem-autoparse: {}
+    rubygem-autoprefixer-rails: {}
+    rubygem-awesome_print: {}
+    rubygem-bootstrap-sass: {}
+    rubygem-bundler_ext: {}
+    rubygem-clamp: {}
+    rubygem-concurrent-ruby:
+      releasers:
+      - koji-foreman-nightly
+      - koji-foreman-plugins-nightly
```
ehelms: This is a good red-flag to find! This should not exist in both of these releasers anymore. The `koji-foreman-plugins-nightly` should be dropped since this is in Foreman core repositories now.

package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    rubygem-concurrent-ruby-edge:
+      releasers:
+      - koji-foreman-nightly
+      - koji-foreman-plugins-nightly
```
ehelms: This is a good red-flag to find! This should not exist in both of these releasers anymore. The `koji-foreman-plugins-nightly` should be dropped since this is in Foreman core repositories now.

package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    rubygem-css_parser: {}
+    rubygem-daemons: {}
+    rubygem-deacon: {}
+    rubygem-deep_cloneable: {}
+    rubygem-domain_name: {}
+    rubygem-dynflow:
+      releasers:
+      - koji-foreman-nightly
+      - koji-foreman-plugins-nightly
```
ehelms: This is a good red-flag to find! This should not exist in both of these releasers anymore. The `koji-foreman-plugins-nightly` should be dropped since this is in Foreman core repositories now.

package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    rubygem-eventmachine: {}
+    rubygem-excon: {}
+    rubygem-extlib: {}
+    rubygem-facter: {}
+    rubygem-faraday:
+      releasers:
+      - koji-foreman-nightly
+      - koji-foreman-plugins-nightly
```
ehelms: This is a good red-flag to find! This should not exist in both of these releasers anymore. The `koji-foreman-plugins-nightly` should be dropped since this is in Foreman core repositories now.

package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    rubygem-fast_gettext:
+      releasers:
+      - koji-foreman-nightly
+      - koji-foreman-plugins-nightly
```
ehelms: This is a good red-flag to find! This should not exist in both of these releasers anymore. The `koji-foreman-plugins-nightly` should be dropped since this is in Foreman core repositories now.

package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    rubygem-ldap_fluff: {}
+    rubygem-little-plugger: {}
+    rubygem-locale: {}
+    rubygem-logging: {}
+    rubygem-multi_json: {}
+    rubygem-multipart-post: {}
+    rubygem-mysql2:
+      releasers:
+      - koji-foreman-nightly
+      - koji-foreman-plugins-nightly
```
ehelms: This is a good red-flag to find! This should not exist in both of these releasers anymore. The `koji-foreman-plugins-nightly` should be dropped since this is in Foreman core repositories now.

package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    rubygem-ruby2ruby: {}
+    rubygem-ruby_parser: {}
+    rubygem-rubyforge: {}
+    rubygem-safemode: {}
+    rubygem-scoped_search: {}
+    rubygem-secure_headers: {}
+    rubygem-sequel:
+      releasers:
+      - koji-foreman-nightly
+      - koji-foreman-plugins-nightly
```
ehelms: This is a good red-flag to find! This should not exist in both of these releasers anymore. The `koji-foreman-plugins-nightly` should be dropped since this is in Foreman core repositories now.

package_manifest.yaml
```diff
@@ -0,0 +1,415 @@
...
+    pulp-katello: {}
+    rubygem-anemone: {}
+    rubygem-foreman_scc_manager: {}
+    rubygem-foreman_virt_who_configure: {}
+    rubygem-hammer_cli_csv: {}
+    rubygem-hammer_cli_foreman_virt_who_configure: {}
+    rubygem-hammer_cli_katello: {}
+    rubygem-jenkins_api_client:
+      releasers:
+      - koji-katello-el7
```
ehelms: This releaser can actually be dropped out of our `releasers.conf` entirely and the packages referencing it use the regular `koji-katello`. This originated from a time when we supported EL6.

package_manifest.yaml
```diff
@@ -0,0 +1,382 @@
...
+katello:
+  vars:
+    package_base_dir: 'packages/katello/'
+    releasers:
+    - koji-katello
+  hosts:
+    katello: {}
+    katello-certs-tools: {}
+    katello-client-bootstrap: {}
+    katello-host-tools:
```
evgeni: do we want a `katello-client` group for `katello-host-tools` and `katello-repos` (why is that built into client anyways?)?

@ehelms what do you think?

ehelms: Hrmmm, since they are their own repositories -- I think that makes sense.

The `katello-repos` RPM has both server repositories and client repositories contained within it. So we build it for all of the buildroots to get the client sub-package into the client repos.

ehelms: @pcreech Want to make this small group? Then I think this is good to go.

pcreech: @ehelms done

evgeni:
The client group is missing from the packages metagroup.

pcreech:
@evgeni not anymore.  Thanks!

package_manifest.yaml
```diff
@@ -0,0 +1,383 @@
...
+    foreman: {}
+    plugins: {}
+    katello: {}
+    katello-client: {}
+
+foreman:
+  vars:
+    package_base_dir: 'packages/foreman/'
+    releasers:
+    - koji-foreman-nightly
```
ekohl: I thought this releaser only worked with foreman, foreman-selinux, foreman-proxy and foreman-installer.

ehelms: This is essentially the "jenkins" releaser.

ehelms:
@ekohl it's a fair question you raise. The fully change requires multiple updates and we can aim for a wholesale change over or small changes like this going in.

ekohl:
> @ekohl it's a fair question you raise. The fully change requires multiple updates and we can aim for a wholesale change over or small changes like this going in.

Since the manifest needs to be updated, I'd prefer to at least ensure the workflows that do touch it are documented. There's a good chance it'll go out of sync which is actually a problem we're trying to solve.

For starters I'd prefer to merge the other changes in this PR since that's already good to go.

ehelms:
I can agree with that -- @pcreech do you mind splitting out the non package manifest changes to a new PR given they are relevant even without the package_manifest?

I believe @evgeni has work to bring the CI job online and will just need to add config to ensure that only `rpm/develop` runs the new PR jobs and branches use the old.

The other big point we need to address is moving away form tagging in the git repository. That can be a stand-alone change that simply requires modifications to developer workflows if that seems acceptable @ekohl

pcreech:
@ehelms @ekohl moved the releaser work to [here](https://github.com/theforeman/foreman-packaging/pull/2214)

package_manifest.yaml
ekohl: Can obal handle using one releaser for a few packages while using another for the rest?

evgeni: Sure, just define them on the package instead of the group.

pcreech: What evgeni said.  The current releasers set in the PR at a group level were a best guess, and I am not familiar with the purpose of each releaser in foreman.  Therefore, if there are ones better suited for a group or package, please let me know.

ehelms: Essentially, this should be `koji-foreman` and either we set nightly package builds, e.g. foreman, to `koji-foreman-nightly` and then change that when we branch off to release branches OR we let the CI jobs handle overriding this when they need to. The nightly releasers that use Jenkins are not as simple as 'tito release' though so in general this particular releaser may require some thought or further configuration

package_manifest.yaml
```diff
@@ -0,0 +1,391 @@
+---
+packages:
+  children:
+    foreman: {}
+    plugins: {}
+    katello: {}
+    katello-client: {}
```
evgeni: dashes in groups are discouraged

package_manifest.yaml
```diff
@@ -0,0 +1,391 @@
+---
+packages:
+  children:
+    foreman: {}
```
evgeni: so uh, a group foreman, and a host foreman, that is gona explode. same for katello. :(

(that's why the pulp group is called pulp_group in sat-packaging)

ehelms:
@ekohl Since this is the first steps towards a bigger change, and you originally reviewed I want to give you a chance to final review and ACK

ekohl:
Looks like 000ebdb8d797b6ea1ae95f5b00fb12de373d9dd0 needs to be done too. It's a bit confusing because rubygem-mysql2 stored in ```packages/foreman/rubygem-mysql2``` but also included in the plugins repository as non-scl package since it's not in EPEL7. I think that the new directory structure kind of breaks down here.

ekohl:
If I understand it correctly then they should be added to ```plugins``` with an overridden ```package_base_dir``` so it can find it in ```packages/foreman```, correct? If so, that'd probably be enough for a first pass.

ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
This is green with https://github.com/theforeman/foreman-infra/pull/493 (note it did not test any packages which is intentional)

ehelms:
Given all current concerns are addressed and ACKs, I am merging this to allow further iteration on this and the scratch test job itself. Thanks @pcreech !

PR #2204, opened 2018-02-16
Fixes #22607 - add rex ssh to backup

chris1984:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/packages/katello/katello/backup.rb
+++ b/packages/katello/katello/backup.rb
@@ -87,7 +87,8 @@ module KatelloUtilities
         '/opt/puppetlabs/puppet/cache/foreman_cache_data',
         '/opt/puppetlabs/puppet/ssl/',
         '/var/lib/puppet/foreman_cache_data',
-        '/var/lib/puppet/ssl'
+        '/var/lib/puppet/ssl',
+        '/usr/share/foreman-proxy/.ssh/'
       ]
 
       scenario_answers = load_scenario_answers(@last_scenario)


chris1984:
@akofink, unfortunately, yes, since we do not encrypt any part of the backup. I can raise a docs bug for downstream and create a doc pr upstream to make sure the backup is safeguarded if that would suffice?

akofink:
We should definitely have a conversation about it. I like the idea of adding documentation that states that this could compromise security, and I also don't see why we can't incorporate an option for encryption of the sensitive data in the future. We obviously don't need to encrypt pulp or anything.

evgeni:
Well, a rouge backup can already compromise any machine by injecting puppet modules or rpms. Or telling a random rex proxy to execute stuff via ssh. So this isn't making it much worse (ssh is easier to start than a puppet master, though ;))

akofink:
True, not saying we should block this PR. :)

packages/katello/katello/backup.rb
```diff
@@ -87,7 +87,8 @@ def configure_configs
         '/opt/puppetlabs/puppet/cache/foreman_cache_data',
         '/opt/puppetlabs/puppet/ssl/',
         '/var/lib/puppet/foreman_cache_data',
-        '/var/lib/puppet/ssl'
+        '/var/lib/puppet/ssl',
+        '/usr/share/foreman-proxy/.ssh/'
```
ekohl: This does make me wonder: why is it writing in ```/usr```? FHS says this shouldn't happen and as an admin I'd be surprised about this behavior. Why doesn't it live in ```/var/lib/foreman-proxy```?

stbenjam: foreman-proxy $HOME is /usr/share/foreman-proxy

ekohl: It sounds like something would should fix, possibly via a symlink.

chris1984:
@akofink 

Downstream:

https://bugzilla.redhat.com/show_bug.cgi?id=1547650

Upstream:

https://projects.theforeman.org/issues/22648

chris1984:
@akofink can this get a merge?

akofink:
I have no perms here, but yes! @jlsherrill?

evgeni:
wonder if I should make it a habit pretending to be @jlsherrill? ;-)

PR #2207, opened 2018-02-16
update foreman-tasks to 0.11.1

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.17


packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -7,13 +7,13 @@
...
-Version: 0.11.0
-Release: 2%{?foremandist}%{?dist}
+Version: 0.11.1
+Release: 1%{?foremandist}%{?dist}
 Group: Development/Libraries
 License: GPLv3
 URL: https://github.com/theforeman/foreman-tasks
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
-Requires: foreman >= 1.15.0
+Requires: foreman >= 1.17.0
```
akofink: Since this is develop/nightly, shouldn't this be 1.18?

mmoll: Nope, this should be perfectly fine. foreman-tasks 0.11.x is compatible with 1.17 also.

PR #2224, opened 2018-02-22
Don't rely on 'tito tag' anymore

pcreech:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ For npm modules (nodejs- packages) you will need:
 
 To build locally or release RPMs from this repo, you also require:
 
+* [obal](https://github.com/theforeman/obal) 0.0.2 or higher
 * [tito](https://github.com/dgoodwin/tito) 0.6.1 or higher
 * [mock](http://fedoraproject.org/wiki/Projects/Mock) or koji client and an account (certificate) on koji.katello.org
 * scl-utils-build package (Fedora or Red Hat repositories)
@@ -34,7 +35,7 @@ Run:
 
 ## HOWTO: test a package
 
-Before tagging a build, please test it.  Using tito's --test flag, you can
+Before releasing a build, please test it.  Using tito's --test flag, you can
 generate test (S)RPMs by first committing your changes locally, then it will
 use the SHA in the RPM version.
 
@@ -49,8 +50,6 @@ the packages locally and quickly.
 
 The last argument is the name of the mock config in mock/, which includes SCL
 and non-SCL variants.
-
-**Notice:** tito works only on committed changes! If you are changing the `.spec` files, make sure you commit those changes before running `tito build` command.
 
 ### With koji access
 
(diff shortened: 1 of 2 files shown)


ehelms:
Thanks! This is a good start. I am going to note a few places I think could use updates with this change in the docs:

 * Talks about tagging and having to do commits for scratch builds (not true with this change) https://github.com/theforeman/foreman-packaging/pull/2224/files#diff-04c6e90faac2675aa89e2176d2eec7d8R36
 * Not true anymore -- https://github.com/theforeman/foreman-packaging/pull/2224/files#diff-04c6e90faac2675aa89e2176d2eec7d8R54

pcreech:
updated to address feedback

ekohl:
I do wonder if we shouldn't continue to use the git commit at least. Having uncommitted local changes is a recipe for disaster. I'd even consider maintaining git tags, just not tito tags which are a separate commit.

pcreech:
Yeah, I wasn't sure if that language should stay or go... I'll but that back in

dLobatog:
I'm not too familiar with this change, what does this mean in terms of how torelease a package?

pcreech:
@dLobatog This change essentially removes the need to do 'tito tag', and dependencies on tito's git tags during the releasing process

pcreech:
@ekohl Obal looks at what's layed down locally

ekohl:
@pcreech I wonder if that's a problem waiting to happen or me being overly cautious.

pcreech:
@ekohl to use git annex like we do, tito needs the annex to be 'unlocked', which means the new annexed source, at least, needs to be committed.  @ehelms thoughts?

ehelms:
I'm not overly worried. In our new world, the only thing that should be releasing is the Jenkins job. All packaging systems have ways to roll back as well. We have yet to hit this issue in other projects. I think if we are worried we can add functionality to 'release' in Obal to ask a user are they sure.

ekohl:
> In our new world, the only thing that should be releasing is the Jenkins job.

This is not true from the beginning. We should at least make a note that this is the case.

ehelms:
That works for me.

ekohl:
I'd be ok with this if we add a section ```Releasing a package``` which goes through the steps to release a package and what to be aware of. Preferably with copy-paste commands.

pcreech:
@ekohl I added a basic 'Releasing a package' section.  Let me know what you think

README.md
```diff
@@ -234,6 +200,22 @@ To build a new release package for foreman project for example, do this:
...
+
+    `$ cd packages/foreman/rubygem-mysql2/`
+
+1. Do a scratch build of the release first:
+
+    `$ tito release --scratch koji-foreman koji-foreman-plugins`
+
+1. Verify package built successfully in koji, and then release:
+
+    `$ tito release koji-foreman koji-foreman-plugins`
```
ehelms: Sorry to add a little bit more work. If the goal is for Obal to be our source of truth, we probably want an Obal section here and a 'manual' mode. This should be easy to write though:

```
obal release rubygem-mysql2
```

README.md
```diff
@@ -234,6 +200,22 @@ To build a new release package for foreman project for example, do this:
 
     $ tito release koji-foreman
 
+## Releasing a package
+
+After a PR has been merged to add or update a package, then the package should be released.
```
ekohl: In case it was added it's important to add the packages to the right koji tags. In manual mode you can use https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rel-eng/compare-with-tags.rb

README.md
pcreech: @ehelms I know this PR has been open a bit, but iirc the intent with this unit of work was to just use the tito builder that is in obal, not re-define the workflow to use obal yet.

ehelms: That's fair, we can leave the full switch to Obal for tooling as a separate change.

README.md
```diff
@@ -234,6 +200,22 @@ To build a new release package for foreman project for example, do this:
 
     $ tito release koji-foreman
 
+## Releasing a package
+
+After a PR has been merged to add or update a package, then the package should be released.  In case it's a new package, be sure to add it to the correct koji tags as well.  [Here's a helper script](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rel-eng/compare-with-tags.rb)
+
+1. Switch to the directory for the package to be released, i.e. for the rubygem-apipie-rails package:
```
ekohl: s/apipie-rails/mysql2/

README.md
```diff
@@ -234,6 +200,22 @@ To build a new release package for foreman project for example, do this:
...
+    `$ cd packages/foreman/rubygem-mysql2/`
+
+1. Do a scratch build of the release first:
+
+    `$ tito release --scratch koji-foreman koji-foreman-plugins`
+
+1. Verify package built successfully in koji, and then release:
+
+    `$ tito release koji-foreman koji-foreman-plugins`
+
```
ekohl: I'd add a note here that it uses local files so be sure to not have local changes.

ekohl:
This does warrant a post on discourse that it's changing, but only for rpm/develop.

pcreech:
https://community.theforeman.org/t/tito-tag-workflow-no-longer-needed-for-rpm-develop-packaging/8505

PR #2228, opened 2018-02-23
Fix plugin repoclosure

ekohl:
e6d1a925989d417b7b08019d3a05ab128b10516e removed these from the plugins repositories but they are in fact needed for repoclosure.


pcreech:
ACK

PR #2230, opened 2018-02-26
create foreman-telemetry package

mmoll:
this should be built into:

* [x] Nightly


ekohl:
I prefer https://github.com/theforeman/foreman/pull/5275 so it's an optional package.

lzap:
I will let you decide guys, will test the core change, sorry!

mmoll:
I'd rather go for telemetry to be optional, if possible.

lzap:
That's what we aim for in core, this is a bug indeed. For packaging, should be definitely a separate package.

PR #2236, opened 2018-02-27
Add nodejs-jquery-ui package

ekohl:
Needed for https://github.com/theforeman/foreman/pull/5212


packages/foreman/nodejs-jquery-ui/nodejs-jquery-ui.spec
```diff
@@ -0,0 +1,45 @@
...
+
+%description
+%{summary}
+
+%prep
+%setup -q -n package
+
+%install
+mkdir -p %{buildroot}%{nodejs_sitelib}/%{npm_name}
+cp -pfr external %{buildroot}%{nodejs_sitelib}/%{npm_name}
```
ekohl: This is actually very ugly and includes many bundled dependencies. Not sure how to get rid of this.

ekohl:
Probably needs a change in the manifest too but RPM job fails now:
```
12:03:26 + echo packages/foreman/nodejs-jquery-ui/nodejs-jquery-ui.spec
12:03:26 /var/lib/workspace/workspace/foreman-packaging-rpm-pr-test@tmp/durable-ce5a9da7/script.sh: line 3: syntax error near unexpected token `|'
```

ehelms:
The `rpm` job needs -- https://github.com/theforeman/foreman-infra/pull/493

ekohl:
@ehelms that's merged and yet I still get the same error.

evgeni:
@ekohl no, the error is different now ;)

evgeni:
[test rpm]

evgeni:
[test rpm]

evgeni:
should be fixed with https://github.com/theforeman/foreman-infra/pull/502 (still fails, because tags blah)

ekohl:
[test rpm]

ekohl:
[test rpm]

ekohl:
New error: ERROR: Unable to lookup latest package info. Perhaps you need to tag first?

Maybe this is because the git annex step is failing? Is the duplicate slash a problem?

ekohl:
@ehelms that's odd because this is a scratch build job - it should never be a production release before it's merged.

ehelms:
It's not that odd when you consider how tito works. If you don't supply --test, then it tries to look up the last tag to base the SRPM build off of. That's fundamental to tito (unless you override it like we did). So `--test` is required if using tags.

ehelms:
[test rpm]

ehelms:
@ekohl The `rpm` job is working now for new packages

ekohl:
I'll wait till the core PR is merged then.

ekohl:
Closing due to the lack of activity on the core PR.

PR #2237, opened 2018-02-27
Format the package manifest

ekohl:
This was automatically formatted using [obal-fmt](https://github.com/theforeman/obal/pull/59).

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -2,144 +2,14 @@
 packages:
   children:
     foreman_packages: {}
-    plugins_packages: {}
-    katello_packages: {}
     katello_client_packages: {}
-
+    katello_packages: {}
+    plugins_packages: {}
 foreman_packages:
-  vars:
-    package_base_dir: 'packages/foreman/'
-    releasers:
-    - koji-foreman
   hosts:
     foreman:
       releasers:
       - koji-foreman-nightly
-    rubygem-activerecord-session_store: {}
-    rubygem-addressable: {}
-    rubygem-algebrick: {}
-    rubygem-ancestry: {}
-    rubygem-apipie-bindings: {}
-    rubygem-apipie-params: {}
-    rubygem-apipie-rails: {}
-    rubygem-audited: {}
-    rubygem-autoparse: {}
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -269,25 +265,25 @@ katello_packages:
...
   vars:
-    package_base_dir: 'packages/katello/'
+    package_base_dir: packages/katello/
     releasers:
-    - koji-katello-client
-  hosts:
-    katello-host-tools: {}
-    katello-repos: {}
+    - koji-katello
+packages:
```
ehelms: I kinda liked this being at the top as it gave a general overview of the groups I expect to find below.

ehelms: Perhaps this can be a convention that is always at the top?

package_manifest.yaml
```diff
@@ -140,115 +232,19 @@ foreman_packages:
...
-    rubygem-rsec: {}
-    rubygem-rubyipmi: {}
-katello_packages:
   vars:
-    package_base_dir: 'packages/katello/'
+    package_base_dir: packages/foreman/
     releasers:
-    - koji-katello
+    - koji-foreman
+katello_client_packages:
```
ehelms: Possible to have the formatter put empty lines between groups? Would be easier to spot start and stop of a group that way.

package_manifest.yaml
ekohl: I like that too and will see how I can force the formatter to behave.

package_manifest.yaml
ekohl: agreed

package_manifest.yaml
```diff
@@ -1,20 +1,104 @@
----
-packages:
-  children:
-    foreman_packages: {}
-    plugins_packages: {}
-    katello_packages: {}
-    katello_client_packages: {}
-
 foreman_packages:
-  vars:
```
ekohl: I also liked vars above hosts because it's usually shorter and sort of a preset for the rest.

ekohl:
Closing this for now.

PR #2244, opened 2018-03-02
update foreman_dhcp_browser to 0.0.8

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.17


packages/plugins/rubygem-foreman_dhcp_browser/rubygem-foreman_dhcp_browser.spec
```diff
@@ -16,19 +16,19 @@
...
 Source0:    https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
-Requires:   foreman >= 1.2.0
+Requires:   foreman >= 1.12.0
 
 Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}rubygems
 
-Requires: %{?scl_prefix}rubygem(deface) < 2.0.0
+Requires: %{?scl_prefix}rubygem(deface) >= 0
```
ekohl: Technically you can leave of the ```>= 0``` here

mmoll: I thought about it, but this reflects the current gemspec 100%

PR #2255, opened 2018-03-06
Update foreman_ansible_core to 2.0.2 + foreman_remote_execution_core to 1.1.0

ekohl:


dLobatog:
[test default]

rubygem-foreman_ansible_core/rubygem-foreman_ansible_core.spec
```diff
@@ -4,18 +4,20 @@
...
 License: GPLv3
 URL: https://github.com/theforeman/foreman_ansible
-Source0: http://rubygems.org/gems/%{gem_name}-%{version}.gem
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}ruby
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
 Requires: %{?scl_prefix}rubygem(foreman-tasks-core) >= 0.1
 Requires: %{?scl_prefix}rubygem(foreman-tasks-core) < 1.0
+Requires: %{?scl_prefix}rubygem(foreman_remote_execution_core) >= 1.1
```
dLobatog: This dependency does not matter for 1.1.1 - in fact we don't want it. I can make a special version, but https://github.com/theforeman/foreman_ansible/blob/master/lib/foreman_ansible_core.rb takes care of it (I've tested it on production). The smart_proxy for example, relies on that file working well, as it cannot load these gems from outside the SCL.

ekohl: You mean 2.0.2? I copied this from the development spec.

dLobatog: Sorry - I meant 'does not matter for 1.16'

dLobatog: Nevermind - it does not matter but I checked and it does not hurt either. It'd just be a useless installed package, I can go into details if you're curious.

dLobatog:
Thanks @ekohl 

http://koji.katello.org/koji/taskinfo?taskID=87056 - REX core
http://koji.katello.org/koji/taskinfo?taskID=87063 - Ansible core

PR #2256, opened 2018-03-08
[rpm] Bump foreman_ansible to 2.1

dLobatog:

* [x] Nightly
* [x] 1.17

---------------

I think this is missing a section to compile the assets on `/webpack`. Katello also has webpack assets, but I couldn't find a `webpack:compile` call anywhere, how is this done?


ekohl:
I'm not aware of any plugin doing any webpack compilation. I think we should add something for this in the foreman.spec macros so any plugin can benefit from it. Not sure how it should exactly look.

packages/foreman/foreman/foreman.spec
```diff
@@ -1070,6 +1070,13 @@ GEMFILE
 # build apipie cache index
 %%%{name}_apipie_cache %%{%{name}_rake} apipie:cache:index >> %%{%{name}_log_dir}/apipie_cache.log 2>&1 || :
 
+# Generate precompiled webpack assets.
+# Plugins may use this to recompile the webpack bundle with its assets.
+# ./script/npm_install_plugins.js does not need to be called, as the npm
+# plugin dependencies should be installed by the time this is called
+%%%{name}_precompile_webpack \\
+%%{?-a:/usr/bin/%%{?scl:%%{scl}-}rake webpack:compile RAILS_ENV=production --trace}
```
dLobatog: @ekohl I think this should suffice, as npm packages should be installed by the time this command is called.

ekohl:
> @ekohl Updated - also how on earth is Katello compiling the webpack assets? It doesn't even have npm dependencies

That's a very good question that I don't know the answer to now. It is pulling in npm via `foreman-assets`.

Currently every plugin runs `foreman-rake plugin:assets::precompile` via the existing macro. That calls Sprockets now, but if we modify it to also handle webpack as well then I think we have a fairly clean path forward. Bonus value is that it will also work for Debian meaning less duplication.

dLobatog:
Do you mean we should modify `plugin:assets:precompile to also handle webpack compilation so that it's automatically done on all packages? I think it makes sense. 

The task is provided by Rails, so we should just call it something else. `plugin:assets:precompile` is terribly slow compared to `webpack:compile`, but think there are no plugins that only have webpack assets.

packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -19,25 +19,29 @@ Requires: %{?scl_prefix_ruby}ruby(release)
...
 Requires: %{?scl_prefix}rubygem(foreman-tasks) >= 0.8
-Requires: %{?scl_prefix}rubygem(foreman-tasks) < 1.0
+Requires: %{?scl_prefix}rubygem(foreman-tasks) < 0.12
 Requires: %{?scl_prefix}rubygem(dynflow) >= 0.8.14
 Requires: %{?scl_prefix}rubygem(dynflow) < 0.9.0
 Requires: %{?scl_prefix}rubygem(foreman_ansible_core) >= 2.0.2
 Requires: %{?scl_prefix}rubygem(foreman_ansible_core) < 3.0
 Requires: %{?scl_prefix}rubygem(foreman_remote_execution) >= 1.4.4
 Requires: %{?scl_prefix}rubygem(foreman_remote_execution) < 2.0
+Requires: npm(react-json-tree) >= 0.1.1
```
ekohl: I think it's only a build requirement because we use ```webpack:compile```.

packages/foreman/foreman/foreman.spec
ehelms: Does this exit gracefully if there are no webpack assets to compile?

packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -19,25 +19,29 @@ Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}rubygems
 Requires: %{?scl_prefix}rubygem(deface) < 2.0
 Requires: %{?scl_prefix}rubygem(foreman-tasks) >= 0.8
-Requires: %{?scl_prefix}rubygem(foreman-tasks) < 1.0
+Requires: %{?scl_prefix}rubygem(foreman-tasks) < 0.12
```
ehelms: This means this is incompatible with Katello FYI given it relies on > 0.12

packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -71,6 +75,8 @@ cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
 %foreman_bundlerd_file
 %foreman_precompile_plugin -a -s
 
+mv -r .%{gem_dir}/public/webpack/%%{gem_name} %{foreman_webpack_plugin}
```
ehelms: Could the macro handle this as the default operation? I think we want plugins with webpack to follow this model so it should be generic enough to turn into macro.

ehelms: This can be dropped now

packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -79,12 +85,15 @@ cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
...
 %{gem_instdir}/locale
+%{gem_instdir}/webpack
+%{gem_instdir}/package.json
 %exclude %{gem_cache}
 %{gem_spec}
 %{foreman_bundlerd_plugin}
 %{foreman_apipie_cache_foreman}
 %{foreman_apipie_cache_plugin}
 %{foreman_assets_plugin}
+%{foreman_webpack_plugin}
```
ehelms: Add in %{foreman_webpack_foreman}

packages/foreman/foreman/foreman.spec
```diff
@@ -1062,6 +1062,8 @@ GEMFILE
 %%%{name}_pluginconf_dir %{_sysconfdir}/%{name}/plugins
 # Common assets locations
 %%%{name}_assets_plugin %%{gem_instdir}/public/assets/%%{gem_name}
+# Common webpack locations
+%%%{name}_webpack_plugin %%{foreman_dir}/public/webpack/%%{gem_name}
```
ehelms: This is now included in Foreman

ekohl:
You labeled this as 1.17 but in 1.17 we don't have the webpack plugin infrastructure present.

Other than that I can't really explain `ERROR:  While executing gem ... (NoMethodError) undefined method `payload' for nil:NilClass`.
[test rpm]

ares:
I tried installing this on nightly setup, but navigating to ansible report page, I'm getting this

```
bundle-ce0196125544525dc65c.js:1 Uncaught Error: Component not found:  ReportJsonViewer among BookmarkContainer, DonutChart, StatisticsChartsList, PowerStatus, NotificationContainer, ToastNotifications, StorageContainer, PasswordStrength, BreadcrumbBar
    at Object.markup (bundle-ce0196125544525dc65c.js:1)
    at Object.t.mount (bundle-ce0196125544525dc65c.js:1)
    at 92:430
2vendor-6f28897291cf7f524049.js:1 Uncaught TypeError: Cannot read property 'series' of null
    at m (vendor-6f28897291cf7f524049.js:1)
    at vendor-6f28897291cf7f524049.js:1
    at vendor-6f28897291cf7f524049.js:1
    at k (vendor-6f28897291cf7f524049.js:1)
    at $ (vendor-6f28897291cf7f524049.js:1)
24vendor-6f28897291cf7f524049.js:1 Uncaught TypeError: Cannot read property 'grid' of null
    at v (vendor-6f28897291cf7f524049.js:1)
    at HTMLCanvasElement.b (vendor-6f28897291cf7f524049.js:1)
    at HTMLCanvasElement.dispatch (vendor-6f28897291cf7f524049.js:1)
    at HTMLCanvasElement.b.handle (vendor-6f28897291cf7f524049.js:1)
vendor-6f28897291cf7f524049.js:1 Uncaught TypeError: Cannot read property 'series' of null
    at m (vendor-6f28897291cf7f524049.js:1)
    at vendor-6f28897291cf7f524049.js:1
    at vendor-6f28897291cf7f524049.js:1
    at k (vendor-6f28897291cf7f524049.js:1)
    at $ (vendor-6f28897291cf7f524049.js:1)
m @ vendor-6f28897291cf7f524049.js:1
(anonymous) @ vendor-6f28897291cf7f524049.js:1
(anonymous) @ vendor-6f28897291cf7f524049.js:1
k @ vendor-6f28897291cf7f524049.js:1
$ @ vendor-6f28897291cf7f524049.js:1
```

ehelms:
This error may be attributed to (https://github.com/theforeman/foreman/pull/5530) which is what fixed things for Katello.

ehelms:
[test rpm]

ares:
It seems, all is green

PR #2257, opened 2018-03-08
[deb] Bump foreman_ansible to 2.1

dLobatog:

* [ ] Nightly
* [x] 1.17

----------

I think this might be the first debian plugin with webpack assets, do you have any ideas on how would it be best to compile them?

### DIFF
--- a/plugins/ruby-foreman-ansible/debian/changelog
+++ b/plugins/ruby-foreman-ansible/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-ansible (2.1.0-1) stable; urgency=low
+
+  * 2.1.0 released
+
+ -- Daniel Lobato Garcia <me@daniellobato.me>  Thu, 08 Mar 2018 10:36:08 +0100
+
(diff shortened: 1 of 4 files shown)


mmoll:
the version needs to be 2.1.0, like the gem

mmoll:
if this is not compatible with foreman-tasks 0.12, that would be 1.17 only, at least for the moment

mmoll:
please open against deb/1.17 then :)

mmoll:
Also, as mentioned in GH-2256, DEB and RPM probably need some work to deal with the new webpack assets, I guess the packaging team needs at least input from the frontend guys here?

ehelms:
[test packaging]

komidore64:
please ignore the `default` pending check.

mmoll:
waiting on GH-2442

mmoll:
@dLobatog could you update this to 2.1.2 and point to (and rebase onto) `deb/1.17`?

mmoll:
does 1.17 bring all the things for webpack assets? And does foreman_ansible 2.1.x use webpack assets?

PR #2264, opened 2018-03-08
Bump foreman_ansible to 1.5.2

dLobatog:
* [x] 1.16

### DIFF
--- a/plugins/ruby-foreman-ansible/debian/changelog
+++ b/plugins/ruby-foreman-ansible/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-ansible (1.5.2-1) stable; urgency=low
+
+  * 1.5.2 released
+
+ -- Daniel Lobato Garcia <me@daniellobato.me>  Mon, 12 Mar 2018 19:50:43 +0100
+
 ruby-foreman-ansible (1.4.6-1) stable; urgency=low
 
   * 1.4.6 released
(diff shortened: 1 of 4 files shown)


akofink:
Why don't you base these PRs off of each other so we can see the complete build?

mmoll:
`19:00:32 Could not find gem 'foreman_ansible_core (> 2.0.2)', which is required by gem`
Is the gemspec of the gem itself OK there?

dLobatog:
` bundler: command not found: /usr/bin/rake`

[test packaging]

mmoll:
```
21:08:19 Bundler could not find compatible versions for gem
21:08:19 "foreman_remote_execution_core":
21:08:19   In Gemfile:
21:08:19     foreman_ansible (= 1.5.2) was resolved to 1.5.2, which depends on
21:08:19       foreman_ansible_core (>= 2.0.2) was resolved to 2.0.2, which depends on
21:08:19         foreman_remote_execution_core (~> 1.1)
21:08:19 
21:08:19 Could not find gem 'foreman_remote_execution_core (~> 1.1)', which is required
21:08:19 by gem 'foreman_ansible_core (>= 2.0.2)', in any of the sources.
```

ares:
Is the foreman-tasks pin still needed? I think new version has been released. We can keep it for next release but it should be removed before 1.18 is built.

mmoll:
I think this whole situation needs to get solved differently for foreman_ansible 1.x. By pulling in foreman_ansible_core 2.x you're creating a dependency onto foreman_remote_execution_core that simply was not there before and I'm puzzled why this should be needed. But maybe some commits that could explain it are missing in the (wrongly named) 1.0-stable branch in the foreman_ansible repo.

PR #2276, opened 2018-03-13
spec changes for katello-host-tools updates

jturel:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

katello-host-tools has been updated to support DNF and some of the code was moved around - requiring these changes. This is WIP because there will be some more changes but it reflects the changes needed to build the proper RPM from current master.

This can be tested by checking out katello-host-tools and tarballing: `git archive HEAD --prefix katello-host-tools-3.1.0/ -o katello-host-tools-3.1.0.tar.gz

Once all changes are made in katello-host-tools I'll bump the version, get it released, and we can merge this.


jturel:
[test]

jturel:
[test packaging]

jturel:
[test rpm]

jturel:
[test rpm]

jturel:
[test rpm]

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -102,18 +112,35 @@ mkdir -p %{buildroot}/%{_prefix}/lib/gofer/plugins
...
 
-mkdir -p %{buildroot}/%{_prefix}/lib/yum-plugins
-cp src/yum-plugins/package_upload.py %{buildroot}/%{_prefix}/lib/yum-plugins
-cp src/yum-plugins/enabled_repos_upload.py %{buildroot}/%{_prefix}/lib/yum-plugins
+%if %{dnf_install}
+%define katello_libdir %{python3_sitelib}/katello
+%define plugins_dir %{python3_sitelib}/dnf-plugins
+%define plugins_confdir %{_sysconfdir}/dnf/plugins
+%else
+%define katello_libdir /%{python_sitelib}/katello
```
ekohl: Is that `/` before `%{python_sitelib}` intentional?

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -165,13 +188,31 @@ exit 0
 
 %files
 %defattr(-,root,root,-)
-%config %{_sysconfdir}/yum/pluginconf.d/package_upload.conf
-%config %{_sysconfdir}/yum/pluginconf.d/enabled_repos_upload.conf
-/var/cache/katello-agent/
+%dir /var/cache/katello-agent/
+%config(noreplace) %{plugins_confdir}/package_upload.conf
+%config(noreplace) %{plugins_confdir}/enabled_repos_upload.conf
+%{katello_libdir}/constants.py*
```
ekohl: Would it make sense to just add ```%{katello_libdir}/*.py*```?

packages/katello/katello-host-tools/katello-host-tools.spec
jturel: interestingly, el5 build breaks without it

packages/katello/katello-host-tools/katello-host-tools.spec
jturel: I don't think so - tracer.py exists in that folder, and that's installed by the tracer subpackage. Also, I think it's better to be more explicit in the files section so nothing can sneak in but maybe that's just me!

ekohl: I guess I'd organize the package a bit more so tracer is in a separate directory but explicit is good.

jturel:
@ekohl re: `make dnf-install` `make yum-install` do you mean to actually add a Makefile here or you just want a way to override the dnf_install global var?

ekohl:
@jturel I think it'd be nice if it was in the projects git repo

jturel:
@ekohl I opened an issue for us to add your suggestions feel free to add any specific details http://projects.theforeman.org/issues/23037

Thanks for looking at this PR!

jturel:
@johnpmitsch since the code itself has been tested I don't think much needs to be done here testing wise.

Installing the RPM on a client and making sure that the plugins load properly (they are found by DNF/yum and no errors are logged in /var/log/{dnf,yum}.log) should be good.

The yum stuff should just work. If you want to test the DNF side of things you'll need a fedora27 box and to build subscription-manager by hand to get it running under python3 which is what is used by DNF. I have instructions for that here: https://github.com/Katello/katello-host-tools/pull/51

jturel:
[test rpm]

jturel:
@ekohl would you do the honors of merging and approving if you're good with the changes here?

packages/katello/katello-host-tools/katello-host-tools.spec
jturel: Something is fishy about it. Taking a look

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -1,6 +1,9 @@
+%global dnf_install (0%{?rhel} > 7) || (0%{?fedora} > 26)
+%{!?python_sitelib: %global python_sitelib %(%{__python} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")}
```
jturel: @ekohl Nice catch on the '/'

In reality there is no python_sitelib macro in el5 so I found this workaround on the internet.

```
[root@2b6de9090795 ~]# python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"
/usr/lib/python2.4/site-packages
```

ekohl: Always pays off to question when something doesn't look right. Either you learn something you didn't know or it's a bug :)

ekohl:
Doesn't appear it was built for f27. @ehelms @evgeni?

ekohl:
Ah right, the releasers ...

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -102,29 +118,42 @@ mkdir -p %{buildroot}/%{_prefix}/lib/gofer/plugins
...
+%define katello_libdir %{python_sitelib}/katello
+%define plugins_dir %{_prefix}/lib/yum-plugins
+%define plugins_confdir %{_sysconfdir}/yum/pluginconf.d
+%endif
+
+mkdir -p %{buildroot}/%{katello_libdir}
+mkdir -p %{buildroot}/%{plugins_dir}
+mkdir -p %{buildroot}/%{plugins_confdir}
+
+cp src/katello/*.py %{buildroot}/%{katello_libdir}/
```
ekohl: I realize this was already present in the old setup, but `%{katello_libdir}` is already an absolute path so the `/` between the two macros is not needed. Doesn't hurt, but it's cleaner not to. This is true in multiple places and I won't comment on them all but basically grep for `%{buildroot}/`.

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -96,35 +112,48 @@ popd
...
-cp src/yum-plugins/package_upload.py %{buildroot}/%{_prefix}/lib/yum-plugins
-cp src/yum-plugins/enabled_repos_upload.py %{buildroot}/%{_prefix}/lib/yum-plugins
+cp src/katello/*.py %{buildroot}%{katello_libdir}/
+cp etc/yum/pluginconf.d/*.conf %{buildroot}%{plugins_confdir}/
 
-mkdir -p %{buildroot}/%{_sysconfdir}/yum/pluginconf.d/
-cp etc/yum/pluginconf.d/package_upload.conf %{buildroot}/%{_sysconfdir}/yum/pluginconf.d/package_upload.conf
-cp etc/yum/pluginconf.d/enabled_repos_upload.conf %{buildroot}/%{_sysconfdir}/yum/pluginconf.d/enabled_repos_upload.conf
+%if %{dnf_install}
+cp src/dnf_plugins/*.py %{buildroot}%{plugins_dir}/
```
ekohl: Technically this line could be before the `%if` since it's in both branches but it's not that important.

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -182,11 +229,26 @@ exit 0
...
 %files tracer
-%{_sysconfdir}/yum/pluginconf.d/tracer_upload.conf
+%{plugins_dir}/tracer_upload.py*
+%{katello_libdir}/tracer.py*
+%{plugins_confdir}/tracer_upload.conf
+
+%if %{dnf_install}
+%{katello_libdir}/__pycache__/tracer.*
+%{plugins_dir}/__pycache__/tracer_upload.*
+%else%
```
ekohl: I'm surprised `%else%` works. Shouldn't it be `%else`?

jturel:
@ekohl updated. If you're good with how things are now I'll ask @johnpmitsch to give this another run-through in light of the changes

packages/katello/katello-host-tools/katello-host-tools.spec
jturel: I'm surprised too, will fix

packages/katello/katello-host-tools/katello-host-tools.spec
jturel: I'll leave it :)

jturel:
@ekohl ditto on John's question.

 I got that same error *repeatedly* before I even made any changes against the previous version of the rpm. I really don't know what to make of it. The file I annexed locally was taken right from the github release and I saw it happen to some others (this was several weeks back) while running setup_sources.sh So it seems like a larger problem to me. If you know, how can we fix it?

ekohl:
@johnpmitsch no, it's a bit underdocumented for maintainers.
@jturel could it be that the tarball was regenerated at some point?

jturel:
@ekohl I don't think so. what do you make of this? not the checksums

```
[jturel@strangeways katello-host-tools]$ pwd
/home/jturel/code/foreman-packaging/packages/katello/katello-host-tools
[jturel@strangeways katello-host-tools]$ md5sum katello-host-tools-3.2.0.tar.gz 
fd6365308174c9c92dcd68d8d576fb62  katello-host-tools-3.2.0.tar.gz
[jturel@strangeways katello-host-tools]$ wget https://github.com/Katello/katello-host-tools/archive/3.2.0.tar.gz
<snip>
2018-04-12 17:06:50 (1.84 MB/s) - ‘3.2.0.tar.gz.1’ saved [22824/22824]

[jturel@strangeways katello-host-tools]$ md5sum 3.2.0.tar.gz
fd6365308174c9c92dcd68d8d576fb62  3.2.0.tar.gz
```

ekohl:
It is odd that a manual `git annex get` did work. I wonder if there's something in the obal version I'm using that couldn't retrieve it. Anyway, it should be building in koji now

PR #2277, opened 2018-03-14
Remove unused packages and tito whitelists

ekohl:
Please carefully look if this is correct. jenkins_api client was used by foreman_pipeline but that was never imported to this repository and not rebuilt for Ruby 2.4 without breaking repoclosure.


ekohl:
obal doesn't really understand package removal :)

ehelms:
I think we should also drop -- tfm-rubygem-foreman_pipeline since these were dependencies of it (and I believe it is unmaintained)

ekohl:
Ah, it is in comps but the spec file isn't there.

ekohl:
Updated.

rel-eng/tito.props
```diff
@@ -395,8 +394,7 @@ whitelist = katello-repos katello-host-tools
 [katello-nightly-rhel7]
 disttag = .el7
 scl = tfm
-whitelist = rubygem-robotex
```
ehelms: This package is still used

ehelms: Oh I see it down below, carry on

PR #2286, opened 2018-03-15
add ansiblerole-insights-client

gavin-romig-koch:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---
This RPM contains an Ansible Role, which once installed on a system can be used to install, configure, register other remote systems with Red Hat Insights.  The role also improves the integration between insights on the remote system and Ansible on connecting systems.

This SRPM largely follows the design of rhel-system-roles.spec, though it packages many roles into one rpm, where as this rpm only packages one role.

I choose the 'ansiblerole-' prefix following the pattern of the 'rubygem-' prefix.

I tested the built RPMs on F27 and RHEL7.4.   The test rpms were built with el7-noscl mock config and with 'f27' (default) mock config.  I could not get el7-scl config to build, something was missing from the root .... 

I'm completely new to tito, and sorta new to packaging in general. I did some packaging for Fedora and RHEL back in the olden days, back before git saved us from svn, so if I've messed up something just let me know.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -335,6 +335,7 @@ plugins_packages:
     releasers:
     - koji-foreman-plugins
   hosts:
+    ansiblerole-insights-client: {}
     nodejs-react-json-tree: {}
     puppet-foreman_scap_client: {}
     rubygem-angular-rails-templates: {}
--- a/packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
+++ b/packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
@@ -1,0 +1,42 @@
+%global repo_orgname RedHatInsights
+%global repo_name insights-client-role
+
+%global role_orgname %{repo_orgname}
+%global role_name insights-client
+
+Name: ansiblerole-insights-client
+Summary: Packaging of the insights-client Ansible role
+Version: 1.5
+Release: 1%{?dist}
+License: ASL 2.0
+
+Source0: https://github.com/%{repo_orgname}/%{repo_name}/archive/v%{version}.tar.gz#/%{role_name}-%{version}.tar.gz
+Url: https://github.com/%{repo_orgname}/%{repo_name}/
+BuildArch: noarch
+
+Requires: ansible
+
+%description
+This package installs the insights-client Ansibile role.
+
+Make sure that "/usr/share/ansible/roles" is on your Ansible role_path.
+
+%prep
+
+%setup -qc
+
+%build
+
+%install
+mkdir -p %{buildroot}%{_datadir}/ansible/roles
+cp -pR %{repo_name}-%{version} %{buildroot}%{_datadir}/ansible/roles/%{role_orgname}.%{role_name}
+
+%files
+%{_datadir}/ansible/roles/%{role_orgname}.%{role_name}
+%doc %{repo_name}-%{version}/examples/*.yml
+%doc %{repo_name}-%{version}/README.md
+%license %{repo_name}-%{version}/LICENSE
+
+%changelog
+* Thu Mar 15 2018 Gavin Romig-Koch <gavin@redhat.com> - 1.5-1
+- Initial release.  Based largely on the pattern of rhel-system-roles.spec
--- a/comps/comps-foreman-plugins-rhel7.xml
+++ b/comps/comps-foreman-plugins-rhel7.xml
@@ -7,6 +7,7 @@
     <description>Plugin packages for the Foreman</description>
     <uservisible>true</uservisible>
     <packagelist>
+      <packagereq type="default">ansiblerole-insights-client</packagereq>
       <packagereq type="default">ipxe-bootimgs</packagereq>
       <packagereq type="default">nodejs-react-json-tree</packagereq>
       <packagereq type="default">puppet-foreman_scap_client</packagereq>
--- a/packages/plugins/ansiblerole-insights-client/insights-client-1.5.tar.gz
+++ b/packages/plugins/ansiblerole-insights-client/insights-client-1.5.tar.gz
@@ -1,0 +1,1 @@
+../../../.git/annex/objects/5z/M8/SHA256E-s6814--96b60e44c5aa1ee6153757d7f348ba6118bbc800426bfb2ef8f7bcdc6fddb4c7.tar.gz/SHA256E-s6814--96b60e44c5aa1ee6153757d7f348ba6118bbc800426bfb2ef8f7bcdc6fddb4c7.tar.gz
--- a/rel-eng/tito.props
+++ b/rel-eng/tito.props
@@ -358,7 +358,8 @@ whitelist = rubygem-angular-rails-templates
 
 [foreman-plugins-nightly-nonscl-rhel7]
 disttag = .el7
-whitelist = nodejs-react-json-tree
+whitelist = ansiblerole-insights-client
+  nodejs-react-json-tree
   puppet-foreman_scap_client
   rubygem-chef-api
   rubygem-faraday


packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,56 @@
...
+
+Requires: ansible
+
+%description
+This package installs the insights-client Ansibile role.
+
+Make sure that "/usr/share/ansible/roles" is on your Ansible role_path.
+
+%prep
+%setup -qc
```
ekohl: Could you add a newline after this to separate sections?

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,56 @@
...
+%build
+%install
+mkdir -p $RPM_BUILD_ROOT%{_datadir}/ansible/roles
```
ekohl: In our other specs we use ```%{buildroot}```

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,56 @@
...
+
+cp -pR %{repo_name}-%{role_version} $RPM_BUILD_ROOT%{_datadir}/ansible/roles/%{role_orgname}.%{role_name}
+
+mkdir -p $RPM_BUILD_ROOT%{_pkgdocdir}/
+
+cp -p $RPM_BUILD_ROOT%{_datadir}/ansible/roles/%{role_orgname}.%{role_name}/README.md \
+      $RPM_BUILD_ROOT%{_datadir}/ansible/roles/%{role_orgname}.%{role_name}/LICENSE \
```
ekohl: I'm not sure this is needed if you already add it via ```%license```.

ekohl:
[test packaging]

ekohl:
[test]

ares:
> If I'm honest I'm not sure I like the idea of packaging Ansible Roles in RPMs.

We need to deliver the configuring piece somehow (see why below), what would be a better way?

> Considering this is probably going to be used with Katello (?) I wounder if we should have our installer configure Pulp to host this role using the Pulp file type repo for now, maybe using pulp_ansible when Pulp 3.0 is released. A user would then be able to use ansible-galaxy to download and use it.

Yeah, sorry for not being clear about the purpose. This is a role that configures insights client on a host. So far we only provided puppet module to do so but with upcoming ansible support, it makes sense to provide similar functionality for ansible users who do not have puppet. It's not required for Katello but rather [redhat_access](https://github.com/redhataccess/foreman-plugin/) plugin. It's the same model we have for foreman_openscap, see https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/plugins/puppet-foreman_scap_client for which we should provide ansible alternative at some point too.

I hope to get the redhat_access plugin packaged here soon too and then integrate it into the installer. At the end, if user allows this plugin they should have the role available in their "ansible master" and they should be able to configure clients the same way they do it today with puppet modules.

sean797:
right, that makes more sense. Thanks @ares :+1:

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
gavin-romig-koch: Yes.  Will do.

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
gavin-romig-koch: OK, I'll change it.

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
gavin-romig-koch: RPM has one convention for where to put package Licence, README, and examples, and Ansible Roles packages have a different convention for where to put them.   

Rather than honor one convention over the other, i choose to install the files into both places.  I'd rather leave it as is, unless someone can thing of a problem this causes.

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,56 @@
+
+%global repo_orgname RedHatInsights
+%global repo_name insights-client-role
+%global repo_version 1.5
+
+%global role_orgname %{repo_orgname}
+%global role_name insights-client
+%global role_version %{repo_version}
```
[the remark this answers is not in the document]
gavin-romig-koch: The repo_xxx vars are git repo names, the role_xxx vars are the names you want installed on disk.

In this case the repo_name (insights-client-role) must be different from the role_name (insights-client).  The repo_name must be insights-client-role because the git repo insight-client contains the actual code that this role installs.  But we don't want the role's name to be insights-client-role, that would be redundant and silly.  Galaxy allows you to change the role's name as you import it, and we are just echoing that functionality here.

As for {repo,role}_orgname.  For this spec file i followed the design of the rhel-system-roles.spec which had this same abiltiy to have a different orgname/namespace from the repo to the role.  In my case i felt the RedHatInsights org name of the repo worked find as an Namespace for the role, but i left the possiblity of changing it in case we needed to.

As for the {repo,role}_version, i don't see any reason to change the version from repo to role, and have to admit i just continued the pattern without thinking about it.

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,56 @@
...
+
+Name: ansiblerole-insights-client
+Summary: Packaging of the insights-client Ansible role
+Version: %{role_version}
+Release: 1%{?dist}
+
+License: ASL 2.0
+
+Source: https://github.com/%{repo_orgname}/%{repo_name}/archive/v%{repo_version}.tar.gz#/%{role_name}-%{role_version}.tar.gz
```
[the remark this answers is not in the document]
gavin-romig-koch: Sorry i don't see a commented section.  ??

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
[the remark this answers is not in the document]
sean797: Sorry, i read this wrong, `role` v `repo` :man_facepalming:

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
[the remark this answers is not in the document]
sean797: %{repo_orgname}/%{repo_name}/archive/v%{repo_version}.tar.gz **#/%{role_name}-%{role_version}.tar.gz**"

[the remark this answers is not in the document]
gavin-romig-koch: Oh, that.  That isn't actually a comment.

That construction is used to change/specify the name of the downloaded file that is written to disk.  The name before the '#' is what is actually GET'ed, the name after the '#" is what is actually written to disk.  In this case GitHub actually serves up the tarball I want as .../v1.5.tar.gz, but i want something more rational as the name of the file on disk: insights-client-1.5.tar.gz.    This convention is described in the RPM specfile documentation, and implemented in, for example, spectool.

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
gavin-romig-koch: fixed

gavin-romig-koch:
This push fixes all the issues i said i'd fix.

ekohl:
ok to build

ares:
ok to test
perhaps :-)

ekohl:
The build failure is because you didn't add it to `package_manifest.yaml`. You also need to add the package to `rel-eng/tito.props`.

I did a manual build @ http://koji.katello.org/koji/taskinfo?taskID=91596 to see the result.

ekohl:
https://github.com/ekohl/foreman-packaging/commit/d3e7a70184306f2f311ee6b6fdccbc73ecdc69f8 are the changes I made to make it build in the most minimal way. See http://koji.katello.org/koji/taskinfo?taskID=91604 for the result. These files are generated for me:
```
drwxr-xr-x    2 root    root                        0 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client
-rw-r--r--    1 root    root                    10175 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/LICENSE
-rw-r--r--    1 root    root                     4759 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/README.md
drwxr-xr-x    2 root    root                        0 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/examples
-rw-r--r--    1 root    root                       66 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/examples/example-insights-client-playbook.yml
drwxr-xr-x    2 root    root                        0 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/files
-rwxr-xr-x    1 root    root                      886 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/files/insights.fact
drwxr-xr-x    2 root    root                        0 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/meta
-rw-r--r--    1 root    root                      325 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/meta/main.yml
drwxr-xr-x    2 root    root                        0 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/tasks
-rw-r--r--    1 root    root                     3228 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/tasks/main.yml
drwxr-xr-x    2 root    root                        0 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/tests
lrwxrwxrwx    1 root    root                       48 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/tests/example-insights-client-playbook.yml -> ../examples/example-insights-client-playbook.yml
drwxr-xr-x    2 root    root                        0 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/tests/roles
lrwxrwxrwx    1 root    root                        5 mrt 15 01:48 /usr/share/ansible/roles/RedHatInsights.insights-client/tests/roles/RedHatInsights.insights-client -> ../..
drwxr-xr-x    2 root    root                        0 mrt 28 21:41 /usr/share/doc/ansiblerole-insights-client-1.5
-rw-r--r--    1 root    root                     4759 mrt 15 01:48 /usr/share/doc/ansiblerole-insights-client-1.5/README.md
-rw-r--r--    1 root    root                       66 mrt 15 01:48 /usr/share/doc/ansiblerole-insights-client-1.5/example-insights-client-playbook.yml
drwxr-xr-x    2 root    root                        0 mrt 28 21:41 /usr/share/licenses/ansiblerole-insights-client-1.5
-rw-r--r--    1 root    root                    10175 mrt 15 01:48 /usr/share/licenses/ansiblerole-insights-client-1.5/LICENSE
```

A more minimal packaging would also strip out `README.md`, `LICENSE` and `examples/*.yml` from the `roles` directory since they're already in other places.

ekohl:
@ehelms could you have a look as well? I feel like I'd be reviewing my own work here, which is always awesome ;)

ehelms:
Does this package only work on EL7? If not, do you want support on EL5/EL6/F26/F27? If so, I'd consider putting this into the Katello client repositories instead of the plugin repositories. This would also be a good example for moving the client repositories to be Foreman client repositories.

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,47 @@
...
+%global repo_name insights-client-role
+%global repo_version 1.5
+
+%global role_orgname %{repo_orgname}
+%global role_name insights-client
+%global role_version %{repo_version}
+
+Name: ansiblerole-insights-client
+Summary: Packaging of the insights-client Ansible role
+Version: %{role_version}
```
ehelms: Normally you just define the version here and RPM makes a `%{version}` macro available for reference.

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,47 @@
+
+%global repo_orgname RedHatInsights
+%global repo_name insights-client-role
+%global repo_version 1.5
+
+%global role_orgname %{repo_orgname}
+%global role_name insights-client
+%global role_version %{repo_version}
+
+Name: ansiblerole-insights-client
```
ehelms: This seems sensible for naming. The only other example I could find was system roles packaging where they used `rhel-system-roles`.

https://access.redhat.com/articles/3050101

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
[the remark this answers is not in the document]
ehelms: I think this is typically `Source0`

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,47 @@
...
+Summary: Packaging of the insights-client Ansible role
+Version: %{role_version}
+Release: 1%{?dist}
+
+License: ASL 2.0
+
```
ehelms: Nitpick -- you could drop this empty line

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,47 @@
...
+%global role_orgname %{repo_orgname}
+%global role_name insights-client
+%global role_version %{repo_version}
+
+Name: ansiblerole-insights-client
+Summary: Packaging of the insights-client Ansible role
+Version: %{role_version}
+Release: 1%{?dist}
+
```
ehelms: Nitpick -- you could drop this empty line

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,47 @@
...
+License: ASL 2.0
+
+Source: https://github.com/%{repo_orgname}/%{repo_name}/archive/v%{repo_version}.tar.gz#/%{role_name}-%{role_version}.tar.gz
+
```
ehelms: Nitpick -- you could drop this empty line

ehelms:
This will need to also be added to the appropriate comps file so it ends up in the repository.

gavin-romig-koch:
In response to @ehelms > Does this package only work on EL7? If not, do you want support on EL5/EL6/F26/F27?

I have my opinion on this, which I detail below, but I leave it to you all, the Forman and Katello developers, to decide what the right thing do here is.

It is slightly more complicated than is usual for a package.  This package installs an Ansible role.  The role itself installs a package that is only available and supported on RHEL6, RHEL7, and is a client (insights-client) that talks to a service (Red Hat Insights) that is only available to paying Red Hat customers using RHEL.  (This is an unfortunate situation for us to be in, but at least for now it is the situation.)

I would rather this package/role didn't go into any distro other than RHEL for that reason.  I feel like Fedora, Centos, etc users might see the package/role, think it's interesting, and after installing it discover that it installs a service they can't use without paying for it (and installing RHEL).  There is no legal or technical reason we couldn't put this role/package into Fedora or Centos or any distro that has Ansible, so I wouldn't want to do significant extra work to keep it out of other distros, but at the same time I don't want to do extra work to put it into other distros.   

RHEL6 is out for now because it does not yet have officially supported RPMs for Ansible.  I understand such RPMs are on the way, but are not yet available to RHEL6 customers.

gavin-romig-koch:
@ekohl, i'm sorry but i don't know what i've left undone that is causing the build to fail.  'obal check ansiblerole-insights-client' runs fine on my machine, and i can't run 'obal scratch' because I don't have koji build permissions.  But from the error message, it looks like the name ansiblerole-insights-client is just not in some config file that it is supposed to be in.

package_manifest.yaml
```diff
@@ -312,9 +312,11 @@ plugins_packages:
   vars:
     package_base_dir: 'packages/plugins/'
     releasers:
-    - koji-foreman-plugins
-  hosts:
+    - koji-foreman-plugin
```
ekohl: Could you revert this change? This breaks obal

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,45 @@
...
```
ekohl: Could you remove this empty line?

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,45 @@
...
+%global repo_orgname RedHatInsights
+%global repo_name insights-client-role
+%global repo_version %{version}
+
+%global role_orgname %{repo_orgname}
+%global role_name insights-client
+%global role_version %{repo_version}
+
+Source: https://github.com/%{repo_orgname}/%{repo_name}/archive/v%{repo_version}.tar.gz#/%{role_name}-%{role_version}.tar.gz
```
ekohl: Could you change this to `Source0`?

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
gavin-romig-koch: yes

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
gavin-romig-koch: yes

package_manifest.yaml
gavin-romig-koch: yes

ares:
ok to test

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
ekohl: (this is old and github doesn't detect this was from another line)

ares:
seems it build fine :-) http://koji.katello.org/kojifiles/work/tasks/3678/93678/ansiblerole-insights-client-1.5-1.el7.noarch.rpm

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,44 @@
...
+
+%global repo_orgname RedHatInsights
+%global repo_name insights-client-role
+%global repo_version %{version}
+
+%global role_orgname %{repo_orgname}
+%global role_name insights-client
+%global role_version %{repo_version}
+
+Source0: https://github.com/%{repo_orgname}/%{repo_name}/archive/v%{repo_version}.tar.gz#/%{role_name}-%{role_version}.tar.gz
```
ehelms: Can this just use %{version} ? The `repo_version` macro seems redundant.

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,44 @@
...
+Release: 1%{?dist}
+License: ASL 2.0
+
+%global repo_orgname RedHatInsights
+%global repo_name insights-client-role
+%global repo_version %{version}
+
+%global role_orgname %{repo_orgname}
+%global role_name insights-client
+%global role_version %{repo_version}
```
ehelms: This seems redundant given `%{version}`

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,44 @@
+Name: ansiblerole-insights-client
+Summary: Packaging of the insights-client Ansible role
+Version: 1.5
+Release: 1%{?dist}
+License: ASL 2.0
+
+%global repo_orgname RedHatInsights
+%global repo_name insights-client-role
+%global repo_version %{version}
```
ehelms: General comment, typically globals are defined at the top of the spec file.

ares:
@ehelms mind to take another look, hopefully it's ready now

ekohl:
ok to test

ares:
any progress with reviewing? it seems it's ready to merge, I've tested the scratch some time ago and it works greatly with our ansible integration

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
```diff
@@ -0,0 +1,42 @@
+%global repo_orgname RedHatInsights
+%global repo_name insights-client-role
+
+%global role_orgname %{repo_orgname}
+%global role_name insights-client
+
+Name: ansiblerole-insights-client
```
evgeni: this could use `%{role_name}` too?

evgeni:
From a packaging perspective, this looks fine and mergeable.

I do wonder about the package name, though. When looking at other roles (as packaged by Red Hat), the pattern seems to be `ansible-%{role_name}` or `%{role_name}-ansible` instead of `ansiblerole-%{role_name}`. I could not fine any guidelines for official naming. I am find with either name, would just like to prevent renames in the future :)

packages/plugins/ansiblerole-insights-client/ansiblerole-insights-client.spec
gavin-romig-koch: Sure.

gavin-romig-koch:
In response to: 
  > I do wonder about the package name, though.  ....

I don't think a pattern/convention/standard for naming role packages has emerged yet, though I think establishing one is worthwhile.  Having such a convention for naming packages that re-package roles, makes it easier to find the RPM that contains that role.  I made this choice based on what I feel is sound reasoning, but I would rather have many people weigh in on this decision.

I choose to follow the pattern of 'rubygem-XXX', rather than that of 'python-XXX'.  

In general, though not always, packages that re-package ruby gems are called 'rubygem-XXX', packages that implement, or extend the implementation of Ruby are called 'ruby-XXX'.   Not always, but mostly.

On the other hand there are no packages named 'pythonpip-XXXX' or 'pythonwheel-XXXX', and both packages that implement Python and packages that simply repackage pip packages, are named 'python-XXXX'.

Both conventions are reasonable for the language/enviroment they were made for.  Python has had several packaging formats and packaging managers, and show every sign that that situation will continue.   Ruby, as  far as I know, only has gems, though I have to admit being far less familiar with Ruby than Python.

For Ansible I felt that the relationship between Ansible the language, and Ansibles packaging format (Roles) was more like that of Ruby and Gems rather than Python and it's packaging format.  

I believe that the Red Hat (and it's associated communities) have packaged very, very, few packages that were just re-packageing of existing roles.  Almost all of the existing packages with names like 'ansible-XXXX' or 'XXX-ansible' are packages that implement or extend the Ansible language and environment, not re-package existing 'roles'.

ekohl:
To me that explanation makes sense. `ansible-*` sounds like a subpackage of ansible like a core module. `ansiblerole-*` or `ansible-role-*` is explicit. @evgeni?

evgeni:
Well, I'd argue, that given a role can contain modules, it can also extend "Ansible the language" and by that should be called `ansible-%{role_name}`, but I am not going to block this review on that, so ACK, whatever name we have :)

The mentioned discussion and guidelines should happen, but this PR is surely the wrong place for both (at least in depth).

ares:
so... let's get it in as it is? :-)

ehelms:
I do like the small enhancement of `ansible-role` as it reads cleaner, but also not going to block on this (though I anticipate we will run into this again when @ares produces a role for openscap similar to the existing puppet-foreman_scap_client)

ares:
I guess we can rename the package if it turns out it's misleading. Seeing everyone saying not-blocking, could we merge and build?

PR #2287, opened 2018-03-16
Set TMPDIR in repoclosure.sh

ekohl:
Repoclosure creates $TMPDIR/$repo which means if multiple processes share the same $TMPDIR you get concurrency issues when they share repos (like el7-epel). Even if the actual repo definitions are the same, then one process might delete a repo the other is still using. By setting a custom TMPDIR we are less efficient but guarantee it does work.

Note that I didn't test this theory yet. Can we replicate the Jenkins setup or should we just apply this?

### DIFF
--- a/repoclosure/repoclosure.sh
+++ b/repoclosure/repoclosure.sh
@@ -11,13 +11,13 @@ yumorig=$1
 url=$2
 shift; shift
 
-TEMPDIR=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


ekohl:
Looking deeper into this it turns out that the `--tempcache` (`-t`) doesn't look at this and just reuses any existing dir matching `/var/tmp/yum-$user-*`

PR #2289, opened 2018-03-19
foreman_maintain pkg should provide foreman-maintain

sean797:
Users call this pkg foreman-maintain, as the main thing it provides is a binary named foreman-maintain, however it is called foreman_maintain.

This commit makes the following work:
- yum install foreman-maintain
- yum install foreman_maintain

We should probably address the confusion properly in another PR at some point by either renaming the binary or renaming the package.

* [x] Nightly

---

### DIFF
--- a/packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec
+++ b/packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec
@@ -9,7 +9,7 @@
 Summary: The Foreman/Satellite maintenance tool
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.1.5
-Release: 1%{?dist}
+Release: 2%{?dist}
 Group: Development/Languages
 License: GPLv3
 URL: https://github.com/theforeman/foreman_maintain
@@ -25,6 +25,8 @@ BuildRequires: %{?scl_prefix_ruby}ruby(rubygems)
 BuildRequires: %{?scl_prefix_ruby}ruby
 BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
+Provides: %{gem_name} = %{version}
+Provides: foreman-maintain = %{version}
 
 %description
 foreman_maintain aims to provide various features that helps keeping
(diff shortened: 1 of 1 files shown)


packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec
```diff
@@ -25,6 +25,8 @@ BuildRequires: %{?scl_prefix_ruby}ruby(rubygems)
 BuildRequires: %{?scl_prefix_ruby}ruby
 BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
+Provides: %{pkg_name} = %{version}
```
ekohl: I don't know everything about RPM packaging but this feels redundant. Shouldn't it actually be ```%{gem_name}``` if you want to make ```foreman_maintain``` work? From the koji build log:
```
Provides: config(rubygem-foreman_maintain) = 0.1.3-1.el7 foreman-maintain = 0.1.3 rubygem(foreman_maintain) = 0.1.3 rubygem-foreman_maintain = 0.1.3 rubygem-foreman_maintain = 0.1.3-1.el7
```

sean797: ah, yes, thanks @ekohl :+1: updated

ekohl: That's what I would expect:

```
Provides: config(rubygem-foreman_maintain) = 0.1.3-1.el7 foreman-maintain = 0.1.3 foreman_maintain = 0.1.3 rubygem(foreman_maintain) = 0.1.3 rubygem-foreman_maintain = 0.1.3-1.el7
```

evgeni:
Yeah, I constantly end up doing `yum search foreman maintain` followed by `yum install rubygem-foreman_maintain`, so :+1: on that change. No idea if we need to provide both, hyphen and underscore (hyphen would suffice for me), but it also does not hurt at all.

ehelms:
Please bump the release.

sean797:
bumped, thanks :+1:

sean797:
Add changelog entry :+1:

ekohl:
Thanks!

PR #2291, opened 2018-03-20
[WIP] Implement generating dependencies while bumping gems

ekohl:
This extends the update-node-dependencies model to be generic. It can now parse files with begin and end markers to get sections. This can then be inserted in another file with matching markers.

The intended use case is calling gem2rpm to generate new dependencies and then inserting those in the trivially bumped specfile. This allows automating bumping while keeping manual changes.

To make this easier we introduce macros to set the minimum foreman(-proxy) versions. While introducing those it turned out there were some inconsistencies in the plugins.

What's missing from this PR is a way to determine the gem2rpm template. For this we could introduce it as a comment in the file which you can grep. An alternative is package_manifest.yaml but that's harder to parse in shell scripts.


ehelms:
This all seems good to me

ekohl:
Updated to grep for the template name in a comment. I'll need to split up the actual updating of spec files into separate commits since at least ```foreman_chef``` doesn't build with the current foreman.

ekohl:
Updated:
* I left out all the spec updates. I'll submit them separately and do more thorough spec updates. Once those get in, it should be able to use the new updating infrastructure. Unmodified specs shouldn't be affected.
* Added a file to get build deps from a Gemfile. This is very raw and not ready to be integrated into the toolchain, but it's already useful to parse the foreman deps and manually paste them in the correct sections. https://github.com/theforeman/foreman-packaging/commit/111cd7e7a206bf510cc5d969bbe576211a907483 is the result of such a run.

ekohl:
Putting it back to WIP. The markers in foreman.spec for both package.json and Gemfile confuses the update script. Need to refactor it further.

ekohl:
Updated. This now should be able to handle bumping dependencies in specfiles. The most notable changes are in the README and `bump_rpm.sh`. The rest implementation details.

My intention is to set up a regular job that automatically creates PRs for this - with some intermediate steps that controls it :)

README.md
```diff
@@ -139,6 +139,25 @@ required non-SCL modes.
...
+#### Foreman
+
+To update foreman we'll assume a git checkout at `~/foreman`:
+
+    ./get-gemfile-deps ~/foreman/Gemfile | ./update-requirements - packages/foreman/foreman/foreman.spec specfile
+    ./update-requirements ~/foreman/package.json packages/foreman/foreman/foreman.spec npm
+
+#### Katello
+
+To update foreman we'll assume a git checkout at `~/foreman`:
```
evgeni: `s/foreman/katello/`

bump_rpm.sh
```diff
@@ -64,13 +64,19 @@ if [[ $CURRENT_VERSION != $NEW_VERSION ]] ; then
 	git annex add *.gem
 	git add $SPEC_FILE
 
-	echo "TODO:"
 	if [[ $PACKAGE_NAME == rubygem-* ]] ; then
```
evgeni: should this also match `tfm-rubygem-*`?

ekohl: There is one case where we have expanded the package name in the directory name so yes.

update-requirements
```diff
@@ -81,29 +103,42 @@ def get_new_content(current_content, requires):
...
-    parser = argparse.ArgumentParser(description='Update the nodejs dependencies from package.json')
-    parser.add_argument('packagesfile', help='Path to the packages file',
-                        type=argparse.FileType('r'))
-    parser.add_argument('specfile', help='path to the specfile', default='foreman.spec', nargs='?',
+    parser = argparse.ArgumentParser(description='Update the dependencies within sections')
+    parser.add_argument('requirementsfile', help='Path to the requirements file',
                         type=argparse.FileType('r'))
+    parser.add_argument('specfile', help='path to the specfile', type=argparse.FileType('r'))
+    parser.add_argument('requirementstype', help='the type of conversion',
+                        choices=('npm', 'specfile', 'direct'))
```
evgeni: I would make the usage `update-requirements <requirementstype> <specfile> <requirementsfile>`, not the other way round, but I'd also not block on such a nice :bike:

evgeni:
:+1:

ekohl:
Updated to address the above issues.

get-gemfile-deps
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env ruby
```
ehelms: We have quite the mix of python and ruby and bash scripts :) I think this is OK for now as long as we are heading towards one set of code.

PR #2295, opened 2018-03-20
Create proxy etc and var directories

dLobatog:
@ekohl I recall you mentioned something about migrating files in packaging, if it was about https://github.com/theforeman/puppet-foreman_proxy/pull/415 ,  `ansible.cfg`, `.ansible`, and `.ansible_galaxy` are good candidates. But also `.ssh` as that's likely setup by REX.

I think the foreman-proxy package itself is probably not the best place to do the migration, in fact I'm not sure if moving files around during `yum install` is going to be expected by any user. Shouldn't that belong in the installer? I'm OK with doing it here, but I think it's really unexpected to see files disappear during `yum install` - whereas the installer at least logs that clearly.

If we have to migrate any files in `/usr/share/foreman-proxy`, I would say that should be done by the  `foreman_ansible` and `foreman_remote_execution` packages, instead of this one.

### DIFF
--- a/packages/foreman/foreman-proxy/foreman-proxy.spec
+++ b/packages/foreman/foreman-proxy/foreman-proxy.spec
@@ -140,6 +140,7 @@ ln -sv %{_tmppath} %{buildroot}%{_datadir}/%{name}/tmp
 %config(noreplace) %{_sysconfdir}/%{name}
 %config(noreplace) %{_sysconfdir}/logrotate.d/%{name}
 %attr(-,%{name},%{name}) %{_localstatedir}/log/%{name}
+%attr(-,%{name},%{name}) %{_localstatedir}/lib/%{name}
 %attr(-,%{name},%{name}) %{_rundir}/%{name}
 %attr(-,%{name},root) %{_datadir}/%{name}/config.ru
 %exclude %{_datadir}/%{name}/bundler.d/development.rb
--- a/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
+++ b/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
@@ -52,7 +52,26 @@ This package contains documentation for rubygem-%{gem_name}.
 
 %build
 
+%pre
+for i in ansible ansible_galaxy; do
+  if [ -d %{foreman_proxy_dir}/.$i ]; then
+    mv %{foreman_proxy_dir}/.$i %{buildroot}%{_localstatedir}/lib/foreman-proxy/$i
+  else
+    mkdir -p %{buildroot}%{_localstatedir}/lib/foreman-proxy/$i
+  fi
+done
+
+if [-f %{foreman_proxy_dir}/.ansible.cfg ]; then
+  mv %{foreman_proxy_dir}/.ansible.cfg %{buildroot}%{_sysconfdir}/foreman-proxy/ansible.cfg
+else
+  touch %{buildroot}%{_sysconfdir}/foreman-proxy/ansible.cfg
+fi
+
 %install
+ln -sv %{buildroot}%{_localstatedir}/lib/foreman-proxy/$i %{buildroot}%{foreman_proxy_dir}/.ansible
+ln -sv %{buildroot}%{_localstatedir}/lib/foreman-proxy/$i %{buildroot}%{foreman_proxy_dir}/.ansible_galaxy
+ln -sv %{buildroot}%{_sysconfdir}/foreman-proxy/ansible.cfg %{buildroot}%{foreman_proxy_dir}/.ansible.cfg
+
 mkdir -p %{buildroot}%{gem_dir}
 
 cp -pa .%{gem_dir}/* \
@@ -75,6 +94,12 @@ EOF
 %{gem_instdir}/settings.d
 %{foreman_proxy_bundlerd_dir}/smart_proxy_ansible.rb
 %config %{foreman_proxy_settingsd_dir}/ansible.yml
+%attr(-,root,root) %{foreman_proxy_dir}/.ansible
+%attr(-,root,root) %{foreman_proxy_dir}/.ansible_galaxy
+%attr(-,root,root) %{foreman_proxy_dir}/.ansible.cfg
+%attr(-,foreman-proxy,foreman-proxy) %{_localstatedir}/lib/foreman_proxy/ansible
+%attr(-,foreman-proxy,foreman-proxy) %{_localstatedir}/lib/foreman_proxy/ansible_galaxy
+%attr(-,root,foreman-proxy) %{_sysconfdir}/foreman-proxy/ansible.cfg
 %doc %{gem_instdir}/LICENSE
 %{smart_proxy_dynflow_bundlerd_dir}/foreman_ansible_core.rb
 
--- a/packages/plugins/rubygem-smart_proxy_remote_execution_ssh/rubygem-smart_proxy_remote_execution_ssh.spec
+++ b/packages/plugins/rubygem-smart_proxy_remote_execution_ssh/rubygem-smart_proxy_remote_execution_ssh.spec
@@ -1,5 +1,4 @@
 %global gem_name smart_proxy_remote_execution_ssh
-
 %global foreman_proxy_dir /usr/share/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_sysconfdir}/foreman-proxy/settings.d
@@ -56,7 +55,16 @@ This package contains documentation for rubygem-%{gem_name}.
 
 %build
 
+%pre
+if [ -d %{foreman_proxy_dir}/.ssh ]; then
+  mv %{foreman_proxy_dir}/.ssh %{_localstatedir}/lib/foreman-proxy/ssh
+else
+  mkdir -p %{_localstatedir}/lib/foreman-proxy/ssh
+fi
+
 %install
+ln -sv %{_localstatedir}/lib/foreman-proxy/ssh %{foreman_proxy_dir}/.ssh
+
 mkdir -p %{buildroot}%{gem_dir}
 
 cp -pa .%{gem_dir}/* \
@@ -81,7 +89,8 @@ EOF
 %config %{foreman_proxy_settingsd_dir}/remote_execution_ssh.yml
 %doc %{gem_instdir}/LICENSE
 %{smart_proxy_dynflow_bundlerd_dir}/foreman_remote_execution_core.rb
-
+%attr(-,root,root) %{foreman_proxy_dir}/.ssh
+%attr(-,foreman-proxy,foreman-proxy) %{_localstatedir}/lib/foreman_proxy/ssh
 %exclude %{gem_instdir}/bundler.plugins.d
 %exclude %{gem_cache}
 %{gem_spec}


packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -117,7 +117,12 @@ cp -p -r bin extra lib modules Rakefile Gemfile.in smart_proxy.gemspec bundler.d
 chmod a+x %{buildroot}%{_datadir}/%{name}/bin/smart-proxy
 rm -rf %{buildroot}%{_datadir}/%{name}/*.rb
 
+# Create etc and var dirs
+%{__mkdir_p} %{buildroot}/etc/foreman-proxy
```
ekohl: This should already be covered by line 101

packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -117,7 +117,12 @@ cp -p -r bin extra lib modules Rakefile Gemfile.in smart_proxy.gemspec bundler.d
...
+%{__mkdir_p} %{buildroot}%{_var}/lib/foreman-proxy
```
ekohl: This should already be covered by line 103, yet on my 1.16.0 proxy it doesn't exist.

dLobatog:
`/var/lib/foreman-proxy` exists on my proxy (1.17), but I think it's because of `smart_proxy_dynflow`. It calls `mkdir -p %{buildroot}%{_localstatedir}/lib/foreman-proxy/dynflow `, and in that case it works. The only difference is using `install -d` vs `mkdir -p`, not sure why it's not working in this case.

packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -118,6 +118,7 @@ chmod a+x %{buildroot}%{_datadir}/%{name}/bin/smart-proxy
 rm -rf %{buildroot}%{_datadir}/%{name}/*.rb
 
 # remove all test units from productive release
+#
```
ekohl: Could you remove this line?

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -86,6 +86,20 @@ EOF
 %doc %{gem_instdir}/README.md
 %doc %{gem_docdir}
 
+%pre
+if [ -d %{foreman_proxy_dir}/.ansible ]
+  mv %{foreman_proxy_dir}/.ansible /var/lib/foreman-proxy/ansible
+  ln -sv /var/lib/foreman-proxy/ansible %{foreman_proxy_dir}/.ansible
```
ekohl: I think this symlink (and others) should be in ```%install``` and ```%files```. See https://github.com/theforeman/foreman-packaging/blob/d09a1169ef51b8854f873f203a9b6aca54baddf8/packages/foreman/foreman/foreman.spec#L1000-L1003 as well. Because this is in ```%pre``` I think it should be installed after this hook is ran.

dLobatog:
Latest update:

 - moved %pre to %install 
 - added `/var/lib/foreman-proxy` to the `%files` section of foreman-proxy.spec
 - added all other files to their respective files section

I haven't tested yet, but my main doubt is how is the %files section meant to be filled when a file might or might not be migrated on %install? I meant to solve this by just `touching` the files and ensuring they're there even if they were not before.

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -75,6 +91,12 @@ EOF
 %{gem_instdir}/settings.d
 %{foreman_proxy_bundlerd_dir}/smart_proxy_ansible.rb
 %config %{foreman_proxy_settingsd_dir}/ansible.yml
+%attr(-,foreman-proxy,foreman-proxy) %{foreman_proxy_dir}/.ansible
+%attr(-,foreman-proxy,foreman-proxy) %{foreman_proxy_dir}/.ansible_galaxy
+%attr(-,foreman-proxy,foreman-proxy) %{foreman_proxy_dir}/.ansible.cfg
+%attr(-,foreman-proxy,foreman-proxy) %{_localstatedir}/lib/foreman_proxy/ansible
+%attr(-,foreman-proxy,foreman-proxy) %{_localstatedir}/lib/foreman_proxy/ansible_galaxy
+%attr(-,foreman-proxy,foreman-proxy) /etc/foreman-proxy/ansible.cfg
```
ekohl: For etc we have ``````%{_sysconfdir}```. I also think this should be owned by root. That means you can leave it out. It should be using ```%config``` though.

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -75,6 +91,12 @@ EOF
 %{gem_instdir}/settings.d
 %{foreman_proxy_bundlerd_dir}/smart_proxy_ansible.rb
 %config %{foreman_proxy_settingsd_dir}/ansible.yml
+%attr(-,foreman-proxy,foreman-proxy) %{foreman_proxy_dir}/.ansible
```
ekohl: Files in ```/usr``` should be owned by root.

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -53,6 +53,22 @@ This package contains documentation for rubygem-%{gem_name}.
 %build
 
 %install
+for i in ansible ansible_galaxy; do
+if [ -d %{foreman_proxy_dir}/.$i ]; then
+  mv %{foreman_proxy_dir}/.$i %{_localstatedir}/lib/foreman-proxy/$i
```
ekohl: Moving this should happen in ```%pre``` because that step happens on the actual system. ```%install``` happens while building in a chroot. Only adding the symlink should happen in ```%install```

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -53,6 +53,22 @@ This package contains documentation for rubygem-%{gem_name}.
...
+else
+  mkdir -p %{_localstatedir}/lib/foreman-proxy/$i
+fi
+  ln -sv %{_localstatedir}/lib/foreman-proxy/$i %{foreman_proxy_dir}/.$i
+done
+
+if [-d %{foreman_proxy_dir}/.ansible.cfg ]; then
```
ekohl: If it should be a dir then ```-f``` should be used instead of ```-d```

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
dLobatog: How would `foreman-proxy` be able to read it if it is owned by root?

ekohl: By default it's 644. If there's a password in there ```%attr(0640,root,foreman-proxy)``` might be a better solution but you're right ```root:foreman-proxy``` is better in any case.

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -52,7 +52,26 @@ This package contains documentation for rubygem-%{gem_name}.
...
+done
+
+if [-f %{foreman_proxy_dir}/.ansible.cfg ]; then
+  mv %{foreman_proxy_dir}/.ansible.cfg %{buildroot}%{_sysconfdir}/foreman-proxy/ansible.cfg
+else
+  touch %{buildroot}%{_sysconfdir}/foreman-proxy/ansible.cfg
+fi
+
 %install
+ln -sv %{buildroot}%{_localstatedir}/lib/foreman-proxy/$i %{buildroot}%{foreman_proxy_dir}/.ansible
```
ekohl: Here `$i` is not defined (and the line below as well).

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -75,6 +94,12 @@ EOF
 %{gem_instdir}/settings.d
 %{foreman_proxy_bundlerd_dir}/smart_proxy_ansible.rb
 %config %{foreman_proxy_settingsd_dir}/ansible.yml
+%attr(-,root,root) %{foreman_proxy_dir}/.ansible
+%attr(-,root,root) %{foreman_proxy_dir}/.ansible_galaxy
+%attr(-,root,root) %{foreman_proxy_dir}/.ansible.cfg
+%attr(-,foreman-proxy,foreman-proxy) %{_localstatedir}/lib/foreman_proxy/ansible
+%attr(-,foreman-proxy,foreman-proxy) %{_localstatedir}/lib/foreman_proxy/ansible_galaxy
+%attr(-,root,foreman-proxy) %{_sysconfdir}/foreman-proxy/ansible.cfg
```
ekohl: This should be marked as a config file so it's not overwritten on upgrades. Maybe also good to set the attr to `0640`?

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -52,7 +52,26 @@ This package contains documentation for rubygem-%{gem_name}.
...
+    mv %{foreman_proxy_dir}/.$i %{buildroot}%{_localstatedir}/lib/foreman-proxy/$i
+  else
+    mkdir -p %{buildroot}%{_localstatedir}/lib/foreman-proxy/$i
+  fi
+done
+
+if [-f %{foreman_proxy_dir}/.ansible.cfg ]; then
+  mv %{foreman_proxy_dir}/.ansible.cfg %{buildroot}%{_sysconfdir}/foreman-proxy/ansible.cfg
+else
+  touch %{buildroot}%{_sysconfdir}/foreman-proxy/ansible.cfg
```
ekohl: I think you could leave this out and place an example file in `%install`. I'll need to verify this.

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -52,7 +52,26 @@ This package contains documentation for rubygem-%{gem_name}.
 
 %build
 
+%pre
+for i in ansible ansible_galaxy; do
+  if [ -d %{foreman_proxy_dir}/.$i ]; then
+    mv %{foreman_proxy_dir}/.$i %{buildroot}%{_localstatedir}/lib/foreman-proxy/$i
+  else
+    mkdir -p %{buildroot}%{_localstatedir}/lib/foreman-proxy/$i
```
ekohl: This should happen in `%install`

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -52,7 +52,26 @@ This package contains documentation for rubygem-%{gem_name}.
...
+  fi
+done
+
+if [-f %{foreman_proxy_dir}/.ansible.cfg ]; then
```
ekohl: Could you add a space between `[` and `-f`? I think this is invalid syntax.

ekohl:
Replacement PR https://github.com/theforeman/foreman-packaging/pull/2348

PR #2305, opened 2018-03-22
[WIP] Move katello to rh-mongodb34

ehelms:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -6,7 +6,7 @@
 
 Name:       katello
 Version:    3.7.0
-Release:    2.nightly%{?dist}
+Release:    3.nightly%{?dist}
 Summary:    A package for managing application life-cycle for Linux systems
 BuildArch:  noarch
 
@@ -51,8 +51,7 @@ Requires: pulp-puppet-tools
 Requires: pulp-selinux
 Requires: pulp-server
 Requires: python-pulp-streamer
-Requires: mongodb >= 2.4
-Requires: mongodb-server >= 2.4
+Requires: rh-mongodb34
 Requires: cyrus-sasl-plain
 Requires: python-crane
 Requires: python-gofer-qpid
@@ -70,6 +69,8 @@ Requires: java-openjdk < 1:1.8.0.45
 Requires: /usr/bin/psql
 Requires: /usr/bin/pg_dump
 Requires: /usr/bin/pg_dumpall
+
+Obsoletes: mongodb
 
 %description
 Provides a package for managing application life-cycle for Linux systems.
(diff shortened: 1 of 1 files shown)


packages/katello/katello/katello.spec
```diff
@@ -51,8 +51,7 @@ Requires: pulp-puppet-tools
 Requires: pulp-selinux
 Requires: pulp-server
 Requires: python-pulp-streamer
-Requires: mongodb >= 2.4
-Requires: mongodb-server >= 2.4
+Requires: rh-mongodb34
```
evgeni: Shouldn't this be `rh-mongodb34-mongodb`, as `rh-mongodb34` is the meta package for the SCL itself?

ehelms: I think it can be -- from my testing both worked.

packages/katello/katello/katello.spec
```diff
@@ -71,6 +70,8 @@ Requires: /usr/bin/psql
 Requires: /usr/bin/pg_dump
 Requires: /usr/bin/pg_dumpall
 
+Obsoletes: mongodb
```
ekohl: This feels like a strong statement to make. In the past we haven't obsoleted elasticsearch in this way.

ehelms: True, but we eliminated elasticsearch all together. In our case, we are wanting to force an upgrade. Further, the impression I was getting is that (https://github.com/Katello/katello-selinux/pull/15) will cause a conflict with the mongodb package?

packages/katello/katello/katello.spec
chris1984: looking here its https://www.softwarecollections.org/en/scls/rhscl/rh-mongodb34/

rh-mongodb34

packages/katello/katello/katello.spec
chris1984: The last thing we want to have is both SCL Mongo and Mongo 2.6 on the system at once.

ekohl: If we add a Conflicts: in katello-selinux, would that remove the package as well or do we need to obsolete it somewhere?

ehelms: Conflicts will prevent installation and requires manual user intervention.

evgeni:
[test rpm]

evgeni:
[test rpm]

evgeni:
[test rpm]

evgeni:
[test rpm]

ekohl:
Should this be closed in favor of https://github.com/theforeman/foreman-packaging/pull/2343?

PR #2311, opened 2018-03-27
fixes #23107 - skips dropping database upon incremental restore

cfouant:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/packages/katello/katello/backup.rb
+++ b/packages/katello/katello/backup.rb
@@ -340,7 +340,8 @@ module KatelloUtilities
       os_version = run_cmd("cat /etc/redhat-release").chomp
       plugins = plugin_list
       rpms = run_cmd("rpm -qa").split("\n")
-      system_facts = {os_version: os_version, plugin_list: plugins, rpms: rpms}
+      incremental = @options.fetch(:incremental, false)
+      system_facts = {os_version: os_version, plugin_list: plugins, rpms: rpms, incremental: incremental}
       File.open('metadata.yml', 'w') do |metadata_file|
         metadata_file.puts system_facts.to_yaml
       end
--- a/packages/katello/katello/restore.rb
+++ b/packages/katello/katello/restore.rb
@@ -3,6 +3,7 @@ require 'socket'
 require 'optparse'
 require 'fileutils'
 require 'date'
+require 'yaml'
 require_relative "helper.rb"
 
 module KatelloUtilities
@@ -19,7 +20,7 @@ module KatelloUtilities
       @accepted_scenarios = accepted_scenarios
       @program = program
 
-      @confirmed = false
+      @options = {}
       @is_foreman_proxy_content = !foreman_rpm_installed?
       @skip_register = false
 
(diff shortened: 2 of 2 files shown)


johnpmitsch:
is http://projects.theforeman.org/issues/22971/ the right issue for this?

cfouant:
@johnpmitsch No, it was not. Great catch, I've updated.

packages/katello/katello/restore.rb
```diff
@@ -264,8 +271,7 @@ def backup_valid?
     end
 
     def confirm
-      puts "WARNING: This script will drop and restore your database."
```
johnpmitsch: I definitely think we should still print this message when its not an incremental backup

cfouant:
:+1: good idea @johnpmitsch, thanks! Let me know what you think of the implementation I proposed.

packages/katello/katello/backup.rb
```diff
@@ -343,6 +343,7 @@ def generate_metadata
       File.open('metadata.yml', 'w') do |metadata_file|
         metadata_file.puts system_facts.to_yaml
```
johnpmitsch: @cfouant I was more thinking of doing something like this - here, we could add incremental as a key in the metadata yaml file:

```
system_facts = {os_version: os_version, plugin_list: plugins, rpms: rpms, incremental: <boolean_for_incremental>}
```

Which means now a backup can be determined as incremental based on its `metadata.yaml`

Then in restore, get rid of the `-i` option and parse the `metadata.yaml` file for the `incremental` boolean and use that to determine if we need to run `reset_katello`

packages/katello/katello/restore.rb
```diff
@@ -178,8 +184,11 @@ def restore
       FileUtils.chown(nil, 'postgres', @dir) unless @is_foreman_proxy_content
       Dir.chdir(@dir)
 
-      set_file_security
-      reset_katello
+      metadata = YAML.load_file(File.join @dir, 'metadata.yml')
+      unless @options[:incremental] or metadata[:incremental]
```
johnpmitsch: `or` is fine here, but rubocop will complain about this when its moved to foreman-maintain, if you don't mind changing it to `||`, it will save me the trouble! :)

packages/katello/katello/restore.rb
```diff
@@ -19,7 +20,8 @@ def initialize(foreman_proxy_content, program, accepted_scenarios=nil)
       @accepted_scenarios = accepted_scenarios
       @program = program
 
-      @confirmed = false
+      @options = {}
+      @options[:confirmed] = false
```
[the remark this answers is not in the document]
cfouant: Nice! thanks, will fix now.

packages/katello/katello/restore.rb
cfouant: absolutely - meant to change these back (had a `not` in other PR)

packages/katello/katello/backup.rb
```diff
@@ -339,7 +339,8 @@ def generate_metadata
       os_version = run_cmd("cat /etc/redhat-release").chomp
       plugins = plugin_list
       rpms = run_cmd("rpm -qa").split("\n")
-      system_facts = {os_version: os_version, plugin_list: plugins, rpms: rpms}
+      incremental = @options.has_key?(:incremental) ? @options[:incremental] : false
```
johnpmitsch: you can do this here:

`incremental = @options.fetch(:incremental, false)`

cfouant:
@johnpmitsch changes made, thanks for the hash fetch option!

johnpmitsch:
@mbacovsky for awareness, this was merged

PR #2325, opened 2018-03-29
Fixes #23059 - foreman-journald package now installable

lzap:
This patch solves the following:
  
  1) The package `foreman-journald` was not in comps, so not available in nightly.
  
  2) There are two gems: `logging-journald` (1.0) and `journald-logger`
  (2.0). We want to depend on the former, which requires the latter. I
  screwed up in the requires, thus the dependency issue during
  installation.
  
  3) Loading of the native version does not work, fixed upstream, creating
  patch:
  
  https://github.com/sandfoxme/journald-native/pull/3

* [x] Nightly

### DIFF
--- a/comps/comps-foreman-rhel7.xml
+++ b/comps/comps-foreman-rhel7.xml
@@ -19,6 +19,7 @@
       <packagereq type="default">foreman-debug</packagereq>
       <packagereq type="default">foreman-ec2</packagereq>
       <packagereq type="default">foreman-gce</packagereq>
+      <packagereq type="default">foreman-journald</packagereq>
       <packagereq type="default">foreman-libvirt</packagereq>
       <packagereq type="default">foreman-mysql2</packagereq>
       <packagereq type="default">foreman-openstack</packagereq>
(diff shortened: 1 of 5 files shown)


lzap:
Wait, found this: http://projects.theforeman.org/issues/23059

lzap:
Fixed, pushed.

lzap:
Hold on, another issue. Jeeez.

lzap:
This is my first package with our new tooling, it's erroring out on jenkins, any tips @ehelms ? I haven't tagged or anything with tito.

lzap:
I still do not understand what's wrong with Jenkins.

I want `foreman-journald` to be installed on systems with journald, where is the best place to do it @ekohl I assume the installer?

lzap:
Oh of course and foreman as well I assume.

lzap:
Bumped the logging gem, not foreman as we don't do this I realized.

lzap:
Done, I haven't built locally yet, fighting with the tools. Won't work here.

ekohl:
It looks like we also haven't added it to `package_manifest.yaml` yet so obal doesn't know how to build it. Could you do that too?

lzap:
Hm, just like that?

lzap:
There is a build failure reported :-(

ekohl:
Yes, nightly is broken due to dynflow

PR #2326, opened 2018-03-29
Refs #25697 - Include katello-installer in foreman-installer

ekohl:
Depends on https://github.com/theforeman/foreman-installer/pull/268


ekohl:
Updated, but will probably fail until https://github.com/theforeman/foreman-installer/pull/268 is merged.

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -30,13 +30,29 @@ BuildRequires: puppet-agent-puppet-strings < 2
...
+Group: Applications/System
+Obsoletes: katello-installer-base < 3.11
+Conflicts: katello-installer-base < 3.11
+Provides: katello-installer-base < 3.11
+
+Requires: %{name} = %{version}-%{release}
+Requires: openssl
+Requires: katello-selinux
+Requires: katello-certs-tools
+Requires: katello-service >= 3.0.0
```
ehelms: Something to put on our list, this was "deprecated" in favor of foreman-maintain but I do see it still referenced in a few hooks.

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -30,13 +30,29 @@ BuildRequires: puppet-agent-puppet-strings < 2
...
 Complete installer for The Foreman life-cycle management system based on Puppet.
 
+%package katello
+Summary: Katello installer bits
+Group: Applications/System
+Obsoletes: katello-installer-base < 3.11
+Conflicts: katello-installer-base < 3.11
+Provides: katello-installer-base < 3.11
+
+Requires: %{name} = %{version}-%{release}
```
ehelms: Needs to be `%{epoch}:%{version}-%{release}`

ehelms:
Upgrade breakage:

```
    Transaction check error:
      file /etc/foreman-installer/scenarios.d/foreman-proxy-content.migrations from install of foreman-installer-katello-1:1.21.0-0.3.develop.el7.noarch conflicts with file from package foreman-installer-katello-3.9.1-1.el7.noarch
      file /etc/foreman-installer/scenarios.d/katello.migrations from install of foreman-installer-katello-1:1.21.0-0.3.develop.el7.noarch conflicts with file from package foreman-installer-katello-3.9.1-1.el7.noarch
```

packages/foreman/foreman-installer/foreman-installer.spec
ekohl: I also realized that when I saw this packaging side and agree it'd be good to avoid the deprecated dependency here.

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -30,13 +30,29 @@ BuildRequires: puppet-agent-puppet-strings < 2
 %description
 Complete installer for The Foreman life-cycle management system based on Puppet.
 
+%package katello
+Summary: Katello installer bits
+Group: Applications/System
+Obsoletes: katello-installer-base < 3.11
```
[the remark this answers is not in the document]
ekohl: I looked at that and hoped the epoch would take care of that.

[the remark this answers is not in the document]
ehelms: This may not be needed after some further testing.

ehelms:
I think I figured out the upgrade issue. In the current foreman-installer-katello, `/etc/foreman-installer/scenarios.d/katello.migrations` is a directory. This RPM is showing this as a symlink. So on upgrade, RPM throws a conflict because you are trying to replace a directory with a symlink. This is a known bug that is not handled [1]. There are suggestions in the bug on how to handle this for upgrades.


[1] https://bugzilla.redhat.com/show_bug.cgi?id=447156

ehelms:
Not only that, but `.applied` lives in the migrations directory and is the record for what migrations have been applied. There seems to be some magic happening inside the Rakefile to deal with the migrations and applied file that won't work when dealing with a real system (cause the Rakefile only applies to the build root and does not handle actions taken when the RPM itself is installed). This may require either reverting this behavior or some post or pre to handle moving everything properly.

ekohl:
http://koji.katello.org/koji/taskinfo?taskID=160385 is a scratch build with the latest patch applied

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -30,13 +30,28 @@ BuildRequires: puppet-agent-puppet-strings < 2
 %description
 Complete installer for The Foreman life-cycle management system based on Puppet.
 
+%package katello
+Summary: Katello installer bits
```
ehelms: Maybe we should update this to something like `Katello scenario for Foreman Installer`

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -30,13 +30,28 @@ BuildRequires: puppet-agent-puppet-strings < 2
...
+Group: Applications/System
+Provides: katello-installer-base < 3.11
```
ehelms: Why do we need this?

packages/foreman/foreman-installer/foreman-installer.spec
ekohl: Good point. I'll make it plural since it also contains a proxy scenario. In the future I hope to add more, like pulp and candlepin standalone.

packages/foreman/foreman-installer/foreman-installer.spec
ekohl: https://docs.fedoraproject.org/en-US/packaging-guidelines/#renaming-or-replacing-existing-packages also lists a Provides but with a `=`. Given we did bump it to 3.11 in nightly, I'm unsure which exact version to use.

PR #2327, opened 2018-03-29
Fixes #23065 - kbackup add validator for --split-pulp-tar

johnpmitsch:
This param gets passed to tar --tape-length, we should validate
the user is passing the right param.

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/packages/katello/katello/backup.rb
+++ b/packages/katello/katello/backup.rb
@@ -372,6 +372,10 @@ module KatelloUtilities
       end
     end
 
+    def tape_length_validator(size)
+      !!(size =~ /^\d+[bBcGKkMPTw]?$/)
+    end
+
     def setup_opt_parser
       @optparse = OptionParser.new do |opts|
         opts.banner = "Usage: #{@program}-backup /path/to/dir [options]\n eg: $ #{@program}-backup /tmp/#{@program}-backup"
@@ -391,7 +395,9 @@ module KatelloUtilities
         end
 
         opts.on("--split-pulp-tar SIZE", "Split pulp data into files of a specified size, i.e. (100M, 50G). See '--tape-length' in 'info tar' for all sizes") do |size|
-          opts.abort("Please specify size according to 'tar --tape-length' format.") unless size
+          unless size && tape_length_validator(size)
+            opts.abort("Please specify size according to 'tar --tape-length' format.")
+          end
           @options[:split_tar] = size
         end
 


johnpmitsch:
This is a follow up from https://github.com/theforeman/foreman-packaging/pull/1974

johnpmitsch:
I made  a test case for this, but cannot add because we do not have tests here :disappointed: 

```
require 'minitest/autorun'
  
def tape_length_validator(size)
# insert method here
end

class SizeValidationTest < Minitest::Test
  def test_size_validation
    assert_equal(false, tape_length_validator("blah"))
    assert_equal(false, tape_length_validator(""))
    assert_equal(false, tape_length_validator("  "))
    assert_equal(false, tape_length_validator("20394023x"))
    assert_equal(false, tape_length_validator("20Z"))
    assert_equal(false, tape_length_validator("20GB"))
    assert_equal(false, tape_length_validator("20xG"))
    assert_equal(false, tape_length_validator("20Gx"))
    assert_equal(false, tape_length_validator("B"))
    assert_equal(false, tape_length_validator("W"))
    assert_equal(false, tape_length_validator("-10B"))
    assert_equal(true, tape_length_validator("10B"))
    assert_equal(true, tape_length_validator("1000b"))
    assert_equal(true, tape_length_validator("4000c"))
    assert_equal(true, tape_length_validator("2134G"))
    assert_equal(true, tape_length_validator("2343K"))
    assert_equal(true, tape_length_validator("343k"))
    assert_equal(true, tape_length_validator("34M"))
    assert_equal(true, tape_length_validator("4P"))
    assert_equal(true, tape_length_validator("2T"))
    assert_equal(true, tape_length_validator("1204w"))
    assert_equal(true, tape_length_validator("1204"))
  end
end
```

@mbacovsky once this is merged, maybe this test case can be added to katello-backup in the move to foreman maintain?

mbacovsky:
@johnpmitsch thanks for heads up. Will add this to `foreman_maintain` along with the test.

packages/katello/katello/backup.rb
```diff
@@ -372,6 +372,12 @@ def validate_directory
       end
     end
 
+    def tape_length_validator(size)
+      with_specific_size = !!(size =~ /^\d+[bBcGKkMPTw]{1}$/)
+      in_bytes = !!(size =~ /^\d+$/)
+      with_specific_size || in_bytes
```
akofink: I think this is the same as `size[/^\d+[bBcGKkMPTw]?$/]`

johnpmitsch:
@mbacovsky fyi the method was updated

packages/katello/katello/backup.rb
mbacovsky: @johnpmitsch, @akofink I think there is a difference. `{1}` means exactly one occurrence (same as no quantificator) while `?` means one or no occurrence. 
```
irb(main):003:0> !!('100' =~ /^\d+[bBcGKkMPTw]?$/)
=> true
irb(main):004:0> !!('100' =~ /^\d+[bBcGKkMPTw]{1}$/)
=> false
irb(main):005:0> !!('100' =~ /^\d+[bBcGKkMPTw]$/)
=> false
irb(main):006:0> !!('100M' =~ /^\d+[bBcGKkMPTw]$/)
=> true
```  
The original didn't allowed size without units. The new version does. While tar accepts no units as bytes I think the new version fits better. I'll update the tests in f-m to cover that.

mbacovsky: Pardon me, I obviously completely and embarrassingly missed the point. Forget my comment. ;)

PR #2331, opened 2018-04-02
Update fog-ovirt to 1.0.1

mmoll:
this should be built into:

* [x] Nightly


ekohl:
The reporting on failed jobs is weird - koji task 92707 passed but because the foreman one failed, it's marked as failed. Thanks @mmoll

PR #2332, opened 2018-04-03
Refs #22888 - Reflect dynflow requirements update

mmoll:
this should be built into:

* [x] Nightly


ekohl:
In https://github.com/theforeman/foreman-packaging/pull/2333 I went through all dependencies and found a few more that could use an update.

iNecas:
The koji failed with some strange state, it seems like it actually got stuck at assets compilation or something like that http://koji.katello.org/koji/taskinfo?taskID=92712

dLobatog:
Closing as

PR #2343, opened 2018-04-03
Refs #23030 - Switch to mongodb from rh-mongodb34

chris1984:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/packages/katello/katello/katello-remove
+++ b/packages/katello/katello/katello-remove
@@ -18,6 +18,7 @@ end
 
 CONFIG_FILES = [
   '/etc/pulp',
+  '/etc/opt/rh/rh-mongodb34/mongod.conf',
   '/etc/candlepin',
   '/etc/katello',
   '/etc/httpd',
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -52,8 +52,7 @@ Requires: pulp-puppet-tools
 Requires: pulp-selinux
 Requires: pulp-server
 Requires: python-pulp-streamer
-Requires: mongodb >= 2.4
-Requires: mongodb-server >= 2.4
+Requires: rh-mongodb34
 Requires: cyrus-sasl-plain
 Requires: python-crane
 Requires: python-gofer-qpid
@@ -171,7 +170,7 @@ Requires: findutils
 Requires: coreutils
 Requires: qpid-tools
 Requires: /bin/ps
-Requires: mongodb
+Requires: rh-mongodb34
 Requires: %{?scl_prefix}rubygem-hammer_cli_katello
 Requires: %{name}-service
 
@@ -185,6 +184,7 @@ Useful utilities for debug info collecting
 Summary: Provides a federation of katello services
 BuildArch: noarch
 Requires: findutils
+Requires: rh-mongodb34
 Requires: foreman-installer-%{name}
 Requires: rubygem-highline
 Obsoletes: katello-capsule
--- a/packages/katello/katello/katello-service
+++ b/packages/katello/katello/katello-service
@@ -5,7 +5,7 @@ require 'optparse'
 ACTIONS = ['restart', 'stop', 'start', 'status', 'list', 'enable', 'disable']
 COMMAND = '/usr/sbin/service-wait'
 SERVICES = {
-  'mongod'                   => 5,
+  'rh-mongodb34-mongod'      => 5,
   'postgresql'               => 5,
   'qpidd'                    => 10,
   'qdrouterd'                => 10,
--- a/packages/katello/katello/katello-service-bash_completion.sh
+++ b/packages/katello/katello/katello-service-bash_completion.sh
@@ -1,7 +1,7 @@
 # Opts for every command
 _katelloservice_help_opts="-h --help"
 _katello_action_opts="restart stop start status list enable disable"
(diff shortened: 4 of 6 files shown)


ehelms:
I think service-wait needs an update too.

Also, https://github.com/theforeman/puppet-service_wait/blob/master/plugins/mongo.sh will likely need an update.

chris1984:
@ehelms updated that with this https://github.com/theforeman/puppet-service_wait/pull/34

packages/katello/katello/katello-debug.sh
```diff
@@ -69,12 +69,6 @@ add_files /var/log/httpd/pulp-http{s,}_access_ssl.log*
 add_files /var/log/httpd/pulp-http{s,}_error_ssl.log*
 add_files /etc/default/pulp*
 
-# MongoDB (*)
```
ekohl: Shouldn't we add the mongo log files? If they move, then we need the new location. It might also be good to add both, even if they're old. I can imagine it could help when there are issues during the migration.

packages/katello/katello/katello-remove
```diff
@@ -43,7 +44,6 @@ LOG_FILES=[
   '/var/log/pulp',
   '/var/log/candlepin',
   '/var/log/httpd',
-  '/var/log/mongodb',
```
ekohl: I think these should still be cleaned. We also have `/var/log/tomcat6` in here which only exists on EL6. We should also clean out the logs from the new location.

packages/katello/katello/katello-remove
```diff
@@ -62,7 +62,7 @@ RPMS=[
   '^pulp',
   '^python-pulp',
   '^pulp-',
-  'mongo',
+  'rh-mongodb',
```
ekohl: Since we grep in `rpm -qa` this is not needed. `mongo` will also match `rh-mongodb` and properly clean up both pre-migration and post-migration systems.

packages/katello/katello/katello-remove
```diff
@@ -172,7 +172,7 @@ end
 puts "\nWARNING: This script will erase many packages and config files."
 puts "Important packages such as the following will be removed:\n"
 puts "  * httpd (apache)"
-puts "  * mongodb"
+puts "  * rh-mongodb"
```
ekohl: I'd leave this as mongo. These are generic service names rather than actual package names.

packages/katello/katello/katello-debug.sh
chris1984: switching to syslog, so not needed anymore

chris1984:
@ekohl updated, can you review again please?

packages/katello/katello/katello-remove
```diff
@@ -43,12 +44,12 @@ LOG_FILES=[
   '/var/log/pulp',
   '/var/log/candlepin',
   '/var/log/httpd',
-  '/var/log/mongodb',
   '/var/log/foreman',
   '/var/log/foreman-proxy',
   '/var/log/foreman-installer',
   '/var/log/tomcat',
   '/var/log/squid',
+  '/var/log/mongodb',
```
ekohl: Any reason you move it here instead of keeping it where it was?

packages/katello/katello/katello-remove
```diff
@@ -62,7 +63,7 @@ RPMS=[
   '^pulp',
   '^python-pulp',
   '^pulp-',
-  'mongo',
+  'mongod',
```
ekohl: Is `mongod` better than `mongo`?

packages/katello/katello/katello-debug.sh
ekohl: Would it still make sense to continue sending the the old log files for the migration case? Could be a case of better to have them & not need them than to not have them & need them.

chris1984:
@ekohl updated with your comments.

packages/katello/katello/katello-remove
```diff
@@ -42,8 +43,8 @@ LOG_FILES=[
   '/var/log/tomcat6',
   '/var/log/pulp',
   '/var/log/candlepin',
-  '/var/log/httpd',
   '/var/log/mongodb',
+  '/var/log/httpd',
```
ekohl: It's nice to minimize the amount of lines changed. In this case it doesn't make a difference so I'd change it back to the way it was.

packages/katello/katello/katello.spec
```diff
@@ -6,7 +6,7 @@
 
 Name:       katello
 Version:    3.7.0
-Release:    2.nightly%{?dist}
+Release:    3.nightly%{?dist}
```
ekohl: version 2 hasn't been released yet so a bump is not needed.

chris1984:
@ekohl Thanks for reviewing, updated based on your comments

ehelms:
I see mongod referenced:

 * https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello/katello-service-bash_completion.sh#L4
 * https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello/service-wait#L70

chris1984:
@ehelms updated.

chris1984:
It wont break anything from the talk in internal IRC, updated the commit message and pr

ehelms:
[test rpm]

PR #2348, opened 2018-04-04
Migrate proxy config to /etc and state to /var

ekohl:


ekohl:
Does this have impact on backup tools that until now have ignored `/var/lib/foreman-proxy` because it didn't exist?

packages/plugins/rubygem-smart_proxy_remote_execution_ssh/rubygem-smart_proxy_remote_execution_ssh.spec
```diff
@@ -56,7 +50,16 @@ This package contains documentation for rubygem-%{gem_name}.
 
 %build
 
+%pre
+if [ -d %{foreman_proxy_dir}/.ssh ]; then
```
lpramuk: if this is already symlink then you must not proceed! Otherwise you would be moving also the symlink!!!

The condition shall be then:
```if [ ! -L %{foreman_proxy_dir}/.ssh -a -d %{foreman_proxy_dir}/.ssh ]; then```

ekohl: So we did handle something similar in 796febe4203eb87f03f24a4d6a223093a1f2f786 with .ansible.cfg but clearly forgot this one.

PR #2350, opened 2018-04-04
fixes #23124 - validates services on prior to backup start

cfouant:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---
To test:
- Run `# katello-service stop --exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd`
- Run `# katello-backup --online-backup`

Backup should be created without failure

### DIFF
--- a/packages/katello/katello/backup.rb
+++ b/packages/katello/katello/backup.rb
@@ -20,7 +20,6 @@ module KatelloUtilities
       @databases = @databases.dup
       @accepted_scenarios = accepted_scenarios
       @last_scenario = self.last_scenario
-      @services_online = true
 
       # keep as variables for easy backporting
       @foreman_proxy_content = foreman_proxy_content
@@ -51,11 +50,13 @@ module KatelloUtilities
     end
 
     def start_services
+      puts "Starting services"
       run_cmd("katello-service start #{@excluded}") unless @services_online
       @services_online = true
     end
 
     def stop_services
+      puts "Stopping services"
       run_cmd("katello-service stop #{@excluded}") if @services_online
       @services_online = false
     end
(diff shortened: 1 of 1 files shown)


packages/katello/katello/backup.rb
```diff
@@ -372,6 +371,11 @@ def validate_directory
       end
     end
 
+    def validate_services
+      result = run_cmd("katello-service status #{@excluded}", [0,1])
+      result.include?('dead') ? false : true
```
johnpmitsch: `result.include?('dead')` should return true/false on its own

cfouant: Yes, but this is backwards. So if services are dead, that means the services are down and thus validate_services would be false.

cfouant: I could negate it in one line, but I believe this is the chosen format rather than `not` or `!`

johnpmitsch: I've never seen this pattern used before, do you know of any docs or examples that say/use ternary true/false is used over `!my_condition`?

johnpmitsch:
Functionality works well!

One suggestions for improvement:  katello-service status spits out a lot of info, grepping for "dead" currently won't show anywhere else, but there is chace some other output in the command could show "dead" and give a false postivie. Is there something more specific we can use? i.e. ` inactive \(dead\) `?

packages/katello/katello/backup.rb
cfouant: I'm happy to change it, I prefer it that way. I recall it from the katello dev guide a while back, but no longer see it. Might have been a different style guide altogether.

johnpmitsch:
When services are stopped and they are started by the scripts, there is no feedback to the user that this is happening, which can be confusing.

```
[root@three-six foreman-packaging]# ./packages/katello//katello/katello-backup --online /backup/
Starting backup: 2018-04-11 14:41:15 +0000
Creating backup folder /backup/katello-backup-20180411144115
### there is a long pause here, which makes one think its taking a long time to create a folder and something may have hung  ###
Generating metadata ... 
...
```
Can we print "Starting Services" and "Stopping Services" in their respective methods?

cfouant:
@johnpmitsch - addition made, thanks!

johnpmitsch:
@mbacovsky this was merged FYI

PR #2385, opened 2018-04-10
Update foreman_templates to 6.0.0

ares:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/dependencies/stretch/hammer_cli_foreman_templates/changelog
+++ b/dependencies/stretch/hammer_cli_foreman_templates/changelog
@@ -1,3 +1,9 @@
+ruby-hammer-cli-foreman-templates (6.0.0-1) stable; urgency=low
+
+  * Update foreman_templates to 6.0.0
+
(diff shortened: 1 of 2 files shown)


mmoll:
Could you please reopen this from a branch which name starts with "deb/develop"? :bowtie:

ares:
is that some temporary problem or a permanent change I've missed? Reopened as https://github.com/theforeman/foreman-packaging/pull/2387

ekohl:
It's something we've had for ages in Debian packaging. We should fix that Jenkins job to use the target branch rather than the source branch to determine if it's Debian.

ares:
seems I just forgot my habits then :-)

ekohl:
I think https://github.com/theforeman/foreman-infra/pull/583 should improve the experience. Too often we've seen it fail because someone has to remember some arbitrary rule.

PR #2386, opened 2018-04-10
clean up el6 references

chris1984:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/packages/katello/katello/katello-service
+++ b/packages/katello/katello/katello-service
@@ -11,7 +11,6 @@ SERVICES = {
   'qdrouterd'                => 10,
   'squid'                    => 10,
   'tomcat'                   => 20,
-  'tomcat6'                  => 20,
   'pulp_workers'             => 20,
   'pulp_celerybeat'          => 20,
   'pulp_resource_manager'    => 20,
@@ -78,9 +77,6 @@ def manage_services(action)
       when 'enable', 'disable'
         puts "systemctl #{action} #{service}.service"
         puts `systemctl #{action} #{service}.service`
-      else
-        puts "#{COMMAND} #{service} #{action}"
-        puts `#{COMMAND} #{service} #{action}`
       end
       failures << service unless $?.success?
     end
--- a/packages/katello/katello/katello-service-bash_completion.sh
+++ b/packages/katello/katello/katello-service-bash_completion.sh
@@ -1,7 +1,7 @@
 # Opts for every command
 _katelloservice_help_opts="-h --help"
 _katello_action_opts="restart stop start status list enable disable"
-_katello_services="mongod postgresql qpidd qdrouterd squid tomcat tomcat6 pulp_workers pulp_celerybeat pulp_resource_manager pulp_streamer foreman-proxy smart_proxy_dynflow_core httpd puppetserver foreman-tasks goferd"
+_katello_services="mongod postgresql qpidd qdrouterd squid tomcat pulp_workers pulp_celerybeat pulp_resource_manager pulp_streamer foreman-proxy smart_proxy_dynflow_core httpd puppetserver foreman-tasks goferd"
 
 _katello-service_exclude-only()
 {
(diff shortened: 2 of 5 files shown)


packages/katello/katello/katello-service
```diff
@@ -78,9 +77,6 @@ def manage_services(action)
       when 'enable', 'disable'
         puts "systemctl #{action} #{service}.service"
         puts `systemctl #{action} #{service}.service`
-      else
```
ekohl: This doesn't sound right. I you're breaking all actions besides `enable` and `disable`. The proper way is to map everything to systemctl (so remove the case statement and just include the 2 `puts` lines).

packages/katello/katello/katello-service-bash_completion.sh
```diff
@@ -1,7 +1,7 @@
 # Opts for every command
 _katelloservice_help_opts="-h --help"
 _katello_action_opts="restart stop start status list enable disable"
-_katello_services="mongod postgresql qpidd qdrouterd squid tomcat tomcat6 pulp_workers pulp_celerybeat pulp_resource_manager pulp_streamer foreman-proxy smart_proxy_dynflow_core httpd puppetserver foreman-tasks goferd"
+_katello_services="mongod postgresql qpidd qdrouterd squid tomcat pulp_workers pulp_celerybeat pulp_resource_manager pulp_streamer foreman-proxy smart_proxy_dynflow_core httpd puppetserver foreman-tasks goferd"
```
ekohl: foreman-tasks should be `dynflowd` in current releases.

packages/katello/katello/service-wait
```diff
@@ -57,25 +53,13 @@ before_start() {
 # after start or restart
 after_start() {
   case "$SERVICE" in
-    tomcat|tomcat6)
```
ekohl: We still have a tomcat service. Would it make sense to keep this in?

packages/katello/katello/service-wait
```diff
@@ -57,25 +53,13 @@ before_start() {
...
-      # RHBZ 789288 - wait until data port is available
-      wait_for_url $TOMCAT_TEST_URL
-      ;;
     foreman)
       wait_for_url $FOREMAN_TEST_URL
       ;;
     katello)
```
ekohl: I wouldn't mind getting rid of this. In production we don't have katello as a standalone service. This probably dates back to the time katello was a standalone rails application rather than a foreman plugin.

packages/katello/katello/service-wait
```diff
@@ -57,25 +53,13 @@ before_start() {
 # after start or restart
 after_start() {
   case "$SERVICE" in
-    tomcat|tomcat6)
-      # RHBZ 789288 - wait until data port is available
-      wait_for_url $TOMCAT_TEST_URL
-      ;;
     foreman)
```
ekohl: I wouldn't mind getting rid of this. With the above configured URL (`http://localhost:5500/foreman/api`) this is probably never going to work. We also don't use the foreman as a standalone service.

packages/katello/katello/service-wait
```diff
@@ -88,15 +72,6 @@ after_start() {
   esac
 }
 
-before_stop() {
-  case "$SERVICE" in
-    tomcat6|tomcat)
```
ekohl: I don't know the exact impact of this, but you're probably right that this is no longer an issue with systemd.

packages/katello/katello/service-wait
```diff
@@ -57,25 +53,13 @@ before_start() {
...
-    mongod)
-      # RHBZ 824405 - wait until service is avaiable
-      for i in {1..$WAIT_MAX}; do mongo --eval "printjson(db.getCollectionNames())" 2>/dev/null 1>&2 && break; sleep 1; done
-      sleep $ADDITIONAL_SLEEP
-      ;;
     postgresql)
-      # RHBZ 800534 for RHEL 6.x - pg sysvinit script return non-zero when PID is not created in 2 seconds
-      RETVAL=0
-      # and wait until port is really avaiable
       for i in {1..$WAIT_MAX}; do ss -ln | grep -q ":$POSTGRES_PORT\s" && break; sleep 1; done
```
ekohl: I kind of wonder how much this is still an issue with systemd.

PR #2388, opened 2018-04-10
Fixes #23397 - Clean up el6 references & remove service-wait

chris1984:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---


chris1984:
Following review comments from https://github.com/theforeman/foreman-packaging/pull/2386

packages/katello/katello/katello-service
```diff
@@ -74,17 +73,12 @@ def manage_services(action)
...
     if service_exists?(service)
-      case action
-      when 'enable', 'disable'
-        puts "systemctl #{action} #{service}.service"
-        puts `systemctl #{action} #{service}.service`
-      else
-        puts "#{COMMAND} #{service} #{action}"
         puts `#{COMMAND} #{service} #{action}`
-      end
+        puts `#{COMMAND} #{service} #{action}`
```
ekohl: Actually I think this code was already correct. I didn't notice `COMMAND` was already `service-wait` rather than `/sbin/service`

packages/katello/katello/katello-service
```diff
@@ -74,17 +73,12 @@ def manage_services(action)
...
+    end
       failures << service unless $?.success?
     end
-  end
+end
```
ekohl: This is not needed.

packages/katello/katello/service-wait
```diff
@@ -57,46 +53,21 @@ before_start() {
...
-      RETVAL=0
-      # and wait until port is really avaiable
       for i in {1..$WAIT_MAX}; do ss -ln | grep -q ":$POSTGRES_PORT\s" && break; sleep 1; done
       # and create lock and pid files if they does not exist
       [ -f "/var/lock/subsys/postgresql" ] || touch "/var/lock/subsys/postgresql"
       [ -f "/var/run/postmaster.${POSTGRES_PORT}.pid" ] || head -n 1 "$POSTGRES_DATA/postmaster.pid" > "/var/run/postmaster.${POSTGRES_PORT}.pid"
       sleep $ADDITIONAL_SLEEP
       ;;
+    tomcat)
+      wait_for_url $TOMCAT_TEST_URL
```
ekohl: I wouldn't move this so the diff is minimal. You've also removed `;;``  which changes the behavior.

packages/katello/katello/service-wait
```diff
@@ -127,8 +98,7 @@ after_disable() {
 
 case "$COMMAND" in
   start|stop|restart|status|enable|disable)
-    before_$COMMAND
-    /bin/systemctl $COMMAND "$SERVICE"
+      /bin/systemctl $COMMAND "$SERVICE"
```
ekohl: Any reason you're changing the indenting?

packages/katello/katello/service-wait
```diff
@@ -127,8 +98,7 @@ after_disable() {
 
 case "$COMMAND" in
   start|stop|restart|status|enable|disable)
-    before_$COMMAND
```
ekohl: This breaks the hooks that might be present and doesn't hurt to keep it in.

packages/katello/katello/service-wait
```diff
@@ -57,46 +53,21 @@ before_start() {
...
+    tomcat)
+      wait_for_url $TOMCAT_TEST_URL
     httpd)
       sleep $ADDITIONAL_SLEEP
       ;;
   esac
 }
 
-before_stop() {
-  case "$SERVICE" in
```
ekohl: If you're removing this check, you should make this hook a noop like the others.

packages/katello/katello/service-wait
```diff
@@ -57,46 +53,21 @@ before_start() {
...
-      # RHBZ 824405 - wait until service is avaiable
-      for i in {1..$WAIT_MAX}; do mongo --eval "printjson(db.getCollectionNames())" 2>/dev/null 1>&2 && break; sleep 1; done
-      sleep $ADDITIONAL_SLEEP
-      ;;
     postgresql)
-      # RHBZ 800534 for RHEL 6.x - pg sysvinit script return non-zero when PID is not created in 2 seconds
-      RETVAL=0
-      # and wait until port is really avaiable
       for i in {1..$WAIT_MAX}; do ss -ln | grep -q ":$POSTGRES_PORT\s" && break; sleep 1; done
       # and create lock and pid files if they does not exist
```
ekohl: I think this could be removed. I'd trust the init scripts to do this right and many systemd scripts no longer care about these files so they might be obsolete anyway.

chris1984:
@ekohl updated with the changes

packages/katello/katello/katello-service
```diff
@@ -73,18 +72,18 @@ def manage_services(action)
   `logger -t #{$0} -p daemon.info '*** #{action} initiated ***'`
 
   services_by_priority(action == 'stop').each do |service|
-    if service_exists?(service)
```
evgeni: why this indentation change? if `if` is inside the `each`

packages/katello/katello/katello-service
```diff
@@ -73,18 +72,18 @@ def manage_services(action)
...
+    when 'enable', 'disable'
+      puts "systemctl #{action} #{service}.service"
+      puts `systemctl #{action} #{service}.service`
+    else
+      puts "#{COMMAND} #{service} #{action}"
+      puts `#{COMMAND} #{service} #{action}`
     end
+    failures << service unless $?.success?
   end
+end
```
evgeni: that closes the `each`, but it's indented as its closing `def manage_services`

packages/katello/katello/service-wait
```diff
@@ -57,44 +53,17 @@ before_start() {
...
 }
 
 before_stop() {
-  case "$SERVICE" in
-    tomcat6|tomcat)
-      # RHBZ 789288 - wait until service port is available
-      for i in {1..$WAIT_MAX}; do ss -ln | grep -q ":$TOMCAT_SERV_PORT\s" && break; sleep 1; done
-      ;;
-  esac
+  test 1 -eq 1 # noop
```
evgeni: `true`?

evgeni: Oh I see, that's used in the other `noop` cases too.

evgeni:
also needs rebase

packages/katello/katello/service-wait
```diff
@@ -82,19 +78,16 @@ after_start() {
       [ -f "/var/run/postmaster.${POSTGRES_PORT}.pid" ] || head -n 1 "$POSTGRES_DATA/postmaster.pid" > "/var/run/postmaster.${POSTGRES_PORT}.pid"
       sleep $ADDITIONAL_SLEEP
       ;;
+=======
+>>>>>>> clean up el6 references
```
ehelms: Rebase leftovers

packages/katello/katello/service-wait
```diff
@@ -57,10 +53,10 @@ before_start() {
 # after start or restart
 after_start() {
   case "$SERVICE" in
-    tomcat|tomcat6)
-      # RHBZ 789288 - wait until data port is available
+    tomcat)
       wait_for_url $TOMCAT_TEST_URL
       ;;
+<<<<<<< HEAD
```
ehelms: Rebase leftovers

packages/katello/katello/katello-service
```diff
@@ -74,15 +73,15 @@ def manage_services(action)
...
-      case action
-      when 'enable', 'disable'
-        puts "systemctl #{action} #{service}.service"
-        puts `systemctl #{action} #{service}.service`
-      else
-        puts "#{COMMAND} #{service} #{action}"
-        puts `#{COMMAND} #{service} #{action}`
-      end
-      failures << service unless $?.success?
+    case action
```
ekohl: Could you revert to the old (correct) indenting?

packages/katello/katello/katello-service-bash_completion.sh
```diff
@@ -1,7 +1,7 @@
 # Opts for every command
 _katelloservice_help_opts="-h --help"
 _katello_action_opts="restart stop start status list enable disable"
-_katello_services="rh-mongodb34-mongod postgresql qpidd qdrouterd squid tomcat tomcat6 pulp_workers pulp_celerybeat pulp_resource_manager pulp_streamer foreman-proxy smart_proxy_dynflow_core httpd puppetserver foreman-tasks goferd"
+_katello_services="rh-mongodb34-mongod postgresql qpidd qdrouterd squid tomcat pulp_workers pulp_celerybeat pulp_resource_manager pulp_streamer foreman-proxy smart_proxy_dynflow_core httpd puppetserver foreman-tasks goferd"
```
ekohl: `foreman-tasks` should be `dynflowd`. It's not related to el6 so I'd be fine with doing it later. It probably also works because it's aliased.

packages/katello/katello/service-wait
```diff
@@ -26,13 +26,9 @@ COMMAND=$2
 WAIT_MAX=${WAIT_MAX:-30}
 RETRY_INTERVAL=${RETRY_INTERVAL:-1}
 TOMCAT_PORT=${TOMCAT_PORT:-8443}
-TOMCAT_SERV_PORT=${TOMCAT_SERV_PORT:-8005}
 TOMCAT_TEST_URL=${TOMCAT_TEST_URL:-https://localhost:$TOMCAT_PORT/candlepin/status}
-MONGO_PORT=${MONGO_PORT:-27017}
 POSTGRES_PORT=${POSTGRES_PORT:-5432}
 POSTGRES_DATA=${POSTGRES_DATA:-/var/lib/pgsql/data}
-FOREMAN_TEST_URL=${FOREMAN_TEST_URL:-http://localhost:5500/foreman/api}
```
ekohl: If you remove the var, you should also remove the service. Same for katello.

ehelms:
This needs to coincide with a katello-installer updated:

https://github.com/Katello/katello-installer/search?utf8=%E2%9C%93&q=service-wait&type=

chris1984:
@ehelms updated

packages/katello/katello/katello.spec
```diff
@@ -5,7 +5,7 @@
 %global confdir common
 
 Name:       katello
-Version:    3.7.0
+Version:    3.7.1
```
ehelms: We keep this inline with the actual Katello release. So just bump the release and leave this as 3.7.0

packages/katello/katello/katello.spec
```diff
@@ -18,20 +18,19 @@ Source1:    katello-debug.sh
...
+Source8:    qpid-core-dump
+Source9:   katello-clean-empty-puppet-environments
+Source10:   katello-change-hostname
+Source11:   katello-repository-publish-check
+Source12:   katello-change-hostname.8.asciidoc
+Source13:   restore.rb
+Source14:   backup.rb
+Source15:   hostname-change.rb
+Source16:   helper.rb
+Source17:   katello-backup-rotate-tar.sh
```
ehelms: Note this shift of source numbers means you have to update all references to these in the spec file. For example, checkout line 109 or 108.

packages/katello/katello/katello.spec
```diff
@@ -209,17 +207,10 @@ Provides a federation of katello services
 Summary: Katello Service utilities
 Group: Applications/System
 
-# service-wait dependency
-Requires:       curl
-Requires:       ruby
```
ehelms: The `katello-service` utility is still written in Ruby so I think we need this.

packages/katello/katello/katello.spec
```diff
@@ -209,17 +207,10 @@ Provides a federation of katello services
...
-Requires:       /usr/sbin/ss
-Requires:       /bin/systemctl
```
ehelms: The `katello-service` still calls out to `systemctl` so I think this is appropriate.

packages/katello/katello/katello-service
```diff
@@ -76,13 +75,13 @@ def manage_services(action)
...
       when 'enable', 'disable'
-        puts "systemctl #{action} #{service}.service"
-        puts `systemctl #{action} #{service}.service`
-      else
-        puts "#{COMMAND} #{service} #{action}"
-        puts `#{COMMAND} #{service} #{action}`
+      puts "systemctl #{action} #{service}.service"
+      puts `systemctl #{action} #{service}.service`
+    else
+      puts "#{COMMAND} #{service} #{action}"
```
ekohl: This needs to be the other way around because systemctl has a different argument order. This also means the whole case statement can be simplified because there's no difference between the actions.

packages/katello/katello/katello-service
```diff
@@ -3,15 +3,14 @@
 require 'optparse'
 
 ACTIONS = ['restart', 'stop', 'start', 'status', 'list', 'enable', 'disable']
-COMMAND = '/usr/sbin/service-wait'
+COMMAND = 'systemctl'
```
ekohl: I think this could be removed or the other case of `systemctl` should also use `COMMAND`

packages/katello/katello/katello-service
johnpmitsch: you can also DRY it up:
```
sys_cmd = "#{COMMAND} #{service} #{action}"
`#{sys_cmd}`
```

chris1984:
@ekohl How does this look?

packages/katello/katello/katello-service
```diff
@@ -74,14 +73,7 @@ def manage_services(action)
...
     if service_exists?(service)
-      case action
-      when 'enable', 'disable'
-        puts "systemctl #{action} #{service}.service"
-        puts `systemctl #{action} #{service}.service`
-      else
-        puts "#{COMMAND} #{service} #{action}"
-        puts `#{COMMAND} #{service} #{action}`
-      end
+      puts `#{COMMAND} #{action} #{service}`
```
ekohl: This no longer prints what it does. I think we should stay compatible. It also no longer appends `.service`. Technically not needed but I'd keep it in since it's more explicit.

packages/katello/katello/katello.spec
```diff
@@ -18,20 +18,19 @@ Source1:    katello-debug.sh
...
-Source14:   restore.rb
-Source15:   backup.rb
-Source16:   hostname-change.rb
-Source17:   helper.rb
-Source18:   katello-backup-rotate-tar.sh
+Source5:    katello-restore
+Source6:    katello-backup
+Source7:    katello-service-bash_completion.sh
+Source8:    qpid-core-dump
+Source9:   katello-clean-empty-puppet-environments
```
ekohl: Could you align this to other sources?

packages/katello/katello/katello.spec
ekohl: I'd change `Source17` to `Source5` and keep the diff small. Perhaps you could even leave out `Source5` and not change anything about these files. A smaller change means its easier to review, easier cherry picks and much smaller risk of regressions because of it.

chris1984:
@ekohl updated

packages/katello/katello/katello.spec
```diff
@@ -96,7 +92,7 @@ install -m 755 %{SOURCE12} %{buildroot}%{_sysconfdir}/cron.daily/katello-reposit
 # symlink script libraries
 mkdir -p %{buildroot}%{_datarootdir}/katello
 install -m 644 %{SOURCE14} %{buildroot}%{_datarootdir}/katello/restore.rb
-install -m 644 %{SOURCE15} %{buildroot}%{_datarootdir}/katello/backup.rb
+install -m 644 %{SOURCE16} %{buildroot}%{_datarootdir}/katello/backup.rb
```
ekohl: This change is no longer needed

packages/katello/katello/katello-service
```diff
@@ -74,14 +73,8 @@ def manage_services(action)
...
     if service_exists?(service)
-      case action
-      when 'enable', 'disable'
-        puts "systemctl #{action} #{service}.service"
-        puts `systemctl #{action} #{service}.service`
-      else
-        puts "#{COMMAND} #{service} #{action}"
-        puts `#{COMMAND} #{service} #{action}`
-      end
+      puts "#{COMMAND} #{action} #{service}"
```
ekohl: I'd add `.service` here as well so it's a smaller change.

chris1984:
@ekohl thanks, updated.

packages/katello/katello/katello.spec
ekohl: (done)

chris1984:
@evgeni removed the command

PR #2397, opened 2018-04-12
Use more modern RPM macros in foreman-proxy.spec

ekohl:
* Use systemd macros where possible

This follows https://fedoraproject.org/wiki/Packaging:Scriptlets#Systemd

* Remove the foreman-proxy.init replacement

This as only relevant in the SCL build (which we don't do), but the
files was removed a long time ago.

* Use %{_sharedstatedir} instead of %{_localstatedir}/lib

This is equivalent, but more specific.

### DIFF
--- a/packages/foreman/foreman-proxy/foreman-proxy.spec
+++ b/packages/foreman/foreman-proxy/foreman-proxy.spec
@@ -46,12 +46,8 @@ Requires:       %{?scl_prefix}rubygem(concurrent-ruby) < 2.0
 Requires:       sudo
 Requires:       wget
 Requires(pre):  shadow-utils
-Requires(post): systemd-sysv
-Requires(post): systemd-units
-Requires(preun): systemd-units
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -183,22 +178,13 @@ if [ $1 == 2 ]; then
   popd >/dev/null
 fi
 
-if [ $1 -eq 1 ]; then
-  /bin/systemctl daemon-reload >/dev/null 2>&1 || :
-fi
+%systemd_post %{name}.service
```
ekohl: This generates a slightly different code that might actually be more appropriate:
```
if [ $1 -eq 1 ] ; then 
        # Initial installation 
        systemctl preset foreman-proxy.service >/dev/null 2>&1 || : 
fi
```

PR #2398, opened 2018-04-12
Update rubygem-sequel to 5.7.1

ekohl:


dLobatog:
nighlty - http://koji.katello.org/koji/taskinfo?taskID=97101

PR #2399, opened 2018-04-12
Remove rubygem-newt

ekohl:
This was intended to be used for the TUI of discovery, but there are no packages dependending on it.

@lzap you initially added this. Could you have a look?


lzap:
Please don't drop those, TUI needs them:

https://github.com/theforeman/foreman-discovery-image/blob/master/20-packages.ks#L41-L42

ekohl:
Thanks for confirming. I did look into the history which implied this was needed, but I wasn't 100% sure. I'll submit a new PR that simplifies the spec by dropping EL6 support.

PR #2401, opened 2018-04-12
Fixes #23265 - Don't build tracer plugin on suse

jturel:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

I haven't tested this on suse yet but it seems to still work from the RHEL side

### DIFF
--- a/packages/katello/katello-host-tools/katello-host-tools.spec
+++ b/packages/katello/katello-host-tools/katello-host-tools.spec
@@ -3,7 +3,7 @@
 
 Name: katello-host-tools
 Version: 3.2.0
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: A set of commands and yum plugins that support a Katello host
 Group:   Development/Languages
 License: LGPLv2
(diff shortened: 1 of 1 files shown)


packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -239,6 +247,7 @@ exit 0
 %else
 %attr(750, root, root) %{_sbindir}/katello-tracer-upload
 %endif
+%endif
```
jlsherrill: instead of wrapping them like this, any objections to using global variables for this?    Such as setting a  'build_tracer' variable at the top, and then using that down below.

jturel: Np, but wouldn't it still be wrapped just the same? Want to make sure I'm not missing something

ekohl:
What's the reason to do this? Does the build fail on SUSE? I thought that with https://github.com/theforeman/foreman-packaging/blob/rpm/develop/comps/comps-katello-client-sles12.xml it wouldn't be included in the repo anyway so there'd be no harm if it built but lacked a dependency.

jturel:
I think the intent is to make it clear that we are not currently interested in supporting it. And by not building, well, we don't have to worry about little quirks here and there that would require things to change.

packages/katello/katello-host-tools/katello-host-tools.spec
jlsherrill: it would still be wrapped, yeah just not based on whether its suse or not, but instead based on whether 'build_tracer' is set.  I think this help better define intent and makes things a bit clearer (I'll have to do more similar things when building this on suse)

jturel:
Closing this in favor of the comps approach which is already there

PR #2403, opened 2018-04-13
Remove ruby-foreman-host-rundeck

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #2405, opened 2018-04-13
SUSE support for katello-repos

stbenjam:
This makes some required updates to get the katello-repos package to build in OBS.

Current builds are here:
  https://build.opensuse.org/package/binaries/home:stbenjam/katello-repos/SLE_12_SP3

### DIFF
--- a/packages/katello/katello-repos/katello-repos.spec
+++ b/packages/katello/katello-repos/katello-repos.spec
@@ -1,6 +1,15 @@
 %global pulp_release stable
 %global pulp_version 2.15
 
+%if 0%{?suse_version}
+%define dist suse%{?suse_version}
+%define repo_dir %{_sysconfdir}/zypp/repos.d
+%define repo_dist sles12
+%else
+%define repo_dir %{_sysconfdir}/yum.repos.d
+%define repo_dist %{dist}
+%endif
+
 Name:           katello-repos
 Version:        3.7.0
 Release:        2.nightly%{?dist}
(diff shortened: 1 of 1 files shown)


packages/katello/katello-repos/katello-repos.spec
```diff
@@ -1,6 +1,15 @@
 %global pulp_release stable
 %global pulp_version 2.15
 
+%if 0%{?suse_version}
+%define dist suse%{?suse_version}
+%define repo_dir %{_sysconfdir}/zypp/repos.d
+%define repo_dist sles12
```
stbenjam: The distro for SLES12sp3 is 1315.  The numbers are in a chart, I'm not sure how we should handle this, since we used `sles12` as the directory for the repo URL.   Should we use suse1315? But then you'd need separate directories for each release of SLES.  :man_shrugging:

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -23,17 +32,18 @@ BuildRequires: sed
 Defines yum repositories for Katello and its sub projects, Candlepin and Pulp.
 
 %package -n katello-client-repos
-Summary:  Definition of yum repositories for Katello clients
-Group:    Applications/System
+Summary:   Definition of yum repositories for Katello clients
+Group:     Applications/System
+Conflicts: katello-repos
```
stbenjam: This won't build in OBS without this, as katello-repos and katello-client-repos have files that conflict with each other.

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -80,8 +90,12 @@ rm -rf %{buildroot}
 
 %files
 %defattr(-, root, root)
-%config %{_sysconfdir}/yum.repos.d/*.repo
+%config %{repo_dir}/*.repo
 %{_sysconfdir}/pki/rpm-gpg/RPM-GPG-KEY-katello
+%dir /etc/pki
+%dir /etc/pki/rpm-gpg
+%dir /etc/zypp
+%dir /etc/zypp/repos.d
```
stbenjam: OBS complains without this too :-\

packages/katello/katello-repos/katello-repos.spec
jlsherrill: i'm okay with this, but we need to add support for sles11 as well (waiting on https://github.com/theforeman/foreman-packaging/pull/2406 to publish the repo)

I don't think i want to get in the business for publishing for each sles minor version.

stbenjam: Ok I updated it, so everything < 1300 is SLES11 otherwise SLES12.  We'll have to update this every time we support a new SLES version.

stbenjam:
Ok this one builds in both OBS and Koji now.

PR #2412, opened 2018-04-18
Various RPM packaging improvements

ekohl:
These are commits from https://github.com/theforeman/foreman-packaging/pull/2291 that prepare for an easier workflow, but already make sense by themselves.


scl_prefixes.json
```diff
@@ -0,0 +1,62 @@
+{
```
ehelms: Would you entertain yaml as a format instead? I find it less cumbersome for human edits than json.

ehelms: Should this file live in gem2rpm directory?

ekohl: I typically choose JSON because it's easier to handle in pure python where YAML is an extension. I originally intended to treat this as a cache where I'd fill it from repositories.

ekohl: In a future patch I'm also going to use this in a `get-gemfile-deps` script which converts `Gemfile` + `bundler.d/*.rb` into a usable format to update all the sections in `foreman.spec`. I'm fine with it living in `gem2rpm`, but it's not strictly a `gem2rpm` thing anymore.

ehelms: Does the YAML extension require installing extra packages with pip? Is that the reason?

ekohl: Yes, it's a pip package. That's why often go for JSON. Especially since in this case it's not really a complex structure that often changes.

PR #2418, opened 2018-04-19
Update rubygem-foreman_remote_execution to 1.5.1

dLobatog:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---

This contains webpack compilation stuff :smile: 

cc @iNecas @adamruzicka

### DIFF
--- a/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
+++ b/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
@@ -13,7 +13,7 @@
 
 Summary:    Plugin that brings remote execution capabilities to Foreman
 Name:       %{?scl_prefix}rubygem-%{gem_name}
-Version:    1.5.0
+Version:    1.5.1
 Release:    1%{?foremandist}%{?dist}
 Group:      Applications/System
 License:    GPLv3
@@ -73,6 +73,7 @@ This package contains documentation for rubygem-%{gem_name}.
 %install
 mkdir -p %{buildroot}%{gem_dir}
 cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
+rm %{buildroot}%{gem_instdir}/.babelrc
 
 %{foreman_bundlerd_file}
 %foreman_precompile_plugin -a -s
@@ -89,6 +90,8 @@ exit 0
 %exclude %{gem_instdir}/.*
 %exclude %{gem_instdir}/%{gem_name}.gemspec
 %exclude %{gem_instdir}/Gemfile
+%exclude %{gem_instdir}/package.json
+%exclude %{gem_instdir}/webpack
 %{gem_instdir}/app
 %{gem_instdir}/config
 %{gem_instdir}/db
@@ -100,6 +103,8 @@ exit 0
 %{foreman_apipie_cache_foreman}
 %{foreman_apipie_cache_plugin}
 %{foreman_assets_plugin}
+%{foreman_webpack_plugin}
+%{foreman_webpack_foreman}
 %{gem_instdir}/Rakefile
 %doc %{gem_instdir}/LICENSE
 %exclude %{gem_cache}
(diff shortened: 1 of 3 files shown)


dLobatog:
I think https://github.com/theforeman/foreman-packaging/pull/2419 has to be merged first. Otherwise REX will try to build using the latest foreman.spec (without this PR) and will not be able to use the path defined there.

iNecas:
Thanks, waiting for the prerequisities, before tesitng

dLobatog:
@iNecas Also https://github.com/theforeman/foreman_remote_execution/pull/346 should get into a release, this is just a test to see if REX compiles, but without that fix, `/job_invocations/xxx` won't work

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -73,6 +73,7 @@ This package contains documentation for rubygem-%{gem_name}.
 %install
 mkdir -p %{buildroot}%{gem_dir}
 cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
+rm %{buildroot}%{gem_dir}/gems/%{gem_name}-%{version}/.babelrc
```
dLobatog: Notice I'm removing Babel's `babelrc` because it's only needed to run `npm run test` on the plugin. But if you try to compile the bundle with it in production, it fails because the `foreman_remote_execution` gem folder does not have - it can't have - a `node_modules` folder. In development it's all fine.

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -95,11 +96,15 @@ exit 0
...
 %{gem_instdir}/public
+%{gem_instdir}/webpack
+%{gem_instdir}/package.json
+%{gem_instdir}
 %{gem_spec}
 %{foreman_bundlerd_plugin}
 %{foreman_apipie_cache_foreman}
 %{foreman_apipie_cache_plugin}
 %{foreman_assets_plugin}
+%{foreman_webpack_plugin}
```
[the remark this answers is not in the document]
dLobatog: Fixd

ekohl:
@dLobatog there was an issue with http://koji.katello.org/koji/buildinfo?buildID=15214 because it was tagged in all tags. This hides the actual package due to priorities. I've now removed it from foreman-nightly-nonscl-rhel7, foreman-plugins-nightly-rhel7 and foreman-plugins-nightly-nonscl-rhel7
[test rpm]

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -95,11 +96,16 @@ exit 0
 %{gem_instdir}/locale
 %{gem_instdir}/lib
 %{gem_instdir}/public
+%{gem_instdir}/webpack
```
ekohl: This and `package.json` can be excluded because they're not needed in production. I do wonder if we should create a -assets package with these files.

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
ekohl: I think we should exclude this from the actual gem.

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -95,11 +96,16 @@ exit 0
...
+%{gem_instdir}/package.json
+%{gem_instdir}
```
ekohl: Is this actually needed?

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
ekohl: This is equal to `rm %{buildroot}%{gem_instdir}/.babelrc` and we already have `%exclude %{gem_instdir}/.*` in `%files`. Does this line actually make a difference? If it does, I'd prefer to use the `%{gem_instdir}` macro instead of writing it out but ideally we'd remove this line.

ekohl:
Thanks! http://koji.katello.org/koji/taskinfo?taskID=97658

ehelms:
Note, this may be affected by https://github.com/theforeman/foreman-packaging/pull/2256#issuecomment-386641581 and need a re-build

PR #2442, opened 2018-04-30
include JS files for webpack assets in foreman-assets package

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.18

I'm unsure what elso needs to get packaged now for the plugin asset builds with webpack to succeed.

### DIFF
--- a/debian/stretch/foreman/foreman-assets.postinst
+++ b/debian/stretch/foreman/foreman-assets.postinst
@@ -29,6 +29,9 @@ fi
 # Own all the core files
 chown -Rf foreman:foreman '/usr/share/foreman'
 
+# make webpack file executable
+chmod 755 '/usr/share/foreman/node_modules/webpack/bin/webpack.js'
+
 #DEBHELPER#
 
 exit 0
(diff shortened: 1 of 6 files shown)


ekohl:
We modified the rake task that precompiles assets so as long as that's called, it should work. It should be noted that some plugins have their own nodejs deps and I don't know how you'd handle that.

mmoll:
this allows now to build _rex 1.5.3 https://ci.theforeman.org/job/packaging_build_deb_plugin/1832/

mmoll:
@ekohl @ehelms @dLobatog @sharvit any more comments? :)

debian/stretch/foreman/foreman-assets.postinst
```diff
@@ -29,6 +29,9 @@ fi
 # Own all the core files
 chown -Rf foreman:foreman '/usr/share/foreman'
 
+# make webpack file executable
+chmod 755 '/usr/share/foreman/node_modules/webpack/bin/webpack.js'
```
sharvit: Shouldn't it live under `./node_modules/.bin/webpack`?

mmoll: That's a symlink to the actiual file - at least in the environment, this got built on.

mmoll:
> It should be noted that some plugins have their own nodejs deps and I don't know how you'd handle that.

I don't know

PR #2448, opened 2018-05-02
[WIP] Introduce a unified add_package script

ekohl:
The goal of this is to handle package addition for all package types and all repos. It's still very much WIP. Most of the gem aspect should work, but npm is missing. It also started work to the comps script which has assumptions on the XML structure with comments that aren't true for katello comps.

### DIFF
--- a/add_package
+++ b/add_package
@@ -1,0 +1,205 @@
+#!/usr/bin/env python3
+
+import argparse
+import configparser
+import logging
+import subprocess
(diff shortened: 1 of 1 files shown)


ehelms:
@ekohl What route would you like to take with this?

ehelms:
Moved to https://github.com/theforeman/obal/issues/236

PR #2449, opened 2018-05-02
[WIP] Handle asset compilation for plugins without webpack

ekohl:


ekohl:
Confirmed to work on tfm-rubygem-foreman_expire_hosts-5.1.0-1.fm1_18.el7.src.rpm (no webpack) but tfm-rubygem-foreman_remote_execution-1.5.1-1.fm1_18.el7.src.rpm (with webpack) breaks.

ehelms:
Is that the updated remote_execution RPM rebuilt against webpack?

ekohl:
Yes, locally rebuilt. I think I used the wrong path to verify - the foreman one which is supposed to the symlink location. Checking that now.

ekohl:
Updated. It worked on webpack, but my previous iteration had a mistake in the apipie macro. Still building locally but could use a new review because I think it's ready.

ekohl:
I could locally build REX as well now. Should be all good to go.

PR #2454, opened 2018-05-04
Update foreman nodejs dependencies

ekohl:


ekohl:
Includes https://github.com/theforeman/foreman/pull/5529 and would unbreak our nightlies.

PR #2455, opened 2018-05-04
update foreman_xen to 0.6.1

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.17


packages/plugins/rubygem-foreman_xen/rubygem-foreman_xen.spec
```diff
@@ -95,6 +95,9 @@ cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
 %doc %{gem_instdir}/README.md
 
 %changelog
+* Sun May 05 2018 Michael Moll <kvedulv@kvedulv.de> 0.6.1-1
```
ekohl: There is no Sunday May 5th

ekohl:
Ah, I now see it:

https://github.com/mmoll/foreman-packaging/blob/d82aae327a655dd05b907e594e05e1281db25a5a/packages/plugins/rubygem-foreman_xen/rubygem-foreman_xen.spec#L86

This should now be `foreman_xen` instead of `compute_resources`

mmoll:
should be all good now

PR #2460, opened 2018-05-07
update foreman_salt to 10.0.0

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.17


tbrisker:
Looks like this didn't get built into the 1.17 repo which still has the incompatible 9.x version of the plugin

PR #2464, opened 2018-05-09
Add rubygem-dalli for foreman_memcache

ekohl:


ekohl:
I noticed the dalli package might be in the ror42 SCL so this might not be needed. Just a change in the dependency for `foreman_memcache`.

PR #2477, opened 2018-05-13
set bundle vars

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.17
* [x] 1.16
* [x] 1.15


ekohl:
Any reason to do this with vars rather than `--jobs=4 --retry=5`?

mmoll:
I don't remeber the exact bundler version, but somewhen in the past "package: didn't take the arguments like "install"

PR #2484, opened 2018-05-15
Fixes #23595 - Set ruby-sequel-pg to conflicting

FrediWeber:
Don't allow ruby-sequel-pg to be installed that the
db:migrate task can run properly.

### DIFF
--- a/debian/stretch/foreman/control
+++ b/debian/stretch/foreman/control
@@ -17,7 +17,7 @@ Depends: ruby2.3, bundler (>= 1.3), zlib1g-dev,
          ruby2.3-dev, build-essential,
          rake (>=0.8.4), ${misc:Depends}
 Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal
-Conflicts: ruby-activerecord-deprecated-finders, ruby-foreman-tasks (<< 0.10.1)
+Conflicts: ruby-activerecord-deprecated-finders, ruby-foreman-tasks (<< 0.10.1), ruby-sequel-pg
 Description: Systems management web interface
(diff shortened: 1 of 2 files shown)


ekohl:
I think this makes sense, but shouldn't it be applied to Ubuntu as well? IIRC it suffers from the same bug.

FrediWeber:
@ekohl Absolutely, changed it also for Ubuntu.

iNecas:
I don't know enough about deb packaging, but if 'conflicts' causes the recommended package (ruby-sequel-pg) not to be installed or to be uninstalled when present, then +1 from me. The ultimate solution although would be preventing the rails stack to use the non-vendored gems (maybe adding `bundle exec` somewhere would help?)

Because the problem is not with the sequel-pg itself, but with the fact, that there is a version of sequel-pg in system ruby that is not compatible with sequel that we ship with Foreman.

FrediWeber:
Yeah but if it's present the required task (e.g. db:migrate) doesn't work so it's definitely conflicting in the current implementation.
Conflicting basically says if ruby-sequel-pg is installed you cannot install foreman or should remove ruby-sequel-pg and you should not install ruby-sequel-pg if it is marked somewhere as recommended which is actually the case and the reason why it gets installed.

mmoll:
@FrediWeber This is clear to me, I just wanted to make sure that this doesn't introduce an error when e.g. using the remote_execution proxy part on the same machine as the Foreman core component.

FrediWeber:
As far as I can judge it works fine on our systems but I haven‘t tested
every possible constellation.

Michael Moll <notifications@github.com> schrieb am Mi. 16. Mai 2018 um
17:09:

PR #2496, opened 2018-05-16
Fixes #23614 - Remove katello-backup and katello-restore.

johnpmitsch:
Katello-backup and katello-restore have moved to foreman-maintain. The commands
in foreman-maintain support remote databases and mongo 3.4. Since katello-backup
and katello-restore do not support these features, we have to remove them in
Katello 3.7 (unfortunately without deprecation)

* [x] Nightly
* [x] 1.18
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---


johnpmitsch:
Depends on https://github.com/theforeman/foreman_maintain/pull/169 to be merged and a fmaintain release containing this fix

packages/katello/katello/katello-backup
```diff
@@ -1,18 +1,5 @@
 #!/usr/bin/env ruby
```
ekohl: How about a simple shell script for this? I'd also `exit 1` so cronjobs start to fail.

packages/katello/katello/katello-restore
```diff
@@ -1,18 +1,5 @@
 #!/usr/bin/env ruby
```
ekohl: How about a simple shell script for this? I'd also `exit 1` so cronjobs start to fail.

johnpmitsch:
@ekohl  thanks, updated to use a bash script and exit 1.

I'm not sure if I am requiring foreman-maintain properly, the latest version is [here](https://yum.theforeman.org//nightly/el7/x86_64/)  as `rubygem-foreman_maintain-0.2.1-1.el7.noarch.rpm`. It is needed on both katello and a foreman-proxy

johnpmitsch:
Update 0.2.2 was [just released](https://github.com/theforeman/foreman-packaging/pull/2504)

@ekohl we are working to hide any functionality that won't work upstream (like the upgrade command) so `foreman-maintain` can be more relied on upstream :)

Should I require `rubygem-foreman_maintain` rather than `foreman-maintain`? I'm not as familiar with our syntax for requiring dependencies.

ekohl:
> @ekohl we are working to hide any functionality that won't work upstream (like the upgrade command) so foreman-maintain can be more relied on upstream :)

I'm also talking about not assuming Katello SSL structures and Proxy port 9090 vs 8443. There are issues open for this, but I'm wondering how much more there is.

> Should I require rubygem-foreman_maintain rather than foreman-maintain? I'm not as familiar with our syntax for requiring dependencies.

I think the original package name (rubygem-foreman_maintain) is cleaner than something it provides but I'm not too familiar with best practices.

packages/katello/katello/katello-backup
```diff
@@ -1,18 +1,10 @@
...
-# check if production install
-if __FILE__.include?('sbin')
-  UTILS_PATH='/usr/share/katello'
-else
-  UTILS_PATH=File.expand_path('..', __FILE__)
-end
+MESSAGE="
+IMPORTANT: katello-backup has been removed in favor of 'foreman-maintain backup'.
+This change happened in order to support remote databases and a newer mongo version.
+Please update your scripts and/or cron jobs.
```
ehelms: How different is the syntax for foreman-maintain backup? I am wondering if for a release we can change this script to essentially do:

```
foreman-maintain backup $@
```

To pass through the options and keep the script "intact" for a single release.

johnpmitsch: we discussed that, it would require a "translation" script since the syntax is different for foreman-maintain backup. It uses subcommands for each backup. i.e.

`foreman-maintain backup online`
is 
`katello-backup --online-backup`

I'm not sure if its worth the added time and complexity to do something like that. I plan on adding a deprecation warning in the next 3.6.z if that helps at all.

johnpmitsch:
Rebased and added a changelog

packages/katello/katello/katello.spec
```diff
@@ -189,15 +181,10 @@ Provides a federation of katello services
 
 %files -n foreman-proxy-content
 %config(missingok) %{_sysconfdir}/cron.weekly/katello-clean-empty-puppet-environments
-%{_sbindir}/katello-backup
```
ekohl: Isn't this still needed since it's the script that mentions the deprecation?

packages/katello/katello/katello.spec
```diff
@@ -215,6 +205,9 @@ Useful utilities for managing Katello services
 %{_sysconfdir}/bash_completion.d/katello-service
 
 %changelog
+* Fri May 23 2018 John Mitsch <jomitsch@redhat.com> - 3.7.0-5.nightly
```
jturel:
```
packages/katello/katello/katello.spec: E: specfile-error warning: bogus date in %changelog: Fri May 23 2018 John Mitsch <jomitsch@redhat.com> - 3.7.0-5.nightly
```

I like where your mind is at but alas it is not yet Friday :-1:

jturel:
Dead code - not sure if you intentionally left it or not

https://github.com/theforeman/foreman-packaging/blob/8b588b471bd4e7825d8e6ae6e944d07191bd5c25/packages/katello/katello/helper.rb#L27

Maybe other methods there which are no longer referenced due to removal of the other scripts...

johnpmitsch:
@jturel nice catch,

I removed `default_split_tar_script` and `foreman_rpm_installed?`, those seem to be the only ones that aren't used elsewhere. I also updated to use the correct date

packages/katello/katello/katello.spec
```diff
@@ -215,6 +205,9 @@ Useful utilities for managing Katello services
 %{_sysconfdir}/bash_completion.d/katello-service
 
 %changelog
+* Wed May 23 2018 John Mitsch <jomitsch@redhat.com> - 3.7.0-5.nightly
+- Remove katello-backup and katello-restore and require foreman-maintian
```
ekohl: s/maintian/maintain/

PR #2505, opened 2018-05-18
fixes #23647 - moves cron jobs to cron.d for greater flexibility

cfouant:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

---
`katello-remove-orphans` can interrupt and interfere with creating a good backup. Moving this process into cron.d, along with `katello-clean-empty-puppet-environments` will allow for greater control and flexibility.

### DIFF
--- a/packages/katello/katello/katello.cron
+++ b/packages/katello/katello/katello.cron
@@ -1,0 +1,14 @@
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+
+# Script for initiating removal of orphaned content
+00 22 * * 0	root	foreman-rake katello:delete_orphaned_content RAILS_ENV=production >/dev/null 2>&1
+
+# cronjob to clean puppet envs for puppet 4
+15 22 * * 0	root	[ -d /etc/puppetlabs/code/environments ] &&  find /etc/puppetlabs/code/environments/KT* -maxdepth 0 -type d -empty -delete > /dev/null 2>&1
+
+# cronjob to clean puppet envs for puppet 3
+20 22 * * 0	root	[ -d /etc/puppet/environments ] && find /etc/puppet/environments/KT* -maxdepth 0 -type d -empty -delete > /dev/null 2>&1
+
+# Script for publishing unpublished repositories
+* 22 * * *     root    foreman-rake katello:publish_unpublished_repositories RAILS_ENV=production >/dev/null 2>&1
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -16,18 +16,16 @@ URL:        https://theforeman.org/plugins/katello
 Source0:    katello-service.8.asciidoc
 Source1:    katello-debug.sh
 Source2:    katello-remove
-Source3:    katello-remove-orphans
 Source4:    katello-service
 Source6:    katello-restore
 Source7:    katello-backup
 Source8:    katello-service-bash_completion.sh
 Source9:    qpid-core-dump
-Source10:   katello-clean-empty-puppet-environments
 Source11:   katello-change-hostname
-Source12:   katello-repository-publish-check
 Source13:   katello-change-hostname.8.asciidoc
 Source16:   hostname-change.rb
 Source17:   helper.rb
+Source18:   katello.cron
 
 BuildRequires: asciidoc
 BuildRequires: util-linux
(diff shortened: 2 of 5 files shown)


packages/katello/katello/katello.spec
```diff
@@ -87,10 +87,10 @@ gzip -f9 %{_sourcedir}/katello-change-hostname.8
...
 
 #copy cron scripts to be scheduled
-install -d -m0755 %{buildroot}%{_sysconfdir}/cron.weekly
+install -d -m0755 %{buildroot}%{_sysconfdir}/cron.d
 install -d -m0755 %{buildroot}%{_sysconfdir}/cron.daily
-install -m 755 %{SOURCE3} %{buildroot}%{_sysconfdir}/cron.weekly/katello-remove-orphans
-install -m 755 %{SOURCE10} %{buildroot}%{_sysconfdir}/cron.weekly/katello-clean-empty-puppet-environments
+install -m 755 %{SOURCE3} %{buildroot}%{_sysconfdir}/cron.d/katello-remove-orphans
+install -m 755 %{SOURCE10} %{buildroot}%{_sysconfdir}/cron.d/katello-clean-empty-puppet-environments
 install -m 755 %{SOURCE12} %{buildroot}%{_sysconfdir}/cron.daily/katello-repository-publish-check
```
ekohl: Should this also be moved into the .d file?

komidore64:
> I installed the built RPM, and /etc/cron.weekly/katello-remove-orphans et al still exist. Is there something we can do in the spec file to remove the old cron jobs?

It looks like they were previously flagged with the `%config` directive, which I believe tells RPM to not delete or overwrite them if they are removed or changed in future `%files` lists.

We could try deleting them in a `%post` section.

akofink:
We must delete them, right? Otherwise we'd have duplicate cron jobs?

akofink:
According to [this](http://ftp.rpm.org/max-rpm/s1-rpm-erase-and-config-files.html) (where is the official spec file docs?), config files *should* be removed or appended with `.rpmsave` during `rpm -e`, but it doesn't mention what happens on upgrade.

cfouant:
@ekohl - i have tested the new cron jobs and they are working, and the changes made here do introduce the file /etc/cron.d/katello with all of our cron tasks. Do you have any other changes requested or could you clarify what changes you're requesting?

@akofink - could you retest? When I upgraded my katello rpm, I was required to also update katello-common rpm, and upon finishing, the tasks were removed from /etc/cron.weekly and a new file was added to /etc/cron.d

packages/katello/katello/katello.cron
```diff
@@ -0,0 +1,12 @@
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+
+# Script for intiating removal of orphaned content
+00 22 * * 0	root	foreman-rake katello:delete_orphaned_content RAILS_ENV=production >/dev/null 2>&1
+
+# cronjob to clean puppet envs for puppet 4
+15 22 * * 0	root	[ -d /etc/puppetlabs/code/environments ] &&  find /etc/puppetlabs/code/environments/KT* -maxdepth 0 -type d -empty -delete > /dev/null 2>&1
```
ekohl: Isn't the find equivalent to `rmdir --ignore-fail-on-non-empty /etc/puppetlabs/code/environments/KT*`? I'm fine with keeping it the current way to keep the change small though.

packages/katello/katello/katello.spec
```diff
@@ -81,10 +80,9 @@ gzip -f9 %{_sourcedir}/katello-change-hostname.8
 mkdir -p %{buildroot}/%{_mandir}/man8
 
 #copy cron scripts to be scheduled
-install -d -m0755 %{buildroot}%{_sysconfdir}/cron.weekly
+install -d -m0755 %{buildroot}%{_sysconfdir}/cron.d
 install -d -m0755 %{buildroot}%{_sysconfdir}/cron.daily
-install -m 755 %{SOURCE3} %{buildroot}%{_sysconfdir}/cron.weekly/katello-remove-orphans
-install -m 755 %{SOURCE10} %{buildroot}%{_sysconfdir}/cron.weekly/katello-clean-empty-puppet-environments
+install -m 644 %{SOURCE18} %{buildroot}%{_sysconfdir}/cron.d/katello
```
ekohl: You can remove the `mkdir` by using `install -D`

cfouant: Not sure what `mkdir` you're referring to? the `install -m` here sets the permissions.

packages/katello/katello/katello.cron
cfouant: Very possible, but would prefer not to change this to keep the change small and limit the scope of the BZ as well.

packages/katello/katello/katello.spec
ekohl: Above you have `install -d %{_sysconfdir}/cron.d` but if you use `install -D` instead of `install` here that should be redundant I think. But feel free to keep it the way it is.

cfouant:
@ekohl - I moved our cron.daily tasks into cron.d, good suggestion :+1: 

@akofink could you do a final code review with this remaining change please and thank you :1st_place_medal:

packages/katello/katello/katello.cron
```diff
@@ -0,0 +1,14 @@
...
+
+# cronjob to clean puppet envs for puppet 3
+20 22 * * 0	root	[ -d /etc/puppet/environments ] && find /etc/puppet/environments/KT* -maxdepth 0 -type d -empty -delete > /dev/null 2>&1
+
+# Script for publishing unpublished repositories
+* 22 * * *     root    foreman-rake katello:publish_unpublished_repositories RAILS_ENV=production >/dev/null 2>&1
```
ekohl: Doesn't this mean every minute of the 22nd hour?

johnpmitsch: yes, it does [see here](http://corntab.com/?c=*_22_*_*_*_)

looks like we need to update to `0 22 * * * ` [see here](http://corntab.com/?c=0_22_*_*_*_)

PR #2514, opened 2018-05-23
Update smart_proxy_ansible to 2.0.3

dLobatog:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [x] 1.17
* [x] 1.16
* [ ] 1.15

See Foreman's [plugin maintainer documentation](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

This PR ought to replace https://github.com/theforeman/foreman-packaging/pull/2506 - the dependency has been removed.

PR #2517, opened 2018-05-24
Release 1.17.1

xprazak2:

* [ ] Nightly
* [x] 1.17
* [ ] 1.16
* [ ] 1.15

---

### DIFF
--- a/debian/stretch/foreman-installer/changelog
+++ b/debian/stretch/foreman-installer/changelog
@@ -1,3 +1,9 @@
+foreman-installer (1.17.1-1) stable; urgency=low
+
+  * Release 1.17.1
+
+ -- Ondrej Prazak <oprazak@redhat.com>  Thu, 24 May 2018 17:45:59 +0200
+
 foreman-installer (1.17.0-1) stable; urgency=low
 
   * 1.17.0 released
--- a/debian/stretch/foreman-proxy/changelog
+++ b/debian/stretch/foreman-proxy/changelog
@@ -1,3 +1,9 @@
+foreman-proxy (1.17.1-1) stable; urgency=low
+
+  * Release 1.17.1
+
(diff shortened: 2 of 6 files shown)


ekohl:
I think this failed because it's comparing the branch to *your* `deb/1.17` which doesn't exist. Not sure that's what we should do, but I think the quickest fix is if you push `deb/1.17` and rerun the tests.

mmoll:
the job is indeed broken, or the parameter passing to it

dLobatog:
Next step of the pipeline kicked off at http://ci.theforeman.org/job/release_packages/125, @xprazak2 please sign the RPMs once the job is done as I think you're the only one who has the keys.

@komidore64 Was this job one of the jobs recently rewritten to JJB? I also see `deb/deb/develop` for example in the tests in https://github.com/theforeman/foreman-packaging/pull/2514

komidore64:
@dLobatog looks like was was [added](https://github.com/theforeman/foreman-infra/commit/403fd0ab) in February.

mmoll:
iirc this stopped to work when jjb on ci.theforeman.org had some problems and after it got moved to slave02(because newer CentOS with newer python) somehow these problems arose...

ekohl:
I wonder if I introduced this in https://github.com/theforeman/foreman-infra/commit/adb5642530827412a5a4cdfefe4b02ae4f905c2c

mmoll:
unsure, but I think that worked...

ekohl:
I think my commit in combination with https://github.com/theforeman/foreman-infra/blob/4815364015b0ae3de9b9376de8f01969fd86a05e/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging_build_deb_coreproject.yaml#L12 could be an issue. In https://github.com/theforeman/foreman-infra/commit/f87a62691b2d296048dab6f1147272d03de42258 the `deb/` was added to fix something, probably to reflect a change from my change.

PR #2521, opened 2018-05-27
[WIP] Regenerate foreman plugin specs

ekohl:
This regenerates the foreman plugin specs to the current template with some minor modification so the diff is smaller. The commit `Temporary ...` should NOT be merged but is useful because it automates most of the regeneration. It's also marked as WIP because this is not every plugin.


ekohl:
Submitted as separate PRs so I can easily see the build status per plugin rather than rebasing every time one fails.

PR #2526, opened 2018-05-28
Regenerate rubygem-foreman_snapshot_management based on the current template

ekohl:


ekohl:
@mdellweg you might want to take a look. Looks like previously assets and apipie caches weren't generated.

mdellweg:
@ekohl LGTM

PR #2527, opened 2018-05-28
Regenerate rubygem-foreman_snapshot based on the current template

ekohl:


ekohl:
@GregSutcliffe we discussed of retiring this plugin. Shall we do that instead?

ekohl:
Replaced by https://github.com/theforeman/foreman-packaging/pull/2557

PR #2540, opened 2018-05-28
Regenerate rubygem-foreman_graphite based on the current template

ekohl:


packages/plugins/rubygem-foreman_graphite/rubygem-foreman_graphite.spec
```diff
@@ -1,47 +1,38 @@
...
-%global foreman_pluginconf_dir /etc/foreman/plugins
+%global plugin_name graphite
+%global foreman_min_version 1.7.0
 
 Summary:    Adds graphite integration to Foreman
 Name:       %{?scl_prefix}rubygem-%{gem_name}
 Version:    0.0.3
-Release:    4%{?foremandist}%{?dist}
-Group:      Applications/System
+Release:    2%{?foremandist}%{?dist}
```
ekohl: This needs to be 5

PR #2549, opened 2018-05-28
Regenerate rubygem-foreman_chef based on the current template

ekohl:


ekohl:
Adding core merge because of https://github.com/theforeman/foreman_chef/pull/86.

PR #2531, opened 2018-05-28
Regenerate rubygem-foreman_remote_execution based on the current template

ekohl:

PR #2553, opened 2018-05-28
Handle webpack for foreman plugins

ekohl:


ekohl:
It might be needed to include package.json dependencies: https://github.com/theforeman/foreman-packaging/pull/2531

PR #2558, opened 2018-05-29
Handle .ssh symlinks in REX SSH

ekohl:
Similar to 796febe4203eb87f03f24a4d6a223093a1f2f786. test -d deferences symlinks so we explicitly need to check if it's not a symlink.


ekohl:
@lpramuk did you create a BZ for this?

PR #2573, opened 2018-05-31
Add 1.18 to the PR template

ekohl:
Also changes the link to Redmine to use HTTPS

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,10 +1,10 @@
 For plugin updates, please indicate which repos this should be built into:
 
 * [x] Nightly
+* [ ] 1.18
 * [ ] 1.17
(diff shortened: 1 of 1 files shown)


mmoll:
looking at the history of this file, do we want do also remove 1.15 with this change?

ekohl:
I thought about that. Have we formally ended support for 1.15? It'd be useful if we said that with the release of 1.17 "now 1.15 is unsupported" and when 1.18.0 is out we stop support with 1.16. Now it's not always that clear.

mmoll:
@dLobatog :point_up: :question:

ehelms:
Our support policy says N and N-1 so even though we haven't been explicitly stating it in release notes this would mean 1.15 was dead.

ekohl:
Updated to remove 1.15

mmoll:
merged, bedankt @ekohl!

PR #2581, opened 2018-06-01
Update url and gpg for 1.18

xprazak2:

* [ ] Nightly
* [x] 1.18
* [ ] 1.17
* [ ] 1.16

---


packages/foreman/foreman/foreman-plugins.repo
```diff
@@ -1,13 +1,13 @@
 [foreman-plugins]
-name=Foreman plugins nightly
```
ekohl: This should be done by a sed command in the .spec already.

packages/foreman/foreman/foreman.repo
```diff
@@ -1,13 +1,13 @@
 [foreman]
-name=Foreman nightly
-baseurl=https://yum.theforeman.org/releases/nightly/$DIST/$basearch
+name=Foreman 1.18
+baseurl=https://yum.theforeman.org/releases/1.18/$DIST/$basearch
 enabled=1
-gpgcheck=0
+gpgcheck=1
```
ekohl: We should automate this in the .spec file so we don't have to change this after branching.

PR #2582, opened 2018-06-01
Fixes #23767 - Enable SCL for Mongo 3.x

chris1984:

* [x] Nightly
* [x] 1.18
* [ ] 1.17
* [ ] 1.16

---

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -6,7 +6,7 @@
 
 Name:       katello
 Version:    3.7.0
-Release:    5.nightly%{?dist}
+Release:    6.nightly%{?dist}
 Summary:    A package for managing application life-cycle for Linux systems
 BuildArch:  noarch
 
@@ -110,6 +110,9 @@ install -m 644 %{SOURCE8} %{buildroot}/etc/bash_completion.d/katello-service
 # install man page
 install -m 644 %{_sourcedir}/katello-service.8.gz %{buildroot}/%{_mandir}/man8
 install -m 644 %{_sourcedir}/katello-change-hostname.8.gz %{buildroot}/%{_mandir}/man8
+
+# Enable SCL for Mongo 3.x
+scl enable rh-mongodb34 bash
 
 %clean
 %{__rm} -rf %{buildroot}


packages/katello/katello/katello.spec
```diff
@@ -111,6 +111,9 @@ install -m 644 %{SOURCE8} %{buildroot}/etc/bash_completion.d/katello-service
 install -m 644 %{_sourcedir}/katello-service.8.gz %{buildroot}/%{_mandir}/man8
 install -m 644 %{_sourcedir}/katello-change-hostname.8.gz %{buildroot}/%{_mandir}/man8
 
+# Enable SCL for Mongo 3.x
+scl enable rh-mongodb34 bash
```
ehelms: This is not persistent for a system, and only enables the SCL for the current session. We either have to consider something similar to this for puppet:

```
# Add /opt/puppetlabs/bin to the path for sh compatible users

if ! echo $PATH | grep -q /opt/puppetlabs/bin ; then
  export PATH=$PATH:/opt/puppetlabs/bin
fi
```

Or we teach our scripts and users they now have to run the following:

```
scl enable rh-mongodb34 'mongo'
```

beav: It would be difficult to get users to remember to run `scl enable rh-mongodb34 'mongo'` every time.

However, I think it's OK to add things to the user's path if needed.

ekohl: The downside of placing it in the profile is that not everything reads profile scripts. The postgres SCL has a package that installs /usr/bin/mongo and friends. That removes the need for the scl enable. Is there a similar package in the mongo SCL?

chris1984:
@ekohl something like this package?

https://centos.pkgs.org/7/centos-sclo-rh/rh-mongodb34-syspaths-3.0-14.el7.x86_64.rpm.html

If so should I just add it as a dep of the mongo in the spec file?

ekohl:
@chris1984 yes, exactly that. I think we should add it as a dep because then we should get a `/usr/bin/mongo` and scripts don't need to be modified

chris1984:
I screwed something up with git where it would not push to this pr, closing this one in favor of https://github.com/theforeman/foreman-packaging/pull/2587

PR #2585, opened 2018-06-01
Fixes #23615 - Redirect katello service

johnpmitsch:
Katello-service functionality has moved to foreman-maintain. We
can redirect katello-service there since it uses the same arguments.
Katello-service is used pretty widely in documenation and scripts,
I don't see a need to deprecate this just yet (we always can later).
I aligned the issue with foreman 1.19(katello 3.8), so  katello
3.7 will still have the original katello-service script.

* [x] Nightly
* [ ] 1.18
* [ ] 1.17
* [ ] 1.16

---


packages/katello/katello/katello-service
```diff
@@ -1,98 +1,7 @@
-#!/usr/bin/env ruby
+#!/usr/bin/bash
```
ekohl: bash is usually found in `/bin`. Might be better to just depend on `/bin/sh` given the simplicity of the script.

johnpmitsch:
[test rpm]

johnpmitsch:
@ekohl installing this on a nightly system doesn't seem to overwrite katello-service and the old command is still there. Am I missing something in the spec file?

johnpmitsch:
@ekohl nvm, I was only installing katello and katello-common, which were installing katello-service from local repos. When I installed the katello-service rpm from the koji build, everything works:

Here is some example output:
```
[root@centos7-katello-nightly ~]# katello-service start
Redirecting to 'foreman-maintain service'
Running Start Services
================================================================================
Check if command is run as root user:                                 [OK]
--------------------------------------------------------------------------------
Start applicable services: 
Starting the following service(s):

postgresql, rh-mongodb34-mongod, qdrouterd, qpidd, squid, pulp_resource_manager, pulp_streamer, foreman-proxy, pulp_celerybeat, pulp_workers, tomcat, dynflowd, httpd, puppetserver

| All services started                                                [OK]      
--------------------------------------------------------------------------------

[root@centos7-katello-nightly ~]# katello-service start --only httpd
Redirecting to 'foreman-maintain service'
Running Start Services
================================================================================
Check if command is run as root user:                                 [OK]
--------------------------------------------------------------------------------
Start applicable services: 
Starting the following service(s):

httpd

\ All services started                                                [OK]      
--------------------------------------------------------------------------------

[root@centos7-katello-nightly ~]# katello-service stop --only httpd
Redirecting to 'foreman-maintain service'
Running Stop Services
================================================================================
Check if command is run as root user:                                 [OK]
--------------------------------------------------------------------------------
Stop applicable services: 
Stopping the following service(s):

httpd

| All services stopped                                                [OK]      
--------------------------------------------------------------------------------

[root@centos7-katello-nightly ~]# 
```

sean797:
Should the RPM now require `/bin/sh` ?

ekohl:
@sean797 that's done automatically:
```
Processing files: katello-service-3.8.0-2.nightly.el7.noarch
Provides: katello-service = 3.8.0-2.nightly.el7
Requires(rpmlib): rpmlib(CompressedFileNames) <= 3.0.4-1 rpmlib(FileDigests) <= 4.6.0-1 rpmlib(PayloadFilesHavePrefix) <= 4.0-1
Requires: /bin/sh
```

PR #2592, opened 2018-06-04
Rename -nightly releasers to -jenkins

ekohl:
This is in line with katello and makes the releaser name consistent between various releases. Jenkins is not 100% correct since you can still deploy from local git repos but it's closer than the alternatives.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,8 @@ than mock (above).
 
 Using a local git checkout, change `source_dir` as appropriate:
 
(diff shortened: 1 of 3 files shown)


ekohl:
I'll first verify if any -infra changes need to be made and send a post on discourse for those who have scripted around it manually.

ekohl:
https://github.com/theforeman/foreman-infra/pull/668 will be needed

PR #2597, opened 2018-06-04
Update friendly_id to 5.2.4

mmoll:
this should be built into:

* [x] Nightly
* [ ] 1.18
* [ ] 1.17
* [ ] 1.16


mmoll:
This is IMHO still compatible with 1.18

PR #2600, opened 2018-06-05
Add rubygem-foreman_wreckingball package

dgoetz:

* [x] Nightly
* [x] 1.18
* [x] 1.17
* [ ] 1.16

---


dgoetz:
Added it to the manifests and tests re-run successfully.

Entry in packages_mainfests is the only thing missing when running the new `add_gem_package.sh` script, thanks for this helpful script.

ekohl:
I've merged this and released to nighly but then I realized this uses the `scl` template rather than the `foreman_plugin`. We should fix that and then backport it.

dgoetz:
Does this require more than changing the spec (and a new pull request)?

ekohl:
No, just change the spec. `add_gem_package.sh` can do that nowadays. It's very recent though so have a good look at what it generates.

PR #2602, opened 2018-06-05
Add rubygem-foreman_rescue package

dgoetz:

* [x] Nightly
* [x] 1.18
* [x] 1.17
* [ ] 1.16

---


packages/plugins/rubygem-foreman_rescue/rubygem-foreman_rescue.spec
```diff
@@ -0,0 +1,89 @@
...
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/README.md
+%{gem_instdir}/Rakefile
+%{gem_instdir}/test
+
+%changelog
+* Tue Jun 05 2018 Dirk Goetz <dirk.goetz@netways.de> 1.0.0-1
+- Add rubygem-foreman_rescue generated by gem2rpm using the scl template
```
ekohl: Should this be using the plugin template?

ekohl: Note I didn't check this for the other 2 packages but they're probably also actual plugins rather than a library.

dgoetz: Yes, I was focused on it being scl instead of nonscl so I did not recognize the foreman_plugin is scl, too.

PR #2607, opened 2018-06-05
Run repoclosure per package

ehelms:

* [x] Nightly
* [ ] 1.18
* [ ] 1.17
* [ ] 1.16

---


ehelms:
Will require https://github.com/theforeman/obal/pull/103

This will allow, for example:

```
obal repoclosure katello-host-tools
```

ehelms:
Related CI updated -- https://github.com/theforeman/foreman-infra/pull/672

ehelms:
The obal PR has been merged, this can be tested and reviewed

komidore64:
this feature doesn't seem to give a nice message when one doesn't have the `repoclosure` command installed in their machine.

komidore64:
i loaded up the obal changes from theforeman/obal#103, checked out this PR and tried to do a repoclosure check on katello-repoclosure-el7`. [no success so far](https://gist.github.com/komidore64/95a35a4c1951e4bd498e6dca5997f566). what did i mess up?

komidore64:
i needed `yum-utils` instead of `dnf-utils`.

package_manifest.yaml
```diff
@@ -11,6 +11,21 @@ foreman_packages:
...
+    repoclosure_lookaside_repos:
+      - el7-base
+      - el7-updates
+      - el7-epel
+      - el7-extras
+      - el7-scl
+      - el7-scl-sclo
+      - el7-scl-ruby
+      - el7-scl-v8
+      - el7-puppet-4
```
ekohl: I'd prefer puppet-5

package_manifest.yaml
```diff
@@ -11,6 +11,21 @@ foreman_packages:
...
+      - el7-subscription-manager
```
ekohl: Should not be needed

package_manifest.yaml
```diff
@@ -11,6 +11,21 @@ foreman_packages:
...
+      - el7-qpid
```
ekohl: Should not be needed

package_manifest.yaml
```diff
@@ -11,6 +11,21 @@ foreman_packages:
...
     - koji-foreman
+    repoclosure_config: repoclosure/yum_el7.conf
+    repoclosure_lookaside_repos:
+      - el7-base
+      - el7-updates
+      - el7-epel
+      - el7-extras
+      - el7-scl
+      - el7-scl-sclo
+      - el7-scl-ruby
```
ekohl: Don't think this is needed anymore

package_manifest.yaml
```diff
@@ -11,6 +11,21 @@ foreman_packages:
...
+      - el7-scl-v8
```
ekohl: Don't think this is needed anymore

package_manifest.yaml
```diff
@@ -11,6 +11,21 @@ foreman_packages:
...
+      - el7-puppet-4
+      - el7-subscription-manager
+      - el7-qpid
+      - el7-tfm-ror51
```
ekohl: Haven't we renamed this to `el7-foreman-rails-nightly`?

package_manifest.yaml
```diff
@@ -285,11 +300,31 @@ foreman_packages:
...
     - koji-katello
+    repoclosure_config: repoclosure/yum_el7.conf
+    repoclosure_lookaside_repos:
+      - el7-base
+      - el7-updates
+      - el7-epel
+      - el7-extras
+      - el7-scl
+      - el7-scl-sclo
+      - el7-scl-ruby
```
ekohl: See comments on `foreman_packages`

package_manifest.yaml
```diff
@@ -327,19 +362,69 @@ katello_packages:
     rubygem-redhat_access_lib: {}
     rubygem-robotex: {}
     rubygem-runcible: {}
+
 katello_client_packages:
   vars:
     package_base_dir: 'packages/katello/'
     releasers:
     - koji-katello-client
+    repoclosure_config: repoclosure/yum.conf
```
ekohl: Doesn't feel right to group all distros together. Perhaps this is really a limitation of the modeling we chose (and why it's always good to have more than 1 distro - your models are better).

komidore64:
I gave this a smoketest on a couple things. It seemed to run fine.

package_manifest.yaml
ehelms: It doesn't feel right but yum/repoclosure is smarter than I even thought. If we don't lump them together, then we have to identify and download each task per distro which is tedious. However, repoclosure if you point it at a directory takes distro into account for us.

ehelms:
@ekohl @komidore64 Updated

rel-eng/tito.props
```diff
@@ -403,26 +403,14 @@ whitelist = ansiblerole-insights-client
...
-
 [katello-nightly-rhel7]
 disttag = .el7
 scl = tfm
 whitelist = katello
   katello-certs-tools
   katello-client-bootstrap
-  katello-host-tools
   katello-installer-base
+  katello-host-tools
```
ekohl: Why are you moving this? Shouldn't we keep it sorted?

ehelms: Accidental move -- its back

rel-eng/tito.props
```diff
@@ -403,18 +403,6 @@ whitelist = ansiblerole-insights-client
...
-  katello-repos
-  python-rhsm
-  subscription-manager
-
-[katello-nightly-rhel6]
-disttag = .el6
-whitelist = katello-host-tools
-  katello-repos
-
 [katello-nightly-rhel7]
```
ekohl: I'm wondering if we should split this in `katello-server-nightly-rhel7` and `katello-client-nightly-rhel7` but that's probably out of scope for this PR.

repoclosure/yum.conf
```diff
@@ -0,0 +1,160 @@
...
+
+[el5-subscription-manager]
+name=el5-subscription-manager
+baseurl=https://repos.fedorapeople.org/candlepin/subscription-manager/epel-5/x86_64/
+
+[el5-katello-client-nightly]
+name=Katello nightly EL5
+baseurl=https://fedorapeople.org/groups/katello/releases/yum/nightly/client/el5/x86_64
+
+[f27-base]
```
ekohl: Do we need equivalent f28 configs here?

ehelms: Not yet -- f28 does not currently work due to a Koji/mock bug.

rel-eng/tito.props
ehelms: I've considered that as well, and out of scope.

PR #2618, opened 2018-06-06
Add puppet 4 repo to repoclosure

ehelms:

* [x] Nightly
* [ ] 1.18
* [ ] 1.17
* [ ] 1.16

---

### DIFF
--- a/repoclosure/yum_el7.conf
+++ b/repoclosure/yum_el7.conf
@@ -42,7 +42,7 @@ baseurl=http://mirror.centos.org/centos/7/sclo/x86_64/sclo/
 
 [el7-pl-prod]
 name=el7-pl-prod
-baseurl=http://yum.puppetlabs.com/el/7/products/$basearch/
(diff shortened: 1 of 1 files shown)


ehelms:
@dLobatog One change that needs to be made

repoclosure/yum_el7.conf
```diff
@@ -42,7 +42,7 @@ baseurl=http://mirror.centos.org/centos/7/sclo/x86_64/sclo/
 
 [el7-pl-prod]
```
ekohl: Do we use this name in the pipeline anywhere?

ehelms: We don't due to refactoring that was done. @dLobatog requested we just one-off run it but this change needs to be made

ekohl: I think we should rather be adding `[el7-puppet-4]`. Since we dropped puppet 3 we could safely drop `[el7-pl-{prod,deps}]` here.

ekohl:
Merged now, but if we do another 1.16 release we should clean it up.

PR #2621, opened 2018-06-06
Fix debian log rotation

mark-wagner:
Based on https://community.theforeman.org/t/smart-proxy-log-rotate-not-working/9904


mark-wagner:
Branch named incorrectly. Closing.

PR #2622, opened 2018-06-07
Fix debian log rotation

mark-wagner:
Based on https://community.theforeman.org/t/smart-proxy-log-rotate-not-working/9904

### DIFF
--- a/debian/stretch/foreman-proxy/logrotate
+++ b/debian/stretch/foreman-proxy/logrotate
@@ -9,6 +9,6 @@
   compress
 	daily
   postrotate
-    [ -e /etc/init.d/foreman-proxy ] && /etc/init.d/foreman-proxy logrotate >/dev/null 2>&1 || true
+    /bin/systemctl kill --signal=SIGUSR1 foreman-proxy >/dev/null 2>&1 || true
   endscript
 }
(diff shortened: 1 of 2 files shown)


ekohl:
I'm not 100% sure how Debian's handled but I think we still maintain support for both systemd and sysvinit. We should prefer systemd but otherwise fall back to the current code.

mark-wagner:
systemd has been the default init for Debian since jessie (8). Stretch is 9. It has been the default for Ubuntu since 15.04. Xenial is 16.04.

According to http://without-systemd.org/wiki/index.php/How_to_remove_systemd_from_a_Debian_Stretch_installation Debian still officially supports sysvinit. I was unable to find Ubuntu's support status but apparently it is possible: http://without-systemd.org/wiki/index.php/How_to_remove_systemd_from_an_Ubuntu_Xenial_installation

I don't want to get involved in an init war but IMHO it is time to drop support for sysvinit and eliminate the complications involved in supporting two init  systems.

mmoll:
I'm quite sure that (almost?) all our DEB users are using the default, which is systemd these days.

However, adding a check (`test -d /run/systemd/system`) here doesn't cost much and will make potential systemd haters happy, so I'm kindly asking to add that check and continue to also support init as long as it's that cheap.

ekohl:
It does sound like it's still easy to support non-systemd for now.

PR #2627, opened 2018-06-07
NPM update script

tstrachota:
Experimental script for updating npm packages based on output of `npm list`.

See the [discourse discussion](https://community.theforeman.org/t/nightlies-assets-are-now-broken/9810) for more details.

### DIFF
--- a/update_npm_dependencies.rb
+++ b/update_npm_dependencies.rb
@@ -1,0 +1,221 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+require 'json'
+
+STRATEGIES = ['single', 'bundle', '']
+
+PROJECT_PACKAGE_PATHS = {
+  'foreman' => ['foreman'],
+  'plugins' => ['plugins', 'foreman'],
+  'katello' => ['katello', 'plugins', 'foreman'],
+}
+
+
+def package_name(npm_module_name)
+  "nodejs-#{npm_module_name}"
+end
+
+def package_dir(npm_module_name, project)
+  "packages/#{project}/#{package_name(npm_module_name)}"
+end
+
+def find_package_project(npm_module_name)
+  PROJECT_PACKAGE_PATHS[@options[:project]].find do |project|
+    File.directory?(package_dir(npm_module_name, project))
+  end
+end
+
+def package_exist?(npm_module_name)
+  !!find_package_project(npm_module_name)
+end
+
+def current_version(npm_module_name)
+  project = find_package_project(npm_module_name)
+  if project
+    spec_file = "#{package_dir(npm_module_name, project)}/#{package_name(npm_module_name)}.spec"
+    version_line = File.readlines(spec_file).grep(/^\W*Version:.*/).first
+    version_line.gsub(/^\W*Version:\W*/, '').strip
+  end
+end
+
+def bundle_strategy(npm_module_name)
+  project = find_package_project(npm_module_name)
+  if project
+    dir = package_dir(npm_module_name, project)
+    zipped_dependencies = Dir["#{dir}/*.tgz"]
+    if zipped_dependencies.length > 1
+      'bundle'
+    else
+      'single'
+    end
+  else
+    @options[:default_strategy]
+  end
+end
+
+def detect_action(npm_module_name, details)
+  if package_exist?(npm_module_name)
+    if details['version'] == current_version(npm_module_name)
+      :skip
+    else
+      :update
+    end
+  else
+    :add
+  end
+end
+
+def add_package_cmd(npm_module_name, module_details)
+  strategy = bundle_strategy(npm_module_name)
+  project = find_package_project(npm_module_name) || @options[:project]
+  cmd = ''
+  if @options[:git_packager]
+    cmd += 'RPM_PACKAGER="$(git config user.name) <$(git config user.email)>" '
+  end
+  cmd += "./add_npm_package.sh #{npm_module_name} #{module_details['version']} '#{strategy}' #{project}"
+  cmd
+end
+
+def action_description(action, npm_module_name, module_details)
+  case action
+  when :skip
+    "skip   #{npm_module_name} (#{module_details['version']})"
+  when :update
+    "update #{npm_module_name} (#{current_version(npm_module_name)} -> #{module_details['version']})"
+  when :add
+    "add    #{npm_module_name} (#{module_details['version']})"
+  end
+end
+
+def format_list(lines)
+  "  " + lines.join("\n  ")
+end
+
+
+@options = {
+  :project => 'foreman'
+}
+opts = OptionParser.new do |opts|
+  opts.banner = [
+    "Usage: #{__FILE__} [options]",
+    "",
+    "Updates npm packages based on json output of npm list.",
+    "Make sure you generate the input json file without development and optional dependencies:",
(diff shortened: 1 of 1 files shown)


update_npm_dependencies.rb
```diff
@@ -0,0 +1,160 @@
...
+  if package_exist?(npm_module_name)
+    spec_file = "#{package_dir(npm_module_name)}/#{package_name(npm_module_name)}.spec"
+    version_line = File.readlines(spec_file).grep(/^\W*Version:.*/).first
+    version_line.gsub(/^\W*Version:\W*/, '').strip
+  end
+end
+
+def bundle_strategy(npm_module_name)
+  if package_exist?(npm_module_name)
+    # new packages default to bundle
```
ekohl: I prefer to not default to bundles. Only if there are too many deps that aren't present or if it would result in conflicting situations where package x requires a version package y can't deal with. You can also leave it empty and let `add_npm_package.sh` deal with this. It queries `api.npms.io` to get the number of dependencies.

update_npm_dependencies.rb
```diff
@@ -0,0 +1,160 @@
...
+  else
+    :add
+  end
+end
+
+def add_package_cmd(npm_module_name, module_details)
+  strategy = bundle_strategy(npm_module_name)
+  cmd = ''
+  if @options[:git_packager]
+    cmd += 'RPM_PACKAGER="$(git config user.name) <$(git config user.email)>" '
```
ekohl: I thought `rpmdev-packager` already looked in git by default so I'd expect this to be redundant.

update_npm_dependencies.rb
```diff
@@ -0,0 +1,160 @@
...
+      exit 1
+    end
+    @options[:package_list_file] = val
+  end
+
+  opts.on('--only NPM_MODULES', 'Comma separated list of npm modules to update, others are skipped') do |val|
+    @options[:only] = val.split(',')
+  end
+
+  opts.on('--except NPM_MODULES', 'Comma separated list of npm modules to be excluded from updating') do |val|
```
ekohl: In `update-requirements` we have a blacklist. I wonder if we could unify that somewhere in a config to prevent duplication.

update_npm_dependencies.rb
```diff
@@ -0,0 +1,160 @@
...
+
+require 'optparse'
+require 'json'
+
+def package_name(npm_module_name)
+  "nodejs-#{npm_module_name}"
+end
+
+def package_dir(npm_module_name)
+  "packages/foreman/#{package_name(npm_module_name)}"
```
ekohl: In the future we might need to make this somewhat more generic but currently `add_npm_package.sh` can't deal with this anyway. Ideally we can use the same for plugins and katello.

update_npm_dependencies.rb
tstrachota: Apparently it doesn't. Without that option all my changelogs contained "vagrant".

update_npm_dependencies.rb
tstrachota: Ok, I'll probably leave it empty for now and we can update it later.

update_npm_dependencies.rb
tstrachota: Yeah, one more option could solve this but as you say, I saw it hardcoded in the other script so I left it as it was for the first iteration.

update_npm_dependencies.rb
tstrachota: I intended the options `--only` and `--except` to be used rather for fixing issues when some packages fail or for some other exceptional cases so I'd like to keep them in regardless of `update-requirements`.

The comment in `update-requirements` says:
```
# We need a blacklist of packages that are devDependencies but not needed to build
```
Is it only dev dependencies that we're filtering now? If it's so we can sort it out with `npm list --production` that lists the modules without dev dependencies already.

tstrachota:
@ekohl updated. I changed the default strategy to empty and let `add_npm_package.sh` to decide.
But I'm afraid this makes the script a bit more fragile, because when there's a new package with dependencies that we haven't got packaged yet they will be missing in the assets.

We could probably at least print a warning about what dependencies are required and are not currently packaged.

update_npm_dependencies.rb
ekohl: I found it tricky. I opted to skip some dependencies that were only used for tests like jest. It's entirely possible I took too much of a shortcut in some places. Sadly I don't think there's a testDependencies to split it out further or bundler groups that give you complete freedom.

update_npm_dependencies.rb
ekohl: In https://github.com/theforeman/foreman-packaging/pull/2632 @ehelms took the chance to make it configurable :)

ekohl:
> But I'm afraid this makes the script a bit more fragile, because when there's a new package with dependencies that we haven't got packaged yet they will be missing in the assets.

You're right, I didn't consider that yet. https://github.com/theforeman/foreman-infra/pull/672 should ensure the PR actually fails when that's the case.

ehelms:
Seems good as a first pass. This will most likely get re-factored, re-written and moved to Obal where we want to centralize package tooling and it's written in python/Ansible. As we already have scripts, I do think itss fine to add things that increase our efficiency.

tstrachota:
I added an option for base directories that defaults to `foreman`.

@ehelms do you think it makes sense to merge this and refactor into obal later, or do you want to skip this and move it directly to obal?

update_npm_dependencies.rb
tstrachota: I'm not sure I understand. Pardon my ignorance, I probably lack some background knowledge.
So you think using `npm list --production` is sufficient or do we still need the blacklist?

If we do need the blacklist, what's the key to decide if a package should go in or be put on the blacklist?

ekohl:
@tstrachota we had some discussions within the team - not specifically about this script but more in general. We agreed it'd be fine to merge scripts in foreman-packaging, let it evolve into something robust and then refactor it into obal. So it's fine to get this merged now.

update_npm_dependencies.rb
ekohl: Odd, we should look into that when we include it in obal.

update_npm_dependencies.rb
```diff
@@ -0,0 +1,201 @@
...
+
+def package_name(npm_module_name)
+  "nodejs-#{npm_module_name}"
+end
+
+def package_dir(npm_module_name)
+  "packages/#{@options[:package_dir]}/#{package_name(npm_module_name)}"
+end
+
+def package_exist?(npm_module_name)
```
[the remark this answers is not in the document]
tstrachota: Ha, good catch!

tstrachota:
@ekohl updated. Detecting the correct target directory made the script a bit more complicated. I renamed `--package-dir` to `--project` and try to find the package in relevant directories now.

tstrachota:
@ekohl ping. Not that I would insist on merging this but I'd like to know what's the status :)

ohadlevy:
@tstrachota @ekohl whats the status of this PR? thanks

ehelms:
@ekohl current state of scripts, do you feel this is needed?

tbrisker:
Looks like there has been no activity and this is likely not valid any more with the current code, thanks everyone! feel free to reopen if you want to work on this.

PR #2633, opened 2018-06-10
katello-host-tools 3.3.1

jturel:

* [x] Nightly
* [x] 1.18
* [ ] 1.17
* [ ] 1.16

---

We can now get away with two versions of gofer. Due to a bug spanning the katello-agent and gofer, goferd minumum reqs have been update to reflect this. 2.12 can only be used on PY2.7+ systems so that's why we have 2.11.5 and 2.12.1


jturel:
Where do we get gofer RPMs from?

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -70,10 +70,16 @@ Group:      Development/Languages
 
 Conflicts: pulp-consumer-client
 
+%if %{dnf_install}
+Requires: gofer >= 2.12
+%else
 %if 0%{?rhel} == 5
 Requires: gofer >= 2.5
+Requires: gofer < 2.11
```
ekohl: Does it actually not work with `2.11` or higher?

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -70,10 +70,16 @@ Group:      Development/Languages
...
 %else
-Requires: gofer >= 2.7.6
+Requires: gofer >= 2.11
+Requires: gofer < 2.12
```
ekohl: Does it actually not work with `2.12`?

packages/katello/katello-host-tools/katello-host-tools.spec
jturel: Untested, but I simply don't envision moving el5 clients from 2.7 to 2.11 do you?

packages/katello/katello-host-tools/katello-host-tools.spec
jturel: 2.12 requires python 2.7 and above so it will work on el7+

ehelms:
[test rpm]

ehelms:
[test rpm]

jturel:
I've received intel that there is an updated gofer we will need coming along which will also need changes in our agent so this will be changing when we have those things in place.

ehelms:
[test rpm]

ehelms:
[test rpm]

ekohl:
What's the status of this now?

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -229,6 +231,10 @@ rm -rf %{buildroot}
 
 %if %{build_agent}
 %post -n katello-agent
+%if %{dnf_install}
+sed 's/bin\/python/bin\/python3/' /etc/sysconfig/goferd -i
```
jturel: This is to make sure goferd starts under the correct python version.

jturel:
@ekohl updated. I've tested that this installs  & goferd starts &loads the katello agent without errors under the correct version of python on all dists except el5 which I would like to do before this is merged, but I expect this to work as-is.

jturel:
@ehelms for your awareness!

packages/katello/katello-host-tools/katello-host-tools.spec
```diff
@@ -70,17 +70,19 @@ Group:      Development/Languages
 
 Conflicts: pulp-consumer-client
 
-%if 0%{?rhel} == 5
-Requires: gofer >= 2.5
+%if %{legacy_agent}
+Requires: gofer >= 2.11.5
+Requires: gofer < 2.12
```
jturel: Remember, 2.12 of gofer requires python 2.7+

jturel:
@ehelms @jlsherrill one or both may want to take a look since it reflects what we talked about last week

ehelms:
[test rpm]

PR #2668, opened 2018-06-18
Relax smart_proxy_omaha nokogiri dependency

ekohl:


ekohl:
The failed koji task is from the previous run.

PR #2673, opened 2018-06-19
Bump NPM dependencies

ekohl:
For https://github.com/theforeman/foreman/pull/5708


ekohl:
It's good to note that this is the first to use a subdirectory in the sources because of `@x/y` package names in NPM. Looks like it builds, but it's good to verify this.

tbrisker:
foreman.spec needs bumping url-loader dependencies, nightlies are now red - http://koji.katello.org/kojifiles/work/tasks/3751/113751/root.log

ekohl:
I just noticed that as well. https://github.com/theforeman/foreman-packaging/pull/2794

PR #2686, opened 2018-06-21
Create a rubygem-katello-assets subpackage

ekohl:


ekohl:
@dLobatog I think it still makes sense. This should allow users to get all the assets, make a change and rebuild the UI. While talking to users I do know some admins make a small change in the UI to test it out. For them it's easier than to set up a development environment. That was very easy with the old ERB templates but without this assets package next to impossible.

We should also look into moving files from the old asset pipeline into the -assets package because I suspect they're not needed on a production installation.

I do remember that I should add the subpackage to comps so it's actually published.

PR #2503, opened 2018-05-17
Bump rubygem-hammer_cli_katello to 0.14.0

akofink:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [ ] 1.17
* [ ] 1.16
* [ ] 1.15

See Foreman's [plugin maintainer documentation](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

PR #2689, opened 2018-06-25
Update rubygem-hammer_cli_katello to 0.14.0

komidore64:

* [x] Nightly
* [ ] 1.18
* [ ] 1.17
* [ ] 1.16

---


komidore64:
#2503 was closed, but I think it shouldn't have been.

komidore64:
> I don't like this change so much. I'd prefer a obal nightly command. In the short term I think our nightly job should explicitly set the releaser. Tomorrow I'll write up my thoughts on this.

Why? This is how Katello's nightly release works.

komidore64:
We release nightly packages much more often than standard releases, so it makes sense for the nightly release to be the default.

An `obal nightly` command does sound convenient, but it's doing the same thing as `obal release`. It's still a release.

ekohl:
The reason it's annoying is that it makes branching very verbose. You can end up with pretty much rewriting the whole manifest. Our goal was to have as little manual work for branching as possible. Since we can't automatically change the manifest we're building technical debt into our process.

As an example: https://github.com/theforeman/foreman-packaging/commit/a47a808eaea7a02e40a710cfdde23fa78cb55f40#diff-9297bc45c5e9bdf810aa55f6c91853dc is with just a few nightly packages. hammer-cli-katello isn't even in there. If we plan to add nightly plugins we'll likely double the amount.

We'll probably still need to modify all the Jenkins jobs in the releasers args but that's a much smaller change compared to also changing all the releasers. Diffs between a release and nightly will be smaller as well.

In short: releasing and nightly apply the same role but from a different context which makes it a different playbook with their own config.

ekohl:
Another thing that I haven't mentioned yet but is where I want to go with nightlies is transforming specs to update dependencies. I want to integrate https://github.com/theforeman/foreman-packaging/tree/rpm/develop#updating-nightly-dependencies into the nightly build process. I'm not sure yet if obal is the right place or we should continuously send in PRs/commits to foreman-packaging but it's good to consider the longer term strategy.

packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
```diff
@@ -8,7 +8,7 @@
 
 Summary: Katello command plugin for the Hammer CLI
 Name:    %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.13.2
+Version: 0.14.0
```
ekohl: The problem with this is that we can't test changes. The current releaser can't find the actual gem. Changing the releaser to nightly means we can't actually release. I still think that we should keep this package spec + releaser to production settings and explicitly set the releaser in the nightly job until obal learns how to build nightly packages.

komidore64: Would you mind explaining more what is preventing us from testing changes or doing an "actual release?" It's not that I want to argue, but I am having trouble understanding what mechanisms wouldn't work.

ekohl: When we release or test (scratch) this releaser is used. It means that if we would want to release an actual `0.14.0` then you need to override the releaser. In a PR that can't happen unless you change `package_manifest.yaml`. Currently the hammer (plugin) maintainer sends in PRs to update the version. These 2 workflows are incompatible.

We either need to change it so the release manager includes hammer updates when doing a Foreman/Katello release or keep the current workflow.

komidore64: When a PR is made to this repo (foreman-packaging) that kicks off the [foreman-packaging-rpm-pr-test](http://ci.theforeman.org/job/foreman-packaging-rpm-pr-test/) which `obal scratch` builds packages using what is listed in the package manifest.

The changes introduced in theforeman/foreman-infra#708 are for a different jenkins job ([hammer-cli-katello-master-release](http://ci.theforeman.org/job/hammer-cli-katello-master-release)) which clones the project repo, builds a gem locally, checks out f-packaging and submits an `obal release` using custom releasers and extra vars.

It sounds like these two workflows won't interfere with one another?

ekohl: Actually this was already a problem when I tried to `obal release` 54d77e1a17d076b3be63e97119bc2a0ca7b1531f. I built a nightly package twice before I realized that it doesn't work that way. That's my main problem. As I understand it now:

For foreman and foreman-{installer,proxy,selinux} we never check in the source in nightly. That's only done in a branch. That's no problem because such a release (1.18.0) by definition only ever lives in a branch (1.18-stable).

For hammer we check in the source gem and release from nightly. Then we cherry pick to older releases as needed. As the linked commit: it's supposed to live in 1.18, 1.19 and nightly. Because there is no dist release we must tag it in koji.

I don't know what's the best workflow here. Should we add a dist tag and just build it once for every branch? Then we should remember that after branching we explicitly need to include a hammer release where we now inherit it automatically from nightly.

ekohl:
My alternative implementation is in https://github.com/theforeman/foreman-packaging/pull/2732. I think there's still some room for improvement to make branches less noisy.

komidore64:
closing in favor of #2732.

PR #2694, opened 2018-06-27
Remove history, warning, and invariant node rpm packages

tbrisker:
Otherwise, when katello builds it expects them to be available via
vendor.js instead of in katello.js

* [x] Nightly
* [x] 1.18
* [ ] 1.17
* [ ] 1.16

---


ekohl:
`Require` is a runtime dependency. Your repoclosure will be broken but you can build it just fine.

dLobatog:
Thanks, I would also go that route then, making `bundled-npm(nodejs-react-router-dom)` to workaround the issue. As far as I can tell from the nightlies, at least nodejs-warning is requested but not found (which breaks the entire page).

screenshot from 2018-06-27 11-26-47

tbrisker:
I'm not sure what will happen if a plugin directly requires them, but I don't want to keep katello broken on a hypothetical question. 
I didn't notice that react-router-dom was not bundled but required theses, it definitely should be to fix this right now. I think we shouldn't include any unbundled modules in plugins atm (unless they are just a single module).

The other path we can go down which we've discussed multiple times before is to stop packaging node modules as rpm but rather push them into the tarballs (which in this case will make warning available as top level in foreman core as well due to npm deduping).

ekohl:
> I'm not sure what will happen if a plugin directly requires them, but I don't want to keep katello broken on a hypothetical question.

That's fine, I'm just pointing out the risk we do have. It's always good to be aware of the risks because other plugins might have the same problem.

For now, can you regenerate react-router-dom? Probably enough to use `./add_npm_package.sh react-router-dom <VERSION> bundle katello` and then ensure the `release` is bumped to 2 with the appropriate changelog entry.

> I didn't notice that react-router-dom was not bundled but required theses, it definitely should be to fix this right now. I think we shouldn't include any unbundled modules in plugins atm (unless they are just a single module).

That could be fair and maybe a good guideline. It's what @tstrachota started in https://github.com/theforeman/foreman-packaging/pull/2627 as well.

tbrisker:
added the change to react-router-dom

tbrisker:
@ekohl looks like this is green now

ekohl:
Let's see if this works.

PR #2697, opened 2018-06-27
Fix fedora comps for katello client

jturel:
(cherry picked from commit b2afa29289df6212e34ebf9b97651cef1e19dbce)


jturel:
@ehelms here's the pick

PR #2716, opened 2018-07-03
Fix linking of webpack assets on debian

tbrisker:
Also, only package the files we need from the build.

* [x] Nightly
* [x] 1.18
* [ ] 1.17
* [ ] 1.16

---

### DIFF
--- a/plugins/ruby-foreman-remote-execution/debian/postinst
+++ b/plugins/ruby-foreman-remote-execution/debian/postinst
@@ -51,6 +51,10 @@ if [ -f /usr/share/foreman/config/database.yml ]; then
   fi
 fi
 
+PLUGIN_ROOT=`$BUNDLE show $PLUGIN`
+mkdir -p $PLUGIN_ROOT/public
+ln -s /var/lib/foreman/public/webpack/ $PLUGIN_ROOT/public/webpack
+
 # Own all the core files
 chown -Rf foreman:foreman '/usr/share/foreman'
 
--- a/plugins/ruby-foreman-remote-execution/debian/changelog
+++ b/plugins/ruby-foreman-remote-execution/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-remote-execution (1.5.3-3) stable; urgency=low
+
+  * Fix links to webpack assets 
+
+ -- Tomer Brisker <tbrisker@gmail.com>  Tue, 03 Jul 2018 15:11:04 +0300
+
 ruby-foreman-remote-execution (1.5.3-2) stable; urgency=low
 
   * add precompiled webpack assets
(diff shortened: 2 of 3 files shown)


mmoll:
```
20:42:26 GEM_PATH=$(cd /usr/share/foreman && /usr/bin/foreman-ruby /usr/bin/bundle show foreman_remote_execution) && \
20:42:26 	cp -rp ${GEM_PATH}/public/assets/ ./ && \
20:42:26 	mkdir ./webpack && \
20:42:26 	cp -rp ${GEM_PATH}/public/webpack/foreman_remote_execution-* ./webpack/ && \
20:42:26 	cp -rp ${GEM_PATH}/public/webpack/manifest.json ./webpack/

20:42:27 cp: cannot stat '/usr/share/foreman/vendor/ruby/2.3.0/gems/foreman_remote_execution-1.5.3/public/webpack/foreman_remote_execution-*': No such file or directory
20:42:27 debian/rules:15: recipe for target 'build' failed
```

tbrisker:
Updated this PR to only fix the linking, I'm not sure about the issue with copying but I think it can be handled separately (the only impact is shipping some files we don't have to, which we already do)

plugins/ruby-foreman-remote-execution/debian/postinst
```diff
@@ -51,6 +51,10 @@ if [ -f /usr/share/foreman/config/database.yml ]; then
   fi
 fi
 
+PLUGIN_ROOT=`$BUNDLE show $PLUGIN`
+mkdir -p $PLUGIN_ROOT/public
+ln -s /var/lib/foreman/public/webpack/ $PLUGIN_ROOT/public/webpack
```
ekohl: In RPM packaging we have 2 macros:

```
%%%{name}_webpack_plugin %%{gem_instdir}/public/webpack/%%{gem_name}
# foreman_webpack_plugin $PLUGIN_ROOT/public/webpack/$PLUGIN

%%%{name}_webpack_foreman %%{foreman_dir}/public/webpack/%%{gem_name}
# foreman_webpack_foreman /var/lib/foreman/public/webpack/$PLUGIN
```

The symlink is also the other way around. A more equivalent version would be:

```
if [ -e $PLUGIN_ROOT/public/webpack/$PLUGIN ] ; then
  mkdir -p /var/lib/foreman/public/webpack
  ln -snf $PLUGIN_ROOT/public/webpack/$PLUGIN /var/lib/foreman/public/webpack/$plugin`
fi
```

I also don't see the asset precompilation here so I'm wondering if the compiled files even exist.

tbrisker: @ekohl The asset compilation occurs in `rules`, and the files are copied there in `install`. I don't mind flipping this around but my knowledge of deb packaging is close to none.

ekohl:
Has anyone verified this works? I still think the symlink should be the other way around.

tbrisker:
@ekohl it does, i just verified it works on nightly. You're welcome to flip the link direction if you think that is better, I just wanted to mess with this as little as possible. we'll need this in 1.19 as well btw.

ekohl:
If the link works that's probably fine but without `-nf` I think this will fail on upgrades because it'll resolve the target rather replace it with the new location.

tbrisker:
@ekohl - opened #2789

PR #2721, opened 2018-07-04
Update foreman_ansible to 2.2.5

ares:

* [x] Nightly
* [x] 1.18
* [ ] 1.17
* [ ] 1.16

---

### DIFF
--- a/plugins/ruby-foreman-ansible/debian/gem.list
+++ b/plugins/ruby-foreman-ansible/debian/gem.list
@@ -1,3 +1,3 @@
 # list of gem urls to download via Jenkins, space separated
-GEMS="https://rubygems.org/downloads/foreman_ansible-2.0.1.gem"
-GEMS="$GEMS https://rubygems.org/downloads/foreman_ansible_core-2.0.2.gem"
+GEMS="https://rubygems.org/downloads/foreman_ansible-2.2.6.gem"
+GEMS="$GEMS https://rubygems.org/downloads/foreman_ansible_core-2.1.2.gem"
--- a/plugins/ruby-foreman-ansible/debian/rules
+++ b/plugins/ruby-foreman-ansible/debian/rules
@@ -16,13 +16,15 @@ build:
 	cp $(PLUGIN).rb /usr/share/foreman/bundler.d/
 	cd /usr/share/foreman && ( \
 		/usr/bin/foreman-ruby /usr/bin/bundle install --local && \
+		/usr/bin/npm install --no-optional --unsafe-perm && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake db:migrate RAILS_ENV=development && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
 		OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
 		)
 	GEM_PATH=$$(cd /usr/share/foreman && /usr/bin/foreman-ruby /usr/bin/bundle show $(PLUGIN)) && \
-    cp -rp $${GEM_PATH}/public/assets/$(PLUGIN) ./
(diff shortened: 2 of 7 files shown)


mmoll:
Does this need webpack asset precompilation?

I think this is only compatible with Foreman 1.18+? If yes, the respective versions constraints need to get updated.

ares:
Oh, I guess that should have been bumped with 2.2.0, good catch. I'll bump Foreman dependencies to 1.18

ares:
@mmoll I bumped Foreman versions and yes it needs webpact assets precompilation. Sorry for the delay, is there something else neeed? Thanks

ares:
putting additional bump on top of this, before this got in, 2.2.2 was released :-)

ares:
and bumped to 2.2.3 because missed dependency. I'd appreciate some help with this, 2.2.3 now depends on ipaddress, which is also a dependency of various fog adapters. That means the gem is bundled with Foreman, but this is a plugin. Should we specify that dependency and package separately? Wouldn't that potentially conflict with bundled version?

ares:
I see, meanwhile I pushed 2.2.4 here too, waiting until the webpack is resolved

mmoll:
https://github.com/theforeman/foreman-packaging/commit/997b24278771f37ac4fa741fb291b90f0dbc4063 and https://github.com/theforeman/foreman-packaging/commit/b71d8ac9b97e9622b13596089f72b84ffb296878 are the respective commits for _rex, please add the webpack stuff similarly here.

plugins/ruby-foreman-ansible/debian/gem.list
```diff
@@ -1,3 +1,3 @@
 # list of gem urls to download via Jenkins, space separated
-GEMS="https://rubygems.org/downloads/foreman_ansible-2.0.1.gem"
+GEMS="https://rubygems.org/downloads/foreman_ansible-2.2.4.gem"
 GEMS="$GEMS https://rubygems.org/downloads/foreman_ansible_core-2.0.2.gem"
```
mmoll: this should get updated to 2.1.1 now, I guess?

mmoll:
https://github.com/theforeman/foreman_ansible/commit/e258437698c7cc5115f5c7b3673bea4cb84273e6#commitcomment-29794543

plugins/ruby-foreman-ansible/debian/gem.list
ares: I don't think that is necessary, the gemspec says ~> 2.0 but probably a good idea to update anyway

ares:
also applied https://github.com/theforeman/foreman-packaging/commit/6afb757916637955506383e69f742cd712ea5cc8 for the webpack, now repushed for 2.2.5 release with the webpack patches

I will remove the pin for net-ssh from gemspec so it will be fixed in next releases

plugins/ruby-foreman-ansible/debian/gem.list
ares: in fact 2.0.2 does not require net-ssh so how about we keep it as is so we can release 2.2.5 and bump core dependency with next version where we'll have net-ssh unpinned?

plugins/ruby-foreman-ansible/debian/rules
```diff
@@ -16,13 +16,15 @@ build:
...
 		/usr/bin/foreman-ruby /usr/bin/bundle install --local && \
+		/usr/bin/npm install --no-optional --unsafe-perm && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake db:migrate RAILS_ENV=development && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
 		OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
 		)
 	GEM_PATH=$$(cd /usr/share/foreman && /usr/bin/foreman-ruby /usr/bin/bundle show $(PLUGIN)) && \
-    cp -rp $${GEM_PATH}/public/assets/$(PLUGIN) ./
+	cp -rp $${GEM_PATH}/public/assets/ ./ && \
```
tbrisker: This needs a matching change in `install` (or we can keep it the same there and here)

mmoll:
updated to 2.2.6 and merged, díky @ares!

PR #2731, opened 2018-07-09
Refs #24213 - Upgrade fog-vsphere gem version to 2.3.0

chris1984:

* [x] Nightly
* [ ] 1.18
* [ ] 1.17
* [ ] 1.16

---

### DIFF
--- a/packages/foreman/rubygem-fog-vsphere/fog-vsphere-2.1.1.gem
+++ b/packages/foreman/rubygem-fog-vsphere/fog-vsphere-2.1.1.gem
@@ -1,1 +1,0 @@
-../../../.git/annex/objects/Mv/Kf/SHA256E-s69632--ef7d6d9de2b575df0a61f8cd3368bcd218f347cd28668a311c5b9f9541e9f5f2.1.gem/SHA256E-s69632--ef7d6d9de2b575df0a61f8cd3368bcd218f347cd28668a311c5b9f9541e9f5f2.1.gem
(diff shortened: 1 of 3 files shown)


dLobatog:
@chris1984 Is there any reason why this shouldn't go in 1.18, 1.17, 1.16? I'm wondering because someone asked today  on IRC as they want https://github.com/fog/fog-vsphere/commit/6da0386a4e0e993f746f954756b39c9899c616e7

ares:
definitely should work with 1.18

dLobatog:
Cherry-picked and tagged in 1.18, 1.17 and 1.16.  cc @sbernhard as afaik you needed it http://koji.katello.org/koji/buildinfo?buildID=29892

dLobatog:
@xprazak2 Could you sign the package for 1.17 and 1.18 inclusion? https://github.com/dLobatog/foreman_release/blob/master/download-builds.sh is what I did

xprazak2:
Signed for 1.17 and 1.18.

timogoebel:
We got reports that this broke vm creation for users, see https://community.theforeman.org/t/foreman-1-16-2-vmware-fails/10565.

We also need to have https://github.com/theforeman/foreman/pull/5756 in 1.16+.

tbrisker:
1.16 is no longer supported, and we won't have another release in that line. looks like we need to remove these packages from the 1.16 repo since it is incompatible with 1.16, to ensure users of 1.16 don't face issues due to the mismatching versions.

ekohl:
For completeness, we have untagged it from 1.16 and 1.17. Even that patch it created a regression in 1.17.2 so we never released that and reverted it in 1.17.3.

PR #2732, opened 2018-07-10
Introduce a nightly_releaser variable

ekohl:


ehelms:
What solution does variablizing it solve to help with context?

ekohl:
I thought it could help in the Jenkins job but there's probably no easy way to use that variable in an override. https://github.com/theforeman/obal/pull/110 is probably a better solution.

package_manifest.yaml
```diff
@@ -343,7 +345,7 @@ katello_packages:
     rubygem-hammer_cli_katello: {}
     rubygem-katello:
       releasers:
-        - koji-katello-jenkins
+        - "{{ nightly_releaser }}"
```
komidore64: @ehelms correct me if I'm wrong, but isn't this going to break releasing Katello? Even for actual Katello releases, it gets the source from the last successful Jenkins nightly build.

ekohl: It shouldn't. This change should be a noop but when https://github.com/theforeman/obal/pull/110 is merged we also need to set `nightly_package_tito_releaser_args`. I can update the PR so they're compatible.

Thinking about it, I'm wondering if we need a `release_is_nightly` parameter for jobs that are only nightly. Then after branching we can flip the boolean when we add the actual released sources.

komidore64: theforeman/obal#110 has been merged.

ekohl:
Updated. I think this config now matches reality. With this Jenkins can be modified to `obal nightly` and with the args for job name and job id. It allows still releasing hammer with the included gems while releasing nightlies.

PR #2735, opened 2018-07-11
Correct call to update-requirements

ekohl:
Between the initial version and the final version the argument order was
changed. This file was not updated accordingly.


ares:
Thanks, fixes the issue :+1:

PR #2748, opened 2018-07-13
Drop initscripts from Foreman

ehelms:

* [x] Nightly
* [ ] 1.18
* [ ] 1.17
* [ ] 1.16

---


packages/foreman/foreman/foreman.spec
```diff
@@ -33,6 +33,7 @@ Source8: %{name}.gpg
 Source9: message_encryptor_extensions.rb
 Source10: %{executor_service_name}.sysconfig
 Source11: %{executor_service_name}.service
+SOURCE12: foreman
```
ekohl: I think this should be lower case

packages/foreman/foreman/foreman.service
```diff
@@ -0,0 +1,16 @@
...
+Description=Foreman
+Documentation=https://theforeman.org
+After=network.target remote-fs.target nss-lookup.target
+
+[Service]
+Type=forking
+User=foreman
+TimeoutSec=90
+WorkingDirectory=/usr/share/foreman
+ExecStart=/usr/sbin/foreman
```
ekohl: Is it not possible to inline `scl enable tfm "rails server -e $FOREMAN_ENV -p $FOREMAN_PORT"` here?

packages/foreman/foreman/foreman.sysconfig
```diff
@@ -1,15 +1,15 @@
...
+FOREMAN_PORT=3000
 
 # the user which runs the web interface
 #FOREMAN_USER=foreman
 
 # the rails environment in which foreman runs
-#FOREMAN_ENV=production
+FOREMAN_ENV=production
 
 # if we're using passenger or not
```
ekohl: If this part is ignored I think we can remove it.

packages/foreman/foreman/foreman.spec
```diff
@@ -1206,6 +1204,9 @@ rm -rf %{buildroot}
 %{_tmpfilesdir}/%{name}.conf
 
 # Service
+%{_sbindir}/%{name}
+%config(noreplace) %{_sysconfdir}/sysconfig/
```
ekohl: Is this missing `%{name}` at the end?

packages/foreman/foreman/foreman.service
ehelms: No -- you need a full path or systemd complains. Also, trying ot using `/usr/bin/tfm-ruby` just failed hard for me.

ekohl: And `/usr/bin/scl ...`?

ehelms:
Updated -- I bumped the timeout to 300 as well

packages/foreman/foreman/foreman.sysconfig
```diff
@@ -1,18 +1,12 @@
 # the location where foreman is installed
-#FOREMAN_HOME=/usr/share/foreman
+FOREMAN_HOME=/usr/share/foreman
```
ekohl: This variable can now be removed since it's unused

packages/foreman/foreman/foreman.sysconfig
```diff
@@ -1,18 +1,12 @@
...
 
 # the port which foreman web server is running at
 # note that if the foreman user is not root, it has to be a > 1024
-#FOREMAN_PORT=3000
+FOREMAN_PORT=3000
 
 # the user which runs the web interface
```
ekohl: Can also be removed

ehelms:
@ekohl Updated

packages/foreman/foreman/foreman.sysconfig
```diff
@@ -1,18 +1,6 @@
-# the location where foreman is installed
-#FOREMAN_HOME=/usr/share/foreman
-
 # the port which foreman web server is running at
 # note that if the foreman user is not root, it has to be a > 1024
-#FOREMAN_PORT=3000
-
-# the user which runs the web interface
-#FOREMAN_USER=foreman
+FOREMAN_PORT=3000
```
ekohl: Technically you could leave these commented out for a smaller diff but it's not that important.

packages/foreman/foreman/foreman.spec
```diff
@@ -1206,6 +1201,9 @@ rm -rf %{buildroot}
 %{_tmpfilesdir}/%{name}.conf
 
 # Service
+%{_sbindir}/%{name}
```
ekohl: This file no longer exists and causes the build to fail

ehelms:
Updated

packages/foreman/foreman/foreman.service
```diff
@@ -0,0 +1,17 @@
...
+Documentation=https://theforeman.org
+After=network.target remote-fs.target nss-lookup.target
+
+[Service]
+Type=simple
+User=foreman
+TimeoutSec=300
+WorkingDirectory=/usr/share/foreman
+ExecStart=/usr/bin/scl enable tfm "rails server -e $FOREMAN_ENV -p $FOREMAN_PORT"
+Environment=FOREMAN_ENV=production FOREMAN_PORT=3000
```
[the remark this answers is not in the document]
ekohl: By default we are but have supported a standalone service for a long time. With switching it to systemd we've dropped some more technical debt. It would be interesting to see if we can use Puma and set up Apache as a reverse proxy to use websockets, but that wasn't the direct goal of this change.

PR #2765, opened 2018-07-17
Prepare Katello 3.8 RC1

ehelms:

* [ ] Nightly
* [x] 1.19
* [ ] 1.18
* [ ] 1.17

---


ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
This is failing at repoclosure and won't pass yet.

ehelms:
Truth, but we don't support upgrades between RCs afaik and unfortunately this is how the versioning the gem is so the RPM should follow suit. So in the end, the proper version should be in the final GA repo.

ehelms:
I will note I have tried many times to manipulate the spec to work the old way of putting the RC into the release but it's difficult given the way gem actions work.

ekohl:
We use a single yum repository. I suspect users might install the latest RC instead of GA unintentionally so we should avoid it even being possible.

ekohl:
It looks like you should be able to use the `prerelease` macro. It is included in all rubygems macros, like:

```
%gem_instdir %{gem_dir}/gems/%{gem_name}-%{version}%{?prerelease}
%gem_extdir_mri %{gem_archdir}/ruby/%{gem_name}-%{version}%{?prerelease}
%gem_libdir %{gem_instdir}/lib
%gem_cache %{gem_dir}/cache/%{gem_name}-%{version}%{?prerelease}.gem
%gem_spec %{gem_dir}/specifications/%{gem_name}-%{version}%{?prerelease}.gemspec
%gem_docdir %{gem_dir}/doc/%{gem_name}-%{version}%{?prerelease}
```

That should allow you to exclude it from `Version` and put it in `Release`. It's ugly, but looks like it's intended for this.

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -4,19 +4,26 @@
 %global foreman_min_version 1.18.0
 %global plugin_name katello
 %global gem_name katello
-# %%global prever .rc1
+%global prerelease .rc1
 %global mainver 3.8.0
-%global release 1.nightly
+%global release 1
+
+%global gem_instdir %{gem_dir}/gems/%{gem_name}-%{version}%{?prerelease}
```
ekohl: These macros are part of `rh-ruby24-rubygem-devel` and you shouldn't need to duplicate them.

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -340,5 +347,8 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/webpack
 
 %changelog
+* Tue Jul 17 2018 Eric D. Helms <ericdhelms@gmail.com> 3.8.0-0.1.rc1
+- rebuilt
```
ekohl: Please fix this here and in other places :)

packages/katello/katello-installer/katello-installer-base.spec
```diff
@@ -2,16 +2,16 @@
...
 Name:    katello-installer-base
 Version: 3.8.0
-Release: 1.nightly%{?prever}%{?dist}
+Release: 0.1%{?prerelease}%{?dist}
 Summary: Puppet-based installer for the Katello and Katello Capsule
 Group:   Applications/System
 License: GPLv3+ and ASL 2.0
 URL:     https://theforeman.org/plugins/katello
-Source0: https://fedorapeople.org/groups/katello/releases/source/tarball/katello-installer-%{version}%{?prever}.tar.gz
+Source0: https://fedorapeople.org/groups/katello/releases/source/tarball/katello-installer-%{version}.tar.gz
```
ekohl: Shouldn't this still include the prerelease part? I'm also missing the annexed file here. If the build succeeds IMHO that indicates an issue by itself because our build pipeline should require it.

packages/katello/katello-installer/katello-installer-base.spec
```diff
@@ -64,7 +64,7 @@ A set of tools for installation of Katello and Katello Capsule,
 including Foreman and Foreman Proxy.
 
 %prep
-%setup -q -n katello-installer-%{version}%{?prever}
+%setup -q -n katello-installer-%{version}
```
ekohl: Shouldn't this still include the prerelease part?

packages/katello/rubygem-katello/rubygem-katello.spec
ehelms: Is the indentation off?

ehelms:
Updated

ehelms:
@ekohl I didn't realize they were built in macros, and I can't explain the installer passing. But all is updated now with the installer annexed source.

PR #2776, opened 2018-07-19
Update foreman_ansible_core to 2.1.1

ares:

* [x] Nightly
* [x] 1.19
* [x] 1.18
* [ ] 1.17

---

### DIFF
--- a/dependencies/bionic/foreman_ansible_core/changelog
+++ b/dependencies/bionic/foreman_ansible_core/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-ansible-core (2.1.2-1) stable; urgency=low
+
+  * Update foreman_ansible_core to 2.1.2
+
+ -- Marek Hulan <mhulan@redhat.com>  Thu, 19 Jul 2018 14:30:44 +0200
+
 ruby-foreman-ansible-core (2.0.2-1) stable; urgency=low
 
   * Initial release for Ubuntu/bionic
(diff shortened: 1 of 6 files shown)


mmoll:
according to the gemspec, net-ssh ~> 4.0 was added as dependency?

mmoll:
https://github.com/theforeman/foreman_ansible/commit/e258437698c7cc5115f5c7b3673bea4cb84273e6#commitcomment-29794543

ares:
just opened https://github.com/theforeman/foreman_ansible/pull/194

mmoll:
I took the liberty to update the control file to the actual gem dependencies. As only Ubuntu/bionic has net-ssh >4, Debian/stretch and Ubuntu/xenial would need net-ssh added in our repo or the net-ssh requirement needs to get lowered... Was there an actual feature required to use >4?

ares:
> Was there an actual feature required to use >4?

No I think it was just developed with 4, I tested with 5 but most likely works with 3 too. I think the recent things we added was ssh key support for REX but since they don't ping anything, we should be fine without too https://github.com/theforeman/foreman_remote_execution/blob/master/foreman_remote_execution_core.gemspec#L21

So, should we drop the pin from gemspec entirely?

mmoll:
I have no idea of the differences between their versions, for the currently supported DEB platforms, 3.0.1 is the lowest version...

mmoll:
updated to 2.1.2 and merged, díky @ares!

PR #2793, opened 2018-07-24
Enable gpgcheck

jturel:


jturel:
@ehelms looks like we have gpgkey set for all the repos so I set gpgcheck=1 for all of them

PR #2795, opened 2018-07-24

ekohl:


ehelms:
You might need to rebase this and a release bump would be nice :)

ekohl:
This is a nightly package so a release bump shouldn't be needed.

ehelms:
@ekohl Given this changes the spec, I think we should be bumping the release to reflect that even on nightly packages. Release generally reflects the "version" of the spec file and I think it would be good to track that over time.

ekohl:
IMHO it adds a lot of noise. We don't track the nightly tarballs either so it's not like you can just build this spec anyway. I'd like to get to the point where we update the deps based on the downloaded tarball in some way, but not entirely sure yet how. Then we'd only update all deps + version/release when there's an actual release. Because I'm unsure yet how, I'd like to continue the current process we we don't bump nightlies

ehelms:
I fail to see how it adds noise. I have run into cases where I look at a nightly RPM and I have to dig down into it to determine if it just includes source changes or changes to the spec or both. To me, this would help clarify when things are broken that I am, for example, `foreman-1.20.0-0.5.gitstuff` and that I was previously passing with something that had 0.4 which tells me spec changes occurred and maybe those broke it. I see it as, I can more quickly get to the set of changes that maybe broke things in nightly. And yes, the field is called 'release' but I find it easier to think of as specfile versioning and not source versioning.

ekohl:
We would have a lot of changelogs and for a normal user the changelog wouldn't be very useful. We could prune those on branching though.

ehelms:
I think that all depends on how you look at a changelog. The changelog represents both changes to your software and changes to your packaging of that software. Knowing we changed dependencies could be quite useful to a user.

ekohl:
That's true, but if we do it automatically you'll maybe have 20 entries of 'Updated dependencies' in a row.

ekohl:
Updated

PR #2796, opened 2018-07-25
Cleanup gofer requirements

jturel:
(cherry picked from commit 4d57f9197d177db8f51588c555aef244c5657309)


jturel:
@ehelms staging repos didn't look quite right - found the problem :+1:

PR #2799, opened 2018-07-25
Switch Foreman specs to use prerelease

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,17 +1,15 @@
-# set and uncomment all three to set alpha tag
-#global alphatag RC1
-#global dotalphatag .%{alphatag}
-#global dashalphatag -%{alphatag}
+%global release 1
+%global prerelease develop
 
 Name:       foreman-installer
 Epoch:      1
 Version:    1.20.0
-Release:    0.develop%{?dotalphatag}%{?dist}
+Release:    %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?dist}
 Summary:    Puppet-based installer for The Foreman
 Group:      Applications/System
 License:    GPLv3+ and ASL 2.0
 URL:        https://theforeman.org
-Source0:    https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?dashalphatag}.tar.bz2
+Source0:    https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?prerelease:-}%{?prerelease}.tar.bz2
 
 BuildArch:  noarch
 
@@ -33,7 +31,7 @@ BuildRequires: puppet-agent-puppet-strings < 2
 Complete installer for The Foreman life-cycle management system based on Puppet.
 
 %prep
-%setup -q -n %{name}-%{version}%{?dashalphatag}
+%setup -q -n %{name}-%{version}%{?prerelease:-}%{?prerelease}
 
 %build
 #replace shebangs for SCL
@@ -75,6 +73,9 @@ foreman-installer --scenario foreman --migrations-only > /dev/null
 %{_mandir}/man8
 
 %changelog
+* Wed Jul 25 2018 Eric D. Helms <ericdhelms@gmail.com> - 1:1.20.0-0.1.develop
+- Add prerelease macro
+
 * Tue Jul 17 2018 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> - 1:1.20.0-0.develop
 - Bump version to 1.20-develop
 
--- a/packages/foreman/foreman-proxy/foreman-proxy.spec
+++ b/packages/foreman/foreman-proxy/foreman-proxy.spec
@@ -4,20 +4,18 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-# set and uncomment all three to set alpha tag
-#global alphatag RC1
-#global dotalphatag .%{alphatag}
-#global dashalphatag -%{alphatag}
+%global release 1
+%global prerelease develop
 
 Name:           foreman-proxy
 Version:        1.20.0
-Release:        0.develop%{?dotalphatag}%{?dist}
+Release:        %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?dist}
 Summary:        Restful Proxy for DNS, DHCP, TFTP, PuppetCA and Puppet
 
 Group:          Applications/System
 License:        GPLv3+
 URL:            https://theforeman.org/projects/smart-proxy
-Source0:        https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?dashalphatag}.tar.bz2
+Source0:        https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?prerelease:-}%{prerelease}.tar.bz2
 Source1:        %{name}.tmpfiles
 Source2:        logrotate.conf
 
@@ -54,7 +52,7 @@ Manages DNS, DHCP, TFTP and puppet settings though HTTP Restful API
 Mainly used by the foreman project (https://theforeman.org)
 
 %prep
-%setup -q -n %{name}-%{version}%{?dashalphatag}
+%setup -q -n %{name}-%{version}%{?prerelease:-}%{?prerelease}
 
 %build
 #build man pages
@@ -191,6 +189,9 @@ fi
 
 
 %changelog
+* Wed Jul 25 2018 Eric D. Helms <ericdhelms@gmail.com> - 1.20.0-0.1.develop
+- Add prerelease macro
+
 * Tue Jul 17 2018 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> - 1.20.0-0.develop
 - Bump version to 1.20-develop
 
--- a/packages/foreman/foreman-selinux/foreman-selinux.spec
+++ b/packages/foreman/foreman-selinux/foreman-selinux.spec
@@ -30,20 +30,18 @@
 
 %define moduletype apps
 
-# set and uncomment all three to set alpha tag
-#global alphatag RC1
-#global dotalphatag .%{alphatag}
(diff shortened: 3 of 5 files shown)


packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -1,17 +1,15 @@
-# set and uncomment all three to set alpha tag
-#global alphatag RC1
-#global dotalphatag .%{alphatag}
-#global dashalphatag -%{alphatag}
+%global prerelease .develop
+%global release 2
```
ekohl: We're still on 1 here

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -1,17 +1,15 @@
-# set and uncomment all three to set alpha tag
-#global alphatag RC1
-#global dotalphatag .%{alphatag}
-#global dashalphatag -%{alphatag}
+%global prerelease .develop
```
ekohl: This isn't entirely going to work I think. I think we need to drop the dot here and explicitly include it in Release. Otherwise the download link will end up looking for `1.19.0-.RC1.tar.bz2` That's why we had a `dotalphatag` and a `dashalphatag`.

packages/foreman/foreman-installer/foreman-installer.spec
ehelms: Since I am making a change to the spec, I feel like the release should get bumped to reflect that.

ekohl: Technically the old release was on 0 so 1 is still an increment :)

ehelms:
[test rpm]

ehelms:
[test rpm]

packages/foreman/foreman/foreman.spec
```diff
@@ -1014,8 +1012,8 @@ install -Dpm0644 %{SOURCE6} %{buildroot}%{_sysconfdir}/yum.repos.d/%{name}.repo
 install -Dpm0644 %{SOURCE7} %{buildroot}%{_sysconfdir}/yum.repos.d/%{name}-plugins.repo
 sed "s/\$DIST/$(echo %{?dist} | cut -d. -f2)/g" -i %{buildroot}%{_sysconfdir}/yum.repos.d/%{name}*.repo
 if [[ '%{release}' != *"develop"* ]];then
```
ekohl: Should we change this to just check `[[ %{prerelease} == "develop" ]]`?

ekohl:
Doesn't look like it fully understands the `-develop` in the filename: http://koji.katello.org/kojifiles/work/tasks/5202/115202/build.log

```
+ cd foreman-proxy-develop.develop
/var/tmp/rpm-tmp.ubQcyV: line 35: cd: foreman-proxy-develop.develop: No such file or directory
```

rel-eng/custom/custom.py
```diff
@@ -83,17 +83,10 @@ def filter_archives(path):
             new_line = "Source%s: %s\n" % (i, base_name)
             replacements.append((source_regex, new_line))
 
-        # Replace version in spec:
-        version_regex = re.compile("^(version:\s*)(.+)$", re.IGNORECASE)
-        self.version = self._get_version()
-        print("Building version: %s" % self.version)
-        replacements.append((version_regex, "Version: %s\n" % self.version))
-        self.replace_in_spec(replacements)
```
ehelms: This was trying to modify the version in place but I couldn't figure out why. It broke RPM building by incorrectly setting version. As far as I can tell, our specs are smart enough to not need this.

rel-eng/custom/custom.py
```diff
@@ -83,17 +83,10 @@ def filter_archives(path):
...
-
         rel_date = datetime.utcnow().strftime("%Y%m%d%H%M")
         self.release = rel_date + gitrev
         print("Building release: %s" % self.release)
-        run_command("sed -i '/^Release:/ s/%%/.%s%%/' %s" % (self.release, self.spec_file))
+        run_command("sed -i '/^Release:/ s/%%/0.%s.%%/' %s" % (self.release, self.spec_file))
```
ehelms: I updated this to put a leading `0.` instead of `.` and a trailing `.` so that it separates things from the intended release a bit more clearly. The nightly release string is a bit verbose now.

rel-eng/custom/custom.py
ekohl: I actually think this could be a good thing to have. If `VERSION` is bumped in git but we forget it here then they won't match. We probably need to drop `-develop` though.

packages/foreman/foreman-selinux/foreman-selinux.spec
```diff
@@ -60,7 +58,7 @@ Requires(postun):   /usr/sbin/semodule, /sbin/restorecon
 SELinux policy module for Foreman
 
 %prep
-%setup -q -n %{name}-%{version}%{?dashalphatag}
+%setup -q -n %{name}-%{version}%{?prerelease:.}%{?prerelease}
```
ekohl: The build is failing because this needs to be a dash. Can't the setup macro figure this out itself from source0?

packages/foreman/foreman/foreman.spec
```diff
@@ -932,7 +930,7 @@ Foreman is based on Ruby on Rails, and this package bundles Rails and all
 plugins required for Foreman to work.
 
 %prep
-%setup -q -n %{name}-%{version}%{?dashalphatag}
+%setup -q -n %{name}-%{version}%{?prerelease:.}%{?prerelease}
```
ekohl: Also needs to be a dash

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -1,17 +1,15 @@
-# set and uncomment all three to set alpha tag
-#global alphatag RC1
-#global dotalphatag .%{alphatag}
-#global dashalphatag -%{alphatag}
+%global prerelease develop
+%global release 1
```
ekohl: One thing to consider here: Previously we had `1.20.0-0.develop` followed by then `1.20.0-0.1RC1`. Because `d < 1` it allowed upgrades. Since we're now including releases we could end up with `1.20.0-0.5.develop` which means we'll also have `1.20.0-0.5RC1`. This will work but because release is more important than prerelease in ordering I'd like it to come first in the file (so swap these 2 lines). Otherwise you might be tempted to reset the release to 1 when bumping to RC1. Small and subtle, but I have made this mistake in the past. Now we have automation that should catch it, but it's better to avoid it in the first place.

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -33,7 +31,7 @@ BuildRequires: puppet-agent-puppet-strings < 2
 Complete installer for The Foreman life-cycle management system based on Puppet.
 
 %prep
-%setup -q -n %{name}-%{version}%{?dashalphatag}
+%setup -q -n %{name}-%{version}%{?prerelease:.}%{?prerelease}
```
ekohl: This also needs to be a dash instead of a dot.

packages/foreman/foreman-installer/foreman-installer.spec
ehelms: Sorry, I don't understand what the problem is that we may encounter by this change. When we do a release, we'll have 1.20.0-0.1.RC1 which is what we had before and want as I understand things.

packages/foreman/foreman/foreman.spec
ehelms: Ehh, this feels fine still to me.

packages/foreman/foreman-installer/foreman-installer.spec
ekohl: It was a long way of saying that release is more significant than preprelease in determining the order. That's why I'd like the release macro to be defined before prerelease. Currently it's the other way around. This is no difference on a technical level but may help prevent human errors in the future.

packages/foreman/foreman/foreman.spec
ekohl: Fair enough.

rel-eng/custom/custom.py
ekohl: I wonder if we should change this code to fail if they don't match. Now it will fail during the build but the earlier you fail, the better.

packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -54,7 +52,10 @@ Manages DNS, DHCP, TFTP and puppet settings though HTTP Restful API
 Mainly used by the foreman project (https://theforeman.org)
 
 %prep
-%setup -q -n %{name}-%{version}%{?dashalphatag}
+echo %{version}
```
ekohl: I'm sure you'll remove this debug code in the final version :)

packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -4,20 +4,18 @@
...
 Version:        1.20.0
-Release:        0.develop%{?dotalphatag}%{?dist}
+Release:        %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?dist}
 Summary:        Restful Proxy for DNS, DHCP, TFTP, PuppetCA and Puppet
 
 Group:          Applications/System
 License:        GPLv3+
 URL:            https://theforeman.org/projects/smart-proxy
-Source0:        https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?dashalphatag}.tar.bz2
+Source0:        https://downloads.theforeman.org/%{name}/%{name}-%{version}-%{prerelease}.tar.bz2
```
ekohl: This could also use the prerelease macro instead of hardcoding. Probably left over debug code :)

rel-eng/custom/custom.py
ehelms: That feels like weird coupling to me to have this custom source strategy doing a validation check when we have builds that will do it or can put it into our pipelines.

packages/foreman/foreman-installer/foreman-installer.spec
ehelms: Heh, I can do that if you want. A bit ironic that prerelease will come after release ;)

ehelms:
Updated

PR #2802, opened 2018-07-25
Remove python-gofer-qpid traces

jturel:
(cherry picked from commit de74c5660b455b071ea6e444a3b4cb997448f20b)


jturel:
@ehelms here is the pick :)

PR #2808, opened 2018-07-27
Fixes #24422 - k-c-h Check hostname is not current hostname

johnpmitsch:
For katello-change-hostname, often times users will use
hostnamectl to change the hostname before running k-c-h.
This causes the script to error out and leaves things in
a bad state. This commit adds a check that the hostname
specified is different than the current system's hostname.

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -77,6 +77,11 @@ module KatelloUtilities
         self.fail_with_message("Please specify a hostname.", @opt_parser)
       end
 
+      if @old_hostname == @new_hostname
+        self.fail_with_message("The hostname specified must be different from the current hostname. If you have changed the hostname" \
+                              " with another utility, please change it back to the original hostname before running this tool.")
+      end
+
       STDOUT.puts "\nChecking hostname validity"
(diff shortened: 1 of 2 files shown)


johnpmitsch:
@ekohl updated, let me know if I bumped the release incorrectly with the new formatting

johnpmitsch:
To review:
- Spin up a "newish" katello production installation (I would say 3.7+, nightly would be best).
- clone foreman-packaging and checkout this PR
- snapshots are highly encouraged at this point since you are dealing with `katello-change-hostname`, which heavily modifies a system. :wink: 
- run `./packages/katello/katello/katello-change-hostname -u admin -p changeme $(hostname -f)`, this should fail with a warning since its trying to change the hostname to current hostname of the box.
- run `./packages/katello/katello/katello-change-hostname -u admin -p changeme <NEW_HOSTNAME>` using a new hostname in place of <NEW_HOSTNAME> and ensure the script completes successfully.
- check anything else you deem important, such as changing the hostname first with `hostnamectl` and trying to use katello-change-hostname to change to that hostname (this is the behavior we are trying to prevent).

PR #2809, opened 2018-07-27
Update katello-repos gpg key

jturel:


jturel:
@jlsherrill tested the package built from this and 3.7 packages installed just fine :+1:

PR #2819, opened 2018-07-31
[1.19] Fix GPG handling of repositories

ekohl:
* Installs the GPG key for the rails repository
* Ensure we can't install with the 1.18 release-scl
* Mark plugin repos as not GPG-signed

Note I intend to use the koji builds from this to actually test it
before pushing.


ekohl:
Going to close this. The proper fix is https://github.com/theforeman/foreman-packaging/pull/2823 and backporting that.

PR #2821, opened 2018-08-01
Add Foreman client repositories

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -298,6 +298,45 @@ foreman_packages:
     tfm: {}
     tfm-rubygem-rest-client: {}
 
+foreman_client_packages:
+  vars:
+    package_base_dir: 'packages/client/'
+    releasers:
+    - koji-foreman-client
+    repoclosure_config: repoclosure/yum.conf
+    repoclosure_lookaside_repos:
+      - el7-base
+      - el7-updates
+      - el7-epel
+      - el7-extras
+      - el7-puppet-5
+      - el7-foreman-client-nightly
+      - el6-base
+      - el6-updates
+      - el6-epel
+      - el6-extras
+      - el6-puppet-5
+      - el6-subscription-manager
+      - el6-qpid
+      - el6-foreman-client-nightly
+      - el5-base
+      - el5-updates
+      - el5-epel
+      - el5-extras
+      - el5-puppet-5
+      - el5-subscription-manager
+      - el5-qpid
+      - el5-foreman-client-nightly
+      - f28-base
+      - f28-updates
+      - f28-puppet-5
+      - f28-foreman-client-nightly
+      - f27-base
+      - f27-updates
+      - f27-puppet-5
+      - f27-foreman-client-nightly
+  hosts: {}
+
 katello_packages:
   vars:
     package_base_dir: 'packages/katello/'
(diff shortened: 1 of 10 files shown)


ekohl:
I assume we'll also be naming these `1.20` rather than `3.9` after branching since we're using the foreman name.

ehelms:
That's the idea @ekohl -- Katello would be reduced to it's server repository only.

ehelms:
Needs https://github.com/theforeman/tool_belt/pull/101

jturel:
Is the goal here to preserve the katello client tooling package names but ship them in a broader, more generic repo for any and all foreman client packages?

ehelms:
@jturel Yup!

ekohl:
There's now a merge conflict but otherwise I think this looks good. We will need to ensure we can manage the repo so we need a `.repo` file in a `-release` RPM and support in forklift.

ehelms:
This should help with these cases -- https://github.com/theforeman/foreman-infra/pull/770

ehelms:
[test rpm]

package_manifest.yaml
```diff
@@ -5,6 +5,7 @@ packages:
     plugins_packages: {}
     katello_packages: {}
     katello_client_packages: {}
```
zjhuntin: Is this still needed if it's being eliminated later?

package_manifest.yaml
```diff
@@ -358,46 +359,45 @@ katello_packages:
...
-      - f28-katello-client-nightly
+      - f28-foreman-client-nightly
       - f27-base
       - f27-updates
       - f27-puppet-5
-      - f27-katello-client-nightly
+      - f27-foreman-client-nightly
   hosts:
     katello-host-tools: {}
-    katello-repos: {}
```
zjhuntin: Why does this need to be removed?

ekohl: The section is being renamed to foreman client repos. I don't think it makes sense for the client tools to ship the katello repos (which should be named `katello-release` if it follows the naming convention), especially if katello-repos is now server-only.

ehelms:
I updated this a bit to put back the Katello client aspect. I think a safer approach is to build out the Foreman client repository and keep existing Katello bits in tact until the new repositories are fully online.

ehelms:
@ekohl @zjhuntin @komidore64 I re-factored this to just add in the structure for the new tags but not mess with the old existing Katello tags. I think it would be safer to make that change once the entire client ecosystem is setup.

rel-eng/tito.props
```diff
@@ -401,6 +401,29 @@ whitelist = ansiblerole-insights-client
   rubygem-smart_proxy_salt
   rubygem-smart_proxy_spacewalk
 
+[foreman-client-nightly-rhel7]
+disttag = .el7
+scl = tfm
+whitelist = katello-host-tools
```
ekohl: Should we keep the whitelists empty as well if we're not adding any packages yet?

rel-eng/tito.props
```diff
@@ -401,6 +401,29 @@ whitelist = ansiblerole-insights-client
   rubygem-smart_proxy_salt
   rubygem-smart_proxy_spacewalk
 
+[foreman-client-nightly-rhel7]
+disttag = .el7
+scl = tfm
```
ekohl: This is the only one with SCL. Should we also have a nonscl version?

ehelms:
Cleaned up tito.props for the tags to start empty.

PR #2822, opened 2018-08-01
Split out foreman-release RPM

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---


ehelms:
I think this would trim down the Foreman spec and make it easier to add a client set of repositories by having a specific foreman-release package for client repos as well.

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,60 @@
+Name:     foreman-release
+Version:  1.19.0
+Release:  1%{?dist}
+
+Summary:  Foreman repositories meta-package
+Group:    Applications/System
+License:  GPLv3+
+URL:      http://theforeman.org
```
ekohl: Mind changing this to https?

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,60 @@
...
+install -Dpm0644 %{SOURCE8} %{buildroot}%{_sysconfdir}/pki/rpm-gpg/RPM-GPG-KEY-foreman
+
+cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-dist << EOF
+# Version to use like a dist tag
+%%%{name}dist .fm$(echo %{version} | awk -F. '{print $1 "_" $2}')
+EOF
+
+
+%files
+%config %{_sysconfdir}/yum.repos.d/%{name}-plugins.repo
```
ekohl: I think this shouldn't be `-plugins` here

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,60 @@
...
+/etc/pki/rpm-gpg/*
+%{_sysconfdir}/rpm/macros.%{name}-dist
+
+%files -n foreman-release-plugins
+%config %{_sysconfdir}/yum.repos.d/%{name}-plugins.repo
+/etc/pki/rpm-gpg/*
```
ekohl: Does it work when multiple packages contain the same file?

packages/foreman/foreman/foreman.spec
```diff
@@ -15,7 +15,7 @@
 
 Name:   foreman
 Version: 1.20.0
-Release: 0.develop%{?dotalphatag}%{?dist}
+Release: 0.develop%{?dotalphatag}%{?dist}.1
```
ekohl: I think the .1 after the dist tag is pretty ugly.

packages/foreman/foreman/foreman.spec
```diff
@@ -1263,6 +1239,9 @@ exit 0
 %systemd_postun_with_restart %{name}.service
 
 %changelog
+* Wed Aug 01 2018 Eric D. Helms <ericdhelms@gmail.com> - 1.20.0-0.develop.1
+- rebuilt
```
ekohl: I'm all for changelogs but this isn't correct :)

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,60 @@
...
+BuildArch: noarch
+
+BuildRequires: sed
+
+%description
+Foreman repository contains open source and other distributable software for
+distributions in RPM format. This package contains the repository configuration
+for Yum.
+
+%package -n foreman-release-plugins
```
ekohl: I think you can use `%package plugins` to automatically name it `foreman-release-plugins`

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,60 @@
...
+Summary:   Definition of yum repositories for Foreman plugins
+Group:     Applications/System
+
+%description -n foreman-release-plugins
+Defines yum repositories for Foreman plugins.
+
+%install
+install -Dpm0644 %{SOURCE6} %{buildroot}%{_sysconfdir}/yum.repos.d/%{name}.repo
```
ekohl: The `SOURCE#` macros need to be updated to refer to the correct numbers.

ekohl:
> To stay compatible I'm leaning to having foreman-release-core, foreman-release-plugins and let foreman-release require them both. That would result in the same user experience and means we don't have to update docs.

I just realized this might not work. Currently we recommend users to download the `foreman-release` RPM and install it. Then you have the core repo and you can install `foreman-release-scl` Perhaps we need to modify that to `foreman-release-all` which is a meta package depending on all subpackages.

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,60 @@
+Name:     foreman-release
+Version:  1.19.0
+Release:  1%{?dist}
+
+Summary:  Foreman repositories meta-package
+Group:    Applications/System
+License:  GPLv3+
+URL:      http://theforeman.org
+Source0:  %{name}.repo
```
ekohl: You can't use `%{name}` macro anymore since that's now `foreman-release` instead of `foreman`.

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,53 @@
...
+Release:  1%{?dist}
+
+Summary:  Foreman repositories meta-package
+Group:    Applications/System
+License:  GPLv3+
+URL:      https://theforeman.org
+Source0:  foreman.repo
+Source1:  foreman-plugins.repo
+Source2:  foreman.gpg
+Source3:  tfm-ror51.repo
```
ekohl: In other places I've named this `foreman-rails.repo` now which means less changes when moving to ror52.

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,53 @@
+Name:     foreman-release
+Version:  1.19.0
+Release:  1%{?dist}
```
ekohl: This needs a develop to differentiate between nightly "1.19" and actual 1.19 (which has GPG checking).

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,53 @@
+Name:     foreman-release
+Version:  1.19.0
```
ekohl: This should be 1.20.0 since that's what nightlies are now.

packages/foreman/foreman-release/tfm-ror51.repo
```diff
@@ -1,5 +1,5 @@
-[foreman-rails]
-name=Foreman Rails nightly
+[tfm-ror51]
```
ekohl: I prefer foreman-rails

ehelms:
@ekohl Can you expand on what the "more work" when releasing that this leads to is?

ekohl:
On branching/releasing it means another package that we need to bump.

Another thing is that we have a dependency in `foreman-plugin`:
https://github.com/theforeman/foreman-packaging/blob/b7e78ed392c754e8276d6bd649722857d9368ba3/packages/foreman/foreman/foreman.spec#L823
This means we must keep them in sync. If we split these into two separate packages it means keeping track of versioning. Everything that requires keeping track of things means more work and easy mistakes. By being part of the main foreman package we get these things for free.

ekohl:
@ehelms and I had some offline discussions. A convincing argument was that we need to build a client repo but for more distros than just EL7 x86-64. Now it'll be a matter of avoid the problems I mentioned.

ehelms:
I have updated this in light of comments and the fact that we will need to be able to build this RPM for varying client repository distributions. The original foreman-release served two purposes:

 * Define and configure the yum repositories
 * Define a foremandist macro for plugins

The foreman-release RPM split out here focuses only on #1 now. The second purpose, foremandist macro, has been defined in a `foreman-plugin-build` RPM that can be added to buildroots much the same as the foreman-release used to be. This splits the duties of foreman-release. Alternative RPM name could be `foreman-plugin-release` instead of `-build`. I originally went with build given this is to be used during the build process. The `-release` nomenclature follows the Red Hat/Fedora current practices of defining dist macros in a package with `-release`.

ehelms:
@ekohl Take note of https://github.com/theforeman/foreman-packaging/pull/2822#issuecomment-414306245 since it posted after your comment so you might have missed it

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+install -Dpm0644 %{SOURCE2} %{buildroot}%{_sysconfdir}/pki/rpm-gpg/RPM-GPG-KEY-foreman
+install -Dpm0644 %{SOURCE1} %{buildroot}%{_sysconfdir}/pki/rpm-gpg/RPM-GPG-KEY-foreman-rails
+
+%files
+%config %{_sysconfdir}/yum.repos.d/foreman.repo
+%config %{_sysconfdir}/yum.repos.d/foreman-plugins.repo
+%config %{_sysconfdir}/yum.repos.d/foreman-rails.repo
+/etc/pki/rpm-gpg/*
```
ekohl: Should this be `%{_sysconfdir}`?

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,51 @@
...
+
+%install
+install -Dpm0644 %{SOURCE0} %{buildroot}%{_sysconfdir}/yum.repos.d/foreman.repo
+install -Dpm0644 %{SOURCE1} %{buildroot}%{_sysconfdir}/yum.repos.d/foreman-plugins.repo
+install -Dpm0644 %{SOURCE3} %{buildroot}%{_sysconfdir}/yum.repos.d/foreman-rails.repo
+
+sed "s/\$DIST/$(echo %{?dist} | cut -d. -f2)/g" -i %{buildroot}%{_sysconfdir}/yum.repos.d/*.repo
+
+if [[ '%{release}' != *"develop"* ]];then
+	VERSION="%{version}"
```
ekohl: Tabs vs spaces

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,51 @@
...
+
+Summary:  Foreman repositories meta-package
+Group:    Applications/System
+License:  GPLv3+
+URL:      https://theforeman.org
+Source0:  foreman.repo
+Source1:  foreman-plugins.repo
+Source2:  foreman.gpg
+Source3:  foreman-rails.repo
+Source4:  copr.gpg
```
ekohl: Should we name this `foreman-rails.gpg`?

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,51 @@
+%global release 1
+%global prerelease develop
+
+Name:     foreman-release
+Version:  1.20.0
+Release: %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?dist}
```
ekohl: Could you align this with the other columns?

packages/foreman/foreman/foreman.spec
```diff
@@ -819,8 +799,7 @@ Meta package to install asset pipeline support.
 %package plugin
 Summary: Foreman plugin support
 Group: Development/Libraries
-Requires: %{name} = %{version}-%{release}
-Requires: %{name}-release = %{version}-%{release}
+Requires: %{name}-plugin-build
```
ekohl: This should also match the version

packages/foreman/foreman/foreman.spec
```diff
@@ -830,6 +809,15 @@ Meta package with support for plugins.
...
 %{_datadir}/%{name}/schema_plugin.rb
 
+%package plugin-build
+Summary: Foreman plugin RPM build support
+Group: Development/Libraries
+
+%description plugin-build
+Meta package with support for plugin RPM building.
+
+%files plugin
```
ekohl: Not only plugins need this. I'd suggest a `foreman-build` if you want to split this off from `foreman-release`.

packages/foreman/foreman/foreman.spec
```diff
@@ -830,6 +809,15 @@ Meta package with support for plugins.
 %{_sysconfdir}/rpm/macros.%{name}-plugin
 %{_datadir}/%{name}/schema_plugin.rb
 
+%package plugin-build
```
ekohl: We already have `foreman-plugin` package that has all the build support for a Foreman plugin.

packages/foreman/foreman/foreman.spec
```diff
@@ -1085,14 +1059,14 @@ cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name} << EOF
...
-cp -pr db/schema.rb %{buildroot}%{_datadir}/%{name}/schema_plugin.rb
-
-cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-dist << EOF
+cat > %{buildroot}%{_sysconfdir}/rpm/macros.foreman-dist << EOF
 # Version to use like a dist tag
-%%%{name}dist .fm$(echo %{version} | awk -F. '{print $1 "_" $2}')
+foremandist .fm$(echo %{version} | awk -F. '{print $1 "_" $2}')
 EOF
 
+# Keep a copy of the schema for quick initialisation of plugin builds
```
ekohl: Any reason you're swapping this around?

packages/foreman/foreman/foreman.spec
```diff
@@ -1085,14 +1059,14 @@ cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name} << EOF
 %%%{name}_restart      (/bin/systemctl try-restart %{name}.service) >/dev/null 2>&1
 EOF
 
-# Keep a copy of the schema for quick initialisation of plugin builds
-cp -pr db/schema.rb %{buildroot}%{_datadir}/%{name}/schema_plugin.rb
-
-cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-dist << EOF
+cat > %{buildroot}%{_sysconfdir}/rpm/macros.foreman-dist << EOF
```
ekohl: What's the benefit of this change?

packages/foreman/foreman/foreman.spec
ehelms: We do, but given this is a special macro that needs to live in the buildroot, I'm worried about a chicken'n'egg situation that could be simplified by just having this simple package with the macro.

ehelms:
Updated based on comments and switched to foreman-build.

packages/foreman/foreman/foreman.spec
```diff
@@ -830,6 +809,15 @@ Meta package with support for plugins.
...
 
+%package build
+Summary: Foreman package RPM support
+Group: Development/Libraries
+
+%description build
+Meta package with support for building RPMs in the Foreman release cycle.
+
+%files build
+%{_sysconfdir}/rpm/macros.foremandist
```
ekohl: It used to be `%{_sysconfdir}/rpm/macros.%{name}-dist` in `-release` Should we keep it like that?

packages/foreman/foreman/foreman.spec
```diff
@@ -819,8 +799,7 @@ Meta package to install asset pipeline support.
 %package plugin
 Summary: Foreman plugin support
 Group: Development/Libraries
-Requires: %{name} = %{version}-%{release}
```
ekohl: The plugin package still needs `foreman` installed so they can compile things like assets.

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -0,0 +1,51 @@
...
+  VERSION="%{version}"
+  sed "s/nightly/${VERSION%.*}/g" -i %{buildroot}%{_sysconfdir}/yum.repos.d/*.repo
+  sed "s/gpgcheck=0/gpgcheck=1/g" -i %{buildroot}%{_sysconfdir}/yum.repos.d/foreman.repo
+fi
+
+install -Dpm0644 %{SOURCE2} %{buildroot}%{_sysconfdir}/pki/rpm-gpg/RPM-GPG-KEY-foreman
+install -Dpm0644 %{SOURCE1} %{buildroot}%{_sysconfdir}/pki/rpm-gpg/RPM-GPG-KEY-foreman-rails
+
+%files
+%config %{_sysconfdir}/yum.repos.d/foreman.repo
```
ekohl: Should we simplify this to *.repo like we do the GPG keys?

ehelms:
Comps updated @ekohl

ehelms:
Going to merge and then fire off some builds, and nightly runs to test out the full range of functionality and ensure things are kosher.

PR #2823, opened 2018-08-01
Move foreman-rails repo to foreman-release

ekohl:
This moves the foreman-rails repo to foreman-release. It's tied to the foreman release cycle and is versioned in the same way. This avoids a lot of the problems with ensuring we have the proper version installed and obsolete the older versions.

We also only depend on centos-release-scl-rh. We used to require SCLo repositories for Rails 4.2 but now we ship our own repository for this.

This is also intended to be backported to 1.17, 1.18 and 1.19. That's why I'd prefer it's merged before https://github.com/theforeman/foreman-packaging/pull/2822 because that's harder to backport.


ekohl:
Will need https://github.com/theforeman/foreman-packaging/pull/2799 to build. After that's merged I'll rebase.

packages/foreman/foreman-release-scl/foreman-release-scl.spec
```diff
@@ -1,17 +1,15 @@
...
 License:  GPLv3+
-URL:      http://theforeman.org
-Source0:  tfm-ror51.repo
-Source1:  copr.gpg
+URL:      https://theforeman.org
 
 BuildArch: noarch
 
-Requires: centos-release-scl
+Requires: centos-release-scl-rh
```
ekohl: I'm leaning to doing this in a `7-2` so we can cherry pick it to 1.17 and 1.18.

packages/foreman/foreman/foreman.spec
```diff
@@ -33,9 +33,12 @@ Source8: %{name}.gpg
 Source9: message_encryptor_extensions.rb
 Source10: %{executor_service_name}.sysconfig
 Source11: %{executor_service_name}.service
+Source12: %{name}-rails.repo
```
ekohl: Should this be a separate file or should we include it in `foreman.repo`?

ekohl:
I'm not sure splitting out foreman-release actually makes it better and this is easier to cherry pick to older releases. It's green now so ready for a final review.

ehelms:
Given the impending nature of 1.19, is it worth backporting to 1.17?

ekohl:
We did talk about a 1.17.2 release and as it stands now that will introduce issues when we release packages so I do think it's worth it.

PR #2824, opened 2018-08-01
[1.19] Move foreman-rails repo to foreman-release

ekohl:
This moves the foreman-rails repo to foreman-release. It's tied to the foreman release cycle and is versioned in the same way. This avoids a lot of the problems with ensuring we have the proper version installed and obsolete the older versions.

We also only depend on centos-release-scl-rh. We used to require SCLo repositories for Rails 4.2 but now we ship our own repository for this.

This is a cherry pick of https://github.com/theforeman/foreman-packaging/pull/2823 for 1.19.


ekohl:
I'd rather not merge this before the develop one is fixed.

PR #2825, opened 2018-08-01
source_files arg to fetchbuilder

komidore64:
this allows one to point to source files that are already built

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/rel-eng/custom/custom.py
+++ b/rel-eng/custom/custom.py
@@ -199,6 +199,10 @@ class ForemanSourceStrategy(SourceStrategy):
 
     def _fetch_prebuilt_local(self):
         source_files = [os.path.expanduser(source_file) for source_file in self.builder.args['source_files']]
+        if self.builder.args['git_hash']:
+            gitsha = self.builder.args['git_hash'][0]
+        else:
+            raise Exception("Specify '--arg git_hash=...' when using '--arg source_files=...[,...]")
         dest_dir = os.path.join(self.builder.rpmbuild_sourcedir, 'archive')
         if not os.path.exists(dest_dir):
             os.mkdir(dest_dir)
@@ -207,11 +211,7 @@ class ForemanSourceStrategy(SourceStrategy):
             debug("Copying %s to %s" % (src, os.path.join(dest_dir, os.path.basename(src))))
             shutil.copy(src, os.path.join(dest_dir, os.path.basename(src)))
 
-        gitsha = self.builder.args['git_hash'][0]
-        if gitsha:
-            return "git%s" % gitsha[0:7]
-        else:
-            return ""
+        return "git%s" % gitsha[0:7]
 
     def _get_version(self):
         """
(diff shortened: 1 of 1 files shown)


rel-eng/custom/custom.py
```diff
@@ -195,6 +197,18 @@ def _fetch_local(self):
...
 
+    def _fetch_prebuilt_local(self):
+        source_files = [os.path.expanduser(source_file) for source_file in self.builder.args['source_files']]
+        dest_dir = os.path.join(self.builder.rpmbuild_sourcedir, 'archive')
+        if not os.path.exists(dest_dir):
+            os.mkdir(dest_dir)
+
+        for src in source_files:
+            debug("Copying %s to %s" % (src, os.path.join(dest_dir, '')))
+            shutil.copy(src, os.path.join(dest_dir, ''))
```
ekohl: I'd prefer to be explicit and use `os.path.basename(srcfile)` instead of an empty string.

komidore64:
@ehelms @ekohl updated based on comment

rel-eng/custom/custom.py
```diff
@@ -195,6 +197,18 @@ def _fetch_local(self):
...
+        source_files = [os.path.expanduser(source_file) for source_file in self.builder.args['source_files']]
+        dest_dir = os.path.join(self.builder.rpmbuild_sourcedir, 'archive')
+        if not os.path.exists(dest_dir):
+            os.mkdir(dest_dir)
+
+        for src in source_files:
+            debug("Copying %s to %s" % (src, os.path.join(dest_dir, os.path.basename(src))))
+            shutil.copy(src, os.path.join(dest_dir, os.path.basename(src)))
+
+        return ""
```
ekohl: What does an empty string mean here? Do we need a return value?

komidore64: Other `_fetch*` methods return a gitrev string for source, but since we're just pointing to a file we don't have any kind of hash, so empty string. That gets concatenated into the release which results in just a timestamp.

komidore64:
don't merge this yet

komidore64:
okay, now it's ready for review again

rel-eng/custom/custom.py
```diff
@@ -195,6 +197,22 @@ def _fetch_local(self):
...
+        source_files = [os.path.expanduser(source_file) for source_file in self.builder.args['source_files']]
+        dest_dir = os.path.join(self.builder.rpmbuild_sourcedir, 'archive')
+        if not os.path.exists(dest_dir):
+            os.mkdir(dest_dir)
+
+        for src in source_files:
+            debug("Copying %s to %s" % (src, os.path.join(dest_dir, os.path.basename(src))))
+            shutil.copy(src, os.path.join(dest_dir, os.path.basename(src)))
+
+        gitsha = self.builder.args['git_hash'][0]
```
ekohl: This will raise an `IndexError` or a `KeyError` if it's not passed in. IMHO that's not a problem and actually a feature, but I'd do it at the start of the function so we reject builds without a git sha. We can also raise an exception with a bit more helpful error message.

ekohl:
Thanks!

PR #2827, opened 2018-08-02
Include redhat_access plugin static assets in the spec

dLobatog:
This plugin is not configured to compile the static assets, hence I'm
adding them to the spec.

Meant to fix https://bugzilla.redhat.com/show_bug.cgi?id=1610971

* [x] Nightly
* [x] 1.19
* [x] 1.18
* [ ] 1.17

---

Thanks for reviewing!


packages/katello/rubygem-redhat_access/rubygem-redhat_access.spec
```diff
@@ -86,6 +86,11 @@ cp -pa $RPM_BUILD_DIR/%{gem_name}-%{version}/config/config.yml.example %{buildro
 %foreman_bundlerd_file
 %foreman_precompile_plugin -s
 
+# This plugin is not able to compile static assets directly
+mkdir -p %{buildroot}/%{rubygem_redhat_access_dir}/public/assets
+cp -r  $RPM_BUILD_DIR/%{gem_name}-%{version}/vendor/assets/images/*  %{buildroot}/%{rubygem_redhat_access_dir}/public/assets
```
ekohl: The `rubygem_redhat_access_dir` macro is undefined. You can see that in the build error as well:

```
   /%{rubygem_redhat_access_dir}/public/assets/insights/Icon_RH_Document_Blueprint_RGB_Button.png
```

packages/katello/rubygem-redhat_access/rubygem-redhat_access.spec
```diff
@@ -3,19 +3,20 @@
 
 %global gem_name redhat_access
 %global plugin_name redhat_access
+%global rubygem_redhat_access_dir %{gem_dir}/gems/%{gem_name}-%{version}
```
ekohl: This is equivalent to `%gem_instdir` so I'd prefer to use that.

dLobatog:
Updated to *not* use the `global` and use %gem_instdir instead, thanks @ekohl

packages/katello/rubygem-redhat_access/rubygem-redhat_access.spec
```diff
@@ -86,6 +86,11 @@ cp -pa $RPM_BUILD_DIR/%{gem_name}-%{version}/config/config.yml.example %{buildro
 %foreman_bundlerd_file
 %foreman_precompile_plugin -s
 
+# This plugin is not able to compile static assets directly
+mkdir -p %{buildroot}/%{gem_instdir}/public/assets
+cp -r  $RPM_BUILD_DIR/%{gem_name}-%{version}/vendor/assets/images/*  %{buildroot}/%{gem_instdir}/public/assets
```
ekohl: Some things do refer to `static/images/Icon_RH_Software_Data_CMYK_Button.png`. I suspect the real reason is that it refers to the exact filenames while asset compilation renames them to hashes so you can use caching.  That leads me to think that any changes to those images will not properly propagate to users on upgrades since we send caching headers.

Another possible reason is that files are moved here from one directory to another which breaks relative paths.

I don't have time to really look into it, but that's just bugs waiting to happen. I'm OK with including this hack now, but would you mind creating an issue upstream that this should be fixed?

lphiri: There is a bit of history here :).
These assets are imported from the Insights app in the Customer Portal and requested by exact filename by the Insights UI Angularjs javascript files (also imported from the Portal app).

ekohl: @lphiri that explains the issue, but we still need to solve the caching. We set expire headers for access plus 1 year which means if any user retrieves `Icon_RH_Software_Data_CMYK_Button.png` then their browser caches that file for a year. This can greatly increase the performance for a user (especially on low bandwidth networks), but it also means you can't change those files on upgrade or the user can get weird inconsistent results. Because of that files are renamed to `<name>-<hash>.extension` and those filenames are used. So this hack works around the issue at first sight, but creates bugs in the long run.

lphiri: @ekohl thats correct - we do need to fix the caching issue long term - I was just adding info for background.

ares:
@ekohl I see the PR is acked, is there something that blocks next step?

PR #2830, opened 2018-08-02
Add gofer to katello-nightly-rhel7

ekohl:


ehelms:
Closing since we dropped katello client repos

PR #2831, opened 2018-08-02
[1.17] Move foreman-rails repo to foreman-release

ekohl:
This moves the foreman-rails repo to foreman-release. It's tied to the
foreman release cycle and is versioned in the same way. This avoids a
lot of the problems with ensuring we have the proper version installed
and obsolete the older versions.


ehelms:
When is the next 1.17 due out? Is it worth pulling this back given 1.19 is on the cusp?

ekohl:
@xprazak2 I recall we talked about doing 1.17.2 but not sure we agreed on a date. It is important to note that the current `foreman-release-scl` in this repo is already different (uses yum.theforeman.org) from the released RPM (uses COPR). This PR corrects the unreleased version to align with the other releases.

xprazak2:
We did not set a specific date, but I would like to start with 1.17.2 in the first half of this week.

PR #2832, opened 2018-08-02
[1.19] Correct foreman-rails repo definition

ekohl:

PR #2833, opened 2018-08-02
Correct foreman-rails repo definition

ekohl:
Equivalent of https://github.com/theforeman/foreman-packaging/pull/2832 for rpm/develop.


ehelms:
Should this get a release bump and changelog to capture the update?

ekohl:
I'm still not sure what to think about many of those small changes in nightly in changelogs. As a user I hardly ever look at RPM changelogs and in our case git log might be more efficient.

PR #2840, opened 2018-08-07
[1.18] Fix foreman spec

ekohl:
* Fix rails GPG key
* Fix NPM dependencies


ekohl:
This is still WIP, still made a mistake.

evgeni:
is this missing the GPG change?

ekohl:
It had an incorrect GPG change. This should be OK now. Waiting for the build to complete and restest it. Note that the GPG file was a binary and github doesn't show a diff.

PR #2857, opened 2018-08-14
Add fast_gettext back to foreman-plugins

lzap:
It was there, but it was removed some time ago. It is already present in
compose XML. I hope this is all I need, Jenkins will hopefully tell.

* [x] Nightly
* [x] 1.19

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -431,6 +431,7 @@ plugins_packages:
     rubygem-deface: {}
     rubygem-diffy: {}
     rubygem-docker-api: {}
+    rubygem-fast_gettext: {}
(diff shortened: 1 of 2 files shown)


lzap:
It's green but I don't see any package built on koji.

ekohl:
Why is this needed? Note you are now asking to create `tfm-rubygem-fast_gettext` (which is already in `foreman-nightly-rhel7`), not `rubygem-fast_gettext`. I think what you actually intended was https://github.com/theforeman/foreman-packaging/pull/2866.

lzap:
Oh sorry, thanks!

PR #2866, opened 2018-08-14
Add rubygem-fast_gettext to plugins nonscl

ekohl:
This package is used by foreman-discovery-image.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -180,7 +180,10 @@ foreman_packages:
     rubygem-extlib: {}
     rubygem-facter: {}
     rubygem-faraday: {}
-    rubygem-fast_gettext: {}
+    rubygem-fast_gettext:
+      releasers:
+        - koji-foreman
+        - koji-foreman-plugins
     rubygem-fog-aws: {}
     rubygem-fog-core: {}
     rubygem-fog-digitalocean: {}
--- a/rel-eng/tito.props
+++ b/rel-eng/tito.props
@@ -371,6 +371,7 @@ whitelist = ansiblerole-insights-client
   rubygem-chef-api
   rubygem-faraday
   rubygem-faraday_middleware
+  rubygem-fast_gettext
   rubygem-foreman_scap_client
   rubygem-infoblox
   rubygem-logify


package_manifest.yaml
```diff
@@ -180,7 +180,10 @@ foreman_packages:
     rubygem-extlib: {}
     rubygem-facter: {}
     rubygem-faraday: {}
-    rubygem-fast_gettext: {}
+    rubygem-fast_gettext:
+      releasers:
+        - koji-foreman
+        - koji-foreman-plugins
```
ehelms: This will work but likely fail if later on we update the package. The repoclosure check for one would be different given this is building an SCL and non-SCL version and targeting two different repositories. Do you want to just roll with it as is or tackle this issue now?

lzap:
Yes thank you this is what I want. I will followup with a metapackage called `foreman-discovery-image-tui` which will contain the three dependencies. In the future, we might move the code into package, but we are not there yet. Making a scratchbuild is a nightmare when a code change is needed in TUI and need to do this pretty often, also RPM is dropped on FDI because of space saving.

package_manifest.yaml
lzap: Guys please fix this for me so I can come up with FDI 3.5.0 build which is a *huge* overdue.

lzap:
For the record the metapackage PR is here: https://github.com/theforeman/foreman-packaging/pull/2868

lzap:
Can I ask for 1.19 as well if it was not already done? 

Unfortunately build failed:

```
Bringing machine 'el7' up with 'openstack' provider...
==> el7: Finding flavor for server...
==> el7: Finding image for server...
==> el7: Launching a server with the following settings...
==> el7:  -- Tenant          : 864112
==> el7:  -- Name            : fdi-builder-vm
==> el7:  -- Flavor          : 4GB Standard Instance
==> el7:  -- FlavorRef       : 5
==> el7:  -- Image           : OnMetal - CentOS 7
==> el7:  -- ImageRef        : 7c2f809e-6f28-42f2-a6f0-c0f3c99f0185
==> el7:  -- KeyPair         : vagrant-generated-6ooktsqd
Image cannot be built with provided flavor.
++ vagrant destroy el7
==> el7: The server hasn't been created yet. Run `vagrant up` first.
+ vagrant destroy el7
==> el7: The server hasn't been created yet. Run `vagrant up` first.
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

Unknown error. It's no good. http://ci.theforeman.org/job/foreman-discovery-image-publish/20/console

lzap:
We finally have a FDI RC build! Thanks. If possible can you also put the fast_gettext package into 1.19 repos so I can use release repo for final FDI build, that's be appreciated.

PR #2868, opened 2018-08-15
Adding FDI metapackage

lzap:

* [x] Nightly

Crossing fingers but since this is not rubygem I probably do it wrong.

---


lzap:
Ok thanks for help, trying again.

packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
```diff
@@ -0,0 +1,36 @@
...
+License: GPLv2+
+URL: https://github.com/theforeman/foreman-discovery-image
+
+Requires:	foreman-proxy
+Requires:	rubygem-smart_proxy_discovery_image
+
+%description
+Metapackage with dependencies for FDI
+
+%package tui
```
ekohl: Looks like this subpackage also needs a summary according to lint

lzap:
Seasoned packager, it is passing locally now. Amended.

packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
```diff
@@ -0,0 +1,37 @@
...
+Summary: Metapackage with dependencies for FDI TUI
+Requires: ruby(release)
+Requires:	rubygem(newt)
+Requires:	rubygem(fast_gettext)
```
ehelms: nitpick formatting

packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
ehelms: What does `tui` stand for?

ehelms: Text User Interface? Makes sense.

packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
```diff
@@ -0,0 +1,37 @@
+Name: foreman-discovery-image-service
```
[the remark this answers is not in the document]
lzap: Sure, so part of FDI there are couple or systemd services like register-host or download-extensions. Those are simple ruby scripts which do few things during startup. My idea is to have a common package for all these services, there is little value in breaking them up as they won't be used in any other context.

Then there's the TUI which is an extra ruby application that provides the user interface. That's the subpackage.

I am amending the formatting changes now.

PR #2880, opened 2018-08-17
[WIP] Rails 52

ehelms:
This heavily relies on the code base moving to Rails 5.2. I am staging in a PR the changes required thus far that I have not been able to push into the main rpm/develop already. Current status:

 * All Foreman core dependencies build
 * Foreman core package fails during RPM build as part of or during db:migrate

Note that in the `package_manifest.yml` I currently broke out more package groupings to make life easier (e.g. `obal srpm foreman_packages` followed by mockchain) when testing to avoid non-SCL packages. I would be curious thoughts on this general approach being done to main `rpm/develop` ahead of time. Useful groupings so far:

 * SCL (aka Foreman core packages)
 * Installer and smart proxy packages
 * Nodejs packages


packages/foreman/rubygem-activerecord-session_store/rubygem-activerecord-session_store.spec
```diff
@@ -74,6 +74,9 @@ cp -a .%{gem_dir}/* \
 %doc %{gem_instdir}/README.md
 
 %changelog
+* Thu Aug 16 2018 Eric D. Helms <ericdhelms@gmail.com> - 1.1.1-2
+- Rebuild for Rails 5.2
```
ekohl: Technically the rebuild is more for the Ruby 2.5 SCL than Rails

ehelms: Technically no :) The Rails 5.2 SCL change (from tfm-ror51 to tfm-ror52) is what necessitates the rebuilding. Whether Ruby changed or not would still result in rebuilding for Rails 5.2. And vice versa. So you could argue both, but primarily Rails is the driver.

ehelms:
Splitting this out a bit -- https://github.com/theforeman/foreman-packaging/pull/2904

PR #2885, opened 2018-08-21
Remove the dependency on SCLo

ekohl:
We used to depend on SCLo for sclo-ror42 but we've now moved to tfm-ror*
so we can depend only on sclo-rh which includes rh-ruby24 which we
depend on. This is not a hard dependency because RHEL users can use the
Ruby SCL from another repository.

### DIFF
--- a/comps/comps-foreman-rhel7.xml
+++ b/comps/comps-foreman-rhel7.xml
@@ -7,7 +7,6 @@
     <description>Packages for the Foreman</description>
     <uservisible>true</uservisible>
     <packagelist>
-      <packagereq type="default">centos-release-scl</packagereq>
       <packagereq type="default">centos-release-scl-rh</packagereq>
       <packagereq type="default">foreman</packagereq>
       <packagereq type="default">foreman-assets</packagereq>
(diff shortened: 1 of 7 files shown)


evgeni:
I'll leave it to the SCL master helms to ACK this, but the change looks good.

Did you run a repoclosure with the new config, just to be sure?

ekohl:
Good point. This doesn't sound right:

```
    package: tfm-rubygem-jquery_pwstrength_bootstrap_4-1.2.2-4.el7.noarch from el7-foreman-nightly
      unresolved deps:
         sclo-ror42-rubygem(railties) >= 0:3.1
```

evgeni:
looks like it was never rebuilt, but maybe, it also not needed anymore?

ekohl:
https://github.com/theforeman/foreman-packaging/pull/2887

evgeni:
@ekohl #2887 is merged, can you untag the RPM from nightly and re-run the repoclosure?

ACK if repoclosure finishes fine

ekohl:
I untagged yesterday. Now that tfm-4.0-4 also was promoted it passes:

```
ok: [foreman-repoclosure-el7] => 
  msg: |-
    Reading in repository metadata - please wait....
    Checking Dependencies
    Repos looked at: 8
       el7-base
       el7-epel
       el7-extras
       el7-foreman-nightly
       el7-foreman-rails-nightly
       el7-puppet-5
       el7-scl
       el7-updates
    Num Packages in Repos: 32894
```

PR #2889, opened 2018-08-23
Fix dynflow dependency for ansible proxy

xprazak2:

* [x] Nightly
* [x] 1.19
* [x] 1.18
* [ ] 1.17

---
The constraint seems [too restrictive](https://github.com/theforeman/smart_proxy_ansible/blob/2.0.3/smart_proxy_ansible.gemspec#L33) to me, this should fix
https://community.theforeman.org/t/problem-with-ansible-smart-proxy-plugin-installation/10850

@ekohl, I think this problem is in 1.19 as well, but it is your call.


xprazak2:
Ugh, wrong branch name. Let me try again.
https://github.com/theforeman/foreman-packaging/pull/2890

PR #2891, opened 2018-08-23
Move REX .ssh directory to /var/lib/foreman-proxy

ekohl:
For now this is untested. This PR is also here to create the packages.

This fix should also be applied to 1.18 and 1.19 so we can fix the bug in the installer.


plugins/smart_proxy_remote_execution_ssh/ruby-smart-proxy-remote-execution-ssh.postinst
```diff
@@ -0,0 +1,43 @@
...
+set -e
+
+USERNAME="foreman-proxy"
+GROUPNAME="foreman-proxy"
+FOREMAN_PROXY_HOME="/usr/share/foreman-proxy"
+VARDIR="/var/lib/foreman-proxy"
+
+case "$1" in
+  configure)
+    if [ -d $VARDIR ] ; then
```
evgeni: create the dir if it exists? did you want a `! -d $VARDIR` here?

plugins/smart_proxy_remote_execution_ssh/ruby-smart-proxy-remote-execution-ssh.postinst
```diff
@@ -0,0 +1,43 @@
...
+      mkdir -p $VARDIR
+      chown $USERNAME:$GROUPNAME $VARDIR
+    fi
+
+    if [ -d $FOREMAN_PROXY_HOME/.ssh ] && [ ! -L $FOREMAN_PROXY_HOME/.ssh ] ; then
+      mv $FOREMAN_PROXY_HOME/.ssh $VARDIR/ssh
```
evgeni: this will move the keys to `$VARDIR/ssh/.ssh` is that correct?

plugins/smart_proxy_remote_execution_ssh/ruby-smart-proxy-remote-execution-ssh.postinst
ekohl: Yes, that's what I meant.

plugins/smart_proxy_remote_execution_ssh/ruby-smart-proxy-remote-execution-ssh.postinst
ekohl: If the directory doesn't exist then it will rename `.ssh` to `ssh`. By default this is the case. If it did exist, then it would indeed end up in a subdirectory. Could be a feature not to overwrite files :)

evgeni: but the dir does exist, you created it.

ekohl: The parent dir exists. It might be cleaner to rely on the main foreman-proxy package to manage `/var/lib/foreman-proxy` but this allows us to ship the fix async. However, given we'll also need an installer update I'm debating if that's the best way forward.

```
$ mkdir /tmp/.ssh
$ mkdir /tmp/bla
$ mv /tmp/.ssh /tmp/bla/ssh
$ ls -l /tmp/bla
total 0
drwxrwxr-x. 2 ekohl ekohl 40 23 aug 16:57 ssh
```

evgeni: oh right, you only create up to VARDIR, sry.

ekohl:
Tested locally, this works as expected. The only thing I can't explain is that `/var/lib/foreman-proxy` already existed.

Another approach I'm considering is doing it as `preinst` and managing the symlink via `links`. This is equivalent to how we do it in RPM packaging. Would that work @evgeni?

evgeni:
@ekohl yeah, that would also work. not sure whats better, though.

mmoll:
Let's bring this in, as it should at least fix this specific issue :green_heart:

PR #2895, opened 2018-08-24
[WIP] fix nightly builds to all be the same and not require jenkins for archive storage

komidore64:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

this is related to and also needs theforeman/obal#125


rel-eng/releasers.conf
```diff
@@ -29,6 +38,10 @@ releaser = tito.release.KojiReleaser
 autobuild_tags = foreman-client-nightly-rhel5 foreman-client-nightly-rhel6 foreman-client-nightly-rhel7 foreman-client-nightly-fedora27 foreman-client-nightly-fedora28
 builder = tito.builder.FetchBuilder
 
+[koji-foreman-client-nightly]
```
ehelms: This releaser seems to be defined exactly the same as `koji-foreman-client`. What is the benefit of re-defining under a new name?

ehelms:
With this, probably a good time to update and capture:

https://github.com/theforeman/foreman-packaging/tree/rpm/develop#howto-build-multiple-packages
https://github.com/theforeman/foreman-packaging/tree/rpm/develop#howto-test-a-package

rel-eng/releasers.conf
komidore64: whooooooops!

ekohl:
What's the bigger design behind this? It only creates more inconsistencies since now we build hammer-cli and hammer-cli-foreman one way while we build hammer-cli-katello in another.

A big downside of not using Jenkins as storage is that we can end up with inconsistencies in artifacts. The result can be that the deb nightlies and RPM nightlies end up using different sources.

komidore64:
closing to redo in a more structured approach

PR #2902, opened 2018-08-27
Build hammer using the regular releaser

ekohl:
This allows us to properly release hammer while still building nightly packages.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -211,15 +211,9 @@ foreman_packages:
     rubygem-google-api-client: {}
     rubygem-gridster-rails: {}
     rubygem-hammer_cli:
-      releasers:
-        - "{{ nightly_releaser }}"
-      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
       nightly_package_tito_releaser_args:
         - "--arg jenkins_job=test_hammer_cli"
     rubygem-hammer_cli_foreman:
-      releasers:
-        - "{{ nightly_releaser }}"
-      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
       nightly_package_tito_releaser_args:
         - "--arg jenkins_job=test_hammer_cli_foreman"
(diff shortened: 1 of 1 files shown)


evgeni:
This mostly reverts @ehelms' #2882, so I'm kinda torn and probably do not understand the issue correctly?

ekohl:
The problem is that right now you can't release hammer with the gem version into nightly here. It always releases a nightly build. This means https://github.com/theforeman/foreman-packaging/pull/2897 and https://github.com/theforeman/foreman-packaging/pull/2898 try to build with a nightly gem. We used to replace the version with that from the nightly, but in https://github.com/theforeman/foreman-packaging/commit/06a47b8c2eacbbb1a4045972b591e41219b65c5b#diff-ea34d906605db96672e9d1060a6834b0L86 that was removed. Since develop has moved to 0.15, it can't find the 0.14 gem either.

ehelms:
I think this will actually break the nightly builds, because hammer-cli source will be version 0.15. So it's a question of which do we prefer. The ability to release into nightly tags packages that do not see the light of day, or to keep releasing nightly packages.

Whichever you want to support right now, I can live with. But this won't get us both as I understand our current code.

mbacovsky:
@ekohl, @ehelms I'm probably missing something important but why do we want to release specific hammer releases into nightly packages? Is that for historical reasons only?
Could we let nightly builds build hammer master snaps and release tagged hammer versions only to specific foreman branches? It seems to fit current workflow well, isn't it?

ekohl:
For now I merged the hammer 0.14 PRs into nightly, cherry picked and only released the RPMs into 1.19. That was mostly to fix it for the 1.19 release. I'll need to think about the right workflow a bit more when things have calmed down a little bit.

mbacovsky:
@ekohl, thanks!

komidore64:
this is related to my [nightly package+ proposal](https://community.theforeman.org/t/nightly-packaging-unification-proposal/10932)

PR #2910, opened 2018-08-30
Require current selinux policy in foreman-selinux

lzap:
https://community.theforeman.org/t/minimum-centos-version-requirement-for-1-18-is-7-5/10908

* [x] Nightly
* [x] 1.19
* [x] 1.18

I am marking also 1.18 but unsure if we are going to do the update so up
to you to decide. The error was reported by 1.18 user.

### DIFF
--- a/packages/foreman/foreman-selinux/foreman-selinux.spec
+++ b/packages/foreman/foreman-selinux/foreman-selinux.spec
@@ -18,19 +18,11 @@
 
 %define selinux_variants targeted
 %define selinux_modules foreman foreman-proxy
-
-%if 0%{?rhel} == 5
-# absolute minimum versions for RHEL 5
-%define selinux_policy_ver 3.11.1-81
-%else
-# absolute minimum versions for RHEL 6
-%define selinux_policy_ver 2.4.6-80
-%endif
-%define selinux_policycoreutils_ver 1.33.12-1
+%global selinux_policy_ver %(rpm --qf "%%{version}-%%{release}" -q selinux-policy)
 
 %define moduletype apps
 
-%global release 1
+%global release 2
 %global prerelease develop
 
 Name:           foreman-selinux
@@ -44,7 +36,7 @@ URL:            https://theforeman.org
 Source0:        https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?prerelease:-}%{?prerelease}.tar.bz2
 
 BuildRequires:  checkpolicy, selinux-policy-devel, hardlink
-BuildRequires:  policycoreutils >= %{selinux_policycoreutils_ver}
+BuildRequires:  policycoreutils
 BuildRequires:  /usr/bin/pod2man
 BuildArch:      noarch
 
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman-selinux/foreman-selinux.spec
```diff
@@ -18,14 +18,7 @@
...
 %define selinux_modules foreman foreman-proxy
-
-%if 0%{?rhel} == 5
-# absolute minimum versions for RHEL 5
-%define selinux_policy_ver 3.11.1-81
-%else
-# absolute minimum versions for RHEL 6
-%define selinux_policy_ver 2.4.6-80
-%endif
+%define selinux_policy_ver 3.13.1-192
```
ekohl: Pulp detects it automatically: `%global selinux_policyver %(rpm --qf "%%{version}-%%{release}" -q selinux-policy)`

https://github.com/pulp/pulp-packaging/blob/master/packages/pulp/pulp.spec#L20

Probably a better way forward.

lzap: Nice, amended. I also removed the policycoreutils version, it was probably for RHEL5, this was copy*paste from Satellite5 packaging.

packages/foreman/foreman-selinux/foreman-selinux.spec
```diff
@@ -44,7 +36,7 @@ URL:            https://theforeman.org
 Source0:        https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?prerelease:-}%{?prerelease}.tar.bz2
 
 BuildRequires:  checkpolicy, selinux-policy-devel, hardlink
-BuildRequires:  policycoreutils >= %{selinux_policycoreutils_ver}
+BuildRequires:  policycoreutils
```
ekohl: This will still be needed. Otherwise the variable is just unused.

lzap: I've removed the variable, it's gone.

ekohl: I meant that we still need to set the minimum version. Otherwise `selinux_policy_ver` is just unused.

lzap: Macro `selinux_policy_ver` is used, I have removed `selinux_policycoreutils_ver` which is not needed anymore, we don't support RHEL6.

ekohl: The `selinux_policy_ver` variable isn't magically used and we must be explicit: `BuildRequires:  policycoreutils >= %{selinux_policy_ver}`

lzap:
Not sure why I can't comment in your review, but what you want is not right:

`BuildRequires: policycoreutils >= %{selinux_policy_ver}`

Package policycoreutils has different versioning scheme than selinux-policy main package. We actually don't need to check core utils package, we used to have in RHEL5 times in Satellite 5 and this is where these lines were coming from. Since we only support RHEL7, we are only interested in selinux-policy minimum version.

PR #2917, opened 2018-08-31
Add Foreman client release repository package

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -208,7 +208,10 @@ foreman_nonscl_packages:
       build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
       nightly_package_tito_releaser_args:
         - "--arg jenkins_job=test_proxy_develop"
-    foreman-release: {}
+    foreman-release:
+      releasers:
+      - koji-foreman
+      - koji-foreman-client
     foreman-release-scl: {}
     foreman-selinux:
       releasers:
--- a/packages/foreman/foreman-release/foreman-release.spec
+++ b/packages/foreman/foreman-release/foreman-release.spec
@@ -1,4 +1,19 @@
-%global release 3
+%if 0%{?suse_version}
+%define dist suse%{?suse_version}
+%define repo_dir %{_sysconfdir}/zypp/repos.d
+
+%if 0%{?suse_version} < 1300
+%define repo_dist sles11
+%else
+%define repo_dist sles12
+%endif
+
+%else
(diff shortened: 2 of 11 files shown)


packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -24,26 +40,44 @@ Foreman repository contains open source and other distributable software for
...
+%package   client
+Summary:   Foreman client repositories meta-package
+Group:     Applications/System
+
+%description client
+Defines yum repositories for Katello clients.
+
+%files client
+%config %{repo_dir}/foreman-client.repo
+%{_sysconfdir}/pki/rpm-gpg/RPM-GPG-KEY-foreman
```
ekohl: Isn't this a duplication? We also don't sign the client repo so this shouldn't be needed

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -24,26 +40,44 @@ Foreman repository contains open source and other distributable software for
 distributions in RPM format. This package contains the repository configuration
 for Yum.
 
+%package   client
+Summary:   Foreman client repositories meta-package
+Group:     Applications/System
+
+%description client
+Defines yum repositories for Katello clients.
```
ekohl: Katello -> Foreman

packages/foreman/foreman-release/foreman-release.spec
ehelms: Duplications are allowed. We don't sign the client repo but should we?

ehelms: @ekohl question for oyu here

ekohl: We don't sign plugins because it's an automated process. If we follow that same logic then it'll depend on the release process for client repos. We could also look at automated signing (with a different key than the manually signed packages). I'd suggest starting with an automated process like plugins and unsigned. See where it goes from there.

ekohl:
There's a merge conflict now

package_manifest.yaml
```diff
@@ -60,7 +60,7 @@ foreman_scl_packages:
     rubygem-excon: {}
     rubygem-extlib: {}
     rubygem-facter: {}
-    rubygem-faraday: {}
+    rubygem-faraday:
```
ekohl: I thought this was already done in develop

ehelms: It was, I just needed to rebase this.

PR #2923, opened 2018-09-04
Split foreman packages into scl and nonscl groups

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -7,12 +7,17 @@ packages:
     katello_client_packages: {}
 
 foreman_packages:
+  children:
+    foreman_scl_packages: {}
+    foreman_nonscl_packages: {}
+    foreman_scl_and_nonscl_packages: {}
+
+foreman_scl_packages:
   vars:
     package_base_dir: 'packages/foreman/'
     diff_package_skip: false
     diff_package_tags:
       - foreman-nightly-rhel7
-      - foreman-nightly-nonscl-rhel7
     releasers:
     - koji-foreman
     nightly_releaser: koji-foreman-jenkins
@@ -23,7 +28,6 @@ foreman_packages:
       - el7-epel
       - el7-extras
       - el7-scl
-      - el7-puppet-5
       - el7-foreman-rails-nightly
       - el7-foreman-nightly
   hosts:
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -7,12 +7,17 @@ packages:
     katello_client_packages: {}
 
 foreman_packages:
+  children:
+    foreman_scl_packages: {}
+    foreman_nonscl_packages: {}
+    foreman_scl_nonscl_packages: {}
```
ehelms: @ekohl Can you re-review, as I made this change here. In the current form, I think this is required to handle dual packages. Even then, diff-package does not work properly due to how package names are derived and them living a double life (goes back to the discussion @evgeni started our SCL and non-SCL naming). I think I will need to modify diff package to accept some special arguments per tag since I doubt we will move forward the naming discussion in time

ehelms: Note @ekohl this really is just whether you agree with this dual group here. The diff package issues can be fixed afterward as they exist no matter what breakdown we choose here.

package_manifest.yaml
```diff
@@ -310,6 +174,187 @@ foreman_packages:
...
+    rubygem-concurrent-ruby: {}
+    rubygem-foreman_maintain: {}
+    rubygem-kafo: {}
+    rubygem-kafo_parsers: {}
+    rubygem-kafo_wizards: {}
+    rubygem-rb-inotify: {}
+    rubygem-rsec: {}
+    rubygem-rubyipmi: {}
+
+foreman_scl_nonscl_packages:
```
zjhuntin: Is there anyway to rename this? At a first glance it feels like the name is negating itself. Is it supposed to be packages in the foreman scl that are not strictly needed as apart of the scl?

ehelms: Would `_and_` help? Other suggestions?

zjhuntin: `foreman_scl_and_nonscl_packages` or `foreman_mixed_scl_nonscl_packages` works for me too.

ehelms:
@zjhuntin Updated with new naming scheme

zjhuntin:
ACK

ekohl:
You missed `rubygem-concurrent-ruby` which is used by both foreman and foreman-proxy.

ekohl:
https://github.com/theforeman/foreman-packaging/pull/2970 should address that

PR #2930, opened 2018-09-06
Add RPM to set dist macro

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/packages/foreman/dist-macro/dist-macro.spec
+++ b/packages/foreman/dist-macro/dist-macro.spec
@@ -1,0 +1,32 @@
+%define base_release_version 7
+%define dist .el%{base_release_version}
+
+Name:     dist-macro
+Version:  1.0
+Release:  1
+
+Summary:  Package to define the dist macro
+Group:    System Environment/Base
+License:  GPLv3+
+URL:      https://theforeman.org
+
+BuildArch: noarch
+
+%description
+Sets the dist macro to basic distribution and major version.
+
+%install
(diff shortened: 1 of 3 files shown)


packages/foreman/dist-macro/dist-macro.spec
```diff
@@ -0,0 +1,31 @@
...
+
+BuildArch: noarch
+
+%description
+Sets the dist macro to basic distribution and major version.
+
+%install
+install -d -m 755 %{buildroot}/etc/rpm
+cat >> %{buildroot}/etc/rpm/dist-macro.dist << EOF
+%%rhel %{base_release_version}
```
ekohl: Where do these values come from? Is it to copy from RHEL's release?

ehelms: Look at the top of the spec file.

ekohl: I meant to comment on the line below. I don't see `%dist` and I think it's not great to depend on the build env for it.

ehelms:
To answer your question about centos-release, at the top of the spec they define this:

```
%define dist .el%{dist_release_version}.centos
```

ekohl:
This was changed recently, see https://bugs.centos.org/view.php?id=14955. Using a recent centos-release version will remove the need for this package. At least, that's the theory :)

ehelms:
Updating Koji with centos updates repository and will re-test builds.

PR #2940, opened 2018-09-07
Use pkill instead of systemctl kill after logrotate for smart-proxy

adamruzicka:
While systemctl kill delivered the signal reliably, it delivered the
signal to every process in the service's control group. The issue was
when the service forked more processes (for example when running
ansible-playbook). pkill signals only the processes with matching
name.

Ansible-playbook reacts to SIGUSR{1,2} by printing "User defined signal
1" and exiting, which is undesired.

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---


mmoll:
merged, díky @adamruzicka!

PR #2963, opened 2018-09-11
Drop hammer requirement from katello-debug

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---


ehelms:
@ekohl I tested and it does not. This captures whatever the error is and moves along happily. We run this on both Katello and smart proxies and there are other elements in the debug not designed for on a proxy. We could split these out in the future to more debug capture files, for now this solves the problem of installing hammer on a smart proxy needlessly.

PR #2969, opened 2018-09-12
Update hammer_cli packaging to include prerelease

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
+++ b/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
@@ -4,18 +4,21 @@
 %global gem_name hammer_cli
 %global confdir hammer
 
+%global release 2
+%global prerelease .pre.develop
+
 %{!?_root_bindir:%global _root_bindir %{_bindir}}
 %{!?_root_mandir:%global _root_mandir %{_mandir}}
 %{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
 
 Summary: Universal command-line interface for Foreman
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.14.0
-Release: 1%{?dist}
+Version: 0.15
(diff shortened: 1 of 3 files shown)


packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
```diff
@@ -57,12 +60,31 @@ Obsoletes: rubygem-%{gem_name}-doc < 0.3.0-2
...
 
 %prep
-%setup -n %{pkg_name}-%{version} -q -c -T
+%{?scl:scl enable %{scl} - << \EOF}
+gem unpack %{SOURCE0}
+%{?scl:EOF}
+
+%setup -q -D -T -n  %{gem_name}-%{version}%{?prerelease}
+
 %{?scl:scl enable %{scl} - <<EOF}
```
ekohl: This shouldn't be here according to the `gem2rpm/scl.spec.erb` template. Is it really needed? If so, do we want to modify our template to allow prereleases for all gems?

ekohl:
Can we give hammer_cli_katello the same treatment so they're really the same? That would save us a lot of trouble.

ehelms:
@ekohl To do so for hammer_cli_katello requires changing how h-c-k handles it source repository. It does not presently include any notion of branch in its version.

ekohl:
Couldn't it use `.pre.master` to reflect it's from the master branch?

ehelms:
Interestingly, the 'pre' part is automatically added by Ruby when using 'gem build' if it finds non-numeric characters at the end of the version. As for adding the branch into the version, thats up to @akofink or other Katello CLI maintainers to decide. If they go that route I am happy to update the spec.

akofink:
Yes, my main goal at this point is to try to package hammer-cli-katello exactly the same as hammer-cli-foreman. I can't see why there should be any differences. Please update the h-c-k spec with this PR. :)

ehelms:
[test rpm]

PR #2970, opened 2018-09-12
Correct manifest

ekohl:
rubygem-concurrent-ruby is used in both scl and nonscl.
rubygem-faraday is used in core scl and plugins nonscl.


ehelms:
We should be getting concurrent-ruby from the Rails 5.2 SCL -- https://github.com/theforeman/rails-packaging/tree/tfm-ror52/packages/tfm-ror52-rubygem-concurrent-ruby

PR #2985, opened 2018-09-13
Move nodejs-jed to foreman

ekohl:
Foreman will [add nodejs-jed as a dependency](https://github.com/theforeman/foreman/pull/5342) so we can move it.


ekohl:
The build failed because it wants a release bump but I think that's not right in this case.

PR #2993, opened 2018-09-13
Fixes #24857 - Require foreman_ansible_core in sp_ansible

ekohl:
We place the gem 'foreman_ansible_core' in dynflows bundler.d so we should ensure it's present.

* [x] Nightly
* [x] 1.19
* [x] 1.18


evgeni:
This has *not* been built into koji due to https://projects.theforeman.org/issues/24955

PR #2994, opened 2018-09-13
Obsolete SCL rubygem-faraday_middleware

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/packages/foreman/tfm/tfm.spec
+++ b/packages/foreman/tfm/tfm.spec
@@ -19,7 +19,7 @@
 Summary: Package that installs %scl
 Name: %scl_name
 Version: 5.0
-Release: 1%{?dist}
+Release: 2%{?dist}
 License: GPLv2+
 Group: Applications/File
 Source0: README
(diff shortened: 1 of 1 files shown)


packages/foreman/tfm/tfm.spec
```diff
@@ -71,6 +71,7 @@ Obsoletes: %{scl_prefix}rubygem-concurrent-ruby < 1.0.1-2
 Obsoletes: %{scl_prefix}rubygem-dalli < 2.6.4-4
 Obsoletes: %{scl_prefix}rubygem-ejs < 1.1.1-4
 Obsoletes: %{scl_prefix}rubygem-flot-rails < 0.0.3-8
+Obsoletes: %{scl_prefix}rubygem-faraday_middleware
```
ekohl: All others are versions so we can later re-introduce them with a newer version/release if needed.

ehelms: The majority are (https://github.com/theforeman/foreman-packaging/pull/2994/files#diff-707343da1976a0fa7839fb1086a261b9R107). What happens if you need to re-introduce the same version that you are obsoleting?

ekohl: They include a release so you can bump the release. See rubygem-flot-rails above. It's `< 0.0.3-8` so you can introduce `0.0.3-8` to replace the old `0.0.3-7` package we used to have.

ehelms: I'll follow the majority convention.

PR #2998, opened 2018-09-14
Added stripping for mmv-statsd

lzap:
This is to workaround RPMDiff error *ELF Binary/Library Stripping*.

### DIFF
--- a/packages/foreman/pcp-mmvstatsd/pcp-mmvstatsd.spec
+++ b/packages/foreman/pcp-mmvstatsd/pcp-mmvstatsd.spec
@@ -2,7 +2,7 @@
 
 Name:    pcp-mmvstatsd
 Version: 0.4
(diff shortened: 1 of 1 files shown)


lzap:
That does not work for Golang binaries actually.

packages/foreman/pcp-mmvstatsd/pcp-mmvstatsd.spec
```diff
@@ -34,6 +34,7 @@ ln -fs $(pwd) _gopath/src
 export GOPATH=$(pwd)/_gopath
 pushd _gopath/src/%{name}-%{version}
 go build
+strip %{name}
```
ekohl: Looks like this needs to be `%{name}-%{version}` according to the `install` command below

lzap:
Good catch, this was not working at all!

PR #3011, opened 2018-09-17
insert nightly macro for nightly built packages

komidore64:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/packages/katello/katello-installer-base/katello-installer-base.spec
+++ b/packages/katello/katello-installer-base/katello-installer-base.spec
@@ -3,11 +3,11 @@
 %global scl_ruby /usr/bin/ruby
 
 %global prerelease master
-%global release 2
+%global release 3
 
 Name:    katello-installer-base
 Version: 3.10.0
-Release: %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?dist}
+Release: %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?nightly}%{?dist}
 Summary: Puppet-based installer for the Katello and Katello Capsule
 Group:   Applications/System
 License: GPLv3+ and ASL 2.0
@@ -118,6 +118,9 @@ ln -sf %{_datadir}/foreman-installer-katello/bin/katello-certs-check %{buildroot
 %doc README.*
 
 %changelog
+* Thu Oct 25 2018 Adam Price <komidore64@gmail.com> - 3.10.0-0.3.master
+- add nightly macro
+
 * Wed Oct 24 2018 Adam Price <komidore64@gmail.com> - 3.10.0-0.2.master
 - breakout period and hyphen from prerelease macro
 - set prerelease to master
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,10 +1,10 @@
-%global release 1
+%global release 2
 %global prerelease develop
 
 Name:       foreman-installer
 Epoch:      1
 Version:    1.21.0
-Release:    %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?dist}
+Release:    %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?nightly}%{?dist}
 Summary:    Puppet-based installer for The Foreman
 Group:      Applications/System
 License:    GPLv3+ and ASL 2.0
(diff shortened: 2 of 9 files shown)


komidore64:
[test]

komidore64:
[test rpm]

ekohl:
Should we use the `prerelease` instead of `nightly` macro? The benefit is that it automatically gets a `0.<release>` instead of `<release>`. That means it'll never have a higher priority than a released version: `0.15.0-1` is newer than `0.15.0-0.1.<nightly>`

komidore64:
`nightly` needs to be independent of `prerelease`, otherwise we'll run into instances of nightly RPMs never evaluating as "higher" than a release RPM.

Take the following example:

```
tfm-rubygem-katello-3.9.0-0.6.rc1.el7
tfm-rubygem-katello-3.9.0-0.6.rc1.20190101120013gitabcdef0.el7
tfm-rubygem-katello-3.9.0-0.6.rc1.20190102120000gitabcdef0.el7
tfm-rubygem-katello-3.9.0-0.6.rc1.20190103120000gitabcdef0.el7
tfm-rubygem-katello-3.9.0-0.7.rc1.el7
tfm-rubygem-katello-3.9.0-0.7.rc1.20190201120000gitabcdef0.el7
tfm-rubygem-katello-3.9.0-1.el7
tfm-rubygem-katello-3.9.0-1.20190301120000gitabcdef0.el7
```

See how it keeps them in order? If you're pulling from a nightly repo, you'd always get the latest.

ekohl:
Note I don't know if this is how it now works, but I'm trying to figure out what should be correct.

In my mind a nightly is always a prerelease. I don't see how we can do a final nightly release. I think the last 2 lines in your example is exactly my point. The final release is `tfm-rubygem-katello-3.9.0-1.el7`. This should IMHO always be more important than a nightly. Once `3.9.0-1` has been released, the next nightly should be `3.9.1-0.1` because that's the next release cycle.

Does that make sense?

komidore64:
"nightly" refers to bleeding edge.

> This should IMHO always be more important than a nightly.

It depends on repo context. If it's a nightly repo, then newest is king. always.

But if it's a release repo (where released packages are tagged into) then I'd argue those repos wouldn't even have nightly packages in them to begin with.

komidore64:
> Once `3.9.0-1` has been released, the next nightly should be `3.9.1-0.1` because that's the next release cycle.

I don't disagree, but there's a possibility that commits could be merged into katello before its spec gets bumped for the next release cycle. Things would still work though I believe.

komidore64:
I think I see the light now.

```
3.9.0-0.1.20190101120000gitabcdef0.el7 # nightly
3.9.0-0.1.20190102120000gitabcdef0.el7 # nightly
3.9.0-0.1.20190103120000gitabcdef0.el7 # nightly

# 3.9.0 branches --> 3.9.0-0.1.rc1.el7

# nightlies continue bumping to next targeted release
3.10.0-0.1.20190104120000gitabcdef0.el7
```

komidore64:
Well, in that case, this PR is no longer needed! Obal will just need to set the `prerelease` macro whenever building a nightly package.

komidore64:
I've since discovered that using the prerelease macro was a bad idea, because that is also used as part of the source's filename, which is no good because this requires meddling with the source's version before it's built and defining a macro at buildtime on koji which isn't possible.

ehelms:
General ACK from me, @ekohl can you take a look

komidore64:
rebased

ekohl:
I assume the installer test failure is on https://github.com/theforeman/obal/pull/141

komidore64:
correct

komidore64:
@ekohl rebased

ekohl:
Since webhooks are still disabled I'd prefer to wait with merging until this really is green.

komidore64:
this is still failing on changelog lint because the epoch isn't in the changelog version. okay if i fix that here?

ehelms:
Absolutely.

packages/katello/katello-installer-base/katello-installer-base.spec
```diff
@@ -118,6 +118,10 @@ ln -sf %{_datadir}/foreman-installer-katello/bin/katello-certs-check %{buildroot
 %doc README.*
 
 %changelog
+* Tue Oct 23 2018 Adam Price <komidore64@gmail.com> - 3.10.0-0.2.master
```
ehelms: It fails the check since the old release its comparing in git against is 3.10.0-1. Means we either just accept this one off and ignore it, or you alter the release on the old changelog commit to 3.10.0-0.1. I can live with either. Might feel weird to alter that release but its a nightly package.

komidore64: @ehelms I tried setting the previous changelog entry to `3.10.0-0.1.master` but the comparison job still fails in the same place.

ekohl: It fails because it does a `git checkout rpm/develop`, calculates the version there and then `git checkout $PR`. It compares those 2. It's a very good check, but in this case we should either accept it and merge it while we purge it from koji and repos or wait till 3.11.0.

komidore64:
this PR now depends on #3143

PR #3012, opened 2018-09-18
Fixes #24928 - add generators for requires and provides of webpack content

evgeni:
This adds automatic Provides and Requires for Foreman's webpack bundles.
    
Foreman ships webpack bundles in its RPMs to be consumed by plugins. To ensure
proper dependency relationships between the plugins and the main Foreman RPM,
this script generates Provides for the Foreman RPM that contain the hash of
the webpack bundles shipped and matching Requires for the plugins packages.
Each time the Foreman bundles change hashes, the plugins become uninstallable
and indicate that a rebuild is required.

### DIFF
--- a/packages/foreman/foreman/foreman.provreq
+++ b/packages/foreman/foreman/foreman.provreq
@@ -37,13 +37,15 @@ def parse_manifest(manifest_path):
     with open(manifest_path) as fh:
         manifest = json.load(fh)
 
-    for name, content in manifest['assetsByChunkName'].items():
-        if name != 'vendor':
-            continue
+    try:
+        name = 'vendor'
+        content = manifest['assetsByChunkName'][name]
         if isinstance(content, basestring):
             content = [content]
         for chunk in content:
             webpack_chunks.add(foreman_webpack(name, chunk))
+    except KeyError:
+        pass
 
     return webpack_chunks
 
--- a/packages/foreman/foreman/foreman.provreq
+++ b/packages/foreman/foreman/foreman.provreq
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+
+"""
+Automatic Provides and Requires generator for Foreman's webpack bundles.
+
+Foreman ships webpack bundles in its RPMs to be consumed by plugins. To ensure
+proper dependency relationships between the plugins and the main Foreman RPM,
+this script generates Provides for the Foreman RPM that contain the hash of
+the webpack bundles shipped and matching Requires for the plugins packages.
+Each time the Foreman bundles change hashes, the plugins become uninstallable
+and indicate that a rebuild is required.
+"""
+from __future__ import print_function
+
+import json
+import os
+import sys
+
+try:
+    basestring
+except NameError:
+    basestring = str
+
+FOREMAN_MANIFEST = '/var/lib/foreman/public/webpack/manifest.json'
+
+
+def foreman_webpack(name, chunk):
+    filename, fileext = os.path.splitext(chunk)
+    fileext = fileext.lstrip('.')
+    filehash = filename.rsplit('-', 1)[1]
+    return 'foreman-webpack-{}-{}({})'.format(name, fileext, filehash)
+
+
+def parse_manifest(manifest_path):
+    webpack_chunks = set()
+
+    with open(manifest_path) as fh:
+        manifest = json.load(fh)
+
+    for name, content in manifest['assetsByChunkName'].items():
+        if name != 'vendor':
+            continue
+        if isinstance(content, basestring):
+            content = [content]
+        for chunk in content:
+            webpack_chunks.add(foreman_webpack(name, chunk))
+
+    return webpack_chunks
+
+
+paths = [path.rstrip() for path in sys.stdin.readlines()]
+
+mode = os.path.splitext(sys.argv[0])[1].lstrip('.')
+
+result = set()
+
+for path in paths:
+    if path.endswith('/webpack/manifest.json') and mode == 'prov':
+        result.update(parse_manifest(path))
+    elif path.endswith('/manifest.json') and mode == 'req':
+        result.update(parse_manifest(FOREMAN_MANIFEST))
+
+print("\n".join(result))
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -8,7 +8,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 10
+%global release 11
 %global prerelease develop
 
 Name:    foreman
@@ -25,6 +25,8 @@ Source2: %{name}.sysconfig
 Source3: %{name}.logrotate
 Source4: %{name}.cron.d
 Source5: %{name}.tmpfiles
+Source6: %{name}.attr
+Source7: %{name}.provreq
 Source10: %{executor_service_name}.sysconfig
 Source11: %{executor_service_name}.service
 BuildArch:  noarch
@@ -854,6 +856,8 @@ Meta package with support for building RPMs in the Foreman release cycle.
 
 %files build
 %{_sysconfdir}/rpm/macros.%{name}-dist
+%{_fileattrsdir}/%{name}.attr
+%{_rpmconfigdir}/%{name}.*
 
 %package console
 Summary: Foreman console support
@@ -1161,6 +1165,11 @@ rm -rf ./usr \\
 %%{?-s:[ -e %%{buildroot}%%{%{name}_webpack_plugin} ] && ln -s %%{%{name}_webpack_plugin} %%{buildroot}%%{%{name}_webpack_foreman}}
 EOF
 
+#copy rpm config
+install -Dpm0644 %{SOURCE6} %{buildroot}%{_fileattrsdir}/%{name}.attr
+install -pm0755 %{SOURCE7} %{buildroot}%{_rpmconfigdir}/%{name}.prov
+install -pm0755 %{SOURCE7} %{buildroot}%{_rpmconfigdir}/%{name}.req
+
 %clean
 rm -rf %{buildroot}
 
@@ -1276,6 +1285,11 @@ exit 0
 %systemd_postun_with_restart %{name}.service
 
 %changelog
+* Fri Oct 12 2018 Evgeni Golov - 1.20.0-0.11.develop
+- Add automatic Provides and Requires for Foreman's webpack bundles.
+  This should ensure that plugins can depend on a specific bundle version and
+  get rebuilt when Foreman's bundle changes.
+
 * Mon Oct 08 2018 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> - 1.20.0-0.10.develop
 - Update Gem and NPM dependencies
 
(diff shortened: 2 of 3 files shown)


packages/foreman/foreman/foreman.spec
```diff
@@ -834,8 +834,15 @@ Group: Development/Libraries
 %description build
 Meta package with support for building RPMs in the Foreman release cycle.
 
+%install build
+install -Dpm0644 foreman.attr %{buildroot}%{_rpmconfigdir}/fileattrs/foreman.attr
```
ekohl: Shouldn't these be listed as `SourceN` and then use the `SOURCEN` macro?

evgeni: picky :P

ekohl:
I don't see it providing those. Am I querying it in the wrong way?

```
$ rpm -q --provides foreman-1.20.0-0.201809181447git42febfe0.10.develop.el7.noarch.rpm
config(foreman) = 1.20.0-0.201809181447git42febfe0.10.develop.el7
foreman = 1.20.0-0.201809181447git42febfe0.10.develop.el7
```

Do those RPM files need to be present before build the package? i.e. a `foreman-packaging` package that provides it and is a BuildRequire?

evgeni:
you need to build foreman with the *new* foreman-build from this PR, AND the path spec in the attr file is still wrong, working on it

evgeni:
this is now working:

```
# rpm -q --provides -p /root/rpmbuild/RPMS/noarch/foreman-1.20.0-0.201809181447git42febfe0.10.develop.el7.centos.noarch.rpm
config(foreman) = 1.20.0-0.201809181447git42febfe0.10.develop.el7.centos
foreman = 1.20.0-0.201809181447git42febfe0.10.develop.el7.centos
foreman-webpack-css(bundle) = bundle-ba06536ae660e9b1a134
foreman-webpack-js(bundle) = bundle-ba06536ae660e9b1a134
foreman-webpack-js(vendor) = vendor-74fdccd61c580192ebd1
```

we might need to bikeshed the exact form of the tags, though.

and I need to test my require script still.

ekohl:
I'm wondering if we should split off the attrs/foreman-build into a separate RPM to avoid a dependency on itself. That feels wrong and doesn't allow obvious from-scratch bootstrapping. Perhaps we should move it back into `foreman-release` which already is a separate package.

packages/foreman/foreman/foreman.req
```diff
@@ -0,0 +1,32 @@
...
+
+paths = [path.rstrip() for path in sys.stdin.readlines()]
+
+def foreman_webpack(name, chunk):
+    filename, fileext = os.path.splitext(chunk)
+    fileext = fileext.lstrip('.')
+    return 'foreman-webpack-{}({}) = {}'.format(fileext, name, filename)
+
+for path in paths:
+    if path.endswith('/manifest.json'):
```
ekohl: This is a check to ensure that we only generate the requires if the plugin uses webpack, correct?

evgeni:
```
# rpm -q --requires -p /root/rpmbuild/RPMS/noarch/tfm-rubygem-foreman_remote_execution-1.6.3-1.fm1_20.el7.centos.noarch.rpm |grep webpack
foreman-webpack-js(vendor) = vendor-74fdccd61c580192ebd1
```

now to bikeshedding.

I think the generated lines should be like this:

```
foreman-webpack-{name}-{ext}({hash})
foreman-webpack-vendor-js(74fdccd61c580192ebd1)
```

also, I think we should not care to add non-vendor ones to the output.

evgeni:
so with the current code (42b209f), I get the following:

```
[root@centos7 ~]# rpm -q --requires -p /root/rpmbuild/RPMS/noarch/tfm-rubygem-foreman_remote_execution-1.6.3-1.fm1_20.el7.centos.noarch.rpm |grep webpack
foreman-webpack-vendor-js(74fdccd61c580192ebd1)
[root@centos7 ~]# rpm -q --provides -p /root/rpmbuild/RPMS/noarch/tfm-rubygem-foreman_remote_execution-1.6.3-1.fm1_20.el7.centos.noarch.rpm |grep webpack
<empty>

[root@centos7 ~]# rpm -q --provides -p /root/rpmbuild/RPMS/noarch/foreman-1.20.0-0.201809181447git42febfe0.10.develop.el7.centos.noarch.rpm |grep webpack
foreman-webpack-vendor-js(74fdccd61c580192ebd1)
[root@centos7 ~]# rpm -q --requires -p /root/rpmbuild/RPMS/noarch/foreman-1.20.0-0.201809181447git42febfe0.10.develop.el7.centos.noarch.rpm |grep webpack
foreman-webpack-vendor-js(74fdccd61c580192ebd1)
```

technically, foreman does not need the require, but it also does not hurt AFAIK.

packages/foreman/foreman/foreman.req
evgeni: yes and no, it's a safenet. non-webpack plugins should not even run this code in the first place, because they don't have a path that would trigger it (see `foreman.attr`).

evgeni:
and yes, I *think* this should not live inside the actual `foreman` RPM, but that's maybe a bikeshedding topic for another day

packages/foreman/foreman/foreman.provreq
```diff
@@ -0,0 +1,51 @@
...
+mode = os.path.splitext(sys.argv[0])[1].lstrip('.')
+
+def foreman_webpack(name, chunk):
+    filename, fileext = os.path.splitext(chunk)
+    fileext = fileext.lstrip('.')
+    filehash = filename.rsplit('-', 1)[1]
+    return 'foreman-webpack-{}-{}({})'.format(name, fileext, filehash)
+
+def parse_manifest(path, mode):
+    result = set()
```
ekohl: Using the same variable name as in the top scope is usually not a good idea. Unless you hide that in a `main()`

packages/foreman/foreman/foreman.provreq
```diff
@@ -0,0 +1,51 @@
...
+
+    if path.endswith('/webpack/manifest.json') and mode == 'prov':
```
ekohl: I'd do this at a higher level and just pass in the filename to this function.

evgeni: like now?

packages/foreman/foreman/foreman.provreq
```diff
@@ -0,0 +1,47 @@
...
+Automatic provides and requires generator for webpack bundles
+"""
+
+import json
+import os
+import sys
+
+FOREMAN_MANIFEST = '/var/lib/foreman/public/webpack/manifest.json'
+
+paths = [path.rstrip() for path in sys.stdin.readlines()]
```
ekohl: Maybe we should move these definitions to below the function definitions?

evgeni: we should

packages/foreman/foreman/foreman.provreq
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/python2
```
ehelms: Should this be `/usr/bin/env python` or some equivalent to future proof it?

ehelms:
I think it would help to add a comment explaining what adding this intends to help solve.

packages/foreman/foreman/foreman.provreq
evgeni: I am not sure. The existing scripts I see use `/usr/bin/python` and `/usr/bin/python3`.

evgeni:
@ehelms added longish comment to the script and a short one to the changelog.

ehelms:
If I understand correctly, with each change to bundle or vendor this will break every webpack based plugin from an RPM installation stand point? Thus if a plugin is not a nightly built plugin, they will be broken for long stretches of time?

evgeni:
Aren't we breaking them (well, their UI part) with the bundle change anyway and this just makes the break more explicit?

ehelms:
I thought as I understood things, not every change breaks a plugins UI today. And that breakage when it does happen is restricted to the UI. This will make them completely uninstallable. I am wondering if we need to make them nightly built packages before we make this change.

ekohl:
While we don't know for sure if the hash vendor changes, it's a good indication it can break. Because we have a fairly well controlled environment with our dependencies, it's unlikely to change lot. I've started to look at the vendor bundle as an ABI rather than an API. The way it's compiled can change the outcome (strip dependencies, present them in a different way). IMHO we should stay on the cautious side and see how it plays out for a while. If it turns out we need to rebuild all webpacked plugins daily, then we should investigate why this is.

packages/foreman/foreman/foreman.provreq
ekohl: The code currently uses python-2 style code. We can make it python 3 compatible as well as python 2. Probably a good idea.

ehelms:
I agree that the `vendor` should be more like an ABI and have stability. The part that concerns me is that the `bundle` will change everytime any line of core webpack code changes and thereby break plugins on an increased cadence. I am basing this on the output from above that foreman now provides and plugins would require:

```
foreman-webpack-js(bundle) = bundle-ba06536ae660e9b1a134
```

I would be ok with just starting with `vendor` as a provides/requires. For `bundle`, I would want to push back to developers and ask is there a set of extension points in the core webpack code that can be combined into their own JS code that plugins can rely on as a versioned entity (aka an API contract).

evgeni:
The current code only does `vendor` :)

packages/foreman/foreman/foreman.provreq
```diff
@@ -0,0 +1,57 @@
...
+def parse_manifest(manifest_path):
+    webpack_chunks = set()
+
+    with open(manifest_path) as fh:
+        manifest = json.load(fh)
+
+    for name, content in manifest['assetsByChunkName'].items():
+        if name != 'vendor':
+            continue
+        if isinstance(content, basestring):
```
ekohl: This and `print` as a statement are the reason I initially made this python2 only. They're easy to work around:
```python
from __future__ import print_function
try:
    basestring
except NameError
    basestring = string
```

packages/foreman/foreman/foreman.provreq
```diff
@@ -0,0 +1,57 @@
...
+
+
+def parse_manifest(manifest_path):
+    webpack_chunks = set()
+
+    with open(manifest_path) as fh:
+        manifest = json.load(fh)
+
+    for name, content in manifest['assetsByChunkName'].items():
+        if name != 'vendor':
```
ekohl: Would it be cleaner to do `manifest['assetsByChunkName']['vendor']` with a `try/except KeyError` instead?

ehelms:
@evgeni Mind posting the latest and greatest of https://github.com/theforeman/foreman-packaging/pull/3012#issuecomment-422501630 since thats what my comments were based on? :)

packages/foreman/foreman/foreman.provreq
evgeni: done.

evgeni:
@ehelms @ekohl updated to support all the pythons, and also made the `vendor` only part more obvious as suggested by Ewoud. Please re-review.

ehelms:
@evgeni Looks to be failing to build now

evgeni:
@ehelms while compiling assets: http://koji.katello.org/kojifiles/work/tasks/6195/136195/build.log

So I dare to say: wasn't me. ;)

ehelms:
@evgeni This needs a rebase please

evgeni:
@ehelms rebased

PR #3017, opened 2018-09-18
Add rubygem-smart_proxy_m2 package

ianballou:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---


stbenjam:
ok to test

comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -133,6 +133,7 @@
       <!--
         Do not edit this section manually and use ./comps_doc.sh script
       -->
+      <packagereq type="default">rubygem-smart_proxy_m2</packagereq>
```
ekohl: Could you move this to the appropriate section where the other plugins also live?

packages/plugins/rubygem-smart_proxy_m2/rubygem-smart_proxy_m2.spec
```diff
@@ -0,0 +1,92 @@
...
+%global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
+%global foreman_proxy_settingsd_dir %{_sysconfdir}/foreman-proxy/settings.d
+
+Name: rubygem-%{gem_name}
+Version: 0.0.1
+Release: 1%{?foremandist}%{?dist}
+Summary: M2 smart proxy plugin for Foreman
+Group: Applications/Internet
+License: GPLv3
+URL: http://github.com/ianballou/smart_proxy_m2
```
ekohl: Could you use https here? I've submitted https://github.com/ianballou/smart_proxy_m2/pull/1 to correct the gemspec as well. Note that the License in the RPM spec is in the **correct** form now and doesn't need updating.

comps/comps-foreman-plugins-rhel7.xml
ianballou: Sure thing, I didn't notice this got put in a strange spot.  I think this got generated by `add_gem_package.sh`, is that script the latest recommended way to generate these?

ekohl: It should be, but the comps adding isn't the greatest code and could use some love.

PR #3024, opened 2018-09-20
Bump nodejs-patternfly-react to 2.18.0

ohadlevy:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---


ohadlevy:
@ehelms do you happen to have an idea why it fails? I've noticed the scripts now made changes to the SPEC structure, could there be an issue there? thanks. sadly this is blocking currently multiple PR's that we hope to land asap to make 1.20. thanks!

ehelms:
RPM does not handle nested sources, so format such as `nodejs-patternfly-react/@babel/runtime-corejs2-7.0.0.tgz` is failing as it not finding it in the buildroot. I think that has to be flattened.

ohadlevy:
in that case what exactly should i do? create a single for runtime-corejs2
and bundle for pf-react?

ekohl:
https://github.com/theforeman/foreman-packaging/pull/3038 replaced this

PR #3026, opened 2018-09-20
Bump rubygem-smart_proxy_dynflow{,_core} to 0.1.11

iNecas:

* [ ] Nightly
* [ ] 1.19
* [ ] 1.18
* [x] 1.17
* [x] 1.16
* [x] 1.15

---


evgeni:
@iNecas you want that tagged into 1.15 (which has [0.1.8](http://koji.katello.org/koji/buildinfo?buildID=17272)) and 1.16 (which has [0.1.7](http://koji.katello.org/koji/buildinfo?buildID=16550)), am I reading this right?

iNecas:
That's correct. I've checked the diffs and there it should be safe to get it to 0.1.11

evgeni:
after discussing this with @iNecas on IRC:
* we only really need to bump smart_proxy_dynflow, there are no changes in core we need
* but the current packaging enforces the same version of smart_proxy_dynflow and smart_proxy_dynflow_core via a strict dependency
* we can't just tag in smart_proxy_dynflow_core from 1.17 to 1.15 and 1.16 as those use a different ruby version

Ivan is going to relax the dependency between smart_proxy_dynflow and smart_proxy_dynflow_core for 1.15 and 1.16 instead

iNecas:
For reference https://github.com/theforeman/foreman-packaging/pull/3032

PR #3038, opened 2018-09-25
Bump nodejs-patternfly-react to 2.19.0

ohadlevy:
trying out another build based on latest pf version (which works fine with foreman too)

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---


ekohl:
@ohadlevy it looks like my way of dealing with the `/` in package names isn't compatible and I need to flatten it. I was sure I tested it and it worked, but I'll have a stab at this in `npm2rpm` where the real fix must be done.

ohadlevy:
@ekohl is that something you currently work on? this is blocking multiple PR's on foreman that I would hate to miss 1.20 release...

ekohl:
Not yet, mostly because I was away on PTO for a week + a public holiday which means I've only spent time catching up until now. It is good that you reminded me though.

ohadlevy:
what should we do? block foreman PR's or merge and break nightly ? (I would
hope to get this work for 1.20 if possible)...

ekohl:
I'll have a stab at changing npm2rpm to replace `/` with `-` to flatten the file structure.

ohadlevy:
thanks, will this also solve https://github.com/dLobatog/npm2rpm/issues/40#issuecomment-423162407 ? (needed for https://github.com/theforeman/foreman/pull/6056)

ekohl:
In all likeliness yes. I was at least planning on having a look a that as well.

ehelms:
For the time being, you could manually update this PR to try what is being scripted. Which I think would be:

 * modify all paths that look like `nodejs-patternfly-react/@babel/runtime-corejs2-7.0.0.tgz` to instead look like `nodejs-patternfly-react/@babel-runtime-corejs2-7.0.0.tgz`
 * update the spec file sources to reflect that change
 * push updates and see what the scratch build does

This is a bit manual but would allow us to get this moving and unblock the dependent code PRs.

ekohl:
It's possible that with my rewrite this is already solved. I'm going to try to do the same thing and see if it works.

ekohl:
Replacement PR https://github.com/theforeman/foreman-packaging/pull/3063

PR #3043, opened 2018-09-27
Add dependent client repos to foreman-client-release

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -396,8 +396,8 @@ foreman_client_packages:
       - el6-epel
       - el6-extras
       - el6-puppet-5
-      - el6-subscription-manager
       - el6-qpid
+      - el6-pulp
       - el6-foreman-client-nightly
       - el5-base
       - el5-updates
@@ -525,7 +525,6 @@ katello_client_packages:
       - el6-epel
       - el6-extras
       - el6-puppet-5
-      - el6-subscription-manager
       - el6-qpid
       - el6-katello-client-nightly
       - el5-base
@@ -769,7 +768,6 @@ repoclosures:
         - el6-epel
         - el6-extras
         - el6-puppet-5
-        - el6-subscription-manager
         - el6-qpid
     foreman-client-repoclosure-el5:
       repoclosure_config: repoclosure/yum_el5.conf
@@ -811,7 +809,6 @@ repoclosures:
         - el6-epel
         - el6-extras
         - el6-puppet-5
-        - el6-subscription-manager
         - el6-qpid
     katello-client-repoclosure-el5:
       repoclosure_config: repoclosure/yum_el5.conf
--- a/packages/foreman/foreman-release/foreman-release.spec
+++ b/packages/foreman/foreman-release/foreman-release.spec
@@ -13,7 +13,7 @@
 %define repo_dist %{dist}
 %endif
 
-%global release 5
+%global release 6
 %global prerelease develop
 
 Name:     foreman-release
@@ -30,6 +30,9 @@ Source2:  foreman.gpg
 Source3:  foreman-rails.repo
 Source4:  foreman-rails.gpg
 Source5:  foreman-client.repo
+Source6:  qpid-copr.repo
+Source7:  subscription-manager.repo
+Source8:  pulp.repo
 
 # Required by RHEL5
 BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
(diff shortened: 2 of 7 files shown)


packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -66,6 +81,18 @@ install -Dpm0644 %{SOURCE1} %{buildroot}%{repo_dir}/foreman-plugins.repo
 install -Dpm0644 %{SOURCE3} %{buildroot}%{repo_dir}/foreman-rails.repo
 install -Dpm0644 %{SOURCE5} %{buildroot}%{repo_dir}/foreman-client.repo
 
+%if 0%{?rhel}
+install -m 644 %{SOURCE6} %{buildroot}%{repo_dir}/pulp.repo
```
ekohl: This should be `%{SOURCE8}`

ekohl:
I can see us using the client repo on a server install. Then you install the stable here but katello itself might want to install an explicit 2.17 for example. I'd prefer to depend on a pulp-release without a version constraint. It's more flexible and avoids the duplication.

Also thinking that the client repo can house hammer and @timogoebel also had a client tool in mind. You don't want to enable pulp repositories on all your clients IMHO.

ehelms:
Got a suggestion then for how to handle these situations? Other ideas I had were:

 1) Pure documentation, less easy for users but avoids all issues raised
 2) Sub-package to capture these dependencies: foreman-client-release-agent, foreman-client-release-katello, foreman-client-release-gofer

ekohl:
I'd prefer having a `pulp-release` package. IMHO we can start with hard depending on it. I think we should remove this and also depend on `pulp-release` in katello-repos`, probably with a `>= 2.17` or similar:
https://github.com/theforeman/foreman-packaging/blob/edf431b1a9e0c5bcd3ed5b3a04e3eb95a94a2d76/packages/katello/katello-repos/katello.repo#L10-L15 
https://github.com/theforeman/foreman-packaging/blob/edf431b1a9e0c5bcd3ed5b3a04e3eb95a94a2d76/packages/katello/katello-repos/katello.repo#L36-L41

It does make me wonder how we should bootstrap this in our installation instructions.

ehelms:
That only solves the Pulp issue. Your concerns seemed to also be in general around configuring additional repositories which are also:

 * EL6: qpid-copr.repo
 * EL5: subscription-manager.repo

How would we handle this?

ehelms:
I am leaning towards starting with updating documentation, dropping this change and updating Forklift roles to handle automated setup of these repositories when agent is set to true.

ekohl:
Personally I don't care as much about EL5 and EL6. In the context of servers those are irrelevant. I don't mind less pretty solutions for legacy :)

Starting with documentation is also a good enough solution. What do you think of doing it in the `%desc` section as well?

ehelms:
Oh! That comment reminded me -- EL7 is the *easy* case. Due to changes in katello-host-tools, it does not require *any* special repositories which include Pulp. Only EL6 and EL5 require special configured repositories. I have updated to try to capture that.

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -66,6 +79,19 @@ install -Dpm0644 %{SOURCE1} %{buildroot}%{repo_dir}/foreman-plugins.repo
 install -Dpm0644 %{SOURCE3} %{buildroot}%{repo_dir}/foreman-rails.repo
 install -Dpm0644 %{SOURCE5} %{buildroot}%{repo_dir}/foreman-client.repo
 
+%if 0%{?rhel}
+%endif
+
+%if 0%{?rhel} == 6
+install -m 644 %{SOURCE6} %{buildroot}%{repo_dir}/pulp.repo
```
ekohl: SOURCE8 here

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -66,6 +79,18 @@ install -Dpm0644 %{SOURCE1} %{buildroot}%{repo_dir}/foreman-plugins.repo
 install -Dpm0644 %{SOURCE3} %{buildroot}%{repo_dir}/foreman-rails.repo
 install -Dpm0644 %{SOURCE5} %{buildroot}%{repo_dir}/foreman-client.repo
 
+%if 0%{?rhel}
```
ekohl: This feels like a dummy

ehelms:
@ekohl Good with this update?

package_manifest.yaml
```diff
@@ -398,6 +398,7 @@ foreman_client_packages:
       - el6-puppet-5
       - el6-subscription-manager
```
ekohl: This doesn't sound correct if I look at the repos here

ehelms: What do you mean?

ekohl: In the client repos we don't install the subscription manager copr for EL6 so why is it needed for this repoclosure?

PR #3051, opened 2018-10-03
Update add_npm_package to use npm2rpm 4.0

ekohl:
Needs https://github.com/dLobatog/npm2rpm/pull/48

### DIFF
--- a/add_npm_package.sh
+++ b/add_npm_package.sh
@@ -7,7 +7,7 @@ TITO_TAG=foreman-nightly-nonscl-rhel7
 DISTRO=${TITO_TAG##*-}
 BASE_DIR=${4:-foreman}
 
-PACKAGE_NAME=nodejs-$NPM_MODULE_NAME
+PACKAGE_NAME=nodejs-${NPM_MODULE_NAME##*/}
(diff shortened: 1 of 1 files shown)


add_npm_package.sh
```diff
@@ -20,7 +20,9 @@ ensure_program() {
   if !(program_exists $1); then
     echo "$1 is not installed - you can install it with"
     if [[ $1 == "npm2pm" ]] ; then
-      echo "sudo npm install npm2rpm"
+      echo "sudo npm install npm2rpm@4.0.0"
```
evgeni: this makes it not pull any further updates, is that desired?

add_npm_package.sh
```diff
@@ -91,6 +90,11 @@ add_npm_to_comps() {
   git add comps/
 }
 
+npm_info() {
+  local name=$(python -c "import urllib ; print(urllib.quote('$NPM_MODULE_NAME', safe=''))")
```
evgeni: this depends on an external variable `NPM_MODULE_NAME` being set, any reason not to use `$1` and then calling `npm_info $NPM_MODULE_NAME`?

ekohl: Good point.

PR #3055, opened 2018-10-04
Fixes #24927 - Fix sending signal after logrotate

adamruzicka:
The solution with using pkill was not ideal as pkill
would match the process it was meant to kill and the
shell running pkill as well. This could lead to
a situation where the shell would be killed first
and the signal would not be delivered to the intended
target process.

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---


mmoll:
merged, díky @adamruzicka!

PR #3065, opened 2018-10-04
[WIP] use bundle exec with dynflowd

mmoll:
this should be built into:

* [x] Nightly

### DIFF
--- a/debian/bionic/foreman/dynflowd.service
+++ b/debian/bionic/foreman/dynflowd.service
@@ -8,8 +8,8 @@ Type=forking
 User=foreman
 TimeoutSec=90
 WorkingDirectory=/usr/share/foreman
-ExecStart=/usr/bin/dynflowd start
-ExecStop=/usr/bin/dynflowd stop
+ExecStart=/usr/bin/bundle exec script/dynflowd start
+ExecStop=/usr/bin/bundle exec script/dynflowd stop
 EnvironmentFile=-/etc/default/dynflowd
 
 [Install]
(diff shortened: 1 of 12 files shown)


mmoll:
@lzap @ekohl @evgeni I have no idea if that's a good idea, but I did build a bionic DEB (https://ci.theforeman.org/view/Packaging/job/packaging_build_deb_coreproject/arch=x86,label=debian,os=bionic/5903/) and it works for me...

evgeni:
I think as a user I'd prefer to have a working `/usr/bin/dynflowd`. Can we work with the shebang?

How does it work in RPM? Ah, we don't bundle in RPMs... Damn.

mmoll:
I think as a user I'd prefer to have a working `/usr/bin/dynflowd`.

Hrm, why? Is this directly uaed somewhere?

evgeni:
Mostly to match RPMs, and secondly to be able to run `--help`. Idk, maybe I'm strange ;)

evgeni:
How did it work before the foreman-tasks -> dynflowd move?

mmoll:
> How did it work before the foreman-tasks -> dynflowd move?

Like dynflowd. It worked until theforeman/foreman@3784e5bc9f17c38a3cde9bf2c5e458, but IMHO using using bundler is cleaner anyway, as this is about Foreman core using the same environmenet (AFAICT).

evgeni:
re-posting from IRC, so that it's archived:

I think there is a certain "a daemon should have a callable binary" somewhere in my head, but I don't think there is a real technical need for it. and given the current setup is broken, and your patch fixes it, we might as well merge it and discuss the utopic beatiful solution afterwards

PR #3067, opened 2018-10-05
Adds logging-journald to nonscl

lzap:
This is my attempt to copy logging-journald to nonscl repository. There
is a patch pending review which adds new dependency for improved logging
stack and abilities:

https://github.com/theforeman/smart-proxy/pull/611

Once the two deps are merged, I will work on packaging changes of
foreman-proxy package (new bundler group).

* [x] Nightly

### DIFF
--- a/comps/comps-foreman-rhel7.xml
+++ b/comps/comps-foreman-rhel7.xml
@@ -289,6 +289,8 @@
       <packagereq type="default">rubygem-concurrent-ruby</packagereq>
       <packagereq type="default">rubygem-hashie</packagereq>
       <packagereq type="default">rubygem-highline</packagereq>
+      <packagereq type="default">rubygem-journald-logger</packagereq>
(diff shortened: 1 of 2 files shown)


lzap:
Do not merge, it looks like for non-SCL version we need `journald-logger` version 1.1.1 because 2.x requires Ruby 2.1. I am not sure how to approach this, that's downgrade in non-SCL but we want to keep latest and greatest in SCL.

lzap:
So I was able to pin 1.1.1 version for plain gem use on Ruby 2.0. See the discussion here, but I don't believe we need to change anything in RPMs now. This is now ready for merge: https://github.com/theforeman/smart-proxy/pull/611

ekohl:
This forgot to change `rel-eng/tito.props` which means it won't be built into the right tags on update.

lzap:
Let me sort it out in: https://github.com/theforeman/foreman-packaging/pull/3084

PR #3074, opened 2018-10-09
Drop java dependency in katello meta RPM

ehelms:

* [x] Nightly
* [ ] 1.19
* [ ] 1.18
* [ ] 1.17

---

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -4,7 +4,7 @@
 %global homedir %{_datarootdir}/%{name}
 %global confdir common
 # %%global prerelease .rc1
-%global release 9
+%global release 10
 
 Name:       katello
 Version:    3.9.0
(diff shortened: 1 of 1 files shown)


packages/katello/katello/katello.spec
```diff
@@ -58,8 +58,6 @@ Requires: mod_xsendfile
 Requires(post): candlepin >= 2.0
```
ekohl: Perhaps this should be a regular Requires now?

packages/katello/katello/katello.spec
```diff
@@ -58,8 +58,6 @@ Requires: mod_xsendfile
...
 Requires: rubygem-foreman_maintain >= 0.2.2
 Requires: candlepin-selinux >= 2.0
-Requires: java-openjdk >= 1:1.7.0
-Requires: java-openjdk < 1:1.8.0.45
 Requires: /usr/bin/psql
 Requires: /usr/bin/pg_dump
```
ekohl: These can probably be removed as well because foreman_maintain now has the backup scripts that require these

ehelms: I can do a targeted further update to clean this up but I'd like to keep this change focused.

PR #3077, opened 2018-10-10
Unbundle nodejs-patternfly-react dependencies

ekohl:
This still contains some helper commits that shouldn't be part of the final merge. This is now here to check if it builds.


package_manifest.yaml
```diff
@@ -7,20 +7,22 @@ packages:
     katello_packages: {}
     katello_client_packages: {}
 
+# Packages in the foreman directory
```
ehelms: I feel like we've failed at naming if these comments are needed.

ekohl: They were a trick to let the YAML reformatting maintain the newline. In the final PR I'll drop this commit and discuss it in a separate commit. It did allow me to script the adding of packages to the manifest though so that was a lot less work.

ekohl:
Currently it fails repoclosure because some packages require warning 3.x and others 4.x. Similar thing with performance-now. Also forgot react-is. I'll see if we can bundle those packages.

ekohl:
I believe this should be ready to merge now. Hopefully the build will be green as well.

tbrisker:
Looks like https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/rubygem-katello/rubygem-katello.spec might also need an update to match this, latest build fails to find all the react-bootstrap submodules (e.g. it is looking for `node_modules/react-bootstrap/node_modules/prop-types-extra/lib/elementType.js` while vendor.js exposes `node_modules/prop-types-extra/lib/elementType.js`)

PR #3081, opened 2018-10-10
Reformat the manifest

ekohl:
This is based on ruemal-yaml which maintains as much style as possible but is strict about indenting levels and quotes. It prefers unquoted when possible and falls back to single quotes.

https://github.com/theforeman/obal/pull/59 was used to format this. I haven't found a way to prefer double quotes over single quotes.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -20,7 +20,7 @@ foreman_scl_packages:
     diff_package_tags:
       - foreman-nightly-rhel7
     releasers:
-    - koji-foreman
+      - koji-foreman
     nightly_releaser: koji-foreman-jenkins
     repoclosure_config: repoclosure/yum_el7.conf
     repoclosure_lookaside_repos:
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -34,10 +34,10 @@ foreman_scl_packages:
   hosts:
     foreman:
       releasers:
-        - "{{ nightly_releaser }}"
-      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
+        - '{{ nightly_releaser }}'
+      build_package_tito_releaser_args: '{{ nightly_package_tito_releaser_args }}'
       nightly_package_tito_releaser_args:
-        - "--arg jenkins_job=test_develop"
+        - --arg jenkins_job=test_develop
```
evgeni: I think this is the only change I don't like, but I can live with it if we can't teach the tool to do what I want.

ekohl: Redone with `preserve_quotes = True`

PR #3092, opened 2018-10-12
Fixes #25195 - Avoid removing mounted filesystems & support for wildcard

chris1984:
Currently, Fileutils.rm_rf(array) does not support wildcards so files are being left behind, we do not see it error out since rm_rf does not stop on errors. I added Dir.glob to support that and now the cleanup works with wildcards. I removed some legacy files/dirs that are no longer present in Foreman and Katello today. I also added color support for more awareness when the script is running. I also sorted the items in the array alphabetical so it is easier to follow.

Results of testing:
```
Mounts:

├─/var/lib/pulp                       /dev/mapper/centos-var_lib_pulp    xfs        rw,relatime,seclabel,attr2,inode64,noquota
├─/var/lib/pgsql                      /dev/mapper/centos-var_lib_pgsql   xfs        rw,relatime,seclabel,attr2,inode64,noquota
└─/var/lib/mongodb                    /dev/mapper/centos-var_lib_mongodb xfs        rw,relatime,seclabel,attr2,inode64,noquota

Removing Databases and Pulp storage on Filesystem

 /var/lib/mongodb is part of a mounted filesystem, skipping. You will need to remove /var/lib/mongodb manually.

Without mounted filesystem:

Type [remove] to continue 
remove

Starting removal

Stopping services

Removing RPMs

Cleaning up configuration files

Cleaning up log files

Cleaning up certs

Cleaning up content

Removing Databases and Pulp storage on Filesystem

Finished
```

### DIFF
--- a/packages/katello/katello/katello-remove
+++ b/packages/katello/katello/katello-remove
@@ -16,39 +16,55 @@ optparse = OptionParser.new do |opts|
   opts.parse!
 end
 
+# Define colors we can use for stdout messages
+def important(message)
+  "\033[31m#{message}\033[0m"
+end
+
+def info(message)
+  "\033[32m#{message}\033[0m"
+end
+
+def warn(message)
+  "\033[33m#{message}\33[0m"
+end
+
 CONFIG_FILES = [
-  '/etc/pulp',
-  '/etc/opt/rh/rh-mongodb34/mongod.conf',
   '/etc/candlepin',
-  '/etc/katello',
-  '/etc/httpd',
-  '/etc/foreman',
-  '/etc/tomcat',
+  '/etc/default/pulp_workers.rpmsave',
   '/etc/foreman-installer',
   '/etc/foreman-proxy',
-  '/etc/pki/katello-certs-tools',
-  '/etc/sudoers.d/foreman-proxy',
+  '/etc/foreman',
   '/etc/hammer',
-  '/etc/tomcat',
-  '/etc/squid',
+  '/etc/httpd',
+  '/etc/opt/rh/rh-mongodb34/mongod.conf',
+  '/etc/opt/theforeman',
+  '/etc/pki/katello-certs-tools',
+  '/etc/pulp',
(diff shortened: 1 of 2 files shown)


ehelms:
Can you bump the release of the katello spec file so we can build the change? If you install obal you can do:

```
obal bump-release katello
```

packages/katello/katello/katello-remove
```diff
@@ -141,38 +164,49 @@ CONTENT.map! do |dir|
 end
 
 def remove
-  puts 'Stopping services'
-  `katello-service stop`
+  puts "\nStopping services".yellow
+  `foreman-maintain service stop`
 
-  puts 'Removing RPMs'
+  puts "\nRemoving RPMs".yellow
```
ehelms: Design wise, it might be a bit cleaner to use semantic naming instead of the colors themselves and avoid opening up the native String class. The example would be calling strings like this:

```
warn("Removing RPMs")
info("")
important("")
```

Not a requirement, but a thought on another way to approach it. I am fine with whichever way you want to go.

packages/katello/katello/katello-remove
```diff
@@ -141,38 +164,49 @@ CONTENT.map! do |dir|
...
+  puts "\nCleaning up content".yellow
+  # Content
+  FileUtils.rm_r Dir.glob(CONTENT)
+
+  puts "\nRemoving Databases and Pulp storage on Filesystem".yellow
+  # Check and see if MOUNTS are mounted filesystems.
+  MOUNTS.each do |mount|
+    `mountpoint -q #{mount}`
+    if $?.success?
+      puts "\n #{mount} is part of a mounted filesystem, skipping. You will need to remove #{mount} manually.".red
```
ekohl: Should this remove `Dir.glob(File.join(mount, '*'))`? Probably good to ensure it's not `/*` just to be sure though.

chris1984:
@ehelms updated

ehelms:
[test rpm]

PR #3093, opened 2018-10-15
logging-journald gem and friends

mmoll:
this should be built into:

* [x] Nightly


mmoll:
The licence files under debian/* are not correct right now, but I'll re-import all of this, once I got all packages upstreamed.

PR #3098, opened 2018-10-16
add foreman-proxy-journald meta package

mmoll:
this should be built into:

* [x] Nightly


ekohl:
If we're going to introduce a subpackage for RPMs, I'd like to have it too for Debs so the installer doesn't have to know. Is that hard to do?

mmoll:
> Is that hard to do?

Nope, will do it later.

lzap:
Just probing how far this is in your TODO @mmoll as I am trying to hit with the logging feature 1.20 ;-)

ekohl:
Does anything needs to happen in the bundler config to ensure it's installed is that irrelevant because it's a system gem?

mmoll:
> is that irrelevant because it's a system gem?

exactly.

PR #3099, opened 2018-10-16
add ruby-jwt recommendation to foreman-proxy

mmoll:
this should be built into:

* [x] Nightly


mmoll:
```
smart-proxy$ cat bundler.d/puppetca_token_whitelisting.rb 
group :puppetca_token_whitelisting do
  if RUBY_VERSION < '2.1'
    gem 'jwt', '~> 1.5.6'
  else
    gem 'jwt'
  end
end
```

PR #3101, opened 2018-10-16
Update fog-vsphere to 2.4.0

timogoebel:
This is only for nightly (Foreman 1.21).


dLobatog:
http://koji.katello.org/koji/taskinfo?taskID=143512

ekohl:
@dLobatog for `rpm/develop` you no longer need to release packages. We have https://ci.theforeman.org/job/foreman-packaging-release/ for that.

timogoebel:
> why is this only for nightly though?

To reduce the risk of breaking stable versions. :-)
We can make a point release for 2.3.1 that contains fixes if you want.

ekohl:
@timogoebel if it's a breaking release, shouldn't that have been a `3.0.0` with semver?

timogoebel:
@ekohl: Well, it's not a breaking release, but it contains new features and not only bugfixes. The version bump was chosen wisely.

PR #3104, opened 2018-10-17
Remove incorrect systemd service alias

ekohl:


packages/foreman/foreman/foreman.service
```diff
@@ -14,4 +14,3 @@ EnvironmentFile=-/etc/sysconfig/foreman
 
 [Install]
 WantedBy=multi-user.target
-Alias=foreman-tasks.service
```
ehelms: Doh! Good catch

PR #3108, opened 2018-10-17
katello-host-tools 3.3.6 release

jlsherrill:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---


jturel:
Might need to add a check for zypper_install here too:

https://github.com/theforeman/foreman-packaging/blob/acc5bf86445a0ec7a8caada1d93a75a60625bafc/packages/client/katello-host-tools/katello-host-tools.spec#L184

jturel:
Here too

https://github.com/theforeman/foreman-packaging/blob/acc5bf86445a0ec7a8caada1d93a75a60625bafc/packages/client/katello-host-tools/katello-host-tools.spec#L301

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -179,7 +181,7 @@ rm %{buildroot}%{katello_libdir}/agent/goferd/legacy_plugin.py
 %endif
 %endif
 
-%if %{dnf_install} || %{yum_install}
+%if %{dnf_install} || %{yum_install} || %{zypper_install}
```
ehelms: Is there a build that does not trigger one of these at this point?

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -296,7 +298,7 @@ exit 0
 %doc LICENSE
 %dir %{_localstatedir}/cache/katello-agent/
 
-%if %{dnf_install} || %{yum_install}
+%if %{dnf_install} || %{yum_install} || %{zypper_install}
```
ehelms: Is there a build that does not trigger one of these at this point?

packages/client/katello-host-tools/katello-host-tools.spec
jlsherrill: There isn't now, but dnf_install will be removed as part of the 3.4 release, so it seemed simpler to just add 1 condition and remove one later than remove the entire check

packages/client/katello-host-tools/katello-host-tools.spec
jlsherrill: same as above

jlsherrill:
@ehelms not really, 3.10 is fine

PR #3110, opened 2018-10-18
Bump katello versions to 3.10

ehelms:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -471,6 +471,7 @@ katello_packages:
       build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
       nightly_package_tito_releaser_args:
         - "--arg jenkins_job=katello-installer-nightly-release"
+    katello-repos: {}
     katello-selinux: {}
     nodejs-bootstrap-select: {}
     nodejs-downshift: {}
@@ -515,8 +516,6 @@ katello_client_packages:
       - katello-nightly-rhel5
       - katello-nightly-fedora27
       - katello-nightly-fedora28
-    releasers:
-      - koji-katello-client
     repoclosure_config: repoclosure/yum.conf
     repoclosure_lookaside_repos:
       - el7-base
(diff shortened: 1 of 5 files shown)


package_manifest.yaml
```diff
@@ -499,6 +501,7 @@ katello_packages:
       build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
       nightly_package_tito_releaser_args:
         - "--arg jenkins_job=katello-master-release"
+        - "--arg jenkins_job_id=lastBuild"
```
ekohl: Why lastBuild instead of the last successful build (which we default to)?

package_manifest.yaml
```diff
@@ -471,6 +471,8 @@ katello_packages:
       build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
       nightly_package_tito_releaser_args:
         - "--arg jenkins_job=katello-installer-nightly-release"
+        - "--arg jenkins_job_id=lastBuild"
```
ekohl: Same here

ehelms:
[test rpm]

package_manifest.yaml
ehelms: Chicken and egg problem the way those jobs work right now. They both build source and the RPM. However, if we are in this current state, the lastSuccessfulBuild will never point to a usable source. Further, we aren't validating the source, so we just need to most recently generated source.

ekohl: But `katello-installer-nightly-release` triggers a build with an explicit job ID so it'd be a matter of restarting that job. I don't think there's a need to pollute our history with this change and then changing it back. Other reasons for failure might be intermittent network failures. We also had jobs that timed out after 2 hours so had no artifact (more relevant on the installer, but still).

ehelms:
This won't pass any tests due to chicken and egg problem so I'm going to merge it.

PR #3083, opened 2018-10-11
Update katello package dependencies

ekohl:

PR #3124, opened 2018-10-23
Remove *.js.map file after asset precompile

evgeni:
Inspired by https://github.com/theforeman/foreman-packaging/pull/3083

* [x] Nightly
* [x] 1.20
* [ ] 1.19
* [ ] 1.18

---

### DIFF
--- a/packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
+++ b/packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
@@ -7,7 +7,7 @@
 Summary: Ansible integration with Foreman (theforeman.org)
 Name:    %{?scl_prefix}rubygem-%{gem_name}
 Version: 2.2.9
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Group:   Applications/System
 License: GPLv3
 URL:     https://github.com/theforeman/foreman_ansible
@@ -76,6 +76,7 @@ cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
 
 %foreman_bundlerd_file
 %foreman_precompile_plugin -a -s
+rm -f %{buildroot}%{foreman_webpack_plugin}/*.js.map
 
 %files
 %dir %{gem_instdir}
@@ -111,6 +112,9 @@ exit 0
 
 
 %changelog
+* Tue Oct 23 2018 Evgeni Golov - 2.2.9-2
+- Remove *.js.map file after asset precompile
+
 * Tue Sep 25 2018 Marek Hulan <mhulan@redhat.com> 2.2.9-1
 - Update to 2.2.9
 


ohadlevy:
@tstrachota doesnt this goes against your original intent? just want to make sure we are not missing anything...

evgeni:
@ohadlevy the `.js.map` files work fine in Foreman itself, but not in Plugins. One of the issues is that they contain build-time filepaths, which Koji/Brew forbids to be shipped in an RPM. So without the above change, the plugin does not build to RPM.

ohadlevy:
@evgeni I understand, I wonder in that case if we shouldn't remove the webpack change (to generate it in the first place) as I believe the original goal was to include it in the RPM...

tstrachota:
Yes, that was the original intent. If we can keep the maps at least for the core then it's also an improvement and I wouldn't remove it from the webpack config.

@evgeni what do we need to change to get the maps in for plugins too?

evgeni:
We do ship the maps in core at the moment, and there it seems to work just fine.

@tstrachota I think the current problem was (and @ekohl can correct me here) that the `.map` file contains build-time paths and RPM does not like this:
```
+ /usr/lib/rpm/check-buildroot
/builddir/build/BUILDROOT/tfm-rubygem-foreman_ansible-2.2.9-3.el7sat.noarch/opt/theforeman/tfm/root/usr/share/gems/gems/foreman_ansible-2.2.9/public/webpack/foreman_ansible/foreman_ansible-3e92a72ba2eea7b84016.js.map:{"version":3,"sources":["webpack:////builddir/build/BUILDROOT/tfm-rubygem-foreman_ansible-2.2.9-3.el7sat.noarch/opt/theforeman/tfm/root/usr/share/gems/gems/foreman_ansible-2.2.9/webpack/components/ReportJsonViewer.js","webpack:////builddir/build/BUILDROOT/tfm-rubygem-foreman_ansible-2.2.9-3.el7sat.noarch/opt/theforeman/tfm/root/usr/share/gems/gems/foreman_ansible-2.2.9/webpack/index.js","webpack:///./webpack/assets/javascripts/foreman_editor.js","webpack:///./webpack/assets/javascripts/foreman_navigation.js",…
Found '/builddir/build/BUILDROOT/tfm-rubygem-foreman_ansible-2.2.9-3.el7sat.noarch' in installed files; aborting
```

the core maps only reference `/usr/lib/node_modules/`, not `/builddir/build/BUILDROOT/`

if we can generate maps that do not contain these (`/builddir/build/BUILDROOT/`) paths, we should be good (at least on the RPM side, no idea how useable these maps are).

ohadlevy:
i wonder why source maps include the full path, but i guess its just to
recreate the source directory structure, nothing more - which imho is
harmless?

tstrachota:
I also think it should be harmless. I wonder if this setting could help us: https://webpack.js.org/configuration/output/#output-devtoolmodulefilenametemplate

evgeni:
Yeah. I thought the same, but didn't try yet.

ekohl:
Replaced by https://github.com/theforeman/foreman-packaging/pull/3152. There's a generic fix in https://github.com/theforeman/foreman-packaging/pull/3151 so if we find a solution it only needs to be fixed in 1 place.

PR #3125, opened 2018-10-23
Fixes #20302 - Update rubygem-fog-vsphere to 2.4.0

dLobatog:

* [x] Nightly
* [x] 1.20
* [x] 1.19
* [x] 1.18

---

I think according to the [Gemfile](https://github.com/theforeman/foreman/blob/develop/bundler.d/vmware.rb#L2 ) and the [foreman spec](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/foreman/foreman.spec#L580) just by releasing a new gem, this would be used right away.

cc @timogoebel @chris1984  as the reason I submitted this mostly https://github.com/fog/fog-vsphere/pull/157 as you can see in the redmine issue,


timogoebel:
See #3101.
Note, that 2.4.0 includes huge performance improvement - but some serious rewrites were necessary for this.

dLobatog:
@timogoebel Oops, thanks, I guess it was too easy to just run .bump_rpm :)

PR #3152, opened 2018-10-31
Correct source map behavior

ekohl:
This is now handled in the foreman_precompile_plugin macro so individual plugins no longer need to care about this.


ekohl:
Repoclosure fails because of an infra issue. We're running repoclosure on rubygem-katello.rpm with the repos from just plugins. Merging because of it.

PR #3180, opened 2018-11-11
Add rubygem-xmlrpc package

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.20
* [ ] 1.19
* [ ] 1.18

This is a new dependency of fog-xenserver 1.0 (to come in with another PR, after this one got merged).


mmoll:
I'm just seeing, this requires Ruby 2.3+, so it's probably only suited for nightly and 1.20, as the ROR51 SCL is Ruby 2.2 based, correct?

ekohl:
It also failed because you didn't add the package to `package_manifest.yaml`.

ekohl:
And yes, I think that ROR51 is still ruby 2.2 which would make it 1.20+ I think.

mmoll:
I added some stuff and it's :green_heart: now, but please check thoroughly. ;)

mmoll:
anybody? :)

PR #3182, opened 2018-11-12
make katello-common truly common

evgeni:
* move Katello specific files to katello itself
* make foreman-proxy-content require k-common

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -4,7 +4,7 @@
 %global homedir %{_datarootdir}/%{name}
 %global confdir common
 # %%global prerelease .rc1
-%global release 3
+%global release 4
 
 Name:       katello
 Version:    3.10.0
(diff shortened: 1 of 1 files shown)


packages/katello/katello/katello.spec
```diff
@@ -172,19 +172,14 @@ Requires: rh-mongodb34
 Requires: foreman-installer-%{name}
 Requires: rubygem-highline
```
ekohl: Why is this still needed?

ekohl:
Overall it looks good.

packages/katello/katello/katello.spec
evgeni: Probably because of https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello/hostname-change.rb#L9 but that should live in -common then too.

PR #3184, opened 2018-11-13
Drop findutils and highline from foreman-proxy-content requires

evgeni:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---


ehelms:
Any idea what required them previously that does not now?

packages/katello/katello/katello.spec
```diff
@@ -167,10 +167,8 @@ Useful utilities for debug info collecting
 %package -n foreman-proxy-content
 Summary: Provides a federation of katello services
 BuildArch: noarch
-Requires: findutils
```
ekohl: This is still used: https://github.com/theforeman/foreman-packaging/blob/0d038ed8b626b0e8dc53690c59cbe6187c473c6a/packages/katello/katello/foreman-proxy-content.cron#L5

evgeni: good catch, reverted that part.

evgeni:
> Any idea what required them previously that does not now?

highline was used in the ruby based helpers we shipped, but those are in `-common` now :)

PR #3185, opened 2018-11-13
Refs #25289 - Add back Puppet clean cron to Katello

evgeni:
this was initially moved to f-p-c only, but the main Katello
installation might have empty Puppet environments too.

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---


evgeni:
@sean797 can you please have a look?

sean797:
I believe Katello always has an ‘internal’ foreman proxy so this would result in 2 of the same crons ?

evgeni:
Yeah, it has an internal proxy, but the `f-p-c` RPM is not installed there.

evgeni:
(rebased)

sean797:
In that case, I agree with @ekohl, rubygem-smart_proxy_pulp would probably be a better place, that or the installer, but I don't think we normally manage crons with that.

@evgeni would you mind moving it?

evgeni:
@sean797 @ekohl like this?

packages/plugins/rubygem-smart_proxy_pulp/rubygem-smart_proxy_pulp.spec
```diff
@@ -60,6 +61,10 @@ mkdir -p  %{buildroot}%{_sysconfdir}/foreman-proxy/settings.d/
 cp -pa .%{gem_instdir}/settings.d/pulp.yml.example %{buildroot}%{foreman_proxy_settingsd_dir}/pulp.yml
 cp -pa .%{gem_instdir}/settings.d/pulpnode.yml.example %{buildroot}%{foreman_proxy_settingsd_dir}/pulpnode.yml
 
+#copy cron scripts to be scheduled
+install -d -m0755 %{buildroot}%{_sysconfdir}/cron.d
+install -m 644 %{SOURCE1} %{buildroot}%{_sysconfdir}/cron.d/%{name}
```
ekohl: Should this be `install -D` and drop the line above it?

evgeni: probably. copy&paste from katello.spec ;)

ekohl:
The build failure is an issue with `obal repoclosure`: it runs it against katello in a non-scl env so that fails.

ehelms:
The better question is, do we even need this anymore? My research says no, as this was a work around for some bad implementation on Pulp's part that was fixed. Please double check me on this based on:

https://projects.theforeman.org/issues/15845 which leads to https://pulp.plan.io/issues/2108

@jlsherrill mind weighing in since you originally added this?

evgeni:
> The better question is, do we even need this anymore? My research says no, as this was a work around for some bad implementation on Pulp's part that was fixed. Please double check me on this based on:
> 
> https://projects.theforeman.org/issues/15845 which leads to https://pulp.plan.io/issues/2108

While the Pulp issue is fixed, I don't think we use that feature as we don't set the `subdir` config element in https://github.com/Katello/runcible/blob/master/lib/runcible/models/puppet_install_distributor.rb.

The patch should quite easy to create, but is nevertheless required.

jlsherrill:
@evgeni i think is correct, pulp added the ability to remove these directories, but we aren't currently configuring the distributor to use that feature.

ehelms:
Then this change is fine but we *should* fix this properly so we can drop a cron job.

evgeni:
Eric, could you please open a redmine for it? Don't have my credentials with me atm.

ehelms:
Your Redmine issue sir -- https://projects.theforeman.org/issues/25576

evgeni:
I've rebased this. But I also moved https://projects.theforeman.org/issues/25576 to "ready to test", so maybe we should just close here?

PR #3199, opened 2018-11-15
Remove rubygem-puppet-strings

ekohl:
This package was used for the Puppet system package. We no longer support that and only use the puppet-agent-puppet-strings package.


ekohl:
Note it wasn't present in tito.props and comps. Probably because this was used on Fedora. Given we had some chats about deploying on Fedora again, it may also be good to leave this is but just update it later.

PR #3201, opened 2018-11-15
drop foremandist from the release of core gems

evgeni:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---


evgeni:
Correct. But: we ship 124 "core" gems in the SCL (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rel-eng/tito.props#L176) and only  these three have foremandist in their release. Makes me wonder why. Are the other gems not dependant on the Ruby version in the SCL?

evgeni:
Looking at commits like https://github.com/theforeman/foreman-packaging/commit/0935aa36cce7f6fcd94ddd46eb432b1a3283c4ae makes me think they are?

ekohl:
Yes, but I think in practice we don't really cherry pick those so haven't ran into it. Perhaps here we have.

evgeni:
obsoleted by https://github.com/theforeman/foreman-packaging/pull/3211

PR #3202, opened 2018-11-16
Release 1.20.0

ekohl:


evgeni:
I assume the repoclosure error is due to missing SCLs?

ekohl:
@evgeni split it into a separate commit. The failure is probably because of the broken repoclosure command in obal.

PR #3207, opened 2018-11-18
Add nodejs-react-ace package

ohadlevy:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---


ohadlevy:
any idea why the tests fails? thanks

ekohl:
[test rpm]

packages/foreman/nodejs-react-ace/nodejs-react-ace.spec
```diff
@@ -0,0 +1,66 @@
...
+Release: 1%{?dist}
+Summary: A react component for Ace Editor
+License: MIT
+Group: Development/Libraries
+URL: https://github.com/securingsincity/react-ace#readme
+Source0: https://registry.npmjs.org/%{npm_name}/-/%{npm_name}-%{version}.tgz
+BuildRequires: nodejs-packaging
+Requires: npm(brace) >= 0.11.1
+Requires: npm(brace) < 1.0.0
+Requires: npm(diff-match-patch) >= 1.0.4
```
ekohl: This and `lodash.{get,isequal}` are not packaged so it'd fail to install. You can either package them as well or bundle them. Since they themselves have no dependencies I'd have a slight preference for packaging them too.

ohadlevy: @ekohl I've tried to address your concerns, it seems that upgrading brace was also required, but I'm not sure as it still fails, do you mind having another look? thanks!

ekohl: I'm not exactly sure, but it looks like brace is being packaged in the bundled method. It doesn't have any dependencies so that could be simplified by using the single method. If you want, I can take it over.

ohadlevy: I would be very happy if you can :) the reason I've bundled brace is due to the fact it failed on missing dependencies on an earlier version (e.g. http://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-rpm-pr-test/detail/foreman-packaging-rpm-pr-test/2500/pipeline#step-169-log-64). please close the PR once you have an alternative - thanks :)

PR #3227, opened 2018-11-22
set the prerelease of katello to master

evgeni:
This technically breaks upgrades between nightlies, but I am not sure we care? :)

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -3,12 +3,12 @@
 
 %global homedir %{_datarootdir}/%{name}
 %global confdir common
-# %%global prerelease .rc1
-%global release 6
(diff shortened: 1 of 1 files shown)


ekohl:
Weren't we waiting for 3.10 to do this?

evgeni:
Why? The current nightlies represent the state of "to be 3.10", no?

ekohl:
Not sure we actually clean nightly directories and we decided to avoid this problem by only applying this change until after 3.10 was branched. So technically we'd start doing this on 3.11.

evgeni:
oh ok, shall I close this then for now?

PR #3259, opened 2018-12-02
Update fog-xenserver to 1.0.0

mmoll:
This should be built into:

* [x] Nightly
* [x] 1.20
* [x] 1.19
* [x] 1.18


mmoll:
Once this is reviewed, I'll open another PR with an updated foreman-xen, please don't merge before that, as repoclosure would break otherwise.

ekohl:
I've opened https://github.com/theforeman/foreman-packaging/pull/3278 to update the spec file itself as well. I'm a bit hesitant to update this in released versions since it requires signing by the release manager as well before they're available.

mmoll:
@ekohl I see. No objections against regenerating of course, but an update to at least 1.20 would be nice as (AFAICT) the new foremn-xen release (which also requires the updated fog-xenserver) would fix some problems for users...

mmoll:
@ekohl What's the status here (and/or over at GH-3278)?

ekohl:
I've merged GH-3278 but not cherry picked it yet. Right now I'm working on the 1.21 release but can have a look later.

PR #3266, opened 2018-12-03
Fixes #25517 - run locale:pack twice

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.20
* [ ] 1.19
* [ ] 1.18


mmoll:
@tbrisker @ekohl please check this :point_up:

While I have no idea about the real cause, this seems to fix it.

mmoll:
> Perhaps we just need to run locale:pack and assets:precompile in separate steps?

That seems to be enough :tada:

PR #3277, opened 2018-12-06
Release 1.20.1

ekohl:


ekohl:
Repoclosure is failing because it's checking the SCL RPMs against a non-SCL config.

PR #3281, opened 2018-12-08
downgrade foreman-proxy-journald to Suggests

mmoll:
this should be built into:

* [x] Nightly


mmoll:
@ekohl I think this should fix the current puppet-foreman_proxy beaker tests.

PR #3283, opened 2018-12-10
split out journald loading to foreman-proxy-journald package

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.20


mmoll:
@ekohl next try...

debian/bionic/foreman-proxy/foreman-proxy.install
```diff
@@ -0,0 +1,22 @@
+foreman-proxy                             etc
+bin                                       usr/share/foreman-proxy
+bundler.d/bmc.rb                          usr/share/foreman-proxy/bundler.d
+bundler.d/dhcp_isc.rb                     usr/share/foreman-proxy/bundler.d
+bundler.d/krb5.rb                         usr/share/foreman-proxy/bundler.d
+bundler.d/libvirt.rb                      usr/share/foreman-proxy/bundler.d
+bundler.d/puppetca_token_whitelisting.rb  usr/share/foreman-proxy/bundler.d
+bundler.d/puppet.rb                       usr/share/foreman-proxy/bundler.d
```
ekohl: I wonder if we should drop this. It's only used if Puppet < 3 is used and there are no packages for it on Bionic.

debian/stretch/foreman-proxy/foreman-proxy.install
```diff
@@ -0,0 +1,22 @@
+foreman-proxy                             etc
+bin                                       usr/share/foreman-proxy
+bundler.d/bmc.rb                          usr/share/foreman-proxy/bundler.d
+bundler.d/dhcp_isc.rb                     usr/share/foreman-proxy/bundler.d
+bundler.d/krb5.rb                         usr/share/foreman-proxy/bundler.d
+bundler.d/libvirt.rb                      usr/share/foreman-proxy/bundler.d
+bundler.d/puppetca_token_whitelisting.rb  usr/share/foreman-proxy/bundler.d
+bundler.d/puppet.rb                       usr/share/foreman-proxy/bundler.d
```
ekohl: Same here as on Bionic

mmoll:
> Looks like this already gives the benefit that it no longer ships development and test dependencies.

That was already the case before this change, I removed the respective command from the `rules` file.

debian/bionic/foreman-proxy/foreman-proxy.install
mmoll: I guess there might be people out there you built funky special setups to stay on the old version, so I'd rather leave this as is for now.

PR #3289, opened 2018-12-11
Rebuild foreman_ansible for webpack

ekohl:


evgeni:
ACK on the diff (katello was already rebuilt in https://github.com/theforeman/foreman-packaging/commit/f7e93aae56033e38515bd2cb8bdd11d4080822a6)

I think ansible needs rex to build, so you'd first have to build rex, wait for it to land in the build repo, and then build ansible (AKA: Jenkins' scratchbuild might fail)

ekohl:
https://github.com/theforeman/foreman-packaging/pull/3290

ehelms:
REX can be dropped from here now, its releasing as I type this

tbrisker:
Still failing, it tries to install the previously released rex version which depends on the 670 hash

ekohl:
[test rpm] Probably because the build was started before the old REX was really built and released

PR #3310, opened 2018-12-14
Bump hammer_cli_katello to 0.17

ekohl:
This matches the commit in master


ekohl:
> Wah wah -- #3292 but we can keep yours since it resets the release

So this is what happens when I focus on my own work and ignore open PRs for a bit :)

PR #3314, opened 2018-12-17
Add more env vars to the foreman service

ekohl:
These are the same ones as used in the old sysvinit script and the installer can be made generic.

### DIFF
--- a/packages/foreman/foreman/foreman.service
+++ b/packages/foreman/foreman/foreman.service
@@ -5,11 +5,11 @@ After=network.target remote-fs.target nss-lookup.target
 
 [Service]
 Type=simple
-User=foreman
+User=$FOREMAN_USER
 TimeoutSec=300
 WorkingDirectory=/usr/share/foreman
-ExecStart=/usr/bin/scl enable tfm "rails server -e $FOREMAN_ENV -p $FOREMAN_PORT"
-Environment=FOREMAN_ENV=production FOREMAN_PORT=3000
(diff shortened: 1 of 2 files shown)


packages/foreman/foreman/foreman.service
```diff
@@ -5,11 +5,11 @@ After=network.target remote-fs.target nss-lookup.target
 
 [Service]
 Type=simple
-User=foreman
+User=$FOREMAN_USER
 TimeoutSec=300
 WorkingDirectory=/usr/share/foreman
-ExecStart=/usr/bin/scl enable tfm "rails server -e $FOREMAN_ENV -p $FOREMAN_PORT"
-Environment=FOREMAN_ENV=production FOREMAN_PORT=3000
+ExecStart=/usr/bin/scl enable tfm "rails server -e $FOREMAN_ENV -p $FOREMAN_PORT -b $FOREMAN_IFACE"
```
ekohl: This is based on https://github.com/theforeman/puppet-foreman/blob/7dfb21677cd1dfe98e397527de2e18b19f8dcf38/templates/foreman.default.erb#L10 but it looks like the installer doesn't set it. I'd be open to to naming it something more correct, like `FOREMAN_BIND`

ehelms: ACK - I like FOREMAN_BIND much better.

ekohl:
Closing in favor of https://github.com/theforeman/foreman-packaging/pull/3315

PR #3315, opened 2018-12-19
Add foreman-service subpackage

ehelms:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---


ehelms:
@ekohl Starter, this could go a lot of ways, but let me know your thoughts.

packages/foreman/foreman/foreman.service
```diff
@@ -5,10 +5,10 @@ After=network.target remote-fs.target nss-lookup.target
 
 [Service]
 Type=simple
-User=foreman
+User=$FOREMAN_USER
 TimeoutSec=300
 WorkingDirectory=/usr/share/foreman
-ExecStart=/usr/bin/scl enable tfm "rails server -e $FOREMAN_ENV -p $FOREMAN_PORT"
+ExecStart=/usr/bin/scl enable tfm "rails server -e $FOREMAN_ENV -p $FOREMAN_PORT -b $FOREMAN_BIND"
 Environment=FOREMAN_ENV=production FOREMAN_PORT=3000
```
ekohl: Don't we need to set defaults here for the new variables? Currently the installer doesn't generate these so users are likely to run into issues. It also lists the sysconfig as optional

packages/foreman/foreman/foreman.spec
```diff
@@ -957,6 +957,21 @@ Meta Package to install requirements for journald logging support
 %files journald
 %{_datadir}/%{name}/bundler.d/journald.rb
 
+%package service
+Summary: Foreman systemd service support
+Group:  Applications/System
+# start specfile service Requires
```
ekohl: My script maps these start/end to bundler.d files. Since no service.rb exists, this will be probably be cleared.

ehelms: What script? Are you wanting me to drop these? or?

packages/foreman/foreman/foreman.service
ehelms: Should we just make sysconfig required? Seems cleaner that way.

ekohl: IMHO we can't rely on the content set in sysconfig. In fact, systemd recommends removing these sysconfig files entirely and moving to systemd drop in files to override values.

packages/foreman/foreman/foreman.spec
ekohl: These comments are used https://github.com/theforeman/foreman-packaging/blob/rpm/develop/update-requirements. In this case you can just drop the comments.

ehelms:
@ekohl This has been updated with suggestions

ehelms:
Attempt at DEB -- https://github.com/theforeman/foreman-packaging/pull/3574

packages/foreman/foreman/foreman.spec
```diff
@@ -981,6 +981,19 @@ Meta Package to install requirements for journald logging support
...
 %{_datadir}/%{name}/bundler.d/journald.rb
 
+%package service
+Summary: Foreman systemd service support
+Group:  Applications/System
+Requires: %{?scl_prefix}rubygem(puma)
+Requires: %{name} = %{version}-%{release}
+
+%description service
+Meta Package to install requirements for Foreman service
```
ekohl: Technically not really a metapackage since it has the service file

ekohl:
Btw, for smart-proxy we ship the unit file in the git repo + tarball so we don't have to duplicate it between packaging. Not a blocker, but bonus points :)

ehelms:
What's the benefit in duplicating it across two code bases?

ekohl:
> What's the benefit in duplicating it across two code bases?

Currently we duplicate it in 2 branches and in Debian packaging in 3 files. Shipping it in foreman.tar.bz2 would allow us to avoid that. For reference:

https://github.com/theforeman/foreman-packaging/blob/82262a1b7f1879a385995bec9226787a6d267af7/packages/foreman/foreman-proxy/foreman-proxy.spec#L104

ehelms:
That logic makes sense. Seems like I should drop it here then?

ekohl:
Currently foreman.git doens't have the unit file, but adding it to extras does make sense. Do note that the sysconfig line is different. foreman-proxy.spec uses sed to add it, on Debian we don't support that and recommend using drop in files. For foreman.service it makes sense to use `/etc/default/foreman` by default to remain compatible with the installer. Sed to the rescue :)

https://github.com/theforeman/foreman-packaging/blob/82262a1b7f1879a385995bec9226787a6d267af7/packages/foreman/foreman-proxy/foreman-proxy.spec#L105-L106

ehelms:
Opened a PR https://github.com/theforeman/foreman/pull/6624 and updated this to reflect that change

PR #3320, opened 2018-12-22
Update foreman_ansible to 2.3.0

mmoll:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19

---


ekohl:
Is this blocked on anything?

plugins/ruby-foreman-ansible/foreman_ansible.rb
```diff
@@ -1 +1,2 @@
-gem 'foreman_ansible', '2.2.9'
+gem 'foreman_ansible', '2.3.0'
+gem 'foreman_ansible_core', '2.2.0'
```
ares: this is no longer dependency of foreman_ansible, it's required by smart_proxy_ansible only

ares: also I may be wrong if it works differently on debian systems, anyway I tried my best and here's the PR for 2.3.1 (new release meanwhile) https://github.com/theforeman/foreman-packaging/pull/3412

PR #3328, opened 2019-01-04
Update rubygem-dynflow to 1.2.0 (including dependencies) (RPM)

iNecas:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

See Foreman's [plugin maintainer documentation](https://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

PR #3329, opened 2019-01-07
Add epoch to changelog

iNecas:
Otherwise the obal lint complains on version comparison.

I've hit issues when updating concurrent-ruby, that is using epoch number (https://github.com/theforeman/foreman-packaging/pull/3328)

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---

### DIFF
--- a/bump_rpm.sh
+++ b/bump_rpm.sh
@@ -54,7 +54,8 @@ if [[ $CURRENT_VERSION != $NEW_VERSION ]] ; then
 		sed -i "s/^\(Release:\s\+\)${RELEASE}/\11/" $SPEC_FILE
 	fi
 
-	$ROOT/add_changelog.sh $SPEC_FILE ${NEW_VERSION}-1 <<-EOF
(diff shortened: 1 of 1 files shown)


bump_rpm.sh
```diff
@@ -48,13 +48,14 @@ if [[ $CURRENT_VERSION != $NEW_VERSION ]] ; then
 	echo "${PACKAGE_NAME}: $CURRENT_VERSION != $NEW_VERSION ; bumping"
 	sed -i "s/^\(Version:\s\+\).\+$/\1${NEW_VERSION}/" $SPEC_FILE
 
+	EPOCH=$(rpmspec --srpm -q --queryformat='%{evr}' --undefine=dist $SPEC_FILE | grep -o '^.*:' || :)
```
ekohl: A better solution is to do this after resetting the release. Then the `%{evr}` is exactly the format and you can drop the grep part.

iNecas: Very good idea, updated and looks much better now. Thanks for the suggestion.

PR #3334, opened 2019-01-08
Fixes #25804 - Update rubygem-gettext_i18n_rails to 1.8.0

ezr-ondrej:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19

---


ekohl:
ok to test

ezr-ondrej:
Probably will be better, sorry, kind of habbit - Refs in my opinion should have link the commit with Redmine, but not close the redmine issue as Fixes

ekohl:
That is indeed the difference. In this case the issue reads as just a packaging issue so doesn't this exactly cover that?

The real reason it wasn't linked is that the commit message itself doesn't have the `Refs` nor `Fixes` part. The bot ignores the actual PR title, but I didn't see that previously. This is true for all repositories where the bot is active, but here a Redmine issue isn't required so it doesn't complain.

ares:
So what's the status here, does the commit message need to be changed? All seems green

ezr-ondrej:
I have amended the commit to have the Fixes #25804, I thought is not necessary as the squash commit would have done it, but I guess is better to be there right away.

PR #3336, opened 2019-01-08
Make bump_rpm.sh more generic

iNecas:
After this changes, it can be used against rails-packaging repo as
well.

Tested with https://github.com/theforeman/rails-packaging/pull/151

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---

### DIFF
--- a/bump_rpm.sh
+++ b/bump_rpm.sh
@@ -8,9 +8,9 @@ elif [[ ! -d $1 ]] ; then
 	exit 1
 fi
 
-cd $1
+SCRIPT_DIR=$(dirname $(readlink -f $BASH_SOURCE))
 
-ROOT=$(git rev-parse --show-toplevel)
+cd $1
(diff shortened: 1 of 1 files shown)


bump_rpm.sh
```diff
@@ -30,7 +30,7 @@ ensure_program() {
 }
 
 if [[ -z $2 ]] ; then
-	if [[ $PACKAGE_NAME == rubygem-* ]] || [[ $PACKAGE_NAME == tfm-rubygem-* ]]; then
+	if echo $PACKAGE_NAME | grep -q rubygem; then
```
ekohl: You could also keep it pure bash by changing `tfm-rubygem-*` to `*-rubygem-*`.

iNecas: Good idea, I've also noticed we have this condition in some other location, so fixed it there as well

PR #3338, opened 2019-01-09
Add NPM packages needed for core

ekohl:


ekohl:
I've omitted babel-plugin-dynamic-import-node for now because I'm not 100% sure it's actually needed. https://github.com/theforeman/foreman/blob/1426e39272146949f0fc4e6a02d76811074d054f/.babelrc#L13-L17 makes me suspect it's only needed in test but I'm hoping @ohadlevy or @sharvit can confirm/deny this.

sharvit:
@ekohl `babel-plugin-dynamic-import-node` only needed to build/test.

ekohl:
> @ekohl `babel-plugin-dynamic-import-node` only needed to build/test.

If it's needed to build, then we must include it. However, as I read it then it's only included in the test env. Will babel ignore this is `NODE_ENV=production`?

sharvit:
> > @ekohl `babel-plugin-dynamic-import-node` only needed to build/test.
> 
> If it's needed to build, then we must include it. However, as I read it then it's only included in the test env. Will babel ignore this is `NODE_ENV=production`?

Mmm, my bad, it's only there for testing env:
https://github.com/theforeman/foreman/blob/96cf840441438d4de602bbe41d0c5dbf9898f06f/.babelrc#L13-L16

Using `NODE_ENV=production` won't use it, instead it will use the `babel-plugin-syntax-dynamic-import`:
https://github.com/theforeman/foreman/blob/96cf840441438d4de602bbe41d0c5dbf9898f06f/.babelrc#L11

btw, babel reads the env var `BABEL_ENV` and if it's not exists it will use the `NODE_ENV`.

ekohl:
Thanks for that clarification.

PR #3343, opened 2019-01-09
Add nodejs-{angular,ngreact} packages

ekohl:


ehelms:
What is the reason for packaging nodejs-angular when Bastion provides the hardened version of Angular we want to be using in production?

ekohl:
Don't ask me. It was added in https://github.com/Katello/katello/commit/c9ad1a640dca56531a592f026e5d1f9d2a686106. It looks like they now also import angular into react.

rel-eng/tito.props
```diff
@@ -23,6 +23,7 @@ whitelist = foreman-bootloaders-redhat
   foreman-release
   foreman-release-scl
   foreman-selinux
+  nodejs-angular
```
ehelms: Shouldn't these be in the katello-nightly-rhel7 whitelist?

ekohl: Probably why the build failed

ekohl:
Updated

ehelms:
According to the output:


```
    package: nodejs-ngreact-0.5.1-1.el7.noarch from downloaded_rpms

      unresolved deps:

         npm(react-dom) < 0:0.15

         npm(react) < 0:0.15

         npm(ngreact) < 0:0.4
```

We use react 16.4+ and it is odd that it requires itself at a version less than itself.

ehelms:
Where are those dependencies even coming from? I see 0 listed:

https://www.npmjs.com/package/ngreact

ekohl:
It looks like those come from `nodejs-packaging`'s autorequire. However, it's interpreting `>= 0.14.0` in a very weird way. I also can't explain the `npm(ngreact) < 0:0.4`. Still investigating.

PR #3347, opened 2019-01-11
Update git to 1.5.0

ares:

* [x] Nightly
* [ ] 1.20
* [ ] 1.19

---


packages/plugins/rubygem-git/rubygem-git.spec
```diff
@@ -37,13 +37,20 @@ cp -a .%{gem_dir}/* \
 %files
 %defattr(-, root, root, -)
 %dir %{gem_instdir}
-%doc %{gem_instdir}/README
+%doc %{gem_instdir}/README.md
+%doc %{gem_instdir}/CHANGELOG.md
+%doc %{gem_instdir}/CONTRIBUTING.md
+%doc %{gem_instdir}/LICENSE
```
ekohl: Could you use `%license`?

packages/plugins/rubygem-git/rubygem-git.spec
```diff
@@ -5,8 +5,8 @@
 
 Summary:        A package for using Git in Ruby code
 Name:           %{?scl_prefix}rubygem-%{gem_name}
-Version:        1.2.5
-Release:        9%{?dist}
+Version:        1.5.0
+Release:        1%{?dist}
 Group:          Development/Languages
 License:        MIT
 URL:            http://rubyforge.org/projects/git/
```
ekohl: Looks like this URL is dead. Could you update it as well? And the Source0 link also doesn't look up to date anymore.

ares:
> The requirement git 1.6.0.0, or greater suggests git should be installed. Is this true?

I think so, but that I believe was always true, I added that as dependency. Also other things fixed as suggested.

PR #3356, opened 2019-01-14
Update rubygem-foreman_remote_execution to 1.7.0 (RPM)

iNecas:
Includes rubygem-foreman_remote_execution_core to 1.1.5

* [x] Nightly
* [ ] 1.20
* [ ] 1.19
* [ ] 1.18

---


ares:
@iNecas can we also push this down to 1.21 and potentially 1.20? New foreman_ansible now depends on 1.7.0 and I'd like it to be available at least for 1.21 users (ideally to 1.20 too). REX 1.7.0 seems 1.20 compatible https://github.com/theforeman/foreman_remote_execution/blob/master/lib/foreman_remote_execution/engine.rb#L37

iNecas:
Rex 1.7.0 is 1.21 only so I'm ok with getting it to 1.21: by the time this PR has been opened, there was no 1.21 yet, so I hoped it would get there automatically (@evgeni could you help us with that?).


Rex 1.7.0 is not suitable for 1.20 though

PR #3417, opened 2019-01-29
Update rubygem-foreman_scap_client to 0.4.1

ares:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---


packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
```diff
@@ -25,6 +25,10 @@ BuildRequires: ruby(abi)
 %else
 BuildRequires: ruby(release)
 %endif
+%if 0%{?el6}
+BuildRequires: rubygem(json) >= 1.0
```
evgeni: I think this is rather a runtime dependency, so a `Requires` instead of `BuildRequires`? Also there are already EL6 branches in the spec above, so this can go into there?

evgeni: the [gemspec says `~> 1.4`](https://github.com/theforeman/foreman_scap_client/blob/master/foreman_scap_client.gemspec#L27), so it should be `>= 1.4` here?

ares: updated

PR #3120, opened 2018-10-20
Add smart_proxy_rex_core dep to smart_proxy_ansible

dLobatog:
In order to run Ansible jobs you would need this dependency. At the
moment users may choose to only install smart_proxy_ansible and jobs
will fail to run, as you can see in
https://community.theforeman.org/t/run-ansible-role-error/11501/2

For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [x] 1.20
* [x] 1.19
* [x] 1.18

See Foreman's [plugin maintainer documentation](https://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

PR #3440, opened 2019-02-11
Fixes #24586 - fix foreman_ansible dependencies

iNecas:
foreman_ansible_core is listed in dependencies in the gemspec, but not
in the spec.

* [x] Nightly
* [x] 1.21
* [ ] 1.20

---

### DIFF
--- a/packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
+++ b/packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
@@ -20,6 +20,8 @@ Requires: %{?scl_prefix_ruby}rubygems
 Requires: %{?scl_prefix}rubygem(deface) < 2.0
 Requires: %{?scl_prefix}rubygem(foreman-tasks) >= 0.8
 Requires: %{?scl_prefix}rubygem(foreman-tasks) < 1.0
+Requires: %{?scl_prefix}rubygem(foreman_ansible_core) >= 2.0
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -30,7 +32,7 @@ BuildRequires: %{?scl_prefix_ruby}rubygems
 BuildRequires: %{?scl_prefix}rubygem(deface) < 2.0
 BuildRequires: %{?scl_prefix}rubygem(foreman-tasks) >= 0.8
 BuildRequires: %{?scl_prefix}rubygem(foreman-tasks) < 1.0
-BuildRequires: %{?scl_prefix}rubygem(foreman_ansible_core) >= 2.0.2
+BuildRequires: %{?scl_prefix}rubygem(foreman_ansible_core) >= 2.0
```
ekohl: I think we can actually really drop this. It's now only a dev dependency so the tests can run together. https://github.com/theforeman/foreman-packaging/pull/3120 might still be good to finish though.

iNecas:
Seems like the issue was this https://github.com/theforeman/foreman_ansible/commit/6a78d7e62b0654b813d66b7a8d70935d13702f0d, and should be fixed in ansible 2.3.1, so we whould pull it to 1.21 (we have 2.3.0 now)

PR #3449, opened 2019-02-12
Rebuild rubygem-foreman_remote_execution for webpack

ekohl:


evgeni:
[test rpm]

evgeni:
I wonder if we should keep the "needs review" workflow for plain rebuilds. There is no real benefit in getting a review here. And we've not been doing reviews for CPs to stable branches anyways.

ekohl:
That's fair. I did send this in because I wanted to be sure it actually built. When I sent it in, I also saw there was a react 16.8 update which would also change the hash so I held off on merging it.

PR #3460, opened 2019-02-18
Split foreman-webpack-helper

sbernhard:
This fixes a issue that you need to build foreman first in order to be able to install foreman-build which ships the foreman.attr / foreman.provreq which sets the webpack js provide/requires. 

Means, to build a foreman which provides the webpack js header you need to do the following, which is pretty unsightly
- build foreman
- install foreman-build
- build foreman again


packages/foreman/foreman-webpack-helper/foreman-webpack-helper.spec
```diff
@@ -0,0 +1,40 @@
+%global release 1
+%global prerelease develop
+
+Name:    foreman-webpack-helper
```
ekohl: Some bikeshedding: I think `foreman-packaging` is a more common name for this.

packages/foreman/foreman-webpack-helper/foreman-webpack-helper.spec
```diff
@@ -0,0 +1,40 @@
+%global release 1
+%global prerelease develop
```
ekohl: We use this for nightly packages but this will probably be rather stable so you can drop this

packages/foreman/foreman-webpack-helper/foreman-webpack-helper.spec
```diff
@@ -0,0 +1,40 @@
...
+
+Name:    foreman-webpack-helper
+Version: 1.0
+Release: %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?nightly}%{?dist}
```
ekohl: Same as the `prerelease` comment. You can simply use `Release: 1`

tbrisker:
@sharvit - you might be interested in this as well

PR #3461, opened 2019-02-19
Fixes #26084 - Restart the passenger service again

ekohl:


ekohl:
We should rebuild all plugins to benefit from the updated macro, right?

evgeni:
Technically, yes. Or we wait for the next plugin update.

ekohl:
I've been thinking about a round of plugin rebuilds before every major release to see if they still build. I think it'd be good to do this right before branching.

PR #3500, opened 2019-03-08
Fixes #26339 - Update foreman-installer dependencies

ekohl:
Incomplete, will require a new kafo release as well. Since that's a build dependency that will be a separate PR.


ekohl:
[test rpm]

ekohl:
Looks like we don't have Puppet 5.5.8+ in our build root. @ehelms can we update our copy of `puppetlabs-puppet5-rhel-7` on Koji or should we go ahead an introduce `puppetlabs-puppet6-rhel-7`? The latter is probably better.

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -14,18 +14,18 @@ Source0:    https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?prerel
...
-Requires:   %{?scl_prefix}rubygem-kafo >= 1.0.5
+Requires:   %{?scl_prefix}puppet-agent >= 5.5.8
+Requires:   %{?scl_prefix}rubygem-kafo >= 3.0.0
 Requires:   foreman-selinux
 Requires:   %{?scl_prefix_ruby}ruby(release)
 Requires:   %{?scl_prefix}rubygem-highline
 
 BuildRequires: asciidoc
 BuildRequires: rubygem(rake)
 BuildRequires: %{?scl_prefix}puppet-agent >= 1.9.0
```
ekohl: I've kept this older since it can still build with older versions. Ideally we'd update this as well when our koji has a newer version.

ekohl:
[test rpm] now that I've mashed to staging (which will include kafo 3.0).

PR #3519, opened 2019-03-13
Update foreman-installer dependencies

ekohl:
We now depend on kafo >= 3.0.0 and Puppet 5.5.8.

https://community.theforeman.org/t/dropping-puppet-4-support-in-our-installer/13029


mmoll:
merged, bedankt @ekohl!

PR #3534, opened 2019-03-19
Fixes #26401 - Remove GPG key from f-release-client

ekohl:
This file belongs to foreman-release and can create conflicts on upgrade.

### DIFF
--- a/packages/foreman/foreman-release/foreman-client.repo
+++ b/packages/foreman/foreman-release/foreman-client.repo
@@ -3,11 +3,11 @@ name=Foreman client nightly
 baseurl=https://yum.theforeman.org/client/nightly/$DIST/$basearch
 enabled=1
 gpgcheck=0
-gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman-client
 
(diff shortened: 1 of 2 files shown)


packages/foreman/foreman-release/foreman-client.repo
```diff
@@ -3,11 +3,11 @@ name=Foreman client nightly
 baseurl=https://yum.theforeman.org/client/nightly/$DIST/$basearch
 enabled=1
 gpgcheck=0
-gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman-client
```
evgeni: where is this file coming from? I don't see it in `%files`

ekohl: The repo is currently not signed (hence the `gpgcheck=0`). I just thought I should rename this to make it clear it's a different file. Users have been setting gpgcheck=1.

evgeni: oooh, I'll be damned for not seeing the `gpgcheck=0` above.

ekohl: See https://projects.theforeman.org/issues/26278 as well. I also consider leaving the line out.

PR #3555, opened 2019-03-27
Update foreman dependencies

tbrisker:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---


ehelms:
Bit of a chicken and egg with the develo-release job so merging

packages/foreman/foreman/foreman.spec
```diff
@@ -554,26 +554,6 @@ Meta package to install requirements for oVirt compute resource support.
 %files ovirt
 %{_datadir}/%{name}/bundler.d/ovirt.rb
 
-%package compute
```
ekohl: Removing this does break the foreman_docker and foreman_one packages. It's also in the `foreman-ovirt` subpackage requirement so that won't pass repoclosure either.

PR #3557, opened 2019-03-27
Remove foreman-compute

ekohl:
We no longer have a bundler.d/fog.rb so this package no longer makes sense.


mmoll:
merged, bedankt @ekohl!

PR #3565, opened 2019-03-28
drop foreman-compute require from foreman-one

evgeni:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---

### DIFF
--- a/packages/plugins/rubygem-foreman_one/rubygem-foreman_one.spec
+++ b/packages/plugins/rubygem-foreman_one/rubygem-foreman_one.spec
@@ -9,7 +9,7 @@
 Summary:    Provision and manage OpenNebula VMs from Foreman
 Name:       %{?scl_prefix}rubygem-%{gem_name}
 Version:    0.4
-Release:    3%{?foremandist}%{?dist}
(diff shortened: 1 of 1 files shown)


ekohl:
Does this even work? 0.4 is from 2015.

evgeni:
:see_no_evil: I don't know

dgoetz:
I do not think so and because of our Managed Service team moving to Openstack, I think there will be no development from our site any longer. :cry:

evgeni:
@dgoetz time to archive the source repo and remote the rpm then?

dgoetz:
If it causes problems, yes. I already mentioned it in the fog upgrade discussion and noone reacted, so I think user base is low to non existing.

ekohl:
Should we sunset it? I think the procedure should be:
* Post to discourse
* Remove from nightly
* Update the README in the git repo pointing to the discourse post
* Archive the git repo

evgeni:
We can also merge this PR and wait till someone complains it's not actually working?

ares:
If we know it does not work, let's set the right expectation. I vote for sunset.

PR #3572, opened 2019-03-28
Remove foreman_one

ares:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---


ekohl:
I think this should also add it to tfm.spec per https://github.com/theforeman/foreman-packaging/tree/rpm/develop#howto-removing-a-package so it's removed from existing installs. We do need to consider the database and not to break the application though.

ares:
yeah we should clean all instances of https://github.com/theforeman/foreman-one/blob/master/app/models/foreman_one/one.rb#L2

I don't know how to do it best, we either need foreman-maintain working in upstream already or have a cleanup task in core, so people could do cleanups without plugins installed. Or in this case, we could have something in core that cleans up unknown STIed compute resources.

ares:
Anyway, it should most likely not block this PR to get in. Why this should be obsoleted by foreman package though? Foreman didn't depend on it right?

evgeni:
Not foreman, tfm. That's the SCL package that had the plugin and doesn't anymore.

ares:
ok, added the Obsolete to tfm and also removed opennebula gem

PR #3580, opened 2019-03-29
Obsolete foreman-compute

tbrisker:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -8,7 +8,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 8
+%global release 9
 %global prerelease develop
 
 Name:    foreman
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman/foreman.spec
```diff
@@ -34,6 +34,8 @@ BuildArch:  noarch
 Conflicts: foreman-tasks < 0.11.0-2
 Conflicts: foreman-release-scl < 7-1
 
+Obsoletes: foreman-compute
```
evgeni: Should this be
```suggestion
Obsoletes: foreman-compute < 1.22.0-0.9.develop
```

evgeni: or `< %{version}`

evgeni:
```
10:31 < ewoud> Zhenech: I think your suggestion is correct
10:33 < ewoud> either can work, but I think the latter one is a bit more robust
10:34 < ewoud> the explicit version number does show that at some point you could drop it more easily, but we don't really do that anyway
```

tbrisker: amended. in the future feel free to push to my branch, especially when i'm afk :)

PR #3585, opened 2019-03-29
check repoclosure against the whole repo, not only the new package

evgeni:
also check dependant repos (plugins/katello for core)

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -22,6 +22,11 @@ foreman_scl_packages:
       - koji-foreman
     nightly_releaser: koji-foreman-jenkins
     repoclosure_config: repoclosure/yum_el7.conf
+    repoclosure_target_repos:
+      - downloaded_rpms
+      - el7-foreman-nightly
+      - el7-foreman-plugins-nightly
+      - el7-katello-nightly
     repoclosure_lookaside_repos:
       - el7-base
       - el7-updates
@@ -29,7 +34,6 @@ foreman_scl_packages:
       - el7-extras
       - el7-scl
       - el7-foreman-rails-nightly
-      - el7-foreman-nightly
   hosts:
     foreman:
       releasers:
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -22,14 +22,18 @@ foreman_scl_packages:
       - koji-foreman
     nightly_releaser: koji-foreman-jenkins
     repoclosure_config: repoclosure/yum_el7.conf
+    repoclosure_target_repos:
+      - downloaded_rpms
```
ehelms: Previously this was 'hidden' as a nuance of how Obal handled scratch builds. This seems strange to have to surface here? Could Obal just include this as a default lookup location?

evgeni: Obal certainly could. We have configs (see below for repoclosure only examples) where we don't have a downloaded_rpms folder. What would happen if we still add it in that case. We could make obal only add it if present?

package_manifest.yaml
```diff
@@ -200,13 +204,17 @@ foreman_nonscl_packages:
...
+      - el7-foreman-nightly
+      - el7-foreman-plugins-nightly
```
ekohl: Does this mean we no longer check repoclosure for nonscl without plugins present? Should we have a separate run for that so you never break repoclosure on nightly, but can break plugins and katello. For example when you're doing the fog migration.

package_manifest.yaml
evgeni: Implemented in https://github.com/theforeman/obal/pull/186

package_manifest.yaml
evgeni: This is a good point. My intention here was "did a change in nonscl break plugins?", but this obviously also makes plugins resolve the deps...

Why are we always finding more layers of complexity in this?

Anyways, this PR will be closed until we sort out https://github.com/theforeman/obal/pull/186

evgeni:
closing until https://github.com/theforeman/obal/pull/186 is finished

PR #3594, opened 2019-04-02
Add foreman service to run as puma for DEB

ehelms:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---

### DIFF
--- a/debian/bionic/foreman/control
+++ b/debian/bionic/foreman/control
@@ -17,7 +17,7 @@ Priority: extra
 Depends: ruby, ruby2.5, bundler (>= 1.3), zlib1g-dev,
          ruby2.5-dev, build-essential,
          rake (>=0.8.4), ${misc:Depends}
-Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal
+Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal, foreman-service
 Conflicts: ruby-activerecord-deprecated-finders, ruby-foreman-tasks (<< 0.10.1), ruby-sequel-pg
 Description: Systems management web interface
  Foreman is aimed to be a single address for all machines life cycle management.
@@ -236,6 +236,17 @@ Priority: extra
 Depends: foreman (= ${binary:Version}), libsystemd-dev, ${misc:Depends}
 Description: metapackage providing journald dependencies for Foreman
  This package provides journald logging dependencies for Foreman,
+ a flexible systems management web application.
+ .
+ The foreman package should be installed along with this package.
+
+Package: foreman-service
+Architecture: all
+Section: web
+Priority: extra
+Depends: foreman (= ${binary:Version}), ${misc:Depends}
+Description: package providing Foreman services running via puma
+ This package provides Foreman systemd service for Foreman,
  a flexible systems management web application.
  .
  The foreman package should be installed along with this package.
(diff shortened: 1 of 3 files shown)


debian/bionic/foreman/control
```diff
@@ -239,3 +239,14 @@ Description: metapackage providing journald dependencies for Foreman
  a flexible systems management web application.
  .
  The foreman package should be installed along with this package.
+
+Package: foreman-service
+Architecture: all
+Section: web
+Priority: extra
+Depends: foreman (= ${binary:Version}), ruby-puma, libsystemd-dev, ${misc:Depends}
```
evgeni: why does it depend on `libsystemd-dev`?

debian/bionic/foreman/control
```diff
@@ -239,3 +239,14 @@ Description: metapackage providing journald dependencies for Foreman
...
+Description: package providing Foreman services running via puma
+ This package provides Foreman systemd service for Foreman,
+ a flexible systems management web application.
+ .
+ The foreman package should be installed along with this package.
```
evgeni: you depend on it, so this seems rather given?

debian/bionic/foreman/control
ehelms: No idea.

debian/bionic/foreman/control
ehelms: Just copy pasting and keeping things in line with others.

debian/bionic/foreman/control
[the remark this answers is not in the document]
evgeni: There *is* https://packages.debian.org/search?keywords=puma, but we might or might not want to use that.

evgeni: why did you add it? :)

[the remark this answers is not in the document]
mmoll: AFAICT, we don't want to use this, but (like the dynflowd thingies) use the one coming with Foreman's bundled gems.

ehelms: Copy/pastin!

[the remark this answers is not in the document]
ehelms: Soooo, does all that translate to 'remove ruby-puma' here? Or more than that needed?

ehelms:
I made some updates, probably still missing a few things. Treat me like I am completely new at this if more changes are needed. Spell it out for me :)

debian/bionic/foreman/control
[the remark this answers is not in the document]
mmoll: Exactly, removing it as dependency. Also, akeen to `dynflowd.service` in DEB packaging, at the end the puma process needs to get started via bundle.

evgeni:
Can you call your branch `deb/develop-foreman-puma` so that jenkins properly fetches the tarball from jenkins and not from downloads?

evgeni:
As the service file is now in the tarball, there is nothing that would install it into the .deb (contrary to the case where it's inside `debian/`, where some magic happen).

You'll need to add a line like this: https://github.com/theforeman/foreman-packaging/blob/deb/develop/debian/stretch/foreman/foreman-test.install with `/usr/lib/systemd/system` as target.

One thing that makes me think: We also ship a traditional `foreman.init` file for SysV init, which will be used by systemd if there is no `foreman.service`. Should we drop that now?

ekohl:
> Should we drop that now?

I think I heard some users mention they use sysvinit only, but I'd be fine with dropping it. Probably good to write up something for the release notes about what the service is and how it's supported though.

ehelms:
Closing in favor of https://github.com/theforeman/foreman-packaging/pull/3617

PR #3604, opened 2019-04-05
Drop tests package from debian builds

tbrisker:
Tests aren't expected to run in production anyways, no point in
packaging them. We don't package them in RPMS as well.

(cherry picked from commit c298716baf118c0a3b4dee994b32a2ce33891ed5)

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---


ekohl:
I'm not in favor of dropping a package from a released version since I suspect the old version will remain available in the archive and cause weird issues. @mmoll @evgeni probably know more.

tbrisker:
ok, we'll go with 1.21.2 to fix this.

PR #3617, opened 2019-04-11
Add foreman service to run as puma for DEB

ehelms:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---

### DIFF
--- a/debian/bionic/foreman/control
+++ b/debian/bionic/foreman/control
@@ -17,7 +17,7 @@ Priority: extra
 Depends: ruby, ruby2.5, bundler (>= 1.3), zlib1g-dev,
          ruby2.5-dev, build-essential,
          rake (>=0.8.4), ${misc:Depends}
-Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal
+Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal, foreman-service
 Conflicts: ruby-activerecord-deprecated-finders, ruby-foreman-tasks (<< 0.10.1), ruby-sequel-pg
 Description: Systems management web interface
  Foreman is aimed to be a single address for all machines life cycle management.
@@ -236,6 +236,17 @@ Priority: extra
 Depends: foreman (= ${binary:Version}), libsystemd-dev, ${misc:Depends}
 Description: metapackage providing journald dependencies for Foreman
  This package provides journald logging dependencies for Foreman,
+ a flexible systems management web application.
+ .
+ The foreman package should be installed along with this package.
+
+Package: foreman-service
+Architecture: all
+Section: web
+Priority: extra
+Depends: foreman, ${misc:Depends}
+Description: package providing Foreman services running via puma
+ This package provides Foreman systemd service for Foreman,
  a flexible systems management web application.
  .
  The foreman package should be installed along with this package.
--- a/debian/bionic/foreman/foreman-service.install
+++ b/debian/bionic/foreman/foreman-service.install
@@ -1,0 +1,1 @@
+bundler.d/service.rb /usr/share/foreman/bundler.d
--- a/debian/stretch/foreman/control
+++ b/debian/stretch/foreman/control
@@ -16,7 +16,7 @@ Priority: extra
 Depends: ruby, ruby2.3, bundler (>= 1.3), zlib1g-dev,
          ruby2.3-dev, build-essential,
          rake (>=0.8.4), ${misc:Depends}
-Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal
+Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal, foreman-service
 Conflicts: ruby-activerecord-deprecated-finders, ruby-foreman-tasks (<< 0.10.1), ruby-sequel-pg
 Description: Systems management web interface
  Foreman is aimed to be a single address for all machines life cycle management.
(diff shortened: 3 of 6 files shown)


debian/bionic/foreman/control
```diff
@@ -17,7 +17,7 @@ Priority: extra
 Depends: ruby, ruby2.5, bundler (>= 1.3), zlib1g-dev,
          ruby2.5-dev, build-essential,
          rake (>=0.8.4), ${misc:Depends}
-Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal
+Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal, foreman-service
```
ekohl: You can add the exact version like `foreman-debug`

evgeni: Actually, I don't see a point in recommending the exact same version. It'll work with most other versions too.

ekohl: Fair point. In that case, the dependency should also be relaxed to just foreman?

ehelms:
Not sure what my next move here is

ekohl:
@ehelms I think you still need to make sure puma is actually present. AFAIK this is typically done by dropping a file in bundler.d.

To test this now:
```bash
apt install -y wget gpg2
wget -O - https://deb.theforeman.org/pubkey.gpg | apt-key add -
cat > /etc/apt/sources.list.d/foreman.list <<EOF
deb http://deb.theforeman.org/ stretch nightly
deb http://stagingdeb.theforeman.org/ stretch ehelms-nightly
EOF
# Needed because we generate inconsistent package names - we need to install the Jenkins timestamp plugin again
cat > /etc/apt/preferences.d/foreman-service <<EOF
Package: foreman*
Pin: version 9999-stretch+scratchbuild+20190411002447
Pin-Priority: 999
EOF
apt update
apt install foreman-service
```

This gives me:
```console
root@debian9:~# systemctl start foreman
root@debian9:~# systemctl status foreman
● foreman.service - Foreman
   Loaded: loaded (/usr/lib/systemd/system/foreman.service; disabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Fri 2019-04-12 13:44:10 UTC; 35s ago
     Docs: https://theforeman.org
  Process: 22018 ExecStart=/usr/bin/scl enable tfm rails server --environment $FOREMAN_ENV --port $FOREMAN_PORT --bind
 Main PID: 22018 (code=exited, status=203/EXEC)

Apr 12 13:44:10 debian9 systemd[1]: Started Foreman.
Apr 12 13:44:10 debian9 systemd[1]: foreman.service: Main process exited, code=exited, status=203/EXEC
Apr 12 13:44:10 debian9 systemd[1]: foreman.service: Unit entered failed state.
Apr 12 13:44:10 debian9 systemd[1]: foreman.service: Failed with result 'exit-code'.
```

The problem is the SCL and that's a mistake we made in core which will also break on EL8 if we build without SCLs. We should probably use `/usr/bin/foreman-ruby` in some way.

ekohl:
> ```shell
> # Needed because we generate inconsistent package names - we need to install the Jenkins timestamp plugin again
> ```

https://github.com/theforeman/foreman-infra/pull/1022

debian/bionic/foreman/control
evgeni: Yeah, I think that's right.

debian/bionic/foreman/control
```diff
@@ -239,3 +239,14 @@ Description: metapackage providing journald dependencies for Foreman
  a flexible systems management web application.
  .
  The foreman package should be installed along with this package.
+
+Package: foreman-service
+Architecture: all
+Section: web
+Priority: extra
+Depends: foreman (= ${binary:Version}), ${misc:Depends}
```
evgeni: As discussed above, the dependency probably does not need to be version strict here.

debian/bionic/foreman/foreman-service.install
```diff
@@ -0,0 +1 @@
+extras/systemd/foreman.service /usr/lib/systemd/system
```
evgeni: turns out, this needs to be `/lib`, not `/usr/lib` for some magic scripts in debian to work :woman_facepalming:

evgeni:
- we should also move the init script (and `foreman.default`?) into this package (or drop it for good), as otherwise systemd will still pick it up and generate a dummy .service from it, thus rendering this package useless
- I am not sure if we need to do some more magic like https://github.com/theforeman/foreman-packaging/blob/5835bbbf9d34fc46f4b5825c6daad0da136dc8d2/debian/stretch/foreman/rules#L36 too -- we will find out after we moved the `.service` file to `/lib`

ehelms:
A few updates pushed. The `.service` file now needs a `sed` somewhere I think to change the start command? 

https://github.com/theforeman/foreman/blob/develop/extras/systemd/foreman.service#L11

ekohl:
Perhaps this should also be `Refs #26628 - ...` as a commit message?

evgeni:
You'll need smth similar to https://github.com/theforeman/foreman-packaging/blob/5a940b1b5d68f337432b10017a44c902fd05c808/debian/stretch/foreman/rules#L23 for the replacement of the `ExecStart` in the `service` file.

Did you drop the `service` file from install intentionally?

My comments from monday still stand :)

evgeni:
Pushed a few changes into https://github.com/theforeman/foreman-packaging/pull/3643, still no idea how to handle the `.init` file nicely.

ekohl:
I wouldn't mind dropping the sysvinit script. Rather appreciate it since we don't really support it.

ehelms:
Closing in favor of https://github.com/theforeman/foreman-packaging/pull/3643

PR #3624, opened 2019-04-12
use the nightly macro in the ForemanSourceStrategy

evgeni:
previously tito would (via our ForemanSourceStrategy) generate versions
like foreman-1.22.0-0.201903280802git97b75e70.8.develop.el7.src.rpm
because it was prepending `0.<date>git<hash>` to the %release macro

but `obal nightly` uses a special `nightly` macro to inject
`<date>git<hash>` into %release.

because of this difference, different jobs would generate different
version structures, resulting in nightly packages that can't be upgraded
because ForemanSourceStrategy generated a newer version than `obal
nightly`.

this commit adjusts ForemanSourceStrategy to also use the same method as
`obal nightly`, resulting in
foreman-1.22.0-0.8.develop.201903280802git97b75e70.el7.src.rpm for the
above example

### DIFF
--- a/rel-eng/custom/custom.py
+++ b/rel-eng/custom/custom.py
@@ -88,7 +88,7 @@ class ForemanSourceStrategy(SourceStrategy):
         rel_date = datetime.utcnow().strftime("%Y%m%d%H%M")
         self.release = rel_date + gitrev
         print("Building release: %s" % self.release)
-        run_command("sed -i '/^Release:/ s/\s\+/ 0.%s/' %s" % (self.release, self.spec_file))
+        run_command("sed -i '1i %%global nightly .%s/' %s" % (self.release, self.spec_file))
 
     """
     Downloads the source files from Jenkins, from a job that produces them as


evgeni:
@ares this is what have been haunting you for a while I think

@komidore64 can you please have a look here, as you were the one doing the most work on nightly building in the past

rel-eng/custom/custom.py
```diff
@@ -88,7 +88,7 @@ def filter_archives(path):
         rel_date = datetime.utcnow().strftime("%Y%m%d%H%M")
         self.release = rel_date + gitrev
         print("Building release: %s" % self.release)
-        run_command("sed -i '/^Release:/ s/\s\+/ 0.%s/' %s" % (self.release, self.spec_file))
+        run_command("sed -i '1i %%global nightly .%s/' %s" % (self.release, self.spec_file))
```
komidore64: my sed knowledge is rusty. `1i` means "insert at line 1" ?

evgeni: Exactly.

komidore64: right on.

komidore64:
:crossed_fingers:

PR #3643, opened 2019-04-17
Deb/develop foreman service

evgeni:
testing, ignore

### DIFF
--- a/debian/bionic/foreman/control
+++ b/debian/bionic/foreman/control
@@ -17,7 +17,7 @@ Priority: extra
 Depends: ruby, ruby2.5, bundler (>= 1.3), zlib1g-dev,
          ruby2.5-dev, build-essential,
          rake (>=0.8.4), ${misc:Depends}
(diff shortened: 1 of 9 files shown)


mmoll:
@evgeni this seems to work fine, thanks! However, I think until puma is becoming the default, the -service package should not be recommended.

evgeni:
> @evgeni this seems to work fine, thanks! However, I think until puma is becoming the default, the -service package should not be recommended.

I agree, but then we also should not ship `foreman.init` as enabled-by-default :)

ekohl:
Because the sysvinit script and the systemd unit work in a very different way I'd prefer to drop the sysvinit script. If there's demand from users we can add it again, but I don't expect any interest.

evgeni:
[test packaging]

evgeni:
it's green :green_heart: 

@ekohl I think I agree, but I'd do it in a separate PR

ekohl:
Thanks!

PR #3658, opened 2019-04-23
Fixes #26661 - don't ship fact-plugin on modern Fedora and EL

evgeni:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---

### DIFF
--- a/packages/client/katello-host-tools/katello-host-tools.spec
+++ b/packages/client/katello-host-tools/katello-host-tools.spec
@@ -8,6 +8,7 @@
 
 %global build_agent (0%{?suse_version} == 0) && ((0%{?fedora} > 28) || (0%{?rhel} > 0))
 %global legacy_agent (0%{?rhel} == 5) || (0%{?rhel} == 6)
+%global build_fact_plugin (0%{?rhel} > 0 && 0%{?rhel} <= 7)
 
 %if 0%{?suse_version}
 %define dist suse%{?suse_version}
@@ -15,7 +16,7 @@
 
 Name: katello-host-tools
 Version: 3.5.0
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: A set of commands and yum plugins that support a Katello host
 Group:   Development/Languages
 License: LGPLv2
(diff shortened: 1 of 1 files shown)


packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -8,14 +8,15 @@
 
 %global build_agent (0%{?suse_version} == 0) && ((0%{?fedora} > 28) || (0%{?rhel} > 0))
 %global legacy_agent (0%{?rhel} == 5) || (0%{?rhel} == 6)
+%global build_fact_plugin ((0%{?rhel} > 0 && 0%{?rhel} <= 7) || (0%{?fedora} > 0 && 0%{?fedora} <= 24))
```
ekohl: We don't support Fedora <= 24 anymore so should we leave that out?

evgeni: This sounds like a good idea, yeah.

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -8,14 +8,15 @@
 
 %global build_agent (0%{?suse_version} == 0) && ((0%{?fedora} > 28) || (0%{?rhel} > 0))
 %global legacy_agent (0%{?rhel} == 5) || (0%{?rhel} == 6)
+%global build_fact_plugin (0%{?rhel} > 0 && 0%{?rhel} <= 7)
```
ekohl: After reading this, I do wonder about Suse.

evgeni: I'll leave that as an exercise to the reader, errr, someone who knows more about our SUSE bits.

evgeni: We ship our own sub-man for suse: http://koji.katello.org/koji/packageinfo?packageID=1006, so we should be good?

@jturel might know more

jturel: The subman version we ship for suse doesn't need our workaround - so this is good! :taco:

PR #3660, opened 2019-04-23
Add nodejs-history package

ohadlevy:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---


ekohl:
d1152eafc6f824ecedd7ef2f0eac68e3db0f9b8f removed history.js because it broke Katello. Probably good to investigate this a bit closer but I don't really have time right now.

ohadlevy:
this doesnt soud relevant, as in either case it will be provided by
foreman, so I don't assume any issue with webpack on core.

regardless, I would be in favor of merging this kind of stuff early on the
release cycle so we can quickly fix it instead of leaving the pr's open for
a long duration... the associated pr is block for many other items....

ekohl:
I'm not going to because I'm not going to deal with a broken nightly now. I simply don't have the time and a way behind schedule.

ohadlevy:
On Tue, Apr 23, 2019 at 5:05 PM Ewoud Kohl van Wijngaarden <
notifications@github.com> wrote:

> I'm not going to because I'm not going to deal with a broken nightly now.
> I simply don't have the time and a way behind schedule.
>

We are OK with waiting for you to release 1.22rc's hopefully that would
unblock you? just let us know when...

ekohl:
That's the first big part, but there's more. It's probably best in this case if someone else steps in to unblock development.

ohadlevy:
:green_heart:

PR #3663, opened 2019-04-23
Release 1.21.3

ekohl:


tbrisker:
Looks like this is failing on some installer deps?

ekohl:
Yes, pretty sure it's because our repoclosure process is broken due to mixing of SCL and non-SCL. @evgeni is working on improving this, but it's harder than it looks.

tbrisker:
how do we proceed then?

ekohl:
Just merge it.

PR #3665, opened 2019-04-23
update configs for Foreman 1.22/Katello 3.12

evgeni:


package_manifest.yaml
```diff
@@ -28,15 +28,15 @@ foreman_scl_packages:
...
+      - el7-foreman-rails-1.22
+      - el7-foreman-1.22
   hosts:
     foreman:
       releasers:
         - "{{ nightly_releaser }}"
       build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
       nightly_package_tito_releaser_args:
-        - "--arg jenkins_job=foreman-develop-source-release"
+        - "--arg jenkins_job=foreman-1.22-source-release"
```
evgeni: these jobs don't exist yet, @ehelms is that on your todo?

ekohl: This has the old naming convention: `test_1_22_stable`

evgeni: "oh"

ekohl: I'm not against introducing the new name, but we missed this in the 1.22 cycle.

ehelms: Yea, I was targeting this new convention for 1.23+

ehelms: However, in a release branch, there is no 'nightly' concept right? The source should be from the official hosting site.

ekohl: It's a bit undefined. In the past few releases I updated them and removed them once we had RC1 tarballs.

ekohl: Correction: I kept it in, but changed the releaser once we had tarballs. That means in theory we can do nightly scratch builds of the next RC, but in practice we don't really have this set up.

evgeni: Yeah, I've been following what's in the 1.21 branch

evgeni: So, it seems there are no `test_develop` or `test_1_22` jobs in Jenkins (this is what 1.21 points to).

@ekohl you suggested https://ci.theforeman.org/job/test_1_22_stable, but that does not have artifacts?

ekohl: Every single item in the matrix has an archive. The job picks one of these. This may not be pretty, but it's how it's worked for a long time.

evgeni: "oh²"

evgeni: all updated now

PR #3682, opened 2019-04-26
Regenerate rubygem-smart_proxy_discovery spec

evgeni:

* [x] Nightly
* [ ] 1.22
* [ ] 1.21
* [ ] 1.20

---

### DIFF
--- a/packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
+++ b/packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
@@ -94,8 +94,6 @@ mv %{buildroot}%{gem_instdir}/settings.d/discovery.yml.example \
 %changelog
 * Fri Apr 26 2019 Evgeni Golov 1.0.5-1
 - Update to 1.0.5-1
-
-* Fri Apr 26 2019 Evgeni Golov 1.0.4-2
 - Regenerate RPM spec based on latest template
 
 * Fri Apr 29 2016 Dominic Cleal <dominic@cleal.org> 1.0.4-1
--- a/packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
+++ b/packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
@@ -80,9 +80,7 @@ mv %{buildroot}%{gem_instdir}/settings.d/discovery.yml.example \
 %dir %{gem_instdir}
 %config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/discovery.yml
 %license %{gem_instdir}/LICENSE
-%{gem_instdir}/bundler.d
 %{gem_libdir}
-%{gem_instdir}/settings.d
 %{foreman_proxy_bundlerd_dir}/%{plugin_name}.rb
 %exclude %{gem_cache}
 %{gem_spec}
--- a/packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
+++ b/packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
@@ -1,38 +1,36 @@
+# Generated from smart_proxy_discovery-1.0.4.gem by gem2rpm -*- rpm-spec -*-
+# template: smart_proxy_plugin
 %global gem_name smart_proxy_discovery
+%global plugin_name discovery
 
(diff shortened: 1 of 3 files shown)


evgeni:
I'll update this when https://github.com/theforeman/smart_proxy_discovery/commit/60adb369c01a37b575158b756a88b47ab3c7caa8 is in a release, as I want to get rid of the `rest-client < 2` dependency.

evgeni:
this also includes the bump to 1.0.5

packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
```diff
@@ -71,19 +78,24 @@ mv %{buildroot}%{gem_instdir}/settings.d/discovery.yml.example \
...
 %dir %{gem_instdir}
+%config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/discovery.yml
+%license %{gem_instdir}/LICENSE
+%{gem_instdir}/bundler.d
 %{gem_libdir}
-%{gem_spec}
-%{foreman_proxy_bundlerd_dir}/discovery.rb
-%config(noreplace) %{foreman_proxy_settingsd_dir}/discovery.yml
-%doc %{gem_instdir}/LICENSE
+%{gem_instdir}/settings.d
```
ekohl: Probably not needed

packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
```diff
@@ -71,19 +78,24 @@ mv %{buildroot}%{gem_instdir}/settings.d/discovery.yml.example \
 
 %files
 %dir %{gem_instdir}
+%config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/discovery.yml
```
ekohl: Would it make sense to use `%{plugin_name}` here as well as in the %install of it?

packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
```diff
@@ -71,19 +78,24 @@ mv %{buildroot}%{gem_instdir}/settings.d/discovery.yml.example \
...
+%license %{gem_instdir}/LICENSE
+%{gem_instdir}/bundler.d
```
ekohl: Probably not needed

packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
evgeni: probably yes

evgeni:
@ekohl all your comments seem correct -- but that's what the template generates right now. shall we fix the template first?

ekohl:
That sounds like a good idea.

packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
evgeni: so this is generated in a loop https://github.com/theforeman/foreman-packaging/blob/360dc68181bea3c3684a2dff713c4cd880dc360d/gem2rpm/smart_proxy_plugin.spec.erb#L130-L133

which makes it not easy to replace with `%{plugin_name}` (as there might be more configs)

ekohl: Hmm ok, fair enough.

packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
evgeni: This comes directly from `gem2rpm`s `main_file_entries` which lists everything from the gems `files` list, so I'm not sure how to tell it not to add it here

packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
evgeni: same as `settings.d`

packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
evgeni: I dropped these from this spec file for now.

PR #3683, opened 2019-04-26
build rest-client for SCL and non-SCL and update to 2.0.2

evgeni:
We removed the non-SCL build of `rest-client` in d60352adaeb9ddbabbc54e927a5dc8f635d479f9 and have been relying on the version in EPEL (1.6.7) for the smart-proxy (`discovery` and `monitoring` require `rest-client`). I'll prefer if we'd use the same version everywhere.

`discovery` used to require `rest-client < 2` to be able to run on Ruby 1.8 (EL6), but since we don't support that anymore, this requirement was removed in https://github.com/theforeman/smart_proxy_discovery/commit/60adb369c01a37b575158b756a88b47ab3c7caa8 (not in a released `discovery` version yet).


evgeni:
Mhh, this fails repoclosure for nonSCL. I'll add them if this PR is deemed worthy :)

package_manifest.yaml
```diff
@@ -151,6 +151,10 @@ foreman_scl_packages:
     rubygem-record_tag_helper: {}
     rubygem-representable: {}
     rubygem-responders: {}
+    rubygem-rest-client:
+      releasers:
+        - koji-foreman
+        - koji-foreman-plugins
```
ehelms: This shouldnt need to exist in plugins if foreman repository has it.

evgeni: foreman: SCL
plugins: nonSCL
:)

ehelms: Helps if I read all your comments instead of just the first.

evgeni:
[test rpm]

evgeni:
closing in favor of https://github.com/theforeman/foreman-packaging/pull/3783

PR #3690, opened 2019-04-29
also build netrc for nonSCL

evgeni:
this is needed for `rest-client` for nonSCL (#3683)

### DIFF
--- a/packages/foreman/rubygem-netrc/rubygem-netrc.spec
+++ b/packages/foreman/rubygem-netrc/rubygem-netrc.spec
@@ -7,7 +7,7 @@
 Summary: Library to read and write netrc files
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.11.0
-Release: 3%{?dist}
+Release: 4%{?dist}
 Group: Development/Languages
 License: MIT
 URL: https://github.com/geemus/netrc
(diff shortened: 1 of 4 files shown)


packages/foreman/rubygem-netrc/rubygem-netrc.spec
```diff
@@ -51,9 +51,11 @@ cp -a .%{gem_dir}/* \
 
 %check
 pushd .%{gem_instdir}
+%if 0%{?scl}
```
ekohl: Can we also skip the pushd and popd when we're outside of the SCL?

packages/foreman/rubygem-netrc/rubygem-netrc.spec
```diff
@@ -71,6 +73,9 @@ popd
 %{gem_instdir}/test
 
 %changelog
+* Mon Apr 29 2019 Evgeni Golov - 0.11.0-4
+- rebuilt
```
ekohl: I'm sure there's a better description here :)

evgeni: Picky.

ekohl: You know me :)

PR #3650, opened 2019-04-18
Bump Foreman Ansible (core) to version 3

ekohl:
This bumps the SCL'ed bits of Foreman Ansible to version 3. It regenerates the spec files and includes the webpack dependencies now.

PR #3628, opened 2019-04-15
Update rubygem-foreman-tasks to 0.15.5

adamruzicka:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [x] 1.22
* [ ] 1.21
* [ ] 1.20

See Foreman's [plugin maintainer documentation](https://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

PR #3697, opened 2019-04-30
Update foreman_ansible to 3.0.2

ares:

* [x] Nightly
* [x] 1.22
* [ ] 1.21
* [ ] 1.20

---

Replaces https://github.com/theforeman/foreman-packaging/pull/3650 which I rebased and also added a bump to 3.0.1, which fixes the npm dependencies. This needs to get to 1.22 ideally this week, before Friday. @evgeni could you please take a look?


ekohl:
Doesn't look like you updated the NPM dependencies. This is not done automatically, but is automated:
```
./update-requirements npm /path/to/package.json packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```

Ideally we'd extract `package.json` from the gem and run that script if the gem contains `package.json` inside `bump_rpm.sh` and I might hack up something to do that now that we're adding more webpack to plugins.

ekohl:
https://github.com/theforeman/foreman-packaging/pull/3698 automates that process as well if it has the correct template.

ares:
Thanks @ekohl, updated dependencies using your command and kept as a separate commit for easier review. Let me know if you prefer squash or will squash on merge (it it turns green)

ekohl:
[test rpm]

packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -1,76 +1,168 @@
...
+BuildRequires: npm(lodash) < 5.0.0
+BuildRequires: npm(patternfly) >= 3.58.0
+BuildRequires: npm(patternfly) < 4.0.0
+BuildRequires: npm(patternfly-react) >= 2.25.4
+BuildRequires: npm(patternfly-react) < 3.0.0
+BuildRequires: npm(prop-types) >= 15.6.2
+BuildRequires: npm(prop-types) < 16.0.0
+BuildRequires: npm(react) >= 16.6.3
+BuildRequires: npm(react) < 17.0.0
+BuildRequires: npm(react-bootstrap) >= 0.32.4
```
ekohl: Looks like we currently ship 0.32.1. Patternfly-react depends on `< 0.33.0` so we should be able to bump it. This needs an independent packaging PR since it's a build dependency.

ares: that would mean bumping also in foreman core, which is probably too late for 1.22, I'll check if we can relax this in foreman_ansible and use 0.32.1 there, seems it will work

ares:
Updated to foreman_ansible 3.0.2 that relaxed that requirement.

ares:
Now I'm confused
```
DEBUG util.py:491:  Error: Package: tfm-rubygem-foreman_remote_execution-1.7.0-13.fm1_23.el7.noarch (build)
DEBUG util.py:491:             Requires: foreman-webpack-vendor-js(c9584967d7ba3b886b50)
```
REX probably needs rebuild? Opened a PR at https://github.com/theforeman/foreman-packaging/pull/3705

packages/plugins/rubygem-foreman_ansible_core/rubygem-foreman_ansible_core.spec
```diff
@@ -1,29 +1,34 @@
...
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
 %global gem_name foreman_ansible_core
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 2.2.1
-Release: 1%{?foremandist}%{?dist}
+Version: 3.0.0
+Release: 1%{?dist}
```
evgeni: Is the drop of foremandist intentional? I think this happened because the scl and not the plug-in template was used.

evgeni:
Damn Android client. I said two comments. ansible drops buildreq on ansible_core, is that OK?

packages/plugins/rubygem-foreman_ansible_core/rubygem-foreman_ansible_core.spec
ares: I took what @ekohl did, originally I kept it but when he regenerated specs, it was removed. He's travelling today so I'm not sure if we can get answer from him. I'd go with merging, we can fix it if needed in separate PR.

ares:
dropping core build dep was probably not intentional, reverting back, let's see if it builds now

packages/plugins/rubygem-foreman_ansible_core/rubygem-foreman_ansible_core.spec
ares: also note that this is in fact smart proxy dynflow core plugin, not foreman plugin, which may be the reason

ares:
Now it fails on foreman-tasks-core 0.3.2 not available, this will be resolved by https://github.com/theforeman/foreman-packaging/pull/3628

I added a commit that restores foremandist too, let me know if I should keep it or drop it

adamruzicka:
Needs a rebase

ares:
rebased

ares:
[test rpm]

packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -31,6 +31,8 @@ Requires: %{?scl_prefix}rubygem(ipaddress) < 1.0
 BuildRequires: foreman-assets >= %{foreman_min_version}
 BuildRequires: foreman-plugin >= %{foreman_min_version}
 BuildRequires: %{?scl_prefix}rubygem(deface) < 2.0
+BuildRequires: %{?scl_prefix}rubygem(foreman_ansible_core) >= 2.0.2
```
ekohl: Why is this here? AFAIK it's not an actual dependency. It's mentioned as a development dependency so the tests can run, but this gem should be called `smart_proxy_ansible_core`. Refactoring that was a bit hard for me, but it'd really clear up the situation for everyone involved.

packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -91,24 +187,27 @@ cp -a .%{gem_dir}/* %{buildroot}%{gem_dir}/
...
+%{foreman_apipie_cache}
+%{foreman_restart}
 exit 0
 
-
 %changelog
+* Fri May 10 2019 Marek Hulan <mhulan@redhat.com> 3.0.2-1
+- Update to 3.0.2
+
+* Fri May 10 2019 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> 3.0.0-1
```
ekohl: Feel free to drop this changelog

ares:
Squashed to 2 commits (core and foreman_ansible separately)

packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
ares: This follows REX, tasks (and probing) pattern where some code needed to be shared. We gave up (I think) on running REX from core without proxy already so you're right, it may make more sense to be renamed, not something that I'd like to solve in this PR though. I suppose we should also extract that to a separate repo. cc @iNecas

ekohl: But it's not actually a dependency anymore. That's why I left it out in the (autogenerated) dependency list. https://github.com/theforeman/foreman_ansible/commit/e88cfd7efd7ad8abe8b37f2159190e622b79c841 removed it and only kept it as a development dependency for tests (https://github.com/theforeman/foreman_ansible/blob/e805df4ba6f4366423b369c5746a58fc662034d8/foreman_ansible.gemspec#L21-L22). I think that one doesn't make sense since it will pull the gem from rubygems rather than using the actual git sources. Will you only see see failures after you release?

ares:
all green, @ekohl could this get in? it would be great to have it in nightlies for tomorrow talk at OSDC (cc @adamruzicka for status visibility)

ekohl:
> all green, @ekohl could this get in? it would be great to have it in nightlies for tomorrow talk at OSDC (cc @adamruzicka for status visibility)

I still think the foreman_ansible_core dependency should be dropped. See https://github.com/theforeman/foreman-packaging/pull/3697#discussion_r283814670

ares:
One face to face discussion worth of hundred github comments :-) Alright, pushed without the build requirement statements for core.

ares:
green now

PR #3703, opened 2019-05-02
remove foreman_userdata package

timogoebel:
This is only for Nightly and removes `foreman_userdata`, which was moved to core in (https://github.com/theforeman/foreman/pull/6057).


evgeni:
@timogoebel this needs a rebase please (and it should then magically build ;))

timogoebel:
@evgeni: Rebased.

evgeni:
thanks!

evgeni:
```
% koji untag-build --all foreman-plugins-nightly-rhel7 tfm-rubygem-foreman_userdata
% koji remove-pkg foreman-plugins-nightly-rhel7 tfm-rubygem-foreman_userdata
```

PR #3752, opened 2019-05-13
Update patternfly-react package

glekner:


ohadlevy:
@ekohl whats the status here? should he just run it again unbundled ? maybe we should add that to the readme as this is a common package upgrade?

mmoll:
Could this get bumped to 2.34.1? So we can update the minimum version in develop and solve the snapshot mismatch...

ekohl:
Unbundled version: https://github.com/theforeman/foreman-packaging/pull/3775

PR #3753, opened 2019-05-13

ekohl:


packages/foreman/foreman/foreman.spec
```diff
@@ -165,6 +165,9 @@ BuildRequires: %{?scl_prefix_ror}rubygem(sqlite3) < 1.4.0
 # end specfile sqlite BuildRequires
 
 # start specfile main BuildRequires
+BuildRequires: %{?scl_prefix}rubygem(strscan)
```
ekohl: I don't know where these came from. They were autogenerated.

ekohl: Manually removed for now.

PR #3760, opened 2019-05-14
Update rubygem-foreman_ansible to 2.3.6

ares:


evgeni:
http://koji.katello.org/koji/taskinfo?taskID=198536

PR #3761, opened 2019-05-14
Remove webpack provides/requires from Foreman

ekohl:
The goal of this was to make sure we always built webpack against a bundle that was compatible, or were forced to rebuilt it. That part sort of worked, but the idea was that a rebuild with an incompatible build would fail.

The reality is that webpack will happy use any dependency, even if it doesn't fit the interface. This can be seen by various failures where a webpack bundle was generated that was utterly broken.

Since it doesn't really provide the guarantees we wanted but does add a lot of overhead, it's better to remove this.


evgeni:
So my original idea, when implementing this was: don't allow plugins to install with an incompatible Foreman bundle. IMHO that part worked. The reverse did not always work, though. An identical hash in the foreman and plugin rpms did not imply that the plugin UI worked fine.

When we drop these constraints, we still have to rebuild nightlies from time to time to make them work. But now it's someone who uses nightly has to inform us instead of Jenkins.

On the other hand, @sharvit has been working on a proper vendor bundle we can offer as an API, which hopefully should fix all the issues? Or am I reading https://github.com/theforeman/foreman/pull/6605 wrong here?

TL;DR: nuke it if we have a replacement until 1.23 branches :)

ekohl:
> don't allow plugins to install with an incompatible Foreman bundle. IMHO that part worked.

No, it doesn't really. Last week we had the issue that we built a katello that was utterly broken until we merged https://github.com/Katello/katello/commit/157f3522dfd1d0612717384f9cbc7fa91a7a2523. Haven't checked since. Foreman Ansible in nightly is also broken, but we can rebuild it. Then it installs, but the reports page is still broken. Webpack simply doesn't appear care about code and assumes it just works. Any normal compiler does static code analysis to verify functions are present. Javascript is a dynamic language so that's not really possible. I've started to view it as a very stupid transformer rather than a compiler.

ehelms:
I'm really starting to wonder if we shouldn't just ship every plugins generated vendor.js, rename it to <plugin>-vendor and serve it. I'm sure there is some nuance I am missing that would break, but if we did that every plugin would have the bindings present it generated against to reference.

ekohl:
> I'm really starting to wonder if we shouldn't just ship every plugins generated vendor.js, rename it to -vendor and serve it. I'm sure there is some nuance I am missing that would break, but if we did that every plugin would have the bindings present it generated against to reference.

I think that could easily have name collisions, but don't know enough about webpack. Given we already have issues with load times (Foreman has become noticeably slow to render) with just 1 vendor pack, I really wouldn't want to multiply that by n plugins.

ares:
@ekohl are you sure ansible reports page is broken due to this? That would be very disappointing.

ekohl:
@ares I don't know the exact error, but the errors it spits out remind me of that. It's just so cryptic that I have no clue. The goal of the vendor js was to find incompatibilities but AFAIK we haven't found a single one because of it.

ohadlevy:
What's the actual error?

ares:
```
foreman_ansible-ec08546ea644d7b98ffb.js:1 Uncaught (in promise) TypeError: Cannot read property 'props' of undefined
    at u (foreman_ansible-ec08546ea644d7b98ffb.js:1)
    at Ga (react-dom.production.min.js:145)
    at Fo (react-dom.production.min.js:174)
    at qi (react-dom.production.min.js:221)
    at $i (react-dom.production.min.js:222)
    at Es (react-dom.production.min.js:238)
    at js (react-dom.production.min.js:237)
    at vs (react-dom.production.min.js:234)
    at Ki (react-dom.production.min.js:232)
    at Object.enqueueSetState (react-dom.production.min.js:125)
```

I can share a reproducing machine if you're interested

ohadlevy:
On Wed, May 15, 2019 at 5:57 PM Marek Hulán <notifications@github.com>
wrote:

> foreman_ansible-ec08546ea644d7b98ffb.js:1 Uncaught (in promise) TypeError: Cannot read property 'props' of undefined
>     at u (foreman_ansible-ec08546ea644d7b98ffb.js:1)
>     at Ga (react-dom.production.min.js:145)
>     at Fo (react-dom.production.min.js:174)
>     at qi (react-dom.production.min.js:221)
>     at $i (react-dom.production.min.js:222)
>     at Es (react-dom.production.min.js:238)
>     at js (react-dom.production.min.js:237)
>     at vs (react-dom.production.min.js:234)
>     at Ki (react-dom.production.min.js:232)
>     at Object.enqueueSetState (react-dom.production.min.js:125)
>
> I can share a reproducing machine if you're interested
>
> might help if you have source maps enabled you can find the offending code.
AFAIR we remove the source maps from the rpm?

ekohl:
Yes, sadly we do. Probably worth investing some time in figuring out if we can make that work.

ekohl:
@ohadlevy btw, the point of this PR is that webpack doesn't provide the guarantees we were looking for. I don't think I've seen a build fail during a rebuild so we're catching no issues. This means the requirement on an exact vendor hash isn't worth what we thought it would be.

ares:
Here's the same error from development

```
Uncaught (in promise) TypeError: Cannot read property 'props' of undefined
    at ReportJsonViewer (ReportJsonViewer.js:13)
    at renderWithHooks (react-dom.development.js:12839)
    at mountIndeterminateComponent (react-dom.development.js:14814)
    at beginWork (react-dom.development.js:15419)
    at performUnitOfWork (react-dom.development.js:19106)
    at workLoop (react-dom.development.js:19146)
    at renderRoot (react-dom.development.js:19229)
    at performWorkOnRoot (react-dom.development.js:20136)
    at performWork (react-dom.development.js:20048)
    at performSyncWork (react-dom.development.js:20022)
```

given I can reproduce there, it seems it's not related to assets compilations. I still have no clue what caused the issue, perhaps some dependency bump.

OT: how do we support our users if we have bugs in frontend and we can't get better traceback from production setups?

Anyway if provides/requires does not reliable tell us, plugin is broken, let's remove it. I'm tired of rebuilding once a week.

evgeni:
> I don't think I've seen a build fail during a rebuild so we're catching no issues. This means the requirement on an exact vendor hash isn't worth what we thought it would be.

It was never about failing builds but failing installs. And these fail. Whether all the hash changes are worth the fail is a different question.

PR #3763, opened 2019-05-14
[WIP] Disable auto-requires/provides from public

ekohl:
This disables RPM from scanning for requires and provides in public. These should all be static files. This results in a faster build.

May require https://github.com/theforeman/foreman-packaging/pull/3761 or break the webpack provides/requires implementation.


ekohl:
Looks like the before and after is both 10 minutes so this doesn't work.

PR #3762, opened 2019-05-14
Add pulpcore_client and pulp_file_client gems

jlsherrill:
after discussion with the pulp team, they have decided
to produce gems for api access that katello can use.  This
will replace the zest gem that was previously created.
I have left that gem here to be removed in the future as
to not break the build

PR #3764, opened 2019-05-14
Add pulp gem deps to katello

jlsherrill:

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -6,7 +6,7 @@
 %global gem_name katello
 %global prerelease .pre.master
 %global mainver 3.13.0
-%global release 1
+%global release 2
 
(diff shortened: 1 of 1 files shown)


jlsherrill:
requires https://github.com/theforeman/foreman-packaging/pull/3762

jlsherrill:
[test katello]

jlsherrill:
[test rpm]

jlsherrill:
[test rpm]

jlsherrill:
[test rpm]

jlsherrill:
[test rpm]

jlsherrill:
The latest build was green, i'm not sure why its keeping around the old red builds

evgeni:
@jlsherrill yeah, it's not smart enough to throw them away. (it's a non-issue for non-nightly packages, where the version is always the same on rebuild, but here we have the hash in the version and you see red).

PR #3771, opened 2019-05-15
Handle foreman-compute package on updates

mmoll:


evgeni:
Oh god, I hate replacing packages.

Should the replaces/breaks be versioned?
Should we also provide compute for anything depending on that name? (Is there anything?)

mmoll:
> Oh god, I hate replacing packages.

It was only a deletion. ;)

> Should the replaces/breaks be versioned?

I hope not. 1.21 has foreman-compute, while 1.22 has not, so if one does add the 1.22 repo, any foreman-compute should get deinstalled.

>  Should we also provide compute for anything depending on that name? (Is there anything?)

There is nothing left. :scissors:

evgeni:
Jenkins error is unrelated. :ship:

evgeni:
@ekohl unrelated to the PR itself, but why was the `DEB` label only added now, and not when the PR was opened?

PR #3772, opened 2019-05-16
Require SCL prefix only on El7 for REX SSH

ehelms:


packages/plugins/rubygem-smart_proxy_remote_execution_ssh/rubygem-smart_proxy_remote_execution_ssh.spec
```diff
@@ -8,14 +8,19 @@
...
 Version: 0.2.1
-Release: 1%{?dist}
+Release: 2%{?dist}
 Group: Applications/System
 License: GPLv3
 URL: https://github.com/theforeman/smart_proxy_remote_execution_ssh
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
+%if 0%{?rhel} == 7
 Requires: %{?rhel:tfm-}rubygem(smart_proxy_dynflow_core) >= 0.1.5
```
evgeni: the `?rhel:` seems redundant now and we can just write `tfm-rubygem(…)`?

ekohl: Looks like there's a precedent for that:
https://github.com/theforeman/foreman-packaging/blob/403a3e050607433634b29e13a8b6e3e8b8ab2a61/packages/plugins/rubygem-smart_proxy_dynflow/rubygem-smart_proxy_dynflow.spec#L20-L26

PR #3774, opened 2019-05-16
Add nodejs-react-ace and its dependencies

ekohl:
Includes https://github.com/theforeman/foreman-packaging/pull/3773


ohadlevy:
thanks! just wondering, why did we end up with missing packages? is it because we didn't bundle ace at the end?

tbrisker:
I'm confused, should this PR be the one we merge or one of the others?

ekohl:
@ohadlevy yes, this doesn't bundle ace which means there's always only one version of brace.

PR #3782, opened 2019-05-17
Update rubygem-redhat_access to 2.2.6

ares:


evgeni:
http://koji.katello.org/koji/taskinfo?taskID=198057

PR #3783, opened 2019-05-17
simplify rest-client packaging and bump to 2.0.2

evgeni:
simplify = don't special case it for scl-only


evgeni:
Initially I planned to build it for nonSCL too (see https://github.com/theforeman/foreman-packaging/pull/3683), but the version from EPEL which we use is "good" enough and this PR makes our packaging cleaner :)

packages/foreman/rubygem-rest-client/rubygem-rest-client.spec
```diff
@@ -1,83 +1,112 @@
...
+%{gem_instdir}/Rakefile
 %doc %{gem_instdir}/history.md
-
-%exclude %{gem_instdir}/.*
-%exclude %{gem_instdir}/*.gemspec
-%exclude %{gem_instdir}/Gemfile
-%exclude %{gem_instdir}/spec
-%exclude %{gem_instdir}/Rakefile
+%{gem_instdir}/rest-client.gemspec
+%{gem_instdir}/rest-client.windows.gemspec
```
ekohl: Should we exclude this?

evgeni: The windows gemspec you mean? could, yes.

PR #3785, opened 2019-05-17
Fix gofer cfg edit when already set to use python3

ShuffleBox:
Fedora 30 ships with /etc/sysconfig/goferd default containing 
```
# Python interpreter.
PYTHON=/usr/bin/python3

# Python optimization (0=disabled, 1=enabled).
PYTHONOPTIMIZE=1
```
The scripted sed line will result in `PYTHON=/usr/bin/python33` causing goferd to fail to start since python33 is not present.  Adding the end of line match check should make the modification only if it's `PYTHON=/usr/bin/python`

### DIFF
--- a/packages/client/katello-host-tools/katello-host-tools.spec
+++ b/packages/client/katello-host-tools/katello-host-tools.spec
@@ -16,7 +16,7 @@
 
 Name: katello-host-tools
 Version: 3.5.0
-Release: 3%{?dist}
+Release: 4%{?dist}
 Summary: A set of commands and yum plugins that support a Katello host
 Group:   Development/Languages
 License: LGPLv2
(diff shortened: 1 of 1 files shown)


evgeni:
This looks correct to me, thanks @ShuffleBox 

Can you please bump the release and add a changelog entry to the spec file?

ShuffleBox:
I think I have all of the steps covered.  Please let me know if I missed anything in the process.

evgeni:
@ShuffleBox thanks, diff looks good and Jenkins is happy :)

@jlsherrill @jturel mind having a look for a second sanity check?

jturel:
Looks good to me. What kind of dolt would have written the original regex like that? Well, probably me!

Thanks for fixing it! Looks like there is a merge conflict on this branch though.

ehelms:
@ShuffleBox If you can rebase this PR, we can get it merged.

ShuffleBox:
I think I got it, but please doublecheck that I did that correctly.  I haven't had to rebase off of a remote project before, so some of the merge resolution was new to me.

jturel:
> @ehelms I think I got it, but please doublecheck that I did that correctly. I haven't had to rebase off of a remote project before, so some of the merge resolution was new to me.

pretty close. i think you'll need to bump the release to '4' from '3' and update your changelog entry similary & also make it the latest entry in the list

ShuffleBox:
One more try!  Thank you for the guidance.

evgeni:
[test rpm]

evgeni:
The repoclosure error us unrelated. Tha ks!

PR #3801, opened 2019-05-22
Require python-requests in smart_proxy_ansible

ehelms:

### DIFF
--- a/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
+++ b/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
@@ -9,7 +9,7 @@
 Summary: Ansible support for Foreman smart proxy
 Name: rubygem-%{gem_name}
 Version: 2.1.2
-Release: 2%{?foremandist}%{?dist}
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -121,6 +126,9 @@ find %{buildroot}%{gem_instdir}/bin -type f | xargs chmod a+x
 %doc %{gem_docdir}
 
 %changelog
+* Wed May 22 2019 Eric D. Helms <ericdhelms@gmail.com>
```
evgeni: this is missing a version :)

ehelms: Odd, I used obal.

evgeni: I've seen this when it could not figure out the current version for some reason.

evgeni:
Repoclosure failures are mine *cough* not sure how to fix that, tho.

Your change is good.

PR #3802, opened 2019-05-23
foreman kubevirt dependencies for 1.22

evgeni:

PR #3803, opened 2019-05-23
foreman kubevirt for 1.22

evgeni:


evgeni:
depends on https://github.com/theforeman/foreman-packaging/pull/3802

evgeni:
@shiramax @masayag could you please have a brief look that I pulled in the right versions?

evgeni:
> So, if we encounter a bug in the plugin, we'll have to send the packaging PR both to rpm/develop and to rpm/1.22 ?

That would make our life easier, yes. You can also mention in the `rpm/develop` PR that it should go to a stable branch and we'll try to make this happen, but having a PR is an easier work item :)

evgeni:
[test rpm]

evgeni:
http://koji.katello.org/koji/taskinfo?taskID=199884

PR #3799, opened 2019-05-22
Update rubygem-fog-ovirt to 1.2.0

shiramax:
Supported Versions:

 * Nightly

PR #3805, opened 2019-05-23
Update fog-ovirt deps

shiramax:
Supported Versions:

 * Nightly


evgeni:
repoclosure failure is expected, https://github.com/theforeman/foreman-packaging/pull/3799 was just merged

and if we don't merge this one, it will never be promoted as foreman is not installable now.

PR #3806, opened 2019-05-23
Update rubygem-foreman_maintain to 0.4.2

mbacovsky:


evgeni:
http://koji.katello.org/koji/taskinfo?taskID=199715

PR #3817, opened 2019-05-24
Include Pulp repo for EL5 clients

evgeni:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -491,7 +491,7 @@ foreman_client_packages:
         - el5-extras
         - el5-puppet-6
         - el5-subscription-manager
-        - el5-qpid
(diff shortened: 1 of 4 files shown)


evgeni:
hmm, this was found by `obal repoclosure katello-host-tools`, but looking at http://koji.katello.org/releases/yum/foreman-client-nightly/el5/x86_64/ we don't ship katello-agent (which requires all the pulp godness) on el5?!

ehelms:
IIRC building and supporting all the qpid bits on EL5 became not worth the cost so we dropped agent support.

evgeni:
Well, we still *build* `katello-agent` on EL5, we just don't ship it because it's not installable.
Withi this change it at least becomes installable, no idea if working.

PR #3818, opened 2019-05-24
Require python-requests in smart_proxy_ansible

ehelms:

### DIFF
--- a/plugins/smart_proxy_ansible/changelog
+++ b/plugins/smart_proxy_ansible/changelog
@@ -1,3 +1,9 @@
+ruby-smart-proxy-ansible (3.0.0-2) stable; urgency=low
+
+  * Require python-requests for the Ansible Foreman callback
+
+ -- Eric D. Helms <ericdhelms@gmail.com>  Fri, 07 Jun 2019 08:32:41 -0400
+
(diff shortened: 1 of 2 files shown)


ehelms:
@ares Any concerns with this approach?

mmoll:
IMHO this is fine, but needs a changelog entry (3.0.0-2)

ares:
I can't connect this to any issue, what is the reason? If this is a dependency that a lot of ansible modules rely on, then I'm fine with installing it.

ekohl:
The reasoning is that the Ansible Foreman callback requires it. Currently the installer installs this, but on EL8 it's python3-requests. Rather than solving it there, we can also solve it in packaging which is generally more aware of those differences.

ares:
Oh, I got it now, perhaps a comment in the spec file (or in worse case the changelog) would be good, so we don't drop it in future when we do some cleanup.

ekohl:
I actually do think an installer issue (and this may refs it) would be good for tracking purposes.

ekohl:
@mmoll please re-review - now has a changelog but since I suggested the command to @ehelms I'd like someone else to review it :)

mmoll:
merged, thanks @ehelms!

PR #3823, opened 2019-05-24
Drop katello-host-tools for F28

jturel:
I was looking into katello-host-tools build failures and noticed separately that the foreman-client repos are missing katello-agent for EL5 in nightly and other released versions. Not sure if it's something to be fixed here, but I found a few other cleanups.

- https://yum.theforeman.org/client/nightly/el5/x86_64/
- https://yum.theforeman.org/client/1.22/el5/x86_64/
- https://yum.theforeman.org/client/1.21/el5/x86_64/
- https://yum.theforeman.org/client/1.20/el5/x86_64/

I also noticed that EL5 and EL6 repos for clients are missing gofer and its related packages. Since we started building gofer I think we need to have them there, like this: https://yum.theforeman.org/client/1.20/el7/x86_64/

### DIFF
--- a/comps/comps-foreman-client-fedora28.xml
+++ b/comps/comps-foreman-client-fedora28.xml
@@ -11,10 +11,6 @@
     <uservisible>true</uservisible>
     <packagelist>
       <packagereq type="default">foreman-client-release</packagereq>
-      <packagereq type="default">katello-agent</packagereq>
-      <packagereq type="default">katello-host-tools</packagereq>
-      <packagereq type="default">katello-host-tools-fact-plugin</packagereq>
-      <packagereq type="default">katello-host-tools-tracer</packagereq>
       <packagereq type="default">python2-apypie</packagereq>
       <packagereq type="default">python3-apypie</packagereq>
       <packagereq type="default">rubygem-foreman_scap_client</packagereq>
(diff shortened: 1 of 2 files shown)


comps/comps-foreman-client-fedora28.xml
```diff
@@ -12,7 +12,6 @@
     <uservisible>true</uservisible>
     <packagelist>
       <packagereq type="default">foreman-client-release</packagereq>
-      <packagereq type="default">katello-agent</packagereq>
```
jturel: katello-agent is not supported for F28

jturel:
@ehelms is fixing up those repos just a matter of tagging in the right RPMs? If so, where is the automation for that missing the instructions to do the tagging?

ehelms:
IIRC we intentionally dropped katello-agent support because it needs a version of qpid-proton-c that is not available anywhere. Are users asking for this?

@jlsherrill do you recall the above? ^

jturel:
I've started a somewhat related discussion here https://community.theforeman.org/t/katello-host-tools-el5-client-support/14035

ekohl:
https://community.theforeman.org/t/dropping-fedora-28-client-support/15101 is probably relevant. F28 is EOL so let's kill it :fire:

ehelms:
@jturel Checking in

jturel:
@ehelms updated - let me know if I missed something!

ekohl:
Given apypie also has issues and it's EOL I still prefer to completely drop F28 in nightly.

PR #3825, opened 2019-05-26
update recompose package

glekner:


ohadlevy:
this seems like its trying to bundle babel/core ?
error from the logs is BUILDSTDERR: ERROR: the path for dependency "@babel/runtime" already exists

perhaps try rebuilding it using the single strategy ?

glekner:
@ohadlevy I did run with single
`./add_npm_package.sh recompose auto single` , ill try again

glekner:
using `useEffect` now in favor of recompose

PR #3828, opened 2019-05-28
[WIP] @theforeman/vendor npm packaging

evgeni:


tbrisker:
waiting on @sharvit to replace copy-webpack-plugin with a simple cp so we don't need to build 100 node modules into an rpm

sharvit:
> waiting on @sharvit to replace copy-webpack-plugin with a simple cp so we don't need to build 100 node modules into an rpm

After spending some time around replacing `copy-webpack-plugin` I believe it should **not** be done.
It seems like an easy task at the beginning, `cp` from `dist` to `public/webpack`, but it only works with production env.

`copy-webpack-plugin` integrates with the webpack API instead using `cp` so it works with `webpack-dev-server` as well.
I would prefer using a plugin in this case with a community who update and maintain it over time and versions.

It is currently working with `webpack-copy-plugin` v4.6.0 which is still supported by webpack-3 and webpack-4.

evgeni:
I think my last question would be: would it be beneficial for us to use cp in prod and the plugin in dev? The answer can very well be no and we should carry on with my original PR ;)

sharvit:
> I think my last question would be: would it be beneficial for us to use cp in prod and the plugin in dev? The answer can very well be no and we should carry on with my original PR ;)

I don't think so, it will add more complexity to the setup so the tradeoff doesn't worth it.

PR #3845, opened 2019-06-10
Update rubygem-fog-vsphere to 3.1.1

ezr-ondrej:


packages/foreman/rubygem-fog-vsphere/rubygem-fog-vsphere.spec
```diff
@@ -88,6 +88,9 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/tests
```
ekohl: Looks like these are no longer part of the gem so this line should be dropped

ezr-ondrej: Yeah, you are right. Any other place I should delete them from?

PR #3846, opened 2019-06-10
Updated katello gem dependencies

ekohl:
This used the built gem with gem2rpm to generate the requirements. By automating the conversion we can easier make sure both are in sync.

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -1,3 +1,4 @@
+# template: foreman_plugin
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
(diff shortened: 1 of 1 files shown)


ekohl:
Depends on https://github.com/theforeman/foreman-packaging/pull/3847

Generated by [fm-packaging-update-deps](https://github.com/ekohl/theforeman-plumbing/blob/master/fm-packaging-update-deps). Will need https://github.com/Katello/katello/pull/8149 to correctly build the gem from any directory.

jlsherrill:
[test rpm]

ekohl:
Not sure why this fails:
```
+ /usr/bin/tfm-rake 'plugin:assets:precompile[katello]' RAILS_ENV=production --trace
Gem loading error: Could not find 'foreman-tasks' (>= 0.14.1, ~> 0.13) among 121 total gem(s)
```

ekohl:
Oh wait, I see - it's the missing build requires. I'll need to look at my script to see why it doesn't generate them.

ekohl:
The issue was that I was building the gem from a different directory. While it does produce a gem, it doesn't include all files. That tripped up the gem2rpm template which detects whether a plugin has assets. If it doesn't, it doesn't add build dependencies.

ekohl:
:green_heart: now

PR #3847, opened 2019-06-10
bump pulp_file_client gem

jlsherrill:


jlsherrill:
@ekohl oh thanks!  didn't know about that, will do !

PR #3852, opened 2019-06-12
change katello's pulp_file_client pin

jlsherrill:


ekohl:
I just realized this build didn't bump the release. Also, in packaging we already have a 3.x version and this pins to a 0.x version. This means we need to downgrade.

jlsherrill:
Doh, forgot to bump release.  As for downgrading, this was not in a
released version, I'm okay if the upgrade doesn't work from a nightly a
week ago to today, but I can fix it when I get back if you'd like.  The gem
authors are still figuring out how to release rubygems and it's been a tad
annoying :)

ekohl:
Probably fine since it's in nightly, but it does mean we need to untag in koji and likely fix the spec so some runaway script doesn't attempt to release it again. Could you submit a PR?

jlsherrill:
Yep I already untagged, what would cause it to be retagged?

I'm afk for a few days now, so can't do much till then as far as a new PR,
other than reply to emails on my phone :)

ekohl:
Our release job in Jenkins looks at modified files and does an `obal release`. Not that likely so it can wait for a bit I think.

PR #3855, opened 2019-06-19
Fixes #27083 - Add service file for smart_proxy_dynflow_core

adamruzicka:
This will allow us to run smart proxy dynflow core the same way it is run on EL-based distros.

### DIFF
--- a/plugins/smart_proxy_dynflow_core/rules
+++ b/plugins/smart_proxy_dynflow_core/rules
@@ -13,6 +13,7 @@ export DH_RUBY_IGNORE_TESTS=require-rubygems
 
 build:
 	/bin/sed -ri '/^ExecStart/ s#/usr/bin/smart_proxy_dynflow_core#/usr/bin/bundle exec smart_proxy_dynflow_core#' deploy/smart_proxy_dynflow_core.service
+	/bin/sed -ri '/^EnvironmentFile/ s#sysconfig#default#'
 
 %:
 	dh $@ --buildsystem=ruby --with ruby
--- a/plugins/smart_proxy_dynflow_core/rules
+++ b/plugins/smart_proxy_dynflow_core/rules
@@ -13,7 +13,7 @@ export DH_RUBY_IGNORE_TESTS=require-rubygems
 
 build:
 	/bin/sed -ri '/^ExecStart/ s#/usr/bin/smart_proxy_dynflow_core#/usr/bin/bundle exec smart_proxy_dynflow_core#' deploy/smart_proxy_dynflow_core.service
-	/bin/sed -ri '/^EnvironmentFile/ s#sysconfig#default#'
+	/bin/sed -ri '/^EnvironmentFile/ s#sysconfig#default#' deploy/smart_proxy_dynflow_core.service
 
 %:
 	dh $@ --buildsystem=ruby --with ruby
--- a/plugins/smart_proxy_dynflow_core/rules
+++ b/plugins/smart_proxy_dynflow_core/rules
@@ -12,7 +12,11 @@ export DH_RUBY_IGNORE_TESTS=require-rubygems
 #export DH_RUBY_GEMSPEC=gem.gemspec
 
 build:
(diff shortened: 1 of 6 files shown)


ekohl:
Could you open a Redmine issue for this so we can track it? We'll also have a matching installer PR

adamruzicka:
done

adamruzicka:
Changed to `/etc/default`. The goal of this is to unify the deployment on EL and debians, I didn't want to introduce another inconsistency (env file vs systemd dropin). I'll be glad to open another PRs? which would change this to use systemd dropins everywhere later.

adamruzicka:
One more thing comes to mind. Does debian automatically start services after they are installed?
If it does, can we get away with leaving the service running there without actually being used or should I try to work around that somehow here?

ekohl:
I think that's fine. We'll also modify the installer to actually use the service so when you're installing it, it's likely you want the service anyway.

plugins/smart_proxy_dynflow_core/rules
```diff
@@ -11,5 +11,9 @@ export DH_RUBY_IGNORE_TESTS=require-rubygems
 # If you need to specify the .gemspec (eg there is more than one)
 #export DH_RUBY_GEMSPEC=gem.gemspec
 
+build:
+	/bin/sed -ri '/^ExecStart/ s#/usr/bin/smart_proxy_dynflow_core#/usr/bin/bundle exec smart_proxy_dynflow_core#' deploy/smart_proxy_dynflow_core.service
+	/bin/sed -ri '/^EnvironmentFile/ s#sysconfig#default#'
```
ekohl:
```
12:25:47 /bin/sed -ri '/^EnvironmentFile/ s#sysconfig#default#'
12:25:47 /bin/sed: no input files
```

```suggestion
	/bin/sed -ri '/^EnvironmentFile/ s#sysconfig#default#' deploy/smart_proxy_dynflow_core.service
```

adamruzicka: Yeah, sometimes I should stop and read the stuff I wrote...

mmoll:
I guess there needs more stuff to be set...
```
Jun 22 13:26:42 sledge bundle[2509]: `/usr/share/foreman-proxy` is not writable.
Jun 22 13:26:42 sledge bundle[2509]: Bundler will use `/tmp/bundler/home/foreman-proxy' as your home directory tempora
Jun 22 13:26:42 sledge bundle[2509]: Could not locate Gemfile or .bundle/ directory
Jun 22 13:26:42 sledge systemd[1]: smart_proxy_dynflow_core.service: Control process exited, code=exited status=10
Jun 22 13:26:42 sledge systemd[1]: smart_proxy_dynflow_core.service: Failed with result 'exit-code'.
Jun 22 13:26:42 sledge systemd[1]: Failed to start Foreman smart proxy dynflow core service.
-- Subject: Unit smart_proxy_dynflow_core.service has failed
```

adamruzicka:
I must admit this is ugly, but I haven't found any other way (I tried building a forest of symlinks and failed miserably) how to make the sources live both in `/usr/share/smart_proxy_dynflow_core` (need to put `Gemfile.in` and `bundler.d` in there) and in `/usr/lib/ruby/vendor_ruby`.

The service is now able to start. Further changes should be made to various `*-core` packages to deploy their stuff into `/usr/share/smart_proxy_dynflow_core/bundler.d/`, but I'd leave that for another PR

mmoll:
TBH, I still think that all these `_core`-gems make things unnecessarily hard... What is the reason to also bring this to Debian and RHEL8 now?

adamruzicka:
I'd say trying to keep things compatible is making things hard. Everything would be easier even with `*-core` gems if we could just drop support for the dynflow-in-proxy deployment from everywhere.

Originally we needed to split the dynflow core from proxy because the proxy wasn't SCLed and we needed SCL bits for dynflow. We used the split deployment on EL* for some time and with a recent change to smart_proxy_dynflow we're now able to run it the same way on debian too. The main drive for this is unification across platforms. Also the dynflow core can hit the cpu pretty hard at times and having it outside the proxy imo makes sense at least from the "keeping proxy light" viewpoint

mmoll:
I see. I guess there's no way to get the goodness from one without the other...

mmoll:
merged, díky @adamruzicka!

PR #3862, opened 2019-06-27

ekohl:

### DIFF
--- a/packages/foreman/foreman-release/foreman-release.spec
+++ b/packages/foreman/foreman-release/foreman-release.spec
@@ -13,7 +13,7 @@
 %define repo_dist %{dist}
 %endif
 
-%global release 2
+%global release 3
 %global prerelease develop
 
 Name:     foreman-release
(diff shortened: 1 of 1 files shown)


ekohl:
Don't merge until https://github.com/theforeman/foreman-packaging/pull/3863 has been verified as well. That's the easiest way to test the if/else branch for (non-)release.

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -98,19 +99,24 @@ if [[ '%{release}' != *"develop"* ]];then
   VERSION="%{version}"
   sed "s/nightly/${VERSION%.*}/g" -i %{buildroot}%{repo_dir}/*.repo
   sed "s/gpgcheck=0/gpgcheck=1/g" -i %{buildroot}%{repo_dir}/foreman.repo
-  sed "s/gpgcheck=0/gpgcheck=1/g" -i %{buildroot}%{repo_dir}/foreman-rails.repo
+  sed "s/gpgcheck=0/gpgcheck=1/g" -i %{buildroot}%{repo_dir}/foreman-{rails,client}.repo
```
evgeni: I'd prefer this to be two separate lines (otherwise you could also write {,-rails,-client}, right? ;))

ekohl: Yes, I considered that as well. Do you prefer separate lines or 1 smart line?

evgeni: I find separate lines more readable, as the user does not have to know/remember how bash/zsh does brackets (and that plain posix sh does not ;))

ekohl: Updated

PR #3863, opened 2019-06-27

ekohl:

(cherry picked from commit 92da00fdf47d80f2c089207c421708c29b4026a9)

### DIFF
--- a/packages/foreman/foreman-release/foreman-release.spec
+++ b/packages/foreman/foreman-release/foreman-release.spec
@@ -13,7 +13,7 @@
 %define repo_dist %{dist}
 %endif
 
-%global release 1
+%global release 2
 
 Name:     foreman-release
 Version:  1.22.0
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -95,19 +96,24 @@ if [[ '%{release}' != *"develop"* ]];then
   VERSION="%{version}"
   sed "s/nightly/${VERSION%.*}/g" -i %{buildroot}%{repo_dir}/*.repo
   sed "s/gpgcheck=0/gpgcheck=1/g" -i %{buildroot}%{repo_dir}/foreman.repo
-  sed "s/gpgcheck=0/gpgcheck=1/g" -i %{buildroot}%{repo_dir}/foreman-rails.repo
+  sed "s/gpgcheck=0/gpgcheck=1/g" -i %{buildroot}%{repo_dir}/foreman-{rails,client}.repo
```
evgeni: same as in #3862 ;)

ekohl: Updated

ekohl:
The repositories should be signed so installing this RPM should allow actual installation. Will test once it's built.

ekohl:
Updated and manually verified on a fresh CentOS 7 that it checks the GPG key when installing.

PR #3875, opened 2019-07-01

ekohl:

(cherry picked from commit 0c9164e8eab5f3ea084654cece4eca98c3e761cb)


ekohl:
We'll do a 1.20.3 for upgrades, which is why I included this but you're right, it's not strictly needed. Also looks like the build is failing so I'll close this.

PR #3879, opened 2019-07-02
Update rubygem-foreman_maintain to 0.4.4

evgeni:


packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec
```diff
@@ -8,8 +8,9 @@
 
 Summary: The Foreman/Satellite maintenance tool
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.4.3
+Version: 0.4.4
 Release: 1%{?dist}
+Epoch: 1
```
evgeni: we bumped that downstream, so it'd be good to have this in sync upstream (cc @ekohl)

evgeni:
the repoclosure errors are unrelated, we currently fail to properly do repoclosure for non-scl packages :/

merging

PR #3887, opened 2019-07-03
Fixes #27225: Fix the Foreman restart macro

ekohl:
Previously it relied on the Foreman build macros to be defined but this is not defined as a build dependency so the macro is not expanded. This results in an error:

    /var/tmp/rpm-tmp.eSpJPO: line 7: fg: no job control

(cherry picked from commit d2314333bce3da45473151bd798cf288a8bb2751)


evgeni:
repoclosure should be fixed by https://github.com/theforeman/foreman-packaging/pull/3891

PR #3888, opened 2019-07-03
Release 1.20.3-2 with ruby 2.3 compatible spring

tbrisker:
Currently we package all of the gems into the foreman package, but we don't actually need to package development and test bundles (we don't for rpm as well)


tbrisker:
modified this to change the gemfile so it uses spring <2.1

mmoll:
merged, thanks @tbrisker!

PR #3889, opened 2019-07-04
Drop development package

tbrisker:
We don't ship development gems in rpm as well, this package isn't tested
to actually work.


tbrisker:
closing in favor of GH-3890

PR #3890, opened 2019-07-04
Drop development package

tbrisker:
We don't ship development gems in rpm as well, this package isn't tested
to actually work.


debian/bionic/foreman/rules
```diff
@@ -11,6 +11,7 @@ build:
 	/usr/bin/gem install -f rake -v '12.3.1'
 	/usr/bin/gem install -f minitest -v '5.10.3'
 	/usr/bin/gem install -f thor -v '0.19.4'
+	/bin/rm -f bundler.d/test.rb bundler.d/development.rb
```
ekohl: Nit: I'd drop the `-f` since that'll tell us when we (for some reason) not have these files which would be highly unexpected.

evgeni:
With this change, if somebody *had* `foreman-devel` installed, they will not be able to upgrade to the new `foreman` package without *manually* removing `foreman-devel` (because of the explicit `Depends` of `foreman-devel` on `foreman (= ${binary:Version})`).

I am not sure what the absolute correct way to fix that is, but adding a `Breaks: foreman-devel` to `foreman` might at least hint `apt` to do the right thing. I'll go and ask somebody who knows stuff about Debian packaging ;)

mmoll:
merged, thanks @tbrisker!

PR #3896, opened 2019-07-05
build pulp_docker gem for katello

jlsherrill:

### DIFF
--- a/comps/comps-katello-server-rhel7.xml
+++ b/comps/comps-katello-server-rhel7.xml
@@ -55,6 +55,7 @@
       <packagereq type="default">tfm-rubygem-justified</packagereq>
       <packagereq type="default">tfm-rubygem-pulpcore_client</packagereq>
       <packagereq type="default">tfm-rubygem-pulp_ansible_client</packagereq>
+      <packagereq type="default">tfm-rubygem-pulp_docker_client</packagereq>
       <packagereq type="default">tfm-rubygem-pulp_file_client</packagereq>
       <packagereq type="default">tfm-rubygem-qpid_messaging</packagereq>
       <packagereq type="default">tfm-rubygem-redhat_access_lib</packagereq>
@@ -83,6 +84,7 @@
       <packagereq type="optional">tfm-rubygem-hpricot-doc</packagereq>
       <packagereq type="default">tfm-rubygem-pulpcore_client-doc</packagereq>
       <packagereq type="default">tfm-rubygem-pulp_ansible_client-doc</packagereq>
+      <packagereq type="default">tfm-rubygem-pulp_docker_client-doc</packagereq>
       <packagereq type="default">tfm-rubygem-pulp_file_client-doc</packagereq>
       <packagereq type="optional">tfm-rubygem-redhat_access-doc</packagereq>
       <packagereq type="optional">tfm-rubygem-redhat_access_lib-doc</packagereq>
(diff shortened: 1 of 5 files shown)


jlsherrill:
@evgeni @ekohl  This will be required for the next pipeline run

packages/katello/rubygem-pulp_docker_client/rubygem-pulp_docker.spec
```diff
@@ -0,0 +1,93 @@
+# Generated from pulp_docker_client-4.0.0b6.dev01562331743.gem by gem2rpm -*- rpm-spec -*-
```
ekohl: Can you rename this script to follow the convention that the spec has the same name as the directory and package name?

packages/katello/rubygem-pulp_docker_client/rubygem-pulp_docker.spec
```diff
@@ -0,0 +1,93 @@
...
+
+%global gem_name pulp_docker_client
+
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 4.0.0b6.dev01562331743
+Release: 1%{?dist}
+Summary: Pulp 3 Docker API Ruby Gem
+Group: Development/Languages
+License: GPLv2
+URL: https://github.com/pulp/pulp_docker/
```
ekohl: This is the Pulp docker plugin itself. Would https://rubygems.org/gems/pulp_docker_client be a better URL?

packages/katello/rubygem-pulp_docker_client/rubygem-pulp_docker.spec
```diff
@@ -0,0 +1,93 @@
...
+%{?scl:EOF}
+
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -pa .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+%files
+%dir %{gem_instdir}
+%{gem_instdir}/git_push.sh
```
ekohl: This sounds like a helper. Should this be excluded or at least moved to the docs package?

packages/katello/rubygem-pulp_docker_client/rubygem-pulp_docker.spec
```diff
@@ -0,0 +1,93 @@
...
+%{gem_libdir}
+%exclude %{gem_cache}
+%{gem_spec}
+
+%files doc
+%doc %{gem_docdir}
+%{gem_instdir}/Gemfile
+%{gem_instdir}/Gemfile.lock
```
ekohl: Libraries shouldn't include a `Gemfile.lock`. Perhaps something the gem generation process can do?

packages/katello/rubygem-pulp_docker_client/rubygem-pulp_docker.spec
jlsherrill: whooops! good catch

packages/katello/rubygem-pulp_docker_client/rubygem-pulp_docker.spec
jlsherrill: if they ever get ruby docs published, it may be it.  Will change it for now.

packages/katello/rubygem-pulp_docker_client/rubygem-pulp_docker.spec
jlsherrill: i will file a pulp bug, and for now delete it

jlsherrill:
Thanks for the review @ekohl , pr updated

PR #3903, opened 2019-07-10
add the owner to namespaced npm packages

evgeni:
npm2rpm 5.0.0 introduced a new package naming scheme, let's follow suit

### DIFF
--- a/add_npm_package.sh
+++ b/add_npm_package.sh
@@ -7,7 +7,14 @@ TITO_TAG=foreman-nightly-nonscl-rhel7
 DISTRO=${TITO_TAG##*-}
 BASE_DIR=${4:-foreman}
 
-PACKAGE_NAME=nodejs-${NPM_MODULE_NAME##*/}
+PACKAGE_MODULE=${NPM_MODULE_NAME##*/}
+PACKAGE_VENDOR=${NPM_MODULE_NAME%%/*}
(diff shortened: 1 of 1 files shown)


add_npm_package.sh
```diff
@@ -7,7 +7,14 @@ TITO_TAG=foreman-nightly-nonscl-rhel7
 DISTRO=${TITO_TAG##*-}
 BASE_DIR=${4:-foreman}
 
-PACKAGE_NAME=nodejs-${NPM_MODULE_NAME##*/}
+PACKAGE_MODULE=${NPM_MODULE_NAME##*/}
+PACKAGE_VENDOR=${NPM_MODULE_NAME%%/*}
+PACKAGE_VENDOR=${PACKAGE_VENDOR##@}
+if [[ $NPM_MODULE_NAME == @* ]]; then
```
ekohl: Should this check if `PACKAGE_VENDOR` is non-empty?

evgeni: It never is -- it will be equal the package module name for a non-namespaced package, but I thought that checking vendor == module is also wrong, as someone might have `@foo/foo`

ekohl: Pehaps it should check if `$PACKAGE_NAME == */*` because we really do this because of the slash.

evgeni: updated.

PR #3906, opened 2019-07-11

ekohl:


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -100,6 +102,8 @@ BuildRequires: npm(babel-preset-react) >= 6.24.1
 BuildRequires: npm(babel-preset-react) < 7.0.0
 BuildRequires: npm(identity-obj-proxy) >= 3.0.0
 BuildRequires: npm(identity-obj-proxy) < 4.0.0
+BuildRequires: npm(react-router-dom) >= 4.2.2
```
evgeni: that looks not sorted to me? :)

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -43,8 +43,9 @@ Requires: %{?scl_prefix}rubygem(runcible) >= 2.11.0
 Requires: %{?scl_prefix}rubygem(runcible) < 3.0.0
 Requires: %{?scl_prefix}rubygem(anemone)
 Requires: %{?scl_prefix}rubygem(pulpcore_client)
-Requires: %{?scl_prefix}rubygem(pulp_ansible_client) < 1.0
-Requires: %{?scl_prefix}rubygem(pulp_file_client) < 1.0
+Requires: %{?scl_prefix}rubygem(pulp_file_client) < 3.0.0
```
evgeni: also here, sorting is weird. (but the version changes are correct)

packages/katello/rubygem-katello/rubygem-katello.spec
ekohl: How so? If you do sort it alphabetically I think `i < rea < red` is correct. Do you mean that `<` comes before `>` in ascii because I prefer it this way.

packages/katello/rubygem-katello/rubygem-katello.spec
ekohl: I think this one isn't sorted by the script and it just follows what's defined in the gemspec. If you look at the file, you'll see there's a lot of deps that aren't sorted. We can change the gem2rpm templates to sort.

packages/katello/rubygem-katello/rubygem-katello.spec
evgeni: Oh wait. It moved from dependencies to devDependencies. That's why it jumped.

I wondered why react-router-dom is not next to react-router-bootstrap anymore

packages/katello/rubygem-katello/rubygem-katello.spec
evgeni: Right. wfm.

ekohl:
[2019-07-11T12:09:43.746Z]     package: tfm-rubygem-katello-3.13.0-0.6.pre.master.201907111142git7b97424.el7.noarch from downloaded_rpms

[2019-07-11T12:09:43.746Z]       unresolved deps:

[2019-07-11T12:09:43.746Z]          tfm-rubygem(pulp_docker_client)

ekohl:
[test rpm] now that https://ci.theforeman.org/job/foreman-packaging-release/207/ has released that package.

ekohl:
[test rpm] because I didn't mash in between

PR #3908, opened 2019-07-12
don't inspect foreman-nighly when doing nonscl repoclosure

evgeni:
only use foreman-nightly as a lookaside.
inspecting it raises errors as the repo also contains scl'ed packages

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -206,11 +206,9 @@ foreman_nonscl_packages:
       - koji-foreman
     nightly_releaser: koji-foreman-jenkins
     repoclosure_config: repoclosure/yum_el7.conf
-    repoclosure_target_repos:
-      rhel7:
-        - el7-foreman-nightly
     repoclosure_lookaside_repos:
       rhel7:
+        - el7-foreman-nightly
         - el7-base
         - el7-updates
         - el7-epel


package_manifest.yaml
```diff
@@ -206,11 +206,9 @@ foreman_nonscl_packages:
       - koji-foreman
     nightly_releaser: koji-foreman-jenkins
     repoclosure_config: repoclosure/yum_el7.conf
-    repoclosure_target_repos:
-      rhel7:
-        - el7-foreman-nightly
     repoclosure_lookaside_repos:
       rhel7:
+        - el7-foreman-nightly
```
ekohl: I think you should actually drop this. It shouldn't even be available. SCL can depend on nonscl but not the other way around. Otherwise we'd also need rails and other SCL repos.
```suggestion
```

evgeni: but `el7-foreman-nightly` contains other non-scl'ed packages our non-scl'ed package could depend on?

ekohl: The koji tag [foreman-nightly-nonscl-rhel7-build](http://koji.katello.org/koji/taginfo?tagID=7) doesn't have the SCL repo in its build tree so I'd say that's unlikely.

evgeni: While correct, `el7-foreman-nightly` is not a koji tag, but a repo (http://koji.katello.org/releases/yum/foreman-nightly/RHEL/7/x86_64/) which gets populated by both the scl and nonscl tags.

So while we build smth in `foreman-nightly-nonscl-rhel7-build`, the downloaded result is checked against the contents of a repo, not against the contents of the tag (which has a hidden repo somewhere)

ekohl: You're right, it is needed as a lookaside.

PR #3909, opened 2019-07-12
don't inspect foreman-plugins-nighly when doing nonscl repoclosure

evgeni:
only use foreman-plugins-nightly as a lookaside.
inspecting it raises errors as the repo also contains scl'ed packages


evgeni:
same as #3908 but for plugins

PR #3917, opened 2019-07-16
Update foreman-tasks to 0.16.0

adamruzicka:


mmoll:
merged, díky @adamruzicka!

PR #3928, opened 2019-07-17
Katello 3.12.1

evgeni:


packages/katello/katello-repos/katello-repos.spec
```diff
@@ -6,10 +6,10 @@
 %define repo_dist %{dist}
 
 #global prerelease .rc2
```
ekohl: In Foreman spec files I remove these when we've released a GA. Not something that needs to happen here, but probably good to unify workflows.

evgeni: True

PR #3929, opened 2019-07-17
Rebuild REX for @theforeman/vendor

evgeni:


evgeni:
This is not needed, 1.8.2 was built against the new Foreman already.

PR #3930, opened 2019-07-17
Rebuild foreman_ansible for @theforeman/vendor

evgeni:

### DIFF
--- a/packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
+++ b/packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
@@ -9,7 +9,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 3.0.3
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Summary: Ansible integration with Foreman (theforeman.org)
 Group: Applications/Systems
 License: GPLv3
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -157,6 +157,7 @@ cp -pa .%{gem_dir}/* \
 %{foreman_assets_plugin}
 %{foreman_webpack_plugin}
 %{foreman_webpack_foreman}
+%exclude %{gem_instdir}/public/webpack/foreman-vendor.*
```
ekohl: Should this be part of the precompile macro?
https://github.com/theforeman/foreman-packaging/blob/2224d7916e19f18f1d81fe5c10fc3f61db2e302e/packages/foreman/foreman/foreman.spec#L901-L938
We already remove `*.js.map` there. Probably also good to see if those files still contain the BUILDROOT (which is why we dropped them).

evgeni: Yeah, I also thought of that and agree that this is probably the best place we can have for this cleanup.

evgeni: https://github.com/theforeman/foreman-packaging/pull/3933

evgeni: #3933 got merged and I've updated this PR to not include the exclude (heh).

PR #3931, opened 2019-07-17
add bootstrap-sass to f-tasks to fix build problems

evgeni:


evgeni:
I have opened https://github.com/theforeman/foreman-tasks/pull/438

PR #3936, opened 2019-07-18
Update rubygem-foreman_ansible to 3.0.5

xprazak2:


xprazak2:
foreman_ansible has slightly different release workflow than other plugins and it removed the translations for me. Updated, they should be back now.

evgeni:
[test rpm]

PR #3937, opened 2019-07-19
Fixes #27424 - Seed the DB in rubygem-katello posttrans

jturel:
**To test:**

- Spin up centos7-katello-nightly

- Delete all the upgrade tasks in the DB:

```
> foreman-rake console
> UpgradeTask.delete_all
```

- yum install http://koji.katello.org/kojifiles/work/tasks/4471/214471/tfm-rubygem-katello-3.13.0-0.8.pre.master.20190724163855gita08a213.el7.noarch.rpm

- check the UpgradeTasks:

```
> foreman-rake console
> UpgradeTask.all
```

You should see all of Katello's Upgrade Tasks!

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -7,7 +7,7 @@
 %global gem_name katello
 %global prerelease .pre.master
 %global mainver 3.13.0
-%global release 7
+%global release 8
 
 Name:    %{?scl_prefix}rubygem-%{gem_name}
 Summary: Content and Subscription Management plugin for Foreman
@@ -240,7 +240,14 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/package.json
 %{gem_instdir}/webpack
 
+%posttrans
+%{foreman_db_seed}
+exit 0
+
 %changelog
+* Wed Jul 24 2019 Jonathon Turel <jturel@gmail.com> - 3.13.0-0.8.pre.master
+- Seed the database after transaction to ensure upgrade steps will run
+
 * Tue Jul 23 2019 Evgeni Golov - 3.13.0-0.7.pre.master
 - Update katello packaging to @theforeman/vendor bundle
 


jturel:
[test packaging]

jturel:
[test rpm]

jturel:
[test packaging]

evgeni:
[test rpm]

mccun934:
@ekohl I don't understand how upgrades ever worked for Katello if we were missing this?

were we just getting lucky that most people were upgrading foreman or other plugins that called db:seed?

doesn't this seem more like a bug in the installer that --upgrade should just call db:seed whenever an upgrade is being run? This 'don't forget to put it in the %posttrans' seems really fragile

mccun934:
.. as a follow on, I'm totally fine with this change if this is the way it works across the stack. I was just wondering if we may want to improve the installer to better detect and handle when seed should be run, eg, you update an RPM that might change schema which is missing the db:seed in post and you run the installer expecting it to handle it.

jlsherrill:
I'd prefer us move away from putting db operations in the spec file IMO, so yeah, lets put it in the installer if its not already being run.   
It looks like seeding today is done via the foreman rpm %post script and via the installer here: https://github.com/theforeman/puppet-foreman/blob/81fcf11a48d3b5bee659e9a84c8927701e5d2674/manifests/database.pp#L33

maybe @ekohl would be able to explain when that does or doesn't happen.

We could add it either as part of https://github.com/theforeman/foreman-installer/blob/ff9143a4a38fd2b0bd35c8103546014effb6b96b/katello/hooks/post/30-upgrade.rb or just before.

chris1984:
Results of testing:

```bash
[root@centos7-katello-nightly ~]# foreman-rake console
Loading production environment (Rails 5.2.1)
irb(main):001:0> UpgradeTask.delete_all
=> 10

[root@centos7-katello-nightly ~]# foreman-rake console
Loading production environment (Rails 5.2.1)
irb(main):001:0> UpgradeTask.all
=> #<ActiveRecord::Relation [#<UpgradeTask id: 11, name: "katello:correct_repositories", task_name: "katello:correct_repositories", long_running: true, always_run: true, skip_failure: true, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">, #<UpgradeTask id: 12, name: "katello:correct_puppet_environments", task_name: "katello:correct_puppet_environments", long_running: true, always_run: true, skip_failure: true, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">, #<UpgradeTask id: 13, name: "katello:clean_backend_objects", task_name: "katello:clean_backend_objects", long_running: true, always_run: true, skip_failure: true, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">, #<UpgradeTask id: 14, name: "katello:upgrades:3.8:clear_checksum_type", task_name: "katello:upgrades:3.8:clear_checksum_type", long_running: false, always_run: false, skip_failure: false, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">, #<UpgradeTask id: 15, name: "katello:upgrades:3.10:clear_invalid_repo_credentia...", task_name: "katello:upgrades:3.10:clear_invalid_repo_credentia...", long_running: false, always_run: false, skip_failure: false, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">, #<UpgradeTask id: 16, name: "katello:upgrades:3.10:update_gpg_key_urls", task_name: "katello:upgrades:3.10:update_gpg_key_urls", long_running: false, always_run: false, skip_failure: false, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">, #<UpgradeTask id: 17, name: "katello:upgrades:3.11:import_yum_metadata", task_name: "katello:upgrades:3.11:import_yum_metadata", long_running: false, always_run: false, skip_failure: false, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">, #<UpgradeTask id: 18, name: "katello:upgrades:3.11:update_puppet_repos", task_name: "katello:upgrades:3.11:update_puppet_repos", long_running: false, always_run: false, skip_failure: false, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">, #<UpgradeTask id: 19, name: "katello:upgrades:3.11:clear_checksum_type", task_name: "katello:upgrades:3.8:clear_checksum_type", long_running: false, always_run: false, skip_failure: false, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">, #<UpgradeTask id: 20, name: "katello:upgrades:3.12:remove_pulp2_notifier", task_name: "katello:upgrades:3.12:remove_pulp2_notifier", long_running: false, always_run: false, skip_failure: false, last_run_time: nil, ordering: 100, subject: "katello", created_at: "2019-07-26 14:48:07", updated_at: "2019-07-26 14:48:07">]>
```

ekohl:
> maybe @ekohl would be able to explain when that does or doesn't happen.

The way the installer detects if it needs to run migrations and seeds is https://github.com/theforeman/puppet-foreman/blob/a14059fdcb7701336396a570def254588e7f955e/manifests/database.pp#L24-L35.
The key to understand it is that the rake define only runs when it's refreshed (to get idempotency).  When the config entry `db_pending_seed` isn't `false` it will trigger seeding.

I don't know if there's a mechanism in Foreman to detect this, but an alternative to this PRs implementation is to make sure the Foreman application knows it needs seeding. Perhaps that means calling `foreman-rake -- config -k db_pending_seed -v true`.

ekohl:
> @ekohl I don't understand how upgrades ever worked for Katello if we were missing this?
> 
> were we just getting lucky that most people were upgrading foreman or other plugins that called db:seed?

That's my likely explanation for the time between 3.9 and 3.12. It used to rely on code in the upgrade hook. https://github.com/Katello/katello-installer/commit/dcc0f330b003c4ac74c8ce1030336a28960c3b9f#diff-1453f17351c9fefb73f41a2dd0837b82L10-L12 removed this and we probably didn't see the bigger picture completely.

> doesn't this seem more like a bug in the installer that --upgrade should just call db:seed whenever an upgrade is being run? This 'don't forget to put it in the %posttrans' seems really fragile

It's not that fragile since it's part of the template. Katello is the only plugin that doesn't do this and instead did it in the installer. That was one of the reasons Katello needed the installer where vanilla Foreman would work fine without rerunning the installer (unless you needed config changes). Note https://theforeman.org/manuals/1.22/index.html#3.6Upgrade states rerunning the installer is an optional step.

There are some reasons why we wouldn't want to seed this way. https://projects.theforeman.org/issues/21965 is open for one reason. The short version is that you don't want to db:migrate and db:seed in n RPM packages because it can really slow it down. A `yum install katello` would probably be much faster if we didn't.

Another is load balanced setups. If you have multiple webservers, doing parallel yum upgrades is very inefficient at best and unpredictable at worst. Implementing the trigger would at least reduce the potential impact.

It is important to note that for a proper API cache you need the database to be migrated. Not sure if it also needs to be seeded.

Note that I'd love to unify this so everything is using the same methods.

jturel:
Thanks for the detailed info Ewoud.

I've opened an installer PR to move this over there. Seems like there are larger refactorings to be made in this area so let's just do what was working for us before - for now.

https://github.com/theforeman/foreman-installer/pull/375

PR #3938, opened 2019-07-22
Fixes #27255 - katello-change-hostname cleans puppet certs

jturel:
Also fixes an additional bug when changing the hostname on a proxy
caused by giving the installer an invalid parameter

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -63,7 +63,7 @@ module KatelloUtilities
     def get_fpc_answers
       register_in_foreman = false
       certs_tar = @options[:certs_tar]
-      " --foreman-proxy-register-in-foreman #{register_in_foreman} --foreman-proxy-content-certs-tar #{certs_tar}"
+      " --foreman-proxy-register-in-foreman #{register_in_foreman} --certs-tar-file #{certs_tar}"
     end
 
     def precheck
@@ -165,6 +165,12 @@ module KatelloUtilities
       scenario_answers["certs"]["server_cert"] &&
       scenario_answers["certs"]["server_key"] &&
       scenario_answers["certs"]["server_cert_req"]
+    end
+
+    def delete_puppet_certs
+      puppet_ssldir = @scenario_answers['puppet']['ssldir']
+
+      run_cmd("rm -rf '#{puppet_ssldir}'")
     end
 
     def all_custom_cert_options_present?
@@ -334,6 +340,8 @@ module KatelloUtilities
         self.run_cmd("rm -rf #{@scenario_answers["foreman"]["client_ssl_cert"]}")
         self.run_cmd("rm -rf #{@scenario_answers["foreman"]["client_ssl_key"]}")
       end
+
+      delete_puppet_certs
 
       STDOUT.puts "backed up #{public_dir} to #{public_backup_dir}"
       STDOUT.puts "updating hostname in /etc/hosts"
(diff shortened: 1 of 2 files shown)


tbrisker:
[test rpm]

tbrisker:
[test rpm] (please disregard me, checking jenkins config)

tbrisker:
[test rpm]

packages/katello/katello/hostname-change.rb
```diff
@@ -167,6 +167,28 @@ def using_custom_certs?(scenario_answers)
...
+        'ca/inventory.txt',
+        'ca/infra_inventory.txt',
+        'ca/infra_serials',
+        'ca/serial',
+        'ca/root_key.pem',
+        'ca/ca_key.pem'
+      ]
+
+      files.each do |file|
+        run_cmd("rm -rf #{File.join(puppet_ssldir, file)}")
```
ekohl: Wouldn't it be cleaner to use native Ruby functions? Or at least quote the argument.

jturel: I think so, but run_cmd is consistent with everything else in the script and apparently it does some cleanup and handling of any errors

tbrisker:
[test packaging]

packages/katello/katello/hostname-change.rb
```diff
@@ -167,6 +167,28 @@ def using_custom_certs?(scenario_answers)
...
+        'ca/inventory.txt',
+        'ca/infra_inventory.txt',
+        'ca/infra_serials',
+        'ca/serial',
+        'ca/root_key.pem',
+        'ca/ca_key.pem'
+      ]
+
+      files.each do |file|
+        run_cmd("rm -f '#{File.join(puppet_ssldir, file)}'")
```
jturel: @ekohl quoted & changed to -f

ekohl:
> This isn't related to this PR, but I see some JS errors on the products page, but I'm not sure if they were pre-existing before the katello change hostname run. Let me know if you've seen something similar!

These are fallout from the vendor js change.

packages/katello/katello/hostname-change.rb
```diff
@@ -167,6 +167,28 @@ def using_custom_certs?(scenario_answers)
...
+      files = [
+        'ca/ca_crt.pem',
+        'ca/ca_crl.pem',
+        'ca/infra_crl.pem',
+        'certs/ca.pem',
+        'crl.pem',
+        'ca/ca_pub.pem',
+        'ca/inventory.txt',
+        'ca/infra_inventory.txt',
+        'ca/infra_serials',
```
ekohl: I don't see this on my Puppet 5 setup and I think you're using Puppet 6. Do we want to be safe and just remove the entire `puppet_ssldir`? AFAIK Puppet recreates it if needed and you don't need to track individual files.

packages/katello/katello/hostname-change.rb
```diff
@@ -167,6 +167,12 @@ def using_custom_certs?(scenario_answers)
       scenario_answers["certs"]["server_cert_req"]
     end
 
+    def delete_puppet_certs
+      puppet_ssldir = @scenario_answers['puppet']['ssldir']
+
+      run_cmd("rm -rf '#{puppet_ssldir}'")
```
jturel: Updated this to remove the whole folder. It appears to work! I tested it on a nightly proxy. @johnpmitsch can you confirm it works on your katello nightly?

evgeni:
[test rpm]

jturel:
Looks like we're all good here, but I don't have merge access :)

jturel:
> Both are issues in 1.22 as well so candidates for a cherry pick.
> 
> Edit: I'll leave it the pick request up to someone else :)

I flagged the issue for Katello 3.12.2 so we'll include it there

PR #3947, opened 2019-07-23
Update katello packaging to @theforeman/vendor bundle

evgeni:


evgeni:
For this to work properly, `foreman` will also need to be rebuilt with the new (0.1.1) vendor js bundle. Other plugins don't need a rebuild! (tested with tasks 0.16.0-2)

evgeni:
[test rpm]

evgeni:
The repoclosure error is OK, we didn't mash yet. And I can't get rid of the bad old build, so the status will never go green :D

PR #3949, opened 2019-07-23
Adjust selinux for proxmox

mdellweg:

### DIFF
--- a/packages/plugins/rubygem-foreman_fog_proxmox/rubygem-foreman_fog_proxmox.spec
+++ b/packages/plugins/rubygem-foreman_fog_proxmox/rubygem-foreman_fog_proxmox.spec
@@ -8,7 +8,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.8.0
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Summary: Foreman plugin that adds Proxmox VE compute resource using fog-proxmox
 Group: Applications/Systems
 License: GPLv3
@@ -35,6 +35,13 @@ BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 Provides: foreman-plugin-%{plugin_name} = %{version}
 # end specfile generated dependencies
+%if 0%{?fedora} >= 27 || 0%{?rhel} >= 8
+Requires(post):   policycoreutils-python-utils
+Requires(postun): policycoreutils-python-utils
+%else
+Requires(post):   policycoreutils-python
+Requires(postun): policycoreutils-python
+%endif
 
 %description
 Foreman plugin adds Proxmox VE compute resource using fog-proxmox. It is
@@ -103,7 +110,18 @@ cp -pa .%{gem_dir}/* \
 %{foreman_restart}
 exit 0
 
+%post
+/sbin/semanage port -a -t http_port_t -p tcp 8006 &> /dev/null || :
+
+%postun
+if [ $1 -eq 0 ] ; then
+  /sbin/semanage port -d -t http_port_t -p tcp 8006 &> /dev/null || :
+fi
+
 %changelog
+* Tue Jul 23 2019 Matthias Dellweg <dellweg@atix.de> 0.8.0-2
+- Manage selinux permissions for proxmox access
+
 * Fri Jul 12 2019 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> 0.8.0-1
 - Update to 0.8.0-1
 


packages/plugins/rubygem-foreman_fog_proxmox/rubygem-foreman_fog_proxmox.spec
```diff
@@ -103,7 +103,16 @@ cp -pa .%{gem_dir}/* \
 %{foreman_restart}
 exit 0
 
+%post
+semanage port -a -t http_port_t -p tcp 8006
```
dgoetz: Should be `semanage port -a -t http_port_t -p tcp 8006 &> /dev/null || :` to suppress output and always run without error even if SELinux is disabled.

packages/plugins/rubygem-foreman_fog_proxmox/rubygem-foreman_fog_proxmox.spec
```diff
@@ -103,7 +103,16 @@ cp -pa .%{gem_dir}/* \
...
+
+%postun
+semanage port -d -t http_port_t -p tcp 8006
```
dgoetz: Should be `semanage port -d -t http_port_t -p tcp 8006 &> /dev/null || :` to suppress output and always run without error even if SELinux is disabled.

mdellweg: @dgoetz Thank you, changed.

packages/plugins/rubygem-foreman_fog_proxmox/rubygem-foreman_fog_proxmox.spec
```diff
@@ -103,7 +103,16 @@ cp -pa .%{gem_dir}/* \
 %{foreman_restart}
 exit 0
 
+%post
+semanage port -a -t http_port_t -p tcp 8006 &> /dev/null || :
+
+%postun
+semanage port -d -t http_port_t -p tcp 8006 &> /dev/null || :
```
ekohl: I thought postun was also called on upgrades. This means it'll flip flop between on/off/on. https://www.golinuxhub.com/2018/05/how-to-execute-script-at-pre-post-preun-postun-spec-file-rpm.html has some examples on how to check for this.

mdellweg: I guess, you are the expert on rpm packaging questions. `postun` sounded right to me, so i had no reason to question that. Feel free to add any needed change to this PR.

ekohl: I think technically an upgrade is an uninstall + install so it's called in both removals + upgrades. I also wouldn't consider myself an expert ;)

dgoetz:
Give me some minutes and I will do the adjustments. I have working code in the icinga2 packages.

packages/plugins/rubygem-foreman_fog_proxmox/rubygem-foreman_fog_proxmox.spec
mdellweg: maybe prepending `[ $1 == 0 ] && ` is enough.

dgoetz:
I added Requires section based on version as packages changed with Fedora 27 and RHEL 8 and the if condition which does the removal of the label only on uninstall of the last package (read real uninstall, but not upgrade).

packages/plugins/rubygem-foreman_fog_proxmox/rubygem-foreman_fog_proxmox.spec
mdellweg: When i read the article correctly, it is even install + uninstall. So my version would have left the port closed on update.
Thank you both for spotting and repairing this.

packages/plugins/rubygem-foreman_fog_proxmox/rubygem-foreman_fog_proxmox.spec
```diff
@@ -34,6 +34,13 @@ BuildRequires: %{?scl_prefix_ruby}rubygems-devel
 BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 Provides: foreman-plugin-%{plugin_name} = %{version}
+%if 0%{?fedora} >= 27 || 0%{?rhel} >= 8
```
ekohl: Can you move it below the comment? Within the block it's autogenerated and replaced on package bumping so it'd be dropped on the next package update.

mdellweg: like this?

ekohl: Exactly.

lzap:
> I wonder if this should be done in https://github.com/theforeman/foreman-selinux instead. @lzap?

I don't mind it here. Looks good.

PR #3954, opened 2019-07-24
Fixes #27401 - katello_repositories debug sql query fixed

lzap:

* [x] Nightly

---

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -4,7 +4,7 @@
 %global homedir %{_datarootdir}/%{name}
 %global confdir common
 %global prerelease .master
-%global release 1
+%global release 2
 
 Name:       katello
 Version:    3.14.0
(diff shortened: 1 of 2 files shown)


jturel:
Looks good @lzap I think we need to bump the release in the spec file as well

lzap:
Done!

packages/katello/katello/katello.spec
```diff
@@ -194,6 +194,9 @@ Useful utilities for managing Katello services
 %{_sysconfdir}/bash_completion.d/katello-service
 
 %changelog
+* Tue Aug 13 2019 Lukas Zapletal <lzap+rpm@redhat.com> - 3.14.0-1.1.master
```
jturel: It seems the release went out of sync with the version indicated in the changelog, possibly when 3.14 was bumped. @evgeni should this be 3.14.0-2.1.master now?

evgeni: -0.2.master

lzap: Updated the changelog

jturel:
@evgeni pls2merge? :)

PR #3955, opened 2019-07-24
Add python-apypie

evgeni:

### DIFF
--- a/packages/client/python-apypie/python-apypie.spec
+++ b/packages/client/python-apypie/python-apypie.spec
@@ -45,7 +45,7 @@ Requires:       python-requests
 Apipie bindings for Python2
 %endif
 
-%if %{with python2}
+%if %{with python3}
 %package -n     python%{python3_pkgversion}-%{pypi_name}
 Summary:        Apipie bindings for Python
 Requires:       python%{python3_pkgversion}-requests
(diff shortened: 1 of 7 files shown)


packages/client/python-apypie/python-apypie.spec
```diff
@@ -0,0 +1,65 @@
...
+
+License:        MIT
+URL:            https://github.com/Apipie/apypie
+Source0:        https://files.pythonhosted.org/packages/source/a/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+BuildArch:      noarch
+
+BuildRequires:  python2-devel
+BuildRequires:  python2-setuptools
+
+BuildRequires:  python%{python3_pkgversion}-devel
```
evgeni: Contrary to EPEL7, our builders don't have Python3 support. So I'll have to disable that.

packages/client/python-apypie/python-apypie.spec
```diff
@@ -0,0 +1,94 @@
...
+%if %{with python2}
+%package -n     python2-%{pypi_name}
+Summary:        Apipie bindings for Python
+Requires:       python-requests
+
+%description -n python2-%{pypi_name}
+Apipie bindings for Python2
+%endif
+
+%if %{with python2}
```
ekohl:
```suggestion
%if %{with python3}
```

PR #3959, opened 2019-07-25
Updated infoblox DNS to 1.0.0

lzap:

* [x] Nightly

---


evgeni:
@lzap given we now use the `timeout` param (via https://github.com/theforeman/smart_proxy_dns_infoblox/commit/4b4fffe996110e3e2cfd0600c413167e0412d273), should this depend on the 3.0 version of the gem?

lzap:
Yes, it will not break without it but I should bump it. It's missing in the gemfile completely, I will add that change for the next release and push out a gem only release 1.0.1. No need of updating RPM for this one.

Pushed the spec change now.

dgoetz:
[test rpm]

lzap:
Damn its 404 already, [test rpm]

lzap:
Green!

PR #3961, opened 2019-07-25
Update rubygem-hammer_cli_katello to 0.18.2

chris1984:


akofink:
For transparency: we skipped 0.18.1 as there was a messed up deployment to rubygems.org that we had to `gem yank` back.

PR #3963, opened 2019-07-30
Bump RPM versions to 1.24-develop

evgeni:

### DIFF
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,9 +1,9 @@
-%global release 2
+%global release 1
 %global prerelease develop
 
 Name:       foreman-installer
 Epoch:      1
-Version:    1.23.0
+Version:    1.24.0
 Release:    %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?nightly}%{?dist}
 Summary:    Puppet-based installer for The Foreman
 Group:      Applications/System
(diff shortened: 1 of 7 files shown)


evgeni:
build obviously fails as the sources are not bumped to 1.24 yet :)

tbrisker:
[test rpm] now they are

evgeni:
So @ekohl pointed out on IRC that this PR failed repoclosure (https://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-rpm-pr-test/detail/foreman-packaging-rpm-pr-test/756/pipeline/):

```
TASK [repoclosure : Run repoclosure for fedora29] ******************************
fatal: [foreman]: FAILED! => 
  msg: |-
    The task includes an option with an undefined variable. The error was: 'dict object' has no attribute u'fedora29'
  
    The error appears to be in '/home/jenkins/workspace/foreman-packaging-rpm-pr-test/obal/obal/data/roles/repoclosure/tasks/repoclosure.yml': line 5, column 3, but may
    be elsewhere in the file depending on the exact syntax problem.
  
    The offending line appears to be:
  
  
    - name: 'Run repoclosure for {{ dist }}'
      ^ here
    We could be wrong, but this one looks like it might be an issue with
    missing quotes. Always quote template expression brackets when they
    start a value. For instance:
  
        with_items:
          - {{ foo }}
  
    Should be written as:
  
        with_items:
          - "{{ foo }}"

PLAY RECAP *********************************************************************
foreman                    : ok=13   changed=1    unreachable=0    failed=1    skipped=1    rescued=0    ignored=0   
```

`fedora29` you say, wtf? Right. This PR contained multiple packages, some of them are built for *all* targets (`foreman-release`) while others are built only for EL7 (`foreman` etc). But then we download *all* builds, place them into the `downloaded_rpms` folder and use *that* to calculate which targets we need to run repoclosure against. Well, and then we run f29 repoclosure for foreman, which is undefined and breaks stuff.

TL;DR: nothing to see here besides a far from perfect repoclosure check. blame @evgeni for trying to make it smarter.

PR #3964, opened 2019-07-30
update configs to use stable 1.23 repos/tags/etc

evgeni:


evgeni:
and this is missing an update of the foreman gpg keys

evgeni:
now with 100% more correct keys

PR #3965, opened 2019-07-30
use pulp 2.19 for repoclosure

evgeni:


evgeni:
> we should also be using it in 1.23-stable i believe

yes, this change was also part of #3964 ;)

PR #3969, opened 2019-07-31
Bump rubygem-hammer_cli_foreman_discovery to 1.0.1-1

ekohl:
Also regenerates the spec file. This has the added benefit is that it doesn't install the config file as executable and automatically updates the dependencies on the next bump.


lzap:
LGTM thanks.

PR #3970, opened 2019-07-31
Fixes #27255 - katello-change-hostname cleans puppet certs

chris1984:
Also fixes an additional bug when changing the hostname on a proxy
caused by giving the installer an invalid parameter

(cherry picked from commit 58b3bd781b8e67dbbf4d3f694d2ddb96f218f6b7)

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -4,7 +4,7 @@
 %global homedir %{_datarootdir}/%{name}
 %global confdir common
 #global prerelease .rc2
-%global release 1
+%global release 2
 
 Name:       katello
 Version:    3.12.1
@@ -194,6 +194,10 @@ Useful utilities for managing Katello services
 %{_sysconfdir}/bash_completion.d/katello-service
 
 %changelog
+* Sun Jul 21 2019 Jonathon Turel <jturel@gmail.com> 3.12.1-2
+- katello-change-hostname: clean puppet certs
+- katello-change-hostname: s/foreman-proxy-content-certs-tar/certs-tar-file/
+
 * Wed Jul 17 2019 Evgeni Golov - 3.12.1-1
 - Release 3.12.1
 
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -63,7 +63,7 @@ module KatelloUtilities
     def get_fpc_answers
       register_in_foreman = false
       certs_tar = @options[:certs_tar]
-      " --foreman-proxy-register-in-foreman #{register_in_foreman} --foreman-proxy-content-certs-tar #{certs_tar}"
+      " --foreman-proxy-register-in-foreman #{register_in_foreman} --certs-tar-file #{certs_tar}"
     end
 
     def precheck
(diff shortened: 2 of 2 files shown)


chris1984:
@jturel here is the foreman-packaging pr

packages/katello/katello/katello.spec
```diff
@@ -194,8 +194,12 @@ Useful utilities for managing Katello services
 %{_sysconfdir}/bash_completion.d/katello-service
 
 %changelog
-* Wed Jul 17 2019 Evgeni Golov - 3.12.1-1
-- Release 3.12.1
+* Sun Jul 21 2019 Jonathon Turel <jturel@gmail.com> 3.13.0-0.4.master
```
jturel: Changelog doesn't look quite right. I think it should be a single changelog entry for 3.12.1-2 ?

chris1984:
@jturel updated

packages/katello/katello/katello.spec
```diff
@@ -194,8 +194,9 @@ Useful utilities for managing Katello services
 %{_sysconfdir}/bash_completion.d/katello-service
 
 %changelog
-* Wed Jul 17 2019 Evgeni Golov - 3.12.1-1
-- Release 3.12.1
+* Sun Jul 21 2019 Jonathon Turel <jturel@gmail.com> 3.12.1-2
```
jturel: This entry should be above Evgeni's for 3.12.1-1 but other than that I think it's good

chris1984:
@jturel updated, thanks for catching that

jturel:
> @jturel updated, thanks for catching that

np!

need to update the release to '2' as well:

https://github.com/theforeman/foreman-packaging/blob/8fec8a8800227401196659d60d27b9658130e84d/packages/katello/katello/katello.spec#L7

chris1984:
@jturel tests are green now, I think we are good for this one.

packages/katello/katello/katello.spec
```diff
@@ -194,6 +194,10 @@ Useful utilities for managing Katello services
 %{_sysconfdir}/bash_completion.d/katello-service
 
 %changelog
+* Sun Jul 21 2019 Jonathon Turel <jturel@gmail.com> 3.12.1-2
```
evgeni: Is that supposed to go out as a fix to 3.12.1? Or should this be part of 3.12.2?

If the latter, I'd just include/update this PR when I'll be doing the overall 3.12.2 packaging for you.

jturel: It's for 3.12.2

Can you reword your suggestion - want to make sure I'm following

evgeni: Chris said there will be a 3.12.2 soon. I think there is no reason to release a 3.12.1-2 now just to be overwritten by 3.12.2-1 days later.

So I'd just fold this PR into the 3.12.2 one as soon that's open (and leave this open until then for documentation purposes).

jturel: Makes sense! thx for the guidance

ekohl:
Closing in favor of https://github.com/theforeman/foreman-packaging/pull/3989

PR #3973, opened 2019-08-01
Update katello-host-tools to match SUSE build service

jturel:
I was updating katello-host-tools for SUSE to get a SLES15 build working
and these were the changes I had to make.

Result:
https://build.opensuse.org/package/show/systemsmanagement:katello/katello-host-tools

### DIFF
--- a/packages/client/katello-host-tools/katello-host-tools.spec
+++ b/packages/client/katello-host-tools/katello-host-tools.spec
@@ -2,7 +2,7 @@
 %global yum_install ((0%{?rhel} <= 7) && (0%{?rhel} >= 5)) || ((0%{?fedora} < 27) && (0%{?fedora} > 0))
 %global zypper_install (0%{?suse_version} > 0)
 
-%global build_tracer 0%{?rhel} >= 7 || 0%{?fedora}
+%global build_tracer (0%{?suse_version} > 0 ) || ((0%{?fedora}) || (0%{?rhel} >= 7))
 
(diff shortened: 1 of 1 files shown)


packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -2,7 +2,7 @@
 %global yum_install ((0%{?rhel} <= 7) && (0%{?rhel} >= 5)) || ((0%{?fedora} < 27) && (0%{?fedora} > 0))
 %global zypper_install (0%{?suse_version} > 0)
 
-%global build_tracer 0%{?rhel} >= 7 || 0%{?fedora}
+%global build_tracer (0%{?suse_version} > 0 ) || ((0%{?fedora}) || (0%{?rhel} >= 7))
```
evgeni: That means we support Tracer on SUSE now?

jturel: Yes! It's not using the tracer library, though. It's much more basic: https://github.com/Katello/katello-host-tools/blob/master/src/zypper_plugins/tracer_upload.py

evgeni: Nice.

ekohl: Isn't the `> 0` redundant? I think logically the following is equivalent:
```suggestion
%global build_tracer 0%{?rhel} >= 7 || 0%{?fedora} || 0%{?suse_version}
```

jturel: While I preferred the inconsistent, confusing, and needlessly verbose solution I'll take your suggestion :smile:

PR #3976, opened 2019-08-01
Update rubygem-hammer-cli to 0.18.0

mbacovsky:


mbacovsky:
@ekohl anything I can do about the failing tests? It does not seem related.

PR #3983, opened 2019-08-02
katello 3.13 packaging picks

evgeni:


evgeni:
this PR is only here to sync the git repo with reality as those changes are in the katello-3.13-rhel7 tag on koji (see https://github.com/theforeman/tool_belt/pull/191 for details why)

PR #3990, opened 2019-08-08
Update rubygem-foreman_virt_who_configure to 0.4.3

ares:


ekohl:
There's files added that I can't place:
```
Compared versions: ["0.4.1", "0.4.2"]
  DIFFERENT files:
    0.4.1->0.4.2:
      * Added:
            locale/en/foreman_virt_who_configure.edit.po
              (!) Unexpected permissions: 100664
            locale/en/foreman_virt_who_configure.po.time_stamp
              (!) Unexpected permissions: 100664
```

Are these intended?

ares:
oh, there's missing .gitignore and I accidentally pushed that to the gem, we can ignore these in spec for now

ekohl:
I think gem build ignores .gitignore and just takes what's in the gemspec and on the filesystem.

ares:
ah yeah, I extracted strings this time from release repo, that's why, do you need me to release new version?

ekohl:
It would be cleaner.

ares:
that will have to wait at least one week, I can't connect to my host in VPN from here :-(

ares:
updated to 0.4.3 which shouldn't have any extra files

PR #3997, opened 2019-08-09
Katello 3.13.0 RC1

evgeni:


evgeni:
I wouldn't know :)
But I can check with @jjeffers

ekohl:
It's something that has failed for the past few releases and then only to surface at the GA release. Let's try to get it right here this time :)

evgeni:
FTR, I see no Katello 3.13 key in the Katello keys repo, so I guess this was not updated yet.

jjeffers:
@evgeni I'll follow up wrt to the GPG keys.

evgeni:
@jjeffers @jlsherrill following the discussion, I've updated this to include the Foreman pubic key as the Katello public key.

This will need a section in the release announcement I guess.

evgeni:
@ekohl yeah I thought similar, but decided not to touch it right now.

evgeni:
repoclosure is broken, expected :)

PR #3999, opened 2019-08-09
regenerate hammer_cli_katello spec

evgeni:

### DIFF
--- a/packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
+++ b/packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
@@ -1,54 +1,55 @@
+# Generated from hammer_cli_katello-0.19.0.gem by gem2rpm -*- rpm-spec -*-
+# template: hammer_plugin
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
-%{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
-
 %global gem_name hammer_cli_katello
-%global confdir hammer
+%global plugin_name katello
 
-%global release 1
+%global release 2
 %global prerelease .pre.master
 
-Summary: Katello command plugin for the Hammer CLI
-Name:    %{?scl_prefix}rubygem-%{gem_name}
+%{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
+%global hammer_confdir %{_root_sysconfdir}/hammer
+
+Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.19
 Release: %{?prerelease:0.}%{release}%{?prerelease}%{?nightly}%{?dist}
-Group:   Development/Languages
+Summary: Katello commands for Hammer
+Group: Development/Languages
 License: GPLv3
-URL:     https://github.com/theforeman/hammer-cli-katello
+URL: https://github.com/Katello/hammer-cli-katello
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}%{?prerelease}.gem
 
-BuildArch: noarch
-Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
-%if 0%{?scl:1}
-Obsoletes: rubygem-%{gem_name} < 0.0.17-3
-Obsoletes: rubygem-hammer_cli_gutterball
-%endif
-%{?scl:Obsoletes: ruby193-rubygem-%{gem_name}}
-
+# start specfile generated dependencies
 Requires: %{?scl_prefix_ruby}ruby(release)
+Requires: %{?scl_prefix_ruby}ruby
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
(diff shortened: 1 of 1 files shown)


packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
```diff
@@ -1,54 +1,55 @@
...
+
+Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.19
 Release: %{?prerelease:0.}%{release}%{?prerelease}%{?nightly}%{?dist}
-Group:   Development/Languages
+Summary: Katello commands for Hammer
+Group: Development/Languages
 License: GPLv3
-URL:     https://github.com/theforeman/hammer-cli-katello
+URL: https://github.com/Katello/hammer-cli-katello
```
evgeni: Funny this was never noticed :)

packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
```diff
@@ -74,26 +75,31 @@ gem build %{gem_name}.gemspec
...
+
 %files
 %dir %{gem_instdir}
-%{gem_instdir}/lib
+%{gem_libdir}
 %{gem_instdir}/locale
-%config(noreplace) %{_root_sysconfdir}/%{confdir}/cli.modules.d/katello.yml
 %exclude %{gem_cache}
 %{gem_spec}
+%config %{hammer_confdir}/cli.modules.d/%{plugin_name}.yml
```
evgeni: `noreplace` was dropped, not sure that's correct

packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
```diff
@@ -1,54 +1,55 @@
...
-
+# start specfile generated dependencies
 Requires: %{?scl_prefix_ruby}ruby(release)
+Requires: %{?scl_prefix_ruby}ruby
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
-Requires: %{?scl_prefix}rubygem(hammer_cli_foreman) >= 0.15
+Requires: %{?scl_prefix}rubygem(hammer_cli_foreman) >= 0.18.0
 Requires: %{?scl_prefix}rubygem(hammer_cli_foreman) < 1.0.0
-Requires: %{?scl_prefix}rubygem(hammer_cli_foreman_tasks) >= 0.0.12
+Requires: %{?scl_prefix}rubygem(hammer_cli_foreman_tasks)
```
evgeni: This might be missing in the gemspec?

packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
ekohl: I *just* had a draft i was about to submit. I've noticed this in other places as well but I think most hammer plugins actually use `noreplace`. Shall we update the template?

evgeni: `noreplace` means it creates an `.rpmnew` instead of moving the existing one to `.rpmsave`? I can never remember. But yeah. Seems sensible to update the template then.

ekohl: https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html has a nice table. `%config` is `rpmsave`. `%config(noreplace)` is `.rpmnew`.

packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
ekohl: @akofink might know but let's match packaging to the gemspec.

ekohl:
The build failed because it couldn't retrieve the source, which is weird because it should be using the nightly releaser that pulls it from Jenkins.

evgeni:
bumped.

packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
akofink: It's there https://github.com/Katello/hammer-cli-katello/blob/master/hammer_cli_katello.gemspec#L52

evgeni: @akofink The question was whether the ` >= 0.0.12` is important or not -- the gemspec does not have a version defined.

akofink: Oh, I'm actually not sure. They hardly ever release hammer_cli_foreman_tasks

evgeni: Then let's stick to what is in the gemspec and if somebody will complain, we can re-introduce the version again :)

evgeni: Thanks!

PR #4006, opened 2019-08-12

ekohl:


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -44,7 +44,7 @@ Requires: %{?scl_prefix}rubygem(runcible) < 3.0.0
 Requires: %{?scl_prefix}rubygem(anemone)
 Requires: %{?scl_prefix}rubygem(pulpcore_client)
 Requires: %{?scl_prefix}rubygem(pulp_file_client) < 3.0.0
-Requires: %{?scl_prefix}rubygem(pulp_ansible_client)
+Requires: %{?scl_prefix}rubygem(pulp_ansible_client) = 0.2.0b2.dev01565187947
```
evgeni: I don't think `=` will work nicely here, as the version has no revision/dist?

ekohl:
https://github.com/Katello/katello/commit/bd04bc6d700838c6e86ef0ee8f41b61c1c57da58 dropped the requirement so there's no more change needed.

PR #4012, opened 2019-08-15
Update rubygem-dynflow to 1.3.0

adamruzicka:
Is the second commit needed? In previous versions we depended on ruby >= 2.0.0, but the dependency was not mentioned anywhere.


ekohl:
In https://github.com/theforeman/foreman-packaging/pull/4030 I've regenerated the spec file so in the future it automatically updates the dependencies, including Ruby itself.

PR #3438, opened 2019-02-08
Update foreman-tasks-core to 0.3.0 (deb-proxy)

iNecas:
For plugin updates, please indicate which repos this should be built into:

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

See Foreman's [plugin maintainer documentation](https://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Release-strategies) for more information.

---

PR #4013, opened 2019-08-15
Update dynflow to 1.3.0

adamruzicka:


mmoll:
Blocked for the same reason as GH-3438, AFAICT.

PR #4018, opened 2019-08-18
Update rubygem-foreman_remote_execution to 1.8.3

iNecas:
Adds rubygem-foreman_remote_execution-cockpit subpackage with the cockpit integration bits and pieces

* [x] Nightly
* [ ] 1.21
* [ ] 1.20

---

### DIFF
--- a/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
+++ b/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
@@ -104,7 +104,6 @@ install -Dp -m0644 %{buildroot}%{gem_instdir}/extra/cockpit/settings.yml.example
 exit 0
 
 %post cockpit
-%systemd_postun_with_restart foreman-cockpit.service
 %systemd_post foreman-cockpit.service
 
 %preun cockpit
@@ -112,7 +111,6 @@ exit 0
 
 %postun cockpit
 %systemd_postun_with_restart foreman-cockpit.service
-%systemd_post foreman-cockpit.service
 
 %files
 %dir %{gem_instdir}
--- a/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
+++ b/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
@@ -42,6 +42,7 @@ BuildRequires: %{?scl_prefix}rubygem(foreman-tasks) >= 0.15.1
 BuildRequires: %{?scl_prefix}rubygem(foreman-tasks) < 1.0.0
 BuildRequires: %{?scl_prefix}rubygem(foreman_remote_execution_core)
 BuildRequires: %{?scl_prefix}rubygem(deface)
+BuildRequires: systemd
 
 BuildArch: noarch
 
@@ -52,6 +53,17 @@ Provides: foreman-plugin-remote_execution
 %description
 A plugin bringing remote execution to the Foreman, completing the config
 management functionality with remote management functionality
+
+%package cockpit
+BuildArch:  noarch
+Requires:   cockpit
+Requires(post): systemd-units
+Requires(preun): systemd-units
+Summary:    Cockpit integration using remote execution connection
+
+%description cockpit
+This package contains files related to Cockpit, mainly foreman-cockpit service
+and corresponding configuration files.
 
 %package doc
 BuildArch:  noarch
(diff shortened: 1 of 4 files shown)


packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -72,19 +85,33 @@ This package contains documentation for rubygem-%{gem_name}.
...
+
 %posttrans
 %{foreman_db_migrate}
 %{foreman_db_seed}
 %{foreman_apipie_cache}
 %{foreman_restart}
 exit 0
 
+%post cockpit
+%systemd_postun_with_restart foreman-cockpit.service
```
ekohl: Why is this here?

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -72,19 +85,33 @@ This package contains documentation for rubygem-%{gem_name}.
...
+%systemd_post foreman-cockpit.service
+
+%postun cockpit
+%systemd_postun_with_restart foreman-cockpit.service
+%systemd_post foreman-cockpit.service
```
ekohl: This also surprises me

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -53,6 +54,18 @@ Provides: foreman-plugin-remote_execution
 A plugin bringing remote execution to the Foreman, completing the config
 management functionality with remote management functionality
 
+%package cockpit
+BuildArch:  noarch
+Requires:   cockpit
+Requires(post): systemd-sysv
```
ekohl: I wonder why this is needed. I thought it's only for legacy sysv init scripts but this is a native systemd service.

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
iNecas: Inspired heavily by https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/foreman/foreman.spec#L1031

Expands to:

```
systemctl daemon-reload >/dev/null 2>&1 || : 
if [ $1 -ge 1 ] ; then 
        # Package upgrade, not uninstall 
        systemctl try-restart foreman-cockpit.service >/dev/null 2>&1 || : 
fi 


if [ $1 -eq 1 ] ; then 
        # Initial installation 
        systemctl preset foreman-cockpit.service >/dev/null 2>&1 || : 
fi 
```

Looks useful

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
iNecas: Ditto

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
iNecas: Because we have it in foreman.spec: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/foreman/foreman.spec#L52 seems like it should not be there as well. Updated this spec

iNecas:
The `%preun` macros added

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
ekohl: I think that one in foreman.spec is already incorrect and we should rather be using `%systemd_post` in `%post`.

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
ekohl: The `%systemd_postun_with_restart` should be sufficient here.

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
iNecas: Updated

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
iNecas: Updated

ehelms:
[test rpm]

PR #4022, opened 2019-08-20
Update foreman_remote_execution to 1.8.4

adamruzicka:

### DIFF
--- a/plugins/ruby-foreman-remote-execution/debian/install
+++ b/plugins/ruby-foreman-remote-execution/debian/install
@@ -1,5 +1,5 @@
 foreman_remote_execution.rb           usr/share/foreman/bundler.d
 cache                                 usr/share/foreman/vendor
 assets                                var/lib/foreman/public
-webpack                               var/lib/foreman/public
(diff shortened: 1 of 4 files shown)


plugins/ruby-foreman-remote-execution/debian/install
```diff
@@ -1,5 +1,5 @@
 foreman_remote_execution.rb           usr/share/foreman/bundler.d
 cache                                 usr/share/foreman/vendor
 assets                                var/lib/foreman/public
-webpack                               var/lib/foreman/public
+webpack/foreman_remote_execution      var/lib/foreman/public/webpack
```
adamruzicka: Interesting, I wasn't aware I'd have to change anything else than the versions

mmoll: See GH-4025 for the full story, I just wanted to take the opportunity to bring this in with the udpate.

adamruzicka: Ok, makes sense. Thanks for stepping in

mmoll:
merged, díky @adamruzicka!

PR #4023, opened 2019-08-20
Add npm @theforeman/vendor update docs

sharvit:
Explain about updating `@theforeman/vendor`


sharvit:
@ekohl @tbrisker would you like to add this commit?
I guess it is not necessary but because the vendor is going to be the main package it would be nice to add a more specific instruction.

PR #4024, opened 2019-08-20
Update nodejs-thforeman-vendor to v1.4.0

sharvit:
PR in foreman: https://github.com/theforeman/foreman/pull/6988


tbrisker:
This will also need another PR to update the foreman rpm spec to accept 1.x - it is currently limited to <1 iirc

sharvit:
> This will also need another PR to update the foreman rpm spec to accept 1.x - it is currently limited to <1 iirc

Thanks!
How do I do that? where is it defined?

PR #4027, opened 2019-08-21
use the bindir defined in gemspec, not "bin"

evgeni:
some gems use "exe" and similar oddities


evgeni:
this was found while working on #4026

PR #4032, opened 2019-08-22
Add foreman azure rm 2.0.0.pre1

apuntamb:
Supported Versions:

 * Nightly
 * 1.23
 * 1.22
 * 1.21


packages/plugins/rubygem-foreman_azure_rm/rubygem-foreman_azure_rm.spec
```diff
@@ -0,0 +1,113 @@
...
+#   2. Delete these lines
+#
+# Generated from foreman_azure_rm-2.0.0.pre1.gem by gem2rpm -*- rpm-spec -*-
+# template: foreman_plugin
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
+%global gem_name foreman_azure_rm
+%global plugin_name azure_rm
+%global foreman_min_version FIXME
```
ekohl: Could you fill this in and then remove the top 4 lines?

packages/plugins/rubygem-foreman_azure_rm/rubygem-foreman_azure_rm.spec
```diff
@@ -0,0 +1,113 @@
...
+
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 2.0.0.pre1
+Release: 1%{?foremandist}%{?dist}
+Summary: Azure Resource Manager as a compute resource for The Foreman
+Group: Applications/Systems
+License: GPL-3.0
```
ekohl: Open in issue in our tooling, but the format gems and RPMs use differs:
```suggestion
License: GPLv3
```

packages/plugins/rubygem-foreman_azure_rm/rubygem-foreman_azure_rm.spec
apuntamb: done!

packages/plugins/rubygem-foreman_azure_rm/rubygem-foreman_azure_rm.spec
apuntamb: updated.

ekohl:
[test rpm] looks like the job was killed for some reason

apuntamb:
> [test rpm] looks like the job was killed for some reason

As per the logs, it seems the `git annex whereis` command shows failure. I'm not sure why is this caused. What do you think could be the solution to this?

ShimShtein:
[test rpm]

ekohl:
Looks like the build is getting killed after 2 hours because that's the maximum time we allow for such a job. I wonder why it's taking that long.

ShimShtein:
No idea, and don't see anything in Jenkins log. Are there additional files to look into?

ekohl:
I've tried to look into it, but I can't really explain it. Maybe it's because of their size and the amount of files. Looks like the SDKs have multiple versions in them making them quite large. RPMs do post processing to see which dependencies are there and that may be slow.

apuntamb:
> I've tried to look into it, but I can't really explain it. Maybe it's because of their size and the amount of files. Looks like the SDKs have multiple versions in them making them quite large. RPMs do post processing to see which dependencies are there and that may be slow.

Okk, yep they do have some multiple versions. Then how do you suggest to go about it?

ekohl:
It's probably best to split up the PR into multiple smaller PRs to work around it. I can't think of a better workaround.

evgeni:
@apuntamb can you please rebase this ontop of `rpm/develop` now that #4179 is merged?

evgeni:
obsoleted by https://github.com/theforeman/foreman-packaging/pull/4257

PR #4040, opened 2019-08-28
Update rubygem-audited to 4.9.0

tbrisker:


tbrisker:
repoclosure failed as we currently require <4.9 in foreman, once this is in we'll need to also update the foreman spec.

ekohl:
In https://github.com/theforeman/foreman-packaging/pull/4041 I've regenerated the spec file to allow future updates to automatically update the dependencies. That one also has corrected the deps.

PR #4042, opened 2019-08-28
Fixes #27727 - Add foreman service to run as puma for DEB

ekohl:
(cherry picked from commit ab3a75c954f616944199a68a2a76b86acd15aa6f)

### DIFF
--- a/debian/bionic/foreman/control
+++ b/debian/bionic/foreman/control
@@ -230,3 +230,14 @@ Description: metapackage providing journald dependencies for Foreman
  a flexible systems management web application.
  .
  The foreman package should be installed along with this package.
+
+Package: foreman-service
+Architecture: all
+Section: web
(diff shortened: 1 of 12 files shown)


ekohl:
@evgeni @mmoll we intended to include this in 1.22.0 but I failed to cherry pick it in time. I think this is safe to do in 1.22.1 but would like an explicit ack.

@tbrisker I've created a Redmine issue to properly track this and linked it manually to the proper commit.

ekohl:
After more testing it looks like the foreman-service was missing something. Opened https://github.com/theforeman/foreman-packaging/pull/4044 which will also need a cherry pick into 1.23.

ekohl:
Build is failing because we need to pin dependencies. Those are part of 1.22-stable now but unreleased.

PR #4053, opened 2019-08-30
smart_proxy_dhcp_infoblox-0.0.15

lzap:

* [x] Nightly

---


ekohl:
I've taken the liberty to open https://github.com/theforeman/foreman-packaging/pull/4054 which cleans up a few more things by regenerating from the template. This gets rids of EL6 compatibility code and also updates the gem dependencies.

PR #4057, opened 2019-09-03
Update puppet-agent-puppet-strings to 2.3.0

dgoetz:
This will allow to use puppet-strings for determine custom types and perhaps help with https://github.com/theforeman/puppet-foreman/pull/601


ekohl:
I should have commented on this before, but there was a very good reason not to do this. This version requires Ruby 2.1.0. I haven't looked at whether this really is a problem, but it may lead to some odd things. This package in general is odd since it installs the gem into the Puppet-bundled-Ruby.

https://github.com/theforeman/puppet-foreman/pull/601 is a limitation in Kafo since it has custom code to parse types which has a bug.

dgoetz:
Correct me if I am wrong. Using the puppet packaged gem means using their ruby version which is 2.4.5p335. So I would not expect any problem. I only did some short testing but with the modules I tested the only difference in json output was the added data types.

ekohl:
You're right. It actually only "installs" on the main system by placing a gem file. I was under the impression it also installed it like anormal RPM. In that case this is fine.

PR #4058, opened 2019-09-03
Drop katello-remove from katello

ehelms:


ehelms:
Here ya go -- https://projects.theforeman.org/issues/27780

ehelms:
@ekohl Is this sufficient now?

ekohl:
The tricky part is that if you use `Fixes #27780` in the commit message it no longer closes the issue. Technically it needs to be in Foreman -> Packaging.

chris1984:
This is a sad day :( All that work I put into it. At least it will live on in my gists

ehelms:
@ekohl I thought the idea for the issue was to show up in Katello's changelog? IF we move to packaging it won't.

ekohl:
It's a bit of a challenge we have and not unique to Katello but rather all plugins. I'm unsure how to properly do this.

One thought I have is another issue and make it block the removal in Katello, but that feels like a lot of bureaucracy.

ehelms:
For this, I would prefer to close the issue myself once this is merged, leaving it under Katello. So that for Katello's changelog, which I believe is the best place for this to live, the change is visible to users of that plugin.

PR #4073, opened 2019-09-09
Bump rubygem-foreman_inventory_upload to 1.0.0.beta2-1

ShimShtein:
Bump version to beta2.

Main change: add `webpack` and `package.js`


packages/katello/rubygem-foreman_inventory_upload/rubygem-foreman_inventory_upload.spec
```diff
@@ -32,6 +35,78 @@ Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
...
+# start package.json devDependencies BuildRequires
+BuildRequires: npm(babel-cli) >= 6.10.1
+BuildRequires: npm(babel-cli) < 7.0.0
+BuildRequires: npm(babel-core) >= 6.26.3
+BuildRequires: npm(babel-core) < 7.0.0
+BuildRequires: npm(babel-loader) >= 7.1.1
+BuildRequires: npm(babel-loader) < 8.0.0
+BuildRequires: npm(babel-plugin-lodash) >= 3.3.4
+BuildRequires: npm(babel-plugin-lodash) < 4.0.0
+BuildRequires: npm(babel-plugin-module-resolver) >= 3.2.0
```
ekohl: Not sure why it can't find this package since it should be available.

packages/katello/rubygem-foreman_inventory_upload/rubygem-foreman_inventory_upload.spec
```diff
@@ -32,6 +35,78 @@ Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
...
+BuildRequires: npm(identity-obj-proxy) < 4.0.0
+BuildRequires: npm(jed) >= 1.1.1
+BuildRequires: npm(jed) < 2.0.0
+BuildRequires: npm(node-sass) >= 4.5.0
+BuildRequires: npm(node-sass) < 5.0.0
+BuildRequires: npm(patternfly) >= 3.58.0
+BuildRequires: npm(patternfly) < 4.0.0
+BuildRequires: npm(raf) >= 3.4.0
+BuildRequires: npm(raf) < 4.0.0
+BuildRequires: npm(react-animated-tree) >= 1.0.10
```
ekohl: This one isn't packaged and needs to be submitted as a separate PR first.

On a side note: I would expect this to be a runtime dependency rather than a build dep, but that's not very important because the whole build vs runtime concept rather breaks down with webpack.

ShimShtein: We're getting rid of this component, since it's not used in our code  right now. Since it will require a rubygems release, closing this PR in favour of the one without the component.

PR #4080, opened 2019-09-09
Update rubygem-redhat_access to 2.2.8

ares:

PR #4081, opened 2019-09-09
Bump rubygem-redhat_access to 2.2.8-1

ekohl:
This also regenerates the spec based on the current template. This may not work: the comments said it can't compile assets.


ekohl:
Needs https://github.com/theforeman/foreman-packaging/pull/4082

ares:
LGTM, @rexwhite any concern from your side re backporting?

ekohl:
[test rpm]

ekohl:
[test rpm] because I was too soon

ares:
I spoke with Rex, 1.1.5 is backwards compatible with 1.1.4, so no concern there. Seems rpm didn't build for some reason

ekohl:
```
BUILDSTDERR: NameError: uninitialized constant RedhatAccess::Engine::ForemanTasks
```

This makes me think that the dependency on `foreman-tasks` that was in packaging but not in the gemspec is actually there. Probably the same with `katello`.

ekohl:
Repoclosure failed because we haven't mashed a repo with the redhat_access_lib package. However, it would be good to test if the rewritten version that uses compiled assets actually works well. @ares do you have a setup to test this?

ehelms:
@ares Checking back with you

ehelms:
[test rpm]

ares:
I installed with few days old nightlies. UI seems to work fine, just some images on rules pages are missing, but I have no idea if that's related. I'd say :+1:

packages/katello/rubygem-redhat_access/rubygem-redhat_access.spec
```diff
@@ -100,14 +99,17 @@ cp -r  $RPM_BUILD_DIR/%{gem_name}-%{version}/vendor/assets/fonts/*  %{buildroot}
...
 %exclude %{gem_cache}
 %{gem_spec}
 %{foreman_bundlerd_plugin}
-
-%config(noreplace) /etc/redhat_access/config.yml
+%{foreman_apipie_cache_foreman}
+%{foreman_apipie_cache_plugin}
+%{foreman_assets_plugin}
+# The next line is foreman_assets_plugin if the gem name was insights
+%{gem_instdir}/public/assets/insights
```
ekohl: I think this is here because of https://github.com/redhataccess/foreman-plugin/tree/master/redhat-access/vendor/assets/images/insights. It might be the cause of your missing images that Rails can't find them. From `redhat-access/vendor/assets/stylesheets/telemetry/sat6.css`

```
.band.ansible{background:url(../insights/images/bg_ansible.jpg) no-repeat;
```

Perhaps the rails assets pipeline is not rewriting that properly? This is actually important because we send caching headers to indefinitely cache files so in a future upgrade users wouldn't pick up on changes. In the assets pipeline files are named with a hash in the filename to work around this. Because of this, I did want to properly compile them rather than move them around as it did before.

ares: so, what are the next steps?

ares:
If this is not progressing, should I reopen https://github.com/theforeman/foreman-packaging/pull/4080 ?

ekohl:
I'll revert the asset handling since this plugin just has broken assets. If the plugin author cares, they can fix it but for now I'm only going to make our own life easier.

xprazak2:
I think this might be obsoleted by https://github.com/theforeman/foreman-packaging/pull/4521

evgeni:
#4755 was merged

PR #4085, opened 2019-09-09
Bump rubygem-foreman_inventory_upload to 1.0.0.beta3-1

ShimShtein:


packages/katello/rubygem-foreman_inventory_upload/rubygem-foreman_inventory_upload.spec
```diff
@@ -1,4 +1,4 @@
-# Generated from foreman_inventory_upload-1.0.0.beta1.gem by gem2rpm -*- rpm-spec -*-
+# Generated from foreman_inventory_upload-1.0.0.beta3.gem by gem2rpm -*- rpm-spec -*-
```
ekohl: You can drop this line. We don't really use this and at this point it's mostly noise. Note the next one (`# template:`) is important.

ShimShtein:
@ekohl, removed, but `babel-plugin-module-resolver` is still missing. Can we do something about it?

ekohl:
[test rpm] since https://github.com/theforeman/tool_belt/commit/96bd5c234f9a36a1962dca54bea24dfcc24686df should have fixed the build issue. Let's found out if I did it right :)

PR #4094, opened 2019-09-10
Fixes #27811 - satellite-change-hostname fails if locale is set to ja…

sufanek1:
…_JP.UTF-8

satellite-change-hostname script should make sure it uses hammer with English locale only
to prevent possible issues with parsing special characters is locale of OS is set to language which has its own special alphabet. 

Supported Versions:

 * Nightly
 * 1.23
 * 1.22
 * 1.21

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -151,7 +151,7 @@ module KatelloUtilities
     end
 
     def hammer_cmd(cmd, exit_codes=[0], message=nil)
-      run_cmd("hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
+      run_cmd("LANG=en_US.utf-8 hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
     end
 
     def get_default_proxy_id
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -151,7 +151,7 @@ module KatelloUtilities
     end
 
     def hammer_cmd(cmd, exit_codes=[0], message=nil)
-      run_cmd("LANG=en_US.utf-8 hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
+      run_cmd("LANG=en_US.UTF-8 hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
     end
 
     def get_default_proxy_id


ekohl:
I don't like this enforcing a locale. Why is this failing in the first place? Is it a bug in hammer or are we parsing output? The exit code is 64 which is for incorrect usage. Does Hammer translate command line options as well?

cc @mbacovsky

sufanek1:
seems like we are parsing hammer output. Enforcing locale on this specific command solves the issue. 
I am not sure ho difficult would be to re-write parsing part, for use of kanji or other similar alphabets.

sufanek1:
> I don't like this enforcing a locale. Why is this failing in the first place? Is it a bug in hammer or are we parsing output? The exit code is 64 which is for incorrect usage. Does Hammer translate command line options as well?
> 
> cc @mbacovsky

Seems like we are parsing hammer output. Enforcing locale on this specific command workarounds the fact that hammer output is localized. 
I am not sure ho difficult would be to re-write parsing part, for use of kanji or other similar alphabets.

ares:
I think it make sense if we rely on hammer output elsewhere (which is the case). I think we do similar things in the installer. This is internal use of hammer and it sets the locale just for this cmd. @ekohl, what was your concern?

ekohl:
What confused me was the redirection to `/dev/null`. From the Redmine:

```
Failed 'hammer -u admin -p redhat capsule update --id --url https://satellite.jp.test:9090 --new-name satellite.jp.test 2> /dev/null' with exit code 64
```

However, I didn't see it was only redirecting stderr. That said, we don't actually use stdout.

I just don't understand why this one is failing on code 64 (invalid usage) where other hammer commands do pass. My personal rule is that I never merge fixes I don't understand and might be incomplete.

I don't object to others merging this and there's already other bugs that need fixing. For example, it doesn't instruct users to set the `--puppet-server-foreman-url` on the proxies if the user is using Puppet. It's using deprecated commands (`katello-service`, `/sbin/service`). It first restarts puppet and then the puppetserver.

ekohl:
Also, this is probably fine since it's changing the general helper. I think on RH-based we can count on en_US.UTF-8 being present but this isn't true on all platforms.

mbacovsky:
@ekohl the failing command is failing with invalid usage error as there is no value for the id parameter in `hammer -u admin -p redhat capsule update --id --url https://satellite.jp`... The value is missing because its computation failed. When the Japanese  local is used the labels of the values are localized. The label they were looking for to get the id was localized too and they didn't found it under the English name. 

IMO if you need to parse Hammer's output you should run it with fixed locale to have predictable labels in the output.

packages/katello/katello/hostname-change.rb
```diff
@@ -151,7 +151,7 @@ def warning_message
     end
 
     def hammer_cmd(cmd, exit_codes=[0], message=nil)
-      run_cmd("hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
+      run_cmd("LANG=en_US.utf-8 hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
```
ekohl: I do think it's more common to use the uppercase encoding. This is what all the manpages use as examples:
```suggestion
      run_cmd("LANG=en_US.UTF-8 hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
```

ekohl:
Thanks!

PR #4095, opened 2019-09-10
Release Katello 3.13.0 RC2

evgeni:


evgeni:
This currently fails to build (see
http://koji.katello.org/kojifiles/work/tasks/6537/226537/root.log)
because "Error: No Package found for tfm-rubygem(pulp_file_client) <
3.0.0"
And indeed, the packaged version is bigger than 3.0.0
(http://koji.katello.org/koji/buildinfo?buildID=43824) as far as RPM
is concerned (it does sort lexically, without the special rc/beta etc
meaning).

evgeni:
[test rpm]

evgeni:
[test rpm]

evgeni:
I did a `koji untag-build katello-3.13-rhel7 tfm-rubygem-pulp_file_client-3.0.0rc2.dev.1558441126-1.el7` to get the older pulp_file_client back, but it still fails to build the gemspec expects very specific versions of the pulp3 gems: https://github.com/Katello/katello/blob/KATELLO-3.13/katello.gemspec#L45-L48
But we don't have these versions packaged.

I could go and package these exact versions for 3.13, but it seems kinda broken to me to expect a specific dev snapshot of a gem for a
stable release?

jjeffers:
I recall that there was a shift in how the pulp team was creating versions for this gem. For example the 3.x versions were created first, then the later (by date) versions were created with versons 0.1.x.

Would it help if we pinned that gem in RC2 to a specific version we know is good?

evgeni:
I would prefer if we could depend on stable-ish (rc or better) releases of those pulp3 gems. Currently we use daily/dev snapshots, which seems odd to me in a supposedly stable Katello release.

I'd be okay if you pin to a specific RC release.

jjeffers:
As far as I know Pulp3 client gems won't be labelled as RC candidates for some time, but I can follow up with them to confirm that assumption. We're in a pickle as far Katello goes. While there is not officially support for Pulp3 in Katello at the moment, we are in the process of developing those features, and that's why we include the pre-release gems.

I'll follow up with @jsherill as well, but it shouldn't be too difficult to pin an available pulp version.

jlsherrill:
@evgeni they are pre-release gems because pulp3 development is still heavy and the pulp3 api is not yet stable.  Keep in mind we don't support pulp3 yet during installation, the gems are required for various load time constants, but are otherwise unused.  When we start deploying wth pulp3 (and pulp3 has had a stable release), we will be using stable Gems

evgeni:
@jjeffers did you also release rc2.1 as a gem? I don't see it on rubygems.org :(

jjeffers:
@evgeni looks like we did not. I'll follow up with @jturel this morning.

jturel:
Released and tagged.

https://rubygems.org/gems/katello/versions/3.13.0.rc2.1

jturel:
[test rpm]

evgeni:
@jturel this PR still points at rc2, you'll have to update it to look for 2.1 (or I'll do when back at a pc).

jturel:
Oh right. I'll leave it to you, no rush. :)

evgeni:
it built \o/

PR #4097, opened 2019-09-10
Update rubygem-foreman-tasks to 0.16.2

ares:


packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -88,8 +88,6 @@ BuildRequires: npm(humanize-duration) >= 3.20.1
 BuildRequires: npm(humanize-duration) < 4.0.0
 BuildRequires: npm(react-intl) >= 2.8.0
 BuildRequires: npm(react-intl) < 3.0.0
-BuildRequires: npm(c3) >= 0.4.11
```
ekohl: This was manually added to make sure c3 was present. It should have been added to package.json as well so that on the next update it wouldn't break. It could also be that c3 is part of the foreman vendor bundle but then the import needs to be corrected.

ares: oh damn, I somehow mischerrypicked the package.json so it's still not part of 0.16.2, which was the plan :-( I'll restore it for now and cp to 0.16.x branch so we'll have it in next release

PR #4108, opened 2019-09-14
Add rkerberos

ehelms:


packages/foreman/rubygem-rkerberos/rubygem-rkerberos.spec
```diff
@@ -0,0 +1,97 @@
...
+
+%global gem_name rkerberos
+
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.1.5
+Release: 1%{?dist}
+Summary: A Ruby interface for the the Kerberos library
+Group: Development/Languages
+License: Artistic 2.0
+URL: http://github.com/domcleal/rkerberos
```
ekohl:
```suggestion
URL: https://github.com/domcleal/rkerberos
```

package_manifest.yaml
```diff
@@ -161,6 +161,7 @@ foreman_scl_packages:
     rubygem-responders: {}
     rubygem-rest-client: {}
     rubygem-retriable: {}
+    rubygem-rkerberos: {}
```
ekohl: Shouldn't it be under `scl_and_nonscl`?

ehelms: I suppose I can. I was mostly aiming to deliver this in the SCL as prep work.

ehelms:
[test rpm]

ehelms:
Needs https://github.com/theforeman/foreman-packaging/pull/4127

packages/foreman/rubygem-rkerberos/rubygem-rkerberos.spec
evgeni: this is marked as resolved, but itsn't? :)

evgeni:
[test rpm]

evgeni:
this doesn't have a buildrequires on rake-compiler?

evgeni:
have a look at https://src.fedoraproject.org/rpms/rubygem-rkerberos/blob/master/f/rubygem-rkerberos.spec, it's moving rake-compiler from requires to build-requires.

and we probably should have a release >= the one in fedora?

evgeni:
(rake-compiler is funny, because you don't need to add any code for it to work, so for some reason the gem still builds)

ehelms:
I think I have addressed the concerns

ehelms:
@ekohl or @evgeni  can you please take a look again

packages/foreman/rubygem-rkerberos/rubygem-rkerberos.spec
```diff
@@ -0,0 +1,98 @@
...
+Group: Development/Languages
+License: Artistic 2.0
+URL: https://github.com/domcleal/rkerberos
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+# start specfile generated dependencies
+Requires: %{?scl_prefix_ruby}ruby(release)
+Requires: %{?scl_prefix_ruby}ruby
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
+Requires: %{?scl_prefix}rubygem(rake-compiler)
```
evgeni: I know, this is autogenerated, but I still think we don't need rake-compiler at runtime.

@ekohl can we blacklist it somehow?

packages/foreman/rubygem-rkerberos/rubygem-rkerberos.spec
```diff
@@ -0,0 +1,98 @@
+# template: scl
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
+%global gem_name rkerberos
+
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.1.5
+Release: 2%{?dist}
```
evgeni: Is the 2 intentional? EPEL7 has 0.1.3-6, so 0.1.5-1 would be newer anyways.
And if we want to be newer than Fedora, it needs to be >15 (see https://src.fedoraproject.org/rpms/rubygem-rkerberos/blob/master/f/rubygem-rkerberos.spec)

packages/foreman/rubygem-rkerberos/rubygem-rkerberos.spec
```diff
@@ -0,0 +1,98 @@
...
+Summary: A Ruby interface for the the Kerberos library
+Group: Development/Languages
+License: Artistic 2.0
+URL: https://github.com/domcleal/rkerberos
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
```
evgeni: Fedora carries a [patch](https://src.fedoraproject.org/rpms/rubygem-rkerberos/blob/master/f/rkerberos-0.1.5-gcc8-argnum-fix.patch) for compilation with gcc8 and ruby 2.5 -- might be needed when we go for CentOS8. Not sure if we should just pull it in now.

packages/foreman/rubygem-rkerberos/rubygem-rkerberos.spec
```diff
@@ -0,0 +1,98 @@
...
+
+# %%gem_install compiles any C extensions and installs the gem into ./%%gem_dir
+# by default, so that we can move it into the buildroot in %%install
+%{?scl:scl enable %{scl} - << \EOF}
+%gem_install
+%{?scl:EOF}
+
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -pa .%{gem_dir}/* \
```
evgeni:
```suggestion
cp -p .%{gem_dir}/* \
```

packages/foreman/rubygem-rkerberos/rubygem-rkerberos.spec
ekohl: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/update-requirements does have an NPM blacklist because that's also incomplete. We could implement it there.

rel-eng/tito.props
```diff
@@ -310,6 +310,7 @@ whitelist = foreman
   rubygem-responders
   rubygem-rest-client
   rubygem-retriable
+  rubygem-rkerberos
```
evgeni: this *only* adds it to the SCL, correct?

package_manifest.yaml
```diff
@@ -408,6 +408,7 @@ foreman_nonscl_packages:
     rubygem-kafo_parsers: {}
     rubygem-kafo_wizards: {}
     rubygem-rb-inotify: {}
+    rubygem-rkerberos: {}
```
evgeni: but here we add it to nonscl?

ehelms:
Updated -- this is to build only for SCL

evgeni:
When installing the resulting gem, I get the following:

```
[root@c78a8ef4f9d8 /]# tfm-ruby /usr/bin/irb 
Ignoring rkerberos-0.1.5 because its extensions are not built. Try: gem pristine rkerberos --version 0.1.5
```

evgeni:
```
[root@c78a8ef4f9d8 /]# touch /opt/theforeman/tfm/root/usr/lib64/gems/ruby/rkerberos-0.1.5/gem.build_complete
[root@c78a8ef4f9d8 /]# tfm-ruby /usr/bin/irb 
irb(main):001:0> require 'rkerberos'
Traceback (most recent call last):
       11: from /usr/bin/irb:12:in `<main>'
       10: from (irb):1
        9: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:39:in `require'
        8: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:128:in `rescue in require'
        7: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems.rb:217:in `try_activate'
        6: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems.rb:224:in `rescue in try_activate'
        5: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/specification.rb:1440:in `activate'
        4: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/specification.rb:1458:in `activate_dependencies'
        3: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/specification.rb:1458:in `each'
        2: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/specification.rb:1469:in `block in activate_dependencies'
        1: from /opt/rh/rh-ruby25/root/usr/share/rubygems/rubygems/dependency.rb:310:in `to_specs'
Gem::MissingSpecError (Gem::MissingSpecError)

[root@c78a8ef4f9d8 /]# sed -i '/rake-compiler/ s/runtime/development/' /opt/theforeman/tfm/root/usr/share/gems/specifications/rkerberos-0.1.5.gemspec
[root@c78a8ef4f9d8 /]# tfm-ruby /usr/bin/irb 
irb(main):001:0> require 'rkerberos'
=> true
```

So we need to ship the `gem.build_complete` file, and patch the `rake-compiler` out of the gemspec (something like `sed -i '/rake-compiler/ s/runtime/development/' %{buildroot}/%{gem_spec}` should do).

packages/foreman/rubygem-rkerberos/rubygem-rkerberos.spec
ehelms: Let's cross that bridge given I am not sure what version of things we will be on when we arrive there.

ehelms:
Updated

package_manifest.yaml
```diff
@@ -453,6 +453,7 @@ foreman_proxy_packages:
     rubygem-rake-compiler: {}
     rubygem-rubyipmi: {}
     rubygem-rb-inotify: {}
+    rubygem-rkerberos: {}
```
evgeni: you add the package to `foreman_proxy_packages`, which obal thinks is built for non-SCL, but you're actually building an SCL'ed version, so the result fails repoclosure.

evgeni:
(other than the repoclosure issue, this looks good!)

ehelms:
Fixed the placement of the package

PR #4109, opened 2019-09-14
Update smart-proxy template for SCL

ehelms:

### DIFF
--- a/gem2rpm/smart_proxy_plugin.spec.erb
+++ b/gem2rpm/smart_proxy_plugin.spec.erb
@@ -11,7 +11,7 @@
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_sysconfdir}/foreman-proxy/settings.d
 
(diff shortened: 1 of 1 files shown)


gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -26,27 +26,27 @@ Source0: <%= download_path %>%{gem_name}-%{version}.gem
...
+Requires: %{?scl_prefix}ruby <%= req %>
 <% end -%>
 <% for req in spec.required_rubygems_version -%>
-Requires: ruby(rubygems) <%= req %>
+Requires: %{?scl_prefix}ruby(rubygems) <%= req %>
 <% end -%>
 <% for d in spec.runtime_dependencies -%>
 <% for req in d.requirement -%>
-Requires: rubygem(<%= d.name %>) <%= req  %>
+Requires: %{?scl_prefix_ruby}rubygem(<%= d.name %>) <%= req  %>
```
ekohl: It could come from `_ror` as well. Could you cope the lookup that we have in the other SCL templates?

PR #4116, opened 2019-09-18
Update smart_proxy_plugin gem2rpm template

ehelms:

### DIFF
--- a/gem2rpm/smart_proxy_plugin.spec.erb
+++ b/gem2rpm/smart_proxy_plugin.spec.erb
@@ -1,17 +1,22 @@
-# FIXME:
-#   1. Edit foreman-proxy requirement and specify minimum version
-#   2. Delete these lines
-#
 # template: smart_proxy_plugin
+<%-
+require 'json'
+SCL_PREFIXES = JSON.load(File.read(File.join(`git rev-parse --show-toplevel`.strip, 'scl_prefixes.json')))
+-%>
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
 %global gem_name <%= spec.name %>
 %global plugin_name <%= spec.name.sub(/\Asmart_proxy_/, '') %>
 
-%global foreman_proxy_min_version FIXME
-%global foreman_proxy_dir %{_datarootdir}/foreman-proxy
+%global foreman_proxy_min_version 1.24
+%{!?_root_datadir:%global _root_datadir %{_datadir}}
+%global foreman_proxy_dir %{_root_datadir}/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
-%global foreman_proxy_settingsd_dir %{_sysconfdir}/foreman-proxy/settings.d
+%global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
+%global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
 
-Name: rubygem-%{gem_name}
+Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: <%= spec.version %>
 Release: 1%{?foremandist}%{?dist}
 Summary: <%= spec.summary.gsub(/\.$/, "") %>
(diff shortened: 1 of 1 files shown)


gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -1,17 +1,18 @@
-# FIXME:
-#   1. Edit foreman-proxy requirement and specify minimum version
-#   2. Delete these lines
-#
+# Generated from <%= package.spec.file_name %> by gem2rpm -*- rpm-spec -*-
```
ekohl: We actually dropped these in 8d6a5d23c90c69c796e359571545dead05822644 so why are you re-introducing them?

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -1,17 +1,18 @@
...
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
 %global gem_name <%= spec.name %>
 %global plugin_name <%= spec.name.sub(/\Asmart_proxy_/, '') %>
 
-%global foreman_proxy_min_version FIXME
-%global foreman_proxy_dir %{_datarootdir}/foreman-proxy
+%global foreman_proxy_min_version 1.24
+%global foreman_proxy_dir /usr/share/foreman-proxy
```
ekohl: Why this change? I assume that this directory is different within the SCL but I'd prefer a line similar to the one we have in various other places:
```
%{!?_root_datarootdir:%global _root_datarootdir %{_datarootdir}}
```

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -1,17 +1,18 @@
...
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_sysconfdir}/foreman-proxy/settings.d
```
ekohl: Here you also want to prefer `%{_root_sysconfdir}` within the SCL but fall back to the non-SCL.

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -1,17 +1,18 @@
...
+%global smart_proxy_dynflow_bundlerd_dir %{?rhel:/opt/theforeman/tfm/root/}%{_datadir}/smart_proxy_dynflow_core/bundler.d
```
ekohl: I think the `%{_datadir}` value actually includes the SCL path automatically within the SCL env so this may end up being redundant.

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -24,29 +25,29 @@ Source0: <%= download_path %>%{gem_name}-%{version}.gem
 
 # start specfile generated dependencies
 Requires: foreman-proxy >= %{foreman_proxy_min_version}
-Requires: ruby(release)
+Requires: %{?scl_prefix_ruby}ruby(release)
 <% for req in spec.required_ruby_version -%>
-Requires: ruby <%= req %>
+Requires: %{?scl_prefix_ruby}ruby <%= req %>
```
ekohl: Could you use the `SCL_PREFIXES` to pull it from the correct SCL? https://github.com/theforeman/foreman-packaging/blob/682a5f1a0803c5aa422c8495fb19f56494058f36/gem2rpm/foreman_plugin.spec.erb#L64

Note it also needs to read these: https://github.com/theforeman/foreman-packaging/blob/682a5f1a0803c5aa422c8495fb19f56494058f36/gem2rpm/foreman_plugin.spec.erb#L6-L9

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -1,17 +1,18 @@
...
 # template: smart_proxy_plugin
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
 %global gem_name <%= spec.name %>
 %global plugin_name <%= spec.name.sub(/\Asmart_proxy_/, '') %>
 
-%global foreman_proxy_min_version FIXME
-%global foreman_proxy_dir %{_datarootdir}/foreman-proxy
+%global foreman_proxy_min_version 1.24
```
ekohl: Is this chosen because that will change from non-SCL to SCL or as a sort of safe default that we should update with every release?

gem2rpm/smart_proxy_plugin.spec.erb
ehelms: You are correct in your assumption. Can you point me to where this line is used? I get you want to be programmatic about things, but I find this more confusing them simply stating the foreman_proxy_dir is at /usr/share/foreman-proxy. Maybe this changes on something like EL8, but guessing not and is way easier to comprehend.

ekohl: We don't do it for datarootdir but there is for sysconfdir:
https://github.com/theforeman/foreman-packaging/blob/e3312eef30898896c3ff427dd22523bb92e4564e/gem2rpm/hammer_plugin.spec.erb#L13-L14

gem2rpm/smart_proxy_plugin.spec.erb
ehelms: From all my testing, this line produces the right behavior.

gem2rpm/smart_proxy_plugin.spec.erb
ehelms: Yessir.

ehelms:
I pushed a round of updates.

gem2rpm/smart_proxy_plugin.spec.erb
ekohl: So I had a look:

```
# yum -y install https://yum.theforeman.org/releases/nightly/el7/x86_64/foreman-release.rpm && yum -y install foreman-release-scl && yum -y install tfm-build
# rpm --eval '%{?scl:%scl_package rubygem-smart_proxy_plugins}%{_datadir}'









BuildRequires: scl-utils-build
%if "tfm" == "rubygem-smart_proxy_plugins"
Requires: tfm-runtime
Provides: scl-package(tfm)
%endif

/opt/theforeman/tfm/root/usr/share
```

So I expect that this will end up as `/opt/theforeman/tfm/root//opt/theforeman/tfm/root/usr/share`

ehelms:
Updated

gem2rpm/smart_proxy_plugin.spec.erb
ekohl: Could you have a look at this? I can also easily be convinced to only define `foreman_proxy_bundlerd_dir` since only 2 spec files actually use `foreman_proxy_dir`.

ehelms:
Added the root_datadir change requested

ehelms:
@ekohl Can you take a look at this one again

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -1,17 +1,22 @@
...
+%{!?_root_datarootdir:%global _root_datarootdir %{_datarootdir}}
```
ekohl:
```suggestion
%{!?_root_datadir:%global _root_datadir %{_datadir}}
```

ehelms:
@ekohl Fixed up the datadir comment

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -1,17 +1,22 @@
...
+%global foreman_proxy_dir %{_root_datadir}/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
-%global foreman_proxy_settingsd_dir %{_sysconfdir}/foreman-proxy/settings.d
+%global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
+%global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
```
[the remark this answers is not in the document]
ehelms: TO close this out, I have noted this elsewhere to track.

PR #4117, opened 2019-09-18
Introduce two new package groups in package_manifest for foreman-proxy

ehelms:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -11,7 +11,8 @@ foreman_packages:
     foreman_scl_packages: {}
     foreman_nonscl_packages: {}
     foreman_scl_and_nonscl_packages: {}
-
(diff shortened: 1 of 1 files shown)


ehelms:
My goal here is to allow easier rebuilds of top level projects. In this case the ability to entirely rebuild the foreman-proxy stack or the foreman-proxy plugin stack.

ehelms:
@pcreech @ekohl @evgeni thoughts on this idea?

ehelms:
In that instance, I would introduce a "common" or some other group that gets included in each group that needs it to denote those packages.

pcreech:
@ehelms Generally, I'm in favor of breaking down large groups into smaller subgroups, to help improve interacting with packaging.  +1000 from me

PR #4118, opened 2019-09-18
Bump rubygem-chef-api to 0.9.0-1

ehelms:


packages/plugins/rubygem-chef-api/rubygem-chef-api.spec
```diff
@@ -1,41 +1,56 @@
...
+Summary: A Chef API client in Ruby
 Group: Development/Languages
-License: ASL 2.0
+License: Apache-2.0
 URL: https://github.com/sethvargo/chef-api
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
+# start specfile generated dependencies
 Requires: ruby(release)
+Requires: ruby >= 2.2
```
ekohl: We don't have ruby 2.2 in the non-SCL context

PR #4119, opened 2019-09-18
Bump rubygem-openscap to 0.4.9-1

ehelms:


packages/plugins/rubygem-openscap/rubygem-openscap.spec
```diff
@@ -1,66 +1,40 @@
...
 URL: https://github.com/OpenSCAP/ruby-openscap
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
+# start specfile generated dependencies
+Requires: ruby(release)
+Requires: ruby
 Requires: ruby(rubygems)
 Requires: rubygem(ffi) >= 1.0.9
-# require libopenscap.so.8 in an arch neutral way
-Requires: openscap >= 1.2.9
```
ekohl: I suspect you need to add this below the generated dependencies to make sure it's present.

packages/plugins/rubygem-openscap/rubygem-openscap.spec
```diff
@@ -1,103 +1,92 @@
...
 
-Name: rubygem-%{gem_name}
-Version: 0.4.7
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.4.9
 Release: 1%{?dist}
 Summary: A FFI wrapper around the OpenSCAP library
 Group: Development/Languages
-License: GPLv2+
+License: GPL-2.0
```
ekohl: I think this should be GPLv2 or GPLv2+ so it matches the Fedora license codes.

ehelms:
Updated

evgeni:
@ekohl we never shipped the `-devel` package in our comps, so while you're *technically* correct -- nobody should *need* the obsoletes.

That said, we also don't have `-doc` in comps :)

PR #4120, opened 2019-09-18
Update foreman-proxy to build for the SCL

ehelms:

### DIFF
--- a/packages/foreman/foreman-proxy/foreman-proxy.spec
+++ b/packages/foreman/foreman-proxy/foreman-proxy.spec
@@ -1,10 +1,12 @@
 %global homedir %{_datadir}/%{name}
 %global confdir config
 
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -1,10 +1,11 @@
 %global homedir %{_datadir}/%{name}
 %global confdir config
 
+%{?scl:%global scl_prefix %{scl}-}
```
ekohl: In foreman.spec there's a helpful comment why this is here:
```
# explicitly define, as we build on top of an scl, not inside with scl_package
```

I suspect here there's the same reason

ehelms: Fair -- added

PR #4123, opened 2019-09-18
Bump rubygem-rubyipmi to 0.10.0-1

ehelms:

### DIFF
--- a/packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
+++ b/packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
@@ -1,55 +1,66 @@
+# template: scl
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
 %global gem_name rubyipmi
 
-Summary: A ruby wrapper for ipmi command line tools
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.10.0
-Release: 3%{?dist}
+Release: 4%{?dist}
+Summary: A ruby wrapper for ipmi command line tools that supports ipmitool and freeipmi
 Group: Development/Languages
 License: LGPLv2
 URL: https://github.com/logicminds/rubyipmi
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
-%if 0%{?fedora} > 18 || 0%{?rhel} >= 7
-Requires: %{?scl_prefix_ruby}ruby(release)
-%else
-Requires: %{?scl_prefix_ruby}ruby(abi)
-%endif
-Requires: %{?scl_prefix_ruby}ruby(rubygems)
+
 Requires: ipmitool
-%if 0%{?fedora} > 18 || 0%{?rhel} >= 7
+
+# start specfile generated dependencies
+Requires: %{?scl_prefix_ruby}ruby(release)
+Requires: %{?scl_prefix_ruby}ruby
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
 BuildRequires: %{?scl_prefix_ruby}ruby(release)
-%else
-BuildRequires: %{?scl_prefix_ruby}ruby(abi)
-%endif
+BuildRequires: %{?scl_prefix_ruby}ruby
 BuildRequires: %{?scl_prefix_ruby}rubygems-devel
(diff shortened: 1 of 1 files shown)


packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
```diff
@@ -1,80 +1,91 @@
...
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.10.0
-Release: 3%{?dist}
+Release: 4%{?dist}
+Summary: A ruby wrapper for ipmi command line tools that supports ipmitool and freeipmi
 Group: Development/Languages
-License: LGPLv2
-URL: https://github.com/logicminds/rubyipmi
+License: LGPLv2.1
+URL: http://github.com/logicminds/rubyipmi
```
evgeni: this is pulled from the gem metadata, but I think we should "correct" it here. (I've opened a PR upstream to fix that: https://github.com/logicminds/rubyipmi/pull/49)

packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
```diff
@@ -1,80 +1,91 @@
...
-%if 0%{?fedora} > 18 || 0%{?rhel} >= 7
+
+# start specfile generated dependencies
 Requires: %{?scl_prefix_ruby}ruby(release)
-%else
-Requires: %{?scl_prefix_ruby}ruby(abi)
-%endif
+Requires: %{?scl_prefix_ruby}ruby
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
-Requires: ipmitool
```
evgeni: as far as I understand the gem, it's a wrapper around `ipmitool` and `freeipmi`, so I think the requires should remain (outside of the generated block)

packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
```diff
@@ -1,80 +1,91 @@
...
-Summary: A ruby wrapper for ipmi command line tools
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.10.0
-Release: 3%{?dist}
+Release: 4%{?dist}
+Summary: A ruby wrapper for ipmi command line tools that supports ipmitool and freeipmi
 Group: Development/Languages
-License: LGPLv2
-URL: https://github.com/logicminds/rubyipmi
+License: LGPLv2.1
```
evgeni: this was changed from 2.1 to 2 in https://github.com/theforeman/foreman-packaging/commit/7833c2d3f31ce9f1e00ad5cb67d7a5a05f641963#diff-a8eaae8f93e67347c371d8bdb1148e66, so your change would be a regression :)

packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
```diff
@@ -1,80 +1,93 @@
...
+# %%gem_install compiles any C extensions and installs the gem into ./%%gem_dir
+# by default, so that we can move it into the buildroot in %%install
+%{?scl:scl enable %{scl} - << \EOF}
+%gem_install
+%{?scl:EOF}
 
 %install
 mkdir -p %{buildroot}%{gem_dir}
-cp -a .%{gem_dir}/* \
+cp -pa .%{gem_dir}/* \
```
ekohl: f02a98ade2c230f3968cd12e0ac220262f10109c simplified this:

```suggestion
cp -a .%{gem_dir}/* \
```

packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
```diff
@@ -1,80 +1,93 @@
...
-%{gem_instdir}/VERSION
-
-%exclude %{gem_instdir}/%{gem_name}.gemspec
-%exclude %{gem_instdir}/Gemfile*
+%{gem_instdir}/rubyipmi.gemspec
+%{gem_instdir}/spec
 
 %changelog
+* Tue Sep 17 2019 Eric D. Helms <ericdhelms@gmail.com> 0.10.0-4
+- Update to 0.10.0-4
```
ekohl:
```suggestion
- Regenerate template based on the current scl gem2rpm template
```

PR #4124, opened 2019-09-18
Bump rubygem-infoblox to 3.0.0-1

ehelms:

### DIFF
--- a/packages/plugins/rubygem-infoblox/rubygem-infoblox.spec
+++ b/packages/plugins/rubygem-infoblox/rubygem-infoblox.spec
@@ -1,60 +1,66 @@
-# Generated from infoblox-1.0.0.gem by gem2rpm -*- rpm-spec -*-
+# template: scl
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
 %global gem_name infoblox
 
-Name: rubygem-%{gem_name}
+Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 3.0.0
-Release: 1%{?dist}
-Summary: A Ruby wrapper to the Infoblox WAPI
+Release: 2%{?dist}
 Group: Development/Languages
+Summary: A Ruby wrapper to the Infoblox WAPI.
 License: MIT
 URL: https://github.com/govdelivery/infoblox
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
-%if 0%{?rhel} == 6
-Requires: ruby(abi)
-%else
-Requires: ruby(release)
-%endif
-Requires: ruby
-Requires: ruby(rubygems)
-Requires: rubygem(faraday)
-Requires: rubygem(faraday_middleware)
-%if 0%{?rhel} == 6
-BuildRequires: ruby(abi)
-%else
-BuildRequires: ruby(release)
-%endif
-BuildRequires: ruby
-BuildRequires: rubygems-devel
+
+# start specfile generated dependencies
+Requires: %{?scl_prefix_ruby}ruby(release)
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-infoblox/rubygem-infoblox.spec
```diff
@@ -1,76 +1,85 @@
...
+
 %global gem_name infoblox
 
-Name: rubygem-%{gem_name}
+Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 3.0.0
-Release: 1%{?dist}
-Summary: A Ruby wrapper to the Infoblox WAPI
+Release: 2%{?dist}
+Summary: This gem is a Ruby interface to the Infoblox WAPI. Using the gem, you can query, create, update, and delete DNS records in your Infoblox instance
```
evgeni: I think summary and description should be swapped?

packages/plugins/rubygem-infoblox/rubygem-infoblox.spec
```diff
@@ -79,6 +88,9 @@ cp -a .%{gem_dir}/* \
 %{gem_instdir}/spec
 
 %changelog
+* Tue Sep 17 2019 Eric D. Helms <ericdhelms@gmail.com> 3.0.0-2
+- Update to 3.0.0-2 for SCL
```
evgeni: the PR title is different, but that's actually the correct summary :)

packages/plugins/rubygem-infoblox/rubygem-infoblox.spec
```diff
@@ -1,76 +1,85 @@
...
-cp -a .%{gem_dir}/* \
+cp -pa .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
 %files
 %dir %{gem_instdir}
-%exclude %{gem_instdir}/.*
-%doc %{gem_instdir}/LICENSE.txt
+%exclude %{gem_instdir}/.gitignore
+%{gem_instdir}/.ruby-version
```
ekohl: This can also be excluded

packages/plugins/rubygem-infoblox/rubygem-infoblox.spec
ehelms: I agree with you, but this is generated by the gem2rpm template and any future updates will keep putting this back the way it is.

ekohl: We don't regen the template that often. When bumping dependencies it only replaces the block between `# start specfile generated dependencies` and `# end specfile generated dependencies`. I've submitted https://github.com/Granicus/infoblox/pull/62 so it should be clean in future releases though.

packages/plugins/rubygem-infoblox/rubygem-infoblox.spec
```diff
@@ -1,76 +1,85 @@
...
 # %%gem_install compiles any C extensions and installs the gem into ./%%gem_dir
 # by default, so that we can move it into the buildroot in %%install
+%{?scl:scl enable %{scl} - << \EOF}
 %gem_install
+%{?scl:EOF}
 
 %install
 mkdir -p %{buildroot}%{gem_dir}
-cp -a .%{gem_dir}/* \
+cp -pa .%{gem_dir}/* \
```
evgeni: we dropped the `-p` in the templates in https://github.com/theforeman/foreman-packaging/commit/f02a98ade2c230f3968cd12e0ac220262f10109c

ehelms:
@ekohl I think you meant that comment for https://github.com/theforeman/foreman-packaging/pull/4119

ekohl:
Yes I did. Too many tabs :(

PR #4128, opened 2019-09-18
add ansible-collection-theforeman-foreman

evgeni:

### DIFF
--- a/packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
+++ b/packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
@@ -1,0 +1,58 @@
+%global debug_package %{nil}
+
+%global collection_namespace theforeman
+%global collection_name foreman
+%global collection_directory %{_datadir}/ansible/collections/ansible_collections/%{collection_namespace}/%{collection_name}
+
+%if 0%{?fedora} || 0%{?rhel} >= 8
+%global ansible_python python3
+%global pyyaml %{ansible_python}-pyyaml
+%else
+%global ansible_python python2
+%global pyyaml PyYAML
+%endif
+
+%global release 1
+
+Name:       ansible-collection-%{collection_namespace}-%{collection_name}
+Version:    0.1.0
+Release:    %{?prerelease:0.}%{release}%{?prerelease}%{?nightly}%{?dist}
+Summary:    The Foreman Project Ansible modules collection
+
+License:    GPLv3+
+URL:        https://theforeman.org/plugins/foreman-ansible-modules
+Source0:    https://galaxy.ansible.com/download/%{collection_namespace}-%{collection_name}-%{version}.tar.gz
+BuildArch:  noarch
+
+Requires: ansible >= 2.8
+Requires: %{ansible_python}-apypie >= 0.0.5
+Requires: %{pyyaml}
+%if 0%{?rhel} == 7
+Requires: python-ipaddress
+%endif
+
+%description
+Collection of Ansible Modules to manage Foreman installations.
+Includes modules for Katello.
+
+%prep
+%setup -q -c
+
+%build
+
+%install
+mkdir -p %{buildroot}%{collection_directory}
+cp -a ./* %{buildroot}%{collection_directory}
+
+
+
+%files
+%{collection_directory}/
+%license %{collection_directory}/LICENSE
+%doc %{collection_directory}/README.md
+
+
+
+%changelog
+* Wed Sep 18 2019 Evgeni Golov - 0.1.0-1
+- Initial package
--- a/comps/comps-foreman-client-fedora29.xml
+++ b/comps/comps-foreman-client-fedora29.xml
@@ -10,6 +10,7 @@
     <description>Foreman Client Packages</description>
     <uservisible>true</uservisible>
     <packagelist>
+      <packagereq type="default">ansible-collection-theforeman-foreman</packagereq>
       <packagereq type="default">foreman-client-release</packagereq>
       <packagereq type="default">katello-agent</packagereq>
       <packagereq type="default">katello-host-tools</packagereq>
--- a/comps/comps-foreman-client-rhel7.xml
+++ b/comps/comps-foreman-client-rhel7.xml
@@ -10,6 +10,7 @@
     <description>Foreman Client Packages</description>
     <uservisible>true</uservisible>
     <packagelist>
+      <packagereq type="default">ansible-collection-theforeman-foreman</packagereq>
       <packagereq type="default">foreman-client-release</packagereq>
       <packagereq type="default">gofer</packagereq>
       <packagereq type="default">katello-agent</packagereq>
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -504,6 +504,7 @@ foreman_client_packages:
         - f29-updates
         - f29-puppet-6
   hosts:
+    ansible-collection-theforeman-foreman: {}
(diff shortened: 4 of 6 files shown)


packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
```diff
@@ -0,0 +1,56 @@
...
+
+Requires: ansible >= 2.8
+Requires: %{ansible_python}-apypie >= 0.0.5
+Requires: %{pyyaml}
+%if 0%{?rhel} == 7
+Requires: python-ipaddress
+%endif
+
+%description
+Collection
```
ekohl: The best collection!

packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
```diff
@@ -0,0 +1,56 @@
+%global debug_package %{nil}
+
+%global collection_namespace theforeman
+%global collection_name foreman
+%global collection_folder %{_datadir}/ansible/collections/ansible_collections/%{collection_namespace}/%{collection_name}
```
ekohl: Folder is a bit of a Windows naming convention.
```suggestion
%global collection_directory %{_datadir}/ansible/collections/ansible_collections/%{collection_namespace}/%{collection_name}
```

evgeni:
repoclosure should be fixed by https://github.com/theforeman/foreman-packaging/pull/4130

ehelms:
Have we discussed this vector vs. Galaxy and the implications to users? I am looking at this from the user perspective and how we communicate using one vs. the other.

evgeni:
"this" is shipping the collection as an rpm? It was blessed by the Ansible team. IMHO the only difference for the user is the install/upgrade part, use is identical. And that matches how rpm compares to pip/gem/npm install too.

ehelms:
I was targeting the question less at can or should we deliver as an RPM, and more around the messaging. I don't recall where we landed on user messaging around supported versions. As in, what's on Galaxy will be use across Foreman versions at your own risk? Will Galaxy represent bleeding edge and the RPM targets a known, working state for a given release? While the RPM will be the only "guaranteed" to work against whichever version of the collection is available for that version of Foreman. I am saying I want us to know this up front across the maintainers so we understand our support burden and communicate it to users.

evgeni:
The repo currently says "works with any supported foreman version" and that's also what I'd expect at Galaxy.

For the rpm... Ideally I'd put it in a non-versioned repo (which we currently don't have) and apply the same rule to it. Until then: rpm us for the version it shipped with. Unless someone disagrees, of course.

evgeni:
[test rpm]

evgeni:
fedora 28 doesn't have ansible 2.8 (epel7 and f29 do). shall we just not build this for f28 then?

evgeni:
@ekohl I've pushed a small change (https://github.com/theforeman/foreman-packaging/compare/b232b43b2f01f95bc2048e857e9adb1b3ac15b21..5d79951ac083c19a96b5c80c5d1d318750c39a25), could you please have a second look?

packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
```diff
@@ -0,0 +1,57 @@
...
+Includes modules for Katello.
+
+%prep
+%setup -q -c
+
+%build
+
+%install
+mkdir -p %{buildroot}%{collection_directory}
+cp -a ./* %{buildroot}%{collection_directory}
```
ekohl: I'm still more a fan of the explicit paths. Looking at what it now copies is also `tests/sanity/ignore-2.10.txt` which doesn't appear all that useful. Are you doing this so `FILES.json` is correct?

packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
```diff
@@ -0,0 +1,57 @@
...
+
+
+
+%files
+%{collection_directory}/
+%license %{collection_directory}/LICENSE
```
ekohl: This is now in /usr/share/ansible rather than /usr/share/licenses

packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
```diff
@@ -0,0 +1,57 @@
...
+
+%build
+
+%install
+mkdir -p %{buildroot}%{collection_directory}
+cp -a ./* %{buildroot}%{collection_directory}
+
+
+
+%files
```
ekohl: Perhaps you should also add `%doc README.md`

packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
evgeni: I am doing this to match what `ansible-galaxy collection install` would do, and it puts *everything* into that folder.

packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
evgeni: is that a problem?

packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
evgeni: good idea

packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
ekohl: I'm not 100% sure. https://rpm-packaging-guide.github.io/#built-in-macros states:
> This identifies the file listed as a LICENSE file and it will be installed and labeled as such by RPM. Example: %license LICENSE

Looking at this, the labeling might still happen and be good enough. The tag was introduced because previously `%doc` was used. Licenses must be delivered and conflicts with `%nodoc`. Since it's shipped within the collection and there's no (standard) way to exclude this, I'm going to guess it's not a problem.

evgeni: I *think* that should be fine.

packages/client/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
```diff
@@ -0,0 +1,58 @@
...
+
+%global collection_namespace theforeman
+%global collection_name foreman
+%global collection_directory %{_datadir}/ansible/collections/ansible_collections/%{collection_namespace}/%{collection_name}
+
+%if 0%{?fedora} || 0%{?rhel} >= 8
+%global ansible_python python3
+%global pyyaml %{ansible_python}-pyyaml
+%else
+%global ansible_python python2
```
ehelms: With CentOS 7.7 out, and Python 3 available now, do we want to switch to python3 in all cases?

evgeni: No, unless ansible also switched to using python3.

ehelms: Well tell them to get with the times. It's been a day already!

PR #4133, opened 2019-09-18
Example SCL nodejs package to agree on structure

ehelms:


ehelms:
This is entirely meant as an example for comment on the structure with SCLs for a nodejs package. The goal would be to move the generic parts into npm2rpm to then apply across the board.

packages/foreman/nodejs-history/nodejs-history.spec
```diff
@@ -16,38 +19,49 @@ Source5: https://registry.npmjs.org/value-equal/-/value-equal-0.4.0.tgz
...
 Source7: https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.13.2.tgz
 Source8: https://registry.npmjs.org/js-tokens/-/js-tokens-4.0.0.tgz
-Source9: %{name}-%{version}-registry.npmjs.org.tgz
-BuildRequires: nodejs-packaging
+Source9: nodejs-%{npm_name}-%{version}-registry.npmjs.org.tgz
+
+Requires:       rh-nodejs10-nodejs
+BuildRequires:  rh-nodejs10-nodejs-devel
+BuildRequires:  rh-nodejs10-build
+BuildRequires:  rh-nodejs10-npm
```
ehelms: We can choose to not hard-code these if we want to make the `tfm` dependent on `rh-nodejs10` and define macros therein like we do for Rails.

packages/foreman/nodejs-history/nodejs-history.spec
```diff
@@ -16,38 +19,49 @@ Source5: https://registry.npmjs.org/value-equal/-/value-equal-0.4.0.tgz
...
+Provides: %{?scl_prefix}npm(%{npm_name}) = %{version}
+Provides: bundled(%{?scl_prefix}npm(@babel/runtime)) = 7.4.3
+Provides: bundled(%{?scl_prefix}npm(history)) = 4.9.0
+Provides: bundled(%{?scl_prefix}npm(js-tokens)) = 4.0.0
+Provides: bundled(%{?scl_prefix}npm(loose-envify)) = 1.4.0
+Provides: bundled(%{?scl_prefix}npm(regenerator-runtime)) = 0.13.2
+Provides: bundled(%{?scl_prefix}npm(resolve-pathname)) = 2.2.0
+Provides: bundled(%{?scl_prefix}npm(tiny-invariant)) = 1.0.4
+Provides: bundled(%{?scl_prefix}npm(tiny-warning)) = 1.0.2
+Provides: bundled(%{?scl_prefix}npm(value-equal)) = 0.4.0
```
ehelms: Is this the correct way to specify bundled with an scl_prefix ?

packages/foreman/nodejs-history/nodejs-history.spec
```diff
@@ -16,38 +19,49 @@ Source5: https://registry.npmjs.org/value-equal/-/value-equal-0.4.0.tgz
...
 AutoReq: no
 AutoProv: no
 
-%define npm_cache_dir /tmp/npm_cache_%{name}-%{version}-%{release}
+%define npm_cache_dir npm_cache
```
ehelms: This change was required for me to get this to work with NodeJS 10.

ekohl: That was always a weird one to me

packages/foreman/nodejs-history/nodejs-history.spec
ekohl: I think so

packages/foreman/nodejs-history/nodejs-history.spec
ekohl: If we make tfm dependent on nodejs, would that require nodejs on every deployment rather than the build? I'd still prefer to use macros and keep it in the build env, but do the dependency in mock and koji. Possibly via a tfm-nodejs subpackage which in turn depends on the correct nodejs. I think this will make it easier when we also want to support EL8 modules.

ehelms: We would omit nodejs from the tfm-runtime package and relegate it only to build level packages. I think this would achieve what you want without having to spawn out a separate package but still provide all the same macros in all the environments we typically expect them in.

packages/foreman/nodejs-history/nodejs-history.spec
ehelms: It does feel a bit odd at first, but everything I tried with using the one defined in `/tmp` didn't work. I think it's due to the cache structure change in newer nodejs versions.

packages/foreman/nodejs-history/nodejs-history.spec
ehelms: I'll open a PR to show my thinking.

packages/foreman/nodejs-history/nodejs-history.spec
ekohl: It probably should at least use some macro to get the temp dir if you want an absolute path.

ehelms: Why would you want an absolute path if this works?

ehelms:
Updated structure based on what tfm with nodejs support would look like. Please re-review and if we are satisfied I will work on updating npm2rpm

packages/foreman/nodejs-history/nodejs-history.spec
ekohl: All I'm saying is that I disliked the hardcoded `/tmp` in the spec file and I'm glad that's gone :)

packages/foreman/nodejs-history/nodejs-history.spec
```diff
@@ -16,38 +19,49 @@ Source5: https://registry.npmjs.org/value-equal/-/value-equal-0.4.0.tgz
 Source6: https://registry.npmjs.org/resolve-pathname/-/resolve-pathname-2.2.0.tgz
 Source7: https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.13.2.tgz
 Source8: https://registry.npmjs.org/js-tokens/-/js-tokens-4.0.0.tgz
-Source9: %{name}-%{version}-registry.npmjs.org.tgz
-BuildRequires: nodejs-packaging
```
ekohl: We rely on this at least for `%nodejs_sitelib` in non-SCL builds. In SCL this is provided by `rh-nodejs10-build`.

packages/foreman/nodejs-history/nodejs-history.spec
```diff
@@ -16,38 +19,49 @@ Source5: https://registry.npmjs.org/value-equal/-/value-equal-0.4.0.tgz
...
+Source9: nodejs-%{npm_name}-%{version}-registry.npmjs.org.tgz
+
+Requires:       %{?scl_prefix_nodejs}nodejs
+BuildRequires:  %{?scl_prefix_nodejs}nodejs-devel
```
ekohl: We don't really use the development headers

packages/foreman/nodejs-history/nodejs-history.spec
```diff
@@ -16,38 +19,49 @@ Source5: https://registry.npmjs.org/value-equal/-/value-equal-0.4.0.tgz
 Source6: https://registry.npmjs.org/resolve-pathname/-/resolve-pathname-2.2.0.tgz
 Source7: https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.13.2.tgz
 Source8: https://registry.npmjs.org/js-tokens/-/js-tokens-4.0.0.tgz
-Source9: %{name}-%{version}-registry.npmjs.org.tgz
-BuildRequires: nodejs-packaging
+Source9: nodejs-%{npm_name}-%{version}-registry.npmjs.org.tgz
```
ekohl: This looks incorrect and may also be the cause of its removal.

ehelms: Why does this look incorrect? (it works) In an SCL context, `name` now includes the prefix

ehelms:
Updated based on the comments

packages/foreman/nodejs-history/nodejs-history.spec
ekohl: Probably something we'll find out in the npm2rpm process but we do rewrite some things. For example, `@theforeman/vendor` becomes `nodejs-theforeman-vendor`. Likely easier to fix in the template though.

packages/foreman/nodejs-history/nodejs-history.spec
```diff
@@ -16,39 +19,57 @@ Source5: https://registry.npmjs.org/value-equal/-/value-equal-0.4.0.tgz
 Source6: https://registry.npmjs.org/resolve-pathname/-/resolve-pathname-2.2.0.tgz
 Source7: https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.13.2.tgz
 Source8: https://registry.npmjs.org/js-tokens/-/js-tokens-4.0.0.tgz
-Source9: %{name}-%{version}-registry.npmjs.org.tgz
+Source9: nodejs-%{npm_name}-%{version}-registry.npmjs.org.tgz
+
+Requires:       %{?scl_prefix_nodejs}nodejs
+%if 1%{!?scl:0}
```
ekohl: Doesn't this become `if 1` or `if 10` so always true?

ehelms:
Adjusted for the wrong scl check, also opened https://github.com/dLobatog/npm2rpm/pull/58/files

ehelms:
This now reflects the changes for a single and bundled package

PR #4134, opened 2019-09-18
Add NodeJS SCL to tfm

ehelms:

### DIFF
--- a/packages/foreman/tfm/tfm.spec
+++ b/packages/foreman/tfm/tfm.spec
@@ -10,6 +10,8 @@
 %global scl_prefix_ror %{scl_ror}-
 %global scl_ruby rh-ruby25
 %global scl_prefix_ruby %{scl_ruby}-
+%global scl_nodejs rh-nodejs10
+%global scl_prefix_nodejs %{scl_nodejs}-
 
 # Do not produce empty debuginfo package.
 %global debug_package %{nil}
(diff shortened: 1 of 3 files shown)


ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
[test rpm]

packages/foreman/tfm/tfm.spec
```diff
@@ -19,12 +21,13 @@
...
 Name: %scl_name
 Version: 5.0
-Release: 8%{?dist}
+Release: 9%{?dist}
 License: GPLv2+
 Group: Applications/File
 Source0: README
 Source1: LICENSE
 Source2: tfm.attr
+Source3: nodejs-symlink-deps
```
ehelms: @ekohl I had to add this in order to fix this bug I found -- https://bugzilla.redhat.com/show_bug.cgi?id=1755675

packages/foreman/tfm/tfm.spec
```diff
@@ -137,6 +143,8 @@ Requires: %{scl_runtime}
 Requires: %{scl_runtime}-assets
 Requires: %{scl_prefix_ror}scldevel
 Requires: %{scl_prefix_ruby}scldevel
+Requires: %{scl_prefix_nodejs}scldevel
+Requires: %{scl_prefix_nodejs}build
```
ehelms: There are certain macros that need to be available at build time for nodejs packages, e.g. nodejs_sitelib

packages/foreman/tfm/tfm.spec
ekohl: That increases my suspicion that we're the only people building RPMs on top of a NodeJS SCL

ehelms: Absolutely

PR #4143, opened 2019-09-24
Update scl gem2rpm template to handle extensions

ehelms:

### DIFF
--- a/gem2rpm/scl.spec.erb
+++ b/gem2rpm/scl.spec.erb
@@ -89,7 +89,7 @@ cp -a .%{gem_dir}/* \
 
 <% unless spec.extensions.empty? -%>
 mkdir -p %{buildroot}%{gem_extdir_mri}
-cp -a .%{gem_extdir_mri}/{gem.build_complete,*.so} %{buildroot}%{gem_extdir_mri}/
+cp -a .%{gem_extdir_mri}/{gem.build_complete,%{gem_name}/*.so} %{buildroot}%{gem_extdir_mri}/
 
 <% for ext in spec.extensions -%>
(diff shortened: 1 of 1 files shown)


gem2rpm/scl.spec.erb
```diff
@@ -89,7 +89,7 @@ cp -a .%{gem_dir}/* \
 
 <% unless spec.extensions.empty? -%>
 mkdir -p %{buildroot}%{gem_extdir_mri}
-cp -a .%{gem_extdir_mri}/{gem.build_complete,*.so} %{buildroot}%{gem_extdir_mri}/
+cp -a .%{gem_instdir}/ext/%{gem_name}/*.so %{buildroot}%{gem_extdir_mri}/
```
ekohl: This no longer copies the `gem.build_complete` file which I think is an issue. Is the real issue that `%{gem_extdir_mri}` is actually incorrect? I think I've seen that under in a package that's build both SCL and non-SCL. I can't find the PR now, but rubygem-mysql is an example of that.

ehelms: Turns out, it was more simple than what I had. The `.so` file appears to end up in a directory up from the `gem.build_complete` and this was not copying both. I updated this line to catch the `.so` at the right spot.

ehelms:
An application of this change can be seen as part of https://github.com/theforeman/foreman-packaging/pull/4144/

evgeni:
"funnily", #4144 doesn't work, as it requires the .so file to be in a subfolder, which I think is a special-case of `unicode`?

ehelms:
I have been seeing mixed results with this so gonna close till I pin that down.

PR #4147, opened 2019-09-25
Add root_sysconfdir macro to smart_proxy_plugin template

ehelms:

### DIFF
--- a/gem2rpm/smart_proxy_plugin.spec.erb
+++ b/gem2rpm/smart_proxy_plugin.spec.erb
@@ -6,15 +6,19 @@ SCL_PREFIXES = JSON.load(File.read(File.join(`git rev-parse --show-toplevel`.str
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
+%{!?_root_datadir:%global _root_datadir %{_datadir}}
+%{!?_root_localstatedir:%global _root_localstatedir %{_localstatedir}}
+%{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
+
 %global gem_name <%= spec.name %>
 %global plugin_name <%= spec.name.sub(/\Asmart_proxy_/, '') %>
 
 %global foreman_proxy_min_version 1.24
-%{!?_root_datadir:%global _root_datadir %{_datadir}}
 %global foreman_proxy_dir %{_root_datadir}/foreman-proxy
+%global foreman_proxy_statedir %{_root_localstatedir}/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
-%global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
+%global smart_proxy_dynflow_bundlerd_dir %{!?scl:/opt/theforeman/tfm/root}%{_datadir}/smart_proxy_dynflow_core/bundler.d
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: <%= spec.version %>


ehelms:
Ok I think this is full yupdated from testing nonscl and SCL.

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -6,15 +6,19 @@ SCL_PREFIXES = JSON.load(File.read(File.join(`git rev-parse --show-toplevel`.str
...
 %global plugin_name <%= spec.name.sub(/\Asmart_proxy_/, '') %>
 
 %global foreman_proxy_min_version 1.24
-%{!?_root_datadir:%global _root_datadir %{_datadir}}
 %global foreman_proxy_dir %{_root_datadir}/foreman-proxy
+%global foreman_proxy_statedir %{_root_localstatedir}/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
-%global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
+%global smart_proxy_dynflow_bundlerd_dir %{_root_datadir}/smart_proxy_dynflow_core/bundler.d
```
ekohl: I think this isn't actually in the root but *always* in the SCL when the platform has SCL support (essentially that's EL7 for our purposes). Regardless of whether this package itself is built in SCL.

ehelms: I don't follow exactly. But I can tell you I had to do define it this way to ensure everything worked in both environments.

ekohl:
On Thu, Sep 26, 2019 at 04:45:36AM -0700, Eric Helms wrote:
>ehelms commented on this pull request.
>>  %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
> %global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
>-%global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
>+%global smart_proxy_dynflow_bundlerd_dir %{_root_datadir}/smart_proxy_dynflow_core/bundler.d
>
>I don't follow exactly. But I can tell you I had to do define it this way to ensure everything worked in both environments.

Let's take the current non-SCL rubygem-smart_proxy_ansible. Note it's 
dropping foreman_ansible_core.rb in the SCL directory while itself isn't 
in the SCL context:

```
$ rpm -ql rubygem-smart_proxy_ansible-3.0.1-1.fm1_22.el7.noarch
/etc/foreman-proxy/ansible.cfg
/etc/foreman-proxy/settings.d/ansible.yml
/opt/theforeman/tfm/root/usr/share/smart_proxy_dynflow_core/bundler.d/foreman_ansible_core.rb
/usr/share/foreman-proxy/.ansible
/usr/share/foreman-proxy/.ansible.cfg
/usr/share/foreman-proxy/.ansible_galaxy
/usr/share/foreman-proxy/bundler.d/smart_proxy_ansible.rb
/usr/share/gems/gems/smart_proxy_ansible-3.0.1
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/LICENSE
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/bin
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/bin/json_inventory.sh
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib/smart_proxy_ansible
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib/smart_proxy_ansible.rb
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib/smart_proxy_ansible/api.rb
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib/smart_proxy_ansible/exception.rb
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib/smart_proxy_ansible/http_config.ru
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib/smart_proxy_ansible/plugin.rb
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib/smart_proxy_ansible/roles_reader.rb
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib/smart_proxy_ansible/variables_extractor.rb
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/lib/smart_proxy_ansible/version.rb
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/settings.d
/usr/share/gems/gems/smart_proxy_ansible-3.0.1/settings.d/ansible.yml.example
/usr/share/gems/specifications/smart_proxy_ansible-3.0.1.gemspec
/var/lib/foreman-proxy/ansible
/var/lib/foreman-proxy/ansible_galaxy
```

That's why it's the complicated line:

```
%global smart_proxy_dynflow_bundlerd_dir %%{?rhel:/opt/theforeman/tfm/root/}%{_datadir}/smart_proxy_dynflow_core/bundler.d
```

Sadly there's no real simplification until we fully drop running the 
proxy outside of the SCL on EL7. I made the assumption that we were 
doing that, but maybe it's another case of assumptions that never work 
out.

ehelms:
Thanks for that explanation. Now I get the inconsistency you are alluding to. Ok I will fix it back for here now.

ehelms:
Updated!

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -6,12 +6,16 @@ SCL_PREFIXES = JSON.load(File.read(File.join(`git rev-parse --show-toplevel`.str
...
 %global gem_name <%= spec.name %>
 %global plugin_name <%= spec.name.sub(/\Asmart_proxy_/, '') %>
 
 %global foreman_proxy_min_version 1.24
-%{!?_root_datadir:%global _root_datadir %{_datadir}}
 %global foreman_proxy_dir %{_root_datadir}/foreman-proxy
+%global foreman_proxy_statedir %{_root_localstatedir}/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
 %global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
```
ekohl: Note that this is actually incorrect and I didn't catch it in 19ada47719ee878439561a5e40027c8bcb5cae0f. I think the correct version is:
```suggestion
%global smart_proxy_dynflow_bundlerd_dir %{?rhel:/opt/theforeman/tfm/root}%{_root_datadir}/smart_proxy_dynflow_core/bundler.d
```

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -6,15 +6,19 @@ SCL_PREFIXES = JSON.load(File.read(File.join(`git rev-parse --show-toplevel`.str
...
 %global plugin_name <%= spec.name.sub(/\Asmart_proxy_/, '') %>
 
 %global foreman_proxy_min_version 1.24
-%{!?_root_datadir:%global _root_datadir %{_datadir}}
 %global foreman_proxy_dir %{_root_datadir}/foreman-proxy
+%global foreman_proxy_statedir %{_root_localstatedir}/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
-%global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
+%global smart_proxy_dynflow_bundlerd_dir %{!?scl:/opt/theforeman/tfm/root}%{_datadir}/smart_proxy_dynflow_core/bundler.d
```
ehelms: I updated this in two ways:

 1) Instead of checking is this rhel, check if this is an SCL and include that beginning path
 2) Use datadir so that this can build in nonscl and SCL contexts

This is a lot of work to support non-SCL for the 1 week or less to convert to SCL :/

ekohl: I agree. Perhaps the next time we should make it clear what the goal is. Initially it wasn't clear to me we were pushing for this within 1.24 and if it isn't done in time, it'll be a blocker. We should also have a look at installer changes before we pull the trigger.

My biggest concern is that I'll be out next week and then there's 3 weeks left for me. I still need to finish the tuning option which we've prioritized. That's why I thought keeping backwards compatibility for a rollback was important.

ehelms:
@ekohl Agreed... but my hope would be that we maybe fix this overall design with the _core packages before then. I've started working on this.

PR #4148, opened 2019-09-25
Drop nodejs packages

ehelms:


ehelms:
This drops every nodejs package that is no longer required by Foreman. I verified this by performing a complete mock rebuild of Foreman. 

This also introduces a `foreman_nodejs_packages` group in the package_manifest to make rebuilds easier.

ehelms:
@ekohl you are, sadly, correct. We now have this position where there are nodejs packages in Foreman core repository that are not used by Foreman but used by plugins. In some cases, due to plugins using theforeman JS and in some cases not using it but still needing dependencies

ehelms:
Gonna close for now as this is a mess with plugins.

PR #4149, opened 2019-09-25
Update packages with extensions to build

ehelms:


ekohl:
I'm wondering if we should regenerate these specs completely and also update to the latest compatible versions where possible.

ehelms:
That's something we across the board need to figure out. It's a bit time consuming. Currently, these do not build using mock locally for me which implies they need these updates. I would prefer this focus on them building and look at consistency later.

ehelms:
Note two of the packages: bcrypt, unf_ext do not follow the convention that the rest do with respect to the location of the `.so` file.

ehelms:
[test rpm]

ehelms:
I don't know why I saw these needed changes locally and not anywhere else.

PR #4146, opened 2019-09-25
Update all smart proxy plugins to SCL specs

ehelms:

PR #4153, opened 2019-10-01
Build foreman-proxy for SCL

ehelms:


ehelms:
There will be a small change required to puppet module for plugins -- https://github.com/theforeman/puppet-foreman_proxy/pull/538

The above PR also has a link to a repository with all built changes for any testing.

ehelms:
Just a repoclosure failure due to rkerberos which isn't in the nightly repositories yet but is on koji repos:

http://koji.katello.org/releases/yum/foreman-nightly/RHEL/7/x86_64/tfm-rubygem-rkerberos-0.1.5-16.el7.x86_64.rpm

package_manifest.yaml
```diff
@@ -445,19 +437,20 @@ foreman_proxy_packages:
...
     nightly_releaser: koji-foreman-jenkins
     repoclosure_config: repoclosure/yum_el7.conf
     repoclosure_lookaside_repos:
       rhel7:
-        - el7-foreman-nightly
         - el7-base
         - el7-updates
         - el7-epel
         - el7-extras
-        - el7-puppet-6
```
evgeni: is puppet not needed for the closure?

ehelms:
https://github.com/theforeman/foreman-packaging/pull/4146 is also a prerequisite and needs some love

packages/foreman/rubygem-concurrent-ruby/rubygem-concurrent-ruby.spec
```diff
@@ -29,6 +29,7 @@ BuildArch: noarch
 
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 %{?scl:Obsoletes: ruby193-rubygem-%{gem_name}}
```
evgeni: IMHO we can drop the 193 obsoletes by now, but does not need to be in this PR

evgeni:
why did it still built concurrent ruby and rsec for both, scl and non-scl?

evgeni:
I *think* the repoclosure errors are OK, as they are from *existing* packages which now are not installable anymore.

ehelms:
Still had concurrent-ruby and rsec in the tito.props in two places. That's been updated and this rebased.

evgeni:
rb-inotify and rubyipmi are still built for nonSCL too. is that on purpose?

evgeni:
So I tested this together with https://github.com/theforeman/foreman-packaging/pull/4551 and https://github.com/theforeman/puppet-foreman_proxy/pull/538 on a CentOS7 box. And it worked with a few quirks ;)

* I had to pass `--foreman-proxy-bind-host '*'` to the installer, as this would default to `::` and then Ruby binds to v6 only for some reason.
* I could install REX + proxy plugin just fine
* I could install Ansible + proxy plugin fine after applying https://github.com/theforeman/foreman_ansible/pull/324 (unrelated)
* I didn't try a standalone proxy
* I filed https://projects.theforeman.org/issues/28717 (unrelated, not blocking)

Overall I'd say this is working as intended.

ekohl:
To clarify: on Ruby 2.0 if you bind to `*`, it only binds to IPv4. if you bind to `::`, it actually binds to dual stack. On modern Rubies `*` means dual stack and `::` is single stack. Modern rubies is the behavior you'd expect. Since https://github.com/theforeman/puppet-foreman_proxy/pull/538 needs this to be merged, it can be part of that PR.

package_manifest.yaml
```diff
@@ -384,14 +384,6 @@ foreman_nonscl_packages:
         - "jenkins_job=foreman-installer-develop-source-release"
         - "jenkins_url=https://ci.theforeman.org"
       build_package_tito_builder: "tito.builder.FetchBuilder"
-    foreman-proxy:
```
ekohl: It surprises me that this was here. Did we have two entries for foreman-proxy?

packages/foreman/rubygem-concurrent-ruby/rubygem-concurrent-ruby.spec
```diff
@@ -29,6 +29,7 @@ BuildArch: noarch
...
+%{?scl:Obsoletes: rubygem-%{gem_name}}
```
ekohl: My concern with this is that it's not a direct replacement. If there's some other tool that does need the system Ruby version of this, it doesn't conflict. It does mean we may need a different strategy for removing the old files because yum doesn't handle leaf packages by default.

packages/foreman/rubygem-ffi/rubygem-ffi.spec
```diff
@@ -14,7 +14,7 @@
 
 Name:           %{?scl_prefix}rubygem-%{gem_name}
```
ekohl: It surprises me this is moved from plugins to foreman. Looks like `foreman-proxy` requires `rb-inotify` which requires ffi. It looks like before we pulled in rubygem-ffi version 1.9.10 from EPEL meaning this will be a downgrade.

packages/foreman/rubygem-gssapi/rubygem-gssapi.spec
```diff
@@ -5,7 +5,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 1.2.0
```
ekohl: There's 1.1.2 so this is actually an upgrade, which is good.

packages/foreman/rubygem-concurrent-ruby/rubygem-concurrent-ruby.spec
ehelms: I can not obsolete these and leave them. I don't know that I care that much. Folks shouldn't be using our packages with their custom tools IMO.

package_manifest.yaml
ehelms: We did

PR #4161, opened 2019-10-04
Update a portion of NodeJS packages to SCL specs

ehelms:


ehelms:
Woo I got a chunk of these to actually pass in time!

package_manifest.yaml
```diff
@@ -245,7 +245,6 @@ foreman_nodejs_packages:
     nodejs-babel-register:
       strategy: bundle
     nodejs-babel-runtime: {}
-    nodejs-bootstrap-sass: {}
```
ekohl: This is no longer needed?

ehelms: I can't find any requires on it.

pcreech:
ACK

package_manifest.yaml
ekohl: https://github.com/theforeman/foreman-packaging/blob/2ba172048a55c97d0e3398a5863ae5764b9e908c/packages/foreman/rubygem-patternfly-sass/rubygem-patternfly-sass.spec#L17-L18

pcreech: @ekohl That's "rubygem(bootstrap-sass)" not "npm(bootstrap-sass)"

pcreech: Looks like the requirement on bootstrap-sass was removed after vendorizing took place:

https://github.com/theforeman/foreman-packaging/commit/dfe70d0640ee8dee7fec92b1b60bc28ea63f4095#diff-3399e5c03dc282964481f8f3ced9c141L660-L661

pcreech: Based on the above, i'm inclined to go ahead and merge this to unblock nodejs scl work.   Worst case, we can always add it back in real quick.

PR #4166, opened 2019-10-04
Update foreman_remote_execution_core to 1.3.0

adamruzicka:

PR #4167, opened 2019-10-04
Update foreman_remote_execution to 2.0.0

adamruzicka:

### DIFF
--- a/plugins/ruby-foreman-remote-execution/debian/gem.list
+++ b/plugins/ruby-foreman-remote-execution/debian/gem.list
@@ -1,3 +1,3 @@
 # list of gem urls to download via Jenkins, space separated
-GEMS="https://rubygems.org/downloads/foreman_remote_execution-1.8.4.gem"
+GEMS="https://rubygems.org/downloads/foreman_remote_execution-2.0.0.gem"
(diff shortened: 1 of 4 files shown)


plugins/ruby-foreman-remote-execution/debian/gem.list
```diff
@@ -1,3 +1,3 @@
 # list of gem urls to download via Jenkins, space separated
-GEMS="https://rubygems.org/downloads/foreman_remote_execution-1.8.4.gem"
+GEMS="https://rubygems.org/downloads/foreman_remote_execution-2.0.0.gem"
 GEMS="$GEMS https://rubygems.org/downloads/foreman_remote_execution_core-1.2.0.gem"
```
mmoll: this should be 1.3.0, I guess? :)

adamruzicka: That's being done in here https://github.com/theforeman/foreman-packaging/pull/4166

mmoll: I see... :eyes:

mmoll:
merged, díky @adamruzicka!

PR #4173, opened 2019-10-06
Update plugin nodejs specs to build for SCL

ehelms:


ekohl:
You are changing them to non-bundled. Normally I like this, but it breaks repoclosure.

ehelms:
Not intentional to unbundle them. I put them back and undated them.

PR #4180, opened 2019-10-08
Add license translation

ekohl:
The SPDX license codes differ from Fedora. This adds a mechanism to map them.


licenses.json
```diff
@@ -0,0 +1,4 @@
+{
+	"GPL-3.0": "GPLv3",
+	"GPL-3.0+": "GPLv3+"
```
ehelms: probably a silly nit, these are indented more than the scl_prefixes.json

PR #4181, opened 2019-10-08
Update templates for https URL

ehelms:

### DIFF
--- a/gem2rpm/foreman_plugin.spec.erb
+++ b/gem2rpm/foreman_plugin.spec.erb
@@ -48,7 +48,7 @@ Summary: <%= spec.summary.gsub(/\.$/, "") %>
 Group: Applications/Systems
 License: <%= spec.licenses.empty? ? 'FIXME' : spec.licenses.map { |l| LICENSES.fetch(l, l) }.join(" and ") %>
 <% if spec.homepage -%>
-URL: <%= spec.homepage %>
+URL: <%= spec.homepage.include?('http://github.com') ? spec.homepage.gsub('http://', 'https://') : spec.homepage %>
 <% end -%>
 Source0: <%= download_path %>%{gem_name}-%{version}.gem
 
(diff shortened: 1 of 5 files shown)


ehelms:
@ekohl This dropped my license implementation and just adds a check to set `http://github.com` to `https://github.com`

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -48,7 +48,7 @@ Summary: <%= spec.summary.gsub(/\.$/, "") %>
 Group: Applications/Systems
 License: <%= spec.licenses.empty? ? 'FIXME' : spec.licenses.map { |l| LICENSES.fetch(l, l) }.join(" and ") %>
 <% if spec.homepage -%>
-URL: <%= spec.homepage %>
+URL: <%= spec.homepage.include?('http://github.com') ? spec.homepage.gsub('http://', 'https://') : spec.homepage %>
```
ekohl: I wonder if `spec.homepage` is ever `nil`. Other than that I still like fixing it in upstream gemspecs to improve the ecosystem a bit.
```suggestion
URL: <%= spec.homepage.start_with?('http://github.com') ? spec.homepage.gsub('http://', 'https://') : spec.homepage %>
```

ehelms:
I agree fixing it in upstream would be a nice contribution on top of working around those that are not fixed up yet.

ekohl:
If we silently fix it up, we'll never notice. Given how little we regenerate the specs and it's nice to catch on new specs, I'm slightly leaning to leaving this out.

ehelms:
I forsee us updating specs in the future more than you might think. But we can leave this out. I think it's going to bite folks time and time again. And there is no guarantee that every project will accept a fix and release that fix in a timely manner.

ekohl:
I'm almost wondering if we should have a helper library where we centralize this. We've copied the SCL prefixes, licenses and homepage. If we have a helper, we can also introduce an env var to opt in to fix it up.

PR #4183, opened 2019-10-08
Adding pulp 2.21 GA

parthaa:


ekohl:
Nightly should be also updated [in forklift](https://github.com/theforeman/forklift/blob/bd5b293c265e16a094b109a18e4a7f2c9b1ef293/vagrant/config/versions.yaml#L48)

PR #4182, opened 2019-10-08
Update tfm-rubygem-jwt to 2.2.1

rahulbajaj0509:

PR #4184, opened 2019-10-10
Update Foreman dependency: JWT

rahulbajaj0509:


rahulbajaj0509:
This is a follow-up pr for https://github.com/theforeman/foreman-packaging/pull/4182, i hope it makes sense.

packages/foreman/foreman/foreman.spec
```diff
@@ -135,7 +135,7 @@ Requires: %{?scl_prefix}rubygem(get_process_mem)
 Requires: %{?scl_prefix}rubygem(rack-cors) >= 1.0.2
 Requires: %{?scl_prefix}rubygem(rack-cors) < 1.1.0
 Requires: %{?scl_prefix}rubygem(jwt) >= 2.1.0
-Requires: %{?scl_prefix}rubygem(jwt) < 2.2.0
+Requires: %{?scl_prefix}rubygem(jwt) = 2.2.1
```
ehelms: Needs a bump in the BuildRequires as well

rahulbajaj0509: Done. @ehelms it that correct?

ekohl: An exact match is usually incorrect. In general never pin this way. In general I use https://github.com/ekohl/theforeman-plumbing/blob/master/fm-packaging-update-deps to automatically do this.

rahulbajaj0509: @ekohl would running this script require https://github.com/theforeman/foreman-packaging/pull/4182 to merged first?

ekohl: No, the core PR which has just been merged.

PR #4186, opened 2019-10-10
Update nodejs-node-gyp for SCL

ehelms:


ehelms:
Requires https://github.com/theforeman/foreman-packaging/pull/4134 to be built first

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
```diff
@@ -0,0 +1,46 @@
...
++        log.verbose('get node dir', 'no --target version specified, falling back to RPM installed headers')
+       }
+
+       if (!release.semver) {
+@@ -75,7 +77,7 @@ function configure (gyp, argv, callback) {
+       gyp.commands.install([ release.version ], function (err, version) {
+         if (err) return callback(err)
+         log.verbose('get node dir', 'target node version installed:', release.versionDir)
+-        nodeDir = path.resolve(gyp.devDir, release.versionDir)
++        nodeDir = '/opt/rh/rh-nodejs10/root/usr/share/node'
```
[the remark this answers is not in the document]
ehelms: We will, which you can argue its an ugly solution, and can solve with two patches. I'm OK crossing that bridge when we get there. Or follow up work to try to simplify this.

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
```diff
@@ -0,0 +1,46 @@
+diff --git a/lib/configure.js b/lib/configure.js
+index 2ff476d..7a69e1c 100644
+--- a/lib/configure.js
++++ b/lib/configure.js
+@@ -62,7 +62,9 @@ function configure (gyp, argv, callback) {
+         log.verbose('get node dir', 'compiling against --target node version: %s', release.version)
+       } else {
+         // if no --target was specified then use the current host node version
```
evgeni: should this line be removed/replaced by the next one?

packages/foreman/nodejs-node-gyp/nodejs-node-gyp.spec
```diff
@@ -1,312 +1,313 @@
...
-Provides: bundled(npm(isexe)) = 1.1.2
-Provides: bundled(npm(os-tmpdir)) = 1.0.2
-Provides: bundled(npm(glob)) = 7.1.1
-Provides: bundled(npm(minimist)) = 0.0.8
+
+Provides: %{?scl_prefix}npm(%{npm_name}) = %{version}
+Provides: bundled(npm(abbrev)) = 1.1.1
+Provides: bundled(npm(ajv)) = 6.10.2
 Provides: bundled(npm(ansi)) = 0.3.1
-Provides: bundled(npm(are-we-there-yet)) = 1.1.2
```
evgeni: it confused me that we didn't touch the gyp version, and yet the versions of the bundled modules changed, but in retrospect this is correct, as we pulled new versions that still satisfy the deps.

packages/foreman/nodejs-node-gyp/nodejs-node-gyp.spec
```diff
@@ -1,312 +1,313 @@
...
 %description
 %{summary}
 
 %prep
-mkdir npm_cache
+mkdir -p %{npm_cache_dir}
+%{?scl:scl enable %{?scl_nodejs} - << \end_of_scl}
 for tgz in %{sources}; do
-  echo $tgz | egrep -q "registry.npmjs.org|\.gypi" || npm cache add --cache ./npm_cache $tgz
+  echo $tgz | grep -q -E "registry.npmjs.org|.gypi" || npm cache add --cache %{npm_cache_dir} $tgz
```
evgeni: I think the `\` is still valuable, as it makes the `.` a dot, not "any char"

rel-eng/tito.props
```diff
@@ -73,8 +73,6 @@ whitelist = foreman-bootloaders-redhat
   nodejs-lodash.debounce
   nodejs-lodash-es
   nodejs-loose-envify
-  nodejs-node-gyp
-  nodejs-node-sass
```
evgeni: is the sass removal here intentional?

evgeni:
needs rebaseneeds rebase

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
ehelms: I think the two lines are saying two different things if you look at the log messages that go with them.

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
```diff
@@ -0,0 +1,46 @@
...
+       })
+     }
+@@ -247,9 +249,10 @@ function configure (gyp, argv, callback) {
+     }
+
+     // this logic ported from the old `gyp_addon` python file
+-    var gyp_script = path.resolve(__dirname, '..', 'gyp', 'gyp_main.py')
+-    var addon_gypi = path.resolve(__dirname, '..', 'addon.gypi')
+-    var common_gypi = path.resolve(nodeDir, 'include/node/common.gypi')
++    var gyp_script = '/usr/bin/gyp'
```
ekohl: If we're building to an SCL, will it be in `/usr/bin` or in the `/opt` version of it?

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
evgeni: they are, but how can it have two behaviors when no `--target` was passed? or am I reading the diff wrong?

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
ehelms: This will use `/usr/bin` since gyp is not SCL (I think its a c/python project) -- https://github.com/theforeman/foreman-packaging/pull/4186/files#diff-075b589a7d745162ae3b74d7d464f577R124

ehelms:
[test rpm]

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
pcreech: Reading this, it seems to be logging multiple assumptions that will happen if "--target" is specified.  My vote is to keep as is

PR #4187, opened 2019-10-10
Update nodejs-node-sass for SCL

ehelms:


ehelms:
Requires https://github.com/theforeman/foreman-packaging/pull/4186 to be built into SCL first

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
```diff
@@ -0,0 +1,46 @@
...
++        log.verbose('get node dir', 'no --target version specified, falling back to RPM installed headers')
+       }
+
+       if (!release.semver) {
+@@ -75,7 +77,7 @@ function configure (gyp, argv, callback) {
+       gyp.commands.install([ release.version ], function (err, version) {
+         if (err) return callback(err)
+         log.verbose('get node dir', 'target node version installed:', release.versionDir)
+-        nodeDir = path.resolve(gyp.devDir, release.versionDir)
++        nodeDir = '/opt/rh/rh-nodejs10/root/usr/share/node'
```
ekohl: Is there a way we could determine this automatically in case we change the SCL? Looks like there's a `nodeDir` parameter and can't we pass that in?

packages/foreman/nodejs-node-sass/nodejs-node-sass.spec
```diff
@@ -1,434 +1,451 @@
...
 %{nodejs_sitelib}/%{npm_name}
-%license LICENSE
-%doc CHANGELOG.md
-%doc README.md
+%{_bindir}/node-sass
+%license node_modules/%{npm_name}/LICENSE
+%doc node_modules/%{npm_name}/CHANGELOG.md
+%doc node_modules/%{npm_name}/README.md
 
 %changelog
```
ekohl: No changelog?

comps/comps-foreman-rhel7.xml
```diff
@@ -245,7 +245,7 @@
       <packagereq type="default">nodejs-lodash.debounce</packagereq>
       <packagereq type="default">nodejs-loose-envify</packagereq>
       <packagereq type="default">nodejs-node-gyp</packagereq>
-      <packagereq type="default">nodejs-node-sass</packagereq>
+      <packagereq type="default">tfm-nodejs-node-sass</packagereq>
```
ekohl: This doesn't look correct

package_manifest.yaml
```diff
@@ -210,9 +210,9 @@ foreman_nodejs_packages:
     diff_package_skip: false
     diff_package_tags:
       - foreman-nightly-nonscl-rhel7
+      - foreman-nightly-rhel7
     releasers:
       - koji-foreman
-    nightly_releaser: koji-foreman-jenkins
```
ekohl: Isn't this still relevant?

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
ehelms: This file actually should be in the node-gyp PR. Will fix

ehelms: As to the question, I honestly don't know. And I on the fence how much time its worth spending diving into this code vs. taking this simple patch that works.

ekohl:
Looks like the build still fails
```
[2019-10-19T00:04:51.674Z]     /home/jenkins/workspace/foreman-packaging-rpm-pr-test/packages/foreman/nodejs-node-sass/nodejs-node-sass.spec:447: E: hardcoded-library-path in /usr/lib/debug

[2019-10-19T00:04:51.674Z]     /home/jenkins/workspace/foreman-packaging-rpm-pr-test/packages/foreman/nodejs-node-sass/nodejs-node-sass.spec: W: invalid-url Source179: nodejs-node-sass-4.12.0-registry.npmjs.org.tgz
```

ehelms:
I expect things to fail until the buildroot gets the newer tfm that adds
Nodejs SCL support.

pcreech:
ACK

PR #4192, opened 2019-10-11
Update Foreman RPM to handle SCL nodejs

ehelms:

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -6,10 +6,11 @@
 
 # explicitly define, as we build on top of an scl, not inside with scl_package
 %{?scl:%global scl_prefix %{scl}-}
+%{?scl:%global _scl_root /opt/theforeman/%{scl}/root}
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 9
+%global release 10
 %global prerelease develop
 
 Name:    foreman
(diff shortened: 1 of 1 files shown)


ehelms:
These changes should be safe to merge ahead of time.

packages/foreman/foreman/foreman.spec
```diff
@@ -485,75 +486,75 @@ Requires: %{name} = %{version}-%{release}
 Requires: %{scl}-runtime-assets >= 5
 Requires: %{scl}-runtime-assets < 6
 %endif
```
ekohl: Does this need the sale `%else` as in the header?

packages/foreman/foreman/foreman.spec
```diff
@@ -759,7 +760,7 @@ mv Gemfile Gemfile.in
 cp config/database.yml.example config/database.yml
 cp config/settings.yaml.example config/settings.yaml
 export BUNDLER_EXT_GROUPS="default assets"
-ln -s %{nodejs_sitelib} node_modules
+ln -s %{?scl_prefix_nodejs:/opt/theforeman/tfm/root}%{nodejs_sitelib} node_modules
```
ekohl: Isn't there a macro for this root path?

packages/foreman/foreman/foreman.spec
```diff
@@ -251,79 +251,80 @@ BuildRequires: %{?scl_prefix}rubygem(graphql-batch)
...
 BuildRequires: %{scl}-runtime-assets < 6
+%else
+BuildRequires: nodejs-packaging
 %endif
-BuildRequires: nodejs >= 6.10
+BuildRequires: %{?scl_prefix_nodejs}nodejs >= 6.10
 BuildRequires: http-parser
 # Temporary dep on libuv until https://bugs.centos.org/view.php?id=10606
 # is resolved
 BuildRequires: libuv
```
ekohl: This bug was filed against CentOS 6 for nodejs. We can probably drop it.

packages/foreman/foreman/foreman.spec
ehelms: Not following this question

packages/foreman/foreman/foreman.spec
ehelms: Happy to hear it... things get ewird for packages built against but not into the SCL. They don't have all the same macro definitions.

packages/foreman/foreman/foreman.spec
ekohl: https://github.com/theforeman/foreman-packaging/pull/4192/files#diff-3399e5c03dc282964481f8f3ced9c141R251-R256

packages/foreman/foreman/foreman.spec
ekohl: Isn't this equal to `%{_scl_root}` since this is built into the tfm SCL?

packages/foreman/foreman/foreman.spec
ehelms: Sorry, I still am not clear what you are asking for.

packages/foreman/foreman/foreman.spec
ehelms: Oooo, maybe! I will give it a try

ehelms:
[test rpm]

ehelms:
[test rpm]

packages/foreman/foreman/foreman.spec
evgeni: I think Ewoud asks if we need to add

```
%else
Requires: nodejs-packaging
```
for non-SCL'ed builds

but I wonder when this is the case, because `scl` refers to *our* SCL, and we always build against TFM these days?

packages/foreman/foreman/foreman.spec
```diff
@@ -251,79 +251,80 @@ BuildRequires: %{?scl_prefix}rubygem(graphql-batch)
...
-BuildRequires: npm(style-loader) < 1.0.0
-BuildRequires: npm(uglifyjs-webpack-plugin) >= 1.2.2
-BuildRequires: npm(uglifyjs-webpack-plugin) < 2.0.0
-BuildRequires: npm(url-loader) >= 1.0.1
-BuildRequires: npm(url-loader) < 2.0.0
-BuildRequires: npm(webpack) >= 3.4.1
-BuildRequires: npm(webpack) < 4.0.0
-BuildRequires: npm(webpack-stats-plugin) >= 0.1.5
-BuildRequires: npm(webpack-stats-plugin) < 1.0.0
+BuildRequires: %{?scl_prefix}npm(argv-parse) >= 1.0.1
```
evgeni: should all these be `scl_prefix_nodejs`?

ekohl: I think these are packages from our own repo so I think it's correct. However, we haven't built these yet in our koji.

packages/foreman/foreman/foreman.spec
ekohl: Yes, that's what I meant.

packages/foreman/foreman/foreman.spec
evgeni: seems to work, at least the reproducer in the bug doesn't trigger

packages/foreman/foreman/foreman.spec
evgeni: makes sense.

packages/foreman/foreman/foreman.spec
evgeni: did you?

packages/foreman/foreman/foreman.spec
ehelms: I suppose not if things are working already?

evgeni: probably, yeah. the diff just looked weird to me

packages/foreman/foreman/foreman.spec
```diff
@@ -6,10 +6,11 @@
 
 # explicitly define, as we build on top of an scl, not inside with scl_package
 %{?scl:%global scl_prefix %{scl}-}
+%{?scl:%global _scl_root /opt/theforeman/%{scl}/root}
```
ehelms: I added this to centralize this reference. The trick with any RPM built against but not into the SCL is that most of the SCL macros are not defined and therefore have to be defined manually.

ekohl: This should be part of the RPM macros. I wonder why it's not set properly. Should it be provided by tfm-build?

ehelms: It could be useful to create special macro to represent this path.

pcreech:
[test rpm]

ehelms:
[test rpm]

ehelms:
[test rpm]

ekohl:
[test rpm]

ekohl:
Looks like a node-sass issue:
```
    ERROR in /opt/theforeman/tfm/root/usr/lib/node_modules/css-loader!/opt/theforeman/tfm/root/usr/lib/node_modules/sass-loader/lib/loader.js!./webpack/assets/javascripts/react_app/components/hosts/powerStatus/powerStatusInner/PowerStatusInner.scss
    Module build failed: Error: ENOENT: no such file or directory, scandir '/opt/theforeman/tfm/root/usr/lib/node_modules/node-sass/vendor'
```

ehelms:
[test rpm]

ehelms:
Only failing on repoclosure which I would expect for the nodejs packages.

packages/foreman/foreman/foreman.spec
ekohl: Perhaps we should have a `%{foreman_nodejs_sitelib}` macro.

ehelms: We only use it in two places right now so I am less inclined to go that route. This works and is localized to it's use right now. I think we can explore the macro when this path starts to be more widely needed.

ekohl: I'm ok with that. My main concern is that `_scl_root` should already be defined and this is just redefining it. Possibly with a different value.

packages/foreman/foreman/foreman.spec
```diff
@@ -485,75 +484,75 @@ Requires: %{name} = %{version}-%{release}
 Requires: %{scl}-runtime-assets >= 5
 Requires: %{scl}-runtime-assets < 6
 %endif
-Requires: nodejs >= 6.10
+Requires: %{?scl_prefix_nodejs}nodejs >= 6.10
 # Temporary dep on libuv until https://bugs.centos.org/view.php?id=10606
 # is resolved
 Requires: libuv
```
evgeni: this probably can be dropped too?

ehelms: Probably... I can do that here or follow up with some clean up once we know things are fully working?

evgeni: given we dropped it from `BuildRequires` in this PR, it would be consistent to also drop it from the `Requires` (or not touch at all).

PR #4188, opened 2019-10-10
Add rubygem-sidekiq{,-reliable-fetch}

adamruzicka:

PR #4194, opened 2019-10-15
Fixes #28065 - package dynflow sidekiq service files

ezr-ondrej:
Depend on:
- [x] Package dependencies #4188
- [x] Prepare init script theforeman/foreman#7109

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -11,7 +11,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 10
+%global release 11
 %global prerelease develop
 
 Name:    foreman
@@ -722,19 +722,9 @@ Requires: %{name} = %{version}-%{release}
 Meta Package to install dynflow sidekiq executor support
 
 %files dynflow-sidekiq
-%{_datadir}/%{name}/bundler.d/sidekiq.rb
+%{_datadir}/%{name}/bundler.d/dynflow_sidekiq.rb
 %{_unitdir}/%{dynflow_sidekiq_service_name}.service
 %{_datadir}/%{name}/extras/dynflow-sidekiq.rb
-
-%install dynflow-sidekiq
-install -Dp -m0644 extras/systemd/%{dynflow_sidekiq_service_name}.service %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
-
-sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|/usr/bin/scl enable tfm "sidekiq \1"|' %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
-
-for i in orchestrator.yml worker.yml; do
-  mv %{buildroot}%{_datadir}/%{name}/config/dynflow/$i %{buildroot}%{_sysconfdir}/%{name}/dynflow/
-  ln -sv %{_sysconfdir}/%{name}/$i %{buildroot}%{_datadir}/%{name}/config/$i
-done
 
 %post dynflow-sidekiq
 %systemd_post %{dynflow_sidekiq_service_name}.service
@@ -820,6 +810,7 @@ install -d -m0755 %{buildroot}%{_datadir}/%{name}
 install -d -m0755 %{buildroot}%{_datadir}/%{name}/plugins
 install -d -m0755 %{buildroot}%{_sysconfdir}/%{name}
 install -d -m0755 %{buildroot}%{_sysconfdir}/%{name}/plugins
+install -d -m0755 %{buildroot}%{_sysconfdir}/%{name}/dynflow
 install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}
 install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}/tmp
 install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}/tmp/pids
@@ -829,6 +820,7 @@ install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}/plugins
 #Copy init scripts and sysconfigs
 install -Dp -m0644 %{SOURCE10} %{buildroot}%{_sysconfdir}/sysconfig/%{executor_service_name}
 install -Dp -m0644 %{SOURCE11} %{buildroot}%{_unitdir}/%{executor_service_name}.service
+install -Dp -m0644 extras/systemd/%{dynflow_sidekiq_service_name}.service %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
 install -Dp -m0755 script/%{executor_service_name} %{buildroot}%{_sbindir}/%{executor_service_name}
 install -Dp -m0755 script/%{name}-debug %{buildroot}%{_sbindir}/%{name}-debug
 install -Dp -m0755 script/%{name}-rake %{buildroot}%{_sbindir}/%{name}-rake
@@ -839,6 +831,7 @@ install -Dp -m0644 %{SOURCE4} %{buildroot}%{_sysconfdir}/cron.d/%{name}
 install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
 install -Dp -m0644 extras/systemd/%{name}.service %{buildroot}%{_unitdir}/%{name}.service
 
+sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|/usr/bin/scl enable tfm "sidekiq \1"|' %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
 sed -i '/^ExecStart/ s|/usr/bin/rails \(.\+\)$|/usr/bin/scl enable tfm "rails \1"|' %{buildroot}%{_unitdir}/%{name}.service
 
 cp -p Gemfile.in %{buildroot}%{_datadir}/%{name}/Gemfile.in
@@ -859,6 +852,12 @@ mv %{buildroot}%{_datadir}/%{name}/config/settings.yaml.example %{buildroot}%{_d
 for i in database.yml settings.yaml foreman-debug.conf; do
 mv %{buildroot}%{_datadir}/%{name}/config/$i %{buildroot}%{_sysconfdir}/%{name}
 ln -sv %{_sysconfdir}/%{name}/$i %{buildroot}%{_datadir}/%{name}/config/$i
+done
+
+for i in orchestrator worker; do
+  mv %{buildroot}%{_datadir}/%{name}/config/dynflow/$i.yml.example %{buildroot}%{_datadir}/%{name}/config/dynflow/$i.yml
+  mv %{buildroot}%{_datadir}/%{name}/config/dynflow/$i.yml %{buildroot}%{_sysconfdir}/%{name}/dynflow/
+  ln -sv %{_sysconfdir}/%{name}/dynflow/$i.yml %{buildroot}%{_datadir}/%{name}/config/dynflow/$i.yml
 done
 
 # Put db in %{_localstatedir}/lib/%{name}/db
@@ -1085,6 +1084,9 @@ exit 0
 %systemd_postun_with_restart %{name}.service
 
 %changelog
+* Tue Oct 22 2019 Ondrej Ezr <ezrik12@gmail.com> - 1.24.0-0.11.develop
+- Add dynflow-sidekiq package providing services for running dynflow on sidekiq
+
 * Fri Oct 11 2019 Eric D. Helms <ericdhelms@gmail.com> - 1.24.0-0.10.develop
 - Updates to support NodeJS packages built into SCL
 
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -2,6 +2,7 @@
 %global confdir extras/packaging/rpm/sources
 %global foreman_rake %{_sbindir}/%{name}-rake
 %global executor_service_name dynflowd
+%global dynflow_sidekiq_service_name dynflow-sidekiq@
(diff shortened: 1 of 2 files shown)


packages/foreman/foreman/dynflow@.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
```
ekohl: Can we ship this file in the foreman.git repo? We do this for smart-proxy (https://github.com/theforeman/smart-proxy/blob/develop/extra/systemd) and foreman (https://github.com/theforeman/foreman/tree/develop/extras/systemd) and it has the benefit of being able to share configs between debs and rpms.

packages/foreman/foreman/dynflow@.service
```diff
@@ -0,0 +1,15 @@
...
+Description=Foreman jobs daemon on sidekiq
+Documentation=https://theforeman.org
+After=network.target remote-fs.target nss-lookup.target
+
+[Service]
+Type=forking
+User=foreman
+TimeoutSec=90
+ExecStart=/usr/bin/bundle exec sidekiq -e ${RAILS_ENV} -r ${DYNFLOW_SIDEKIQ_SCRIPT} -C /etc/foreman/dynflow/%i.yml
```
ekohl: There is no bundle on production installs.

packages/foreman/foreman/dynflow@.service
```diff
@@ -0,0 +1,15 @@
...
+WorkingDirectory=/usr/share/foreman
+EnvironmentFile=-/etc/sysconfig/dynflow
```
ekohl: For new services we don't use `/etc/sysconfig` anymore but prefer systemd drop in configs. This saves us from having to manage the difference between RPMs (/etc/sysconfig) and debs (/etc/default).

packages/foreman/foreman/foreman.spec
```diff
@@ -711,6 +714,38 @@ Meta Package to install requirements for Redis caching support
 %files redis
 %{_datadir}/%{name}/bundler.d/redis.rb
 
+%package dynflow-sidekiq
+Summary: Foreman Redis caching support
+Group:  Applications/System
```
ekohl: Could you add a line to allow automatic updating of this?
```suggestion
Group:  Applications/System
# start specfile sidekiq Requires
```

packages/foreman/foreman/foreman.spec
```diff
@@ -711,6 +714,38 @@ Meta Package to install requirements for Redis caching support
...
+Requires: %{?scl_prefix}rubygem(sidekiq) >= 5.0
+Requires: %{?scl_prefix}rubygem(sidekiq) < 6
+Requires: %{?scl_prefix}rubygem(gitlab-sidekiq-fetcher)
```
ekohl:
```suggestion
Requires: %{?scl_prefix}rubygem(gitlab-sidekiq-fetcher)
# end specfile sidekiq Requires
```

packages/foreman/foreman/foreman.spec
```diff
@@ -2,6 +2,7 @@
 %global confdir extras/packaging/rpm/sources
 %global foreman_rake %{_sbindir}/%{name}-rake
 %global executor_service_name dynflowd
+%global dynflow_sidekiq_service_name dynflow-sidekiq
```
ekohl: I think `@` should be part of the service name

packages/foreman/foreman/foreman.spec
```diff
@@ -711,6 +714,38 @@ Meta Package to install requirements for Redis caching support
...
+%config(noreplace) %{_sysconfdir}/sysconfig/%{dynflow_sidekiq_service_name}
+%{_unitdir}/%{dynflow_sidekiq_service_name}@.service
+%{_datadir}/%{name}/script/dynflow-sidekiq.rb
+
+%install dynflow-sidekiq
+install -d -m0755 %{buildroot}%{_sysconfdir}/%{name}/dynflow
+install -Dp -m0644 %{SOURCE12} %{buildroot}%{_sysconfdir}/sysconfig/%{dynflow_sidekiq_service_name}
+install -Dp -m0644 %{SOURCE13} %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}@.service
+
+sed -i '/^ExecStart/ s|/usr/bin/bundle \(.\+\)$|/usr/bin/scl enable tfm "bundle \1"|' %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}@.service
```
ekohl: This is more of a `%build` step.

packages/foreman/foreman/foreman.spec
```diff
@@ -711,6 +714,38 @@ Meta Package to install requirements for Redis caching support
...
+
+mv %{buildroot}%{_datadir}/%{name}/config/dynflow/orchestrator.yml.example %{buildroot}%{_datadir}/%{name}/config/orchestrator.yml
```
ekohl: Can we move it just once instead of twice?

packages/foreman/foreman/foreman.spec
```diff
@@ -711,6 +713,34 @@ Meta Package to install requirements for Redis caching support
...
+Group:  Applications/System
+# start specfile redis Requires
+Requires: %{?scl_prefix}rubygem(sidekiq) >= 5.0
+Requires: %{?scl_prefix}rubygem(sidekiq) < 6
+Requires: %{?scl_prefix}rubygem(gitlab-sidekiq-fetcher)
+# end specfile redis Requires
+Requires: %{name} = %{version}-%{release}
+
+%description dynflow-sidekiq
+Meta Package to install requirements for Redis caching support
```
ehelms: This description mentions Redis caching support but this is for Dynflow support right?

packages/foreman/foreman/foreman.spec
```diff
@@ -711,6 +713,34 @@ Meta Package to install requirements for Redis caching support
 %files redis
 %{_datadir}/%{name}/bundler.d/redis.rb
 
+%package dynflow-sidekiq
```
ehelms: Does this package conflict with the regular dynflow service? Can both run at the same time? What happens if they do? Is the new dynflow-sidekiq optional?

packages/foreman/foreman/foreman.spec
ezr-ondrej: My bad, too much of copy pasting.

packages/foreman/foreman/foreman.spec
ezr-ondrej: No, it doesn't conflict, and it is not trying to remove it, if someone runs both, they will share the load and loadbalance.
`dynflow-sidekiq` is installed in installer only if the option is turned on (theforeman/puppet-foreman#761) so yes, it's optional.

packages/foreman/foreman/foreman.spec
```diff
@@ -711,6 +713,43 @@ Meta Package to install requirements for Redis caching support
...
+
+%files dynflow-sidekiq
+%{_datadir}/%{name}/bundler.d/sidekiq.rb
+%{_unitdir}/%{dynflow_sidekiq_service_name}.service
+%{_datadir}/%{name}/script/dynflow-sidekiq.rb
+
+%install dynflow-sidekiq
+install -Dp -m0644 extras/systemd/%{dynflow_sidekiq_service_name}.service %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
+
+sed -i '/^ExecStart/ s|/usr/local/bin/sidekiq \(.\+\)$|/usr/bin/scl enable tfm "sidekiq \1"|' %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
```
adamruzicka: Does `scl enable tfm "sidekiq -C something"` work? I always used `scl enable tfm -- sidekiq -C something`

ehelms: We do a similar thing for running Foreman as a service using puma.

packages/foreman/foreman/foreman.spec
ehelms: Does the singleton nature of the dynflow sidekiq orchestrator play well with the current dynflow executor if both are deployed simultaneously?

ezr-ondrej: Yeah, that doesn't really matter, the separation is not really affecting the old deployment.

packages/foreman/foreman/foreman.spec
ezr-ondrej: I took it from there actually.

packages/foreman/foreman/foreman.spec
```diff
@@ -27,6 +28,7 @@ Source4: %{name}.cron.d
 Source5: %{name}.tmpfiles
 Source10: %{executor_service_name}.sysconfig
 Source11: %{executor_service_name}.service
+Source12: extras/systemd/%{dynflow_sidekiq_service_name}.service
```
ekohl: This needs to be removed.
```suggestion
```

packages/foreman/foreman/foreman.spec
```diff
@@ -711,6 +713,43 @@ Meta Package to install requirements for Redis caching support
...
+%{_unitdir}/%{dynflow_sidekiq_service_name}.service
+%{_datadir}/%{name}/script/dynflow-sidekiq.rb
+
+%install dynflow-sidekiq
+install -Dp -m0644 extras/systemd/%{dynflow_sidekiq_service_name}.service %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
+
+sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|/usr/bin/scl enable tfm "sidekiq \1"|' %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
+
+for i in orchestrator.yml worker.yml; do
+mv %{buildroot}%{_datadir}/%{name}/config/dynflow/$i %{buildroot}%{_sysconfdir}/%{name}/dynflow/
```
ekohl: Can you indent these lines?

packages/foreman/foreman/foreman.spec
```diff
@@ -711,6 +713,43 @@ Meta Package to install requirements for Redis caching support
...
+
+%description dynflow-sidekiq
+Meta Package to install dynflow sidekiq executor support
+
+%files dynflow-sidekiq
+%{_datadir}/%{name}/bundler.d/sidekiq.rb
+%{_unitdir}/%{dynflow_sidekiq_service_name}.service
+%{_datadir}/%{name}/script/dynflow-sidekiq.rb
+
+%install dynflow-sidekiq
```
ekohl: I don't think you can have separate `%install` sections. At least, we don't use it anywhere else.

ezr-ondrej: So I should install it in the main package directly? or to use build section?

ekohl: Main `%install` is fine. Then within `%files` you split it up again.

ekohl:
Also needs an entry in comps.

ezr-ondrej:
Added comps section, moved install to main install.

ezr-ondrej:
Add changelog line

ekohl:
The build failure is due to the nodejs changes that are going on.

ehelms:
This current design installs and starts "classic" dynflow by default. As I understood things, the target was for the new dynflow-sidekiq to be the default?

If we are keeping "classic" dynflow as the default, installing this new subpackage will configure and start dynflow-sidekiq while leaving classic dynflow running as is. Is this what a user should expect? That feels like it will lead to inconsistencies.

ezr-ondrej:
@ehelms this is now addressed in the installer PR theforeman/puppet-foreman#761 it will disable the old service for you once you select the sidekiq to be installed. If you are not using installer you're on your own to figure it out though.

tbrisker:
this is now blocked on the core PR: `BUILDSTDERR: install: cannot stat 'extras/systemd/dynflow-sidekiq@.service': No such file or directory`

ezr-ondrej:
Any idea why the tests are failing now? :disappointed:

ekohl:
Because there wasn't a new source tarball with the new file. There is now.
[test rpm]

packages/foreman/foreman/foreman.spec
```diff
@@ -2,6 +2,7 @@
 %global confdir extras/packaging/rpm/sources
 %global foreman_rake %{_sbindir}/%{name}-rake
 %global executor_service_name dynflowd
+%global dynflow_sidekiq_service_name dynflow-sidekiq@
```
ekohl: Given the repetition I'm inclined to add `.service` here but I'm fine either way. I'd probably do a cleanup of all service handling later anyway.

ezr-ondrej: I would like to stay consistent with the `dynflowd` service, otherwise I would agree :thinking:

ekohl:
I'm looking at https://github.com/mperham/sidekiq/blob/master/examples/systemd/sidekiq.service and it has some neat tricks:
Reloading:
```
# use `systemctl reload sidekiq` to send the quiet signal to Sidekiq
# at the start of your deploy process.
ExecReload=/usr/bin/kill -TSTP $MAINPID
```

Memory fragmentation fix:
```
# Greatly reduce Ruby memory fragmentation and heap usage
# https://www.mikeperham.com/2018/04/25/taming-rails-memory-bloat/
Environment=MALLOC_ARENA_MAX=2
```

I'm wondering if the `type=forking` that's currently used is incorrect and should be `type=simple`. Does sidekiq fork itself?

tbrisker:
considering that this is experimental in 1.24, perhaps we can merge if it works and improve afterwards if needed?

ehelms:
@tbrisker That was my plan. My comment was to indicate I did not get it to work, and to identify some issues that if we don't fix now, are captured in issues that block it going from "beta" to "GA".

ezr-ondrej:
I am trying to address the comments in theforeman/foreman#7127 - not to forget them.

adamruzicka:
Addressing the start timeout (it indeed was `simple` vs `forking`) and syslog identifier issues in https://github.com/theforeman/foreman/pull/7128

tbrisker:
what's the status here @ehelms @ekohl ?

ehelms:
@tbrisker I wanted to see this in action to know this packaging change plus the code works if someone installs it and then manually configures Redis. I have not reached that point yet, working through it.

ekohl:
Should we add a `%exclude` commit to rpm/develop in the mean time to get nightly to build again?

ehelms:
[test rpm]

PR #4195, opened 2019-10-16
Regenerate comps file

adamruzicka:


comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -271,10 +271,10 @@
       <packagereq type="default">tfm-rubygem-hammer_cli_foreman_ssh-doc</packagereq>
       <packagereq type="default">tfm-rubygem-hammer_cli_foreman_tasks-doc</packagereq>
       <packagereq type="default">tfm-rubygem-hammer_cli_foreman_templates-doc</packagereq>
+      <packagereq type="default">tfm-rubygem-httparty-doc</packagereq>
```
ekohl: I'm not sure about this sorting.

adamruzicka: That's what the script does. I can move it by hand, but it will go back on next run

ekohl: Thanks! We should really have a check that runs this command and sees if there's a diff while having the correct locale available.

PR #4196, opened 2019-10-16
Add ruby-gitlab-sidekiq-fetcher

adamruzicka:


dependencies/bionic/gitlab-sidekiq-fetcher/copyright
```diff
@@ -0,0 +1,26 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: gitlab-sidekiq-fetcher
+Upstream-Contact: Valery Sizov <valery@gitlab.com>
+Source: https://gitlab.com/gitlab-org/sidekiq-reliable-fetch
+
+Files: *
+Copyright: 2016, Rémi Bauzac
+License: LGPL-3.0
```
adamruzicka: Is this all correct? I wasn't sure about who is the copyright holder so I picked the one who put the license in the repo.

dependencies/bionic/gitlab-sidekiq-fetcher/control
```diff
@@ -0,0 +1,16 @@
...
+Section: ruby
+Priority: optional
+Maintainer: Adam Ruzicka <aruzicka@redhat.com>
+Uploaders: Adam Ruzicka <aruzicka@redhat.com>
+DM-Upload-Allowed: yes
+Build-Depends: debhelper, gem2deb
+Standards-Version: 3.9.3
+XS-Ruby-Versions: all
+
+Package: ruby-dynflow
```
ekohl: This looks odd. Are you sure this package is generated from the source?

adamruzicka: Odd indeed. It wasn't generated per-se, I created the package by copying dynflow over and changing the contents. Looks like I missed this when editing it

ekohl:
Given https://github.com/theforeman/foreman-packaging/pull/4230, I don't know if this is needed. On Debian we do bundle various dependencies. @mmoll could you take a look?

mmoll:
For what exactly is this needed? If it's in core, it's already bundled in the foreman DEB package. Only for e.g. proxy packages it would be needed as dependency.

adamruzicka:
It implements the reliable fetch pattern for sidekiq and is used when deploying dynflow the new way. I don't think it is bundled in core, but I could be wrong.

zjhuntin:
@mmoll @adamruzicka How should we continue with this PR?

adamruzicka:
I'll have to spin up a debian nightly box and see what's there, will report back when i do

ehelms:
@adamruzicka Any updates here to guide what we should do with tihs PR? it looks like it will also need a rebase

evgeni:
ping @adamruzicka is this still needed? otherwise I'm inclined to close this.

adamruzicka:
```
root@8ae45edf3ea9:~# grep PRETTY_NAME /etc/os-release 
PRETTY_NAME="Ubuntu 18.04.5 LTS"

root@8ae45edf3ea9:~# dpkg-query -s foreman | grep Version
Version: 2.2.1-1

root@8ae45edf3ea9:~# dpkg-query -L foreman | grep gitlab-sidekiq-fetcher
/usr/share/foreman/vendor/cache/gitlab-sidekiq-fetcher-0.6.1.gem
```

Looks like it is bundled in `foreman` package, so unless we start using sidekiq on the smart proxy, this can be closed.

PR #4197, opened 2019-10-17
Add rubygem-smart_proxy_dhcp_bluecat package

mdellweg:


package_manifest.yaml
```diff
@@ -414,6 +414,7 @@ foreman_nonscl_packages:
     rubygem-kafo: {}
     rubygem-kafo_parsers: {}
     rubygem-kafo_wizards: {}
+    rubygem-smart_proxy_dhcp_bluecat: {}
```
ekohl: I think we need to update our add script to the updated structure. Could you move it to the `foreman_proxy_plugin_packages` group?

mdellweg: Maybe the scripts are ok. I think if you only specify the gem name and the template, they end up in the right category. But The usage message does not suggest that.

comps/comps-foreman-rhel7.xml
```diff
@@ -323,6 +323,7 @@
       <packagereq type="default">rubygem-rb-inotify</packagereq>
       <packagereq type="default">rubygem-rsec</packagereq>
       <packagereq type="default">rubygem-rubyipmi</packagereq>
+      <packagereq type="default">rubygem-smart_proxy_dhcp_bluecat</packagereq>
```
ehelms: I believe this should end up in the plugins comps file.

PR #4201, opened 2019-10-18
Add rubygem-sentry-raven package

dgoetz:
This is a required dependency for the Foreman plugin foreman_concrete which will follow as a separate PR. I will also cherry-pick it for Foreman 1.23 afterwards.


dgoetz:
I do not understand the error (including since when the spec is also in git annex). So if anyone has some advice I would be happy!

ehelms:
The spec file appears to think its a git annex added file rather than the spec file itself. That is the first big issue its hitting.

dgoetz:
I created a new system to create and push the changes and now it works fine, so it seems like an issue with my Fedora 30.

dgoetz:
Rebased to resolve merge conflicts. Can someone do me the favor of a review? I want to get foreman_concrete packaged.

packages/plugins/rubygem-sentry-raven/rubygem-sentry-raven.spec
```diff
@@ -0,0 +1,100 @@
...
+mkdir -p %{buildroot}%{_bindir}
+cp -a .%{_bindir}/* \
+        %{buildroot}%{_bindir}/
+find %{buildroot}%{gem_instdir}/exe -type f | xargs chmod a+x
+
+%files
+%dir %{gem_instdir}
+%{_bindir}/raven
+%exclude %{gem_instdir}/.gitignore
+%{gem_instdir}/.gitmodules
```
ekohl: Is this one needed? It implies there's git submodules.

PR #4206, opened 2019-10-21
Automatically set gpgcheck=1 for release versions

evgeni:


packages/katello/katello-repos/katello.repo
```diff
@@ -31,7 +31,7 @@ name=Katello @REPO_NAME@ Source
 baseurl=https://fedorapeople.org/groups/katello/releases/source/srpm/@REPO_VERSION@/katello/@DIST@/
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-katello
 enabled=0
-gpgcheck=0
+gpgcheck=@REPO_GPGCHECK@
```
ekohl: We sign sources as well, right?

evgeni: I did for 3.13, yes.

PR #4208, opened 2019-10-21
Use RPM-GPG-KEY-foreman from foreman-release

evgeni:


evgeni:
@ehelms not sure I follow your comment. My mumbling that this should live in `foreman-release` (in https://github.com/theforeman/tool_belt/pull/205/) was still suggesting a separate sub-package, that would not be published to yum.tfm.o/releases.

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -17,11 +17,12 @@ Group:          Applications/Internet
...
 URL:            https://theforeman.org/plugins/katello/
 Source0:        katello.repo
-Source1:        RPM-GPG-KEY-katello-2015
 
 BuildArch:      noarch
 
-BuildRequires: sed
+BuildRequires:  sed
+
+Requires:       foreman-release
```
ekohl: Do we want to enforce a match here?
```suggestion
Requires:       foreman-release >= 1.24
Requires:       foreman-release < 1.25
```

ekohl:
My idea was a `katello` subpackage in `foreman-release`. I'd also publish that to yum.tf.o in the main foreman repo. That way the katello installation instructions become:

```
yum -y install epel-release http://yum.puppet.com/puppet-release-el-7.noarch.rpm https://yum.theforeman.org/releases/latest/el7/x86_64/foreman-release.rpm
yum -y install foreman-release-katello
yum -y install foreman-installer-katello
foreman-installer --scenario katello
```

ehelms:
What @ekohl said here https://github.com/theforeman/foreman-packaging/pull/4208#issuecomment-544525306 much more elegantly than I did :P

packages/katello/katello-repos/katello-repos.spec
evgeni: That would require updating the lock on every major release. I'd defer that until we merge the two packages? :)

ekohl: I guess that's fine. Users will find out it's incompatible anyway when the GPG signatures don't match ;)

evgeni:
Oh, you wrote "advantageous" not "adventurous", that changes everything.

I think moving this completely into `f-release` is the right thing to do, but requires more work, while this PR get's us most of the benefits (don't have to care about `k-repos` besides version-bump) without the headache of the move. So I'd prefer to merge this and drop "merge `f-release` and `k-repos`" to the backlog?

PR #4211, opened 2019-10-22
Update rubygem-foreman_virt_who_configure to 0.4.5

ares:


ares:
this is a 1.24 release

evgeni:
> this is a 1.24 release

what does that mean? it doesn't define a min_version of 1.24 and we didn't branch 1.24 yet :)

ares:
This should be available in 1.24 / 3.14 repos, so either get it to nightlies or if branchinh happens first, we'll need to cherrypick. The only patch didn't need 1.24 bump, hence it's just minor version update.

evgeni:
branching didn't happen yet, so you're good.

PR #4214, opened 2019-10-22
Another round of SCL nodejs packages to build

ehelms:


tbrisker:
looks like just foreman repoclosure failed which i guess is expected.

packages/foreman/nodejs-patternfly/nodejs-patternfly.spec
```diff
@@ -5,7 +5,7 @@
```
ekohl: Why is the filesize for this increasing by 120%? Shouldn't it just be a trivial rebuild?

ehelms: I assume you mean the registry tarball. Any bundled nodejs module has to have a tarball generated based on the Nodejs 10 structure since it changed from nodejs 6.

ehelms:
Since the modules cannot all be rebuilt at the same time, we end up with having to do them in chunks or many many individual PRs. The repoclosures should fix themselves once all are built and thus I would ask to merge this to avoid this churn.

packages/foreman/nodejs-patternfly/nodejs-patternfly.spec
ekohl: Still odd that it's so much larger.

PR #4231, opened 2019-10-24
Fixes #28125 - Drop foreman_cockpit

ekohl:
The plugin is broken and replaced with (optional) integration in REX.


ekohl:
Should this also be obsoleted in tfm so it's automatically removed? I think that might break but we don't really have a good way to deal with that in general.

ehelms:
Maybe we do that next release and give users one release to manually remove it?

ekohl:
You mean to merge this as-is and commit the Obsoletes right after branching? I'd be ok with that.

PR #4233, opened 2019-10-25
Bump foreman_openscap to 2.0.0

xprazak2:


packages/plugins/rubygem-foreman_openscap/rubygem-foreman_openscap.spec
```diff
@@ -24,7 +24,6 @@ Requires: %{?scl_prefix_ruby}ruby(rubygems)
 Requires: %{?scl_prefix}rubygem(deface) < 2.0
```
evgeni: you droped the `BuildRequires`, but the `Requires` is still there?

xprazak2: My mistake, fixed now.

PR #4234, opened 2019-10-25
Bump foreman_ansible to 4.0.0

xprazak2:


lzap:
Can you bump ReX dependency to 2.0? This will probably need a new gem release. Need this for https://github.com/theforeman/foreman-packaging/pull/4234

xprazak2:
@lzap, seems like that dependency is already bumped in [gemspec](https://github.com/theforeman/foreman_ansible/blob/master/foreman_ansible.gemspec#L26), I made corresponding changes here as well.

sbernhard:
[test rpm]

ares:
seems the scratch build is there, the rpm job still fails on

```
unresolved deps:
[2019-11-05T06:56:34.821Z]          tfm-rubygem(foreman_remote_execution) >= 0:2.0
```
even though 2.0.2 was merged yesterday, the pluging didn't make it to nightlies yet, which may be the cause?

evgeni:
and nightlies now fail repoclosure, because the ansible and salt in there needs older rex, going to merge as is.

PR #4240, opened 2019-10-28
don't hardcode the dot in the prerelease macro

evgeni:

### DIFF
--- a/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
+++ b/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
@@ -4,8 +4,9 @@
 %global gem_name hammer_cli
 %global confdir hammer
 
-%global release 1
-%global prerelease .pre.develop
+%global release 2
+%global prereleasesource pre.develop
+%global prerelease %{?prereleasesource:.}%{?prereleasesource}
 
 %{!?_root_bindir:%global _root_bindir %{_bindir}}
 %{!?_root_mandir:%global _root_mandir %{_mandir}}
@@ -127,6 +128,9 @@ install -m 644 .%{gem_instdir}/config/cli_config.template.yml \
 %{gem_instdir}/test
 
 %changelog
+* Mon Nov 18 2019 Evgeni Golov - 0.20-0.2.pre.develop
+- Unify prerelease macro handling
+
 * Mon Oct 28 2019 Evgeni Golov - 0.20-0.1.pre.develop
 - Bump to 0.20-develop
 
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,5 +1,6 @@
-%global release 1
-%global prerelease develop
+%global release 2
+%global prereleasesource develop
+%global prerelease %{?prereleasesource}
 
 Name:       foreman-installer
 Epoch:      1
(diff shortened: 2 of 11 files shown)


ehelms:
The prerelease macro is used in a few more places including the Source0

evgeni:
> The prerelease macro is used in a few more places including the Source0

That's what I've just force-pushed :)

evgeni:
turns out `%gem_install` uses `prerelease` internally and fails if that doesn't contain the dot :fire:

packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
```diff
@@ -4,8 +4,8 @@
 %global gem_name hammer_cli
 %global confdir hammer
 
-%global release 1
-%global prerelease .pre.develop
+%global release 2
+%global prerelease pre.develop
```
evgeni: I wonder if we should add a 
```
%global dotprerelease %{?prerelease:.}%{?prerelease}
```

ehelms: That makes the logic less repetitive so +1. Might eventually be worth a macro in tfm-build to capture it in a single place.

ehelms:
The Katello build error looks unrelated but real as it seems to be finding a source map? Which I thought we were omitting.

evgeni:
> The Katello build error looks unrelated but real as it seems to be finding a source map? Which I thought we were omitting.

So this is another place which fails because the ruby macros use `prerelease` internally. See:

```
$ grep prerelease /usr/lib/rpm/macros.d/macros.rubygems
%gem_instdir %{gem_dir}/gems/%{gem_name}-%{version}%{?prerelease}
%gem_extdir_mri %{gem_archdir}/ruby/%{gem_name}-%{version}%{?prerelease}
%gem_cache %{gem_dir}/cache/%{gem_name}-%{version}%{?prerelease}.gem
%gem_spec %{gem_dir}/specifications/%{gem_name}-%{version}%{?prerelease}.gemspec
%gem_docdir %{gem_dir}/doc/%{gem_name}-%{version}%{?prerelease}
```

I'd prefer not to have to redefine all those macros :(

So I'm seeking guidance how to continue here.

To recap, the idea of this PR was to unify the way we handle prereleases in packaging, to make it easier for obal and friends to go and "do a rc1 of $project" and that would basically be a "set prerelease to `rc1`". Currently this is not possible, as we sometimes need the value to be `.rc1` and sometimes `rc1`.

One possible solution would be to *always* include the dot in the `prerelease` macro, thus adjusting foreman, foreman-proxy, foreman-selinux and foreman-installer instead.

While I think not having the dot is cleaner, I am willing to jump through just not enough hops to make it happen ;)

ehelms:
Would your strategy proposed in (https://github.com/theforeman/foreman-packaging/pull/4240#discussion_r339960365) solve the issues you ran into with https://github.com/theforeman/foreman-packaging/pull/4240#issuecomment-549339940 ?

evgeni:
@ehelms no, the ruby rpm macros expect `%{prerelease}` to contain the dot, not any other var.

So the only cheap solution is to always have the dot there (or ignore the inconsistency, but that will make my automation life harder)

ehelms:
I :heart: consistency, I saw always have the dot there! Unless you want to capture a term for the pre state of prerelease. e.g. candidate, branch, prereleasenodot, nodotprerelease

evgeni:
Always using the dot doesn't work, as the tarball-based packages (foreman etc), actually need a dash as the prerelease-delimiter in the source line.

I went ahead and added a `prepreleasesource` macro, that contains the branch/rc-state, and then adjusted `prerelease` to contain that with either a dot prefix or not.

evgeni:
builds fine, repoclosure is red because cross-repo PR

ekohl:
Does this mean we should update obal as well to update this macro rather than the prerelease one?

evgeni:
> Does this mean we should update obal as well to update this macro rather than the prerelease one?

Yes, that was my plan.

PR #4241, opened 2019-10-28
Drop psql requirement on katello-debug

ehelms:

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -4,7 +4,7 @@
 %global homedir %{_datarootdir}/%{name}
 %global confdir common
 %global prerelease .master
-%global release 3
+%global release 4
 
 Name:       katello
 Version:    3.14.0
@@ -152,7 +152,6 @@ Requires: qpid-tools
 Requires: /bin/ps
 Requires: rh-mongodb34
 Requires: %{name}-service
-Requires: /usr/bin/psql
 
 %description debug
 Useful utilities for debug info collecting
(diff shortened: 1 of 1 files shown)


packages/katello/katello/katello.spec
```diff
@@ -152,7 +152,6 @@ Requires: qpid-tools
 Requires: /bin/ps
 Requires: rh-mongodb34
 Requires: %{name}-service
-Requires: /usr/bin/psql
```
ehelms: The foreman-debug itself uses psql and thus should handle this. Further, this ties to a specific binary that can change based on Postgres via SCL or not.

ekohl: Given https://community.theforeman.org/t/rfc-deprecate-foreman-debug-in-favor-of-sosreport/12465 should we drop it altogether?

packages/katello/katello/katello.spec
```diff
@@ -152,7 +152,6 @@ Requires: qpid-tools
 Requires: /bin/ps
 Requires: rh-mongodb34
 Requires: %{name}-service
```
ekohl: This should be changed to `foreman-maintain` since `katello-service` is no longer used.

packages/katello/katello/katello.spec
ehelms: We probably should, but is that RFC fully implemented? My read on it is no.

ekohl: sosreport has implemented all the functionality but that version has only landed in EL7. Not Debian derivatives. That doesn't really matter for Katello. Also not sure if katello users really use the upload functionality to our servers.

packages/katello/katello/katello.spec
ehelms: Can I do an atomic follow up for that change?

ekohl: Yes.

PR #4249, opened 2019-10-29
Drop katello-service in favor of foreman-maintain

ehelms:


packages/katello/katello/hostname-change.rb
```diff
@@ -400,8 +400,8 @@ def run
         if success
           STDOUT.puts result
           STDOUT.puts 'Restarting puppet services'
-          run_cmd('/sbin/service puppet restart')
-          run_cmd('katello-service restart --only puppetserver')
+          run_cmd('/bin/systemctl puppet restart')
```
ekohl: The order is reversed. We can also make it safer by using `try-restart` so it doesn't require Puppet to be running.
```suggestion
          run_cmd('/bin/systemctl try-restart puppet')
```

ehelms:
A Redmine issue where?

ekohl:
Normally I'd say https://projects.theforeman.org/projects/rpms since this is packaging but in the past you've preferred https://projects.theforeman.org/projects/katello. I just think an issue in either one is good so we can point users to the fact we've dropped katello-service.

PR #4251, opened 2019-10-30
Bump version to 1.25-develop

ekohl:


ekohl:
Repoclosure fails because it's mixing SCL and non-SCL.

PR #4255, opened 2019-10-30
update configs to use stable 1.24 repos/tags/etc

zjhuntin:


package_manifest.yaml
```diff
@@ -36,15 +36,9 @@ foreman_scl_packages:
...
+        - el7-foreman-rails-1.24
   hosts:
-    foreman:
-      releasers:
-        - "{{ nightly_releaser }}"
-      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
-      nightly_package_tito_releaser_args:
-        - "jenkins_job=foreman-develop-source-release"
-        - "jenkins_url=https://ci.theforeman.org"
+    foreman: {}
```
ehelms: The line below this needs to be dropped as well, the one about FetchBuilder

mock/el7-scl.cfg
```diff
@@ -49,17 +49,17 @@ baseurl=http://mirror.centos.org/centos/7/sclo/x86_64/rh/
...
 
 [foreman-rails]
 name=foreman-rails
-baseurl=https://yum.theforeman.org/rails/foreman-nightly/el7/x86_64
+baseurl=https://yum.theforeman.org/rails/foreman-1.24/el7/x86_64
 
 [katello]
 name=katello
-baseurl=https://fedorapeople.org/groups/katello/releases/yum/nightly/katello/el7/x86_64/
+baseurl=https://fedorapeople.org/groups/katello/releases/yum/1.24/katello/el7/x86_64/
```
ekohl:
```suggestion
baseurl=https://fedorapeople.org/groups/katello/releases/yum/3.14/katello/el7/x86_64/
```

repoclosure/yum_el7.conf
```diff
@@ -45,32 +45,32 @@ baseurl=https://yum.puppetlabs.com/el/7/PC1/x86_64/
...
-name=Katello Pulp nightly EL7
+[el7-katello-pulp-3.14]
+name=Katello Pulp 3.14 EL7
 baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/stable/2.19/7Server/x86_64/
 
-[el7-katello-candlepin-nightly]
-name=Katello Candlepin nightly EL7
-baseurl=http://koji.katello.org/releases/yum/katello-nightly/candlepin/el7/x86_64/
+[el7-katello-candlepin-3.14]
+name=Katello Candlepin 3.14EL7
```
ekohl:
```suggestion
name=Katello Candlepin 3.14 EL7
```

repoclosure/yum_el7.conf
```diff
@@ -45,32 +45,32 @@ baseurl=https://yum.puppetlabs.com/el/7/PC1/x86_64/
...
-baseurl=http://koji.katello.org/releases/yum/katello-nightly/katello/el7/x86_64/
+[el7-katello-3.14]
+name=Katello 3.14 EL7
+baseurl=http://koji.katello.org/releases/yum/katello-3.14/katello/el7/x86_64/
 
-[el7-katello-pulp-nightly]
-name=Katello Pulp nightly EL7
+[el7-katello-pulp-3.14]
+name=Katello Pulp 3.14 EL7
 baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/stable/2.19/7Server/x86_64/
```
ekohl: Do we still have this? Should it at least be updated to 2.21? That means we probably need to do the same in develop.

repoclosure/yum.conf
```diff
@@ -45,35 +45,35 @@ baseurl=https://yum.puppetlabs.com/el/7/PC1/x86_64/
 name=el7-puppet-6
 baseurl=https://yum.puppetlabs.com/puppet6/el/7/x86_64/
 
-[el7-foreman-rails-nightly]
-name=Foreman Rails nightly EL7
+[el7-foreman-rails-1.24]
+name=Foreman Rails 1.24 EL7
 baseurl=http://koji.katello.org/releases/yum/tfm-ror52/el7/x86_64/
```
ekohl: For 1.23 we've created a `foreman-rails-1.23` tag that branched off `tfm-ror52`. so this probably needs an update. I'd check with 1.23.

repoclosure/yum_el7.conf
```diff
@@ -45,32 +45,32 @@ baseurl=https://yum.puppetlabs.com/el/7/PC1/x86_64/
...
-name=Katello nightly EL7
-baseurl=http://koji.katello.org/releases/yum/katello-nightly/katello/el7/x86_64/
+[el7-katello-3.14]
+name=Katello 3.14 EL7
+baseurl=http://koji.katello.org/releases/yum/katello-3.14/katello/el7/x86_64/
 
-[el7-katello-pulp-nightly]
-name=Katello Pulp nightly EL7
-baseurl=https://repos.fedorapeople.org/repos/pulp/pulp/stable/2.19/7Server/x86_64/
+[el7-katello-pulp-2.21]
```
ekohl: I think we use 3.14 here because in Jenkins/obal we don't care about the exact version. We'll just say "we need the pulp config for katello 3.14, whatever that may be".
```suggestion
[el7-katello-pulp-3.14]
```

rel-eng/tito.props
```diff
@@ -459,7 +459,7 @@ whitelist = ansiblerole-foreman_scap_client
   rubygem-smart_proxy_salt
   rubygem-smart_proxy_spacewalk
 
-[foreman-client-nightly-rhel7]
+[foreman-client-1.24-rhel7]
 disttag = .el7
 whitelist =
   # don't whitelist foreman-release here, it's already in foreman-nightly-nonscl-rhel7
```
ekohl: FYI: this is a limitation in obal. This is because obal can't release the same package to multiple tags and you must manually tag it in after every release. You may want to update the comment to refer to 1.24 instead of nightly as well.

PR #4266, opened 2019-11-01
Bump Katello to 3.15.0

jturel:


ekohl:
Can you bump `katello-repos` and `rubygem-katello` while you're at it? That will also unbreak nightlies.

jturel:
> Can you bump `katello-repos` and `rubygem-katello` while you're at it? That will also unbreak nightlies.

It's all there - I think you looked before my force-push yesterday :)

PR #4268, opened 2019-11-03
foreman_datacenter 1.24.0

Zenya:


mmoll:
There are no changes in the plugin source from 1.23.1 to 1.24, AFAICT. Also, 1.2.4.0 is of course not yet released.

PR #4276, opened 2019-11-04
Refs #28158 - Bump katello rpm for hostname-change

chris1984:
@ekohl here is the packaging pr for the hostname-change with removing the `csr`


chris1984:
@ehelms is this good for a merge?

PR #4279, opened 2019-11-04
update pulp_file_client gem

jlsherrill:

### DIFF
--- a/packages/katello/rubygem-pulp_file_client/rubygem-pulp_file_client.spec
+++ b/packages/katello/rubygem-pulp_file_client/rubygem-pulp_file_client.spec
@@ -6,7 +6,7 @@
 %global gem_name pulp_file_client
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 3.0.0rc2.dev.1558441126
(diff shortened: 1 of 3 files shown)


packages/katello/rubygem-pulp_file_client/rubygem-pulp_file_client.spec
```diff
@@ -6,7 +6,8 @@
 %global gem_name pulp_file_client
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 3.0.0rc2.dev.1558441126
+Version: 0.1.0b5.dev01571253617
+Epoch: 1
```
ekohl: Do we need to use epoch? I thought we removed the older version from the repo. If it's only to get a green build, I'd say we should ignore that.

jlsherrill: hahah, so this was never released in upstream, but did appear in nightlies.  I'm okay not using epoch, but figured we'd want to stick to the letter of the law.  But if you're good with no epoch, so am i :)

PR #4287, opened 2019-11-05
don't update taxonomies settings -- they are on by default

evgeni:


evgeni:
@tbrisker double check me on those, but we shouldn't need to sed around anymore, right?

mmoll:
merged, спасибо Женя!

PR #4292, opened 2019-11-05
Update rubygem-foreman_discovery to 16.0.0

ares:


packages/plugins/rubygem-foreman_discovery/rubygem-foreman_discovery.spec
```diff
@@ -17,7 +17,7 @@
 
 Summary:    MaaS Discovery Plugin for Foreman
 Name:       %{?scl_prefix}rubygem-%{gem_name}
-Version:    %{mainver}
+Version:    16.0.0
```
evgeni: any reason not to keep this at `%{mainver}`?

ares: bump_rpm.sh did that... reverting to %{mainver}

evgeni: Mhh. Not sure the mainver is really needed anyways, but let's keep this small. Ack

PR #4297, opened 2019-11-06
Bump foreman_ansible to 4.0.0

xprazak2:
(cherry picked from commit 8523b3a41bb6f60f59c688403bc53dfcfaa5f5fd)

PR #4294, opened 2019-11-05
Bump foreman_salt / smart_proxy_salt versions

sbernhard:

PR #4298, opened 2019-11-06
REX and Ansbile for 1.24

adamruzicka:


zjhuntin:
Is there also supposed to be an update for Ansible in this?

adamruzicka:
I included https://github.com/theforeman/foreman-packaging/pull/4297 at first, but the build failed, so I stripped it down to just REX

tbrisker:
The repoclosure is failing because the salt pr (GH-4294) isn't merged yet, and that pr is failing because this one isn't in yet - we had the same issue in rpm/develop. we should merge this and retrigger the ansible (GH-4297) and salt (GH-4294) ones.

PR #4301, opened 2019-11-06
Release foreman 1.22.2

zjhuntin:


packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -137,6 +137,9 @@ done
 %{_sbindir}/foreman-proxy-certs-generate
 
 %changelog
+* Wed Nov 06 2019 b'Zach Huntington-Meath <zhunting@redhat.com>' - b'1:1.22.2-1'
```
tbrisker: not sure where these `b'` came from but looks like they are causing the build to fail (also in all other specs except foreman.spec)

PR #4308, opened 2019-11-08
Update rubygem-foreman_fog_proxmox to 0.8.2

tristanrobert:
WARNING: needs #4306 to be approved previously!!!


evgeni:
@tristanrobert if this needs #4306, should the dependency on the proxmox gem be bumped too?

tristanrobert:
@evgeni Yes!

tristanrobert:
You can forget this PR because it has been approved after 0.9.x so conflicts appears.

PR #4310, opened 2019-11-08
Updated foreman-fog-proxmox to 0.8.2

tristanrobert:
Nightly

### DIFF
--- a/plugins/ruby-foreman-fog-proxmox/debian/changelog
+++ b/plugins/ruby-foreman-fog-proxmox/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-fog-proxmox (0.8.2-1) stable; urgency=low
+
+  * 0.8.2 released
+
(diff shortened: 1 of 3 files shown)


mmoll:
@tristanrobert this version is for nightly all the way down to 1.22, I presume?

tristanrobert:
Yes it is for nightly. The PR one for 1.22 is asked in #4311.

mmoll:
@tristanrobert I'm asking because in this case usually we just "want"" the PR for nightly and then cherry-pick that change into the otehr brnaches, as 1.23 and 1.24 would get missed with a per-version approach.

So my real question is: Can the change in this PR be used for all versions from nightly to 1.22?

tristanrobert:
Sorry I didn't understand your question. Yes this change can be used to 1.22. I have tried cherry-pick but the number of diff commits were to high between the two branches.

mmoll:
merged, merci @tristanrobert!

mmoll:
- 1.24: a2172bf
- 1.23: 57bd189
- 1.22: 5a96940

PR #4318, opened 2019-11-11
begin pulpcore packaging

evgeni:


evgeni:
so good you can't flip a PR to WIP after creating it.

package_manifest.yaml
```diff
@@ -953,3 +954,30 @@ repoclosures:
...
+        - el7-updates
+        - el7-epel
+        - el7-extras
+        - el7-scl
+        - el7-puppet-6
+        - el7-foreman-rails-nightly
+        - el7-foreman-nightly
+        - el7-foreman-plugins-nightly
+        - el7-katello-pulp-nightly
+        - el7-katello-candlepin-nightly
```
ehelms: All these are needed as lookasides for Pulp?

rel-eng/releasers.conf
```diff
@@ -43,3 +43,8 @@ autobuild_tags = katello-nightly-rhel7
 builder = tito.builder.FetchBuilder
 builder.jenkins_url = https://ci.theforeman.org
 builder.rpmbuild_options = --define "foremandist .fm1_25"
+
+[koji-katello-pulpcore]
+releaser = tito.release.KojiReleaser
+autobuild_tags = katello-nightly-pulpcore-el7
+builder.rpmbuild_options = --define "foremandist .fm1_25"
```
ehelms: I assume this is not needed for Pulp packages.

package_manifest.yaml
evgeni: no

rel-eng/tito.props
```diff
@@ -529,3 +529,8 @@ whitelist = katello
   rubygem-runcible
   rubygem-typhoeus
   rubygem-zest
+
+[katello-nightly-pulpcore-el7]
+disttag = .el7
+whitelist =
+        python-setuptools-scm
```
ehelms: Nit, indented more than other whitelists in the file.

evgeni: blame the tools ;)
but fair

add_pypi_package.sh
```diff
@@ -73,12 +74,16 @@ add_to_tito_props() {
 }
 
 add_pypi_to_comps() {
-  local comps_scl="nonscl"
   local comps_packages=$(rpmspec --query --builtrpms --queryformat '%{NAME}\n' $PACKAGE_DIR/*.spec)
-  if [[ $TITO_TAG == katello-* ]]; then
+  if [[ $TITO_TAG == katello-*-pulpcore-* ]]; then
+    local comps_file="katello-pulpcore-server"
```
ekohl: Will we also have a `pulpcore-client`?

comps/comps-katello-pulpcore-server-el7.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<!DOCTYPE comps PUBLIC "-//Red Hat, Inc.//DTD Comps info//EN" "comps.dtd">
+<comps>
+  <!--  <meta> -->
+  <!-- Meta information will go here eventually -->
+  <!--  </meta> -->
+  <group>
+    <id>katello-pulpcore-server</id>
```
ekohl: Do we forsee client packages? Won't we put those in `foreman-client`?

packages/pulp/python-setuptools-scm/python-setuptools-scm.spec
```diff
@@ -0,0 +1,56 @@
...
+
+BuildRequires:  python3-devel
+BuildRequires:  python3-setuptools
+
+%description
+setuptools_scm setuptools_scm handles managing your Python package versions in
+SCM metadata instead of declaring them as the version argument or in a SCM
+managed file.It also handles file finders for the supported SCMs. setup.py
+usage To use setuptools_scm just modify your project's setup.py file like
+this:* Add setuptools_scm to the setup_requires parameter. * Add the
```
ekohl: Do we care about a readable description?

pyp2rpm/pulpcore.spec
```diff
@@ -0,0 +1,104 @@
+{{ data.credit_line }}
```
ekohl: Wondering if this file should have `.j2` file extension.

add_pypi_package.sh
```diff
@@ -39,7 +40,7 @@ generate_pypi_package() {
   else
     RPM_NAME_ARG=""
   fi
-  pyp2rpm --no-autonc -s -d $PACKAGE_DIR -v $VERSION $RPM_NAME_ARG $PYPI_NAME
+  pyp2rpm --no-autonc -s -t $TEMPLATE -b 3 -d $PACKAGE_DIR -v $VERSION $RPM_NAME_ARG $PYPI_NAME
```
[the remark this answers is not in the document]
evgeni: yes, -b3 means "use python3 as base python"

add_pypi_package.sh
evgeni: no, this was mainly following the fact that all katello-related comps are called -server

pyp2rpm/pulpcore.spec
evgeni: only if we want `pyp2rpm` not find it ;)

packages/pulp/python-setuptools-scm/python-setuptools-scm.spec
evgeni: IMHO not, this one is actually almost readable.

pyp2rpm/pulpcore.spec
evgeni: (it uses splitext internally, and having two extensions would confuse that)

add_pypi_package.sh
evgeni: @pcreech any naming preferences from your POV? I think droping `-server` is the right way?

pcreech: Kill "-server" if it's not needed

evgeni:
this now has been updated for all comments but the `%description` one, as I don't foresee us to fix up the generated descriptions manually for all packages.

add_pypi_package.sh
```diff
@@ -39,18 +41,14 @@ generate_pypi_package() {
...
   fi
-  pyp2rpm --no-autonc -s -d $PACKAGE_DIR -v $VERSION $RPM_NAME_ARG $PYPI_NAME
+  pyp2rpm --no-autonc -s -t $TEMPLATE -b $BASE_PYTHON -d $PACKAGE_DIR -v $VERSION $RPM_NAME_ARG $PYPI_NAME
   echo "FINISHED"
   if [[ $UPDATE == true ]]; then
     echo "Restoring changelogs..."
     cat OLD_CHANGELOG >> $PACKAGE_DIR/*.spec
     sed -i '/^%changelog/,/^%changelog/{0,//!d}' $PACKAGE_DIR/*.spec
     rm OLD_CHANGELOG
-    CHANGELOG="- Update to $VERSION"
```
ekohl: Isn't this part still useful?

rel-eng/tito.props
```diff
@@ -529,3 +529,8 @@ whitelist = katello
   rubygem-runcible
   rubygem-typhoeus
   rubygem-zest
+
+[katello-nightly-pulpcore-el7]
```
ekohl: If we look at the foreman tags then I think `katello-pulpcore-nightly-el7` might be more consistent.

add_pypi_package.sh
evgeni: `pyp2rpm` currently does not support setting the changelog message, so this would generate two entries. I'd like to fix `pyp2rpm`, but not right now.

ekohl: So you end up with the old changelog and a new 'Initial package` entry that you need to manually modify?

evgeni: at the moment: yes

rel-eng/tito.props
evgeni: maybe, but that's how @pcreech created it in Koji via https://github.com/theforeman/tool_belt/pull/222

ehelms: That's a fair thought, should be easy to edit the tag name I believe. I agree it meets with more of a convention of:

<name>-<version>-<dist>

We should write that convention down.

ekohl: Edited your post to add code tags so Github doesn't attempt to parse it.

ekohl:
Now that the tag is still empty, I'd prefer to do it right. Probably quite trivial to do it now.

ehelms:
It is trivial, but seems unnecessary to block the follow up work on when it's something trivial that can be fixed later by the folks working on this change.

pcreech:
It's not as trivial as it initially may seem, as the tag config is already in tool_belt, necessitating a round of review before creating the new tag, potentially causing 1 day+ delays to the process.  The pulp package count is also small enough that it will be almost as trivial to move to the new tag later, and this will unblock work now.

I'm all for ensuing it's the most correct it can be, but in this case I don't see the value in halting everything to change what is essentially a label.

PR #4319, opened 2019-11-11
add add_pypi script

evgeni:


ekohl:
I've been wondering about finishing https://github.com/theforeman/foreman-packaging/pull/2448. The complexity there is mostly in the comps file handling, other than that it should be mostly ready IIRC.

pcreech:
@ekohl That would be nice to have, but I don't think it should block this for the time being.

@evgeni This worked well, minus the add_host script bombing out on ruamel initially (not a problem for this PR)

ACK

PR #4327, opened 2019-11-12
add build binary python packages to comps

evgeni:
they will start with python2 or python3 instead of python

### DIFF
--- a/add_pypi_package.sh
+++ b/add_pypi_package.sh
@@ -77,10 +77,12 @@ add_to_tito_props() {
 
 add_pypi_to_comps() {
   local comps_scl="nonscl"
-  local comps_package="${PACKAGE_NAME}"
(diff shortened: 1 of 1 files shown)


add_pypi_package.sh
```diff
@@ -77,10 +77,12 @@ add_to_tito_props() {
 
 add_pypi_to_comps() {
   local comps_scl="nonscl"
-  local comps_package="${PACKAGE_NAME}"
+  local comps_packages=$(rpmspec --query --builtrpms --queryformat '%{NAME}\n' $PACKAGE_DIR/*.spec)
```
ekohl: Can there be macros so it either outputs Python 2 or Python 3 depending on the host platform? For example, when I run it on Fedora, can it still output Python 2 for EL7 if needed?

evgeni: "hah" (no)

pcreech:
confirmed pre and post commit behavior, and post-commit behavior is what we want

PR #4335, opened 2019-11-13
Refs #28257 - Drop mysql support

tbrisker:
I'm guessing i did this the wrong way, but just to start the ball rolling.

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -11,7 +11,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 3
+%global release 4
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -131,10 +131,6 @@ foreman_scl_packages:
     rubygem-locale: {}
     rubygem-memoist: {}
     rubygem-multipart-post: {}
-    rubygem-mysql2:
```
ekohl: This is currently still needed for the `smart_proxy_dns_powerdns` package but I've deprecated the direct DB versions so I can release a new version that just supports the API. https://github.com/theforeman/smart_proxy_dns_powerdns/commit/deec2285488109c66b78c9ebae62456098602ee7 was already merged. Since it has no extra dependencies I also considered merging it into `smart-proxy` itself but I'll first release 0.5.0.

packages/plugins/rubygem-smart_proxy_dns_powerdns/rubygem-smart_proxy_dns_powerdns.spec
```diff
@@ -79,6 +77,9 @@ cp -pa .%{gem_instdir}/config/dns_powerdns.yml %{buildroot}%{foreman_proxy_setti
 %doc %{gem_instdir}/README.md
 
 %changelog
+* Wed Nov 13 2019 Tomer Brisker <tbrisker@gmail.com> - 0.4.0-2
```
ekohl: I'll release 0.5.0 with https://github.com/theforeman/smart_proxy_dns_powerdns/commit/deec2285488109c66b78c9ebae62456098602ee7 since that's cleaner IMHO.

tbrisker: ping @ekohl - still waiting on the new release

ekohl:
If you want to unblock yourself you can limit the PR to just dropping mysql from the foreman and leave the other cleanup till later.

tbrisker:
Updated to only drop the foreman-mysql package, build is now expectadly failing on:
```
BUILDSTDERR: error: Installed (but unpackaged) file(s) found:
BUILDSTDERR:    /usr/share/foreman/bundler.d/mysql2.rb
```

tbrisker:
Would it be better to make foreman obsolete foreman-mysql2 like we did with foreman-compute?

ekohl:
Not in this case. In foreman-compute it was a meta package that others could depend on. There was a straight migration path that yum could do. Now it involves a data migration so I prefer the explicit in your face "it breaks".

mmoll:
merged, thanks @tbrisker!

PR #4336, opened 2019-11-13
Refs #28257 - Drop mysql support

mmoll:


debian/bionic/foreman/control
```diff
@@ -100,20 +100,6 @@ Description: metapackage providing libvirt dependencies for Foreman
  .
  The foreman package should be installed along with this package.
 
-Package: foreman-mysql2
-Architecture: all
-Section: web
-Priority: extra
-Replaces: foreman-mysql
```
tbrisker: Should we add `Replaces: foreman-mysql2` on the main foreman package to make sure it gets removed on upgrades?

mmoll: Added.

debian/bionic/foreman/control
```diff
@@ -3,8 +3,10 @@ Maintainer: Greg Sutcliffe <greg.sutcliffe@gmail.com>
 Section: web
 Priority: extra
 Standards-Version: 3.9.1
+Replaces: foreman-mysql2
```
tbrisker: don't these belong on line 22-23?

mmoll: Indeed.

tbrisker:
build is failing because it still tries to install the mysql2 gem and fails

debian/bionic/foreman/control
```diff
@@ -19,8 +19,8 @@ Depends: ruby, ruby2.5, bundler (>= 1.3), zlib1g-dev,
          rake (>=0.8.4), ${misc:Depends}
 Recommends: foreman-proxy, foreman-debug (= ${binary:Version}), python-minimal
 Conflicts: ruby-activerecord-deprecated-finders, ruby-foreman-tasks (<< 0.10.1), ruby-sequel-pg
-Replaces: foreman-compute
-Breaks: foreman-compute
+Replaces: foreman-compute, foreman-mysql2
```
ekohl: I don't think this is a good idea. If the user hasn't migrated, this like breaks things. I'd rather it doesn't work and tell the user to remove `foreman-mysql2` in the migration guide.

tbrisker: seems reasonable to me. let's make sure people explicitly remove the package when upgrading

mmoll: updated.

PR #4345, opened 2019-11-18
pulpcore packaging

evgeni:


evgeni:
this includes #4318

packages/pulp/python-yarl/python-yarl.spec
```diff
@@ -0,0 +1,47 @@
...
+ :target: :target:
+
+%package -n     python3-%{pypi_name}
+Summary:        %{summary}
+%{?python_provide:%python_provide python3-%{pypi_name}}
+
+Requires:       python3-idna >= 2.0
+Requires:       python3-multidict >= 4.0
+%description -n python3-%{pypi_name}
+ :target: :target:
```
ehelms: This seems curious.

evgeni: it's what's in their README.rst (or rather, how it's mangled on pyp2rpm) and I don't feel like fixing that given the packages are not for general use outside of being a dependency for pulpcore.

packages/pulp/python-django/python-django.spec
```diff
@@ -0,0 +1,69 @@
...
+%py3_build
+
+%install
+%py3_install
+
+%files -n python3-%{srcname}
+%license django/dispatch/license.txt django/contrib/gis/gdal/LICENSE django/contrib/gis/geos/LICENSE django/contrib/admin/static/admin/js/vendor/xregexp/LICENSE.txt django/contrib/admin/static/admin/js/vendor/jquery/LICENSE.txt django/contrib/admin/static/admin/js/vendor/select2/LICENSE.md django/contrib/admin/static/admin/css/vendor/select2/LICENSE-SELECT2.md django/contrib/admin/static/admin/fonts/LICENSE.txt django/contrib/admin/static/admin/img/LICENSE docs/_theme/djangodocs/static/fontawesome/LICENSE.txt LICENSE LICENSE.python
+%doc django/contrib/admin/static/admin/fonts/README.txt django/contrib/admin/static/admin/img/README.txt docs/_theme/djangodocs/static/fontawesome/README.md tests/README.rst README.rst extras/README.TXT
+%{_bindir}/django-admin
+%{_bindir}/django-admin.py
```
ekohl: I'm wondering if we need to clutter $PATH with both

packages/pulp/python-aiofiles/python-aiofiles.spec
```diff
@@ -0,0 +1,54 @@
...
+%global pypi_name aiofiles
+
+Name:           python-%{pypi_name}
+Version:        0.4.0
+Release:        1%{?dist}
+Summary:        File support for asyncio
+
+License:        Apache 2.0
+URL:            https://github.com/Tinche/aiofiles
+Source0:        https://files.pythonhosted.org/packages/source/a/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
```
ekohl: Can we use the `%pypi_source` macro or does that break other tools?

packages/pulp/python-dotenv/python-dotenv.spec
```diff
@@ -0,0 +1,54 @@
...
+ _______ .__ __. ____ ____ | ____|| \ | | \ \ / / | |__ | \| | \ \/ / | __| | .
+| \ / __ | |____ | |\ | \ / (__)|_______||__| \__| \__/python-dotenv | [![Build
+Status]( [![Coverage Status]( [![PyPI version]( [![Say Thanks!](
+
+%prep
+%autosetup -n %{pypi_name}-%{version}
+# Remove bundled egg-info
+rm -rf %{pypi_name}.egg-info
+
+sed -i 's/long_description = f.read.*/long_description = "dotenv"/' setup.py
```
ekohl: What's the benefit of this?

packages/pulp/python-pulpcore/python-pulpcore.spec
```diff
@@ -0,0 +1,82 @@
...
+License:        GPLv2+
+URL:            http://www.pulpproject.org
+Source0:        https://files.pythonhosted.org/packages/source/p/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+BuildArch:      noarch
+
+BuildRequires:  python3-devel
+BuildRequires:  python3-setuptools
+
+%description
+[![Build Status]( [![PyPI]( [![codecov]( [![Code Quality: Python]( [![Total
```
ekohl: Given how central this package is, this probably deserves some special attention.

packages/pulp/python-pyyaml/python-pyyaml.spec
```diff
@@ -0,0 +1,63 @@
...
+License:        MIT
+URL:            https://github.com/yaml/pyyaml
+Source0:        https://files.pythonhosted.org/packages/source/P/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+
+BuildRequires:  gcc
+BuildRequires:  libyaml-devel
+
+BuildRequires:  python3-devel
+BuildRequires:  python3-setuptools
+%if 0%{?rhel} == 8
```
ekohl: Cython is just available on EL7?

packages/pulp/python-pyyaml/python-pyyaml.spec
```diff
@@ -0,0 +1,63 @@
...
+
+%build
+%py3_build
+
+%install
+%py3_install
+
+%files -n python3-%{srcname}
+%license LICENSE
+%{python3_sitearch}/*
```
ekohl: Isn't this considered a bad practice?

packages/pulp/python-six/python-six.spec
```diff
@@ -0,0 +1,47 @@
...
+%build
+%py3_build
+
+%install
+%py3_install
+
+%files -n python3-%{pypi_name}
+%license LICENSE
+%doc README.rst
+%{python3_sitelib}/__pycache__/*
```
ekohl: Can't this include the `pypi_name` to be explicit?

packages/pulp/python-six/python-six.spec
```diff
@@ -0,0 +1,47 @@
...
+License:        MIT
+URL:            https://github.com/benjaminp/six
+Source0:        https://files.pythonhosted.org/packages/source/s/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+BuildArch:      noarch
+
+BuildRequires:  python3-devel
+BuildRequires:  python3-setuptools
+
+%description
+ Six is a Python 2 and 3 compatibility library. It provides utility functions
```
ekohl: A `lstrip()` might be nice on the description. Think I saw it elsewhere too.

packages/pulp/python-typing-extensions/python-typing-extensions.spec
```diff
@@ -0,0 +1,55 @@
...
+%build
+%py3_build
+
+%install
+%py3_install
+
+%files -n python3-%{pypi_name}
+%license LICENSE
+%doc README.rst
+%{python3_sitelib}/__pycache__/*
```
ekohl: Can this also be explicit?

pyp2rpm/pulpcore.spec
```diff
@@ -0,0 +1,104 @@
...
+
+{%- if not data.has_extension %}
+BuildArch:      noarch
+{%- endif %}
+{%- for pv in data.sorted_python_versions %}
+{{ dependencies(data.build_deps, False, pv, data.base_python_version, False) }}
+{%- endfor %}
+
+%description
+{{ data.description|truncate(400)|wordwrap }}
```
ekohl: This probably also needs a strip.

packages/pulp/python-aiofiles/python-aiofiles.spec
evgeni: I intentionally didn't use it, as it breaks with packages where the package name does not match the PyPI name

packages/pulp/python-dotenv/python-dotenv.spec
evgeni: not failing on Unicode errors on EL7, probably warrants a comment

packages/pulp/python-pyyaml/python-pyyaml.spec
evgeni: it's only available for py2 in EL7.

evgeni: fixed by not building the c code :)

packages/pulp/python-six/python-six.spec
evgeni: fixed by not building the c code :)

packages/pulp/python-pyyaml/python-pyyaml.spec
evgeni: fixed by not building the c code :)

packages/pulp/python-aiofiles/python-aiofiles.spec
evgeni: For Django, `pyp2rpm` sets `srcname` to `django` and `pypi_name` to `Django`, as `srcname` wins in `pypi_source` and that 404s.

It might be an issue how pyp2rpm uses these two macros, but I didn't want to investigate further.

packages/pulp/python-django/python-django.spec
evgeni: Don't *need*, but that's what's generated and I tried not to touch unless really required.

packages/pulp/python-six/python-six.spec
evgeni: that's what's in the default pyp2rpm template, I didn't investigate further.

packages/pulp/python-box/python-box.spec
```diff
@@ -0,0 +1,51 @@
...
+
+%description -n python3-%{srcname}
+%{summary}
+
+%prep
+%autosetup -n %{pypi_name}-%{version}
+# Remove bundled egg-info
+rm -rf %{pypi_name}.egg-info
+
+sed -i '/setup_requires/d' setup.py
```
ekohl: Could also use a comment.

evgeni:
The pulpcore naming comes from avoiding pulp3 as 3 is just the API version but also not wanting pulp to not confuse with pulp2. I'm good with renaming the folder to pulpcore.

@pcreech?

packages/pulp/python-pulpcore/python-pulpcore.spec
```diff
@@ -0,0 +1,94 @@
...
+- Manage content from multiple sources in one place
+- Promote content through different repos in an organized way
+
+%package -n     python3-%{pypi_name}
+Summary:        %{summary}
+%{?python_provide:%python_provide python3-%{pypi_name}}
+Requires:       python3-coreapi >= 2.3.3
+Conflicts:      python3-coreapi >= 2.4
+Requires:       python3-djangorestframework-queryfields >= 1.0.0
+Conflicts:      python3-djangorestframework-queryfields >= 1.1
```
ehelms: Is this Requires/Conflicts a newer convention? I am used to doing a greater and less than?

evgeni: Yes. See https://github.com/fedora-python/pyp2rpm/pull/196

Doing two requires as you say can result in different packages satisfying the requirement. On Fedora/EL8 we could use rich requires, but not here on EL7.

pcreech:
@evgeni @ekohl Personally, I don't have a strong preference on what the actual folder names are.  There is no explicit 1-1 between koji tags and obal package groups, although some entries do have correlation.  There is some benefits to consistency, but i'm personally unwilling to block merging over bike-shedding the group folder.

pcreech:
[rpm]

pcreech:
[test rpm]

packages/pulp/python-pulpcore/python-pulpcore.spec
ehelms: I'm not getting it. Is there an example of how `Requires: python3-djangorestframework-queryfields < 1.1.0` results in a different dependency resolution than the conflicts? Consider this not a blocker, but more of I am interested to understand this RPM nuance. And also understand why we wouldn't adopt this pattern in other places.

pcreech: We ran into this when we were attempting to lock down python-celery and friends.  The short of it is, on yum based systems, nothing *just works* on enforcing the maximum versions, and using requires alone to enforce the issue causes misleading errors.  i.e. it will just say "nothing provides xxxx", while using an explicit conflicts will give you an error expressing the explicit conflicts.

It would seem that pyp2rpm's way of handling this issue is to forego the requires aspect of it for a conflicts only approach.

ehelms: My takeway is: both cases work properly with respect to yum, conflicts just gives an error output that is easier for a user to parse than double requires?

packages/pulp/python-pulpcore/python-pulpcore.spec
```diff
@@ -0,0 +1,94 @@
...
+Requires:       python3-aiohttp
+Requires:       python3-backoff
+Requires:       python3-django >= 2.2.3
+Conflicts:      python3-django >= 2.3
+Requires:       python3-django-filter >= 2.2.0
+Conflicts:      python3-django-filter >= 2.3
+Requires:       python3-django-rest-framework >= 3.10.2
+Conflicts:      python3-django-rest-framework >= 3.11
+Requires:       python3-dynaconf < 2.3
+Requires:       python3-dynaconf >= 2.2
```
ehelms: Related to above, only cause this is damn confusing. And I get that this is auto-generated so we shouldn't have to care. But here is a use of Requires/Requires. And in other places, Requires/Conflicts. How am I supposed to know when to use what?

packages/pulp/python-pulpcore/python-pulpcore.spec
pcreech: "work properly" is arguable, they more "work identically".  With multiple repos, yum will never do the wrong thing, but it also will not always find the right one, necessitating either repo priority, or an alternative solution (hence the high epoch on the celery environment in pulp).

But yes, that would be a good assumption.

packages/pulp/python-pulpcore/python-pulpcore.spec
pcreech: agreed, the mixed usage is confusing at best.   This appears to come down to handling of the pep440  dependencies in setup.py 

https://github.com/pulp/pulpcore/blob/master/setup.py#L13
vs 
https://github.com/pulp/pulpcore/blob/master/setup.py#L23

in other words, it's doing what pulpcore's setup.py told it to do

evgeni: Yes, `~=` gets translated to `Requires/Conflicts`, and an explicit range gets translated to two `Requires`.

packages/pulp/python-typing-extensions/python-typing-extensions.spec
evgeni: yes

evgeni:
lol, the build failed during notification to GH :woman_shrugging:

PR #4353, opened 2019-11-18
Refs #26631 - Set MALLOC_ARENA_MAX via the service

ekohl:
By setting it via the service, there is no .rpmnew changes when a user made manual changes. If the user set it manually via /etc/sysconfig/dynflowd that will still take precedence.

This is Refs since it should be cherry picked into 1.24. It will fix it also for upgrading users.


ekohl:
Note I didn't add a changelog since the nightly will take care of that and avoids a conflict in cherry picking where the next RC changelog will take care of it.

ehelms:
I am less of a fan of the nightly argument as it adds an exception to the workflow rather than just bump release for any change. If you want a clean cherry pick you can do two commits?

PR #4367, opened 2019-11-19
Add apipie-cache macros to scc_manager

m-bucher:
@evgeni seems that I have forgotten to add the apipie-cache creation in the spec-file, sorry:frowning_face:
Fix for #4331


m-bucher:
Thanks @evgeni !
I opened PRs, cherry-picking the fix to the relevant branches.

PR #4375, opened 2019-11-20
Add python-keycloak-httpd-client-install package

upadhyeammit:
This package contains libraries and tools which can automate and simplify configuring an Apache HTTPD authentication module and registering as a client of a Keycloak IdP.

### DIFF
--- a/comps/comps-foreman-rhel7.xml
+++ b/comps/comps-foreman-rhel7.xml
@@ -121,8 +121,8 @@
       <packagereq type="default">tfm-nodejs-redux</packagereq>
       <packagereq type="default">tfm-nodejs-redux-thunk</packagereq>
       <packagereq type="default">tfm-nodejs-regenerator-runtime</packagereq>
-      <packagereq type="default">tfm-nodejs-reselect</packagereq>
       <packagereq type="default">tfm-nodejs-regenerator-runtime</packagereq>
+      <packagereq type="default">tfm-nodejs-reselect</packagereq>
       <packagereq type="default">tfm-nodejs-reselect</packagereq>
       <packagereq type="default">tfm-nodejs-sass-loader</packagereq>
       <packagereq type="default">tfm-nodejs-seamless-immutable</packagereq>
@@ -297,12 +297,14 @@
       <packagereq type="default">foreman-installer-katello</packagereq>
       <packagereq type="default">katello-certs-tools</packagereq>
       <packagereq type="default">katello-selinux</packagereq>
+      <packagereq type="default">keycloak-httpd-client-install</packagereq>	      
       <packagereq type="default">mod_passenger</packagereq>
       <packagereq type="default">pcp-mmvstatsd</packagereq>
       <packagereq type="default">puppet-agent-oauth</packagereq>
       <packagereq type="default">puppet-agent-puppet-strings</packagereq>
       <packagereq type="default">puppet-agent-rgen</packagereq>
       <packagereq type="default">puppet-agent-yard</packagereq>
+      <packagereq type="default">python2-keycloak-httpd-client-install</packagereq>
       <packagereq type="default">rubygem-bundler_ext</packagereq>
       <packagereq type="default">rubygem-clamp</packagereq>
       <packagereq type="default">rubygem-concurrent-ruby</packagereq>
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -410,6 +410,7 @@ foreman_nonscl_packages:
       build_package_tito_builder: "tito.builder.FetchBuilder"
     katello-certs-tools: {}
     katello-selinux: {}
+    keycloak-httpd-client-install: {}
     pcp-mmvstatsd: {}
     puppet-agent-oauth: {}
     puppet-agent-puppet-strings: {}
--- a/packages/foreman/keycloak-httpd-client-install/keycloak-httpd-client-install-1.2.2.tar.gz
+++ b/packages/foreman/keycloak-httpd-client-install/keycloak-httpd-client-install-1.2.2.tar.gz
@@ -1,0 +1,1 @@
+../../../.git/annex/objects/K5/JG/SHA256E-s47282--983afacfc9a1a743e2f7a3812c2738c09f58a997c5d1549790aecbb7bb23ac71.tar.gz/SHA256E-s47282--983afacfc9a1a743e2f7a3812c2738c09f58a997c5d1549790aecbb7bb23ac71.tar.gz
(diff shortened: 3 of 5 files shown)


ehelms:
I think this needs to build a python 2 package for EL7 so changing all the `3` to `2` should suffice.

upadhyeammit:
> I think this needs to build a python 2 package for EL7 so changing all the `3` to `2` should suffice.

I fixed this one. I tried to test this with tito mock build on my laptop which fails with below error,
~~~
+ /usr/bin/python2 setup.py build '--executable=/usr/bin/python2 -s'
BUILDSTDERR: /usr/bin/python2: can't open file 'setup.py': [Errno 2] No such file or directory
~~~

I can see that setup.py is there in .tar file, and I am able to build package manually,
~~~
]$ python setup.py build
running build
running build_py
creating build
creating build/lib
creating build/lib/keycloak_httpd_client
...
...
~~~

Not sure where its failing now!

comps/comps-foreman-rhel7.xml
```diff
@@ -303,6 +303,7 @@
       <packagereq type="default">puppet-agent-puppet-strings</packagereq>
       <packagereq type="default">puppet-agent-rgen</packagereq>
       <packagereq type="default">puppet-agent-yard</packagereq>
+      <packagereq type="default">python3-keycloak-httpd-client-install</packagereq>
```
evgeni:
```suggestion
      <packagereq type="default">python2-keycloak-httpd-client-install</packagereq>
```

(the generator gets that wrong I think)

packages/foreman/python-keycloak-httpd-client-install/python-keycloak-httpd-client-install.spec
```diff
@@ -0,0 +1,50 @@
...
+%install
+%py2_install
+
+%files -n python2-%{pypi_name}
+%license LICENSE.txt
+%doc README.md
+%{_bindir}/keycloak-httpd-client-install
+%{_bindir}/keycloak-rest
+%{python2_sitelib}/keycloak_httpd_client
+%{python2_sitelib}/usr
```
evgeni:
```suggestion
```

not sure why this was added by the generator, but it's wrong :)

evgeni:
@upadhyeammit I think it should build if you apply my suggestions :)

packages/foreman/python-keycloak-httpd-client-install/python-keycloak-httpd-client-install.spec
```diff
@@ -0,0 +1,50 @@
...
+Source0:        https://files.pythonhosted.org/packages/source/k/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+BuildArch:      noarch
+ 
+BuildRequires:  python2-devel
+BuildRequires:  python2-setuptools
+
+%description
+
+
+%package -n     python2-%{pypi_name}
```
evgeni: Hm. [`setup.py` defines requires](https://github.com/latchset/keycloak-httpd-client-install/blob/master/setup.py#L23) but for some reason they were not picked up here :(

additionally, the script needs six, but there is no six in setup.py: https://github.com/latchset/keycloak-httpd-client-install/blob/master/bin/keycloak-httpd-client-install#L16-L17

You will need:
```
Requires:       python-requests
Requires:       python-requests-oauthlib
Requires:       python-jinja2
Requires:       python-six
```

and then probably also `setuptools`: https://github.com/latchset/keycloak-httpd-client-install/blob/master/keycloak_httpd_client/__init__.py#L1 and `lxml`: https://github.com/latchset/keycloak-httpd-client-install/blob/master/keycloak_httpd_client/utils.py#L22

upadhyeammit:
I have downloaded the rpm and able to run command. The rpm is pulling required dependencies from repository.

evgeni:
Thanks @upadhyeammit!

From a pure "this is a working python package" perspective, this looks correct to me and should work.
However, this is a package also present in EL7 (and EL8) and as such we should be careful that the users really get our version with all its glory. Because of that, I have a few questions/suggestions:
* Will a similar package be required for EL8 (when we ship Foreman on EL8)? Currently, I see version `0.8-7.el8` in CentOS8 AppStream repository.
* The spec in [Fedora](https://src.fedoraproject.org/rpms/keycloak-httpd-client-install/blob/master/f/keycloak-httpd-client-install.spec) (which is also used in EL7 I think) currently produces multiple packages: `keycloak-httpd-client-install` which contains `/usr/share/keycloak-httpd-client-install/` and `pythonX-keycloak-httpd-client-install` which contains the library and the binaries/scripts. The user is supposed to install `keycloak-httpd-client-install` and that will pull the correct `pythonX-...` package as a requirement. Given your spec does not have `keycloak-httpd-client-install`, we might end up with mixed packages (and possibly file conflicts).
* One possibility to solve the above would be to split the package in the same way it's done in Fedora/EL.
* Another would be having a `Obsoletes: keycloak-httpd-client-install < %{version}` and `Provides:  keycloak-httpd-client-install = %{version}-%{release}` in the specfile. But IMHO following the Fedora split is cleaner.

upadhyeammit:
> Thanks @upadhyeammit!
> 
> From a pure "this is a working python package" perspective, this looks correct to me and should work.
> However, this is a package also present in EL7 (and EL8) and as such we should be careful that the users really get our version with all its glory. Because of that, I have a few questions/suggestions:
> 
>     * Will a similar package be required for EL8 (when we ship Foreman on EL8)? Currently, I see version `0.8-7.el8` in CentOS8 AppStream repository.
> 
>     * The spec in [Fedora](https://src.fedoraproject.org/rpms/keycloak-httpd-client-install/blob/master/f/keycloak-httpd-client-install.spec) (which is also used in EL7 I think) currently produces multiple packages: `keycloak-httpd-client-install` which contains `/usr/share/keycloak-httpd-client-install/` and `pythonX-keycloak-httpd-client-install` which contains the library and the binaries/scripts. The user is supposed to install `keycloak-httpd-client-install` and that will pull the correct `pythonX-...` package as a requirement. Given your spec does not have `keycloak-httpd-client-install`, we might end up with mixed packages (and possibly file conflicts).
> 
>     * One possibility to solve the above would be to split the package in the same way it's done in Fedora/EL.
> 
>     * Another would be having a `Obsoletes: keycloak-httpd-client-install < %{version}` and `Provides:  keycloak-httpd-client-install = %{version}-%{release}` in the specfile. But IMHO following the Fedora split is cleaner.

@evgeni Thanks a lot for suggestions, I have changed the spec file and now its very similar to what we have in Fedora. It generates two packages with dependencies. All looks green now :)

evgeni:
[test rpm]

evgeni:
[test rpm]

evgeni:
[test rpm]

PR #4376, opened 2019-11-20
Blacklist dnd testing npms

ezr-ondrej:
Needed for: theforeman/foreman#6762


ezr-ondrej:
Updated to keep the list sorted.

PR #4378, opened 2019-11-20
Release hammer-cli-katello 0.20.1

jturel:


jturel:
cc @evgeni

evgeni:
http://koji.katello.org/koji/taskinfo?taskID=251172

PR #4379, opened 2019-11-20
Upgrade asset gems

ezr-ondrej:
~Not entirely sure why the asset gems are packaged though... ?~ :thinking:

#5099 should go first!
theforeman/foreman#7187 depends on this.


ekohl:
> Not entirely sure why the asset gems are packaged though... ? thinking

You need them to build the application.

ezr-ondrej:
Ah right, we are building it on the client, right?
Why don't we do not package precompliled assets, though? :thinking:

ekohl:
> Ah right, we are building it on the client, right?

No, they are built during the RPM build process. However, RPM builds have no network connectivity and all build dependencies must be in the buildroot. These are typically not installed on the users system. It also allows users to take the source RPMs and rebuild Foreman.

>  Why don't we package precompliled assets, though? thinking

We do

ezr-ondrej:
Ah right, cool, thanks for the explanation! :heart:

packages/foreman/rubygem-patternfly-sass/rubygem-patternfly-sass.spec
```diff
@@ -4,8 +4,8 @@
 %global gem_name patternfly-sass
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 3.37.0
-Release: 2%{?dist}
+Version: 3.59.4
```
mmoll: This version has different dependency versions, which need get updated.

mmoll:
I think this will get related to https://github.com/theforeman/rails-packaging/pull/153 as the sassc gem will be needed and it depends on ffi.

packages/foreman/rubygem-patternfly-sass/rubygem-patternfly-sass.spec
mmoll: Specificallty on my DEB install it was:
```
#Installing jquery-ui-rails 6.0.1 (was 4.2.1)
Installing sassc 2.2.1 with native extensions
Installing bootstrap-sass 3.4.1 (was 3.3.7)
Installing patternfly-sass 3.59.4 (was 3.37.10)
```

ezr-ondrej:
> I think this will get related to [theforeman/rails-packaging#153](https://github.com/theforeman/rails-packaging/pull/153) as the sassc gem will be needed and it depends on ffi.

You are right, not only related, but we depend on it :(

mmoll:
@ezr-ondrej After looking into it again, I'm quite sure this PR here is not blocked by the rails-packaging one. We do have ffi available in foreman-packaging.

ezr-ondrej:
> @ezr-ondrej After looking into it again, I'm quite sure this PR here is not blocked by the rails-packaging one. We do have ffi available in foreman-packaging.

But not `sassc`, so we would need to add `sassc` here and than it would really make no sense to have `sass-rails` in rails-packaging, right?

mmoll:
We can add `sassc` here, but on in the `rails-packaging` repo. So there's no dependency right now.

ezr-ondrej:
> We can add `sassc` here, but not in the `rails-packaging` repo. So there's no dependency right now.

Yeah, but what for the `rails-packaging` is than? :thinking:

mmoll:
fails :broken_heart:

ezr-ondrej:
Any idea why? I don't know what the obal lint is complaining about. :(

Dne st 27. 11. 2019 11:25 uživatel Michael Moll <notifications@github.com>
napsal:

comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -227,6 +226,7 @@
       <packagereq type="default">tfm-rubygem-down-doc</packagereq>
       <packagereq type="default">tfm-rubygem-elastic-apm-doc</packagereq>
       <packagereq type="default">tfm-rubygem-faraday-cookie_jar-doc</packagereq>
+      <packagereq type="default">tfm-rubygem-ffi-doc</packagereq>
```
mmoll: What's going on here? :upside_down_face:

comps/comps-foreman-rhel7.xml
```diff
@@ -121,9 +121,9 @@
       <packagereq type="default">tfm-nodejs-redux</packagereq>
       <packagereq type="default">tfm-nodejs-redux-thunk</packagereq>
       <packagereq type="default">tfm-nodejs-regenerator-runtime</packagereq>
-      <packagereq type="default">tfm-nodejs-reselect</packagereq>
       <packagereq type="default">tfm-nodejs-regenerator-runtime</packagereq>
       <packagereq type="default">tfm-nodejs-reselect</packagereq>
+      <packagereq type="default">tfm-nodejs-reselect</packagereq>
```
mmoll: double line?

mmoll:
I think the XMLs are wrong, but it's only a hunch...

mmoll:
https://github.com/theforeman/foreman-packaging/pull/4390#issuecomment-558042711 - maybe just moving the result will fix it, but no idea... :man_shrugging:

ezr-ondrej:
This is trying to do too much at once :)
I'll go ahead and upgrade to `sassc-rails` first: #4661

zjhuntin:
[test rpm]

tbrisker:
repoclosure is expected as foreman-assets still requires `tfm-rubygem(sass-rails) < 0:6.0` and needs to be updated

ezr-ondrej:
> Could this PR get rebased?

I've rebased, but we should merge #5099 first to knowingly switch to `sassc`!

package_manifest.yaml
```diff
@@ -113,6 +113,7 @@ foreman_scl_packages:
       releasers:
         - koji-foreman
         - koji-foreman-plugins
+    rubygem-ffi: {}
```
ekohl, line 116: We already have FFI, this shouldn't be needed.

ezr-ondrej: I've checked and it's not in manifest. So it doesn't need to be in `package_manifest.yaml`? or will it get refreshed automatically at some point?

ekohl: It's under the foreman proxy packages, but I think that's wrong:
https://github.com/theforeman/foreman-packaging/blob/d6c2dcb38d6ae0aad3ecbe3b1c8261eaed58b30a/package_manifest.yaml#L599

@ehelms you changed this in 3c6439a288d21e93f02efb7481048a4d19fe52fd but `rubygem-sassc` also depends on it for example.

ezr-ondrej: right, I've removed it from there:
https://github.com/theforeman/foreman-packaging/pull/4379/files#diff-9297bc45c5e9bdf810aa55f6c91853dcL588
looks better now?

ezr-ondrej:
[test rpm]

package_manifest.yaml
ekohl: I think so. @ehelms?

ehelms: Yep, looks good.

zjhuntin:
[test rpm]

zjhuntin:
[test rpm]

zjhuntin:
Thanks @ezr-ondrej

PR #4388, opened 2019-11-22
Add foreman_acd

sbernhard:


packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
```diff
@@ -0,0 +1,181 @@
...
+# start package.json dependencies BuildRequires
+BuildRequires: %{?scl_prefix}npm(c3) >= 0.4.11
+BuildRequires: %{?scl_prefix}npm(c3) < 1.0.0
+BuildRequires: %{?scl_prefix}npm(humanize-duration) >= 3.20.1
+BuildRequires: %{?scl_prefix}npm(humanize-duration) < 4.0.0
+BuildRequires: %{?scl_prefix}npm(jquery) >= 2.2.4
+BuildRequires: %{?scl_prefix}npm(jquery) < 2.3.0
+BuildRequires: %{?scl_prefix}npm(react-intl) >= 2.8.0
+BuildRequires: %{?scl_prefix}npm(react-intl) < 3.0.0
+# end package.json dependencies BuildRequires
```
ehelms: Not that this blocks, but please consider moving to use of the foreman vendor JS.

sbernhard: Thanks @ehelms for your remark. I had a look at other plugins and no-one is currently using it (unfortunately). 
Should I then ONLY use the foreman-vendor package and remove all other npm packages or should I add foreman-vendor and still use the npm packages?

ekohl: Plugins like `foreman_ansible` are using it. They also have additional dependencies. Probably best to ask the UI team. I'm sure @sharvit will happily guide you in the right direction.

sharvit: The build process will replace all the modules you are using with the modules that exist in the vendor anyway. (if they exist in the vendor)

It's a good practice to install the vendor as a dep and clean up modules from your package.json so your plugin declares what vendor version it needs. It will help with upgrading once there are breaking changes to the vendor since you will know about them in advance.

sbernhard:
Thanks @sharvit. Means, remove all dependencies and replace them with just and only foreman-vendor?

sharvit:
> Thanks @sharvit. Means, remove all dependencies and replace them with just and only foreman-vendor?

Only if they exist in the vendor, see: https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/package.json#L27

sbernhard:
Thanks. BTW, jQuery is really old and has some security issues (says GitHub)

sharvit:
> Thanks. BTW, jQuery is really old and has some security issues (says GitHub)

Yes, actually we want to be in a point where we can drop `jquery` entierly.
See:
- https://github.com/theforeman/foreman-js/blob/master/roadmap.md#theforemanvendor-core
- https://community.theforeman.org/t/legacy-js-things-we-might-want-to-remove-replace-at-some-point/16195

sharvit:
> > Thanks. BTW, jQuery is really old and has some security issues (says GitHub)
> 
> Yes, actually we want to be in a point where we can drop `jquery` entierly. (I would recommend avoiding using `jquery` if possible)
> See:
> 
> * https://github.com/theforeman/foreman-js/blob/master/roadmap.md#theforemanvendor-core
> * https://community.theforeman.org/t/legacy-js-things-we-might-want-to-remove-replace-at-some-point/16195

sbernhard:
Thanks @sharvit for your explanation. I will have a look how to get rid of jquery (soon).

Currently, the build RPM tasks failes because of:
http://koji.katello.org/koji/getfile?taskID=252074&volume=DEFAULT&name=build.log&offset=-4000

Is it related to different version of lodash in https://github.com/theforeman/foreman-js/blob/f0e3fe1d27255f77e2a6303185ea3e578c9845a4/packages/vendor-core/package.json#L50 and https://github.com/theforeman/foreman-packaging/blob/f779889e127326f1061d28cf29093c42b5cb7f7d/packages/foreman/nodejs-lodash/nodejs-lodash.spec#L7 ?

sharvit:
It might be that but I don't think it's related since `cloneDeep` exists in lodash since v1

sbernhard:
hmm. Do you have another idea whats wrong? Previously. without using foreman-vendor-js the build was successfully (used lodash 4.17.11).

sharvit:
> hmm. Do you have another idea whats wrong? Previously. without using foreman-vendor-js the build was successfully (used lodash 4.17.11).

- Do you have the same issue when building locally against @theforeman/vendor (v1.4)?
- Do you have the same issue when building locally against `lodash` v4.17.11?

sbernhard:
Package build was successfully after:
- adding lodash 4.17.11
- adding patternfly (because of some necesary variables)
- adding patternfly-react (because of _inline_edit variables)

sharvit:
> Package build was successfully after:
> 
> * adding lodash 4.17.11
Very odd

> * adding patternfly (because of some necesary variables)
> * adding patternfly-react (because of _inline_edit variables)


You shouldn't import patternfly in your `scss` files to get their variables, instead you should import them from the `vendor`.
See: https://github.com/theforeman/foreman-js/tree/master/packages/vendor#stylesheets

sbernhard:
I can the _variables -> foreman-js/variables import but I can not find the
_loading-state and the _inline_edit. Are they not included in foreman-js? Do I need to use them from patternfly and patternfly-react?

sharvit:
You should not load them internally, all the pf partials should exist in https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/scss/vendor-core.scss

If they are not, we should create a PR and add them to the `vendor-core.scss` file.
The `vendor-core.scss` is the opinionated 3rd-party css bundle that we deliver to plugins.
Other than that you can also use the 3rd-party variables and mixins from: https://github.com/theforeman/foreman-js/tree/master/packages/vendor#stylesheets

sbernhard:
Removed patternfly-(react) dependencies. Unfortunately, lodash must still be a dependency - otherwise it doesn't build.

ehelms:
[test rpm]

zjhuntin:
[test rpm]

ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
@sbernhard Needs updates to latest dependencies

PR #4390, opened 2019-11-22
add rubygem-sassc

mmoll:


mmoll:
somehow fails :man_shrugging:

evgeni:
For some reason, the `.so` file gets installed as `/usr/share/gems/gems/sassc-2.2.1/ext/libsass.so`, but the RPM (correctly) expects it to be `/usr/lib64/gems/ruby/sassc-2.2.1/ext/libsass.so` and then stuff fails.

mmoll:
Leaving this to the RPM experts...

PR #4396, opened 2019-11-25
Drop packaging for stretch and xenial

tbrisker:


mmoll:
merged, thanks @tbrisker!

PR #4409, opened 2019-11-26
Update rubygem-foreman_discovery to 16.0.1

lzap:

* [x] Nightly
* [x] 1.24

Hello, heed the stable as well, blocker bug for discovery.


packages/plugins/rubygem-foreman_discovery/rubygem-foreman_discovery.spec
```diff
@@ -17,7 +17,7 @@
 
 Summary:    MaaS Discovery Plugin for Foreman
 Name:       %{?scl_prefix}rubygem-%{gem_name}
-Version:    %{mainver}
+Version:    16.0.1
```
evgeni: can you please update `mainver` above instead of setting the version here?

(or remove all that mainver handling, but that's a different discussion I guess)

evgeni:
thanks @lzap!

please open a cherrypick PR for 1.24

lzap:
https://github.com/theforeman/foreman-packaging/pull/4434

PR #4413, opened 2019-11-27
Add nodejs-theforeman-builder package

evgeni:

PR #4416, opened 2019-11-27
Bump nodejs-babel-loader to 8.0.6

evgeni:


evgeni:
The following PRs need to be merged "together": #4417 #4416 #4413

However, they are blocked on the discussion in https://github.com/theforeman/foreman/pull/7178

PR #4417, opened 2019-11-27
Bump nodejs-theforeman-vendor to 3.3.2

evgeni:


evgeni:
The following PRs need to be merged "together": #4417 #4416 #4413

However, they are blocked on the discussion in https://github.com/theforeman/foreman/pull/7178

packages/foreman/nodejs-theforeman-vendor/nodejs-theforeman-vendor.spec
```diff
@@ -4,16 +4,14 @@
...
+Release: 1%{?dist}
 Summary: foreman supported 3rd-party node_modules
 License: MIT
 Group: Development/Libraries
 URL: https://github.com/theforeman/foreman-js#readme
 Source0: https://registry.npmjs.org/%{npm_name}/-/%{npm_name}-%{version}.tgz
-%if 0%{?scl:1}
-BuildRequires: %{?scl_prefix_nodejs}npm
-%else
+%if 0%{?!scl:1}
```
evgeni: Meh, something is wrong with my nom2rpm..

ekohl: I kind of assumed that was intended since non-bundled packages don't really call NPM. Or is it needed for macros?

evgeni: Well, I don't know. It wasn't intended, so I was going with whatever was there before

evgeni: I still backed out that change, will revisit later.

PR #4420, opened 2019-11-27

ekohl:


packages/foreman/foreman/foreman.spec
```diff
@@ -450,8 +450,9 @@ Meta package to install requirements for Rackspace compute resource support.
 Summary: Foreman VMware support
 Group:  Applications/System
 # start specfile vmware Requires
-Requires: %{?scl_prefix}rubygem(fog-vsphere) >= 3.2.1
```
ekohl: https://github.com/theforeman/foreman/pull/7213 should clear this up. I'll fix it manually now.

packages/foreman/foreman/foreman.spec
```diff
@@ -696,9 +696,8 @@ Meta Package to install requirements for Redis caching support
 Summary: Foreman Redis caching support
 Group:  Applications/System
 # start specfile redis Requires
-Requires: %{?scl_prefix}rubygem(sidekiq) >= 5.0
-Requires: %{?scl_prefix}rubygem(sidekiq) < 6
-Requires: %{?scl_prefix}rubygem(gitlab-sidekiq-fetcher)
+Requires: %{?scl_prefix}rubygem(redis) >= 4.0
+Requires: %{?scl_prefix}rubygem(redis) < 5.0
 # end specfile redis Requires
```
tbrisker: This seems odd for dynflow-sidekiq package? perhaps a mix with the redis package above?

ekohl: This comes from https://github.com/theforeman/foreman/blob/develop/bundler.d/redis.rb. Not to be confused with https://github.com/theforeman/foreman/blob/develop/bundler.d/dynflow_sidekiq.rb

ekohl: Now I see what you mean. 15b1329b7267c2312071c86ae107a58a74053e20 was indeed not entirely correct and is causing this problem.

evgeni:
#4435 has been merged

PR #4426, opened 2019-11-29
Update rubygem-foreman_remote_execution to 2.0.4

adamruzicka:


evgeni:
thanks @adamruzicka!

please open a cherrypick PR for 1.24 if that should go there too

PR #4432, opened 2019-12-02
Use package names, not provides in Obsoletes of the f_openscap package

evgeni:
rpm 4.15 does not allow anything else but package names in obsoletes
anymore


evgeni:
this also drops old `ruby193` Obsoletes, because :woman_shrugging:

PR #4433, opened 2019-12-02
Bump rubygem-pg to 1.1.4-1

evgeni:


evgeni:
our CI was running with this version since ages, so why not also our users? :)

packages/foreman/rubygem-pg/rubygem-pg.spec
```diff
@@ -1,101 +1,103 @@
...
-# Generated from pg-0.11.0.gem by gem2rpm -*- rpm-spec -*-
 %global gem_name pg
 
-Summary: A Ruby interface to the PostgreSQL RDBMS
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.21.0
-Release: 3%{?dist}
+Version: 1.1.4
+Release: 1%{?dist}
+Summary: Pg is the Ruby interface to the {PostgreSQL RDBMS}[http://www.postgresql.org/]
```
ekohl:
```suggestion
Summary: Pg is the Ruby interface to the PostgreSQL RDBMS
```

packages/foreman/rubygem-pg/rubygem-pg.spec
```diff
@@ -1,101 +1,103 @@
...
-Obsoletes: %{?scl_prefix}ruby(postgres) <= 0.7.9-2010.01.28.2
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
-%{?scl:Obsoletes: ruby193-rubygem-%{gem_name}}
+# end specfile generated dependencies
+BuildRequires: postgresql-devel
 
 %description
-This is the extension library to access a PostgreSQL database from Ruby.
-This library works with PostgreSQL 7.4 and later.
+Pg is the Ruby interface to the {PostgreSQL
```
ekohl: Probably good to keep the old description

evgeni: no (as it would be lying about the PostgreSQL version), but I made the new description as close to the old as possible ;)

PR #4421, opened 2019-11-27
Update katello package dependencies

ekohl:

PR #4436, opened 2019-12-02
Handle rename of pulp_docker_client

jlsherrill:
to pulp_container_client


jlsherrill:
omitting the katello.spec changes, so it can be handled by https://github.com/theforeman/foreman-packaging/pull/4421

packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
```diff
@@ -86,9 +81,9 @@ rm %{buildroot}/%{gem_instdir}/Gemfile.lock
...
 %doc %{gem_instdir}/docs
-%doc %{gem_instdir}/pulp_docker_client.gemspec
+%{gem_instdir}/pulp_container_client.gemspec
 %{gem_instdir}/spec
 
 %changelog
-* Fri Jul 5 2019 Justin Sherrill <jlsherrill@gmail.com> 4.0.0b6.dev01562331743-1
-- initial build
+* Mon Dec 02 2019 Justin Sherrill <jlsherrill@gmail.com> 1.0.0rc1-1
+Initial build
```
ekohl: Is this correct?

packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
```diff
@@ -1,26 +1,23 @@
...
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 4.0.0b6.dev01562331743
+Version: 1.0.0rc1
 Release: 1%{?dist}
-Summary: Pulp 3 Docker API Ruby Gem
+Summary: Pulp container plugin for Pulp3 API Ruby Gem
 Group: Development/Languages
-License: GPLv2
-URL: https://rubygems.org/gems/pulp_docker_client
+License: GPL-2.0
```
ekohl: Looks like the license is missing in https://github.com/theforeman/foreman-packaging/blob/rpm/develop/licenses.json

```suggestion
License: GPLv2
```

packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
jlsherrill: ah, the changelog script generated this part!  i will fix

licenses.json
```diff
@@ -1,4 +1,5 @@
 {
 	"GPL-3.0": "GPLv3",
-	"GPL-3.0+": "GPLv3+"
+	"GPL-3.0+": "GPLv3+",
+	"GPL-2.0+": "GPLV2+"
```
ekohl: Mind keeping it sorted?

jlsherrill:
any idea why the build is failing?
```
BUILDSTDERR:     ".gem\n" is not an octal string
```

packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
```diff
@@ -1,26 +1,25 @@
...
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 4.0.0b6.dev01562331743
+Version: 1.0.0rc1
 Release: 1%{?dist}
-Summary: Pulp 3 Docker API Ruby Gem
+Summary: Pulp container plugin for Pulp3 API Ruby Gem
 Group: Development/Languages
-License: GPLv2
-URL: https://rubygems.org/gems/pulp_docker_client
+License: GPL-2.0+
```
ekohl: I think the RPM version for that is different
```suggestion
License: GPLv2+
```

licenses.json
ekohl: And lowercase `v`

packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
```diff
@@ -1,26 +1,25 @@
...
 # template: scl
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
-%global gem_name pulp_docker_client
+%global gem_name pulp_container_client
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 4.0.0b6.dev01562331743
+Version: 1.0.0rc1
```
ekohl: I think `rc1` here should be in `%prerelease`. Pretty sure this is making it fail in some way.

packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
```diff
@@ -1,26 +1,29 @@
...
+%global release 1
+%global prerelease .rc1
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 4.0.0b6.dev01562331743
-Release: 1%{?dist}
-Summary: Pulp 3 Docker API Ruby Gem
+Version: 1.0.0
+
+Release %{?prerelease:0.}%{release}%{?prerelease}%{?dist}
```
ekohl:
```suggestion
Release: %{?prerelease:0.}%{release}%{?prerelease}%{?dist}
```

jlsherrill:
obsoletes added and now :green_heart:

packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
```diff
@@ -1,26 +1,30 @@
...
 # template: scl
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
-%global gem_name pulp_docker_client
+%global gem_name pulp_container_client
+
+%global release 1
+%global prerelease rc1
+%global prereleaserpm %{?prerelease:.}%{?prerelease}
```
jlsherrill: note in other places i see we use prereleasesource and prerelease, but the macro that defines gem_instdir uses prerelease, so that model won't work

ekohl: We use this in the katello gem:
https://github.com/theforeman/foreman-packaging/blob/6f54cca5530a9ad10bc545aa86ff51f05aadcada/packages/katello/rubygem-katello/rubygem-katello.spec#L8-L9

ekohl: @evgeni what do you think about including this by default in our gem2rpm templates?

jlsherrill: that didn't work here because the source was '3.0.0rc1' and NOT '3.0.0.rc1'

ehelms: Do we have control over that? Gem has built-in support for prerelease when it finds a dash in the version (https://github.com/Katello/katello/blob/master/lib/katello/version.rb#L2) which it translates to a `.` and then some builtin mechanisms handle this properly in the RPMs which is how Katello and hammer gems work.

ehelms: Also, I am good with solving this for now this way. But if we can improve this for the future for consistency with our other gems that would be nice.

evgeni: Eric said everything I wanted to say, but I'll repeat, just for the sake of my own understanding:
* ruby gems (usually) use `.rc1` or similar suffixes to denote a pre-release version, even tho `rc1` (without the dot) would be sufficient (see https://guides.rubygems.org/patterns/#prerelease-gems, any character triggers "prerelease" mode)
* the generated gems use a more pythonic-aproach (https://www.python.org/dev/peps/pep-0440/#pre-releases) without the dot
* the rubygems rpm macros (which we don't control) allow to use the version without the dot too.
* it would be awesome if we could use the same template for every gem, which would require the pulp gems to switch to include the dot
* the current state is perfectly fine to go with for now

jlsherrill: @evgeni @ekohl sooooo ACK?   :)

packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
```diff
@@ -1,26 +1,30 @@
...
+Summary: Pulp container plugin for Pulp3 API Ruby Gem
 Group: Development/Languages
-License: GPLv2
-URL: https://rubygems.org/gems/pulp_docker_client
-Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+License: GPLv2+
+URL: https://github.com/pulp/pulp_container/
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}%{?prerelease}.gem
+
+Obsoletes: %{?scl_prefix_ruby}rubygem(pulp_docker_client)
```
evgeni: Can you please use rubygem-name notation, not rubygem(name)? The latter freaks out rpm 4.15 (fedora 31)

jlsherrill: you mean just on this obsoletes line?  Or everywhere throughout the file?

evgeni: Just obsoletes

jlsherrill:
@evgeni this should be good, failure is on repoclosure for unrelated reasons

evgeni:
Yeah, repoclosure is expected until we fix katello to use the new vendor v3.

PR #4442, opened 2019-12-04
Fixes #28421 - drop dynflowd service

ezr-ondrej:
Drops the dynflowd service with optional replacement - `dynflow-sidekiq`
Kind of harsh change, but I believe if we do the installer part and notify users in upgrade notes, it should be ok.


ekohl:
Looks good, but blocked by https://github.com/theforeman/puppet-foreman/pull/761

ekohl:
Rebased on develop. Should be ready to merge now if it's green.

ekohl:
Thanks!

PR #4445, opened 2019-12-05
Update katello-client-bootstrap to 1.7.4

evgeni:


evgeni:
repoclosure is unrelated

PR #4461, opened 2019-12-07
Bump nodejs-theforeman-vendor to 3.5.2

ekohl:


ekohl:
I really dislike the huge 4 MB binary. Not sure if we can solve that in another way.

mmoll:
Does patternfly-react also need to get an update?

https://github.com/theforeman/foreman-js/commit/38cb968961edc0c4d0c6119a91c046215cdf2c29

sharvit:
> Does patternfly-react also need to get an update?
> 
> [theforeman/foreman-js@38cb968](https://github.com/theforeman/foreman-js/commit/38cb968961edc0c4d0c6119a91c046215cdf2c29)

Already bundled in the vendor.



> I really dislike the huge 4 MB binary. Not sure if we can solve that in another way.

Since it contains 2 compiled versions of the vendor (dev, prod)

ekohl:
> > I really dislike the huge 4 MB binary. Not sure if we can solve that in another way.
> 
> Since it contains 2 compiled versions of the vendor (dev, prod)

It's 4 MB of gzipped JSON API responses from npmjs.com. It's 13 MB unpacked. AFAIK these don't include the sources but I didn't look too closely.

PR #4466, opened 2019-12-10
Update foreman_expire_hosts to 7.0.0

timogoebel:


dgoetz:
Is it by intention that the cronjob is removed? Also I think the old description was better.

timogoebel:
> Also I think the old description was better.

That's something we should change in the gemspec.

timogoebel:
Re-added the cronjob.

PR #4468, opened 2019-12-10
Add recognition of cronjobs to gem2rpm templates

dgoetz:
I think this would be a useful addition to the templates, so we hopefully do not lose a cronjob if someone regenerates a spec. 

Search for them in a subdirectory `extra` from the gem with pattern `*cron*` was my best idea and it works at least for one cronjob, but if someone has a better idea, I am more than happy to change the logic.

### DIFF
--- a/gem2rpm/foreman_plugin.spec.erb
+++ b/gem2rpm/foreman_plugin.spec.erb
@@ -11,6 +11,7 @@ LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
 -%>
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
+%{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
 
 %global gem_name <%= spec.name %>
 %global plugin_name <%= spec.name.sub(/\Aforeman_/, '') %>
@@ -18,6 +19,7 @@ LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
 <%
 has_assets = spec.files.any? { |f| f.start_with?('app/assets/') }
 has_apidoc = spec.files.any? { |f| f =~ %r{\Aapp/controllers/?.*/(api|concerns)/} }
+has_cronjob = spec.files.any? { |f| f =~ %r{\Aextra/.*\.cron\z} }
 has_db_migrations = spec.files.any? { |f| f.start_with?('db/migrate/') }
 has_db_seeds = spec.files.any? { |f| f.start_with?('db/seeds.d/') }
 has_webpack = spec.files.any? { |f| f.start_with?('webpack/') }
@@ -141,6 +143,10 @@ mkdir -p %{buildroot}%{gem_dir}
 cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
+<% if has_cronjob -%>
+mkdir -p %{buildroot}%{_root_sysconfdir}/cron.d/
+mv %{buildroot}%{gem_instdir}/extra/*.cron %{buildroot}%{_root_sysconfdir}/cron.d/%{gem_name}
+<% end -%>
 <% unless spec.extensions.empty? -%>
 mkdir -p %{buildroot}%{gem_extdir_mri}
 cp -a .%{gem_extdir_mri}/{gem.build_complete,*.so} %{buildroot}%{gem_extdir_mri}/
@@ -188,6 +194,9 @@ find %{buildroot}%{gem_instdir}/<%= spec.bindir %> -type f | xargs chmod a+x
 <% end -%>
 <% if has_assets -%>
 %{foreman_assets_plugin}
+<% end -%>
+<% if has_cronjob -%>
+%config(noreplace) %{_root_sysconfdir}/cron.d/%{gem_name}
 <% end -%>
 <% if has_webpack -%>
 %{foreman_webpack_plugin}
--- a/gem2rpm/smart_proxy_plugin.spec.erb
+++ b/gem2rpm/smart_proxy_plugin.spec.erb
@@ -4,6 +4,7 @@ require 'json'
 root = `git rev-parse --show-toplevel`.strip
 SCL_PREFIXES = JSON.load(File.read(File.join(root, 'scl_prefixes.json')))
 LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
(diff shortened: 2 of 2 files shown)


gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -133,6 +136,10 @@ mkdir -p %{buildroot}%{foreman_proxy_settingsd_dir}
 <% config_file = config_file_path.split('/').last -%>
 mv %{buildroot}%{gem_instdir}/<%= config_file_path %> \
    %{buildroot}%{foreman_proxy_settingsd_dir}/<%= config_file.chomp('.example') %>
+<% if has_cronjob -%>
+mkdir -p %{buildroot}%{?scl:%_root_sysconfdir}%{!?scl:%_sysconfdir}/cron.d/
```
ekohl: On line 13 we ensure the `%_root_sysconfdir` macro is defined. Same below
```suggestion
mkdir -p %{buildroot}%{%_root_sysconfdir}/cron.d/
```

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -21,6 +21,9 @@ LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
 %global smart_proxy_dynflow_bundlerd_dir %{!?scl:/opt/theforeman/tfm/root}%{_datadir}/smart_proxy_dynflow_core/bundler.d
+<%
```
ekohl: Maybe move this to line 7?

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -18,6 +18,7 @@ LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
 <%
 has_assets = spec.files.any? { |f| f.start_with?('app/assets/') }
 has_apidoc = spec.files.any? { |f| f =~ %r{\Aapp/controllers/?.*/api/} }
+has_cronjob = spec.files.any? { |f| f =~ %r{\Aextra/.*cron.*} }
```
ekohl: Should we enforce the `.cron` extension? When does it end with something else?

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -141,6 +142,10 @@ mkdir -p %{buildroot}%{gem_dir}
 cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
+<% if has_cronjob -%>
+mkdir -p %{buildroot}%{?scl:%_root_sysconfdir}%{!?scl:%_sysconfdir}/cron.d/
+mv %{buildroot}%{gem_instdir}/extra/*cron* %{buildroot}%{?scl:%_root_sysconfdir}%{!?scl:%_sysconfdir}/cron.d/%{gem_name}
```
ekohl: I think this breaks when there's multiple cron files. Probably a bad idea regardless and maybe something we don't want to support.

gem2rpm/smart_proxy_plugin.spec.erb
dgoetz: I changed the other template too so both have the same logic.

gem2rpm/smart_proxy_plugin.spec.erb
dgoetz: Moved it to line 7.

gem2rpm/foreman_plugin.spec.erb
dgoetz: Not sure about, as I found `.cron` and `.cron.d`. I can also think someone used `.crontab` and `.cronjob`. But I can restrict it to this list and this should not be to restrictive. What do you think?

gem2rpm/foreman_plugin.spec.erb
dgoetz: Yes, it will break. Any ideas for a safeguard? Or is simply "out of support" ok?

gem2rpm/foreman_plugin.spec.erb
ekohl: I'd prefer to standardize on one since we can easily influence/control the sources. I don't have a lot of examples checked out, but a quick check shows `.cron` is used at least in `smart_proxy_omaha` and `smart_proxy_openscap`.

gem2rpm/foreman_plugin.spec.erb
ekohl: Given how few of these we have it's probably safe enough to let it fail and let the packager deal with the build failure.

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -163,6 +169,7 @@ find %{buildroot}%{gem_instdir}/<%= spec.bindir %> -type f | xargs chmod a+x
 %foreman_precompile_plugin <%= precompile_flags.strip %>
 <% end -%>
 
+
```
ekohl: I'm slightly hestitant to changing the whitespace in all templates.

gem2rpm/foreman_plugin.spec.erb
dgoetz: I updated it to only use `.cron`.

gem2rpm/foreman_plugin.spec.erb
dgoetz: Sorry, my bad. It is gone again.

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -189,11 +195,15 @@ find %{buildroot}%{gem_instdir}/<%= spec.bindir %> -type f | xargs chmod a+x
...
 <% end -%>
+<% if has_cronjob -%>
+%config(noreplace) %{_root_sysconfdir}/cron.d/%{gem_name}
+<% end -%>
 <% if has_webpack -%>
 %{foreman_webpack_plugin}
 %{foreman_webpack_foreman}
 <% end -%>
 
+
```
ekohl: Same here:

```suggestion
```

ehelms:
@dgoetz can you rebase this and then I'll merge

dgoetz:
@ehelms I rebased and pushed it, so ready to merge.

ekohl:
Thanks!

PR #4478, opened 2019-12-11
only update if the version differs

evgeni:

### DIFF
--- a/add_gem_package.sh
+++ b/add_gem_package.sh
@@ -116,6 +116,8 @@ TITO_TAG=$3
 PACKAGE_SUBDIR=$4
 ROOT=$(git rev-parse --show-toplevel)
 
+REWRITE_ON_SAME_VERSION=${REWRITE_ON_SAME_VERSION:-true}
+
 if [[ -z $TEMPLATE_NAME ]] ; then
 	if [[ $GEM_NAME == smart_proxy_*_core ]] ; then
 		TEMPLATE_NAME="scl"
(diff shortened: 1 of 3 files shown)


ekohl:
In other scripts I've used this code path to regenerate the spec file. [bump_rpm.sh](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/bump_rpm.sh) is then for just bumping. That could learn some code to get the latest version on pypi though.

evgeni:
I can add pypi support to `bump_rpm`, but my use-case is slightly different and the current `bump_rpm` design doesn't match the (intended) workflow. Maybe you have an idea.

I have a list of Python packages (with their PyPI names) and versions, as PyPI names don't map 1:1 to RPM names, I can't just call `bump_rpm **/python-%{name} %{version}`. We could use something like https://github.com/theforeman/foreman-packaging/blob/9ce817b06b8a841c14e67e66eb67e9f57a3ae06a/add_pypi_package.sh#L11-L12

But then I still need to teach `bump_rpm` to get the pypi name from the spec file and re-execute pyp2rpm with the right params to update dependencies/file lists etc.

ekohl:
I'd like the `add_*_package.sh` to all behave the same way. How about an env var defaulting to true: `REWRITE_ON_SAME_VERSION=${REWRITE_ON_SAME_VERSION:-1}`

evgeni:
@ekohl hah, it's nice how the simplest solutions don't come to ones mind, thanks, implemented.

add_gem_package.sh
```diff
@@ -174,13 +176,20 @@ if [[ ! -e $TEMPLATE ]] ; then
...
+    git commit -m "Bump $PACKAGE_NAME to $VERSION"
+  else
+    echo "$PACKAGE_NAME is already at version $VERSION"
+  fi
 else
-	add_to_manifest
-	add_to_tito_props
-	add_gem_to_comps
-	git commit -m "Add $PACKAGE_NAME package"
+  generate_gem_package
```
ekohl: The rest of the file uses tabs for indenting. Consistency would be lovely :)

evgeni: You want me to convert the whole file to spaces? ;-)

PR #4491, opened 2019-12-12
Update rubygem-foreman_ansible to 4.0.3

ares:


ares:
```
BUILDSTDERR: ModuleNotFoundError: Module not found: Error: Can't resolve 'react-json-tree' in '/builddir/build/BUILDROOT/tfm-rubygem-foreman_ansible-4.0.3-1.fm1_24.el7.noarch/opt/theforeman/tfm/root/usr/share/gems/gems/foreman_ansible-4.0.3/webpack/components'
```

I have no idea what that means but the `react-json-tree` is a dependency of foreman_ansible and I don't recall a change in this in recent history.

ekohl:
So 783652406abdfb49b615a5acbbd9eb2af545da34 is missing in 1.24 and causing this issue.

evgeni:
[test rpm]

evgeni:
eh, the test would not help, needs re-gen of the PR

ares:
> You're reverting the SCL changes. It may still appear to work because the packages are still in koji, but they're not usable by webpack. I don't know how you generated this and why it broke.

as always, using bump script, perhaps I need bump script from rpm/develop? 1.24 ends up with spec like this

evgeni:
@ares also when you use `rpm/1.24` *after* https://github.com/theforeman/foreman-packaging/pull/4495 was merged?

ares:
I was missing the last commit https://github.com/theforeman/foreman-packaging/commit/246b839e346d0ca7a618b81ccc077228855ba4c4

PR #4500, opened 2019-12-13
rename pulpcore tag

evgeni:


evgeni:
seems @pcreech updated koji, but not the configs ;)

PR #4503, opened 2019-12-13
update pulpcore and dependencies

evgeni:


evgeni:
I have expected this to fail repoclosure, but differently, investigating

evgeni:
[test rpm]

evgeni:
contrary to what the repoclosure says, this would break pulpcore :/

ehelms:
So repoclosure passing is a false positive? What should be breaking?

evgeni:
@ehelms the pulpcore version in the repo does not allow pyyaml 5.2 (3.0 GA does, that's why it's here), but repoclosure does not honour conflicts :(

PR #4506, opened 2019-12-13
Bump nodejs-theforeman-vendor to 3.8.0

jeremylenz:
I can't get `tito` to run, so this will need testing.


packages/foreman/nodejs-theforeman-vendor/nodejs-theforeman-vendor.spec
```diff
@@ -39,6 +39,9 @@ cp -pfr scss %{buildroot}%{nodejs_sitelib}/%{npm_name}
 %doc readme.md
 
 %changelog
+* Fri Dec 13 2019 vagrant 3.8.0-1
```
xprazak2: You are running inside vagrant box, aren't you? :wink:

Could you add your name and email? See the log entries below.

xprazak2:
I see we have [theforeman-builder](https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/foreman/nodejs-theforeman-builder) in packaging, you will need to bump the version for it as well.

laviro:
not sure if I did it correct, opened a PR locally to my branch https://github.com/laviro/foreman-packaging/pull/1 maybe it can help you

jeremylenz:
Not sure why the test is failing, maybe @laviro your PR will work better?

comps/comps-foreman-rhel7.xml
```diff
@@ -298,8 +298,9 @@
       <packagereq type="default">foreman-installer-katello</packagereq>
       <packagereq type="default">katello-certs-tools</packagereq>
       <packagereq type="default">katello-selinux</packagereq>
-      <packagereq type="default">keycloak-httpd-client-install</packagereq>	      
+      <packagereq type="default">keycloak-httpd-client-install</packagereq>
       <packagereq type="default">mod_passenger</packagereq>
+      <packagereq type="default">nodejs-theforeman-env</packagereq>
```
ekohl: This is unexpected. It looks like this added it in the non-scl style while nowadays it needs to be in SCL. Perhaps you based it off an old branch? It is the reason for the failure though.

jeremylenz: @ekohl The first time I attempted to add `@theforeman/env`, it errored near the end because it couldn't find a Python module called `click`.  I installed `pip3` and the `click` module, did a `git stash` to return to the previous state before adding `env`, and then tried again.  Perhaps it was something to do with that flow..

jeremylenz: built again, now failing differently

evgeni:
replacing with https://github.com/theforeman/foreman-packaging/pull/4510

PR #4509, opened 2019-12-17
Bump/nodejs theforeman packages

laviro:
try to continue #4506 work, and see if tests pass


evgeni:
replacing with https://github.com/theforeman/foreman-packaging/pull/4510

PR #4510, opened 2019-12-17
update vendor and builder to 3.9.0

evgeni:


evgeni:
@jeremylenz @laviro this now built fine, but I didn't test it further. also I didn't include `env` as I still think it should not be needed.

sharvit:
Thanks @evgeni,

> also I didn't include `env` as I still think it should not be needed.

The `env` is not needed and should not get packaged.

@laviro `3.9.0` should be packaged instead?

laviro:
thanks @evgeni !

> @laviro 3.9.0 should be packaged instead?

yes, we should upgrade to `v3.9.0`, 
tested it in foreman: https://github.com/theforeman/foreman/pull/7266
and I also found no regression in the main plugins while using it on forklift luna

evgeni:
updated to 3.9.0

packages/foreman/nodejs-theforeman-builder/nodejs-theforeman-builder.spec
```diff
@@ -346,6 +346,12 @@ rm -rf %{buildroot} %{npm_cache_dir}
 %doc node_modules/%{npm_name}/README.md
 
 %changelog
+* Tue Dec 17 2019 Evgeni Golov 3.9.0-1
+- Update to 3.9.0
+
+* Tue Dec 17 2019 Evgeni Golov 3.8.1-1
```
ekohl: Nit: does this really make sense to keep in?

sharvit:
Thanks, can you please ack https://github.com/theforeman/foreman/pull/7266

evgeni:
yes @pcreech can you ack stuff if you merge it? ;))))

pcreech:
ah, didn't initially see the attached PR, looking

PR #4519, opened 2019-12-18
Update rubygem-foreman_inventory_upload to 1.0.3

laviro:


evgeni:
http://koji.katello.org/koji/taskinfo?taskID=262241

PR #4523, opened 2019-12-18

ekohl:


evgeni:
```
LoadError: cannot load such file -- prometheus/client
```
:thinking:

evgeni:
Ah, it's missing a `BuildRequires: %{?scl_prefix}rubygem(prometheus-client)`

evgeni:
And I am lost why `update-requirements` wouldn't pick that up from `bundler.d/telemetry.rb`

evgeni:
resolved conflicts

PR #4539, opened 2020-01-03
Fix path to autocomplete script

xprazak2:

### DIFF
--- a/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
+++ b/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
@@ -4,7 +4,7 @@
 %global gem_name hammer_cli
 %global confdir hammer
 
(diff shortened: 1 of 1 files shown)


xprazak2:
All fixed now.

packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
```diff
@@ -109,6 +109,7 @@ install -m 644 .%{gem_instdir}/config/cli_config.template.yml \
 %files
 %dir %{gem_instdir}
 %{_root_bindir}/hammer
+%{_root_bindir}/hammer-complete
```
ekohl: A side effect of this is that `ha<tab>` wil mean the user needs to type a space and then tab again. I'd recommend to look at options to avoid this. A common way to deal with this is using libexec.

PR #4554, opened 2020-01-07
Drop migrate seed

ehelms:


ehelms:
[test rpm]

ehelms:
This updates all available plugins that will currently build. The repoclosure is due to building plugins and Katello so is expected.

My plan for dropping these:

 * Merge this PR with everything we know that will build
 * Merge PR that drops macros from Foreman
 * Any plugins still with these macros can get updated when next they have open PRs to update their builds to work

PR #4556, opened 2020-01-08
Move omaha update script to correct directory for cron

lhw:
By my mistake the update cron script is installed into a subdirectory of cron.daily which is not being executed by cron. This change corrects that mistake.


mmoll:
ok to test

mmoll:
@lhw please add a new changelog entry for a new `0.0.5-2` version.

mmoll:
merged, danke @lhw!

PR #4576, opened 2020-01-13
Drop turbolinks dependency

ezr-ondrej:
Drops dependency on turbolinks, Obsoletes it in the `scl` but leaves it out in the repo.


ezr-ondrej:
> This is missing the addition of coffee-rails that's in assets.

Like this in the `assets` should be enough, right?

packages/foreman/foreman/foreman.spec
```diff
@@ -98,8 +98,6 @@ Requires: %{?scl_prefix}rubygem(rails-i18n) >= 5.0
 Requires: %{?scl_prefix}rubygem(rails-i18n) < 6.0
 Requires: %{?scl_prefix_ror}rubygem(i18n) >= 1.1
 Requires: %{?scl_prefix_ror}rubygem(i18n) < 2.0
-Requires: %{?scl_prefix_ror}rubygem(turbolinks) >= 2.5.4
```
tbrisker: should we also obsolete this?

packages/foreman/tfm/tfm.spec
```diff
@@ -61,6 +61,7 @@ Obsoletes: ruby193-ruby-wrapper
 
 # Obsolete packages no longer carried within this SCL to both clean up
 # and to ensure smooth upgrades when old packages deps aren't satisified
+Obsoletes: %{scl_prefix_ror}rubygem-jquery-turbolinks < 3.0
```
ekohl: IMHO it's not very clean to obsolete something from another collection here. However, this particular example is also from the regular scl. It's also custom to pin on `Version-(Release+1)`. See https://github.com/theforeman/foreman-packaging/tree/rpm/develop#howto-removing-a-package
```suggestion
Obsoletes: %{scl_prefix}rubygem-jquery-turbolinks < 2.1.0-9
```

package_manifest.yaml
```diff
@@ -126,7 +126,6 @@ foreman_scl_packages:
     rubygem-httpclient: {}
     rubygem-http-cookie: {}
     rubygem-ipaddress: {}
-    rubygem-jquery-turbolinks: {}
```
ekohl: Please also remove the sources in `packages/foreman/rubygem-jquery-turbolinks`

packages/foreman/foreman/foreman.spec
ekohl: No, obsoleting something from a different collection isn't correct IMHO.

packages/foreman/foreman/foreman.spec
```diff
@@ -310,8 +306,7 @@ BuildRequires: %{?scl_prefix}npm(react-intl) < 3.0.0
 # end package.json dependencies BuildRequires
 
 # start specfile assets BuildRequires
-BuildRequires: %{?scl_prefix_ror}rubygem(jquery-turbolinks) >= 2.1
-BuildRequires: %{?scl_prefix_ror}rubygem(jquery-turbolinks) < 3.0
+BuildRequires: %{?scl_prefix_ror}rubygem(coffee-rails) >= 4.2.2
```
ekohl: You're missing the upper bounds. https://github.com/ekohl/theforeman-plumbing/blob/master/fm-packaging-update-deps would have done all of this for you and made sure that in the future it'd also remain consistently sorted.

ezr-ondrej:
Any idea why it is looking for `tfm-rubygem-jquery-turbolinks = 0:2.1.0-8.el7`?

ezr-ondrej:
Ok, it passes if I obsolated the `-doc` packages. So what to do about it? Is it acceptable to obsolete both? :thinking:

zjhuntin:
I think it's acceptable to obsolete both, may seem stringent, but it covers all the bases.

ekohl:
Anyone from delivery minds merging and doing the cleanup?

tbrisker:
Looks like this needs a rebase please

ezr-ondrej:
@tbrisker done :)

PR #4577, opened 2020-01-13
Require foreman-maintain from foreman-installer

ehelms:
The use of foreman-maintain in the installer as part of the PostgreSQL
upgrade adds a requirement for foreman-maintain to be present.


ehelms:
Related to https://github.com/theforeman/foreman-installer/pull/444#discussion_r365871700

ehelms:
I swear I did one, will update.

ehelms:
Updated.... making a change locally, but not committing it...:facepalm:

PR #4578, opened 2020-01-13
Add all the packages for moving over the rails SCL part 1

zjhuntin:


zjhuntin:
Requires https://github.com/theforeman/foreman-packaging/pull/4579

zjhuntin:
This failed on repoclosure, but I think those errors are expected.

packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -32,8 +32,8 @@ Requires:      %{?scl_prefix_ruby}ruby(release)
 Requires:       foreman-debug
 Requires:       %{?scl_prefix_ruby}rubygems
 Requires:       %{?scl_prefix_ruby}rubygem(rake) >= 0.8.3
-Requires:       %{?scl_prefix_ror}rubygem(sinatra)
-Requires:       %{?scl_prefix_ror}rubygem(rack) >= 1.1.0
+Requires:       %{?scl_prefix}rubygem(sinatra)
```
ekohl: If you keep these changes out of it and only add packages, repoclosure should pass. I'd recommend doing that as part of the last step.

zjhuntin: Hrm, it actually failed repoclosures without these changes in, is there something else that can be done to avoid repoclosures then?

PR #4579, opened 2020-01-13
Add all the packages for moving over the rails SCL part 2

zjhuntin:


zjhuntin:
Requires https://github.com/theforeman/foreman-packaging/pull/4578

packages/foreman/rubygem-sprockets/rubygem-sprockets.spec
```diff
@@ -0,0 +1,105 @@
...
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+Obsoletes: tfm-ror52-rubygem-%{gem_name} <= 3.7.2
+
+# start specfile generated dependencies
+Requires: %{?scl_prefix_ruby}ruby(release)
+Requires: %{?scl_prefix_ruby}ruby >= 1.9.3
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
+Requires: %{?scl_prefix_ruby}rubygem(rack) > 1
+Requires: %{?scl_prefix_ruby}rubygem(rack) < 3
```
ehelms: This should be `scl_prefix` as we are building this package and not come from the Ruby SCL.

ehelms: There are other packages that require Rack and are having the same issue.

ehelms:
I think the Obsoletes for the `tfm-ror52` packages on these might be causing the repoclosure failures. This will mean these all have to get built, then all Foreman packages updated before adding these to the comps and pushing out to the main repository. 

The alternative is to drop the Obsoletes for now, build these, update Foreman packages and then come back through and add Obsoletes to clean up for upgrades.

I tend to like the latter as it's more forgiving. @evgeni @ekohl thoughts?

evgeni:
Yeah, just adding the packages w/o the obsoletes now allows for the smoothest transition, as we can rebuild the dependant packages as we go, and then add the obsoletes as soon as nothing depends on the old ror52 names.

PR #4583, opened 2020-01-14
Define scl for Foreman rubygems and nodejs packages

ehelms:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -18,6 +18,7 @@ foreman_packages:
 
 foreman_scl_packages:
   vars:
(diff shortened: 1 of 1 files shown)


ehelms:
Is useful once https://github.com/theforeman/obal/pull/225 is merged, this should speed up larger rebuilds as we will skip trying to build them all together rather than Koji attempting to build and failing.

zjhuntin:
were you also meaning to remove that mime-types-data gem?

ehelms:
Oops, I was not. Fixed.

PR #4585, opened 2020-01-15
Update rubygem-foreman_ansible to 5.0.0

ares:


ares:
this should fix ansible in nightlies

PR #4586, opened 2020-01-15
Update foreman_ansible to 5.0.0

ares:

### DIFF
--- a/plugins/ruby-foreman-ansible/debian/control
+++ b/plugins/ruby-foreman-ansible/debian/control
@@ -2,10 +2,10 @@ Source: ruby-foreman-ansible
 Section: ruby
 Priority: extra
 Maintainer: Daniel Lobato Garcia <elobatocs@gmail.com>
-Build-Depends: debhelper, foreman (>= 1.24.0~rc1), ruby-foreman-deface (<<2.0.0),
-  foreman-assets(>= 1.24.0~rc1), ruby-foreman-tasks (>= 0.8.2-2), ruby-foreman-tasks (<<1.0.0),
-  ruby-foreman-remote-execution (>= 2.0.0-1), ruby-foreman-remote-execution (<<3.0.0),
-  foreman-sqlite3 (>= 1.24.0~rc1)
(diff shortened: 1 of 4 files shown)


plugins/ruby-foreman-ansible/debian/control
```diff
@@ -2,18 +2,17 @@ Source: ruby-foreman-ansible
 Section: ruby
 Priority: extra
 Maintainer: Daniel Lobato Garcia <elobatocs@gmail.com>
-Build-Depends: debhelper, foreman (>= 1.24.0~rc1), ruby-foreman-deface (<<2.0.0),
-  foreman-assets(>= 1.24.0~rc1), ruby-foreman-tasks (>= 0.8.2-2), ruby-foreman-tasks (<<1.0.0),
-  ruby-foreman-remote-execution (>= 2.0.0-1), ruby-foreman-remote-execution (<<3.0.0),
-  foreman-sqlite3 (>= 1.24.0~rc1)
+Build-Depends: debhelper, foreman (>= 2.0.0~rc1), ruby-foreman-deface (<<2.0.0),
```
tbrisker: Looks like this is failing because we don't have 2.0.0~rc1 yet

mmoll: We have that in nightly. :)

plugins/ruby-foreman-ansible/debian/control
```diff
@@ -2,18 +2,17 @@ Source: ruby-foreman-ansible
...
+  foreman-assets(>= 2.0.0~rc1), ruby-foreman-tasks (>= 0.8.2-2), ruby-foreman-tasks (<<1.0.0),
```
mmoll: `ruby-foreman-tasks (<<1.0.0)` is the problem here.

ares: oops, missed to drop this dependecy as well, will update in a sec

mmoll:
merged, díky @ares!

PR #4600, opened 2020-01-16
FDI service: depend on SCL'ed FDI plugin package

evgeni:


lzap:
Thank you I will start using that in the kickstart immediately, later I will move the code from FDI repo into a RPM.

PR #4601, opened 2020-01-16
Drop rubygem-mysql2 and cleanup tito.props

ehelms:


ehelms:
Updated to fix journald packages

ehelms:
When I look at this PR locally, that directory is cleaned up. Am I missing something?

PR #4606, opened 2020-01-16
Add pulp3 debug info to katello-debug

sjha4:
Adding some pulp3 conf and debug info to katello-debug


packages/katello/katello/katello-debug.sh
```diff
@@ -65,6 +65,11 @@ add_files /var/log/httpd/pulp-http{s,}_access_ssl.log*
 add_files /var/log/httpd/pulp-http{s,}_error_ssl.log*
 add_files /etc/default/pulp*
 
+#Pulp3
```
ekohl: Nit: ideally we keep the whitespace consistent.
```suggestion
# Pulp3
```

PR #4611, opened 2020-01-17
Fixing pulpcore tags

pcreech:


pcreech:
@ekohl Yeah, i only searched rel-eng initially.

Should have hit all the spots now

ekohl:
Confirmed that my verification script now recognizes all the pulpcore packages correctly as released.

PR #4613, opened 2020-01-17
Remove _ror scl from rails packages

zjhuntin:


ekohl:
I'd combine this PR with removing all dependencies on `_ror` packages. For some context, it's used in gem2rpm templates to (re)generate package specs. Since gem2rpm is called by [update-requirements](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/update-requirements), it's also used in bumping packages.

PR #4614, opened 2020-01-17
Package puppet gem for installer

ehelms:
This is to track the packaging work that would be required as part of the RFC to switch the installer to use a Puppet gem packaged puppet solely. More details on the RFC:

https://community.theforeman.org/t/rfc-drop-reliance-on-puppet-agent-aio-for-foreman-installer/16987/6

This will be followed up by Redmine trackers and other code change PRs to show all the changes necessary. For now, I wanted to have some code to look at along side the RFC itself as I have had successful installer runs.


ehelms:
How this looks in kafo:

https://github.com/theforeman/kafo/pull/246
https://github.com/theforeman/kafo_parsers/pull/36

ekohl:
Don't forget you need to update Facter to version 3 as well. That means pulling more complex build dependencies (https://github.com/puppetlabs/facter#build-requirements) since it's a C++ tool. The puppet 6 gem also doesn't bundle some modules. See https://src.fedoraproject.org/rpms/puppet/pull-request/6 for more details.

ehelms:
Gimmie a little more insights.

What drives the need to update facter to version 3?
Why do I need worry about bundled puppet modules?

ekohl:
> What drives the need to update facter to version 3?

Because Puppet modules have started to rely on modern facts provided by Facter 3. Even the installer itself. For example, the `mountpoints` fact we use in the hooks doesn't exist in Facter 2.

>  Why do I need worry about bundled puppet modules?

In Puppet 5 and before resources like `cron` and `yumrepo` were bundled in the gem but in Puppet 6 these are hosted standalone modules (like [cron_core](https://github.com/puppetlabs/puppetlabs-cron_core) and [yumrepo_core](https://github.com/puppetlabs/puppetlabs-yumrepo_core)) bundled in the puppet-agent package. We don't use these a lot, but we have some cases where they're used.

ehelms:
For informational purposes, I was able to mock build `leatherman`, `hocon-cpp` and then `facter 3.14` locally using our `mock/el7-scl.cfg`. So upgrading Facter shouldn't be an issue. I thought I recall talking of wanting to update it for other purposes as well within Foreman.

PR #4619, opened 2020-01-17
Another round of plugins dropping posttrans macros

ehelms:


ehelms:
Expected failure given this contains plugins and katello RPMs

PR #4623, opened 2020-01-20
rpm no longer likes "(" and ")" in obsoletes

pcreech:


pcreech:
build: http://koji.katello.org/koji/taskinfo?taskID=273667

PR #4625, opened 2020-01-20
Fix nokogiri.so placement

zjhuntin:


packages/foreman/rubygem-nokogiri/rubygem-nokogiri.spec
```diff
@@ -71,8 +71,8 @@ cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
 mkdir -p %{buildroot}%{gem_extdir_mri}
-cp -a .%{gem_extdir_mri}/gem.build_complete %{buildroot}%{gem_extdir_mri}/
-cp -a .%{gem_extdir_mri}/%{gem_name}/*.so %{buildroot}%{gem_extdir_mri}/
+cp -a .%{gem_extdir_mri}/* %{buildroot}%{gem_extdir_mri}/
```
ekohl: I'd be more specific and copy it directly
```suggestion
cp -a .%{gem_extdir_mri}/%{gem_name}/*.so %{buildroot}%{gem_extdir_mri}/%{gem_name}
```

Normally I like that since it's easier to detect changes. You also don't accidentally copy other files like `.la` if they're present.

zjhuntin:
@ekohl updated and it's passing!

packages/foreman/rubygem-nokogiri/rubygem-nokogiri.spec
```diff
@@ -71,8 +71,8 @@ cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
 mkdir -p %{buildroot}%{gem_extdir_mri}
-cp -a .%{gem_extdir_mri}/gem.build_complete %{buildroot}%{gem_extdir_mri}/
```
ekohl: This file is still needed since it signals Ruby that it should expect a .so file.

packages/foreman/rubygem-nokogiri/rubygem-nokogiri.spec
```diff
@@ -71,8 +71,9 @@ cp -a .%{gem_dir}/* \
...
-cp -a .%{gem_extdir_mri}/%{gem_name}/*.so %{buildroot}%{gem_extdir_mri}/
+cp -a .%{gem_extdir_mri}/%{gem_name}/gem.build_complete %{buildroot}%{gem_extdir_mri}/%{gem_name}
```
ekohl: I don't know if it needs to live in the subdirectory or not. Did you test this out?

zjhuntin: I did not but looking back at what was in the spec it looks like that's the wrong place for it to live. So I'm gonna change it.

packages/foreman/rubygem-nokogiri/rubygem-nokogiri.spec
zjhuntin: I never knew that, thanks for the info!

PR #4628, opened 2020-01-21
Refs #28068 - Drop old dynflowd service

ekohl:
Now that foreman-dynflow-sidekiq is in place, there's no more need for the old dynflowd.

Note that the old dynflowd had a sysvinit script but the new one only works with systemd. This is because it uses multiple instances and this keeps the installer simple. Technically it could work with sysvinit, but no such support is planned.


mmoll:
merged, bedankt @ekohl!

PR #4629, opened 2020-01-21
Create gem name folder to put nokogiri.so into it.

zjhuntin:
Realized that it wasn't copying over nokogiri.so correctly, it was instead renaming it as nokogiri. This creates the folder it's supposed to go into and then correclty copies nokogiri.so there.

### DIFF
--- a/packages/foreman/rubygem-nokogiri/rubygem-nokogiri.spec
+++ b/packages/foreman/rubygem-nokogiri/rubygem-nokogiri.spec
@@ -6,7 +6,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 1.8.4
-Release: 4%{?dist}
+Release: 5%{?dist}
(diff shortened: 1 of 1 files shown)


packages/foreman/rubygem-nokogiri/rubygem-nokogiri.spec
```diff
@@ -71,8 +71,9 @@ cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
 mkdir -p %{buildroot}%{gem_extdir_mri}
+mkdir -p %{buildroot}%{gem_extdir_mri}/%{gem_name}
```
ekohl: I was already wondering why no mkdir was needed for it. Technically the line above is no longer needed since it's `-p`.

zjhuntin: Good call, went ahead and removed the part above.

PR #4637, opened 2020-01-22
Update rubygem-foreman_scc_manager to 1.8.0 for foreman-1.24

m-bucher:
Cherry-Pick of #4636
(cherry picked from commit a3f0ba2bdd68e74096f9f7ae9fa22e9ff9949e11)


adamruzicka:
Nope, 1.x is 1.25/2.0 only

m-bucher:
The dependency changes to `foreman-tasks >= 0.10`, so as long as this is the case for 1.24 it should be OK.

m-bucher:
see #5102

PR #4640, opened 2020-01-22
Include .babelrc.js in foreman-assets

ehelms:
Fixes the REX build issue


ekohl:
I don't think you should include bumping REX since that needs foreman to actually be built with the patch.

PR #4641, opened 2020-01-22
Update foreman_snapshot_management to 1.7.0

mdellweg:


tbrisker:
[test rpm]

tbrisker:
repoclosure failure was due to salt's dependency conflict that's been fixed since. merging, thanks @mdellweg !

mdellweg:
Thank you for looking into that failure, @tbrisker !

PR #4642, opened 2020-01-23
Update rubygem-foreman_azure_rm to 2.0.6

apuntamb:
Supported Versions:

 * Nightly
 * 1.24
 * 1.23
 * 1.22


apuntamb:
I have a concern here. I can see in that a change has been made here in #4619 to drop the `posttrans` macro. My queries are:
1) Will it affect the by-default running of `apipie:cache`?
2) Since I will be cherry-picking this commit to `rpm/1.24`, will this change create conflicts in the stable branch?
@evgeni  @ehelms  thoughts?

evgeni:
The repoclosure error is unrelated, so this can be merged as is.

WRT your question, you can just CP this to 1.24, without the posttrans changes, and the only thing you'll have to fix up is the revision/changelog entry. It should work fine otherwise.
And yes, this does affect the running, but it does move it to the installer, so we should be still fine.

apuntamb:
> The repoclosure error is unrelated, so this can be merged as is.
> 
> WRT your question, you can just CP this to 1.24, without the posttrans changes, and the only thing you'll have to fix up is the revision/changelog entry. It should work fine otherwise.
> And yes, this does affect the running, but it does move it to the installer, so we should be still fine.

Just to confirm, `posttrans` section **should not be present** in CP pr, right?

PR #4648, opened 2020-01-23
pulpcore selinux

evgeni:


ehelms:
I'm wondering what should require this to ensure it gets installed.

ekohl:
For `foreman-selinux` it's `foreman-installer` in packaging but also `foreman::install` in the installer. Then `foreman-installer-katello` depends on `katello-selinux`. It probably makes sense to ensure that in the katello module as well.

That makes me think that `foreman-installer-katello` should depend on it but also `pulpcore::install` should ensure it.

evgeni:
In pulp2, `pulp-server` requires `pulp-selinux` in https://github.com/pulp/pulp-packaging/blob/dca4a9ebe9f46aa50e14f366d084fa418eb75278/packages/pulp/pulp.spec#L381

However, https://fedoraproject.org/wiki/SELinux_Policy_Modules_Packaging_Draft#Runtime_Dependencies says: "If you are not building an integrated package, then the myapp-selinux package needs to have a dependency on the myapp package. This is necessary to ensure that the packages are installed in the correct order if both packages are installed in the same rpm transaction, which in turn is necessary in order to ensure that the post-install scriptlet of the myapp-selinux package can set the correct file contexts."

And this funnily contradicts https://fedoraproject.org/wiki/SELinux/IndependentPolicy#Adding_dependency_to_the_spec_file_of_corresponding_package :)

evgeni:
@jlsherrill @dkliban I'd appreciate some testing ;-)

packages/pulpcore/pulpcore-selinux/pulpcore-selinux.spec
```diff
@@ -0,0 +1,81 @@
...
+Summary:        SELinux policy for Pulp 3
+
+License:        GPL2+
+URL:            https://pulpproject.org
+Source0:        https://codeload.github.com/pulp/%{name}/tar.gz/%{version}#/%{name}-%{version}.tar.gz
+
+BuildRequires:  checkpolicy
+BuildRequires:  selinux-policy-devel
+Requires:       selinux-policy >= %{_selinux_policy_version}
+Requires:       python3-pulpcore
```
evgeni: I am still unsure whether pulpcore-selinux should require python3-pulpcore or the other way round :woman_shrugging:

dgoetz: You only want to ensure when the policy is installed it installs also the base package in most cases. If someone has selinux disabled you do not want to install the policy package. So pulpcore-selinux should require python3-pulpcore and not the other way round.

Does RPM in EL7 already support the rich rule syntax to get the conditional dependency? I am not sure, but I think no. If yes, this could also solve the problem in a more elegant way.

evgeni: No, rich deps are not a thing in EL7, sadly :(

PR #4652, opened 2020-01-27
Add Obsoletes of all the rails-packaging versions

zjhuntin:


packages/foreman/rubygem-actionview/rubygem-actionview.spec
```diff
@@ -34,6 +34,8 @@ BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 # end specfile generated dependencies
 
+Obsoletes: tfm-ror52-rubygem-%{gem_name} <= 5.2.1
```
ekohl: Looks like you forgot to bump this one

packages/foreman/rubygem-multi_json/rubygem-multi_json.spec
```diff
@@ -31,6 +31,8 @@ BuildRequires: %{?scl_prefix_ruby}ruby(rubygems)
 BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 
+Obsoletes: tfm-ror52-rubygem-%{gem_name} <= 1.12.2
```
ekohl: Missed the bump here

ehelms:
Adding the collection to here https://github.com/theforeman/forklift/blob/master/bats/fb-verify-packages.bats#L15 as part of this getting merged will be useful for catching any missing packages.

evgeni:
the passenger repoclosure issue *seems* unrelated?

ekohl:
Is that because of the version mismatch in our repo and the one in EPEL?

ehelms:
The current repoclosure error is just nightly pipeline needing to run and update. So yes, it is unrelated.

PR #4656, opened 2020-01-27
Bump nodejs-theforeman-vendor to 4.0.7

tbrisker:


packages/foreman/nodejs-theforeman-vendor/nodejs-theforeman-vendor.spec
```diff
@@ -4,13 +4,13 @@
...
 Name: %{?scl_prefix}nodejs-theforeman-vendor
-Version: 4.0.2
+Version: 4.0.7
 Release: 1%{?dist}
 Summary: foreman supported 3rd-party node_modules
 License: MIT
 Group: Development/Libraries
 URL: https://github.com/theforeman/foreman-js#readme
-Source0: https://registry.npmjs.org/%{npm_name}/-/%{npm_name}-%{version}.tgz
+Source0: https://registry.npmjs.org/@theforeman/vendor/-/vendor-%{version}.tgz
```
ekohl: This should also work
```suggestion
Source0: https://registry.npmjs.org/%{npm_name}/-/vendor-%{version}.tgz
```

tbrisker: it would, but would require more work to fix in npm2rpm. do we get much benefit from it?

tbrisker:
rpm is actually green, the failure is in the notify gh step

packages/foreman/nodejs-theforeman-vendor/nodejs-theforeman-vendor.spec
ekohl: Now that I've seen https://github.com/theforeman/npm2rpm/pull/62 it makes sense. I wonder why this worked before.

ekohl:
[test rpm]

PR #4657, opened 2020-01-28
Rebuild webpack using plugins

tbrisker:
needed due to change in the way core compiles its bundle to fix issue
caused by babel 7 upgrade.

### DIFF
--- a/packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec
+++ b/packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec
@@ -11,7 +11,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 1.0.1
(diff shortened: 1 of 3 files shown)


packages/plugins/rubygem-foreman_templates/rubygem-foreman_templates.spec
```diff
@@ -139,6 +139,9 @@ cp -pa .%{gem_dir}/* \
 exit 0
```
ekohl: I think this section will be an issue since the `foreman_restart` macro was dropped by @ehelms

ekohl:
Looks like templates is still on an old vendor package. There are commits in master for compatibility with vendor v4 but needs a release. @xprazak2 any plans to do so soon?

tbrisker:
dropped templates from this pr, i missed the fact it wasn't updated for new vendor yet.

PR #4661, opened 2020-01-28
Upgrade sass rails

ezr-ondrej:
Switch from unsupported `sass` to `sassc` by upgrading to `sass-rails >= 6.0`, what is only an wrapper around the `sassc-rails`.


packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -0,0 +1,109 @@
...
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+
+mkdir -p %{buildroot}%{gem_extdir_mri}/%{gem_name}
+
+# hack to get the C extension where we expect it to be
+cp %{buildroot}/opt/theforeman/tfm/root/usr/share/gems/gems/sassc-%{version}/ext/libsass.so .%{gem_extdir_mri}/
```
ezr-ondrej: Any idea why it is building in "wrong" dir? :thinking:

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -0,0 +1,109 @@
...
+
+cp -a .%{gem_extdir_mri}/gem.build_complete %{buildroot}%{gem_extdir_mri}/
+cp -a .%{gem_extdir_mri}/*.so %{buildroot}%{gem_extdir_mri}/%{gem_name}/
+
+# Prevent dangling symlink in -debuginfo (rhbz#878863)
+# But this gem fails on it with error
+# File not found: /builddir/build/BUILDROOT/tfm-rubygem-sassc-2.2.1-1.el7.x86_64/opt/theforeman/tfm/root/usr/share/gems/gems/sassc-2.2.1/ext.
```
ezr-ondrej: No idea why this is searched for just for this gem, maybe because we know the extension should be there?

ezr-ondrej:
Why koji fails on non existent package `sass-rails >= 6`? Should I update the dependency only we merge the gem bump?

packages/foreman/rubygem-sassc/rubygem-sassc.spec
ekohl: https://github.com/theforeman/foreman-packaging/commit/472109d923e96022f5ffd1bb89bc240679cefa8e was recently merged. Perhaps this is the same issue you're running into?

ezr-ondrej: No, it doesn't seem so. It failed anyway (I've created this PR with that version in already).

ekohl:
> Why koji fails on non existent package `sass-rails >= 6`? Should I update the dependency only after we merge the gem bump?

Yes. Currently all packages in PRs to foreman-packaging are built against what's then built in koji. That means you can't depend on anything you've done inside the same PR. In this case it means you should update foreman.spec in a PR after this one is merged.

ezr-ondrej:
ehm... yeah, but now the repoclosure fails, because I removed the old deps :thinking:

ekohl:
That suggests sass can't be removed yet. It's usually worth looking at those projects to see if they plan on migrating. It may be valid to ship both as well.

ezr-ondrej:
> That suggests sass can't be removed yet. It's usually worth looking at those projects to see if they plan on migrating. It may be valid to ship both as well.

Well it needs `sass-rails`, what is the one I'm bumping, but it's out of required range, what I cant move now, as it would fail on missing package.

zjhuntin:
[test rpm]

ezr-ondrej:
I need to leave out `sass` in as it is being used by patternfly, opened up a PR there: patternfly/patternfly#1199 but not more I can do. There is a PR to update foreman references, but what I get, it needs to be done in following PR. Here it is: #4663

tbrisker:
[test rpm]

tbrisker:
repoclosure failure is expected because foreman-assets still requires `gem 'sass-rails', '~> 5.0'` but that will be fixed by https://github.com/theforeman/foreman/pull/7396/

rel-eng/tito.props
```diff
@@ -320,12 +320,11 @@ whitelist = foreman
   rubygem-spice-html5-rails
   rubygem-sprockets
   rubygem-sprockets-rails
+  rubygem-sqlite3
   rubygem-sshkey
   rubygem-statsd-instrument
-  rubygem-sqlite3
   rubygem-syntax
   rubygem-text
-  rubygem-tilt
```
ezr-ondrej: Is it correct this one is deleted? Everything else just got reordered, this one is removed though :thinking:

tbrisker: yes, it was a dependency of sass-rails 5.0

tbrisker: actually should probably be removed from the manifest and comps files as well

packages/foreman/tfm/tfm.spec
```diff
@@ -103,6 +103,7 @@ Obsoletes: %{scl_prefix}rubygem-smart_proxy_abrt < 0.0.8-4
 Obsoletes: %{scl_prefix}rubygem-sprockets < 3.5.2-2
 Obsoletes: %{scl_prefix}rubygem-sprockets-rails < 2.3.3-2
 Obsoletes: %{scl_prefix}rubygem-table_print < 1.5.1-7
+Obsoletes: %{scl_prefix}rubygem-tilt < 2.0.8-4
```
tbrisker: I think this should be done in rubygem-sass-rails that actually obsoleted this gem

packages/foreman/rubygem-sass/rubygem-sass.spec
```diff
@@ -25,6 +25,7 @@ Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 # end specfile generated dependencies
 
 Obsoletes: tfm-ror52-rubygem-%{gem_name} <= 3.4.25
+Obsoletes: %{scl_prefix}rubygem-tilt < 2.0.8-4
```
tbrisker: this isn't needed here i believe

packages/foreman/rubygem-sass-rails/rubygem-sass-rails.spec
```diff
@@ -35,6 +28,7 @@ Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 # end specfile generated dependencies
 
 Obsoletes: tfm-ror52-rubygem-%{gem_name} <= 5.0.7
+Obsoletes: %{scl_prefix}rubygem-tilt < 2.0.8-4
```
tbrisker: Linter doesn't like this, try 
```suggestion
Obsoletes: tfm-rubygem-tilt < 2.0.8-4
```

ezr-ondrej: The question mark did the trick :)

ezr-ondrej:
We are at :green_heart:ish again (only repoclosure fail)

ekohl:
@tbrisker FYI: merging PRs that add packages is still something that someone with koji access needs to merge. https://ci.theforeman.org/job/foreman-packaging-rpm-nightly-release/ will now fail until someone actually adds the package to koji.

packages/foreman/rubygem-sassc-rails/rubygem-sassc-rails.spec
```diff
@@ -0,0 +1,95 @@
...
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+%files
+%dir %{gem_instdir}
+%exclude %{gem_instdir}/.gitignore
+%exclude %{gem_instdir}/.travis.yml
+%{gem_instdir}/CODE_OF_CONDUCT.md
```
ekohl:
```suggestion
%doc %{gem_instdir}/CODE_OF_CONDUCT.md
```

tbrisker:
@ekohl oh, i thought that on develop it is all automated already?

ekohl:
Nope, only updates. I also think more review was needed around the actual RPM specs.

ezr-ondrej:
> Nope, only updates. I also think more review was needed around the actual RPM specs.

I would agree here, but that means we need more people understanding and reviewing them. As this was 5 weeks untouched, we should get more attention for the reviews here (that would mean more developer understand packaging and the reviews could be simplier as such developer would do better job on the first go).

ekohl:
I can't speak for others, but I considered it blocked on the patternfly side. My reading was that it was pointless until everything was compatible with it. I was also too busy with making nightly green again and getting 2.0 into a working state.

Next time I'll be more explicit: I had issues with the .so handling and that does indeed turn out to be broken now.

ezr-ondrej:
> My reading was that it was pointless until everything was compatible with it

It is not pointless, we will just have two sass compilation libraries, but we will use sassc for our assets. It make sense to have the new solution ready asap. Especially because new bootstrap-sass uses the new library already, so we can't upgrade until we get it in.

ekohl:
Ok, in the future I'll leave a clear message that it shouldn't be merged. For the record, the manual whitelisting in koji is precisely because of this: new packages require thorough review.

While looking into it I've found another issue that we'll need to deal with: https://github.com/sass/sassc-ruby/blob/275fe38f56cf07eb3e3c46e41d18402a94e66a5e/ext/extconf.rb#L22-L25 means the sassc only works on the CPU it was compiled on (and CPUs with a superset of instructions). That is generally not the best idea.

I can't figure it out short term and I'd like to get green nightlies for all pipelines after a month of being red: https://github.com/theforeman/foreman-packaging/pull/4836 After that, please work with our delivery team to get it to build.

PR #4662, opened 2020-01-29
Bump python-setuptools-scm to 3.4.3

evgeni:


packages/pulpcore/python-setuptools-scm/python-setuptools-scm.spec
```diff
@@ -26,6 +26,7 @@ use_scm_version...
 Summary:        %{summary}
 %{?python_provide:%python_provide python3-%{pypi_name}}
 
+Requires:       python3-toml
```
ekohl: Nit: perhaps this should have an empty line after it

evgeni: It's a "bug" in our jinja2 template, but yeah I can move that manually.

PR #4546, opened 2020-01-06
update needed pulp3 clients

jlsherrill:

PR #4673, opened 2020-01-30
bump rubygem-pulpcore_client epoch

jlsherrill:
so that its newer than the 3.0.0rc2


jlsherrill:
This repoclosure issue seemed to happen previously as well: 
https://github.com/theforeman/foreman-packaging/pull/4546#issuecomment-571290263

PR #4678, opened 2020-02-03
Remove ror_scl from tfm

zjhuntin:


packages/foreman/tfm/tfm.spec
```diff
@@ -199,7 +190,6 @@ export LD_LIBRARY_PATH=%{_libdir}\${LD_LIBRARY_PATH:+:\${LD_LIBRARY_PATH}}
 export MANPATH=%{_mandir}:\${MANPATH}
 export CPATH=%{_includedir}\${CPATH:+:\${CPATH}}
 export PKG_CONFIG_PATH=%{_libdir}/pkgconfig\${PKG_CONFIG_PATH:+:\${PKG_CONFIG_PATH}}
-export GEM_PATH=%{gem_dir}:\${GEM_PATH:+\${GEM_PATH}}\${GEM_PATH:-\`scl enable %{scl_ror} -- ruby -e "print Gem.path.join(':')"\`}
```
ehelms: We don't want to drop this as it sets the GEM_PATH for the tfm collection

packages/foreman/tfm/tfm.spec
```diff
@@ -253,7 +243,6 @@ source scl_source enable %{scl_name}
 exec rake "\$@"
 EOF
 
-scl enable %{scl_ror} - << \EOF
```
ehelms: I think we want to switch this to `scl_ruby`

packages/foreman/tfm/tfm.spec
```diff
@@ -21,7 +19,7 @@
 Summary: Package that installs %scl
 Name: %scl_name
 Version: 5.0
-Release: 11%{?dist}
+Release: 12%{?dist}
```
ehelms: I think dropping tfm-ror52 is enough to warrant a full version bump

zjhuntin: Thought that might be a thing, but good call!

PR #4679, opened 2020-02-04
Package pre-compiled python files

sbernhard:
This fixes the following issue:

  /usr/lib/yum-plugins/foreman-protector.pyc
  /usr/lib/yum-plugins/foreman-protector.pyo

  /usr/lib/yum-plugins/foreman-protector.pyc
  /usr/lib/yum-plugins/foreman-protector.pyo

### DIFF
--- a/packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec
+++ b/packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec
@@ -10,7 +10,7 @@
 Summary: The Foreman/Satellite maintenance tool
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.6.8
-Release: 1%{?dist}
+Release: 2%{?dist}
 Epoch: 1
 Group: Development/Languages
 License: GPLv3
@@ -21,6 +21,11 @@ Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
 Requires: %{?scl_prefix}rubygem(clamp) >= 0.6.2
 Requires: %{?scl_prefix}rubygem(highline)
+%if 0%{?rhel} < 8
+BuildRequires: python2-devel
+%else
+BuildRequires: python3-devel
+%endif
 BuildRequires: %{?scl_prefix_ruby}rubygems-devel
 BuildRequires: %{?scl_prefix_ruby}ruby(release)
 BuildRequires: %{?scl_prefix_ruby}ruby(rubygems)
@@ -76,7 +81,13 @@ install -D -m0640 %{buildroot}%{gem_instdir}/config/passenger-recycler.yaml %{bu
 
 install -D -m0640 %{buildroot}%{gem_instdir}/extras/foreman_protector/foreman-protector.conf %{buildroot}%{_sysconfdir}/yum/pluginconf.d/foreman-protector.conf
 install -D -m0640 %{buildroot}%{gem_instdir}/extras/foreman_protector/foreman-protector.whitelist %{buildroot}%{_sysconfdir}/yum/pluginconf.d/foreman-protector.whitelist
-install -D -m0640 %{buildroot}%{gem_instdir}/extras/foreman_protector/foreman-protector.py %{buildroot}%{yumplugindir}/foreman-protector.py 
+install -D -m0640 %{buildroot}%{gem_instdir}/extras/foreman_protector/foreman-protector.py %{buildroot}%{yumplugindir}/foreman-protector.py
+
+%if 0%{?rhel} < 8
+%py_byte_compile %{__python2} %{buildroot}%{yumplugindir}/foreman-protector.py
+%else
+%py_byte_compile %{__python3} %{buildroot}%{yumplugindir}/foreman-protector.py
+%endif
 
 %files
 %dir %{gem_instdir}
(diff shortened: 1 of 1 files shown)


sbernhard:
Best practice is to package pre compiled files:

https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#Byte_Compiled_Files

sbernhard:
[test rpm]

ekohl:
In the resulting RPM I don't see any .pyc files and I think that's because it's not a Python package. https://docs.fedoraproject.org/en-US/packaging-guidelines/Python_Appendix/#manual-bytecompilation suggests you can do it manually.

sbernhard:
I guess, this depends. In our build workflow the *.pyc / *.pyo are created. Changing it to "_protector.py*" shouldn't harm.

ehelms:
@sbernhard this needs a rebase

evgeni:
@sbernhard as Eric said this needs a rebase, but even then it requires manual compilation as Ewoud writes, as the files are in a custom folder where the normal build process does not build them.

However, I don't see a huge benefit of providing this versus making Python generate that on demand? It's not like it's a lot of code that takes long to compile.

dgoetz:
@evgeni: Problem is not compile time, it leaves a file that will not be uninstalled if the package gets removed and depending on SELinux and permissions something can go wrong during compile.

evgeni:
@dgoetz see, the PR description didn't say what it tries to fix :)

evgeni:
repoclosure errors unrelated. however, the package still does not contain byte compiled files.

sbernhard:
ping!?

evgeni:
@sbernhard the built RPMs still don't contain the byte compiled files as nothing compiles them.

sbernhard:
This depends on the RPM build configuration. In other environments (like our internal one) this issue happens. Therefore it would be nice to merge this as it doesn't break anything.

evgeni:
Which configuration (and how) would I need to change?
I'd prefer if the package would do it, so the change is fully self contained.
Otherwise this is environment dependent and the changelog entry is kinda misleading for users of this repository.

sbernhard:
Use "rpm --showrc | less"

In a "default" centos installations, this is included: 
-14: __os_install_post  
    /usr/lib/rpm/redhat/brp-compress 
    %{!?__debug_package:
    /usr/lib/rpm/redhat/brp-strip %{__strip} 
    /usr/lib/rpm/redhat/brp-strip-comment-note %{__strip} %{__objdump} 
    } 
    /usr/lib/rpm/redhat/brp-strip-static-archive %{__strip} 
    /usr/lib/rpm/brp-python-bytecompile %{__python} %{?_python_bytecompile_errors_terminate_build} 
    /usr/lib/rpm/redhat/brp-python-hardlink 
    %{!?__jar_repack:/usr/lib/rpm/redhat/brp-java-repack-jars} 
%{nil}

I guess, on the build host which is used to create the foreman/katello packages this is turned off.

ekohl:
I'd suggest to explicitly do this. In Fedora there's https://fedoraproject.org/wiki/Changes/No_more_automagic_Python_bytecompilation so this is likely to be going away. Have you verified it also works on EL8?

The correct way is:
```
%py_byte_compile %{__python2} %{buildroot}%{_datadir}/mypackage/foo
%py_byte_compile %{__python3} %{buildroot}%{_datadir}/mypackage/bar
```

sbernhard:
@ekohl, I guess you mean to change the foreman build environment and "byte compile" every python file? 
I would love to get this done. Don't want to do another rebase :)

evgeni:
I (personally) think the foreman-maintain packaging should take care of the byte compilation.

sbernhard:
Nobody interested in this change? :(

evgeni:
Not before the byte compilation is in this PR, no.

evgeni:
This seems to have worked on EL7, thanks!

```
-rw-r-----    1 root     root                     2707 Aug  3 13:21 /usr/lib/yum-plugins/foreman-protector.py
-rw-r--r--    2 root     root                     3318 Aug  3 13:21 /usr/lib/yum-plugins/foreman-protector.pyc
-rw-r--r--    2 root     root                     3318 Aug  3 13:21 /usr/lib/yum-plugins/foreman-protector.pyo
```

On EL8 however, it fails now:

```
+ xargs -0 /usr/libexec/platform-python -O -m py_compile
+ find ./usr/lib/yum-plugins/foreman-protector.py -type f -a -name '*.py' -print0
  File "./usr/lib/yum-plugins/foreman-protector.py", line 29
    except urlgrabber.grabber.URLGrabError, e:
                                          ^
SyntaxError: invalid syntax
+ failure=1
+ xargs -0 /usr/libexec/platform-python -m py_compile
+ find ./usr/lib/yum-plugins/foreman-protector.py -type f -a -name '*.py' -print0
  File "./usr/lib/yum-plugins/foreman-protector.py", line 29
    except urlgrabber.grabber.URLGrabError, e:
                                          ^
SyntaxError: invalid syntax
```

Seems f-maintain's code is not Python 3 compatible :scream: (it needs to be `except Exc as e`)

sbernhard:
Disable python3 byte compile til https://github.com/theforeman/foreman_maintain/pull/382 is available?

evgeni:
Let me ping foreman maintain people if they plan a release *soon*, otherwise this sounds like a plan.

tbrisker:
This has been incorporated into https://github.com/theforeman/foreman-packaging/pull/5618, thanks @sbernhard and all!

sbernhard:
Thanks!

PR #4690, opened 2020-02-05
Update rubygem-fog-vsphere to 3.2.3

chris1984:


chris1984:
@tbrisker updated 1 more time with a small fix.

PR #4694, opened 2020-02-09
Bump hammer to 2.0-develop

tbrisker:

### DIFF
--- a/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
+++ b/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
@@ -4,7 +4,7 @@
 %global gem_name hammer_cli
 %global confdir hammer
 
-%global release 3
+%global release 1
 %global prereleasesource pre.develop
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 
@@ -14,7 +14,7 @@
 
 Summary: Universal command-line interface for Foreman
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.20
+Version: 2.0.0
 Release: %{?prerelease:0.}%{release}%{?prerelease}%{?nightly}%{?dist}
 Group: Development/Languages
 License: GPLv3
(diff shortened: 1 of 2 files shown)


tbrisker:
This will likely fail until https://github.com/theforeman/hammer-cli/pull/324 and https://github.com/theforeman/hammer-cli-foreman/pull/491 are merged

tbrisker:
[test rpm]

packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
```diff
@@ -14,7 +14,7 @@
 
 Summary: Universal command-line interface for Foreman
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.20
+Version: 2.0
```
ekohl: Should we make it `2.0.0`? Otherwise it might bump to `2.1` and the question becomes whether that's for Foreman 2.1.0 or Foreman 2.0.1. This should also be reflected in hammer itself.

tbrisker: I kept this inline with previous nightlies (e.g. 0.20). In the stable branch we actually do have e.g. 0.19.3, so once we branch this will become 2.0.0 indeed.

ekohl: I'd advise to make 2.0.0 everywhere to make it clear we follow a x.y.z scheme.

tbrisker:
@ekohl The idea was to make the "x.y" part in sync with foreman so it's easy to tell which versions are compatible, but keep the .z part free to do releases as needed out of sync with foreman

packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
tbrisker: I'm not sure the current pipelines will properly pick it up but i can try.

tbrisker:
@ekohl build is now failing to find the correct version of the gem (was green previously).

tbrisker:
changed back to 2.0.

ekohl:
It failed because hammer itself is 2.0 and it has to match. So you'd need to bump hammer.git and then update the packaging PR.

tbrisker:
looks like hammer itself does indeed use 2.0-develop (or 0.20-develop previously) for nightlies, and 2.0.0 for actual releases. do you think those should change now? it may break other workflows that expect this versioning scheme.

ekohl:
IMHO it should change that no because switching between x.y and x.y.z hurts my brain. I always have the feeling RPMs and gems treat those two different but I don't recall any specific instance where it actually caused a problem.

tbrisker:
OK, updated in hammer and here (will fail now likely until hammer-source-release finishes)

tbrisker:
[test rpm]

tbrisker:
all green now :)

PR #4697, opened 2020-02-11
[DO NOT MERGE] Updating pulp_2to3_migration_client

johnpmitsch:


packages/katello/rubygem-pulp_2to3_migration_client/rubygem-pulp_2to3_migration_client.spec
```diff
@@ -84,5 +84,8 @@ cp -a .%{gem_dir}/* \
 %{gem_instdir}/spec
 
 %changelog
+* Mon Feb 10 2020 <John Mitsch> 0.0.1b2.dev01581272120-1
```
ekohl: Normally this is in the form of `My Name <myname@example.com>`. Interesting that Justin also didn't do this.

zjhuntin:
[test rpm]

johnpmitsch:
I was able to build the new gem faked out to the old version, which will work for now until the actual branching happens and we can update the katello RPM to require a later version of this gem. 

If you need the RPM, it is available here: http://koji.katello.org/koji/taskinfo?taskID=284342

PR #4698, opened 2020-02-11
Fixes #28979 - Drop the foreman-rails repo

ekohl:
All packages were included in the normal repository as regular packages so this repository is no longer needed.


tbrisker:
repoclosure is failing for el8 for some reason:
```
The task includes an option with an undefined variable. The error was: 'dict object' has no attribute u'el8'

tbrisker:
pushed another rebase

PR #4701, opened 2020-02-11
Bump nightly to version 3.16.0

jjeffers:


ekohl:
We haven't branched yet so this shouldn't be merged until we do.

PR #4706, opened 2020-02-11
Fixes #28939 - change-hostname: update DNS records when changing hostname

jeremylenz:
When applicable, this change updates DNS records using `nsupdate` during the `[scenario]-change-hostname` script.

To test on a sat-deploy box:
* make sure DNS is managed by satellite (run the satellite-provisioning playbook)
* View the contents of `/var/named/dynamic`

```
sudo su
cd /var/named/dynamic
ls -l
```
You should see a DNS zone file named for your zone, such as `db.example.com`.  Inspect the contents and note that they contain the current satellite hostname in the SOA, A, and NS records:
```
[root@sat-6-7-qa-rhel7 dynamic]# cat db.example.com
$TTL 10800
@ IN SOA sat-6-7-qa-rhel7.redhatlaptop.example.com. root.example.com. (
	1	;Serial
	86400	;Refresh
	3600	;Retry
	604800	;Expire
	3600	;Negative caching TTL
)

@ IN NS sat-6-7-qa-rhel7.redhatlaptop.example.com.

sat-6-7-qa-rhel7.redhatlaptop.example.com. IN A 192.168.73.1
```
The reverse zone file (named similar to `db.73.168.192.in-addr.arpa`) will also contain records referring to the current hostname.

* Set permissions on `hostname-change.rb`:
```
cd /usr/share/katello
chmod a+w hostname-change.rb
```

* On your local machine in the `sat-deploy` directory, run
```
vagrant ssh-config [box_name]
```
* Take note of the IdentityFile and HostName values (HostName will actually be an IP)

* Copy the new `hostname-change.rb` from this PR to the sat-deploy box using `scp`:
```
scp -i $IDENTITY_FILE $LOCAL_HOSTNAME_CHANGE.RB vagrant@$HOSTNAME:/usr/share/katello/hostname-change.rb
```
* Verify using `cat` or `vim` that the `hostname-change.rb` contents successfully copied over
* Repeat the above with `helper.rb`

* Run
```
vim /etc/foreman-installer/scenarios.d/satellite-answers.yaml
```
* Edit the `dns_provider` value (under `foreman_proxy:`) to read `nsupdate`

* Run 
```
satellite-change-hostname -u admin -p changeme newhost.mydomain.example.com
```
* View new output
```
gathering DNS data...
updating DNS records:
forward...
reverse...
updating dynamic zone files...
DNS records updated
```

* Verify that the contents of the zone files `/var/named/db.$ZONE` now reflect the new hostname

This should allow PXE discovery to continue to work properly.

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -344,6 +344,39 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       "echo -e \"#{dns_entries}\" | nsupdate -l -k #{key_file}"
     end
 
+    def update_zone(zone, nameserver_ip)
+      resolver = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
+      commands = []
+
+      soa = resolver.getresource(zone, Resolv::DNS::Resource::IN::SOA)
+      if soa.mname.to_s == @old_hostname
+        commands << "update add #{zone} #{soa.ttl} SOA #{@new_hostname}. #{soa.rname} #{soa.serial + 1} #{soa.refresh} #{soa.retry} #{soa.expire} #{soa.minimum}"
+      end
+
+      nameservers = resolver.getresources(zone, Resolv::DNS::Resource::IN::NS)
+      if nameservers.any? { |ns| ns.name.to_s == @old_hostname }
+        # TODO: replace the correct nameserver
+      end
+
+      begin
+        a_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::A)
+        commands << "update delete #{@old_hostname} A"
+        commands << "update add #{@new_hostname} #{a_record.ttl} A #{a_record.address}"
+      rescue resolv.ResolvError
+        # This is fine
+      end
+
+      begin
+        aaaa_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::AAAA)
+        commands << "update delete #{@old_hostname} AAAA"
+        commands << "update add #{@new_hostname} #{aaaa_record.ttl} A #{aaaa_record.address}"
+      rescue resolv.ResolvError
+        # This is fine
+      end
+
+      commands
+    end
+
     def forward_dns_command(d)
       <<-HEREDOC
 local #{d.dns_server}
@@ -355,6 +388,14 @@ update delete #{d.old_fqdn} A
 update add #{d.new_fqdn} 86400 A #{d.ip}
 send
       HEREDOC
+    end
+
+    def assembled_nsupdate_command(d)
+      commands = ["local #{d.dns_server}", "zone #{d.zone}"]
+      commands += update_zone(d.zone, d.dns_server)
+      commands << "send\n"
+      STDOUT.puts(commands.join("\n"))
+      commands.join("\n")
     end
 
     def reverse_dns_command(d)
@@ -379,10 +420,11 @@ update delete #{reverse_zone} IN NS #{d.old_fqdn}
       # Update SOA record to new hostname; add new A and NS records; delete old A and NS records.
       # Multi-line strings are not indented because Ruby 2.0 doesn't support <<~ heredocs
 
-      STDOUT.puts 'forward...'
-      run_cmd nsupdate_command(forward_dns_command(d), d.key_file)
-      STDOUT.puts 'reverse...'
-      run_cmd nsupdate_command(reverse_dns_command(d), d.key_file)
+      # STDOUT.puts 'forward...'
+      # run_cmd nsupdate_command(forward_dns_command(d), d.key_file)
+      # STDOUT.puts 'reverse...'
+      # run_cmd nsupdate_command(reverse_dns_command(d), d.key_file)
+      run_cmd nsupdate_command(assembled_nsupdate_command(d), d.key_file)
       STDOUT.puts 'updating dynamic zone files...'
       run_cmd('rndc freeze')
       run_cmd('rndc thaw')
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -362,7 +362,7 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
         a_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::A)
         commands << "update delete #{@old_hostname} A"
         commands << "update add #{@new_hostname} #{a_record.ttl} A #{a_record.address}"
-      rescue resolv.ResolvError
+      rescue Resolv::ResolvError
         # This is fine
       end
 
@@ -370,7 +370,7 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
         aaaa_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::AAAA)
         commands << "update delete #{@old_hostname} AAAA"
         commands << "update add #{@new_hostname} #{aaaa_record.ttl} A #{aaaa_record.address}"
-      rescue resolv.ResolvError
+      rescue Resolv::ResolvError
         # This is fine
       end
 
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -349,13 +349,19 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       resolver = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
       commands = []
 
+      soa = resolver.getresource(zone, Resolv::DNS::Resource::IN::SOA)
+      # if soa.mname.to_s == @old_hostname
+        commands << "update add #{zone} #{soa.ttl} SOA #{@new_hostname}. #{soa.rname} #{soa.serial + 1} #{soa.refresh} #{soa.retry} #{soa.expire} #{soa.minimum}"
+      # end
+
       nameservers = resolver.getresources(zone, Resolv::DNS::Resource::IN::NS)
-      if nameservers.any? { |ns| ns.name.to_s == @old_hostname }
+      # if nameservers.any? { |ns| ns.name.to_s == @old_hostname }
         # TODO: replace the correct nameserver
         commands << "update add #{zone}. 3600 IN NS #{@new_hostname}."
         commands << "update delete #{zone}. IN NS #{@old_hostname}"
-      end
+      # end
 
+      return commands if zone.include?('arpa')
       begin
         a_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::A)
         commands << "update delete #{@old_hostname} A"
@@ -370,11 +376,6 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
         commands << "update add #{@new_hostname} #{aaaa_record.ttl} A #{aaaa_record.address}"
       rescue Resolv::ResolvError
         # This is fine
-      end
-
-      soa = resolver.getresource(zone, Resolv::DNS::Resource::IN::SOA)
-      if soa.mname.to_s == @old_hostname
-        commands << "update add #{zone} #{soa.ttl} SOA #{@new_hostname}. #{soa.rname} #{soa.serial + 1} #{soa.refresh} #{soa.retry} #{soa.expire} #{soa.minimum}"
       end
 
       commands
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -105,6 +105,7 @@ module KatelloUtilities
           STDOUT.puts "\nAssembling data for DNS update"
           @new_dns_values = new_dns_values
         rescue StandardError => e # could not reach the DNS server, or something broke while assembling the data
+          STDOUT.puts e
           fail_with_message("Error querying local DNS server for #{@old_hostname}. Make sure the 'named' service is running, or re-run with the --skip-dns option.")
         end
 
@@ -322,13 +323,13 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       new_vals.old_fqdn = @old_hostname
       new_vals.new_fqdn = @new_hostname
 
-      reverse_zones = [@scenario_answers['foreman_proxy']['dns_reverse']].flatten
+      new_vals.reverse_zones = [@scenario_answers['foreman_proxy']['dns_reverse']].flatten
       ip_serial_data = query_dns_server(@old_hostname, new_vals.zone, new_vals.dns_server)
       new_vals.ip = ip_serial_data[:ip]
       serial = ip_serial_data[:serial]
       new_vals.new_serial = serial + 1
 
-      new_vals.reverse_soa_records = reverse_zones.map do |reverse_zone|
+      new_vals.reverse_soa_records = new_vals.reverse_zones.map do |reverse_zone|
         reverse_serial = query_dns_server(@old_hostname, reverse_zone, new_vals.dns_server)[:serial]
         {
           reverse_zone: reverse_zone,
@@ -348,14 +349,11 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       resolver = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
       commands = []
 
-      soa = resolver.getresource(zone, Resolv::DNS::Resource::IN::SOA)
-      if soa.mname.to_s == @old_hostname
-        commands << "update add #{zone} #{soa.ttl} SOA #{@new_hostname}. #{soa.rname} #{soa.serial + 1} #{soa.refresh} #{soa.retry} #{soa.expire} #{soa.minimum}"
-      end
-
       nameservers = resolver.getresources(zone, Resolv::DNS::Resource::IN::NS)
       if nameservers.any? { |ns| ns.name.to_s == @old_hostname }
         # TODO: replace the correct nameserver
+        commands << "update add #{zone}. 3600 IN NS #{@new_hostname}."
+        commands << "update delete #{zone}. IN NS #{@old_hostname}"
       end
 
       begin
@@ -372,6 +370,11 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
         commands << "update add #{@new_hostname} #{aaaa_record.ttl} A #{aaaa_record.address}"
       rescue Resolv::ResolvError
         # This is fine
+      end
+
+      soa = resolver.getresource(zone, Resolv::DNS::Resource::IN::SOA)
+      if soa.mname.to_s == @old_hostname
+        commands << "update add #{zone} #{soa.ttl} SOA #{@new_hostname}. #{soa.rname} #{soa.serial + 1} #{soa.refresh} #{soa.retry} #{soa.expire} #{soa.minimum}"
       end
 
       commands
@@ -394,6 +397,11 @@ send
       commands = ["local #{d.dns_server}", "zone #{d.zone}"]
       commands += update_zone(d.zone, d.dns_server)
       commands << "send\n"
+      d.reverse_zones.each do |zone|
+        commands << "zone #{zone}"
+        commands += update_zone(zone, d.dns_server)
+        commands << "send\n"
+      end
       STDOUT.puts(commands.join("\n"))
       commands.join("\n")
     end
@@ -421,10 +429,8 @@ update delete #{reverse_zone} IN NS #{d.old_fqdn}
       # Multi-line strings are not indented because Ruby 2.0 doesn't support <<~ heredocs
 
       # STDOUT.puts 'forward...'
-      # run_cmd nsupdate_command(forward_dns_command(d), d.key_file)
-      # STDOUT.puts 'reverse...'
-      # run_cmd nsupdate_command(reverse_dns_command(d), d.key_file)
       run_cmd nsupdate_command(assembled_nsupdate_command(d), d.key_file)
+      # run_cmd nsupdate_command(forward_dns_command(d), d.key_file)
       STDOUT.puts 'updating dynamic zone files...'
       run_cmd('rndc freeze')
       run_cmd('rndc thaw')
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -292,15 +292,17 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
 
     def dns_managed?
       @scenario_answers['foreman_proxy']['dns'] &&
-        @scenario_answers['foreman_proxy']['dns_managed']
+        @scenario_answers['foreman_proxy']['dns_managed'] &&
+        %w[nsupdate nsupdate_gss]
+          .include?(@scenario_answers['foreman_proxy']['dns_provider'])
     end
 
     def should_update_dns?
       dns_managed? && !@options[:skip_dns]
     end
 
-    def query_dns_server(fqdn, zone, nameserver_ip)
-      query_dns = Resolv::DNS.new(nameserver: [nameserver_ip], search: [zone], ndots: 1)
+    def query_dns_server(fqdn, nameserver_ip)
+      query_dns = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
       a_record = query_dns.getresource(fqdn, Resolv::DNS::Resource::IN::A)
       ip = a_record.address.to_s if a_record
       soa_record = query_dns.getresource(zone, Resolv::DNS::Resource::IN::SOA)
@@ -322,12 +324,12 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       new_vals.old_fqdn = @old_hostname
       new_vals.new_fqdn = @new_hostname
 
-      ip_serial_data = query_dns_server(@old_hostname, new_vals.zone, new_vals.dns_server)
+      ip_serial_data = query_dns_server(@old_hostname, new_vals.dns_server)
       new_vals.ip = ip_serial_data[:ip]
       serial = ip_serial_data[:serial]
       new_vals.new_serial = serial + 1
 
-      reverse_serial = query_dns_server(@old_hostname, new_vals.reverse_zone, new_vals.dns_server)[:serial]
+      reverse_serial = query_dns_server(@old_hostname, new_vals.dns_server)[:serial]
       new_vals.new_reverse_serial = reverse_serial + 1
       run_cmd('rndc thaw')
 
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -8,6 +8,8 @@ require "uri"
 require "fileutils"
 require "highline/import"
 require "tempfile"
+require "resolv"
+require "ostruct"
 require_relative "helper.rb"
 
 module KatelloUtilities
@@ -87,27 +89,49 @@ module KatelloUtilities
       # Taken from https://www.safaribooksonline.com/library/view/regular-expressions-cookbook/9781449327453/ch08s15.html
       hostname_regex = /^([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,}$/
       unless hostname_regex === @new_hostname
-        self.fail_with_message("#{@new_hostname} is not a valid fully qualified domain name, please use a valid FQDN and try again. " \
+        self.fail_with_message("#{@new_hostname} is not a valid fully qualified domain name. Please use a valid FQDN and try again. " \
                           "No changes have been made to your system.");
       end
 
       unless @foreman_proxy_content
         STDOUT.puts "\nChecking overall health of server"
-        self.run_cmd("hammer ping", [0], "There is a problem with the server, please check 'hammer ping'")
+        self.run_cmd("hammer ping", [0], "There is a problem with the server; please check 'hammer ping'")
         STDOUT.puts "\nChecking credentials"
-        self.hammer_cmd("capsule list", [0], "There is a problem with the credentials, please retry")
+        self.hammer_cmd("capsule list", [0], "There is a problem with the credentials; please retry")
       end
 
-      if @options[:confirm]
-        response = true
-      else
-        STDOUT.print(self.warning_message)
-        response = agree('Proceed with changing your hostname? [y/n]')
+      if should_update_dns?
+        begin
+          STDOUT.puts "\nAssembling data for DNS update"
+          @new_dns_values = new_dns_values
+        rescue StandardError => e # could not reach the DNS server, or something broke while assembling the data
+          fail_with_message("Error querying local DNS server for #{@old_hostname}. Make sure the 'named' service is running, or re-run with the --skip-dns option.")
+        end
+
+        %w[dns_server zone reverse_zone soa_admin_domain key_file old_fqdn new_fqdn ip new_serial new_reverse_serial].each do |field|
+          next if @new_dns_values[field]
+
+          fail_with_message """
+    Error gathering DNS data: couldn't find value for '#{field}'.
+    Make sure /etc/foreman-installer/scenarios.d/#{@options[:scenario]}-answers.yaml
+    'foreman-proxy' section has values for dns_zone, dns_reverse, and keyfile.
+    Make sure A and SOA records are present for #{@old_hostname}.
+"""
+        end
       end
 
-      unless response
-        self.fail_with_message("Hostname change aborted, no changes have been made to your system")
+      return if @options[:confirm]
+      questions = []
+      questions << dns_skip_warning if @options[:skip_dns]
+      questions << warning
+      questions.each do |q|
+        STDOUT.puts
+        STDOUT.puts q[:message]
+        unless agree(q[:question])
+          self.fail_with_message("Hostname change aborted; no changes have been made to your system")
+        end
       end
+      STDOUT.puts "Precheck passed"
     end
 
     def next_steps_message
@@ -140,14 +164,37 @@ module KatelloUtilities
       end
     end
 
-    def warning_message
-      STDOUT.print("***WARNING*** This script will modify your system. " \
-                   "You will need to re-register any #{@options[:program]} clients registered to this system after script completion.")
+    def warning
+      msg = <<-HEREDOC
+***WARNING*** This script will modify your system.
+You will need to re-register any #{@options[:program]} clients registered to this system after script completion.
+      HEREDOC
       unless @foreman_proxy_content
-        STDOUT.print(" #{ @plural_proxy } will have to be re-registered and reinstalled. If you are using custom certificates, you " \
-                     "will have to run the #{@options[:program]}-installer again with custom certificate options after this script completes.")
+        msg << "#{@plural_proxy} will have to be re-registered and reinstalled. If you are using custom certificates,\n" \
+                     "you will have to run the #{@options[:program]}-installer again with custom certificate options after this script completes.\n"
       end
-      STDOUT.print(" Have you taken the necessary precautions (backups, snapshots, etc...)?\n")
+      msg << " Have you taken the necessary precautions (backups, snapshots, etc...)?\n"
+      {
+        message: msg,
+        question: 'Proceed with changing your hostname? [y/n]'
+      }
+    end
+
+    def dns_skip_warning
+      msg = <<-HEREDOC
+***WARNING*** Since the --skip-dns option was specified, nsupdate will NOT be run and
+DNS records will NOT be created for #{@new_hostname}. You will need to do the following manually:
+
+- Remove the A and NS records for #{@old_hostname}
+- Update the SOA record for #{@new_hostname}
+- Create new A and NS records for #{@new_hostname}
+
+If not done, all hosts will lose connection to #{@options[:scenario]} and discovery may not work.
+      HEREDOC
+      {
+        message: msg,
+        question: 'Proceed without updating DNS? [y/n]'
+      }
     end
 
     def hammer_cmd(cmd, exit_codes=[0], message=nil)
@@ -197,7 +244,7 @@ module KatelloUtilities
         opt.banner = "usage: #{@command_prefix}-change-hostname hostname [options]"
         opt.separator  ""
         opt.separator  "example:"
-        opt.separator  "#{@command_prefix}-change-hostname foo.example.com -u admin -p changeme"
+        opt.separator  "#{@command_prefix}-change-hostname newhost.example.com -u admin -p changeme"
         opt.separator  ""
         opt.separator  "options"
 
@@ -211,6 +258,10 @@ module KatelloUtilities
 
         opt.on("-y", "--assumeyes", "answer yes for all questions") do |confirm|
           @options[:confirm] = confirm
+        end
+
+        opt.on("--skip-dns", "skip updating DNS records even if they are managed by #{@options[:scenario]}") do |skip_dns|
+          @options[:skip_dns] = skip_dns
         end
 
         if @foreman_proxy_content
@@ -239,6 +290,102 @@ module KatelloUtilities
       @opt_parser.parse!
     end
 
+    def dns_managed?
+      @scenario_answers['foreman_proxy']['dns'] &&
+        @scenario_answers['foreman_proxy']['dns_managed']
+    end
+
+    def should_update_dns?
+      dns_managed? && !@options[:skip_dns]
+    end
+
+    def query_dns_server(fqdn, zone, nameserver_ip)
+      query_dns = Resolv::DNS.new(nameserver: [nameserver_ip], search: [zone], ndots: 1)
+      a_record = query_dns.getresource(fqdn, Resolv::DNS::Resource::IN::A)
+      ip = a_record.address.to_s if a_record
+      soa_record = query_dns.getresource(zone, Resolv::DNS::Resource::IN::SOA)
+      serial = soa_record.serial if soa_record
+      {
+        ip: ip,
+        serial: serial
+      }
+    end
+
+    def new_dns_values
+      new_vals = OpenStruct.new
+
+      new_vals.dns_server = @scenario_answers['foreman_proxy']['dns_server']
+      new_vals.zone = @scenario_answers['foreman_proxy']['dns_zone']
+      new_vals.reverse_zone = [@scenario_answers['foreman_proxy']['dns_reverse']].flatten.first
+      new_vals.soa_admin_domain = "root.#{new_vals.zone}"
+      new_vals.key_file = @scenario_answers['foreman_proxy']['keyfile']
+      new_vals.old_fqdn = @old_hostname
+      new_vals.new_fqdn = @new_hostname
+
+      ip_serial_data = query_dns_server(@old_hostname, new_vals.zone, new_vals.dns_server)
+      new_vals.ip = ip_serial_data[:ip]
+      serial = ip_serial_data[:serial]
+      new_vals.new_serial = serial + 1
+
+      reverse_serial = query_dns_server(@old_hostname, new_vals.reverse_zone, new_vals.dns_server)[:serial]
+      new_vals.new_reverse_serial = reverse_serial + 1
+      run_cmd('rndc thaw')
+
+      new_vals
+    end
+
+    def nsupdate_command(dns_entries, key_file)
+      "echo -e \"#{dns_entries}\" | nsupdate -l -k #{key_file}"
+    end
+
+    def update_dns_records(d)
+      STDOUT.puts 'updating DNS records:'
+      # Use nsupdate to update DNS records
+      # Update SOA record to new hostname; add new A and NS records; delete old A and NS records.
+      # Multi-line strings are not indented because Ruby 2.0 doesn't support <<~ heredocs
+
+      forward_dns_command = <<-HEREDOC
+local #{d.dns_server}
+zone #{d.zone}
+update add #{d.zone} 10800 SOA #{d.new_fqdn} #{d.soa_admin_domain}. #{d.new_serial} 86400 3600 604800 3600
+update add #{d.zone}. 3600 IN NS #{d.new_fqdn}.
+update delete #{d.zone}. IN NS #{d.old_fqdn}
+update delete #{d.old_fqdn} A
+update add #{d.new_fqdn} 86400 A #{d.ip}
+send
+      HEREDOC
+
+      reverse_dns_command = <<-HEREDOC
+local #{d.dns_server}
+zone #{d.reverse_zone}
+update add #{d.reverse_zone} 10800 SOA #{d.new_fqdn} root.#{d.reverse_zone}. #{d.new_reverse_serial} 86400 3600 604800 3600
+update add #{d.reverse_zone} 3600 IN NS #{d.new_fqdn}
+update delete #{d.reverse_zone} IN NS #{d.old_fqdn}
+send
+      HEREDOC
+
+      STDOUT.puts 'forward...'
+      run_cmd nsupdate_command(forward_dns_command, d.key_file)
+      STDOUT.puts 'reverse...'
+      run_cmd nsupdate_command(reverse_dns_command, d.key_file)
+      STDOUT.puts 'updating dynamic zone files...'
+      run_cmd('rndc freeze')
+      run_cmd('rndc thaw')
+      STDOUT.puts 'DNS records updated'
+    end
+
+    def restore_last_scenario_yaml
+      STDOUT.puts 'restoring last_scenario.yaml'
+      if File.exist?(last_scenario_yaml)
+        run_cmd("cp #{temp_last_scenario_yaml.path} #{last_scenario_yaml}")
+      else
+        # if the installer failed early the last_scenario symlink won't exist
+        scenario_path = "#{scenarios_path}/#{@options[:scenario]}.yaml"
+        run_cmd("cp #{temp_last_scenario_yaml.path} #{scenario_path}")
+        File.symlink(scenario_path, last_scenario_yaml)
+      end
+    end
+
     def run
       raise 'Must run as root' unless Process.uid == 0
 
@@ -263,6 +410,23 @@ module KatelloUtilities
 
       self.precheck
 
+      if should_update_dns?
+        update_dns_records(@new_dns_values)
+      end
+
+      STDOUT.puts "updating hostname in /etc/hostname"
+      self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' /etc/hostname")
+      STDOUT.puts "setting hostname"
+      self.run_cmd("hostnamectl set-hostname #{@new_hostname}")
+
+      # override environment variable (won't be updated until bash login)
+      ENV['HOSTNAME'] = @new_hostname
+
+      STDOUT.puts "checking if hostname was changed"
+      if self.get_hostname != @new_hostname
+        self.fail_with_message("The new hostname was not changed successfully, exiting script")
+      end
+
       unless @foreman_proxy_content
         STDOUT.puts "\nUpdating default #{@proxy}"
         proxy_id = self.get_default_proxy_id
@@ -278,19 +442,6 @@ module KatelloUtilities
           new_path = new_path.to_s
           hammer_cmd("medium update --id #{medium['Id']} --path #{new_path}")
         end
-      end
-
-      STDOUT.puts "updating hostname in /etc/hostname"
-      self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' /etc/hostname")
-      STDOUT.puts "setting hostname"
-      self.run_cmd("hostnamectl set-hostname #{@new_hostname}")
-
-      # override environment variable (won't be updated until bash login)
-      ENV['HOSTNAME'] = @new_hostname
-
-      STDOUT.puts "checking if hostname was changed"
-      if self.get_hostname != @new_hostname
-        self.fail_with_message("The new hostname was not changed successfully, exiting script")
       end
 
       STDOUT.puts "stopping services"
@@ -345,6 +496,10 @@ module KatelloUtilities
       STDOUT.puts "updating hostname in foreman installer scenarios"
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' #{scenarios_path}/*.yaml")
 
+      STDOUT.puts "updating hostname in hammer configuration"
+      self.run_cmd("sed -i.bak -e 's/#{@old_hostname}/#{@new_hostname}/g' #{hammer_root_config_path}/*.yml")
+      self.run_cmd("sed -i.bak -e 's/#{@old_hostname}/#{@new_hostname}/g' #{hammer_config_path}/*.yml")
+
       if File.exist?(last_scenario_yaml)
         STDOUT.puts 'backing up last_scenario.yaml'
         temp_last_scenario_yaml = Tempfile.new('last_scenario')
@@ -377,15 +532,7 @@ module KatelloUtilities
       run_cmd(installer, [0], installer_failure_message) do |result, success|
         if temp_last_scenario_yaml && temp_last_scenario_yaml.path
           unless success
-            STDOUT.puts 'restoring last_scenario.yaml'
-            if File.exist?(last_scenario_yaml)
-              run_cmd("cp #{temp_last_scenario_yaml.path} #{last_scenario_yaml}")
-            else
-              # if the installer failed early the last_scenario symlink won't exist
-              scenario_path = "#{scenarios_path}/#{@options[:scenario]}.yaml"
-              run_cmd("cp #{temp_last_scenario_yaml.path} #{scenario_path}")
-              File.symlink(scenario_path, last_scenario_yaml)
-            end
+            restore_last_scenario_yaml
           end
           STDOUT.puts 'cleaning up temporary files'
           temp_last_scenario_yaml.unlink
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -332,8 +332,7 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       "echo -e \"#{dns_entries}\" | nsupdate -l -k #{key_file}"
     end
 
-    def update_zone(zone, nameserver_ip)
-      resolver = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
+    def update_zone(zone, nameserver_ip, resolver)
       commands = []
 
       soa = resolver.getresource(zone, Resolv::DNS::Resource::IN::SOA)
@@ -370,12 +369,13 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
     end
 
     def assembled_nsupdate_command(d)
+      resolver = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
       commands = ["local #{d.dns_server}", "zone #{d.zone}"]
-      commands += update_zone(d.zone, d.dns_server)
+      commands += update_zone(d.zone, d.dns_server, resolver)
       commands << "send\n"
       d.reverse_zones.each do |zone|
         commands << "zone #{zone}"
-        commands += update_zone(zone, d.dns_server)
+        commands += update_zone(zone, d.dns_server, resolver)
         commands << "send\n"
       end
       STDOUT.puts(commands.join("\n"))
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -108,7 +108,7 @@ module KatelloUtilities
           fail_with_message("Error querying local DNS server for #{@old_hostname}. Make sure the 'named' service is running, or re-run with the --skip-dns option.")
         end
 
-        %w[dns_server zone reverse_zone soa_admin_domain key_file old_fqdn new_fqdn ip new_serial new_reverse_serial].each do |field|
+        %w[dns_server zone reverse_zones soa_admin_domain key_file old_fqdn new_fqdn ip new_serial reverse_soa_records].each do |field|
           next if @new_dns_values[field]
 
           fail_with_message """
@@ -301,7 +301,7 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       dns_managed? && !@options[:skip_dns]
     end
 
-    def query_dns_server(fqdn, nameserver_ip)
+    def query_dns_server(fqdn, zone, nameserver_ip)
       query_dns = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
       a_record = query_dns.getresource(fqdn, Resolv::DNS::Resource::IN::A)
       ip = a_record.address.to_s if a_record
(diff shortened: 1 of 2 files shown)


chris1984:
Starting to test

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +345,53 @@ def run
...
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' #{scenarios_path}/*.yaml")
 
+      if @scenario_answers['foreman_proxy']['dns']
+        STDOUT.puts 'gathering DNS data...'
+        local_ip = '127.0.0.1'
+        zone = @scenario_answers['foreman_proxy']['dns_zone']
+        old_fqdn = @old_hostname
+        soa_admin_domain = "root.#{zone}"
+        zone_file = "/var/named/dynamic/db.#{zone}"
+        zone_file_data = File.read(zone_file).split("\n")
```
ekohl: Rather than parsing a potentially out of date file, I'd use an AXFR query to the server. If you want to go this way, you should first use `rndc freeze` and later thaw to allow updates again.

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +345,53 @@ def run
...
+        a_record = zone_file_data.find { |line| line.include?(' A ') }
+        existing_ip = a_record.split(' A ').last if a_record
+        serial_line = zone_file_data.find { |line| line.include?('Serial') }
```
ekohl: Is it needed to increase the serial when using nsupdate? Wouldn't it be safer to do a query for the SOA record and rely on the actual structure of the record rather than relying on a comment?

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +345,53 @@ def run
...
+        existing_serial = serial_line.split('Serial').first.to_i if serial_line
+        new_serial = (existing_serial + 1).to_s if existing_serial
+        new_fqdn = @new_hostname
+        key_file = '/etc/rndc.key'
```
ekohl: The foreman proxy config also has the correct location.

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +345,53 @@ def run
...
+          new_fqdn: new_fqdn,
+          key_file: key_file,
+          reverse_zone: reverse_zone
+        }
+        dns_data.each { |k, v| raise "Error gathering DNS data: value for '#{k}' is missing" unless v }
+
+        STDOUT.puts 'updating DNS records:'
+        # Use nsupdate to update DNS records
+        # Update SOA record to new hostname; add new A and NS records; delete old A and NS records.
+        forward_dns_command = "echo -e \"local #{local_ip}\n zone #{zone}\n update add #{zone} 10800 SOA #{new_fqdn} #{soa_admin_domain}. #{new_serial} 86400 3600 604800 3600\n update add #{zone}. 3600 IN NS #{new_fqdn}.\n update delete #{zone}. IN NS #{old_fqdn}\n update delete #{old_fqdn} A\n update add #{new_fqdn} 86400 A #{existing_ip}\n send\n\" | nsupdate -l -k #{key_file}"
```
ekohl: Would it make sense to use a multi line string (heredoc) for readability?

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl nsupdate does update the serial for you, except it doesn't seem to when you're updating the SOA record.

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +345,53 @@ def run
       STDOUT.puts "updating hostname in foreman installer scenarios"
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' #{scenarios_path}/*.yaml")
 
+      if @scenario_answers['foreman_proxy']['dns']
```
johnpmitsch: this file isn't the best about this, but it does seem like this could be refactored into method(s) called here for readability. maybe even 3 or more methods. My first instinct is to separate into: `get_dns_data` , `forward_dns_command`, `reverse_dns_command`, and `update_dns_records`,  or even have one `dns_command` that takes a large string to echo into nsupdate and create additional methods returning those strings for forward and reverse (in heredoc or %Q{foo} to split into multiple lines like ewoud suggested). 

Hopefully this make sense!

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +345,53 @@ def run
...
+        STDOUT.puts 'gathering DNS data...'
+        local_ip = '127.0.0.1'
+        zone = @scenario_answers['foreman_proxy']['dns_zone']
+        old_fqdn = @old_hostname
```
johnpmitsch: why reassign @old_hostname?

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +345,53 @@ def run
...
+        soa_admin_domain = "root.#{zone}"
+        zone_file = "/var/named/dynamic/db.#{zone}"
+        zone_file_data = File.read(zone_file).split("\n")
+        a_record = zone_file_data.find { |line| line.include?(' A ') }
+        existing_ip = a_record.split(' A ').last if a_record
+        serial_line = zone_file_data.find { |line| line.include?('Serial') }
+        existing_serial = serial_line.split('Serial').first.to_i if serial_line
+        new_serial = (existing_serial + 1).to_s if existing_serial
+        new_fqdn = @new_hostname
```
johnpmitsch: Same question here, why reassign @new_hostname?

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +345,53 @@ def run
...
+      if @scenario_answers['foreman_proxy']['dns']
+        STDOUT.puts 'gathering DNS data...'
+        local_ip = '127.0.0.1'
+        zone = @scenario_answers['foreman_proxy']['dns_zone']
+        old_fqdn = @old_hostname
+        soa_admin_domain = "root.#{zone}"
+        zone_file = "/var/named/dynamic/db.#{zone}"
+        zone_file_data = File.read(zone_file).split("\n")
+        a_record = zone_file_data.find { |line| line.include?(' A ') }
+        existing_ip = a_record.split(' A ').last if a_record
```
johnpmitsch: will `existing_ip` not exist if there is no `a_record`? wouldn't it cause errors when it is referenced below if so?

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +345,53 @@ def run
...
+        serial_line = zone_file_data.find { |line| line.include?('Serial') }
+        existing_serial = serial_line.split('Serial').first.to_i if serial_line
+        new_serial = (existing_serial + 1).to_s if existing_serial
+        new_fqdn = @new_hostname
+        key_file = '/etc/rndc.key'
+        reverse_zone = @scenario_answers['foreman_proxy']['dns_reverse'].first
+
+        dns_data = {
+          local_ip: local_ip,
```
johnpmitsch: There certainly is no harm in assigning and referencing here, but seems like doing `local_ip: '127.0.0.1'` will be fine if its not referenced anywhere else. Same goes for some other values in the hash. Building this hash feels like a good candidate for its own method

packages/katello/katello/hostname-change.rb
jeremylenz: @johnpmitsch  It's just for readability, because I wanted to be very clear about what values I'm giving to nsupdate.  To me "hostname" could mean either the FQDN (a.b.c.d) or just the basic hostname (a), so calling it "fqdn" removes the ambiguity.

packages/katello/katello/hostname-change.rb
jeremylenz: @johnpmitsch My idea was that I'll gather every value I'm about to give to nsupdate, and make sure all of them are present.  `local_ip` will always be present, so I guess I included it just for completeness.  It's not strictly necessary here I suppose.

packages/katello/katello/hostname-change.rb
jeremylenz: @johnpmitsch thanks, updated!  Couldn't figure out an elegant way to pass data to `forward_dns_command` and `reverse_dns_command`, so keeping them together in `update_dns_records` for now.  (If only Ruby had JS ES6 destructuring ;)

And I had to use `<<-` heredocs because Ruby 2.0 doesn't support `<<~` :cry:

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl Thanks, updated.  I'm now querying first, and parsing the file (with the freeze and thaw) as a fallback.

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +240,103 @@ def setup_opt_parser
...
+echo -e "local #{local_ip}
+zone #{zone}
+update add #{zone} 10800 SOA #{new_fqdn} #{soa_admin_domain}. #{new_serial} 86400 3600 604800 3600
+update add #{zone}. 3600 IN NS #{new_fqdn}.
+update delete #{zone}. IN NS #{old_fqdn}
+update delete #{old_fqdn} A
+update add #{new_fqdn} 86400 A #{existing_ip}
+send
+" | nsupdate -l -k #{key_file}
+      HEREDOC
```
johnpmitsch: Some of this comes down to coding style, so this is just a suggestion, but what about just referencing the hash and creating separate methods?

```ruby
def forward_dns_command(dns_data)
<<-HEREDOC
echo -e "local #{dns_data[:local_ip]}
zone #{dns_data[:zone]}
update add #{dns_data[:zone]} 10800 SOA #{dns_data[:new_fqdn]} #{dns_data[:soa_admin_domain]}. #{dns_data[:new_serial]} 86400 3600 604800 3600
update add #{dns_data[:zone]}. 3600 IN NS #{dns_data[:new_fqdn]}.
update delete #{dns_data[:zone]}. IN NS #{dns_data[:old_fqdn]}
update delete #{dns_data[:old_fqdn]} A
update add #{dns_data[:new_fqdn]} 86400 A #{dns_data[:existing_ip]}
send
" | nsupdate -l -k #{dns_data[:key_file]}
      HEREDOC
end
```

and could even dry up the nsupdate flags in case they need updating later.

```ruby
def nsupdate_command(dns_entries, keyfile)
    "echo -e \"#{dns_entries}\" | nsupdate -l -k #{keyfile}"
end

def forward_dns_command(dns_data)
  dns_entries = <<-HEREDOC
local #{dns_data[:local_ip]}
zone #{dns_data[:zone]}
update add #{dns_data[:zone]} 10800 SOA #{dns_data[:new_fqdn]} #{dns_data[:soa_admin_domain]}. #{dns_data[:new_serial]} 86400 3600 604800 3600
update add #{dns_data[:zone]}. 3600 IN NS #{dns_data[:new_fqdn]}.
update delete #{dns_data[:zone]}. IN NS #{dns_data[:old_fqdn]}
update delete #{dns_data[:old_fqdn]} A
update add #{dns_data[:new_fqdn]} 86400 A #{dns_data[:existing_ip]}
send
      HEREDOC
  nsupdate_command(dns_entries, dns_data[:keyfile])
end

# do the same for reverse_dns_command
 ```

Comment (jeremylenz) [reply packages/katello/katello/hostname-change.rb]: @johnpmitsch yeah I just thought it was really ugly to have `dns_data` everywhere.   I feel like adding that 7 characters dozens of times kind of cancels out the benefit of separating out the methods.  But I'll change it if you feel strongly about it.

Comment (johnpmitsch) [reply packages/katello/katello/hostname-change.rb]: @jeremylenz It would be my preference for readability, but I don't  see any technical or best-practices reason to do one over the other and we don't enforce method limits in this repo, so the final choice is up to you!

Comment (johnpmitsch) [reply packages/katello/katello/hostname-change.rb]: @jeremylenz either way, I would consider DRYing up the nsupdate flags. I seem to remember a bug in k-c-hostname where we only updated cli flags in once place and missed another

Comment (jeremylenz) [reply packages/katello/katello/hostname-change.rb]: @johnpmitsch Added the `nsupdate_command` method per your suggestion.  Also separated out `query_dns_server` and `parse_zone_file` methods.

Also, I realized I needed to handle the situation where `Serial` is different for the forward and reverse zones, so we're handling that now too.

**Comment by chris1984** [comment]
Starting to retest

File: packages/katello/katello/hostname-change.rb [hunk 11]
```diff
@@ -239,6 +240,135 @@ def setup_opt_parser
...
+      zone_file_data = File.read(zone_file).split("\n")
+      # find the character A enclosed by any whitespace chars
+      a_record = zone_file_data.find { |line| !(line =~ /\sA\s/).nil? }
+      existing_ip = a_record.split('A').last.strip if a_record
+      serial_line = zone_file_data.find { |line| line.downcase.include?('serial') }
+      existing_serial = serial_line.split('Serial').first.to_i if serial_line
+      new_serial = (existing_serial + 1).to_s if existing_serial
+      {
+        existing_ip: existing_ip,
+        new_serial: new_serial
```
jturel: Ditto feedback on `query_dns_server`

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +240,135 @@ def setup_opt_parser
...
+      STDOUT.puts 'gathering DNS data...'
+      local_ip = '127.0.0.1'
+      zone = @scenario_answers['foreman_proxy']['dns_zone']
+      reverse_zone = @scenario_answers['foreman_proxy']['dns_reverse']
+      if reverse_zone.is_a?(Array)
+        reverse_zone = reverse_zone.first
+      end
+      old_fqdn = @old_hostname
+      soa_admin_domain = "root.#{zone}"
+      new_fqdn = @new_hostname
```
jturel: If you really want to do this, I'd recommend supplying old_fqdn and new_fqdn as arguments to the method, or just reference the instance variables

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +240,135 @@ def setup_opt_parser
...
+      key_file = @scenario_answers['foreman_proxy']['keyfile']
+      begin
+        STDOUT.puts 'querying local DNS server...'
+        ip_serial_data = query_dns_server(old_fqdn, zone, local_ip)
+        existing_ip = ip_serial_data[:existing_ip]
+        new_serial = ip_serial_data[:new_serial]
```
jturel: seems like right around here is where we should create the new serial based on the old one

packages/katello/katello/hostname-change.rb
```diff
@@ -345,6 +475,11 @@ def run
       STDOUT.puts "updating hostname in foreman installer scenarios"
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' #{scenarios_path}/*.yaml")
 
+      if @scenario_answers['foreman_proxy']['dns']
+        dns_data = get_dns_data
+        update_dns_records(dns_data)
```
jturel: This variable assignment seems unnecessary

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +240,135 @@ def setup_opt_parser
...
+    end
+
+    def update_dns_records(dns_data = {})
+      dns_data.each do |key, value|
+        unless value
+          raise "Error gathering DNS data: couldn't find value for '#{key}'"
+        end
+      end
+
+      local_ip = dns_data[:local_ip]
```
jturel: I'd like to see all these variable reassignments go away. You can do that by making `forward_dns_command` a method that takes required keyword arguments with a pattern like this:

```
hsh = {                                                                                                                                                       
  foo: 'its foo!',                                                                                                                                            
  bar: 'its bar!',                                                                                                                                            
}                                                                                                                                                             
                                                                                                                                                              
def expand_hash(foo:, bar:)                                                                                                                                   
  puts foo                                                                                                                                                    
  puts bar                                                                                                                                                    
end                                                                                                                                                           
                                                                                                                                                              
expand_hash(**hsh) 
```

outputs:

```
its foo!
its bar!
```

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +240,135 @@ def setup_opt_parser
...
     end
 
+    def query_dns_server(fqdn, zone, local_ip)
+      query_dns = Resolv::DNS.new(nameserver: [local_ip], search: [zone], ndots: 1)
+      a_record = query_dns.getresource(fqdn, Resolv::DNS::Resource::IN::A)
+      existing_ip = a_record.address.to_s if a_record
+      soa_record = query_dns.getresource(zone, Resolv::DNS::Resource::IN::SOA)
+      existing_serial = soa_record.serial if soa_record
+      new_serial = (existing_serial + 1).to_s if existing_serial
+      {
```
jturel: I don't think this method should care about the 'new serial' and can be generated by whoever uses the output of this method.

Maybe the response should be:

```
{
 ip: a_record&.address,
 serial: soa_record&.serial
}
```

With the safe navigation operator we can safely call the instance methods even if the values of those objects is nil - but they do need to be defined. I think it's better having fewer if statements - therefore fewer execution paths thru the code. Omitted `existing` prefix be cause we can already assume that

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +240,135 @@ def setup_opt_parser
...
+      serial_line = zone_file_data.find { |line| line.downcase.include?('serial') }
+      existing_serial = serial_line.split('Serial').first.to_i if serial_line
+      new_serial = (existing_serial + 1).to_s if existing_serial
+      {
+        existing_ip: existing_ip,
+        new_serial: new_serial
+      }
+    end
+
+    def get_dns_data
```
jturel: rubocop would complain about a method called `get_<whatever>` since it could just be called `whatever`.

But I think the method name should be changed to reflect the fact that it's data representing the future state of the dns records, and not just the current state.

I'm having trouble coming up with a clearer name ...  :thinking:

packages/katello/katello/hostname-change.rb
jeremylenz: :open_mouth:  This is what I was looking for!

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel The `sat-deploy` box I tested on only had Ruby 2.0, which means the safe navigation operator isn't available :cry:

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel It will be once we come up with a better name for `get_dns_data` :yum:

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel Looks like keyword arguments with default values were introduced in Ruby 2.0, but required keyword args like you outlined didn't [arrive](https://thoughtbot.com/blog/ruby-2-keyword-arguments) until Ruby 2.1.  :sob:

packages/katello/katello/hostname-change.rb
jturel: Dang! My feedback on the output response is still relevant though Ditto for parse_zone_file

chris1984:
@jeremylenz @jturel let me know when it is good to retest

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel I started to make this change but realized I feel like it doesn't make as much sense as one would think.  I have to create both `serial` and `reverse_serial` values for both `query_dns_server` and `parse_zone_file` methods, and moving that incrementation to `get_dns_data` creates a lot of repeated code.

jeremylenz: specifically, instead of getting the incremented value directly from the method I would have to return the unincremented value, then increment it in 2 places in `get_dns_data`.  It would add 2 lines to `get_dns_data` and subtract 1 line from each of the other methods... doesn't seem worth it to me.

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel Renamed to `new_dns_values`

packages/katello/katello/hostname-change.rb
jturel: Having a clear separation of responsibility is more important than saving a few lines of code. `get_dns_data` is the method concerned with determining the future state of the DNS so it should know/care about incrementing the serial. If we wanted to change how we generate the new serial then we'd have to update two methods - that's an increased risk of breaking something and the code is still duplicated in two places.

jeremylenz:
* renamed `existing_ip` to `ip`
* renamed `get_dns_data` to `new_dns_values`
* eliminated `old_fqdn` and `new_fqdn` within `new_dns_values` and use existing instance variables instead
* eliminated unnecessary variable assignment

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel Hadn't thought of it that way, you're right.  Updated.

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +240,135 @@ def setup_opt_parser
...
+      local_ip = new_dns_values[:local_ip]
+      zone = new_dns_values[:zone]
+      old_fqdn = new_dns_values[:old_fqdn]
+      soa_admin_domain = new_dns_values[:soa_admin_domain]
+      ip = new_dns_values[:ip]
+      new_serial = new_dns_values[:new_serial]
+      new_reverse_serial = new_dns_values[:new_reverse_serial]
+      new_fqdn = new_dns_values[:new_fqdn]
+      key_file = new_dns_values[:key_file]
+      reverse_zone = new_dns_values[:reverse_zone]
```
johnpmitsch: @jeremylenz I'm having a hard time coming to terms with these lines, while its valid, it feels like a lot of work to get information from an already structured hash.  Are there any other alternatives?

How about using an [openstruct](https://docs.ruby-lang.org/en/2.0.0/OpenStruct.html) if you are concerned about the amount of characters in [my other suggestion](https://github.com/theforeman/foreman-packaging/pull/4706#discussion_r378980457)?
```ruby
d = OpenStruct.new(new_dns_values)
## then refer to the struct
d.key_file
d.new_fqdn
```
ideally `d` is as locally scoped as possible, breaking into individual methods like my first example. Open to other ideas here though, that is just first the one comes to mind that works with ruby 2.0.

jeremylenz: @johnpmitsch this looks like a good solution!  Updated.

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +241,126 @@ def setup_opt_parser
...
+    end
+
+    def update_dns_records(new_dns_values = {})
+      new_dns_values.each do |key, value|
+        unless value
+          raise "Error gathering DNS data: couldn't find value for '#{key}'"
+        end
+      end
+
+      d = OpenStruct.new(new_dns_values)
```
jturel: nice

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +241,126 @@ def setup_opt_parser
...
+      soa_admin_domain = "root.#{zone}"
+      key_file = @scenario_answers['foreman_proxy']['keyfile']
+      begin
+        STDOUT.puts 'querying local DNS server...'
+        ip_serial_data = query_dns_server(@old_hostname, zone, local_ip)
+        ip = ip_serial_data[:ip]
+        serial = ip_serial_data[:serial]
+        new_serial = (serial + 1).to_s if serial
+        reverse_serial = query_dns_server(@old_hostname, reverse_zone, local_ip)[:serial]
+        new_reverse_serial = (reverse_serial + 1).to_s if reverse_serial
```
jturel: why are the new serials cast to strings?

jeremylenz: @jturel they eventually become part of the nsupdate command which is a string; I just felt better explicitly converting it since all the other values are strings, but if you think I should remove it I don't really feel strongly either way.

jturel: I think it's unnecessary since the interpolation where it's used in the HEREDOC will cast it to a string. Just wanted to see if you'd run into something that made it required. IMO - it should be removed

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +241,126 @@ def setup_opt_parser
...
+        serial = ip_serial_data[:serial]
+        new_serial = serial + 1 if serial
+        reverse_serial = parse_zone_file("/var/named/dynamic/db.#{reverse_zone}")[:serial]
+        new_reverse_serial = reverse_serial + 1 if reverse_serial
+      ensure
+        run_cmd('rndc thaw')
+      end
+
+
+      {
```
jturel: So now that you're using the OpenStruct in `update_dns_records` I think we can change this a little to tidy up a bit and avoid creating this hash in such an explicit way

In this method you can create the struct and assign its values and return it - and avoid the OpenStruct.new in `update_dns_records`

```
new_dns = OpenStruct.new

new_dns.new_serial = ....
```

jturel: Alternatively, just create the hash at the start of the method and populate it along the way, and keep the OpenStruct cast in the other method

johnpmitsch: > In this method you can create the struct and assign its values and return it - and avoid the OpenStruct.new in update_dns_records

> Alternatively, just create the hash at the start of the method and populate it along the way, and keep the OpenStruct cast in the other method

+1 to either of these

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +241,126 @@ def setup_opt_parser
...
+    end
+
+    def nsupdate_command(dns_entries, key_file)
+      "echo -e \"#{dns_entries}\" | nsupdate -l -k #{key_file}"
+    end
+
+    def update_dns_records(new_dns_values = {})
+      new_dns_values.each do |key, value|
+        unless value
+          raise "Error gathering DNS data: couldn't find value for '#{key}'"
```
johnpmitsch: its nice to have an actionable when things go wrong. What do you expect the user to do here or where should they check?

jeremylenz: @johnpmitsch Thanks, updated

jeremylenz:
Removed the file parsing fallback since `nsupdate` won't work anyway if `named` service isn't working.

packages/katello/katello/hostname-change.rb
```diff
@@ -213,6 +226,10 @@ def setup_opt_parser
           @options[:confirm] = confirm
         end
 
+        opt.on("--skip-dns", "skip updating DNS records") do |skip_dns|
```
jturel: I think this should come with some kind of warning text.

If foreman is configured to manage DNS but is not able to pass the pre-check, or this flag is given, something is probably really wrong. Although I suggested adding this flag, I think there could be a very good argument to not provide it. 

@ekohl - any input on this?

johnpmitsch: I would advocate for keeping it, I'm sure we could rationalize that there would be no way it should be needed, but I've found in actual user environments it helps to have an way to workaround issues even if its just for debugging purposes or if someone has a very opinionated "special" setup and will do the updating themselves. However, I think we should be very explicit about what the option is and when to use it.

The helper text could be more specific about when it would be applicable. `"skip updating DNS records even if they are managed by #{@default_program}".` I wonder if we should only show the option if `@scenario_answers['foreman_proxy']['dns']` is truthy as well?

And a confirmation prompt as @jturel suggested if they use that option and have dns managed. A buy-in `y/n` does help us sleep better at night putting in something that is not the recommended happy path.

johnpmitsch:
I can confirm the script still works fine on a Katello nightly box without DNS management (no regressions introduced). I would like the test it on a box with managed DNS or hear confirmation from someone else who tested that part. 

We have to figure out exactly how the skip-dns flag will be handled (I'll wait on this before ACKing), but otherwise the code looks good :+1:

jeremylenz:
updated:
* Added new `should_update_dns?` method correcting the decision logic (thanks @chris1984)
* Moved actual hostname change logic up so that users can try again after failing
* Added hammer config update so that `hammer ping` continues to work after hostname change

packages/katello/katello/hostname-change.rb
jeremylenz: If we do keep it @johnpmitsch I will update the message per your suggestion, thanks.

Holding off for now on "keep the flag?" and "add the confirm?" until I hear more opinions.

jturel: +1 keep +1 add confirmation with a sternly worded message that they're going into uncharted waters preferably mentioning some of the things that could fail to work after the script completes

jeremylenz:
@jturel updated with strongly worded message

packages/katello/katello/hostname-change.rb
jeremylenz: @johnpmitsch Message updated
```
options
    -u, --username username          admin username (required)
    -p, --password password          admin password (required)
    -y, --assumeyes                  answer yes for all questions
        --skip-dns                   skip updating DNS records even if they are managed by satellite
    -h, --help                       help
```

johnpmitsch:
All set to merge from me. I saw some comments from @ekohl and just want to make sure his concerns are addressed before merging. 

@ekohl can you confirm?

packages/katello/katello/hostname-change.rb
```diff
@@ -87,27 +89,40 @@ def precheck
...
 
-      if @options[:confirm]
-        response = true
-      else
-        STDOUT.print(self.warning_message)
-        response = agree('Proceed with changing your hostname? [y/n]')
+      if should_update_dns?
+        begin
+          STDOUT.puts "\nRunning test DNS query"
+          query_dns = Resolv::DNS.new(nameserver: ['127.0.0.1'], search: [], ndots: 1)
```
ekohl: In the answers file you can find the actual nameserver used by nsupdate (`answers['foreman_proxy']['dns_server']`). There's also a `query_dns_server` helper that duplicates the constructor. Should there be a `dns_resolver` helper?

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +281,116 @@ def setup_opt_parser
...
+      @scenario_answers['foreman_proxy']['dns'] &&
+        @scenario_answers['foreman_proxy']['dns_managed']
+    end
+
+    def should_update_dns?
+      dns_managed? && !@options[:skip_dns]
+    end
+
+    def query_dns_server(fqdn, zone, local_ip)
+      query_dns = Resolv::DNS.new(nameserver: [local_ip], search: [zone], ndots: 1)
```
ekohl: I think we should avoid `search` and just assume `fqdn` is absolute.

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +281,116 @@ def setup_opt_parser
...
+    end
+
+    def new_dns_values
+      STDOUT.puts 'gathering DNS data...'
+      new_vals = OpenStruct.new
+      new_vals.local_ip = '127.0.0.1'
+      new_vals.zone = @scenario_answers['foreman_proxy']['dns_zone']
+      new_vals.reverse_zone = @scenario_answers['foreman_proxy']['dns_reverse']
+      if new_vals.reverse_zone.is_a?(Array)
+        new_vals.reverse_zone = new_vals.reverse_zone.first
```
ekohl: I'm never happy with the pattern that you ignore the rest of an array. Should we instead loop over every one?

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +281,116 @@ def setup_opt_parser
...
+      new_vals.new_serial = serial + 1 if serial
+      reverse_serial = query_dns_server(@old_hostname, new_vals.reverse_zone, new_vals.local_ip)[:serial]
+      new_vals.new_reverse_serial = reverse_serial + 1 if reverse_serial
+      run_cmd('rndc thaw')
+
+      new_vals
+    end
+
+    def nsupdate_command(dns_entries, key_file)
+      "echo -e \"#{dns_entries}\" | nsupdate -l -k #{key_file}"
```
ekohl: Can you rely on `open3` and pass in the stdin directly? `echo` might transform things.

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +281,116 @@ def setup_opt_parser
...
+      new_vals.new_fqdn = @new_hostname
+
+      STDOUT.puts 'querying local DNS server...'
+      ip_serial_data = query_dns_server(@old_hostname, new_vals.zone, new_vals.local_ip)
+      new_vals.ip = ip_serial_data[:ip]
+      serial = ip_serial_data[:serial]
+      new_vals.new_serial = serial + 1 if serial
+      reverse_serial = query_dns_server(@old_hostname, new_vals.reverse_zone, new_vals.local_ip)[:serial]
+      new_vals.new_reverse_serial = reverse_serial + 1 if reverse_serial
+      run_cmd('rndc thaw')
```
ekohl: Every zone update thaws, so is this needed?

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +281,116 @@ def setup_opt_parser
...
+      STDOUT.puts 'updating dynamic zone files...'
+      run_cmd('rndc freeze')
+      run_cmd('rndc thaw')
+      STDOUT.puts 'DNS records updated'
+    end
+
+    def restore_last_scenario_yaml
+      STDOUT.puts 'restoring last_scenario.yaml'
+      if File.exist?(last_scenario_yaml)
+        run_cmd("cp #{@temp_last_scenario_yaml.path} #{last_scenario_yaml}")
```
ekohl: Not sure what the convention is, but should this use native ruby methods to copy a file instead?

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl I think it is needed; in my experience `nsupdate` commands came back `REJECTED` if I happened to try them while frozen.  (If already thawed, it seems to be idempotent.)

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +281,116 @@ def setup_opt_parser
...
+update delete #{d.reverse_zone} IN NS #{d.old_fqdn}
+send
+      HEREDOC
+
+      STDOUT.puts 'forward...'
+      run_cmd nsupdate_command(forward_dns_command, d.key_file)
+      STDOUT.puts 'reverse...'
+      run_cmd nsupdate_command(reverse_dns_command, d.key_file)
+      STDOUT.puts 'updating dynamic zone files...'
+      run_cmd('rndc freeze')
```
ekohl: Should we explicitly freeze and thaw the zones we touched?

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +281,116 @@ def setup_opt_parser
...
+
+      STDOUT.puts 'updating DNS records:'
+      # Use nsupdate to update DNS records
+      # Update SOA record to new hostname; add new A and NS records; delete old A and NS records.
+      # Multi-line strings are not indented because Ruby 2.0 doesn't support <<~ heredocs
+
+      forward_dns_command = <<-HEREDOC
+local #{d.local_ip}
+zone #{d.zone}
+update add #{d.zone} 10800 SOA #{d.new_fqdn} #{d.soa_admin_domain}. #{d.new_serial} 86400 3600 604800 3600
```
ekohl: What if the admin changed values on the SOA?

jeremylenz: @ekohl The custom values would be overwritten.  I could collect and pass through the five additional values, but I feel like this specific case wouldn't happen often.  What about adding a blurb to the confirmation?
```
DNS records will be updated using nsupdate.  If you have any custom values in the SOA
 record, they will be overwritten.  If you'd rather update DNS records manually, re-run with the 
--skip-dns option.  Continue with changing hostname? [y/n]
```
or something like that

packages/katello/katello/hostname-change.rb
```diff
@@ -345,13 +501,17 @@ def run
...
       self.run_cmd("sed -i -e 's/#{@old_hostname}/#{@new_hostname}/g' #{scenarios_path}/*.yaml")
 
+      STDOUT.puts "updating hostname in hammer configuration"
+      self.run_cmd("sed -i.bak -e 's/#{@old_hostname}/#{@new_hostname}/g' #{hammer_root_config_path}/*.yml")
+      self.run_cmd("sed -i.bak -e 's/#{@old_hostname}/#{@new_hostname}/g' #{hammer_config_path}/*.yml")
+
       if File.exist?(last_scenario_yaml)
         STDOUT.puts 'backing up last_scenario.yaml'
-        temp_last_scenario_yaml = Tempfile.new('last_scenario')
+        @temp_last_scenario_yaml = Tempfile.new('last_scenario')
```
jturel: There's an attr_accessor defined for @temp_last_scenario_yaml so we should use that unless there was a reason to switch back to the instance variable

packages/katello/katello/hostname-change.rb
jturel: Would there be more than one item in the array? And is it always an array? I'd like to remove the check for 'Array' if possible

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel I'm used to using `attr_accessor` to access instance variables from outside the class, not from within it.  Tbh I didn't even realize why it was there until this comment.

packages/katello/katello/hostname-change.rb
jturel: eh, let's leave well-enough alone. it was already written this way, but I do agree on the change. this PR is big enough IMO

packages/katello/katello/hostname-change.rb
jturel: I think I added it a while back when I was making changes in here. I like to avoid referencing instance variables whenever possible. They can be too easily changed and break things especially when they are used across several methods.  Referencing something like a method that we can easily change the implementation of is better

packages/katello/katello/hostname-change.rb
jturel: Since we're generating and paying close attention to the input given here, it's probably fine

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel on the box I tested with it was a string, but for @chris1984 it was an array.  (or vice versa, can't remember)

chris1984:
Starting to test again

jturel:
@ekohl you OK with the adjustments here? I'd like to see this make the next snap

packages/katello/katello/hostname-change.rb
jeremylenz: Updated to `[@scenario_answers['foreman_proxy']['dns_reverse']].flatten.first`

jturel:
@ekohl are you waiting on / expecting any changes here?

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +290,102 @@ def setup_opt_parser
       @opt_parser.parse!
     end
 
+    def dns_managed?
+      @scenario_answers['foreman_proxy']['dns'] &&
+        @scenario_answers['foreman_proxy']['dns_managed']
```
ekohl: This logic has changed for Foreman 2.0 to also take the provider into account (https://github.com/theforeman/puppet-foreman_proxy/commit/54067fe5ac742000d3c2cdb8505d8f191383aeb7)

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +290,102 @@ def setup_opt_parser
...
+    end
+
+    def should_update_dns?
+      dns_managed? && !@options[:skip_dns]
+    end
+
+    def query_dns_server(fqdn, zone, nameserver_ip)
+      query_dns = Resolv::DNS.new(nameserver: [nameserver_ip], search: [zone], ndots: 1)
+      a_record = query_dns.getresource(fqdn, Resolv::DNS::Resource::IN::A)
```
ekohl: In Foreman 2.1 we'll also set set AAAA records (https://github.com/theforeman/puppet-foreman_proxy/commit/fecf9c8b8bc3a50bdc0ace5e96840fae90c9d240)

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +290,102 @@ def setup_opt_parser
...
+      @scenario_answers['foreman_proxy']['dns'] &&
+        @scenario_answers['foreman_proxy']['dns_managed']
+    end
+
+    def should_update_dns?
+      dns_managed? && !@options[:skip_dns]
+    end
+
+    def query_dns_server(fqdn, zone, nameserver_ip)
+      query_dns = Resolv::DNS.new(nameserver: [nameserver_ip], search: [zone], ndots: 1)
```
ekohl: IMHO you shouldn't provide a `search` here. Searching is providing `www` and expanding that to `www.$zone` and since the FQDN should be fully qualified, that should never happen. Worst case it can actually introduce bugs.

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +290,102 @@ def setup_opt_parser
...
+        serial: serial
+      }
+    end
+
+    def new_dns_values
+      new_vals = OpenStruct.new
+
+      new_vals.dns_server = @scenario_answers['foreman_proxy']['dns_server']
+      new_vals.zone = @scenario_answers['foreman_proxy']['dns_zone']
+      new_vals.reverse_zone = [@scenario_answers['foreman_proxy']['dns_reverse']].flatten.first
```
ekohl: Reverse zones in the installer can be an array. If you're updating SOA records, shouldn't you update them all in a loop? Why is the first one special?

jeremylenz: @ekohl If it's possible to have multiple reverse zones legitimately, then it makes sense to loop over them.  But I don't understand how or why one would ever have multiple reverse zones for the satellite server specified in the answer file, since you only have one record for the satellite server and a reverse zone is for reverse IP lookups.  My assumption was that someone just formatted the yaml file that way and it stuck.

ekohl: No, this is a legitimate feature. People are actually annoyed that we don't have a parameter for multiple forward zones so the Katello manual even documents a way to achieve this (https://theforeman.org/plugins/katello/3.14/installation/index.html#multiple-subnets-and-domains). Your script will not catch these, but parsing `/etc/foreman-installer/custom-hiera.yaml` is probably a bad idea.

>  since you only have one record for the satellite server and a reverse zone is for reverse IP lookups

This is a misunderstanding of the feature. If you provision new hosts, foreman-proxy can (if it manages the zone) create reverse DNS entries. If you have multiple subnets, you probably want to manage them all.

> My assumption was that someone just formatted the yaml file that way and it stuck.

Its data type is `Optional[Variant[String, Array[String]]]` (see https://github.com/theforeman/puppet-foreman_proxy/blob/ef0d8745eecccd6475a9b8d75b917c3eaf01b45b/manifests/init.pp#L428)

jeremylenz: @ekohl Thanks for clarifying.  Will update to loop over the array.

packages/katello/katello/hostname-change.rb
jeremylenz: thanks, updating

packages/katello/katello/hostname-change.rb
jeremylenz: Thanks, updating

jeremylenz:
Updated to address @ekohl comments:
* don't pass in a `search` to DNS resolver
* update definition of `dns_managed?`
* loop over multiple reverse zones

Also updated the testing instructions in the PR description.  (For testing, you must now change your answers file so that `dns_provider` is `nsupdate`)

packages/katello/katello/hostname-change.rb
```diff
@@ -292,15 +292,17 @@ def setup_opt_parser
...
+          .include?(@scenario_answers['foreman_proxy']['dns_provider'])
     end
 
     def should_update_dns?
       dns_managed? && !@options[:skip_dns]
     end
 
-    def query_dns_server(fqdn, zone, nameserver_ip)
-      query_dns = Resolv::DNS.new(nameserver: [nameserver_ip], search: [zone], ndots: 1)
+    def query_dns_server(fqdn, nameserver_ip)
```
jeremylenz: I removed the `zone` arg here in error (added back in next commit)

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +290,119 @@ def setup_opt_parser
...
+      new_vals.key_file = @scenario_answers['foreman_proxy']['keyfile']
+      new_vals.old_fqdn = @old_hostname
+      new_vals.new_fqdn = @new_hostname
+
+      ip_serial_data = query_dns_server(@old_hostname, new_vals.zone, new_vals.dns_server)
+      new_vals.ip = ip_serial_data[:ip]
+      serial = ip_serial_data[:serial]
+      new_vals.new_serial = serial + 1
+
+      new_vals.reverse_soa_records = []
```
jturel: Up to you, if you like it more. I prefer writing this type in this way:

```
new_vals.reverse_soa_records = new_vals.reverse_zones.map do |zone|
  reverse_serial = ...
  {
    reverse_zone: ....,
    new_reverse_serial: .....
  }
end
```

Either way, it looks like we don't need to include `reverse_zones` on the returned struct. Correct me if wrong :)

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +290,119 @@ def setup_opt_parser
...
+          new_reverse_serial: reverse_serial + 1
+        }
+      end
+      run_cmd('rndc thaw')
+
+      new_vals
+    end
+
+    def nsupdate_command(dns_entries, key_file)
+      "echo -e \"#{dns_entries}\" | nsupdate -l -k #{key_file}"
```
jturel: Does this need to be updated to reflect the 'provider' @ekohl mentioned?

jeremylenz: @jturel I thought that provider value is only used to determine whether we should be touching DNS.  What would need to change about the nsupdate command?

jturel: You might be correct. I asked since I don't know the answer

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel Thanks, updated

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl I created a separate [issue](https://projects.theforeman.org/issues/29268) so this can be addressed in a follow-up.

packages/katello/katello/hostname-change.rb
ekohl: The provider can be `nsupdate_gss` which means you need to add options to pass the tsig keytab and principal. You can also just say that the particular provider is not supported.

jeremylenz: @ekohl Thanks, after discussing with @chris1984 I feel like it will be best to only support the `nsupdate` provider for now.

ekohl:
I think the whole loop structure is very weird. I'd say you should use a generic update_zone method that just generates commands
```ruby
def update_zone(zone, fqdn = nil)
  resolver = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
  commands = []

  soa = resolver.getresource(zone, Resolv::DNS::Resource::IN::SOA)
  if soa.mname.to_s == old_fqdn
    # TODO: delete old SOA?
    commands << "update add #{zone} #{soa.ttl} SOA #{new_fqdn}. #{soa.rname} #{soa.serial + 1} #{soa.refresh} #{soa.retry} #{soa.expire} #{soa.minimum}"
  end

  nameservers = resolver.getresources(zone, Resolv::DNS::Resource::IN::NS)
  if nameservers.any? { |ns| ns.name.to_s == old_fqdn }
    # TODO: replace the correct nameserver
  end

  begin
    a_record = resolver.getresource(old_fqdn, Resolv::DNS::Resource::IN::A)
    commands << "update delete #{old_fqdn} A"
    commands << "update add #{new_fqdn} #{a_record.ttl} A #{a_record.address}"
  rescue resolv.ResolvError
    # This is fine
  end

  begin
    aaaa_record = resolver.getresource(old_fqdn, Resolv::DNS::Resource::IN::AAAA)
    commands << "update delete #{old_fqdn} AAAA"
    commands << "update add #{new_fqdn} #{aaaa_record.ttl} A #{aaaa_record.address}"
  rescue resolv.ResolvError
    # This is fine
  end

  commands
end
```

Then you can call that method in a loop and generate a single long nsupdate statement.

Note that in my code I have made several changes:
* It respects all the current SOA fields (this will fail if the nameserver is a recursor and thus serving lower TTLS)
* Left a TODO so that you don't break slave nameservers (note the plural getresources)

Now that I read man nsupdate, you can even replace ruby code with pure nsupdate using `prereq`. See https://linux.die.net/man/8/nsupdate and https://docstore.mik.ua/orelly/networking_2ndEd/dns/ch10_02.htm. I won't insist on that since I don't exactly know how they work.

chris1984:
Starting to test

chris1984:
@jeremylenz did you want to make the requested changes by @ekohl then ill do a final test with the different scenarios we talked about to cover all aspects?

jeremylenz:
@chris1984 sounds good

chris1984:
Starting to test

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +291,122 @@ def setup_opt_parser
...
+    def new_dns_values
+      new_vals = OpenStruct.new
+
+      new_vals.dns_server = @scenario_answers['foreman_proxy']['dns_server']
+      new_vals.zone = @scenario_answers['foreman_proxy']['dns_zone']
+      new_vals.key_file = @scenario_answers['foreman_proxy']['keyfile']
+
+      new_vals.reverse_zones = [@scenario_answers['foreman_proxy']['dns_reverse']].flatten
+      ip_serial_data = query_dns_server(@old_hostname, new_vals.zone, new_vals.dns_server)
+      new_vals.ip = ip_serial_data[:ip]
```
ekohl: With the new update_zone this is no longer needed.

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +291,122 @@ def setup_opt_parser
...
+
+    def update_zone(zone, nameserver_ip)
+      resolver = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
+      commands = []
+
+      soa = resolver.getresource(zone, Resolv::DNS::Resource::IN::SOA)
+      commands << "update add #{zone} #{soa.ttl} SOA #{@new_hostname}. #{soa.rname} #{soa.serial + 1} #{soa.refresh} #{soa.retry} #{soa.expire} #{soa.minimum}"
+
+      nameservers = resolver.getresources(zone, Resolv::DNS::Resource::IN::NS)
+      # TODO: replace the correct nameserver
```
ekohl: You are breaking the workflow where a user has multiple nameservers. This is relevant if they set up replication to other nameservers.

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +291,122 @@ def setup_opt_parser
...
+      commands << "update add #{zone}. 3600 IN NS #{@new_hostname}."
+      commands << "update delete #{zone}. IN NS #{@old_hostname}"
+
+      return commands if zone.include?('arpa') # don't modify A records for reverse zones
```
ekohl: I'd make this safer and check if it ends with `arpa` so you don't end up skipping `arpa.com`

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +291,122 @@ def setup_opt_parser
...
+
+      new_vals
+    end
+
+    def nsupdate_command(dns_entries, key_file)
+      "echo -e \"#{dns_entries}\" | nsupdate -l -k #{key_file}"
+    end
+
+    def update_zone(zone, nameserver_ip)
+      resolver = Resolv::DNS.new(nameserver: [nameserver_ip], search: [], ndots: 1)
```
ekohl: I left it out of the example, but you can also create the resolver outside of the loop and pass it in. No need to instantiate a new one every time.

packages/katello/katello/hostname-change.rb
ekohl: The new update_zone addresses this.

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl I removed the condition because if we don't update NS records, it fails the post-update name server sanity check and the entire `nsupdate` command errors.  The `update add` and `update delete` commands specify which hostname is applicable, so I don't understand what would be breaking?  What should I do for now and what needs to be in the todo for later?

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl I wanted to leave a couple of those values in because the `precheck` method also calls this.  If we get to the point in the script where we're updating DNS and it fails, the change-hostname script fails quite badly and leaves the system in a state where it's not easy to just run again.  So we query the DNS resolver for values before the process starts.

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl Thanks, updated

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl Updated, thanks

packages/katello/katello/hostname-change.rb
ekohl: In DNS records are sets. In practice you often only see one record, but it must be treated as a set. You can see this:
```console
$ dig ns example.com

; <<>> DiG 9.11.14-RedHat-9.11.14-2.fc31 <<>> ns example.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 27338
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;example.com.			IN	NS

;; ANSWER SECTION:
example.com.		13975	IN	NS	b.iana-servers.net.
example.com.		13975	IN	NS	a.iana-servers.net.

;; Query time: 11 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Fri Mar 06 15:52:55 CET 2020
;; MSG SIZE  rcvd: 88
$ dig ns google.com

; <<>> DiG 9.11.14-RedHat-9.11.14-2.fc31 <<>> ns google.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 16103
;; flags: qr rd ra; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;google.com.			IN	NS

;; ANSWER SECTION:
google.com.		13667	IN	NS	ns3.google.com.
google.com.		13667	IN	NS	ns4.google.com.
google.com.		13667	IN	NS	ns2.google.com.
google.com.		13667	IN	NS	ns1.google.com.

;; Query time: 10 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Fri Mar 06 15:56:29 CET 2020
;; MSG SIZE  rcvd: 111
```

There's more than 1 nameserver. The way bind replication works is that it will send a notify to all nameservers (besides itself) when there's an update. Assuming all permissions are correct, you can add a second server next to the foreman and have it slave the zone. If delegation is set up correct, when Foreman is down you'll still have your DNS zones available.

What you are doing here is destroying that if the user has set it up by issuing a `delete` and adding back a single record. To avoid this, look at all NS records, see if `old_fqdn` is in there. If so, that should be replaced with `new_fqdn` and add the rest of the set again.

You can argue that this really is an edge case and few users will have set this up, but I like to be correct rather than getting a bug report later that we broke a production setup.

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl Found a better way to do the precheck; updating soon

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl Thanks for explaining.   The `delete` command I issue is
```ruby
"update delete #{zone}. IN NS #{@old_hostname}"
```

In [`man nsupdate`](https://linux.die.net/man/8/nsupdate) I found

> __update delete domain-name [ ttl ] [ class ] [ type [ data...] ]__
> Deletes any resource records named domain-name. __If type and data is provided, only matching resource records will be removed.__

Since I am providing `type` (NS) and `data` (`@old_hostname` which is the fqdn) in the command, will this address the issue?

ekohl: Yes, I think it does. In that case you can remove the TODO. You can also remove the `namservers = ...` call if you're not going to use the result. I still think the `if` I wrote originally is correct though I used `old_fqdn` and you use `old_hostname`. Perhaps that was the reason it didn't work for you?

jeremylenz: @ekohl, yeah I agree `old_fqdn` is a better variable name because it makes clear that it's fully qualified.  But earlier in this PR I was convinced by @jturel to just use the existing instance variable `@old_hostname`, since we make sure that's the FQDN.  I think the `if` would work fine, but it's not needed because the specificity of the `delete` command provides the same result.

ekohl: Technically the user could have changed their nameserver to `ns1.example.com` instead of `foreman.example.com`. My version caught that. Same story with the SOA record. However, I' won't push on not catching those since I really doubt anyone who cares about DNS will use the `change-hostname` script.

jeremylenz: @ekohl Ahh I see what you were saying now.

chris1984:
Starting another round of tests

packages/katello/katello/hostname-change.rb
jeremylenz: Updated to use `assembled_nsupdate_command` in the precheck.  Thanks @jturel for the suggestion :)

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +292,107 @@ def setup_opt_parser
...
+        @scenario_answers['foreman_proxy']['dns_managed'] &&
+        @scenario_answers['foreman_proxy']['dns_provider'] == 'nsupdate'
+    end
+
+    def should_update_dns?
+      dns_managed? && !@options[:skip_dns]
+    end
+
+    def answers_dns_values
+      new_vals = OpenStruct.new
```
jturel: I think the recent refactoring makes the need for the OpenStruct a lot less compelling. what do you think?

Just using a hash will make your pre-check code simpler as well - you can just do this:

```
if @answers_dns_values.values.any? { |v| v.nil? }
  # show the error
end

```

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +292,107 @@ def setup_opt_parser
...
+      run_cmd('rndc thaw')
+
+      new_vals
+    end
+
+    def nsupdate_command(dns_entries, key_file)
+      "echo -e \"#{dns_entries}\" | nsupdate -l -k #{key_file}"
+    end
+
+    def update_zone(zone, nameserver_ip, resolver)
```
ekohl: nameserver_ip is now unused

```suggestion
    def update_zone(zone, resolver)
```

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +292,107 @@ def setup_opt_parser
...
+
+      return commands if zone.end_with?('arpa') # don't modify A records for reverse zones
+
+      begin
+        a_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::A)
+        commands << "update delete #{@old_hostname} A"
+        commands << "update add #{@new_hostname} #{a_record.ttl} A #{a_record.address}"
+      rescue Resolv::ResolvError => e
+        # This is fine
+        STDOUT.puts e
```
ekohl: In Foreman 2.1 (develop) we support IPv6-only and this is not an error

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +292,107 @@ def setup_opt_parser
...
+        STDOUT.puts "no A record found; skipping"
+      end
+
+      begin
+        aaaa_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::AAAA)
+        commands << "update delete #{@old_hostname} AAAA"
+        commands << "update add #{@new_hostname} #{aaaa_record.ttl} A #{aaaa_record.address}"
+      rescue Resolv::ResolvError => e
+        # This is fine
+        # TODO: Following two lines commented out until IPv6 is officially supported
```
ekohl: IPv6 is already supported in 2.1 (develop) and the installer will create the records.

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +292,107 @@ def setup_opt_parser
...
+
+    def answers_dns_values
+      new_vals = OpenStruct.new
+
+      new_vals.nameserver_ip = @scenario_answers['foreman_proxy']['dns_server']
+      new_vals.zone = @scenario_answers['foreman_proxy']['dns_zone']
+      new_vals.key_file = @scenario_answers['foreman_proxy']['keyfile']
+
+      new_vals.reverse_zones = [@scenario_answers['foreman_proxy']['dns_reverse']].flatten
+      run_cmd('rndc thaw')
```
ekohl: Thawing while determining answers feels like a weird side effect. I'd move this into `update_dns_records` before calling `nsupdate_command`.

packages/katello/katello/hostname-change.rb
jeremylenz: thanks, updated

packages/katello/katello/hostname-change.rb
jeremylenz: @jturel For the error message I needed the key also (`field`), so the resulting code isn't that much cleaner here.  But I still think it's better in case we need to change any of the keys in the future, since they won't be hard-coded.

jturel:
To the merger: don't forget to squash :)

packages/katello/katello/hostname-change.rb
```diff
@@ -87,27 +88,52 @@ def precheck
...
+        rescue StandardError => e # could not reach the DNS server, or something broke while assembling the data
+          STDOUT.puts e
+          fail_with_message("Error querying local DNS server for #{@old_hostname}. Make sure the 'named' service is running, or re-run with the --skip-dns option.")
+        end
+
+        @answers_dns_values.each do |field, v|
+          next unless v.nil?
+            fail_with_message """
+      Error gathering DNS data: couldn't find value for '#{field}'.
+      Make sure /etc/foreman-installer/scenarios.d/#{@options[:scenario]}-answers.yaml
```
ekohl: Perhaps this should reuse `scenarios_path` from `KatelloUtilities::Helper`. There's already another definition that uses the full path so perhaps time for another helper to avoid duplication?

I also don't like the implication that the user should change the answers file. They should **never** manually change the answers file (which is why I want to move it to `/var/lib/foreman-installer` in a future version).

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +292,99 @@ def setup_opt_parser
...
+    def should_update_dns?
+      dns_managed? && !@options[:skip_dns]
+    end
+
+    def answers_dns_values
+      {
+        nameserver_ip: @scenario_answers['foreman_proxy']['dns_server'],
+        zone: @scenario_answers['foreman_proxy']['dns_zone'],
+        key_file: @scenario_answers['foreman_proxy']['keyfile'],
+        reverse_zones: [@scenario_answers['foreman_proxy']['dns_reverse']].flatten
```
ekohl: The value may be `nil` and you don't want to end up with `[nil]`
```suggestion
        reverse_zones: [@scenario_answers['foreman_proxy']['dns_reverse']].flatten.compact
```

packages/katello/katello/hostname-change.rb
```diff
@@ -87,27 +88,52 @@ def precheck
...
+          STDOUT.puts e
+          fail_with_message("Error querying local DNS server for #{@old_hostname}. Make sure the 'named' service is running, or re-run with the --skip-dns option.")
+        end
+
+        @answers_dns_values.each do |field, v|
+          next unless v.nil?
+            fail_with_message """
+      Error gathering DNS data: couldn't find value for '#{field}'.
+      Make sure /etc/foreman-installer/scenarios.d/#{@options[:scenario]}-answers.yaml
+      'foreman-proxy' section has values for dns_zone, dns_reverse, and keyfile.
```
ekohl: `dns_reverse` can be nil and is valid.

packages/katello/katello/hostname-change.rb
```diff
@@ -87,27 +88,52 @@ def precheck
...
+      Make sure A and SOA records are present for #{@old_hostname}.
```
ekohl: This isn't really relevant anymore in the current iteration

packages/katello/katello/hostname-change.rb
```diff
@@ -239,6 +292,99 @@ def setup_opt_parser
       @opt_parser.parse!
     end
 
+    def dns_managed?
+      @scenario_answers['foreman_proxy']['dns'] &&
```
ekohl: If you want to be defensive you'll first check whether `@scenario_answers['foreman_proxy']` since its valid values are `true`, `false` and a hash. This avoids `NoMethodError`s

chris1984:
@ekohl let me know when i can test this please

ekohl:
@chris1984 You can test the actual script. All my current comments are polishing issues that prevent edge cases or clean up messages.

packages/katello/katello/hostname-change.rb
jeremylenz: @ekohl Ok, see if you prefer the reworded error message instead.

packages/katello/katello/hostname-change.rb
jeremylenz: With our flattened compacted array I think this situation would give us a `[]` which wouldn't trigger this error anyway since `[].nil?` is false.  (But this message is reworded anyway)

packages/katello/katello/hostname-change.rb
jeremylenz: hmm, if it's `true` then this also could error.  I'm going to check if it's a hash, then.

ehelms:
This will need a release bump to make it available in nightly.

jturel:
https://github.com/theforeman/foreman-packaging/pull/4830

PR #4712, opened 2020-02-13
Update configs for 2.0 repos

tbrisker:


tbrisker:
isn't the 2.0 tag already there and what we need is to create 2.1 before updating the rpm/develop branch to use that?

ekohl:
They are now since @zjhuntin created them.

tbrisker:
so is this good to go now?

PR #4713, opened 2020-02-13
Update rpm/develop for 2.1-develop

tbrisker:


tbrisker:
@ekohl good point, we should add it to the procedure for next time, for now i opened it separately at https://github.com/theforeman/foreman-packaging/pull/4714

tbrisker:
repoclosure is expected. merging.

PR #4718, opened 2020-02-13
Bump version to 3.16

jjeffers:

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -2,13 +2,13 @@
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
-%global foreman_min_version 1.24.0
+%global foreman_min_version 2.0
 %global plugin_name katello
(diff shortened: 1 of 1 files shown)


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -229,6 +229,9 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/webpack
 
 %changelog
+* Thu Feb 13 2020 James Jeffers <jjeffers@redhat.com> - 3.15.0-0.1.pre.master
```
tbrisker:
```suggestion
* Thu Feb 13 2020 James Jeffers <jjeffers@redhat.com> - 3.16.0-0.1.pre.master
```

tbrisker:
repoclosure is failing because of https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec#L30 which should be updated as well, the gem doesn't define a specific version.

tbrisker:
[test rpm]

PR #4720, opened 2020-02-17
Update rubygem-foreman_azure_rm to 2.0.7

apuntamb:
Supported Versions:

 * Nightly
 * 2.0
 * 1.24
 * 1.23

PR #4721, opened 2020-02-17
Update rubygem-foreman_azure_rm to 2.0.7

apuntamb:
Supported Versions:

 * Nightly
 * 2.0
 * 1.24
 * 1.23

### DIFF
--- a/packages/plugins/rubygem-foreman_azure_rm/rubygem-foreman_azure_rm.spec
+++ b/packages/plugins/rubygem-foreman_azure_rm/rubygem-foreman_azure_rm.spec
@@ -7,7 +7,7 @@
 %global foreman_min_version 1.17
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 2.0.6
+Version: 2.0.7
 Release: 1%{?foremandist}%{?dist}
 Summary: Azure Resource Manager as a compute resource for The Foreman
 Group: Applications/Systems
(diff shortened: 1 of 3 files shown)


packages/plugins/rubygem-foreman_azure_rm/rubygem-foreman_azure_rm.spec
```diff
@@ -111,11 +112,15 @@ cp -a .%{gem_dir}/* \
 %{gem_instdir}/Rakefile
 
 %posttrans
+%{foreman_db_migrate}
```
apuntamb: @evgeni @zjhuntin Adding the rake task here for db migrate. Let me know if `db` needs to be mentioned at other places too.

packages/plugins/rubygem-foreman_azure_rm/rubygem-foreman_azure_rm.spec
```diff
@@ -96,6 +96,7 @@ cp -a .%{gem_dir}/* \
 %dir %{gem_instdir}
 %license %{gem_instdir}/LICENSE
 %{gem_instdir}/app
+%{gem_instdir}/db
```
apuntamb: Added `%{gem_instdir}/db` manually here, similar to #4720

ehelms:
You will want to open a PR to rpm/2.0 as well

PR #4731, opened 2020-02-19
Rpm/2.0 hammer plugins update

tbrisker:


tbrisker:
repoclosure is expected as 2.0 repos haven't been published yet

PR #4736, opened 2020-02-19
Update rubygem-pulp_2to3_migration to 0.0.1rc1

jlsherrill:


jlsherrill:
Note that this will require https://github.com/Katello/katello/pull/8569/files and a packaging update for that first

jlsherrill:
Also, reminder to myself to backport this to 2.0

jlsherrill:
katello pr here: https://github.com/theforeman/foreman-packaging/pull/4738

jlsherrill:
katello pkg pr is merged, repoclosure failure i think is an issue with repoclosure

jlsherrill:
[test rpm]

evgeni:
this lowers the version from 0.0.1a1-1 to 0.0.1-0.1.rc1, so repoclosure is not happy and so won't the clients who already have a1 installed. you might want to add an epoch?

ekohl:
I'd avoid an epoch and just keep a `1` release instead of `0.1`.

evgeni:
@ekohl 0.0.1a1-1 > 0.0.1-1

ekohl:
Given they're so low in version numbers I'd prefer to release a new version rather than to start with epochs already.

jlsherrill:
I ended up bumping epoch, i don't think asking the pulp team to bump their version is the right thing to do

jlsherrill:
all green!

PR #4751, opened 2020-02-24
Update rubygem-pulp_2to3_migration to 0.0.1rc1

jlsherrill:
2.0 backport of https://github.com/theforeman/foreman-packaging/pull/4736


jlsherrill:
repoclosure is broken it seems, i assume due to repos not existing yet

evgeni:
Yeah, we didn't publish any 3.15 bits yet, so repoclosure fails. that's fine.

PR #4521, opened 2019-12-18
Update rubygem-redhat_access to 2.2.9

ehelms:

PR #4755, opened 2020-02-25
Bump rubygem-redhat_access to 2.2.11-1

evgeni:


evgeni:
@ares @ekohl mind having a look at this?

evgeni:
should obsolete #4521 and #4081

evgeni:
not sure we actually need apipie and assets as added by the template?

ares:
I was able to verify the plugin installs and generally works, however the planner and manage page are still making the menu disappear. :+1: for this version though.

evgeni:
You're saying "still", so that error existed before and is not something I introduced by  slightly changing how the package is built? Cool.

ares:
yes, I've seen that before (https://bugzilla.redhat.com/show_bug.cgi?id=1805141)

evgeni:
> I think `%{gem_instdir}/public` in files is now redundant.

No, because `foreman_assets_plugin` resolves to `%{gem_instdir}/public/assets/%{gem_name}`, so only a part of `public`, but we install into `public/assets/insights`, not `public/assets/redhat_access`

evgeni:
@ares shall we backport that to any stable releases?

ares:
I'd wait until some user asks for it, something tells me, we won't see such request

PR #4722, opened 2020-02-17
Release katello-host-tools 3.5.3

jturel:

PR #4764, opened 2020-02-26
[WIP] add gofer packages

evgeni:


evgeni:
damn you dependencies

evgeni:
we dropped k-agent for el8 in https://github.com/theforeman/foreman-packaging/pull/4722, so this is obsolete

PR #4766, opened 2020-02-27
Update rubygem-foreman-tasks to 0.17.6

adamruzicka:


adamruzicka:
Oh boy, `react-html-parser` is not packaged yet, this will be fun.

adamruzicka:
@ekohl Do I read it correctly I'll need to bring that `react-html-parser` in in a separate PR?

ekohl:
Correct. It's a build dependency (it's listed as a `BuildRequire`) and our CI job can't use results built in the same PR.

zjhuntin:
@adamruzicka did the PR to add `react-html-parser` get created?

adamruzicka:
Only for develop so far (https://github.com/theforeman/foreman-packaging/pull/4768 and https://github.com/theforeman/foreman-packaging/pull/4835)

pcreech:
[test rpm]

pcreech:
[test rpm]

package_manifest.yaml
```diff
@@ -368,6 +368,7 @@ foreman_nonscl_packages:
     katello-certs-tools: {}
     katello-selinux: {}
     keycloak-httpd-client-install: {}
+    nodejs-react-html-parser: {}
```
pcreech, line 371: I think this needs to be moved up to the foreman_nodejs_packages group above this group.  It's whitelisted for scl, but the foreman_nonscl_packages group doesn't build against that tag.

pcreech: Also, I think this needs to be broken out into it's own PR, as it seems this is a new build time dependency for foreman-tasks

ehelms:
@adamruzicka Given we are starting the branch process for Foreman 2.2, can I close this?

adamruzicka:
Alright, let's scratch this

PR #4768, opened 2020-02-27
Add nodejs-react-html-parser package

adamruzicka:
This is needed for foreman-tasks-0.17.6 for 1.24 and will be needed when new foreman-tasks in the 1.x series gets released.


ekohl:
Looks like this was generated with the released version of npm2rpm, but I should do a release that supports SCL. Currently we rely on master.

adamruzicka:
I see, sorry about making this a bit messy.

ekohl:
It's NPM packaging, I expect that to be messy. I've opened https://github.com/theforeman/npm2rpm/pull/63 and will dust off my NPM package release skills

comps/comps-foreman-rhel7.xml
```diff
@@ -357,6 +357,7 @@
       <packagereq type="default">katello-selinux</packagereq>
       <packagereq type="default">keycloak-httpd-client-install</packagereq>
       <packagereq type="default">mod_passenger</packagereq>
+      <packagereq type="default">nodejs-react-html-parser</packagereq>
```
ekohl: Looks like our package add scripts aren't updated for SCLs either

```suggestion
      <packagereq type="default">tfm-nodejs-react-html-parser</packagereq>
```

adamruzicka:
I rebuilt it with npm2rpm from master, but it seems to fail the same way

packages/foreman/nodejs-react-html-parser/nodejs-react-html-parser.spec
```diff
@@ -0,0 +1,52 @@
...
+Release: 1%{?dist}
+Summary: Parse HTML into React components
+License: MIT
+Group: Development/Libraries
+URL: https://github.com/wrakky/react-html-parser#readme
+Source0: https://registry.npmjs.org/react-html-parser/-/react-html-parser-%{version}.tgz
+%if 0%{?!scl:1}
+BuildRequires: nodejs-packaging
+%endif
+Requires: npm(htmlparser2) >= 3.9.0
```
ekohl: Looks like this should also be looking at the SCL package. Not sure why it isn't. You may also need to package that.

packages/foreman/nodejs-react-html-parser/nodejs-react-html-parser.spec
```diff
@@ -0,0 +1,52 @@
+%{?scl:%scl_package nodejs-%{npm_name}}
```
ekohl: The file should be in `packages/plugins`

ekohl:
[test rpm]

ekohl:
Relevant error:
```
[2020-03-03T09:27:35.436Z]     package: tfm-nodejs-react-html-parser-2.0.2-1.el7.noarch from downloaded_rpms
[2020-03-03T09:27:35.436Z]       unresolved deps:
[2020-03-03T09:27:35.436Z]          tfm-npm(htmlparser2) < 0:4.0.0
```

Note that this has htmlparser2 4.1.0 which isn't < 4.0.0.

rel-eng/tito.props
```diff
@@ -463,8 +465,6 @@ whitelist = nodejs-babel-plugin-module-resolver
   rubygem-ruby-nmap
   rubygem-safe_yaml
   rubygem-sentry-raven
-  rubygem-smart_proxy_dynflow_core
-  rubygem-smart_proxy_salt_core
```
adamruzicka: I can undo these, it's what one of the scripts we have did

PR #4770, opened 2020-02-27
Cherry pick tfm changes to 2.0

ekohl:
This fixes the builds of Ruby packages that provide bins. It also obsoletes turbolinks which allows users to upgrade.


evgeni:
github didn't get the memo that the test passed green

evgeni:
`koji tag-build foreman-2.0-rhel7 tfm-6.0-4.el7`

PR #4777, opened 2020-02-28
Bump pulpcore for repoclosure

zjhuntin:


package_manifest.yaml
```diff
@@ -1017,16 +1017,23 @@ pulpcore_packages:
...
+      el8:
+        - el8-katello-pulpcore-nightly
     repoclosure_lookaside_repos:
       el7:
         - el7-base
         - el7-updates
+      el8:
+        - el8-base
+        - el8-updates
+        - appstream
```
pcreech: I think this should be 'el8-appstream'

PR #4803, opened 2020-03-04
Update dynflow dependency to 1.4.3

adamruzicka:

### DIFF
--- a/dependencies/bionic/dynflow/changelog
+++ b/dependencies/bionic/dynflow/changelog
@@ -1,3 +1,9 @@
+ruby-dynflow (1.4.3-1) stable; urgency=low
+
+  * 1.4.3 released
+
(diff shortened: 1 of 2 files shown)


adamruzicka:
@mmoll It seems we don't explicitly depend on `concurrent-ruby-edge`. Should it be handled somehow or is it fine this way?

mmoll:
@adamruzicka the concurrent-ruby DEB also brings -edge. It's all super-weird, but a consequence of that gem's git repository layout. For pulling in concurrent-ruby-edge 0.6.0 you should depend on ruby-concurrent >= 1.1.6.

adamruzicka:
It already depends on ruby-concurrent >= 1.1.6 if I'm reading things right

mmoll:
Ah, indeed

mmoll:
Merged, díky @adamruzicka!

PR #4810, opened 2020-03-06
Bump rubygem-foreman_maintain to 0.6.1

upadhyeammit:
Bumping foreman-maintain to 0.6.1


upadhyeammit:
Package built successfully, however I doubt changes of spec file is legitimate ? I used below command,

./add_gem_package.sh foreman_maintain scl foreman-nightly-rhel7

dgoetz:
Yes, the template does not cover the yum repository and bash-completion section which has to be added again. Also the epoch has to be keeped because without it would always be older than the already installed package.

ekohl:
> ./add_gem_package.sh foreman_maintain scl foreman-nightly-rhel7

If you just want to bump, I'd recommend using `bump_rpm.sh` instead.

upadhyeammit:
@dgoetz I kept spec file as it is, changed version and changelog. Which should be fine ?

PR #4813, opened 2020-03-06
Fixes #29313 - require same katello-host-tools version on subpackages

jturel:
This will prevent katello-host-tools-tracer from breaking when changes are made in katello-host-tools

### DIFF
--- a/packages/client/katello-host-tools/katello-host-tools.spec
+++ b/packages/client/katello-host-tools/katello-host-tools.spec
@@ -16,7 +16,7 @@
 
 Name: katello-host-tools
 Version: 3.5.3
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: A set of commands and yum plugins that support a Katello host
 Group:   Development/Languages
 License: LGPLv2
@@ -35,7 +35,7 @@ BuildArch: noarch
 
 Requires: subscription-manager
 %if %{build_fact_plugin}
-Requires: %{name}-fact-plugin
+Requires: %{name}-fact-plugin = %{version}-%{release}
 %else
 Obsoletes: %{name}-fact-plugin < %{version}-%{release}
 %endif
@@ -101,7 +101,7 @@ Requires: python-gofer-proton >= 2.5
 %endif
 
 Requires: subscription-manager
-Requires: katello-host-tools = %{version}-%{release}
+Requires: %{name} = %{version}-%{release}
 
 %if 0%{?rhel} == 6
 Requires: yum-plugin-security
@@ -135,7 +135,7 @@ BuildArch:  noarch
 Summary:    Adds Tracer functionality to a client managed by katello-host-tools
 Group:      Development/Languages
 
-Requires: katello-host-tools
+Requires: %{name} = %{version}-%{release}
 %if %{dnf_install}
 Requires: python3-tracer
 %else
@@ -397,6 +397,9 @@ exit 0
 %endif
 
 %changelog
+* Mon Mar 23 2020 Jonathon Turel - 3.5.3-2
+- require matching katello-host-tools version in subpackages
+
 * Mon Feb 17 2020 Jonathon Turel - 3.5.3-1
 - Release 3.5.3
 


packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -34,6 +34,11 @@ BuildArch: noarch
 %endif
 
 Requires: subscription-manager
+
+%if %{build_tracer}
+Requires: %{name}-tracer = %{version}-%{release}
```
evgeni: It feels weird to have a separate package, yet always install it.

What is the reasoning for always requiring tracer?

jturel: The request is that tracer is always enabled - so I did the simple thing and installed it as a dep of host-tools. Do you think it would be more appropriate to combine -tracer RPM into host-tools itself?

evgeni: If this would be a fresh package, then I'd say yes. Given it's an existing one, your approach is fine.

I think what I (personally) find weird about that: as a user, I perceive tracer as an addon that is not required for general day to day use of a system as a katello client. As such I'd expect tracer not to be installed by default (or at least removable w/o removing all of kht). But it might be just me being me and disliking to hoard code on my systems that I don't use.

@ehelms @ekohl thoughts?

BTW, is there a redmine that could be referenced?

jturel:
@evgeni updated. Someone from my team will test as well

chris1984:
Looks good here:

```bash
[root@centos7 ~]# yum localupdate katello-host-tools-3.5.3-2.el7.noarch.rpm

Examining katello-host-tools-3.5.3-2.el7.noarch.rpm: katello-host-tools-3.5.3-2.el7.noarch
Marking katello-host-tools-3.5.3-2.el7.noarch.rpm as an update to katello-host-tools-3.5.3-1.el7.noarch
Resolving Dependencies
--> Running transaction check
---> Package katello-host-tools.noarch 0:3.5.3-1.el7 will be updated
---> Package katello-host-tools.noarch 0:3.5.3-2.el7 will be an update
--> Processing Dependency: katello-host-tools-tracer = 3.5.3-2.el7 for package: katello-host-tools-3.5.3-2.el7.noarch
Loading mirror speeds from cached hostfile
 * base: repos-va.psychz.net
 * epel: iad.mirror.rackspace.com
 * extras: distro.ibiblio.org
 * updates: repos-va.psychz.net
--> Finished Dependency Resolution
Error: Package: katello-host-tools-3.5.3-2.el7.noarch (/katello-host-tools-3.5.3-2.el7.noarch)
           Requires: katello-host-tools-tracer = 3.5.3-2.el7
           Installed: katello-host-tools-tracer-3.5.3-1.el7.noarch (@foreman-client)
               katello-host-tools-tracer = 3.5.3-1.el7
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest

[root@centos7 ~]# ls
katello-host-tools-3.5.3-2.el7.noarch.rpm katello-host-tools-tracer-3.5.3-2.el7.noarch.rpm
```

jturel:
[test rpm]

chris1984:
@jturel @evgeni tests are green :)

packages/client/katello-host-tools/katello-host-tools.spec
jturel: If I had to a pick a concern it would be the additional requests to the server on package transactions if the user isn't interested in the tracer feature. @sean797 just want to check with you: 'enable tracer by default' means 'always installed and enabled along with katello-host-tools', right?

sean797: @jturel I think so, but I didn't write that requirement. Personally, I could go either way if we should install/enable this by default. I also have concerns regarding unnecessary client side execution, but I also want to advertise this feature because I think it's unique and very useful. Would using a weak dependency help?

Another route could be something like `katello-host-tools` includes tracer, but `katello-host-tools-minimal` does not include tracer?

evgeni: Weak deps are not a thing on EL7 and older, so while "yes, this is the correct approach", it wouldn't work for everyone.

jturel:
Marking as WIP as I now think enabling this by default is perhaps too much. I'll update based on the decision with PM

jturel:
Based on conversations outside this PR we won't be enabling tracer by default. I've cannibalized this PR to address another issue :)

chris1984:
[test rpm]

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -397,6 +398,9 @@ exit 0
 %endif
 
 %changelog
+* Fri Mar 06 2020 Jonathon Turel - 3.5.3-2
```
ekohl: Perhaps this should be updated to today

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -135,7 +136,7 @@ BuildArch:  noarch
 Summary:    Adds Tracer functionality to a client managed by katello-host-tools
 Group:      Development/Languages
 
-Requires: katello-host-tools
+Requires: katello-host-tools = %{version}-%{release}
```
ekohl: This isn't done on line 105 either, but I think it can be:
```suggestion
Requires: %{name} = %{version}-%{release}
```

jturel:
@ekohl thanks for the suggestions - updated

jturel:
@ekohl updated!

ekohl:
I don't see the change I expected. Meant this line:
https://github.com/theforeman/foreman-packaging/blob/1ffdc97aa6b0a9f8b83ecf032a883388372dcf2b/packages/client/katello-host-tools/katello-host-tools.spec#L38
Should that also require a matching version or is it not that relevant?

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -122,6 +122,7 @@ BuildArch:  noarch
 Summary:    Adds an fqdn fact plugin for subscription-manager
 Group:      Development/Languages
 
+Requires: %{name} = %{version}-%{release}
```
ekohl: Nit: perhaps keep this aligned?
```suggestion
Requires:   %{name} = %{version}-%{release}
```

jturel:
@ekohl misunderstood - fixed that. it's pretty much irrelevant for the fact plugin but it can't hurt

PR #4814, opened 2020-03-08
Update patternfly and bootstrap dependencies

ezr-ondrej:
Reminder to bump dependencies after
- [x] #4379 is merged
- [ ] bump in core theforeman/foreman#7621


ezr-ondrej:
I think, this repository doesn't like me :thinking: even though I have coockies! :cookie: 
I think, the failures are related to Rails upgrade (as it fails on no rails 5.2), but I don't know enough to understand what koji wants from me

ekohl:
I'll include this in https://github.com/theforeman/foreman-packaging/pull/5115 once https://github.com/theforeman/foreman/pull/7621 is merged.

PR #4815, opened 2020-03-09
Update fog-vsphere to 3.2.5

ezr-ondrej:
Bump fog-vsphere


ezr-ondrej:
Some nasty Yankees yanked the gem right after publishing, as it's imposible to undo the mistake, I've released the `3.2.5` and it becomes bump to that version now xD

PR #4819, opened 2020-03-09
Update rubygem-foreman_ansible to 5.0.1

ares:


evgeni:
I guess this needs a pick into 2.0?

ares:
yeah, I forgot, do we still need to open cp for that?

evgeni:
Yes, please

PR #4831, opened 2020-03-10
Add ability to choose a version of the gem to the add_gem script

ShimShtein:


ShimShtein:
@ekohl, your bash fu is much better than mine. Fixed.

PR #4834, opened 2020-03-10
Set mock package_manager on EL7

ekohl:
When building on Fedora 31, this is needed. Otherwise mock attempts to use dnf.


ekohl:
Hold off merging this, I'm now getting
```
   libelf.so.1: cannot open shared object file: No such file or directory
```

ekohl:
The libelf.so.1 part was fixed by a `dnf update` on my host, but this PR is still needed to get mock to build.

PR #4838, opened 2020-03-10
Adding infrastructure repo comps

jlsherrill:


jlsherrill:
needed for https://github.com/theforeman/foreman-infra/pull/1285

jlsherrill:
I don't particularly like this solution, for the reasons mentioned here: https://community.theforeman.org/t/adding-postgresql-evr-to-jenkins-slaves/17686/11  but its here we end up going that route

PR #4845, opened 2020-03-12
Update GH pull request template

lzap:
We are not doing good job in keeping this one up to date in `deb/develop` and `rpm/develop` branches. Only CLI GH tools use them, GH Web appears to be using the template from master. It's expected that CLI user knowns what she is doing, so simplified version is ok.


lzap:
https://github.com/theforeman/foreman-packaging/pull/4846

PR #4852, opened 2020-03-12
Clean plugin comps

ekohl:
* 5c254bf99b4308ecd9c5bb3e6b790571626df68e added ffi but this actually lives in foreman rather than plugins.
* 04bfbd6bb4a791f06d5a1ec040e11339e31c7341 added xmlrpc to plugins but it actually lives in foreman.
* algebrick, faraday and sequel are harder to trace, but all appear unused. May have been for a long time.


pcreech:
ACK

PR #4856, opened 2020-03-13
Adjust foreman-nightly tag to only mention el8

zjhuntin:


ekohl:
A just as critical part is actually https://github.com/theforeman/foreman-packaging/blob/86eef8dec21e5ecc7337cf0a954c294aec1030b6/rel-eng/tito.props#L351

PR #4858, opened 2020-03-13
Fixes #29456: Add SCL support for installer

ehelms:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -464,6 +464,28 @@ foreman_installer_packages:
         - el8-puppet-6
         - el8-devel
   hosts:
+    foreman-installer:
+      releasers:
+        - "{{ nightly_releaser }}"
+      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
+      nightly_package_tito_releaser_args:
+        - "jenkins_job=foreman-installer-develop-source-release"
+        - "jenkins_url=https://ci.theforeman.org"
+      build_package_tito_builder: "tito.builder.FetchBuilder"
+    foreman-selinux:
+      releasers:
+        - "{{ nightly_releaser }}"
+      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
+      nightly_package_tito_releaser_args:
+        - "jenkins_job=foreman-selinux-develop-source-release"
+        - "jenkins_url=https://ci.theforeman.org"
+      build_package_tito_builder: "tito.builder.FetchBuilder"
+    katello-certs-tools: {}
+    katello-selinux: {}
+    puppet-agent-oauth: {}
+    puppet-agent-puppet-strings: {}
+    puppet-agent-rgen: {}
+    puppet-agent-yard: {}
     rubygem-ansi: {}
     rubygem-kafo: {}
     rubygem-kafo_parsers: {}
@@ -500,31 +522,9 @@ foreman_nonscl_packages:
         - el8-puppet-6
   hosts:
     foreman-bootloaders-redhat: {}
-    foreman-installer:
-      releasers:
-        - "{{ nightly_releaser }}"
-      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
-      nightly_package_tito_releaser_args:
-        - "jenkins_job=foreman-installer-develop-source-release"
-        - "jenkins_url=https://ci.theforeman.org"
-      build_package_tito_builder: "tito.builder.FetchBuilder"
     foreman-release-scl: {}
-    foreman-selinux:
-      releasers:
-        - "{{ nightly_releaser }}"
-      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
-      nightly_package_tito_releaser_args:
-        - "jenkins_job=foreman-selinux-develop-source-release"
-        - "jenkins_url=https://ci.theforeman.org"
-      build_package_tito_builder: "tito.builder.FetchBuilder"
-    katello-certs-tools: {}
-    katello-selinux: {}
     keycloak-httpd-client-install: {}
     pcp-mmvstatsd: {}
-    puppet-agent-oauth: {}
-    puppet-agent-puppet-strings: {}
-    puppet-agent-rgen: {}
-    puppet-agent-yard: {}
     rubygem-foreman_maintain: {}
 
 foreman_proxy_packages:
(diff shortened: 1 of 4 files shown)


packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -15,16 +22,16 @@ Source0:    https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?prerel
...
 Requires:   %{?scl_prefix}rubygem(kafo) >= 4.0.0
 Requires:   %{?scl_prefix}rubygem(kafo) < 5.0.0
 Requires:   foreman-selinux
 Requires:   %{?scl_prefix_ruby}ruby(release)
 Requires:   foreman-maintain
 
 BuildRequires: asciidoc
-BuildRequires: rubygem(rake)
-BuildRequires: %{?scl_prefix}puppet-agent >= 1.9.0
+BuildRequires: puppet-agent >= 1.9.0
```
ekohl: 1.9.0 was Puppet 4. We didn't bump it then because the build root still had Puppet 4, but it's probably safe to require `>= 5.5.0` now. Perhaps also define it as a macro so they don't go out of sync?

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -56,20 +63,21 @@ Various scenarios and tools for the Katello ecosystem
...
   sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X/usr/bin/%{?scl:%{scl_prefix}}rubyX' bin/foreman-installer bin/foreman-proxy-certs-generate bin/katello-certs-check
 %endif
-rake build \
+
+%{scl_rake} build \
   VERSION=%{version} \
-  LOCALSTATEDIR=%{_localstatedir} \
+  LOCALSTATEDIR=%{_root_localstatedir} \
   PREFIX=%{_prefix} \
   SBINDIR=%{_sbindir} \
```
ekohl: Isn't this also changed in the SCL context to `/opt`?

ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
[test rpm]

package_manifest.yaml
```diff
@@ -500,14 +513,6 @@ foreman_nonscl_packages:
...
-    foreman-installer:
-      releasers:
-        - "{{ nightly_releaser }}"
-      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
-      nightly_package_tito_releaser_args:
-        - "jenkins_job=foreman-installer-develop-source-release"
-        - "jenkins_url=https://ci.theforeman.org"
-      build_package_tito_builder: "tito.builder.FetchBuilder"
     foreman-release-scl: {}
     foreman-selinux:
```
ekohl, line 517: Why isn't this part of it? The installer depends on it and AFAIK is the only package that depends on foreman-selinux (besides katello-selinux).

packages/foreman/foreman-installer/foreman-installer.spec
ehelms: I think we can to keep it here so that the foreman-installer commands are all available system wide.

package_manifest.yaml
ehelms: Will do

packages/foreman/foreman-installer/foreman-installer.spec
ekohl: My point is IIRC `%{_sbindir}` resolves to `/opt/.../bin` if SCL is set and we want `%{_root_sbindir}` because I agree they should be available system wide.

ehelms: Ahhh, I think that's only true when you build *in* the SCL vs. against the SCL. You can see from my upgrade run that it installed and used the SCL:

```
tfm-rubygem-kafo.noarch              4.1.0-2.el7            @foreman-koji       
tfm-rubygem-kafo_parsers.noarch      1.1.0-2.el7            @foreman-koji       
tfm-rubygem-kafo_wizards.noarch      0.0.1-3.el7            @foreman-koji       
[root@upgrade-katello-nightly-centos7 vagrant]# which foreman-installer
/sbin/foreman-installer
[root@upgrade-katello-nightly-centos7 vagrant]# cat /sbin/foreman-installer
#!/usr/bin/tfm-ruby
require 'rubygems'
require 'kafo'

CONFIG_DIR = '/etc/foreman-installer/scenarios.d/'

# Run the install
@result = Kafo::KafoConfigure.run

# handle exit code when help was invoked or installer ended with '2' (success in puppet)
if @result.nil? || (!@result.config.app[:detailed_exitcodes] && @result.exit_code == 2)
  exit(0)
else
  exit(@result.exit_code)
end
```

ekohl: Ok, sounds good.

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -94,10 +102,10 @@ done
...
 %doc README.*
 %license LICENSE
-%dir %{_sysconfdir}/%{name}
+%dir %{_root_sysconfdir}/%{name}
 %config(noreplace) %{_sysconfdir}/%{name}/custom-hiera.yaml
-%dir %{_sysconfdir}/%{name}/scenarios.d
-%{_sysconfdir}/%{name}/scenarios.d/foreman.migrations
+%dir %{_root_sysconfdir}/%{name}/scenarios.d
+%{_root_sysconfdir}/%{name}/scenarios.d/foreman.migrations
 %config(noreplace) %attr(600, root, root) %{_sysconfdir}/%{name}/scenarios.d/foreman.yaml
```
ekohl, line 109: These need to be `_root_sysconfdir` as well. Otherwise I wonder if we need these root_* macros at all.

ehelms:
Updated with consistent macro use. All files are in the same locations they are in a standard install.

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -1,4 +1,8 @@
-%global release 1
+%{?scl:%global scl_prefix %{scl}-}
+%global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
```
ekohl, line 2: I don't think this is used now

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -56,7 +60,8 @@ Various scenarios and tools for the Katello ecosystem
 %if %{?scl:1}%{!?scl:0}
```
ekohl, line 60: Can be simplified as well

```suggestion
%if 0%{?scl:1}
```

ekohl:
Also, does this deserve a Redmine issue?

ehelms:
Yep, I had a Redmine created, just now updating commit message with it.

PR #4862, opened 2020-03-16
Refs #29356 - Package puma-plugin-systemd gem

ekohl:


ekohl:
Looks like it's marked to only support `puma ~> 3.6` but we're on version 4.

ehelms:
Given there is a Redmine issue, and this waits on action within that gem I am gonna close for now to track in other places -- https://projects.theforeman.org/issues/29356

PR #4866, opened 2020-03-16
Add el8 modules repo to repoclosures

zjhuntin:


repoclosure/yum.conf
```diff
@@ -11,6 +11,12 @@ assumeyes=1
 syslog_ident=repoclosure
 syslog_device=
 
+[el8-modules]
```
ekohl: Are these modules we built or modules from CentOS? If the former, then I'd expect to see that in the name.

zjhuntin: They are modules from CentOS that we flattened @pcreech is this the best way to explain?

pcreech: To be _super_ technical, it's:
BaseOS
AppStream
PowerTools

all flattened into a single repo, with modules flattened into a standard RPM repo.  Ruby 2.6 module has been removed

zjhuntin:
[test redmine]

PR #4871, opened 2020-03-17
Update rubygem-foreman_vault to 0.1.0

UXabre:
(cherry picked from commit 8678681723503712124ee2665eef9a9630fa1ce1)

### DIFF
--- a/packages/plugins/rubygem-foreman_vault/rubygem-foreman_vault.spec
+++ b/packages/plugins/rubygem-foreman_vault/rubygem-foreman_vault.spec
@@ -8,7 +8,7 @@
 %global foreman_min_version 1.20
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.0.1
+Version: 0.1.0
 Release: 1%{?foremandist}%{?dist}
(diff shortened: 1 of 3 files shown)


packages/plugins/rubygem-foreman_vault/rubygem-foreman_vault.spec
```diff
@@ -105,5 +105,11 @@ cp -pa .%{gem_dir}/* \
 exit 0
 
 %changelog
+* Mon Mar 16 2020  0.1.0-1
+- Update to 0.1.0
+
+* Tue Jan 07 2020 Eric D. Helms <ericdhelms@gmail.com> - 0.0.1-2
+- Drop migrate, seed and restart posttans
```
ekohl: This part wasn't dropped here, and shouldn't be in 1.24. Mind dropping this log?

UXabre: Yes, will drop it! :-)

UXabre:
Companion PR for 2.0: https://github.com/theforeman/foreman-packaging/pull/4872

PR #4879, opened 2020-03-17
Update nodejs packages for el8 part 3

zjhuntin:


zjhuntin:
[test rpm]

packages/foreman/nodejs-invariant/nodejs-invariant.spec
```diff
@@ -15,6 +15,7 @@ Source0: https://registry.npmjs.org/%{npm_name}/-/%{npm_name}-%{version}.tgz
 BuildRequires: %{?scl_prefix_nodejs}npm
 %else
 BuildRequires: nodejs-packaging
+BuildRequires: npm
```
ekohl: I really still wonder if this is needed. It shouldn't be, at least on EL7 and Fedora as far as I checked.

zjhuntin:
[test rpm]

packages/foreman/nodejs-node-sass/nodejs-node-sass.spec
```diff
@@ -395,9 +397,11 @@ AutoProv: no
 %prep
 mkdir -p %{npm_cache_dir}
 %{?scl:scl enable %{?scl_nodejs} - << \end_of_scl}
+%if 0%{?scl:1}
```
ehelms: This shouldn't be needed, we want to run the code below in both cases.

packages/foreman/nodejs-node-sass/nodejs-node-sass.spec
```diff
@@ -406,16 +410,21 @@ done
 export SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=true
 
 %{?scl:scl enable %{?scl_nodejs} - << \end_of_scl}
+%if 0%{?scl:1}
```
ehelms: Same, not needed in this case

packages/foreman/nodejs-node-sass/nodejs-node-sass.spec
```diff
@@ -406,16 +410,21 @@ done
...
 # Replace node-gyp with packaged version that uses nodejs-devel headers
 rm -rf node_modules/%{npm_name}/node_modules/node-gyp
 ln -s %{nodejs_sitelib}/node-gyp node_modules/%{npm_name}/node_modules/
 
 pushd node_modules/%{npm_name}
+
 %{?scl:scl enable %{scl_nodejs} - << \end_of_scl}
+%if 0%{?scl:1}
 npm_config_nodedir=/opt/rh/%{scl_nodejs}/%{_root_datadir}/include/node node scripts/build -f
+%endif
```
ehelms: This will need an else condition similar to:

```
%else
node scripts/build -f
%endif
```

packages/foreman/nodejs-node-sass/nodejs-node-sass.spec
```diff
@@ -400,22 +399,25 @@ for tgz in %{sources}; do
...
 
-%setup -T -q -a 179 -D -n %{npm_cache_dir}
+%setup -T -q -a 177 -D -n %{npm_cache_dir}
 
 %build
 export SKIP_SASS_BINARY_DOWNLOAD_FOR_CI=true
 
 %{?scl:scl enable %{?scl_nodejs} - << \end_of_scl}
-npm install --ignore-scripts --cache-min Infinity --cache %{?scl:../}%{npm_cache_dir} --no-shrinkwrap --no-optional --global-style true %{npm_name}@%{version}
+npm install --ignore-scripts --cache-min Infinity --cache %{?scl:../}%{npm_cache_dir} --no-shrinkwrap --no-optional -
```
ehelms: Looks like the end of this accidentally got cut off.

ehelms: Since EL8 uses a newer NodeJS version, it needs to reference the npm cache the same way as SCL. I think at this point its safe to say we aren't packaging these for older NodeJS anymore. Therefore I would do the following:

Modify line 385 - 389 to be:

```
%define npm_cache_dir npm_cache
```

Then update this line to:

```
--cache ../%{npm_cache_dir}
```

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
```diff
@@ -1,46 +1,59 @@
...
+index 564564e..d3974ba 100644
 --- a/lib/configure.js
 +++ b/lib/configure.js
-@@ -62,7 +62,9 @@ function configure (gyp, argv, callback) {
+@@ -37,7 +37,7 @@ function configure (gyp, argv, callback) {
+ 
+     if (gyp.opts.nodedir) {
+       // --nodedir was specified. use that for the dev files
+-      nodeDir = gyp.opts.nodedir.replace(/^~/, os.homedir())
++      nodeDir = '/opt/rh/rh-nodejs10/root/usr/share/node'
```
ehelms: Switch out for non-scl pathing

packages/foreman/nodejs-node-gyp/nodejs-node-gyp.spec
```diff
@@ -66,125 +66,135 @@ Source52: https://registry.npmjs.org/json-schema/-/json-schema-0.2.3.tgz
...
+%if 0%{?rhel} >= 8
+Requires:   python3-devel
+%else
 Requires:   python-devel
+%endif
 Requires:   make
 
 BuildRequires: %{?scl_prefix_nodejs}nodejs-devel
+
 %if 0%{!?scl:1}
```
ehelms: Drop the `!`

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
```diff
@@ -1,46 +1,49 @@
...
+index 564564e..d3974ba 100644
 --- a/lib/configure.js
 +++ b/lib/configure.js
-@@ -62,7 +62,9 @@ function configure (gyp, argv, callback) {
+@@ -37,7 +37,7 @@ function configure (gyp, argv, callback) {
+
+     if (gyp.opts.nodedir) {
+       // --nodedir was specified. use that for the dev files
+-      nodeDir = gyp.opts.nodedir.replace(/^~/, os.homedir())
++      nodeDir = '/root/usr/share/node'
```
ehelms: Drop the `/root` here

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
```diff
@@ -1,46 +1,49 @@
...
 +        log.verbose('get node dir', 'no --target version specified, falling back to RPM installed headers')
        }
 
        if (!release.semver) {
-@@ -75,7 +77,7 @@ function configure (gyp, argv, callback) {
-       gyp.commands.install([ release.version ], function (err, version) {
-         if (err) return callback(err)
-         log.verbose('get node dir', 'target node version installed:', release.versionDir)
--        nodeDir = path.resolve(gyp.devDir, release.versionDir)
-+        nodeDir = '/opt/rh/rh-nodejs10/root/usr/share/node'
```
ehelms: Just `/usr/share/node` here

zjhuntin: This is so odd, because I don't see it locally.

packages/foreman/nodejs-node-gyp/node-gyp-node-dirs.patch
```diff
@@ -1,46 +1,49 @@
...
-         createBuildDir()
-       })
-     }
```
ehelms: This patch removes this whole block, the same is true in the scl version of this patch. We still need this part of the patch present with the nodeDir set how it was in the patch prior to this update.

ehelms:
I think it could be useful to pull nodejs-node-gyp and possibly nodejs-node-sass into independent PRs due to how much they are changing and since nodejs-node-gyp has to get built before node-sass.

zjhuntin:
[test rpm]

zjhuntin:
[test rpm]

zjhuntin:
[test rpm]

zjhuntin:
This has failed from repoclosure which is expected

package_manifest.yaml
```diff
@@ -482,6 +482,9 @@ foreman_installer_packages:
       build_package_tito_builder: "tito.builder.FetchBuilder"
     katello-certs-tools: {}
     katello-selinux: {}
+    keycloak-httpd-client-install: {}
+    nodejs-node-gyp: {}
+    pcp-mmvstatsd: {}
```
ehelms, line 487: Where did these packages come from?

comps/comps-foreman-rhel7.xml
```diff
@@ -360,6 +360,7 @@
       <packagereq type="default">katello-selinux</packagereq>
       <packagereq type="default">keycloak-httpd-client-install</packagereq>
       <packagereq type="default">mod_passenger</packagereq>
+      <packagereq type="default">nodejs-node-gyp</packagereq>
```
ehelms, line 363: Was this handled by nodejs-node-gyp PR we merged ?

rel-eng/tito.props
```diff
@@ -24,6 +24,7 @@ whitelist = foreman-bootloaders-redhat
   katello-certs-tools
   katello-selinux
   keycloak-httpd-client-install
+  nodejs-node-gyp
```
ehelms, line 27: Similar question

comps/comps-foreman-rhel7.xml
zjhuntin: I don't think it was, which... is problematic now that I think of it.

rel-eng/tito.props
zjhuntin: Same here, it was not covered.

package_manifest.yaml
zjhuntin: what the hell, I have no idea. I'll remove the extras

ehelms, line 487: I don't think these packages should be here right now.

PR #4885, opened 2020-03-18
Rebuild against newer foreman-webpack-vendor

ehelms:


ehelms:
@evgeni

evgeni:
NACK, as pointed out by @tbrisker we don't have any webpack dep here anyways.

```console
% rpm -q --requires -p ~/Downloads/tfm-rubygem-katello-3.13.4-1.el7.noarch.rpm |grep webpack
<empty>

% rpm -q --requires -p ~/Downloads/tfm-rubygem-katello-3.13.4-2.el7.noarch.rpm |grep webpack
<empty>
```

PR #4886, opened 2020-03-19
Pulp dependencies

sjha4:
Supported Versions:

 * Nightly

PR #4891, opened 2020-03-20
Updated pulp deps

evgeni:
this splits out the "bumped" packages from #4886 and ignores dynaconf for now


evgeni:
repoclosure errors expected, needs #4892 to fix

sjha4:
@evgeni : Is the python-requests failure expected?

evgeni:
No it's not

sjha4:
It passed in the el7 build.. :confused:

evgeni:
My best guess is rpm on EL8 uses stricter patch settings when it comes to fuzz and that fails. I'll look into it on Monday

evgeni:
repoclosure expected

PR #4894, opened 2020-03-20

ekohl:
This gives reliable service restarts since systemd keeps the socket open. bundle exec needs to be called with --keep-file-descriptors since systemd passes these in. The replacement is changed since the service now uses /usr/share/foreman/bin/rails which doesn't need replacing.

It also makes sure the service enable is called on the foreman-service subpackage rather than the main package.

Depends on https://github.com/theforeman/foreman/pull/7536


mmoll:
[test packaging]

mmoll:
What up with these tests?

ekohl:
Looks like I submitted the test from an incorrect branch name.

PR #4895, opened 2020-03-20

ekohl:
This gives reliable service restarts since systemd keeps the socket open.

Replacement of ExecStart is no longer needed since foreman core now calls /usr/share/foreman/bin/rails which is already modified to use /usr/bin/tfm-ruby.

It also correcs the macros to only execute on the subpackage. The incorrect systemd_postun_with_restart (systemctl try-restart) is changed to the appropriate systemd_post (systemctl preset).

Depends on https://github.com/theforeman/foreman/pull/7536


ehelms:
Looks good, might as well bump the release.

ehelms:
[test rpm]

ehelms:
[test rpm]

packages/foreman/foreman/foreman.spec
```diff
@@ -771,7 +772,6 @@ install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
 install -Dp -m0644 extras/systemd/%{name}.service %{buildroot}%{_unitdir}/%{name}.service
```
ehelms: Need to install the socket as well here

ekohl:
Thanks, updated and now green.

ekohl:
Also updated now to drop the unused sysconfig file.

PR #4896, opened 2020-03-23
Rpm/release foreman 2.0.0 rc3 update foreman release 2.0.0 update foreman 2.0.0 update foreman installer 2.0.0 update foreman selinux 2.0.0 update foreman proxy 2.0.0

ehelms:


packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -145,6 +145,9 @@ done
 %{_sbindir}/foreman-proxy-certs-generate
 
 %changelog
+* Mon Mar 23 2020 Eric D. Helms <ericdhelms@gmail.com> - 1:2.0.0-0.3.rc3
+- Release foreman-installer 2.0.0
```
ekohl: In the tooling I had this at least indicated the RC. Now we have rc1, rc2 and rc3 all looking like it's going GA.

ehelms: Updated manually, this just came from running the scripts.

PR #4901, opened 2020-03-24

ekohl:
This gives reliable service restarts since systemd keeps the socket open. bundle exec needs to be called with --keep-file-descriptors since systemd passes these in. The replacement is changed since the service now uses /usr/share/foreman/bin/rails which doesn't need replacing.

It also makes sure the service enable is called on the foreman-service subpackage rather than the main package.

Depends on theforeman/foreman#7536

This PR replaces https://github.com/theforeman/foreman-packaging/pull/4894 with a correct branch name.


ekohl:
Looks like libsass.so is failing to build on ARM.

mmoll:
As we build only non-nightly releases on ARM, let's ignore this.

mmoll:
merged, bedankt @ekohl!

PR #4904, opened 2020-03-24
Build gyp

zjhuntin:


packages/foreman/gyp/gyp.spec
```diff
@@ -0,0 +1,194 @@
...
+%files
+%license LICENSE
+%doc AUTHORS
+%{_bindir}/gyp
+%{python3_sitelib}/*
+
+
+%changelog
+* Tue mar 24 2020 Zach Huntington-Meath <zhunting@redhat.com> - 0.1-0.33.fcd686f1git.2
+-
```
ekohl: What are the changes here?

zjhuntin: Thanks!

zjhuntin: I decided I wasn't changing it, so I just kept it as is for now. Should I go ahead and bump it just to show we are building it here?

zjhuntin:
@ekohl yes, the patches are from the src rpm which came from https://centos.pkgs.org/8/epel-x86_64/gyp-0.1-0.33.fcd686f1git.el8.1.noarch.rpm.html

packages/foreman/gyp/gyp.spec
ekohl: If there's no change at all, I'd keep the exact same version as in EPEL including the same release.

ekohl:
I wonder if we should rebuild this or rather rely on EPEL since it saves us from updating it. Given there are patches, that's IMHO a strong reason not to copy it. Any idea how many more packages we'll need that are already in EPEL?

zjhuntin:
I do not know exactly, @pcreech can you think of any off the top of your head?

pcreech:
I don't have an immediate list off the top of my head.

We are being a little more intentional this time around with trying our best to not rely on epel if we can.

As far as management of the package, looks like this is a very stable codebase, both the source repository and in fedora's packaging repositories.  There hasn't been really any changes in months.

Personally, my vote is to continue down the no-epel path, at least a little bit further.

zjhuntin:
@ehelms added

PR #4907, opened 2020-03-25
Don't build libsass with native CPU code tuning

mmoll:


ekohl:
:+1: for disabling that. We'll need to do the same thing when we do package it in RPMs.

PR #4909, opened 2020-03-25
fix katello requires for pulp 2to3

jlsherrill:


jlsherrill:
Katello 3.15 pr: https://github.com/Katello/katello/pull/8628

PR #4913, opened 2020-03-25
add version dependency also to foreman-service

mmoll:


mmoll:
signet gem related arm build failures unrelated :sweat_smile:

PR #4915, opened 2020-03-26
Fixes #29281: Stop dynflow before service is removed

ehelms:

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -9,7 +9,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 9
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman/foreman.spec
```diff
@@ -672,6 +672,7 @@ Meta Package to install dynflow sidekiq executor support
 %systemd_post %{dynflow_sidekiq_service_name}.service
 
 %preun dynflow-sidekiq
+%systemd_preun dynflow.service
```
ekohl: It used to be part of the main foreman package, not `dynflow-sidekiq`. Upgrading users won't have that installed yet. IIRC it was also `dynflowd` and not `dynflow`.

ehelms:
Added one for rpm/develop -- https://github.com/theforeman/foreman-packaging/pull/4920

The spec changes were different enough I didn't go with a cherrypick.

ehelms:
Updated with explicit use of systemctl

PR #4919, opened 2020-03-26
Add -p to mkdir in add_npm_package script to enable version bumping

zjhuntin:


add_npm_package.sh
```diff
@@ -45,7 +45,7 @@ generate_npm_package() {
     sed -n '/%changelog/,$p' $PACKAGE_DIR/*.spec > OLD_CHANGELOG
     git rm -r $PACKAGE_DIR
   fi
-  mkdir $PACKAGE_DIR
+  mkdir -p $PACKAGE_DIR
```
ekohl: This is weird because just above there's code that does `git rm -r $PACKAGE_DIR`, so it shouldn't be an issue. Did you have untracked files in the directory or something?

zjhuntin: I had some in the general directory, but not in that specific directory :confused:

ehelms: Does the `-p` harm anything to exist?

ekohl: The entire script has always assumed the directory doesn't exist and operates in a clean environment. If it does exist, it can indicate that it's not detecting `$UPDATE` correctly and there might be a bug somewhere else.

PR #4920, opened 2020-03-26
Fixes #29281: Stop dynflowd service before removal

ehelms:

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -9,7 +9,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 5
+%global release 6
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -955,6 +955,10 @@ rm -rf %{buildroot}
 %ghost %attr(0640,root,%{name}) %config(noreplace) %{_datadir}/%{name}/config/initializers/local_secret_token.rb
 %{_tmpfilesdir}/%{name}.conf
 
+%preun
+systemctl --no-reload disable dynflowd.service > /dev/null 2>&1 || :
+systemctl stop dynflowd.service > /dev/null 2>&1 || :
+
 %pre
 # Add the "foreman" user and group
 getent group %{name} >/dev/null || groupadd -r %{name}
@@ -999,6 +1003,9 @@ exit 0
 %systemd_postun_with_restart %{name}.service
 
 %changelog
+* Thu Mar 26 2020 Eric D. Helms <ericdhelms@gmail.com> - 2.1.0-0.6.develop
+- Stop dynflow service before removal
+
 * Fri Mar 20 2020 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> - 2.1.0-0.5.develop
 - Use systemd socket activation (#29144)
 


ekohl:
If I look at https://github.com/systemd/systemd/blob/a85daa0dfb3eb03be9845760e90e54b9af8fb00e/src/core/macros.systemd.in#L56-L62 this shouldn't trigger anything on upgrade. Am I reading that correct?

ehelms:
The comment suggests that, but I don't see why the code would prevent it. My testing indicated this worked.

ehelms:
This is easier to test with https://github.com/theforeman/forklift/pull/1137 as you can run the stages a little easier for inspection.

ekohl:
The generated macros:
```
$ rpm -q --scripts foreman-2.1.0-0.6.develop.20200326210321git40fab71.el7.noarch.rpm
preinstall scriptlet (using /bin/sh):
# Add the "foreman" user and group
getent group foreman >/dev/null || groupadd -r foreman
getent passwd foreman >/dev/null || \
useradd -r -g foreman -d /usr/share/foreman -s /sbin/nologin -c "Foreman" foreman
exit 0
postinstall scriptlet (using /bin/sh):
# secret token used for cookie signing etc.
if [ ! -f /usr/share/foreman/config/initializers/local_secret_token.rb ]; then
  touch /usr/share/foreman/config/initializers/local_secret_token.rb
  chmod 0660 /usr/share/foreman/config/initializers/local_secret_token.rb
  chgrp foreman /usr/share/foreman/config/initializers/local_secret_token.rb
  /usr/sbin/foreman-rake security:generate_token >/dev/null 2>&1 || :
  chmod 0640 /usr/share/foreman/config/initializers/local_secret_token.rb
fi

# encryption key used to encrypt DB contents
# move the generated key file to /etc/foreman/ so users back it up, symlink to it from ~foreman
if [ ! -e /usr/share/foreman/config/initializers/encryption_key.rb -a \
     ! -e /etc/foreman/encryption_key.rb ]; then
  touch /usr/share/foreman/config/initializers/encryption_key.rb
  chmod 0660 /usr/share/foreman/config/initializers/encryption_key.rb
  chgrp foreman /usr/share/foreman/config/initializers/encryption_key.rb
  /usr/sbin/foreman-rake security:generate_encryption_key >/dev/null 2>&1 || :
  chmod 0640 /usr/share/foreman/config/initializers/encryption_key.rb
  mv /usr/share/foreman/config/initializers/encryption_key.rb /etc/foreman/
fi
if [ ! -e /usr/share/foreman/config/initializers/encryption_key.rb -a \
     -e /etc/foreman/encryption_key.rb ]; then
  ln -s /etc/foreman/encryption_key.rb /usr/share/foreman/config/initializers/
fi
exit 0
preuninstall scriptlet (using /bin/sh):

if [ $1 -eq 0 ] ; then 
        # Package removal, not upgrade 
        systemctl --no-reload disable dynflowd.service > /dev/null 2>&1 || : 
        systemctl stop dynflowd.service > /dev/null 2>&1 || : 
fi
```

AFAIK on upgrade the scriptlet is called with `1` as a parameter so it shouldn't be executed. See https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_syntax as well. That part is weird to me because in theory it shouldn't work.

ehelms:
So putting an explicit stop and disable into the pretrans would work?

ekohl:
Doesn't have to be `%pretrans`, `%preun` would also work but I think we should copy the result and unconditionally execute it. We can drop it after branching 2.1

ehelms:
Updated with explicit use of systemctlUpdated with explicit use of systemctl

PR #4925, opened 2020-03-27
Bump rubygem-qpid_messaging to 1.39.0-1

ekohl:
Note EPEL has already bumped to 1.39.0 so rebuilding 1.38.0 is no longer possible.


zjhuntin:
We are closing this because Candlepin is changing to Artemis and this will no longer be needed.

PR #4926, opened 2020-03-29
added foreman_datacenter

Zenya:


Zenya:
@mmoll hi
can you please help me with releasing rpm package?
I got some errors and can't find out why it can't be builded

mmoll:
I'm only doing DEB here...

dgoetz:
The package tries to include apipie cache which seems to be not available. If this is correct because of the plugin not having an api, you can remove `%{foreman_apipie_cache_foreman}` and `%{foreman_apipie_cache_plugin}` from the spec file.

zjhuntin:
[test rpm]

evgeni:
[test rpm]

evgeni:
Interesting, your plugin *does* have an API, but it only writes to `/usr/share/foreman/public/apipie-cache/plugin/foreman_datacenter` not `%{gem_instdir}/public/apipie-cache/plugin/foreman_datacenter`. I am not sure if that will work, but we can try :)

so instead of removing both apipie lines, you need to only have `%{foreman_apipie_cache_foreman}` in the `%files` section.

evgeni:
oh, and I don't see a 2.0 release on https://github.com/theforeman/foreman_datacenter, and https://github.com/theforeman/foreman_datacenter/pull/35 ;)

evgeni:
obsoleted by #5116

PR #4927, opened 2020-03-30
Add rubygem-multi_json to EL8

ehelms:


ehelms:
Build already exists but needs to land in the repo.

PR #4932, opened 2020-03-31
remove js dev-dependecies

sharvit:
As far as I understand, those deps are not required for the build.

Locally, I could build katello without them


sharvit:
@evgeni @tbrisker thoughts?

evgeni:
That still will be pulled in by `foreman-assets`

tbrisker:
What @evgeni said - we have to have the builder required either directly or indirectly for the packages to build, and i'm guessing same goes for the other packages.

tbrisker:
Additionally, i expect there will be issues with `package.json` requiring different versions to what is in the rpm spec (directly or indirectly) leading to interesting results.
I opened an alternative suggestion to revert the 5.0 release - https://github.com/theforeman/foreman-js/pull/132. I still believe that will be much easier to do than release a bunch of plugins and update their specs, even though it might require a bit of manual work outside of lerna (e.g. yanking the wrong release from node modules)

sharvit:
@tbrisker the plugin doesn't actually need any of those packages to do the build.
In the package json the exist only for listing, testing and building the storybook.

In other words, this Is working:
```bash
cd ../katello
rm -rf node_modules
npm i --production
cd ../foreman
RAILS_ENV=production bundle exec rake plugin:assets:precompile[katello]
```

evgeni:
I talked with @sharvit about that a bit, and the idea is that yes, we would end up with the same packages installed in the buildroot (as they are pulled in by foreman-assets), but this change would relieve us from having to update the plugin specs all the time stuff changes. and we don't support different versions in the `package.json` anyways, as we don't package different versions in parallel. so as soon core bumps to vendor v5, plugins have to use the same or we can't build them.

tbrisker:
Wouldn't that mean that plugins may be broken once we bump the js version and we won't know it and they might stop working until we rebuild them?

evgeni:
@tbrisker is that different to today tho? When we bump vendor today, nothing checks that plugins are buildable (and repoclosure does not validate BuildRequires [which we should change maybe, but that's a different topic]) or work (@ekohl's smoker does, but I don't think we gate on these result yet).

ekohl:
This makes sense if the only dev dependencies are the ones that are in Foreman core. If there's ever one that isn't, this will be a problem.

As for the build requires compatibility: in the past we tried to set generated compatibility requirements in the RPM but that proved to be painful.

sharvit:
> This makes sense if the only dev dependencies are the ones that are in Foreman core. If there's ever one that isn't, this will be a problem.

Yes, I only meant removing the once that already exits in foreman-assets.

ekohl:
This approach is not going to work in the current form since our tooling is going to place these right back. We either need to remove these from package.json files in the plugins or modify [update-requirements](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/update-requirements) to special case these. IMHO the former is better but I may be missing something.

PR #4934, opened 2020-04-01
Update rubygem-openscap to 0.4.9

evgeni:


evgeni:
draft until https://github.com/OpenSCAP/ruby-openscap/issues/17 is sorted

the changes should also allow building on EL8

evgeni:
it has been sorted, all good.

PR #4921, opened 2020-03-27
Fixes #29424 - %check section for gems with native libs

ekohl:
Since we manually move file to their correct locations, it happens that they end up in the wrong place. The RPM build then passes but the resulting package is broken. By adding a %check section that requires the gem, this should be prevented in most cases.

PR #4935, opened 2020-04-01
Update rubygem-ffi to 1.12.2 and build on el8

evgeni:


evgeni:
repoclosure fails beause it can't find `libruby.so.2.5` on el8 :woman_shrugging: 

@zjhuntin @pcreech @ehelms I only enabled ffi build on el8, not all of `proxy` group as I wanted to have ffi for openscap. do you want to have the other gems in the group built for el8 in the same PR, or individually later?

oh, and I should update comps.

ekohl:
I've been working on adding `%check` sections to gems with native extensions because they break silently right now. See https://github.com/theforeman/foreman-packaging/pull/4921/files#diff-e11b4537673837d14032dc8531d601f8

ehelms:
Individually later is fine, wouldn't mind if you tackle the group but I don't think its required. I'd be curious if that repoclosure failure is fixed by the dnf repoclosure change.

ehelms:
[test rpm]

ehelms:
@evgeni Now with repoclosure passing, will let you respond about the native extensions

evgeni:
@ekohl interestingly, this fails on EL8 now, but when I take the RPM and install it, it works:

```console
$ podman run -ti --rm centos:8 bash                 
[root@588420ca9ba1 /]# yum install http://koji.katello.org/kojifiles/work/tasks/6050/306050/rubygem-ffi-1.12.2-1.el8.x86_64.rpm

[root@588420ca9ba1 /]# irb
irb(main):001:0> require 'ffi'
=> true
```

ekohl:
I saw the same thing on another gem that was built on EL8: https://github.com/theforeman/foreman-packaging/pull/4921#discussion_r399280489
To debug this, a working mock env is needed so I excluded it first.

ehelms:
Current mock in rpm/develop works. I looked into tihs a bit and tried a few things but I am stumped too on how to tell that ruby installation to find the right library path

ekohl:
I've used `strace` because it can track file opens. That's been the most useful tool to me.

evgeni:
@ekohl @ehelms this is a huge abomination, but it works by faking the same folder structure that ruby is looking for (thanks `strace`). interestingly this also works on EL7, which I didn't expect.

evgeni:
:green_apple:

packages/foreman/rubygem-ffi/rubygem-ffi.spec
```diff
@@ -70,6 +70,16 @@ cp -a .%{gem_extdir_mri}/{gem.build_complete,*.so} %{buildroot}%{gem_extdir_mri}
...
 rm -rf %{buildroot}%{gem_instdir}/ext/
 
+%check
+%{?scl:scl enable %{scl} - << \EOF}
+mkdir gem_ext_test
+cp -a %{buildroot}%{gem_dir} gem_ext_test/
+mkdir -p gem_ext_test/gems/extensions/%{_arch}-%{_target_os}/$(ruby -r rbconfig -e 'print RbConfig::CONFIG["ruby_version"]')/
+cp -a %{buildroot}%{gem_extdir_mri} gem_ext_test/gems/extensions/%{_arch}-%{_target_os}/$(ruby -r rbconfig -e 'print RbConfig::CONFIG["ruby_version"]')/
+GEM_PATH="./gem_ext_test/gems:$GEM_PATH" ruby -e "require '%{gem_name}'"
+rm -rf gem_ext_test
```
ehelms, line 80: That's such a pretty hammer! I suppose I have to ask:

 * Can we macro this ?
 * Or just add it to gem2rpm to get included in native library gems?

evgeni: I will not try to break our macros twice in a week.

That said, gem2rpm template is what @ekohl did in his original check code and so I would just update that.

ehelms: Hah, ok. I would maybe nominate this if we turned it into a macro to be its own special package just for build macros not tied to building foreman itself.

ekohl: In my experience with https://github.com/theforeman/foreman-packaging/pull/4921 it turns out that a lot of gems aren't a `require '%{gem_name}'` but actually a different name.

@evgeni is this something you'd like to take over? I'm getting the feeling that I've opened Pandora's box with it and won't have time to finish.

evgeni: I like boxes. You can stack them.

Yes, I could take that one over as part of "make it el8"

evgeni:
> My only concern is that this installs it to a directory that's not the same as we actually install it to. That means it's not a representative test. Most issues we've seen are with installing the .so file to some directory that's never read.

Technically, I agree, but I couldn't find a way to make EL8 ruby find it in the right folder structure but the wrong prefix.

I think we can run a two-step verification here: have a `%check` that does some basic checking of the built .so, and then have a `obal test` command, that will act on a set of downloaded RPMs like `obal repoclosure` does, and then tests the built RPM in a container or somesuch?

ekohl:
I think that makes sense. In Fedora they're also looking at automated tests and I think we should look at what they're doing to avoid duplicating efforts.

evgeni:
So I updated #4957 and #4942 to follow the same structure as this PR and got them to pass on EL7 and EL8, so I'd say we can add it to the template now without too much pain.

I also had to revert https://github.com/theforeman/foreman-packaging/pull/4631 as it installs the .so file into a folder where ruby does not look for it, and I could not understand why the change was made.

evgeni:
going to merge this, as it had acks and all the changes that came afterwards were cosmetics only

PR #4939, opened 2020-04-01
correct BuildRequires for new pulp gems

evgeni:


evgeni:
Repoclosure expected, as the pulp rpms are not published yet

PR #4951, opened 2020-04-02
fix contentdir for smart_proxy_omaha

dgoetz:
The wrong content_dir breaks update to 2.0 if openscap and omaha are installed.


dgoetz:
@ekohl can we get this in before 2.0 release? It was a small error when moving Smart-Proxy to SCL but will potentially break the update for some users.

ehelms:
@dgoetz can you open a PR to the 2.0 branch?

PR #4955, opened 2020-04-02
Build node-sass for el8

zjhuntin:


ehelms:
I don't think these can be easy built in the same PR since nodejs-node-sass needs nodejs-node-gyp. Could you split nodejs-node-gyp to it's own PR?

zjhuntin:
[test rpm]

PR #4956, opened 2020-04-03
Obsolete non-scl version of kafo, kafo_parsers, kafo_wizards

ehelms:

### DIFF
--- a/packages/foreman/rubygem-kafo/rubygem-kafo.spec
+++ b/packages/foreman/rubygem-kafo/rubygem-kafo.spec
@@ -7,7 +7,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 4.1.0
-Release: 2%{?dist}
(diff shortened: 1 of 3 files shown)


packages/foreman/rubygem-kafo/rubygem-kafo.spec
```diff
@@ -34,6 +34,8 @@ BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 # end specfile generated dependencies
 
+Obsoletes: rubygem-%{gem_name} <= 4.1.0
```
ekohl, line 37: Could you hide this in non-SCL context? True for all files
```suggestion
%{?scl:Obsoletes: rubygem-%{gem_name} <= 4.1.0}
```

ekohl:
@ehelms SCL support was merged so this can be merged too now?

PR #4957, opened 2020-04-03
Add %check to rubygem-journald-native

ekohl:
Split off from https://github.com/theforeman/foreman-packaging/pull/4921

### DIFF
--- a/packages/foreman/rubygem-journald-native/rubygem-journald-native.spec
+++ b/packages/foreman/rubygem-journald-native/rubygem-journald-native.spec
@@ -2,6 +2,7 @@
 %{!?scl:%global pkg_name %{name}}
 
 %global gem_name journald-native
+%global gem_require_name journald/native
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 1.0.11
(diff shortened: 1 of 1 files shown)


zjhuntin:
[test rpm]

evgeni:
@ekohl I made this build, but it's not pretty.

the two issues I faced:
1. the gem name (journald-native) does not match the require name (journald/native), so I introduced a gem_require_name macro, which I'd default to gem_name in our templates, but here we'd override it. not sure if just `s#-#/#` is a nice workaround for that?
2. for some reason the build process didn't create the `gem.build_complete` file on EL8 (it does on EL7, and it does for other EL8 builds), so I `touch`ed it.

cc @zjhuntin @ehelms for visibility too

ekohl:
> the gem name (journald-native) does not match the require name (journald/native), so I introduced a gem_require_name macro, which I'd default to gem_name in our templates, but here we'd override it. not sure if just `s#-#/#` is a nice workaround for that?

I considered this too. If you look at https://github.com/theforeman/foreman-packaging/pull/4921 you'll see a few variations:
* nio4r -> nio
* websocket-driver -> websocket/driver

So a separate `gem_require_name` macro is certainly a good idea. I'd start with manual overrides and not an automatic replacement. It's easy enough to see if it's wrong because the build will fail.

PR #4965, opened 2020-04-03
Update rubygem-foreman_ansible to 4.0.3.5

ares:


ares:
closing as there's not big value, new templates wouldn't work with older Foreman, thanks koji for scratch build :-)

PR #4967, opened 2020-04-03
Package manifests as a directory

ehelms:
This was an idea I had with how large the package manifest is to break it up into more functional areas of smaller files easier to parse at a time.

Relies on https://github.com/theforeman/obal/pull/237

@evgeni @pcreech @zjhuntin @ekohl I converted just a few to see the gist of the idea and allow y'all to weigh in if oyu'd find this helpful or not.


ekohl:
I'll be honest and say that I don't care too much. Just make sure the tools are updated as well
https://github.com/theforeman/foreman-packaging/blob/1d4f76bbe73e9ac7aa11ab84dc63aed993deb4af/add_gem_package.sh#L84-L109
https://github.com/theforeman/foreman-packaging/blob/1d4f76bbe73e9ac7aa11ab84dc63aed993deb4af/add_npm_package.sh#L102-L108
https://github.com/theforeman/foreman-packaging/blob/1d4f76bbe73e9ac7aa11ab84dc63aed993deb4af/add_pypi_package.sh#L107-L130

PR #4973, opened 2020-04-07
Move installer packages to foreman-nightly-rhel7 tag

ehelms:


ehelms:
None of these packages have any SCL context so we can reduce the ease of managing them by moving them to foreman-nightly-rhel7 tag.

ehelms:
I updated and grouped all of the nonscl-scl packages that belong to foreman-maintain in a single group. To merge this I will need to add these packages to foreman-nightly-rhel7 and then cleanup nonscl tag to prevent older versions from winning in Koji heirarchy.

PR #4974, opened 2020-04-07
Rely on default repoclosure config

ehelms:
Requires https://github.com/theforeman/obal/pull/239


evgeni:
the tests are failing because they still refer to `el7.conf` ;)

ehelms:
@evgeni Passing now!

PR #4975, opened 2020-04-07
Switch to using koji_tags_variable

ehelms:


ehelms:
Requires: https://github.com/theforeman/obal/pull/240

This is a step towards introducing dropping of tito in a more step by step approach.

PR #4977, opened 2020-04-07
Add images/ and fonts/ to exclude of plugin webpack builds

evgeni:
new @theforeman/vendor makes webpack deploy pf4 artifacts into those
folders, but they are only really needed for foreman itself, not plugins


evgeni:
example failure: http://koji.katello.org/kojifiles/work/tasks/7543/307543/build.log

ekohl:
Can't plugins have images?

evgeni:
@ekohl they can, but they would end up in the plugin folder, not in the top level, or so I hope :)

sharvit:
Thanks @evgeni  👍

PR #4984, opened 2020-04-07
Build selinux packages for EL8

ehelms:

### DIFF
--- a/packages/foreman/foreman-selinux/foreman-selinux.spec
+++ b/packages/foreman/foreman-selinux/foreman-selinux.spec
@@ -22,7 +22,7 @@
 
 %define moduletype apps
 
-%global release 1
+%global release 2
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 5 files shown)


packages/foreman/foreman-selinux/foreman-selinux.spec
```diff
@@ -36,16 +36,28 @@ License:        GPLv3+
...
+Requires(post):     /sbin/restorecon
+Requires(post):     /usr/sbin/setsebool
+Requires(post):     /usr/sbin/selinuxenabled
+Requires(post):     /usr/sbin/semanage
 Requires(post):     selinux-policy-targeted
-Requires(postun):   /usr/sbin/semodule, /sbin/restorecon
+Requires(postun):   /usr/sbin/semodule
+Requires(postun):   /sbin/restorecon
+
+%if 0%{?rhel} > 7
```
ekohl, line 56: Just thinking out loud: this won't be true on Fedora. Shouldn't we match EL7 which we can consider legacy? While we don't build on Fedora, it'd be nice to be able to.

packages/foreman/foreman-selinux/foreman-selinux.spec
```diff
@@ -36,16 +36,28 @@ License:        GPLv3+
...
+Requires(post):     /sbin/restorecon
+Requires(post):     /usr/sbin/setsebool
+Requires(post):     /usr/sbin/selinuxenabled
+Requires(post):     /usr/sbin/semanage
 Requires(post):     selinux-policy-targeted
-Requires(postun):   /usr/sbin/semodule, /sbin/restorecon
+Requires(postun):   /usr/sbin/semodule
+Requires(postun):   /sbin/restorecon
+
+%if 0%{?rhel} < 8
```
ekohl, line 56: Isn't this still resulting in using `policycoreutils-python` on Fedora since that doesn't define `%{rhel}`
```suggestion
%if 0%{?rhel} == 7
```

ehelms:
All failures are nodejs related. This is good for another review.

PR #4985, opened 2020-04-08
Fix ignoring fonts and images for plugins once more

evgeni:


evgeni:
@ehelms there is a German saying "all the good things are three". here we are :see_no_evil:

ehelms:
What was the error?

evgeni:
The files didn't get deleted as the rm commands are not part of the macro

PR #4988, opened 2020-04-08
Update foreman_omaha to 4.0.0

mmoll:


timogoebel:
@mmoll: Thanks, I believe the release is broken. The graphql plugin API in core needs the db to exist. So currently `rails db:create` fails because the plugin API code runs during rails initialization and before the db can be created.

timogoebel:
https://github.com/theforeman/foreman_omaha/pull/49 should implement a workaround for this. I'll need to cut a new release, though.

timogoebel:
This can be closed in favor of #5015.

PR #4989, opened 2020-04-08
Move koji_tags to a dictionary

ehelms:
Requires: https://github.com/theforeman/obal/pull/242

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -23,8 +23,11 @@ foreman_scl_packages:
     scl: 'tfm'
     package_base_dir: 'packages/foreman/'
     koji_tags:
-      - foreman-nightly-rhel7
-      - foreman-nightly-el8
+      - name: foreman-nightly-rhel7
+        scl: 'tfm'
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -934,7 +964,9 @@ plugin_nonscl_packages:
   vars:
     package_base_dir: 'packages/plugins/'
     koji_tags:
-      - foreman-plugins-nightly-nonscl-rhel7
+      - name: foreman-plugins-nightly-nonscl-rhel7
+        scl: 'tfm'
```
evgeni, line 968:
```suggestion
```

evgeni: ACK pending this change, btw

ehelms:
Fair, I did try to think about but had no good answer on how to reduce repetition. As there is also a notion of defining the properties of the tag once and referencing the needed tag by a package or set of packages. Anchors sound like they would solve both. Or maybe the use of vars files?

PR #5000, opened 2020-04-08
Foreman Bumped to build on el8 p4

zjhuntin:


ehelms:
A bunch of these have already been built for EL8 can you re-check them. You should see such in the changelog.

PR #5002, opened 2020-04-08
Elevate most common repoclosure config to top level

ehelms:
I found myself repeating myself a few times and thought that we should just define exceptions to the rule within each group and have a top level rule otherwise.


pcreech:
@ehelms this could use a rebase

PR #5005, opened 2020-04-08
Build foreman for el8

zjhuntin:


zjhuntin:
[test rpm]

zjhuntin:
[test rpm]

ehelms:
[test rpm]

zjhuntin:
[test rpm]

zjhuntin:
[test rpm]

ehelms:
[test rpm]

packages/foreman/foreman/foreman.spec
```diff
@@ -65,6 +65,7 @@ Requires: %{?scl_prefix}rubygem(will_paginate) >= 3.1.7
 Requires: %{?scl_prefix}rubygem(will_paginate) < 4
 Requires: %{?scl_prefix}rubygem(ancestry) >= 2.0
 Requires: %{?scl_prefix}rubygem(ancestry) < 4
+BuildRequires: %{?scl_prefix}rubygem(bigdecimal)
```
ehelms, line 68: Can you move this down to the BuildRequires section. This should also be `%{?scl_prefix_ruby}` for EL7. I would also make this a Requires as well as we are highly likely to need this at runtime.

packages/foreman/foreman/foreman.spec
```diff
@@ -57,6 +57,7 @@ Requires: %{name}-debug
 
 # start specfile main Requires
 Requires: %{?scl_prefix}rubygem(rails) = 5.2.1
+Requires: %{?scl_prefix}rubygem(bigdecimal)
```
ehelms, line 60: scl_prefix_ruby

packages/foreman/foreman/foreman.spec
```diff
@@ -905,6 +907,10 @@ rm -rf ./usr \\
 %%{?-s:rm -rf %%{buildroot}%%{gem_instdir}/public/webpack/images}
 EOF
 
+%if 0%{?fedora} || 0%{?rhel} >= 8
+grep -lr '#!/usr/bin/python' usr/share/foreman/extras/noVNC | xargs sed -i 's/#!\/usr\/bin\/python/#!\/usr\/bin\/python3/g'
```
ehelms, line 911: At this point it's too late, as this file has already been copied into the buildroot. I'd move this to just after `%install` section and you can drop the `usr/share/foreman` path

packages/foreman/foreman/foreman.spec
```diff
@@ -737,6 +739,10 @@ webpack --bail --config config/webpack.config.js
 rm config/database.yml config/settings.yaml
 
 %install
+%if 0%{?fedora} || 0%{?rhel} >= 8
+grep -lr '#!/usr/bin/python' /extras/noVNC | xargs sed -i 's/#!\/usr\/bin\/python/#!\/usr\/bin\/python3/g'
```
ehelms, line 743: The `/` makes it an absolute path from the root. You want `extras/noVNC`

packages/foreman/foreman/foreman.spec
```diff
@@ -57,6 +57,7 @@ Requires: %{name}-debug
 
 # start specfile main Requires
 Requires: %{?scl_prefix}rubygem(rails) = 5.2.1
+Requires: %{?scl_prefix_ruby}rubygem(bigdecimal)
```
ekohl, line 60: This section is autogenerated using [update-requirements](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/update-requirements) so the next update will wipe this out. If it's needed, either add it in a section that's not autogenerated or add it to https://github.com/theforeman/foreman/blob/develop/Gemfile

packages/foreman/foreman/foreman.spec
```diff
@@ -147,9 +148,10 @@ Requires: %{?scl_prefix}rubygem(rack-jsonp)
 
 # Build dependencies
 %{?systemd_requires}
-BuildRequires: gettext
 BuildRequires: asciidoc
+BuildRequires: gettext
 BuildRequires: %{scl_ruby_bin}
+BuildRequires: %{?scl_prefix_ruby}rubygem(bigdecimal)
```
ekohl, line 154: This one is a bit odd - the other bigdecimal was in the managed section, this one isn't. Below is `# start specfile main BuildRequires` which is the managed section.

packages/foreman/foreman/foreman.spec
```diff
@@ -737,6 +739,10 @@ webpack --bail --config config/webpack.config.js
 rm config/database.yml config/settings.yaml
 
 %install
+%if 0%{?fedora} || 0%{?rhel} >= 8
+grep -lr '#!/usr/bin/python' extras/noVNC | xargs sed -i 's/#!\/usr\/bin\/python/#!\/usr\/bin\/python3/g'
```
ekohl, line 743: Why is the grep needed? Can't you just sed it? I'd also use pipes as separators so you don't need to escape everything.
```suggestion
sed -i 's|#!/usr/bin/python|#!/usr/bin/python3|g' extras/noVNC
```

On a side note: it'd be great if we could actually depend on websockify as a standalone package and solve it there. That would also update our bundled copy by a lot.

ehelms: I think there is also the use of `/usr/bin/env python` in a few of the files. So need to account for each.

packages/foreman/foreman/foreman.spec
ehelms: It is a bit odd, seems like the packaged Ruby on EL7 requires bigdecimal and on EL8 they broke it out to not be required by default.

ehelms:
> On a side note: it'd be great if we could actually depend on websockify as a standalone package and solve it there. That would also update our bundled copy by a lot.

Could you write this up in an issue so we can tackle it as a stand alone fix? I'd like to get foreman built with minimal changes as it unlocks the ability to build and test many, many things.

ekohl:
It came up in https://community.theforeman.org/t/consolidating-the-console/11215 but seeing as how the original author doesn't respond I think we should take it over. https://projects.theforeman.org/issues/29539 is there now just for unbundling.

packages/foreman/foreman/foreman.spec
ekohl: I now see that extras/noVNC is a directory, not a file. Still, the comment about the pipes stands.

zjhuntin: @ekohl Thanks for the advice!

PR #5006, opened 2020-04-08
Rebuild rubygem-gettext for el8

zjhuntin:


ehelms:
Looks like this should be re-generated with our gem2rpm template to help fix the issues

zjhuntin:
This is at repoclosures now.

zjhuntin:
[test rpm]

packages/foreman/rubygem-gettext/rubygem-gettext.spec
```diff
@@ -1,181 +1,108 @@
...
-
-
-# For --short-circult
-%{__rm} -f *.lang
-
-# modify find-lang.sh to deal with gettext .mo files under
-# %%{gem_instdir}/locale (required on EL6)
-sed -e 's|/share/locale/|/locale/|' \
-	/usr/lib/rpm/find-lang.sh \
-	> find-lang-modified.sh
```
ehelms, line 121: I don't feel like any of the stuff in here should be used in production so it ought to be OK to omit it. @ekohl do you know?

packages/foreman/rubygem-gettext/rubygem-gettext.spec
```diff
@@ -1,181 +1,108 @@
...
+%doc %{gem_docdir}
+%doc %{gem_instdir}/README.md
+%{gem_instdir}/Rakefile
+%doc %{gem_instdir}/doc
+%{gem_instdir}/gettext.gemspec
+%{gem_instdir}/test
 
 %changelog
+* Wed Apr 08 2020 Zach Huntington-Meath <zhunting@redhat.com> - 3.1.4-9
+- Bump release to build for el8
```
ekohl, line 104: This is more than just a release bump

packages/foreman/rubygem-gettext/rubygem-gettext.spec
ekohl: Looks like it's commented out in https://koji.fedoraproject.org/koji/buildinfo?buildID=1432155 at least. However there's various changes here that actually trim things down, like removing the `po` directory.

ehelms: I don't think we need that either in production? This appears as just a runtime dependency of the rails i18n gem in our deployment.

packages/foreman/rubygem-gettext/rubygem-gettext.spec
zjhuntin: Ack, sorry this was before I had revamped it.

packages/foreman/rubygem-gettext/rubygem-gettext.spec
```diff
@@ -1,181 +1,108 @@
...
-
-%exclude	%{gem_cache}
+%{_bindir}/rxgettext
+%{_bindir}/rmsgfmt
+%{_bindir}/rmsginit
+%exclude %{gem_instdir}/.yardopts
+%{gem_instdir}/bin
+%{gem_libdir}
+%{gem_instdir}/locale
+%{gem_instdir}/po
```
ekohl, line 88: Looks like this was stripped in the past and was not needed to ship.

packages/foreman/rubygem-gettext/rubygem-gettext.spec
```diff
@@ -1,181 +1,108 @@
...
+%{gem_instdir}/samples
```
ekohl, line 89: This was previously in the doc package

PR #5011, opened 2020-04-08
Rebuild rubygem-passenger for el8

zjhuntin:


ehelms:
There is passenger 6.0 in EPEL that I think we should rely on while we still support passenger. And drop carrying this old version. @ekohl thoughts?

PR #5019, opened 2020-04-09
Rebuild rubygem-sassc for el8

zjhuntin:


packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -70,7 +70,7 @@ cp -a .%{gem_dir}/* \
 mkdir -p %{buildroot}%{gem_extdir_mri}/%{gem_name}
 
 # hack to get the C extension where we expect it to be
-cp %{buildroot}/opt/theforeman/tfm/root/usr/share/gems/gems/sassc-%{version}/ext/libsass.so .%{gem_extdir_mri}/
+cp %{buildroot}/usr/share/gems/gems/sassc-2.2.1/ext/libsass.so .%{gem_extdir_mri}/
```
ehelms, line 73: This will break on EL7 since it needs the SCL path present. I think you will need an if-else to handle both.

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -70,7 +72,11 @@ cp -a .%{gem_dir}/* \
 mkdir -p %{buildroot}%{gem_extdir_mri}/%{gem_name}
 
 # hack to get the C extension where we expect it to be
+%if 0%{?rhel} >= 8
```
ehelms, line 75: Let's reverse the if-else check such that it uses the SCL path if its equal to RHEL 7 and falls back to standard paths as the else condition.

ekohl:
Before you continue this, please look at https://github.com/theforeman/foreman-packaging/pull/4921/files#diff-1aeb3a639cd0b50f5c5c8e2e2b4af65e because this gem is broken right now. That part fixes it.

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -25,10 +28,14 @@ BuildRequires: %{?scl_prefix_ruby}rubygems-devel
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 # end specfile generated dependencies
 
+BuildRequires: libsass
+Requires: libsass
+# We explicitly point it to libsass.so.0 in %build
+Requires: libsass.so.0()
```
ekohl, line 34: Fedora has `libsass.so.1` so this is a bit of an ugly hack. Maybe EL8 has `libsass.so.1` as well but by making this explicit, repoclosure would pick it up. It was inspired by rubygem-sassc in Fedora.

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -66,44 +80,38 @@ mkdir -p %{buildroot}%{gem_dir}
...
+%doc %{gem_instdir}/CODE_OF_CONDUCT.md
 %{gem_instdir}/Gemfile
 %doc %{gem_instdir}/README.md
 %{gem_instdir}/Rakefile
 %{gem_instdir}/sassc.gemspec
 %{gem_instdir}/test
 
 %changelog
+* Thu Apr 09 2020 Zach Huntington-Meath <zhunting@redhat.com> - 2.2.1-2
+- Bump release to build for el8
```
ekohl, line 113: Please update this.

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -1,12 +1,15 @@
 # template: scl
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
+%{!?_root_libdir:%global _root_libdir %{_libdir}}
 
+%global debug_package %{nil}
+%global __debug_install_post /bin/true
```
ekohl, line 7: Since there's no native parts to it anymore, is this still needed?

evgeni:
[test rpm]

zjhuntin:
I'm gonna close this in favor of: https://github.com/theforeman/foreman-packaging/pull/5053

PR #5022, opened 2020-04-09
Adds patch to raise header size to 16380 bytes

bmbouter:

### DIFF
--- a/packages/pulpcore/python-aiohttp/python-aiohttp.spec
+++ b/packages/pulpcore/python-aiohttp/python-aiohttp.spec
@@ -3,12 +3,13 @@
 
 Name:           python-%{pypi_name}
 Version:        3.6.2
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        Async http client/server framework (asyncio)
 
 License:        Apache 2
 URL:            https://github.com/aio-libs/aiohttp
 Source0:        https://files.pythonhosted.org/packages/source/a/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+Patch0:         allow-larger-headers.patch
 
 BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools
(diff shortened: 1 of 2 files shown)


jlsherrill:
Mind bumping the release to '3' : https://github.com/theforeman/foreman-packaging/blob/d3fe4f8c13ef522a8296a44827ed3460379ffcff/packages/pulpcore/python-aiohttp/python-aiohttp.spec#L6

and add a changelog entry for the new 3.6.2-2: 
https://github.com/theforeman/foreman-packaging/blob/d3fe4f8c13ef522a8296a44827ed3460379ffcff/packages/pulpcore/python-aiohttp/python-aiohttp.spec#L56

packages/pulpcore/python-aiohttp/python-aiohttp.spec
```diff
@@ -9,6 +9,7 @@ Summary:        Async http client/server framework (asyncio)
 License:        Apache 2
 URL:            https://github.com/aio-libs/aiohttp
 Source0:        https://files.pythonhosted.org/packages/source/a/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+Patch0:         allow-larger-headers.patch
```
jlsherrill, line 12: is something like this required as well in %prep? 

%patch0 -p1

i always use it when adding a patch

ekohl: We use `%autosetup` so maybe that already takes care of it?

bmbouter: I left it as is because I took it off from here https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/pulpcore/python-requests/python-requests.spec#L12

bmbouter:
@jlsherrill and @ekohl  I pushed the changes what do you think?

packages/pulpcore/python-aiohttp/python-aiohttp.spec
ekohl: They have a different autosetup line with `-p1`: https://github.com/theforeman/foreman-packaging/blob/3964a5c2d1aafb2e2d3092ab428de7ce80306bc4/packages/pulpcore/python-requests/python-requests.spec#L38

PR #5024, opened 2020-04-09
Add fx dependency to katello

ianballou:
Adds fx dependency to Katello for Katello Applicability.

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -8,7 +8,7 @@
 %global prereleasesource pre.master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global mainver 3.16.0
-%global release 7
+%global release 8
 
 Name:    %{?scl_prefix}rubygem-%{gem_name}
 Summary: Content and Subscription Management plugin for Foreman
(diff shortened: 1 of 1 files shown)


jlsherrill:
release needs a bump

jlsherrill:
The reason this is failing is because the fx rpm is in the wrong comps.  It looks like it was put in comps/comps-foreman-rhel7.xml  but it needs to go in comps/comps-katello-server-rhel7.xml

sorry for not catching that earlier!

jlsherrill:
and i would fix it in a 2nd pr

ianballou:
[test rpm]

jlsherrill:
[test rpm]

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -57,6 +57,7 @@ Requires: %{?scl_prefix}rubygem(deface) >= 1.0.2
 Requires: %{?scl_prefix}rubygem(deface) < 2.0.0
 Requires: %{?scl_prefix}rubygem(angular-rails-templates) >= 1.0.2
 Requires: %{?scl_prefix}rubygem(angular-rails-templates) < 1.1
+Requires: %{?scl_prefix}rubygem(fx) < 1.0
```
ekohl, line 60: This is not preset in the gemspec so the next automated update will remove this: https://github.com/theforeman/foreman-packaging/pull/5062

ianballou:
Alright, I guess then we'll need to add it only after the Katello PR lands that requires Fx.

PR #5027, opened 2020-04-13
Refs #29514: Add puma status

ehelms:


ehelms:
Repoclosures are all related to mass EL8 rebuilds happening

sthirugn:
I created a foreman issue last week to track this change - https://projects.theforeman.org/issues/29514

packages/foreman/rubygem-net_http_unix/rubygem-net_http_unix.spec
```diff
@@ -0,0 +1,97 @@
...
+cp -a .%{_bindir}/* \
+        %{buildroot}%{_bindir}/
+find %{buildroot}%{gem_instdir}/bin -type f | xargs chmod a+x
+
+%files
+%dir %{gem_instdir}
+%{_bindir}/net_http_unix
+%exclude %{gem_instdir}/.gitignore
+%exclude %{gem_instdir}/.travis.yml
+%{gem_instdir}/Guardfile
```
evgeni, line 78: can be excluded

gem2rpm/scl.spec.erb
```diff
@@ -4,6 +4,7 @@ require 'json'
 root = `git rev-parse --show-toplevel`.strip
 SCL_PREFIXES = JSON.load(File.read(File.join(root, 'scl_prefixes.json')))
 LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
+config.rules[:ignore].push(*['Guardfile'])
```
ehelms, line 7: @evgeni Added a commit that adds Guardfile to our ignore list from here on out.

ekohl: We should do that more often. There's more files like that that we often end up ignoring.

ehelms: Agreed. That's why I laid the groundwork.

evgeni: @ehelms could you update to the syntax Ewoud posted from the other template?

ehelms: The irony is, I started with the syntax and the voice in my head said why repeat yourself. >.< Updated.

PR #5033, opened 2020-04-14
Add .rpmlintrc

ehelms:

### DIFF
--- a/.rpmlintrc
+++ b/.rpmlintrc
@@ -1,0 +1,1 @@
+addFilter('macro-in-comment')
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -2,6 +2,7 @@
 packages:
   vars:
     foremandist: .fm2_1
+    rpmlint_rc_file: "{{ inventory_dir }}/.rpmlintrc"
   children:
     foreman_packages: {}
     plugins_packages: {}


.rpmlintrc
```diff
@@ -0,0 +1 @@
+addFilter('macro-in-comment')
```
ekohl, line 1: IMHO we should clean these up but I'm ok with merging it if it gives us other lint checks in the short term. Don't we currently run rpmlint now?

ehelms: We do, this hides this warning from the output. I'd prefer a model of hiding warnings via rpmlint config and then removing them as we/if we want to fix them.

ekohl: Can we make warnings fatal? Currently warnings don't cause a build failure so they're ignored, even when shown.

ehelms: Does not look like we can. But we can at least hide this output since I think there is an actual error that can surface related to macros being expanded in comments.

ekohl: Can we check if stdout/stderr is empty perhaps?

ehelms: Probably can check for and error out in Obal if a warning exists easily. This however, I think we want to ignore all together since we use macros in comments for documentation and we'd have to remove that use everywhere.

PR #5040, opened 2020-04-14
Add python-scikit-build package

evgeni:


evgeni:
it's a build dependency for pulp-rpm (and its deps)

PR #5042, opened 2020-04-14
Update rubygem-gettext to remove .po and samples folders

zjhuntin:


packages/foreman/rubygem-gettext/rubygem-gettext.spec
```diff
@@ -85,8 +85,6 @@ find %{buildroot}%{gem_instdir}/bin -type f | xargs chmod a+x
 %{gem_instdir}/bin
 %{gem_libdir}
 %{gem_instdir}/locale
-%{gem_instdir}/po
```
ekohl, line 88: Will this pass or do you need to use `%exclude`?

zjhuntin: probably need to exclude

ehelms:
[test rpm]

zjhuntin:
[test rpm]

zjhuntin:
[test rpm]

PR #5043, opened 2020-04-14
Have rubyforge require the correct json lib from ruby

zjhuntin:


ekohl:
Why did repoclosure fail here on the old RPM? It should be using the one from `downloaded_rpms_repo`, right?
```
[2020-04-14T20:57:16.249Z]     Added downloaded_rpms repo from ./downloaded_rpms/rhel7
[2020-04-14T20:57:16.249Z]     package: tfm-rubygem-rubyforge-2.0.4-9.el7.noarch from el7-foreman-nightly
[2020-04-14T20:57:16.250Z]       unresolved deps:
[2020-04-14T20:57:16.250Z]         tfm-rubygem(json) >= 1.1.7
```

PR #5047, opened 2020-04-15
Fixes #29535 - Exclude devtoolset

stejskalleos:

### DIFF
--- a/mock/el8.cfg
+++ b/mock/el8.cfg
@@ -2,7 +2,7 @@ config_opts['root'] = 'epel-8-x86_64'
 config_opts['target_arch'] = 'x86_64'
 config_opts['legal_host_arches'] = ('x86_64',)
 
-config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep --exclude=devtoolset*'
+config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
 config_opts['dist'] = 'el8'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '8'
 config_opts['package_manager'] = 'dnf'
--- a/mock/el8.cfg
+++ b/mock/el8.cfg
@@ -2,7 +2,7 @@ config_opts['root'] = 'epel-8-x86_64'
 config_opts['target_arch'] = 'x86_64'
 config_opts['legal_host_arches'] = ('x86_64',)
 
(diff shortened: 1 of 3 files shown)


mock/el8.cfg
```diff
@@ -2,7 +2,7 @@ config_opts['root'] = 'epel-8-x86_64'
 config_opts['target_arch'] = 'x86_64'
 config_opts['legal_host_arches'] = ('x86_64',)
 
-config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
+config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep --exclude=devtoolset*'
```
ekohl, line 5: I'm not sure devtoolset ships on EL8. @ehelms @pcreech @zjhuntin @evgeni any idea?

ehelms:
Does not fix my issues :/

mock/el8.cfg
pcreech: Looks like it is not available on el8 at the moment.

ekohl:
@ehelms is there any broken cache present perhaps? Maybe remove `/var/lib/mock/*`

ehelms:
I tried cleaning all of that. I'm good with this change going in just wanted to note it still plagues me.

mock/el8.cfg
ehelms: Not to my knowledge there isn't.

ekohl:
So with this you still get the libelf issue? https://github.com/rpm-software-management/mock/pull/268 disables sclo but not a good solution since we need that. https://github.com/rpm-software-management/mock/pull/268#issuecomment-534384263 talks about what provides it and lists `elfutils-libelf`. Perhaps we should explicitly install that to force it?

ehelms:
Tried all of that, no dice. Maybe Fedora 32 will provide me better luck.

stejskalleos:
@ehelms I also have problem with missing `libelf.so.1`, only fix that worked for me was adding `config_opts['yum_install_command'] += ' --disablerepo sclo*'` to `el7-scl.cfg`.

ehelms:
@stejskalleos Oddly, I tried that tactic a week back and did not work for me. Tried it today and boom I got past the error! Thanks! Mind including that here?

stejskalleos:
@ehelms I think we should not add it as default options, as @ekohl wrote in previous comment:

> ... disables sclo but not a good solution since we need that.

ehelms:
@stejskalleos From all my testing, that option does not break anything and only fixes things. This has the potential to be a common issue more folks run into as newer Fedora's release and folks upgrade. I think it is safe to include it.

ekohl:
@ehelms I thought sclo was used for the Red Hat SCLs. devtoolset comes from the RH SCLs.

ehelms:
This doesn't appear to break my ability to build SCL packages. And I believe that is because this command is run during chroot setup and is not a permanent change within the build environment for yum itself.

PR #5053, opened 2020-04-16
rubygem-sassc %check updates

evgeni:
This takes the Fedora RPM spec approach of using the native libsass
instead of the bundled version.


ekohl:
This is very similar to https://github.com/theforeman/foreman-packaging/pull/5019

evgeni:
it is, and it fails on el8 too :D

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -1,12 +1,14 @@
 # template: scl
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
+%{!?_root_libdir:%global _root_libdir %{_libdir}}
```
evgeni, line 4: So this ends up as `/usr/lib`, not as `/usr/lib64` and it explodes from there.

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -49,7 +58,14 @@ gem unpack %{SOURCE0}
 gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
 %{?scl:EOF}
 
+# disable building bundled libsass
+sed -i "/s\.extensions/d" %{gem_name}.gemspec
+
 %build
+# use libsass.so.0 from host
+sed -i "s/libsass\.\#{dl_ext}/libsass\.\#{dl_ext}\.0/" lib/sassc/native.rb
+sed -i "s!__dir__!\"%{_root_libdir}\"!" lib/sassc/native.rb
```
evgeni, line 67: When I don't use a macro, but `/usr/lib64` here, it works…

packages/foreman/rubygem-sassc/rubygem-sassc.spec
ekohl: https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/ suggests that `%{_libdir}` should be `/usr/lib64` on 64 bit machines. Is that different on EL8?

evgeni: No, `rpm -E '%{_libdir}` does lib64 on my F31, EL7 and EL8. But Koji for some reason ends up with lib. :woman_shrugging:

evgeni: on both, el7 and el8 build that is!

ekohl: @pcreech any idea?

evgeni: And to make matters worse, it is lib64 when I run it in mock with our el7-scl config.

ekohl: That explains why it worked when I debugged it locally using mock.

evgeni: maybe because it's a noarch build? :)

evgeni: yepp, friggin noarch

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -22,9 +25,16 @@ Requires: %{?scl_prefix}rubygem(ffi) < 2
...
 BuildRequires: %{?scl_prefix_ruby}rubygems-devel
+BuildRequires: %{?scl_prefix}rubygem(ffi) >= 1.9
+BuildRequires: %{?scl_prefix}rubygem(ffi) < 2
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 # end specfile generated dependencies
 
+BuildRequires: libsass
+Requires: libsass
+# We explicitly point it to libsass.so.0 in %build
+Requires: libsass.so.0()
```
evgeni, line 36: @ekohl what is this? :)
It's not in the Fedora spec (https://src.fedoraproject.org/rpms/rubygem-sassc/blob/master/f/rubygem-sassc.spec) which does similar shenanigans to use the system libsass.

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -66,44 +84,42 @@ mkdir -p %{buildroot}%{gem_dir}
...
-cp %{buildroot}/opt/theforeman/tfm/root/usr/share/gems/gems/sassc-%{version}/ext/libsass.so .%{gem_extdir_mri}/
-
-cp -a .%{gem_extdir_mri}/gem.build_complete %{buildroot}%{gem_extdir_mri}/
-cp -a .%{gem_extdir_mri}/*.so %{buildroot}%{gem_extdir_mri}/%{gem_name}/
-
-# Prevent dangling symlink in -debuginfo (rhbz#878863)
-# But this gem fails on it with error
-# File not found: /builddir/build/BUILDROOT/tfm-rubygem-sassc-2.2.1-1.el7.x86_64/opt/theforeman/tfm/root/usr/share/gems/gems/sassc-2.2.1/ext.
-# rm -rf %{buildroot}%{gem_instdir}/ext/
+%check
```
evgeni, line 87: given this package does not actually contain a native extension, we don't need the mkdir/copy dance and can use `ruby -I "%{buildroot}%{gem_libdir}" -e "require '%{gem_name}'"` here. do we want that? I'd say yes.

packages/foreman/rubygem-sassc/rubygem-sassc.spec
ekohl: It was my thinking weird idea to ensure that `libsass.so.0` really does exist. If libsass is bumped to `.so.1` then it would break. Perhaps there's another mechanism to really require this.

ekohl: Maybe the braces are actually incorrect and it should just be `libsass.so.0`

packages/foreman/rubygem-sassc/rubygem-sassc.spec
ekohl: I think that's good to verify the hack we've applied.

evgeni: You mean to try to "require 'sassc'"? Yeah, that definitely (even tho it does not fail on EL7 even when it can't find libsass.so - but does so on EL8 :woman_shrugging: )

I was more talking about the syntax of the check. Simple, or in line with the other packages?

packages/foreman/rubygem-sassc/rubygem-sassc.spec
evgeni: I tried both, two sets of braces and no braces, and still no cigar, so I dropped this for now.

evgeni:
it's funny that on EL8 it still doesn't find libsass :exploding_head:

packages/foreman/rubygem-sassc/rubygem-sassc.spec
ekohl: So I'll be a bit more verbose this time: I think it's good to very the gem actually works. If there's no native paths, there's no need to copy the directory I think.

evgeni: simplified, thanks

PR #5058, opened 2020-04-17
Switch to using nulldb for tasks that need a database

ehelms:
@tbrisker @mmoll @ekohl @evgeni 

I had a successful Foreman + foreman_ansible build with using the nulldb adapter. This shows it working for RPM build process and can be cleaned up. I wanted y'all to see it is possible and the changes I had to make.

The assumptions we make today are that `database.yml.example` has a basic sqlite database defined by it. If we drop Sqlite (which I think we should to avoid having to keep SQL compatiable with it in our applications and drop more out of our test matrix) we will need to decide what the example and default database.yml deployed is (perhaps "basic" postgresql configuration is the right answer).

Note that nulldb requires a schema.rb file to exist, but the contents of that file do not matter.

I would suggest that we also modify CI to run asset precompile and apipie cache testing using nulldb to ensure that it continues to work rather than catching it down in the RPM layer if a problem arises.

This has a nice advantage as well of not needing to do any database migrating.

To summarize:

 * nulldb approach works
 * nulldb feels clean and reduces some of the RPM build overhead
 * we should test this in CI IMO and should decide if we want to include it in foreman core codebase or just how I have it here for build tasks

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -9,7 +9,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 11
+%global release 12
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -243,6 +243,7 @@ BuildRequires: %{?scl_prefix}rubygem(jwt) < 2.3.0
 BuildRequires: %{?scl_prefix}rubygem(graphql) >= 1.8.0
 BuildRequires: %{?scl_prefix}rubygem(graphql) < 1.9.0
 BuildRequires: %{?scl_prefix}rubygem(graphql-batch)
+BuildRequires: %{?scl_prefix}rubygem-activerecord-nulldb-adapter
 # end specfile main BuildRequires
 
 # assets
@@ -542,13 +543,15 @@ Group: Development/Libraries
 Requires: %{name} = %{version}-%{release}
 Requires: %{name}-build = %{version}-%{release}
 Requires: %{name}-sqlite = %{version}-%{release}
+Requires: %{?scl_prefix}rubygem(activerecord-nulldb-adapter)
 
 %description plugin
 Meta package with support for plugins.
 
 %files plugin
 %{_sysconfdir}/rpm/macros.%{name}-plugin
-%{_datadir}/%{name}/schema_plugin.rb
+%{_datadir}/%{name}/schema.rb.nulldb
+%{_datadir}/%{name}/bundler.d/nulldb.rb
 
 %package build
 Summary: Foreman package RPM support
@@ -727,8 +730,7 @@ make -C locale all-mo
 
 #use Bundler_ext instead of Bundler
 mv Gemfile Gemfile.in
-cp config/database.yml.example config/database.yml
-cp config/settings.yaml.example config/settings.yaml
+cp db/schema.rb.nulldb db/schema.rb
 export BUNDLER_EXT_GROUPS="default assets"
 ln -s %{?scl_prefix_nodejs:%{_scl_root}}%{nodejs_sitelib} node_modules
 # Calls webpack manually since webpack:compile uses the config which uses
@@ -737,10 +739,9 @@ export NODE_ENV=production
 %{?scl:scl enable %{scl} "}
 webpack --bail --config config/webpack.config.js
 %{?scl:"}
-%{scl_rake} assets:precompile RAILS_ENV=production --trace
-%{scl_rake} db:migrate db:schema:dump RAILS_ENV=production --trace
-%{scl_rake} apipie:cache RAILS_ENV=production cache_part=resources --trace
-rm config/database.yml config/settings.yaml
+%{scl_rake} assets:precompile RAILS_ENV=production DATABASE_URL=nulldb://nohost --trace
+%{scl_rake} apipie:cache RAILS_ENV=production cache_part=resources DATABASE_URL=nulldb://nohost --trace
+rm db/schema.rb
 
 %install
 %if 0%{?fedora} || 0%{?rhel} >= 8
@@ -807,7 +808,7 @@ for i in orchestrator worker; do
 done
 
 # Put db in %{_localstatedir}/lib/%{name}/db
-cp -pr db/migrate db/seeds.rb db/seeds.d %{buildroot}%{_datadir}/%{name}
+cp -pr db/schema.rb.nulldb db/migrate db/seeds.rb db/seeds.d %{buildroot}%{_datadir}/%{name}
 mkdir %{buildroot}%{_localstatedir}/lib/%{name}/db
 
 ln -sv %{_localstatedir}/lib/%{name}/db %{buildroot}%{_datadir}/%{name}/db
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -59,12 +59,14 @@ foreman_scl_packages:
         - "jenkins_job=foreman-develop-source-release"
         - "jenkins_url=https://ci.theforeman.org"
       build_package_tito_builder: "tito.builder.FetchBuilder"
+      nightly_githash: '1'
```
ekohl, line 62: Debug

mock/el7-scl.cfg
```diff
@@ -6,6 +6,7 @@ config_opts['dist'] = 'el7'  # only useful for --resultdir variable subst
 # beware RHEL use 7Server or 7Client
 config_opts['releasever'] = '7'
 config_opts['package_manager'] = 'yum'
+config_opts['yum_install_command'] += ' --disablerepo sclo*'
```
ekohl, line 9: Mind separating this to a separate commit? https://github.com/theforeman/foreman-packaging/pull/5047 was intended to solve the same issue

packages/foreman/foreman/foreman.spec
```diff
@@ -245,14 +245,16 @@ BuildRequires: %{?scl_prefix}rubygem(graphql) < 1.9.0
...
+
 # assets
 %if 0%{?scl:1}
 BuildRequires: %{scl}-runtime-assets >= 6
 BuildRequires: %{scl}-runtime-assets < 7
 %else
 BuildRequires: nodejs-packaging
 %endif
-BuildRequires: %{?scl_prefix_nodejs}nodejs >= 6.10
+BuildRequires: %{?scl_prefix_nodejs}nodejs
```
ekohl, line 257: I actually would like to set this to the correct version (is that 10 now?)

packages/foreman/foreman/foreman.spec
```diff
@@ -725,7 +728,17 @@ make -C locale all-mo
...
+cat <<DATABASE > config/database.yml
+production:
+  adapter: nulldb
+development:
+  adapter: nulldb
+DATABASE
+cat <<DB_SCHEMA > db/schema.rb
+ActiveRecord::Schema.define(version: 1) do
+end
+DB_SCHEMA
```
ekohl, line 741: As you suggested, I think it'd be good if this lived in Foreman itself as `bundler.d/nulldb.rb` and `config/database.yml.nulldb`. Not sure about the schema.

packages/foreman/foreman/foreman.spec
```diff
@@ -792,7 +804,7 @@ find %{buildroot}%{_datadir}/%{name} -type d -name "test" |xargs rm -rf
 rm -f %{buildroot}%{_datadir}/%{name}/bin/spring
 
 # Move config files to %{_sysconfdir}
-mv %{buildroot}%{_datadir}/%{name}/config/database.yml.example %{buildroot}%{_datadir}/%{name}/config/database.yml
+#mv %{buildroot}%{_datadir}/%{name}/config/database.yml %{buildroot}%{_datadir}/%{name}/config/database.yml
```
ekohl, line 807: This is IMHO still good for fresh installs to keep

packages/foreman/foreman/foreman.spec
```diff
@@ -885,8 +897,11 @@ cp -r %%{%{name}_dir} ./%{_datadir} || echo 0 \\
 mkdir -p ./%{_localstatedir}/lib/%{name} \\
 cp -r %{_localstatedir}/lib/%{name}/db ./%{_localstatedir}/lib/%{name} || echo 0 \\
 unlink ./%{_datadir}/%{name}/db \\
-ln -sv \`pwd\`/%{_localstatedir}/lib/%{name}/db ./%{_datadir}/%{name}/db \\
 pushd ./%%{%{name}_dir} \\
+mkdir db/ \\
+cp -rf %{_datadir}/%{name}/db/* db/ \\
```
ekohl, line 902: Doesn't this work:
```suggestion
cp -rf %{_datadir}/%{name}/db ./ \\
```

mock/el7-scl.cfg
ehelms: This is not the final commit. I just kept things in here I need present to allow me to test and re-test.

packages/foreman/foreman/foreman.spec
ehelms: This is purely to allow testing this.

packages/foreman/foreman/foreman.spec
ehelms: I don't know. But this may change depending on how much of this nulldb code we put into Foreman. I wouldn't focus too much on the code itself as much as the "process" that is required. I'll open a Foreman PR with some of the proposed idea changes to help.

evgeni:
Oh, one thing, the nulldb readme does not mention AR 6.0, do you know if it still works with Rails6?

tbrisker:
I like this! 

> Oh, one thing, the nulldb readme does not mention AR 6.0, do you know if it still works with Rails6?

Looks like they forgot to update the readme, but version 0.4.0 does seem to include AR 6 support: https://github.com/nulldb/nulldb/tree/v0.4.0

ohadlevy:
nice - would you mind fixing the dockerfile too? :)

ehelms:
This currently includes a fix related to https://github.com/theforeman/foreman-packaging/pull/5078 but I'd prefer that #5078 go in first and then I update this one.

packages/foreman/foreman/foreman.spec
```diff
@@ -252,7 +253,7 @@ BuildRequires: %{scl}-runtime-assets < 7
 %else
 BuildRequires: nodejs-packaging
 %endif
-BuildRequires: %{?scl_prefix_nodejs}nodejs >= 6.10
+BuildRequires: %{?scl_prefix_nodejs}nodejs >= 10
```
evgeni, line 256: doesn't feel like it belongs here but :woman_shrugging:

packages/foreman/foreman/foreman.spec
```diff
@@ -885,22 +883,23 @@ cp -r %%{%{name}_dir} ./%{_datadir} || echo 0 \\
...
+mkdir db/ \\
+cp -rf %{_datadir}/%{name}/db/* db/ \\
+mv db/schema.rb.nulldb db/schema.rb
 \\
 ln -s %{?scl_prefix_nodejs:%{_scl_root}}%{nodejs_sitelib} node_modules \\
 export GEM_PATH=%%{buildroot}%%{gem_dir}:\${GEM_PATH:+\${GEM_PATH}}\${GEM_PATH:-\`%{?scl:scl enable %%{scl} -- }ruby -e "print Gem.path.join(':')"\`} \\
 unlink tmp \\
 \\
 rm \`pwd\`/config/initializers/encryption_key.rb \\
+rm \`pwd\`/config/database.yml \\
```
evgeni, line 896: what would have created this file?

ehelms: This line copies the example to `database.yml` so there is a default: https://github.com/theforeman/foreman-packaging/pull/5058/files#diff-3399e5c03dc282964481f8f3ced9c141R796

When we drop SQlite we need to re-think this but for now we are just switching the nulldb usage.

evgeni:
@ehelms #5078 is **IN**

packages/foreman/foreman/foreman.spec
evgeni: that link goes nowhere, stupid GH?

tbrisker:
We need source-release to finish, 
`cp: cannot stat 'db/schema.rb.nulldb': No such file or directory`

mmoll:
[test rpm]

PR #5061, opened 2020-04-20
Revert unf_ext to 0.0.7.2 as that's what ms_rest_azure requires

evgeni:


packages/foreman/rubygem-unf_ext/rubygem-unf_ext.spec
```diff
@@ -104,7 +104,10 @@ rm -rf gem_ext_test
 %{gem_instdir}/test
 %{gem_instdir}/unf_ext.gemspec
 
-%changelog
+%changeloga
```
ekohl, line 107:
```suggestion
%changelog
```

evgeni:
scratch builds:
  - http://koji.katello.org/koji/taskinfo?taskID=324651
  - http://koji.katello.org/koji/taskinfo?taskID=324652

(because jenkins hates time going backwards)

packages/foreman/rubygem-unf_ext/rubygem-unf_ext.spec
```diff
@@ -105,6 +105,9 @@ rm -rf gem_ext_test
 %{gem_instdir}/unf_ext.gemspec
 
 %changelog
+* Mon Apr 20 2020 Evgeni Golov 0.0.7.2-3
```
ekohl, line 108: Does this work? Going back and forth in a changelog? Shouldn't we just revert the git commit and untag the new version?

evgeni: Sure, why not. I still want the changes that came in for the spec, just not the version. So I could drop the older changelog entry, but I don't think it will give us anything.

ekohl: Fair enough.

PR #5068, opened 2020-04-21
Fix node gyp repoclosure

ehelms:
There is multiple commits here, designed to address the EL8 repoclosure issues we are seeing.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -50,6 +50,7 @@ foreman_scl_packages:
         - el8-devel
         - el8-epel
         - el8-extras
+        - el8-powertools
         - el8-puppet-6
   hosts:
     foreman:
@@ -1072,6 +1073,7 @@ repoclosures:
           - el8-devel
           - el8-epel
           - el8-extras
+          - el8-powertools
           - el8-puppet-6
     plugins-repoclosure-el7:
       repoclosure_target_repos:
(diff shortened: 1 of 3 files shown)


packages/foreman/nodejs-react-popper/nodejs-react-popper.spec
```diff
@@ -4,31 +4,30 @@
...
-Requires: %{?scl_prefix}npm(create-react-context) < 0.3.0
-Requires: %{?scl_prefix}npm(popper.js) >= 1.14.1
-Requires: %{?scl_prefix}npm(popper.js) < 2.0.0
-Requires: %{?scl_prefix}npm(prop-types) >= 15.6.1
-Requires: %{?scl_prefix}npm(prop-types) < 16.0.0
-Requires: %{?scl_prefix}npm(typed-styles) >= 0.0.5
-Requires: %{?scl_prefix}npm(typed-styles) < 1.0.0
-Requires: %{?scl_prefix}npm(warning) >= 4.0.0
-Requires: %{?scl_prefix}npm(warning) < 5.0.0
+Requires: npm(@babel/runtime) >= 7.1.2
```
ekohl, line 17: Is the non-SCL part intentional here?

ehelms: nope, that came from the script

PR #5069, opened 2020-04-21
Update Remote execution spec to build for el8

zjhuntin:


packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -1,15 +1,18 @@
 # template: foreman_plugin
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
+%{!?_root_sbindir:%global _root_sbindir %{_sbindir}}
+%{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
 
+%global debug_package %{nil}
```
ekohl, line 7: For my understanding: why is this needed?

zjhuntin:
[test rpm]

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
evgeni: `noarch` packages should not produce any debug symbols, so it should not be needed. but with el8 you never know.

PR #5075, opened 2020-04-21
Add more repos to the plugins_scl_packages lookaside for el8

zjhuntin:


package_manifest.yaml
```diff
@@ -1048,10 +1048,13 @@ plugin_nonscl_packages:
...
         - el8-baseos
-        - el8-appstream
+        - el8-devel
+        - el8-epel
         - el8-extras
+        - el8-powertools
         - el8-puppet-6
-        - el8-devel
+        - el8-foreman-nightly
+        - el8-foreman-plugins-nightly
```
ehelms, line 1057: This is a target repository so we do not want it to be a lookaside as well.

PR #5078, opened 2020-04-22
Only use scl for dynflow if available

ehelms:

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -9,7 +9,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 10
+%global release 11
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman/foreman.spec
```diff
@@ -778,7 +778,9 @@ install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
 install -Dp -m0644 extras/systemd/%{name}.service %{buildroot}%{_unitdir}/%{name}.service
 install -Dp -m0644 extras/systemd/%{name}.socket %{buildroot}%{_unitdir}/%{name}.socket
 
+%if %{?scl:1}%{!?scl:0}
```
evgeni, line 781:
```suggestion
%if 0%{?scl:1}
```

Isn't that more readable?

And yeah, probably also made simpler in "replace shebangs and binaries in scripts for SCL" section :)

actually, why don't we do this step in the replace shebangs part of `%build`? :)

ekohl: > actually, why don't we do this step in the replace shebangs part of `%build`? :)

Good question and feel free to blame me for approving it. I've also wondered the same thing in the past.

PR #5084, opened 2020-04-22
Fix react-popper dependency on typed-styles

evgeni:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -319,6 +319,7 @@ foreman_nodejs_packages:
         - el8-epel
         - el8-extras
         - el8-puppet-6
+        - el8-foreman-nightly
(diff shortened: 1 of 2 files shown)


evgeni:
eh, why does that fail closure?! halp

evgeni:
because it didn't look at foreman at all -.-

evgeni:
According to @ehelms this is another rabbit hole we don't want right now

ehelms:
Ya... we *should* update it, but the build issues I hit were strange and a rabbit hole as @evgeni mentions. I'd rather us tackle that when things have calmed down and stick with what we know works.

ehelms:
Filed a ticket to track: https://projects.theforeman.org/issues/29599

PR #5085, opened 2020-04-22
Fixes #29601: Drop foreman-release-scl

ehelms:


ehelms:
Manual updates -- https://github.com/theforeman/theforeman.org/pull/1571

ehelms:
Koji mash update: https://github.com/theforeman/foreman-infra/pull/1343

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -40,6 +40,8 @@ BuildArch: noarch
 
 BuildRequires: sed
 
+Obsoletes: foreman-release-scl < 7-3
```
ekohl, line 43: Just wondering if everyone who installed `foreman-release-scl` also installed `foreman-release`. Perhaps it'd be safer to obsolete it in `tfm`?

ehelms: Yea, suppose so. Dunno what we rae gonna do when we get to pure EL8 and we don't have a tfm package as a catch all for cleaning up user systems.

ekohl: Then we'll have less things to obsolete since unused gems are no longer in our own SCL so they could be used by other tools. We can introduce an optional foreman-obsoletes similar to how Fedora removes obsolete packages.

ehelms:
Related fix: https://github.com/theforeman/forklift/pull/1156

ekohl:
This might cause some issues in the puppet modules, but since those currently are red I can't verify it it. I'm working on making them green and hopefully say something useful.

ehelms:
@ekohl Are the issues blocking this still on going? Thinking this is a change we want prior to branching if we are pulling it in this release.

ekohl:
Currently puppet-foreman (and puppet-foreman_proxy) rely on this in the acceptance tests because they use `foreman::repo`. That would need a change. Given all the things that are going on I'm wondering if we should push this out to the next release. We don't really win that much and I struggle to find time to even to do the critical things.

ehelms:
@ekohl are we ready for this change yet?

ehelms:
Updated https://github.com/theforeman/foreman-infra/pull/1343 -- @evgeni could you take a look?

PR #5100, opened 2020-04-23
Create foreman-nulldb package

mmoll:

### DIFF
--- a/plugins/ruby-foreman-tasks/debian/changelog
+++ b/plugins/ruby-foreman-tasks/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-tasks (1.1.0-1) stable; urgency=low
+
+  * 1.1.0 released
+
+ -- Michael Moll <kvedulv@kvedulv.de>  Thu, 23 Apr 2020 17:45:18 +0200
+
 ruby-foreman-tasks (1.0.1-1) stable; urgency=low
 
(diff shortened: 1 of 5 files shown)


debian/bionic/foreman/control
```diff
@@ -54,15 +54,15 @@ Description: metapackage providing PostgreSQL dependencies for Foreman
  .
  The foreman package should be installed along with this package.
 
-Package: foreman-sqlite3
+Package: foreman-nulldb
```
ekohl, line 57: I feel that we should mirror these subpackage names between RPM and debs. In RPM we have `foreman-build`. Would it make sense to do the same on debs?

mmoll: Sure :heavy_check_mark:

ekohl: @theforeman/packaging this is correct, right?

evgeni:
The f-build RPM contains build macros and stuff like that. The nulldb bundler file and schema is in foreman-plugin which is "this is what you need installed to build a plugin".

mmoll:
I think all that doesn't apply to DEB, so I'm inclined to name it "foreman-nulldb" again... Any strong arguments against this?

The plugin build seems to work just fine.

ekohl:
In that case I think `foreman-nulldb` is fine.

PR #5109, opened 2020-04-27
pulp-rpm and dependencies

evgeni:
EL8 has python3-createrepo_c, python3-productmd and python3-libcomps packages, so we don't really need to build them there. Thoughts?

Edit: pulp requires newer versions than EL8 provides :(


evgeni:
[test rpm]

evgeni:
this is now using fedora-based packaging for createrepo/libcomps/libsolv as those are not really python packages, just happen to have python bindings.

PR #5116, opened 2020-04-28
Added foreman_datacenter

Zenya:

PR #5121, opened 2020-04-28
foreman_datacenter 1.24.1

Zenya:


Zenya:
@evgeni hi
I decided to add older version to rpm before move to 2.0.0 (PR #4926)
so as you recommended:
I removed %{foreman_apipie_cache_plugin} 
fixed homepage in gemspec file
created new gem file
created new release

rel-eng/tito.props
```diff
@@ -530,3 +530,8 @@ whitelist = katello
   rubygem-runcible
   rubygem-typhoeus
   rubygem-zest
+
+
+[foreman-plugins-nightly-rhel7]
```
evgeni, line 535: please add it to the `foreman-plugins-1.24-rhel7` section instead, roughly line 333

evgeni:
I'd like to first get it into develop in #5116, but then 2.0 and 1.24 can follow quickly :)

evgeni:
this is quite outdated, I'll close it and follow up with what we have in #5335

PR #5126, opened 2020-04-29
new pulp deps for 3.3

evgeni:
3.3 depends on tablib and that pulls in a big chunk of deps (which we probably will never need, but hey)


packages/pulpcore/python-tablib/python-tablib.spec
```diff
@@ -0,0 +1,79 @@
...
+Summary:        Format agnostic tabular data library (XLS, JSON, YAML, CSV)
+
+License:        MIT
+URL:            https://tablib.readthedocs.io
+Source0:        https://files.pythonhosted.org/packages/source/t/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
+BuildArch:      noarch
+
+BuildRequires:  python3-devel
+BuildRequires:  python3-markuppy
+BuildRequires:  python3-markuppy
```
ekohl, line 16: Odd that so many lines are duplicated

evgeni: Yeah. It's because tablib actually mentions those multiple times in `setup.py` :(

PR #5128, opened 2020-04-29
Build nodejs-htmlparser2 for EL8

ehelms:


packages/plugins/nodejs-htmlparser2/nodejs-htmlparser2.spec
```diff
@@ -28,6 +28,7 @@ Source13: nodejs-htmlparser2-%{version}-registry.npmjs.org.tgz
 BuildRequires: %{?scl_prefix_nodejs}npm
 %else
 BuildRequires: nodejs-packaging
+BuildRequires: npm
```
ekohl, line 31: I still think this is an ugly template: https://github.com/theforeman/npm2rpm/pull/64#discussion_r403508777

ehelms: I don't know what an ugly template means. But `npm` the command is required per our spec file, e.g.:

https://github.com/theforeman/foreman-packaging/pull/5128/files#diff-b201095925b1d659b12ef2fc00b17222R66

ekohl: Because you specify npm twice: once for SCL and once for non-SCL.

PR #5115, opened 2020-04-28
Update foreman package dependencies

ekohl:

PR #5137, opened 2020-04-30
Update coffee-rails and rails-i18n

zjhuntin:


evgeni:
this rightfuly fails repoclosure, I guess this needs to go hand-in-hand with a core/foreman.spec change?

zjhuntin:
@evgeni yes, I think so

tbrisker:
yes, spec change is in https://github.com/theforeman/foreman-packaging/pull/5115 that is blocked on this, and this matches the versions in core.

PR #5142, opened 2020-04-30
Update roadie-rails to 2.1.1 & rack to 2.2.2

zjhuntin:


ehelms:
I think you can exclude the two files causing issues (and add them to the ignores for teh template in gem2rpm):

```
   /opt/theforeman/tfm/root/usr/share/gems/gems/rack-2.2.2/CONTRIBUTING.md
   /opt/theforeman/tfm/root/usr/share/gems/gems/rack-2.2.2/SPEC.rdoc
```

PR #5140, opened 2020-04-30
Fixes #29677 - Properly stop dynflowd on upgrade

jlsherrill:

PR #5146, opened 2020-04-30
Properly stop dynflowd on upgrade

jlsherrill:


ekohl:
This is the development version of https://github.com/theforeman/foreman-packaging/pull/5140. For background, see https://github.com/theforeman/foreman-packaging/pull/5140#pullrequestreview-403843415

PR #5159, opened 2020-05-04
Update rubygem-foreman_scc_manager to 1.8.4

m-bucher:
see https://github.com/ATIX-AG/foreman_scc_manager/pull/39


ekohl:
Looks like it's failing to build:
`I18n::InvalidLocale: :en is not a valid locale`

tbrisker:
Looks like this is caused by the upgrade to the i18n gem to version 1.8.2 as part of the rails upgrade. Not sure why this is failing here but not elsewhere, but we might be seeing this in other plugins if they are rebuilt on top of rails 6

m-bucher:
@tbrisker, anything I can do to fix this?

ehelms:
Katello rebuilt a few times since the upgrade. I tested another plugin `foreman_virt_who_configure` as well which makes this even more odd.

tbrisker:
[test rpm] just to make sure we aren't chasing a wild goose here...

ehelms:
I'm wondering if we are seeing something related to this bit of code:

https://github.com/ATIX-AG/foreman_scc_manager/blob/master/lib/foreman_scc_manager/engine.rb#L102

tbrisker:
> I'm wondering if we are seeing something related to this bit of code:
> 
> https://github.com/ATIX-AG/foreman_scc_manager/blob/master/lib/foreman_scc_manager/engine.rb#L102

That's common to all plugins - e.g. https://github.com/theforeman/foreman_virt_who_configure/blob/master/lib/foreman_virt_who_configure/engine.rb#L114-L119

tbrisker:
for the record, the failure now is on repoclosure, which is looking at the previous version of the rpm for some reason.

PR #5161, opened 2020-05-04
Refs #29603 - katello-selinux now requires candlepin-selinux

jturel:


jturel:
This is because of https://github.com/Katello/katello-selinux/pull/22

jturel:
[test rpm]

ehelms:
You'll need to git annex add the new source

jturel:
oh right :)

PR #5166, opened 2020-05-05
drop python-scikit-build package

evgeni:
This reverts commit d789376223d29138960368f461d2ff874cd60e95.

We don't need the package after all.


evgeni:
I've untagged the builds already

PR #5167, opened 2020-05-05
Update openscap dependency to match EL8

evgeni:


evgeni:
@zjhuntin this should fix your repoclosure issues you've seen

ekohl:
It would be nice if https://github.com/OpenSCAP/ruby-openscap stated something about openscap version compatibility.

evgeni:
> It would be nice if https://github.com/OpenSCAP/ruby-openscap stated something about openscap version compatibility.

The best I can give you is https://github.com/OpenSCAP/ruby-openscap/blob/master/lib/openscap/openscap.rb#L7

But I don't know how to write that as a (Build)Requires without sacrificing a kitten.

PR #5168, opened 2020-05-05
Add epel repo to el8 plugin lookaside repos

zjhuntin:


evgeni:
Funnily (?) we just removed it last week (https://github.com/theforeman/foreman-packaging/pull/5135) as @ehelms had issues with epel8 providing some gems in the wrong version and thus freaking out repoclosure.

PR #5169, opened 2020-05-05
Add Candlepin repo to foreman lookasides

zjhuntin:


ekohl:
This doesn't feel right. Foreman does not depend on Candlepin and thus shouldn't be using the lookaside IMHO.

ehelms:
Installer requires katello-selinux which requires candlepin-selinux which
lives in candlepin repo.

ekohl:
Yes, but before we didn't have this. We've always tried to keep Foreman dependencies minimal. Also, the instructions to install theforeman.org don't mention Candlepin at all. This introduces the risk that something starts to depend on candlepin. Granted, it's a small risk but IMHO we should look for a way to revert this.

Should `candlepin-selinux` be tagged in foreman instead?

evgeni:
Especially, this now requires one to sign and mash katello, before one can release foreman. Which is backwards in the very sense of the word. :sob:

evgeni:
Oh, and the new dependency made katello-selinux uninstallable on Koji, so we can't build plugins that depend on Katello anymore:

http://koji.katello.org/kojifiles/work/tasks/951/340951/root.log

ehelms:
The further joys of that puppet SELinux bug. 

I do think we could drop rubygem-katello requiring katello-selinux and leave that up to the installer packaging to handle. I think that still puts a candlepin-selinux burden on the release workflow. If we need to just put all the selinux packages in the main foreman repository then so be it.

PR #5170, opened 2020-05-06
Require glibc-langpack-en in foreman-postgresql

ekohl:
On EL8 and Fedora the locales are not present by default. Since the installer installs with the en_US.UTF-8 locale, this is needed in most cases and doesn't hurt in others.


ekohl:
To answer a question from @evgeni stated offline: this is on foreman-postgresql rather than foreman-installer because you can install Foreman without the installer while still using the Puppet modules.

ekohl:
Replaced by https://github.com/theforeman/foreman-packaging/pull/5171 because I pushed it from the wrong branch.

PR #5171, opened 2020-05-06
Require glibc-langpack-en in foreman-postgresql

ekohl:
On EL8 and Fedora the locales are not present by default. Since the installer installs with the en_US.UTF-8 locale, this is needed in most cases and doesn't hurt in others.

It's in foreman-postgresql rather than foreman-installer since users can install Foreman without the installer. For example, using the Puppet modules directly.

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -9,7 +9,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 19
+%global release 20
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman/foreman.spec
```diff
@@ -987,6 +994,9 @@ exit 0
 %systemd_postun_with_restart %{name}.service
 
 %changelog
+* Wed May 06 2020 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl>
```
ehelms, line 997: The version and release are missing at the end here

ekohl: Odd - I used obal for it.

ehelms:
I'm good with this. This won't cover Postgres on Proxies where Pulpcore is present but we can address that in pulpcore packaging as well?

ekohl:
Yes, I think that requires a similar trick.

packages/foreman/foreman/foreman.spec
evgeni: And outdated version? It had bugs around fetching those on Py3 installs **whistle**

ekohl: I first had a macro error (`%end` instead of `%endif`) so it probably failed to parse the version out of it.

PR #5172, opened 2020-05-06
Update foreman required version for rubygem-katello

zjhuntin:


zjhuntin:
@ekohl I added the upper bound, but wanted to make sure this is what you were thinking.

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -24,6 +24,7 @@ Requires: katello-selinux
 Requires: foreman-postgresql
 # start specfile generated dependencies
 Requires: foreman >= %{foreman_min_version}
+Requires: foreman < 2.2
```
ehelms, line 27: Could be good to follow the same pattern for foreman_max_version

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -234,6 +235,9 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/webpack
 
 %changelog
+* Wed May 06 2020 Zach Huntington-Meath <zhunting@redhat.com> - 3.16.0-0.12.pre.master
+- Update foreman minimum version
```
ekohl, line 239: How about
```suggestion
- Update foreman version bounds
```

packages/katello/rubygem-katello/rubygem-katello.spec
ekohl: That, or perhaps you can do some math so it becomes `%{foreman_min_version} + 0.1`. Not sure if RPM specs allow this.

zjhuntin:
[test rpm]

zjhuntin:
[test rpm]

evgeni:
this needs a rebase now, sorry

ehelms:
[test rpm]

zjhuntin:
[test rpm]

zjhuntin:
[test rpm]

packages/plugins/ansiblerole-insights-client/insights-client-1.7.1.tar.gz
```diff
@@ -0,0 +1 @@
+../../../.git/annex/objects/V8/WV/SHA256E-s11781--6c22e8ea717b86e5e77a42b2a42dfd2ab6b9beae0171bb78ff230bae6e9c322b.tar.gz/SHA256E-s11781--6c22e8ea717b86e5e77a42b2a42dfd2ab6b9beae0171bb78ff230bae6e9c322b.tar.gz
```
ehelms, line 1: Don't think you meant t include this file with this update

ehelms:
Ah ha! So I don't think you need el8-katello-nightly in the lookasides, its just all around missing from here:

https://github.com/theforeman/foreman-packaging/blob/rpm/develop/repoclosure/yum.conf

PR #5174, opened 2020-05-06
Update dynflow dependency to 1.4.4

adamruzicka:


mmoll:
merged, díky @adamruzicka!

PR #5186, opened 2020-05-07
Update foreman_remote_execution to 3.2.0

adamruzicka:

### DIFF
--- a/plugins/ruby-foreman-remote-execution/debian/changelog
+++ b/plugins/ruby-foreman-remote-execution/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-remote-execution (3.2.0-1) stable; urgency=low
+
+  * 3.2.0 released
+
+ -- Adam Ruzicka <aruzicka@redhat.com>  Thu, 07 May 2020 22:29:59 +0200
+
 ruby-foreman-remote-execution (3.1.0-2) stable; urgency=low
 
   * Switch build to nulldb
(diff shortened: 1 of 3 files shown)


plugins/ruby-foreman-remote-execution/debian/control
```diff
@@ -3,7 +3,8 @@ Section: ruby
 Priority: extra
 Maintainer: Stephen Benjamin <stephen@redhat.com>
 Build-Depends: debhelper (>= 8.0.0), git | git-core, foreman-assets (>= 1.20.0~rc1), foreman (>= 1.25.0~rc1),
-  foreman-nulldb (>= 1.25.0~rc1), ruby-foreman-tasks (>= 0.15.1), ruby-foreman-deface
+  foreman-nulldb (>= 1.25.0~rc1), ruby-foreman-tasks (>= 0.15.1), ruby-foreman-deface,
+  ruby-dynflow (>= 1.0.2), ruby-dynflow (<< 2.0.0)
```
mmoll, line 7: For core plugins, dynflow is provided by the foreman (core) package, so this line needs to get removed.

adamruzicka: So if we need a minimum version (1.0.2), we have to hope foreman (core) package provides it?

mmoll: Yes, but if the core package in the build process doesn't bring it, the build would fail. So we detect this at least with the latest core package that's available for the target branch.

mmoll: BTW, 1.0.2 was release two years ago, so I'd say this is super-safe. ;)

mmoll:
merged, díky @adamruzicka!

PR #5190, opened 2020-05-11
Add configmanagement Ansible repository to repoclosure

ehelms:

### DIFF
--- a/repoclosure/yum.conf
+++ b/repoclosure/yum.conf
@@ -26,6 +26,12 @@ enabled=1
 mirrorlist=http://mirrorlist.centos.org/?release=8&arch=x86_64&repo=AppStream
 failovermethod=priority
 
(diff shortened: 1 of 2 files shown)


repoclosure/yum.conf
```diff
@@ -26,6 +26,13 @@ enabled=1
 mirrorlist=http://mirrorlist.centos.org/?release=8&arch=x86_64&repo=AppStream
 failovermethod=priority
 
+[el8-configmanagement-ansible]
+name=Ansible configmanagement
+enabled=1
+#mirrorlist=http://mirror.centos.org/centos/8/configmanagement/x86_64/ansible-29/
```
adamruzicka, line 32: Should this be kept here?

ehelms: Nah, dropped it.

PR #5191, opened 2020-05-11
python3-createrepo_c needs same version of createrepo_c

ehelms:
```
   Problem: package python3-pulp-rpm-3.3.1-1.el8.noarch requires python3-createrepo_c >= 0.15.10, but none of the providers can be installed
    - conflicting requests
    - nothing provides createrepo_c-libs = 0.15.10-1.el8 needed by python3-createrepo_c-0.15.10-1.el8.x86_64
```

@evgeni Seems we have to go this route due to tight coupling

### DIFF
--- a/comps/comps-katello-pulpcore-el8.xml
+++ b/comps/comps-katello-pulpcore-el8.xml
@@ -10,6 +10,9 @@
     <description>pulpcore Packages</description>
     <uservisible>true</uservisible>
     <packagelist>
+      <packagereq type="default">createrepo_c</packagereq>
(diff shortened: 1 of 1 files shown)


evgeni:
Oh, I thought I added those to comps... But yes, this is certainly needed!

evgeni:
We might also need the py2 packages if pulp2 needs them?

ehelms:
That's a good question, that I don't know the answer for on EL7. This is just updating El8 which won't see Pulp 2 on it. I will reach out and check on the EL7 side.

PR #5192, opened 2020-05-11
Add foreman max version for rubygem-katello

ehelms:


ehelms:
@tbrisker the backport

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -22,6 +23,7 @@ Requires: katello-selinux
 Requires: foreman-postgresql
 # start specfile generated dependencies
 Requires: foreman >= %{foreman_min_version}
+Requires: foreman < %{foreman_max_version}
```
ekohl, line 26: Just when you press submit you see it: this line will be removed in the next update since it's not part of the template. Needs to be above `# start`. If this is a cherry pick, it will also show up in develop soon.

PR #5193, opened 2020-05-11
Add puppetlabs-stdlib

zjhuntin:


packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,76 @@
+%global commit 45454b8de4ff8f860f6f78438107133e510336ed
+%global shortcommit %(c=%{commit}; echo ${c:0:7})
+
+Name:		puppetlabs-stdlib
+Version:	4.25.1
```
ekohl, line 5: I'd prefer not to ship such an old version.

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,76 @@
...
+Release:	2%{?dist}
+Summary:	Puppet Labs Standard Library
+License:	ASL 2.0
+URL:		https://github.com/puppetlabs/puppetlabs-stdlib
+Source0:	https://github.com/puppetlabs/puppetlabs-stdlib/archive/%{commit}/puppetlabs-stdlib-%{shortcommit}.tar.gz
```
ekohl, line 10: How about using the one from the forge, which is the official build:
```suggestion
Source0:	https://forge.puppet.com/v3/files/%{name}-%{version}.tar.gz
```

This also means `%prep` needs to be modified to be more standard.

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,76 @@
...
+mkdir -p %{buildroot}/%{_datadir}/puppet/modules/stdlib
+cp -rp lib/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/lib
+cp -rp spec/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/spec
+cp -rp manifests/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/manifests
+cp -rp types/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/types
+cp -p metadata.json %{buildroot}/%{_datadir}/puppet/modules/stdlib/
+
+
+%files
+%doc MAINTAINERS.md CONTRIBUTING.md CHANGELOG.md README.md README_DEVELOPER.markdown
```
ehelms, line 32: Could split these into individual lines to match our typical standards

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
ehelms: We can update to 5.2 since latest foreman_scap_client requires `< 6`. https://github.com/theforeman/puppet-foreman_scap_client/blob/master/metadata.json#L42

If this starts a discussion of updating foreman_scap_client I am going to vote instead to take this as is to match EL7 and then discuss updating this further.

ekohl: https://github.com/theforeman/puppet-foreman_scap_client/pull/61

zjhuntin:
@ekohl @ehelms updated and building for 5.2

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,78 @@
...
+mkdir -p %{buildroot}/%{_datadir}/puppet/modules/stdlib
+cp -rp lib/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/lib
+cp -rp spec/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/spec
+cp -rp manifests/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/manifests
+cp -rp types/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/types
+cp -p metadata.json %{buildroot}/%{_datadir}/puppet/modules/stdlib/
+
+
+%files
+%doc MAINTAINERS.md
```
ehelms, line 30: I think we can safely exclude this file.

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,78 @@
...
+cp -rp lib/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/lib
+cp -rp spec/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/spec
+cp -rp manifests/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/manifests
+cp -rp types/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/types
+cp -p metadata.json %{buildroot}/%{_datadir}/puppet/modules/stdlib/
+
+
+%files
+%doc MAINTAINERS.md 
+%doc CONTRIBUTING.md
```
ehelms, line 31: I think we can safely exclude this file.

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,78 @@
...
+cp -rp types/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/types
+cp -p metadata.json %{buildroot}/%{_datadir}/puppet/modules/stdlib/
+
+
+%files
+%doc MAINTAINERS.md 
+%doc CONTRIBUTING.md 
+%doc CHANGELOG.md 
+%doc README.md 
+%doc README_DEVELOPER.markdown
```
ehelms, line 34: I think we can safely exclude this file.

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,78 @@
...
+cp -rp manifests/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/manifests
+cp -rp types/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/types
+cp -p metadata.json %{buildroot}/%{_datadir}/puppet/modules/stdlib/
+
+
+%files
+%doc CHANGELOG.md 
+%doc README.md 
+%license LICENSE
+%exclude /MAINTAINERS.md
```
ekohl, line 33: The `/` is a bit odd here. I think that since you're not copying them in `%install`, you don't need to mention these.

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,78 @@
...
+
+%prep
+%setup -qn %{name}-%{version}
+
+%build
+
+%install
+mkdir -p %{buildroot}/%{_datadir}/puppet/modules/stdlib
+cp -rp lib/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/lib
+cp -rp spec/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/spec
```
ekohl, line 23: I'd consider leaving these out.

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,78 @@
...
+%description
+Puppet Labs Standard Library module.
+
+%prep
+%setup -qn %{name}-%{version}
+
+%build
+
+%install
+mkdir -p %{buildroot}/%{_datadir}/puppet/modules/stdlib
```
ekohl, line 21: `%{_datadir}` includes a leading slash so I this will suffice:
```suggestion
mkdir -p %{buildroot}%{_datadir}/puppet/modules/stdlib
```

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,78 @@
...
+cp -rp lib/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/lib
+cp -rp spec/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/spec
+cp -rp manifests/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/manifests
+cp -rp types/ %{buildroot}/%{_datadir}/puppet/modules/stdlib/types
+cp -p metadata.json %{buildroot}/%{_datadir}/puppet/modules/stdlib/
```
ekohl, line 26: You can also consider combining these:
```suggestion
cp -rp lib/ manifests/ types/ metadata.json %{buildroot}%{_datadir}/puppet/modules/stdlib/
```

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
```diff
@@ -0,0 +1,78 @@
...
+%doc README.md 
+%license LICENSE
+%exclude /MAINTAINERS.md 
+%exclude /CONTRIBUTING.md 
+%exclude /README_DEVELOPER.markdown
+%{_datadir}/puppet/modules/stdlib
+
+%changelog
+* Mon May 11 2020 <zhunting@redhat.com> - 5.2.0-1
+- Build in foreman-packaging
```
ekohl, line 40: Feels like this doesn't quite do it justice

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
zjhuntin: it failed to build because it didn't have the `/`

packages/plugins/puppetlabs-stdlib/puppetlabs-stdlib.spec
zjhuntin: ack

zjhuntin:
@ekohl good catch, thank you! I think it's fixed now I extracted the new rpm and found the locales directory.

ehelms:
I think we could follow up with an EL7 update for parity.

PR #5194, opened 2020-05-11
Fdi service update

zjhuntin:


ekohl:
Looks like it's not built as an SCL package.

packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
```diff
@@ -1,14 +1,14 @@
...
+Release: 3%{?dist}
 Summary: Metapackage with dependencies for FDI
 
 Group: Applications/System
 License: GPLv2+
 URL: https://github.com/theforeman/foreman-discovery-image
 
 Requires:	foreman-proxy
-Requires:	tfm-rubygem-smart_proxy_discovery_image
+Requires:	rubygem-smart_proxy_discovery_image
```
ehelms, line 11: This still needs the scl_prefix on EL7. The problem is that we build this into the nonscl tag so the scl_prefix macro is not available. I think you will need to do an if-else check here on the version of the rhel macro.

PR #5210, opened 2020-05-14
Update rubygem-foreman_remote_execution to 3.2.1

adamruzicka:


evgeni:
@adamruzicka my non-educated guess is: this should also go into 2.1, could you please open a cherry pick PR for that?

adamruzicka:
Here it is https://github.com/theforeman/foreman-packaging/pull/5242

PR #5229, opened 2020-05-15
Fix statedir and shebang for smart_proxy_salt

adamruzicka:


ehelms:
[test rpm]

ehelms:
This repoclosure failure would indicate we need to update the Requires on python to the python from the module. The salt-master requirement appears to be a bit more problematic as I do not see it in the configmanagement repo for CentOS or in EPEL.

m-bucher:
This fix is included in #5371, so I guess this one can be closed, @adamruzicka!?

adamruzicka:
That's true, closing.

PR #5237, opened 2020-05-17
Drop the shebang in ruby2ruby

ekohl:
This file has a /usr/bin/ruby shebang, but doesn't need it since rubygems generates a wrapper in %{_bindir}. In an SCL environment this results in a dependency on /usr/bin/ruby while it's actually only using the SCL ruby.

### DIFF
--- a/packages/foreman/rubygem-ruby2ruby/rubygem-ruby2ruby.spec
+++ b/packages/foreman/rubygem-ruby2ruby/rubygem-ruby2ruby.spec
@@ -7,7 +7,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 2.4.2
-Release: 2%{?dist}
+Release: 3%{?dist}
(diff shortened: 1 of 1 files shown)


ekohl:
Note to self: node.rb was then broken for me because I removed Ruby 2.0 altogether.

ekohl:
The installer ensures rubygem-json is installed so that shouldn't be an issue on fresh installs - just yum remove will break things. We can also tell users to rerun the installer.

packages/foreman/rubygem-ruby2ruby/rubygem-ruby2ruby.spec
```diff
@@ -55,6 +55,8 @@ gem unpack %{SOURCE0}
 gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
 %{?scl:EOF}
 
+sed -i 1d bin/r2r_show
```
evgeni, line 58: can we have a comment what/why we're deleting the line here?

PR #5197, opened 2020-05-12
Fixes #29737 - Put smart proxy plugin local state into /var/lib/foreman-proxy

adamruzicka:

PR #5240, opened 2020-05-18
Fixes #29737 - Put smart proxy plugin local state into /var/lib/foreman-proxy

evgeni:
cherry pick of #5197


evgeni:
repoclosure unrelated

PR #5251, opened 2020-05-18
Drop the shebang in ruby2ruby

ekohl:
This file has a /usr/bin/ruby shebang, but doesn't need it since rubygems generates a wrapper in %{_bindir}. In an SCL environment this results in a dependency on /usr/bin/ruby while it's actually only using the SCL ruby.

(cherry picked from commit bbbfe9c24d495caf92821db18266f4da5084683a)


evgeni:
tagged

PR #5255, opened 2020-05-19
Fixes #29883 - SELinux wrapper for sidekiq service

lzap:
Sidekiq service must be started under foreman_rails_t domain which can
only be achieved using a shell wrapper and proper SELinux label.

[x] Develop
[x] 2.1

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -9,7 +9,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 21
+%global release 22
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -703,9 +703,9 @@ plugins required for Foreman to work.
   done
   # script content
   sed -ri 'sX/usr/bin/rakeX%{scl_rake}X' extras/dbmigrate script/foreman-rake
-
-  sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|/usr/bin/scl enable tfm "sidekiq \1"|' extras/systemd/%{dynflow_sidekiq_service_name}.service
 %endif
+# sidekiq service SELinux helper path update
+sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|%{_libexecdir}/%{name}/sidekiq-selinux \1|' extras/systemd/%{dynflow_sidekiq_service_name}.service
 
 #build locale files
 make -C locale all-mo
@@ -752,6 +752,7 @@ install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}/tmp/pids
 install -d -m0755 %{buildroot}%{_localstatedir}/run/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}/plugins
+install -d -m0755 %{buildroot}%{_libexecdir}/%{name}
 #Copy init scripts and sysconfigs
 install -Dp -m0644 extras/systemd/%{dynflow_sidekiq_service_name}.service %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
 install -Dp -m0755 script/%{name}-debug %{buildroot}%{_sbindir}/%{name}-debug
@@ -762,6 +763,16 @@ install -Dp -m0644 %{SOURCE4} %{buildroot}%{_sysconfdir}/cron.d/%{name}
 install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
 install -Dp -m0644 extras/systemd/%{name}.service %{buildroot}%{_unitdir}/%{name}.service
 install -Dp -m0644 extras/systemd/%{name}.socket %{buildroot}%{_unitdir}/%{name}.socket
+
+# SELinux libexec wrappers
+cat > %{buildroot}%{_libexecdir}/%{name}/sidekiq-selinux <<EOF
+#!/bin/bash
+# Shell wrapper with SELinux transition into foreman_rails_t domain.
+%if 0%{?scl:1}
+source scl_source enable %{scl}
+%endif
+exec sidekiq "\$@"
+EOF
 
 cp -p Gemfile.in %{buildroot}%{_datadir}/%{name}/Gemfile.in
 cp -p -r app bin bundler.d config config.ru extras lib locale Rakefile script webpack .babelrc.js %{buildroot}%{_datadir}/%{name}
@@ -934,6 +945,7 @@ rm -rf %{buildroot}
 %{_sbindir}/%{name}-rake
 %{_sbindir}/%{name}-tail
 %{_mandir}/man8
+%attr(755,root,root) %{_libexecdir}/%{name}/*
 %config(noreplace) %{_sysconfdir}/%{name}
 %ghost %attr(0640,root,%{name}) %config(noreplace) %{_sysconfdir}/%{name}/encryption_key.rb
 %config(noreplace) %{_sysconfdir}/logrotate.d/%{name}
@@ -998,6 +1010,9 @@ exit 0
 %systemd_postun_with_restart %{name}.service
 
 %changelog
+* Thu May 21 2020 Lukas Zapletal <lzap+rpm@redhat.com> - 2.2.0-0.22.develop
+- Added SELinux wrapper for sidekiq
+
 * Thu May 21 2020 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> - 2.2.0-0.21.develop
 - Update Gem and NPM dependencies
 


lzap:
Rebased, bumped release version.

packages/foreman/foreman/foreman-sidekiq
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec /usr/bin/scl enable tfm "sidekiq $*"
```
ekohl, line 2: Rather than this wrapper, can't we use `/usr/bin/tfm-ruby` which basically already does this? Can we call `/opt/theforeman/tfm/root/usr/bin/sidekiq` and put that in the right context?

lzap: We cannot, unfortunately. This would mean every Ruby application would end up in `foreman_rails_t` domain, which we don't want to.

Your second suggestion is possible, however again this means every single sidekiq instance would run in our Foreman RoR domain which might not be wanted.

The wrapper is the cleanest solution, I used `exec` so there is not another subprocess hanging in the `ps` output.

lzap:
I see a repoclosure error which seems unrelated?

```
[2020-05-19T10:11:35.530Z]     package: rubygem-hammer_cli-2.2.0-0.1.pre.develop.20200518090133git919dc54.el8.noarch from el8-foreman-nightly
[2020-05-19T10:11:35.530Z]       unresolved deps:
[2020-05-19T10:11:35.530Z]         rubygem(amazing_print)
```

lzap:
> I don't think this will work on EL8 where we don't use SCL.

Good find. I haven't realized that at all.

packages/foreman/foreman/foreman-sidekiq
ekohl: > We cannot, unfortunately. This would mean every Ruby application would end up in `foreman_rails_t` domain, which we don't want to.

I meant that a better wrapper is doing something like (untested):
```suggestion
exec /usr/bin/tfm-ruby "sidekiq $*"
```

The benefit is that we can use this on both EL7 and EL8 since `/usr/bin/tfm-ruby` is a constant on all platforms. Not sure it will find sidekiq though.

lzap:
How about this?

packages/foreman/foreman/foreman-sidekiq
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec /usr/bin/sidekiq $*
```
ekohl, line 2: I'd be concerned about spaces in arguments:
```suggestion
exec /usr/bin/sidekiq "$@"
```

packages/foreman/foreman/foreman.spec
```diff
@@ -691,6 +693,9 @@ plugins required for Foreman to work.
   SYSCONFDIR=%{_sysconfdir} \
   --trace
 
+#use SELinux wrapper for sidekiq
+sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|/usr/share/foreman/bin/foreman-sidekiq \1|' extras/systemd/%{dynflow_sidekiq_service_name}.service
```
ekohl, line 697: I'd use the same macros as you use in the `install` line so you're sure they're in sync.

packages/foreman/foreman/foreman.spec
```diff
@@ -758,6 +763,7 @@ install -Dp -m0644 %{SOURCE4} %{buildroot}%{_sysconfdir}/cron.d/%{name}
 install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
 install -Dp -m0644 extras/systemd/%{name}.service %{buildroot}%{_unitdir}/%{name}.service
 install -Dp -m0644 extras/systemd/%{name}.socket %{buildroot}%{_unitdir}/%{name}.socket
+install -Dp -m0755 %{SOURCE6} %{buildroot}%{_datadir}/%{name}/bin/%{name}-sidekiq
```
ekohl, line 766: `/usr/share/foreman/bin` is an odd place for this. Granted, we already have some precedent.

lzap: Yeah, the ideal place is `/var/lib/libexec/foreman` or something like that (for binaries which are not meant for end users). I can do this if you insist.

packages/foreman/foreman/foreman.spec
lzap: Sorry what macros? Not following.

lzap:
> Generally we want to keep platforms in sync

I agree in general, but I thought that the other extreme is introducing wrapper for debian when it's not necessary. But it's probably cleaner.

lzap:
Rebased, this now needs https://github.com/theforeman/foreman/pull/7666 to be merged in order to pass tests.

packages/foreman/foreman/foreman.spec
ekohl: I don't insist in this PR. If we do that, we should move everything from that bin dir to `/usr/libexec/foreman`. That probably takes a bit more coordinated effort.

packages/foreman/foreman/foreman.spec
```diff
@@ -699,8 +700,8 @@ plugins required for Foreman to work.
   done
   # script content
   sed -ri 'sX/usr/bin/rakeX%{scl_rake}X' extras/dbmigrate script/foreman-rake
-
-  sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|/usr/bin/scl enable tfm "sidekiq \1"|' extras/systemd/%{dynflow_sidekiq_service_name}.service
+  # sidekiq wrapper
+  sed -i '/^exec/ s|/usr/bin/sidekiq "\(.\+\)"$|/usr/bin/scl enable tfm "/usr/bin/sidekiq \1"|' extras/%{name}-sidekiq
```
ekohl, line 704: The whole point of the SCL wrapper is that you don't need to specify the full path. `/usr/bin/sidekiq` doesn't exist, but after activating the SCL env it is in `$PATH`.
```suggestion
  sed -i '/^exec/ s|/usr/bin/sidekiq "\(.\+\)"$|/usr/bin/scl enable tfm "sidekiq \1"|' extras/%{name}-sidekiq
```

However, this still removes the quotes from `"$@"` which I'm not sure actually works.

packages/foreman/foreman/foreman.spec
```diff
@@ -758,6 +759,7 @@ install -Dp -m0644 %{SOURCE4} %{buildroot}%{_sysconfdir}/cron.d/%{name}
 install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
 install -Dp -m0644 extras/systemd/%{name}.service %{buildroot}%{_unitdir}/%{name}.service
 install -Dp -m0644 extras/systemd/%{name}.socket %{buildroot}%{_unitdir}/%{name}.socket
+install -Dp -m0755 extras/%{name}-sidekiq %{buildroot}%{_datadir}/%{name}/bin/%{name}-sidekiq
```
ekohl, line 762: On line 765 we copy the entire bin directory. This makes the line redundant if you keep it in `bin/` in foreman.git.

packages/foreman/foreman/foreman.spec
lzap: Honestly, I don't like solving artificial problems. We don't quote in the systemd unit either, there are no spaces and very unlikely will be. BUT since you asked:

```
[root@sat68 ~]# scl enable tfm -- ruby -e "puts :hello, :world"
hello
world

[root@sat68 ~]# scl enable tfm 'ruby -e "puts :hello, :world"'
hello
world

[root@sat68 ~]# scl enable tfm -- 'ruby -e "puts :hello, :world"'
/var/tmp/sclDm2GG6: line 8: ruby -e puts: command not found
```

I tend to prefer `--` syntax which I often use for git. I will refactor it this way.

packages/foreman/foreman/foreman.spec
lzap: Oh good catch. I made sure attributes are correct.

lzap:
Rebased.

lzap:
For the record it fails now with `File not found: /builddir/build/BUILDROOT/foreman-2.2.0-0.21.develop.20200519143528gita68ffcb.el8.noarch/usr/share/foreman/bin/foreman-sidekiq` because core PR needs to be merged first.

lzap:
Wait with merging, I need to actually create second wrapper also for rails app. We cannot launch it this way, otherwise this rule is needed: `corecmd_exec_shell(foreman_rails_t)` - this effectively allows spawning ANY shell command. Thist must be avoided, otherwise SELinux policy is pretty much useless.

Feel free to review the wrapper for sidekiq and I will do the same for rails too. Creating a ticket for this as I will need to backport this into 2.1: https://projects.theforeman.org/issues/29883

lzap:
Wait I need to investigate, it's the `/usr/bin/scl` what spawns `/bin/bash /var/tmp/sclIV6Zq0` not us. This needs discussion with the SCL team.

ekohl:
> Wait I need to investigate, it's the `/usr/bin/scl` what spawns `/bin/bash /var/tmp/sclIV6Zq0` not us. This needs discussion with the SCL team.

I told you this in https://community.theforeman.org/t/puma-production-deployment-port/18818/8. That's why I'm suggesting to use tfm-ruby. It does not spawn a shell.

You can also copy the activation code from it:
https://github.com/theforeman/foreman-packaging/blob/1e5327eff2b7c2211186c028f0ece58ef57b3e23/packages/foreman/tfm/tfm.spec#L247-L251

lzap:
That's what I did not understand, let me try this. FYI I reported this to SCL guys: https://github.com/sclorg/scl-utils/issues/31

lzap:
Ok I suggest to create this helper in libexec directory. We would keep those which are distributed with Foreman core in `/usr/share/foreman/bin` (this includes sidekiq helper) but those we create only for purpose of SCL would go into libexec what you think?

lzap:
Ok I created RM ticket and also added SELinux wrapper for TFM Ruby (Rails):

```diff
diff --git a/packages/foreman/foreman/foreman.spec b/packages/foreman/foreman/foreman.spec
index d7eeedcf6..aff16b175 100644
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -700,6 +700,8 @@ plugins required for Foreman to work.
   done
   # script content
   sed -ri 'sX/usr/bin/rakeX%{scl_rake}X' extras/dbmigrate script/foreman-rake
+  # ruby on rails wrapper
+  sed -i '/^ExecStart/ s|/usr/share/foreman/bin/rails \(.\+\)$|/usr/lib/libexec/%{name}/%{scl_name}-rails \1|' extras/systemd/%{name}.service
   # sidekiq wrapper
   sed -i '/^exec/ s|/usr/bin/sidekiq \(.\+\)$|/usr/bin/scl enable tfm -- sidekiq \1|' extras/sidekiq
 %endif
@@ -749,6 +751,7 @@ install -d -m0755 %{buildroot}%{_localstatedir}/lib/%{name}/tmp/pids
 install -d -m0755 %{buildroot}%{_localstatedir}/run/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}
 install -d -m0750 %{buildroot}%{_localstatedir}/log/%{name}/plugins
+install -d -m0755 %{buildroot}%{_libexecdir}/%{name}
 #Copy init scripts and sysconfigs
 install -Dp -m0644 extras/systemd/%{dynflow_sidekiq_service_name}.service %{buildroot}%{_unitdir}/%{dynflow_sidekiq_service_name}.service
 install -Dp -m0755 script/%{name}-debug %{buildroot}%{_sbindir}/%{name}-debug
@@ -765,6 +768,14 @@ cp -p -r app bin bundler.d config config.ru extras lib locale Rakefile script we
 rm -rf %{buildroot}%{_datadir}/%{name}/extras/{jumpstart,spec}
 find %{buildroot}%{_datadir}/%{name}/script/%{name}-tail.d/* -type d |xargs rm -rf
 
+# SCL libexec wrappers
+%if 0%{?scl:1}
+  cat >> %{buildroot}%{_libexecdir}/%{name}/%{scl_name}-rails <<-EOF
+  #!/bin/bash
+  source scl_source enable %{scl_name}
+  exec ruby "\$@"
+EOF
+
 # remove all test units from produciton release
 find %{buildroot}%{_datadir}/%{name} -type d -name "test" |xargs rm -rf
 
@@ -931,6 +942,7 @@ rm -rf %{buildroot}
 %{_sbindir}/%{name}-rake
 %{_sbindir}/%{name}-tail
 %{_mandir}/man8
+%attr(755,root,root) %{_libexecdir}/%{name}/*
 %config(noreplace) %{_sysconfdir}/%{name}
 %ghost %attr(0640,root,%{name}) %config(noreplace) %{_sysconfdir}/%{name}/encryption_key.rb
 %config(noreplace) %{_sysconfdir}/logrotate.d/%{name}
```

This is now ready for review and testing.

packages/foreman/foreman/foreman.spec
```diff
@@ -699,8 +699,10 @@ plugins required for Foreman to work.
   done
   # script content
   sed -ri 'sX/usr/bin/rakeX%{scl_rake}X' extras/dbmigrate script/foreman-rake
-
-  sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|/usr/bin/scl enable tfm "sidekiq \1"|' extras/systemd/%{dynflow_sidekiq_service_name}.service
+  # ruby on rails wrapper
+  sed -i '/^ExecStart/ s|/usr/share/foreman/bin/rails \(.\+\)$|/usr/libexec/%{name}/%{scl_name}-rails \1|' extras/systemd/%{name}.service
```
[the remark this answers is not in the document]
adamruzicka: The original line is
```
ExecStart=/usr/share/foreman/bin/rails server --environment $FOREMAN_ENV --port $FOREMAN_PORT --binding $FOREMAN_BIND
```
After this `sed` it becomes
```
ExecStart=/usr/libexec/%{name}/%{scl_name}-rails server --environment $FOREMAN_ENV --port $FOREMAN_PORT --binding $FOREMAN_BIND
```

So the libexec wrapper is run, and in turn runs
```
ruby server --environment $FOREMAN_ENV --port $FOREMAN_PORT --binding $FOREMAN_BIND
```
which fails with
```
May 20 11:22:04 foreman-selinux.example.com foreman[7657]: ruby: No such file or directory -- server (LoadError)
```

Either the sed should add another `rails` before the captured group or `/usr/libexec/tfm-rails` should run rails instead of just ruby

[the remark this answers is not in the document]
lzap: Yeah that was a copy and paste error. Rails helper should call rails. Fixed as well as the libexec macro use.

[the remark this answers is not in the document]
ekohl: Why is this helper needed? Why can't the existing `bin/rails` be used with the right label? Using bash wrappers was an absolute pain when trying to keep the file descriptors. Keeping everything in pure Ruby avoided forking

lzap:
I have updated this PR according to explanation in https://github.com/theforeman/foreman-selinux/pull/100#issuecomment-632018235

Core PR is no longer needed as explained in the comment.

packages/foreman/foreman/foreman.spec
```diff
@@ -759,6 +762,22 @@ install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
...
 install -Dp -m0644 extras/systemd/%{name}.socket %{buildroot}%{_unitdir}/%{name}.socket
 
+# SCL libexec wrappers
+%if 0%{?scl:1}
+  cat > %{buildroot}%{_libexecdir}/%{name}/%{scl_name}-ruby-selinux <<-EOF
+  #!/bin/bash
+  # Shell wrapper for %{scl_name} with SELinux transition into foreman_rails_t domain.
+  source scl_source enable %{scl_name}
+  exec ruby "\$@"
+EOF
```
adamruzicka, line 772: Wouldn't this lead to all the lines being prefixed with two empty spaces and possibly messing up the shebang?

lzap:
Fixed.

packages/foreman/foreman/foreman.spec
```diff
@@ -759,6 +762,33 @@ install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
...
+  #!/bin/bash
+  # Shell wrapper with SELinux transition into foreman_rails_t domain.
+  source scl_source enable %{scl_name}
+  exec ruby "\$@"
+EOF
+  cat > %{buildroot}%{_libexecdir}/%{name}/sidekiq-selinux <<-EOF
+  #!/bin/bash
+  # Shell wrapper with SELinux transition into foreman_rails_t domain.
+  source scl_source enable %{scl_name}
+  exec sidekiq "\$@"
```
adamruzicka, line 777: The built packages contain unexpanded `%{scl_name}`. Is it the right macro?

```
[root@foreman-selinux foreman]# cat /usr/libexec/foreman/ruby-selinux
  #!/bin/bash
  # Shell wrapper with SELinux transition into foreman_rails_t domain.
  source scl_source enable %{scl_name}
  exec ruby "$@"
[root@foreman-selinux foreman]# /usr/libexec/foreman/ruby-selinux -e 'puts "Hello"'
Can't read /etc/scl/conf/%{scl_name}, %{scl_name} is probably not installed.
Hello
```

ekohl: It's just `%{scl}`

lzap:
> I would prefer to standardize everything on `/usr/libexec/foreman` and that includes Debian.

We cannot for the reason I explained. Startup script `rails` must be in the RAILS_DIR/bin directory. I could put everything into RAILS_DIR/bin but I don't want to. Let's only keep scripts which are required there, the rest should go into libexec.

> This would avoid patching the service files altogether (which we do everywhere).

Not possible. And I've spend hours today to find out! Let me explain - since we use systemd notify mechanism to start the service, any wrapper scripts which exec, or other indirections in ExecStart, may result in activated socket file descriptors being closed before they reach the puma master process.

More details here:

* https://github.com/puma/puma/blob/master/docs/systemd.md#socket-activation
* https://github.com/puma/puma/issues/1499
* https://serverfault.com/questions/960132/puma-with-systemd-and-socket-activation-errnoeaddrinuse

We would need to create a special program that would keep file descriptors open or alternatively we could use TCP sockets for systemd socket activation instead. Both to me look like an overkill.

> Then the script should not have a `-selinux` suffix. I'd suggest two wrappers: `/usr/libexec/foreman/{rails,sidekiq}`.

I suggest to be explicit here, these binaries has SELinux exec label associated to them, therefore any start will cause the process to fail. With this suffix, we are telling the user "hey watch out you are starting something special". But if you insist, I can remove this.

packages/foreman/foreman/foreman.spec
lzap: That's indeed incorrect, let me check.

lzap: Replaced with scl_prefix that should do it.

ekohl:
>> I would prefer to standardize everything on `/usr/libexec/foreman` and that includes Debian.
>
> We cannot for the reason I explained. Startup script `rails` must be in the RAILS_DIR/bin directory. I could put everything into RAILS_DIR/bin but I don't want to. Let's only keep scripts which are required there, the rest should go into libexec.

Does it need to be there or does `$CWD` need to be there? Note in our `foreman-rake` wrapper we `cd` there (see `/usr/sbin/foreman-rake` on an installed system).

> > This would avoid patching the service files altogether (which we do everywhere).
> 
> Not possible. And I've spend hours today to find out! Let me explain - since we use systemd notify mechanism to start the service, any wrapper scripts which exec, or other indirections in ExecStart, may result in activated socket file descriptors being closed before they reach the puma master process.

I know this exactly because I've been through exactly the same. I don't understand why we need these additional wrappers anyway. We have `/usr/share/foreman/bin/rails` and that can be labeled, right? It already has the correct shebang (`/usr/bin/tfm-ruby` on EL7 and `/usr/bin/ruby` on EL8).

> We would need to create a special program that would keep file descriptors open or alternatively we could use TCP sockets for systemd socket activation instead. Both to me look like an overkill.

That doesn't make sense to me. We have TCP sockets opened by systemd. That passes on the file descriptors. The whole reason we have this is that systemd keeps the socket open. That means we can restart foreman.service without dropping any requests from Apache. If you don't do that Apache returns HTTP 502 for a while and breaks upgrades.

packages/foreman/foreman/foreman.spec
```diff
@@ -759,6 +762,33 @@ install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
...
+# SELinux libexec wrappers
+%if 0%{?scl:1}
+  cat > %{buildroot}%{_libexecdir}/%{name}/ruby-selinux <<-EOF
+  #!/bin/bash
+  # Shell wrapper with SELinux transition into foreman_rails_t domain.
+  source scl_source enable %{scl_prefix}
+  exec ruby "\$@"
+EOF
+  cat > %{buildroot}%{_libexecdir}/%{name}/sidekiq-selinux <<-EOF
+  #!/bin/bash
```
adamruzicka, line 774: I'm getting

```
systemd[1]: Started Foreman jobs daemon - orchestrator on sidekiq.
systemd[2852]: Failed at step EXEC spawning /usr/libexec/foreman/sidekiq-selinux: Exec format error
systemd[1]: dynflow-sidekiq@orchestrator.service: main process exited, code=exited, status=203/EXEC
systemd[1]: Unit dynflow-sidekiq@orchestrator.service entered failed state.
systemd[1]: dynflow-sidekiq@orchestrator.service failed.
```

removing the leading whitespace from the shebang fixes it

packages/foreman/foreman/foreman.spec
```diff
@@ -759,6 +762,33 @@ install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
...
+  # Shell wrapper with SELinux transition into foreman_rails_t domain.
+  source scl_source enable %{scl_prefix}
```
adamruzicka, line 776: Looks like `%{scl_prefix}` contains a trailing dash, `%{scl}` should be just the name
```
systemd[1]: Started Foreman jobs daemon - orchestrator on sidekiq.
dynflow-sidekiq@orchestrator[2898]: Can't read /etc/scl/conf/tfm-, tfm- is probably not installed.
systemd[1]: dynflow-sidekiq@orchestrator.service: main process exited, code=exited, status=127/n/a
systemd[1]: Unit dynflow-sidekiq@orchestrator.service entered failed state.
systemd[1]: dynflow-sidekiq@orchestrator.service failed.
```

lzap:
> Does it need to be there or does `$CWD` need to be there? Note in our `foreman-rake` wrapper we `cd` there (see `/usr/sbin/foreman-rake` on an installed system).

At least `rails` scripts uses `__FILE__` location and then constructs Rails.root from it.

> I don't understand why we need these additional wrappers anyway. We have `/usr/share/foreman/bin/rails` and that can be labeled, right?

Yeah this is a left over from my previous version when I planned to have `tfm-ruby-selinux` to be shared between sidekiq and rails. That could not be done. In fact, it is already labelled, so I can just remove that extra step.

> That doesn't make sense to me. We have TCP sockets opened by systemd.

Systemd socket activation uses two sockets. one is TCP 3000 for incoming connections. Then there's notification socket that subprocess uses for IPC and by default it uses UNIX socket which is passed as opened handler to subprocess. But you can also configure systemd to open a named UNIX socket (e.g. `/run/puma.socket`) or even a TCP socket (`localhost:14567`).

Anyway, I am going to remove that extra `tfm-ruby-selinux` which is not necessary actually. I am going to keep sidekiq in libexec as a good precetent.

packages/foreman/foreman/foreman.spec
lzap: Well I have learned a thing today. I tought `<<-EOF` removes intendation from the whole string. It does in fact only from the EOF line, and also these must be tabs not spaces.

packages/foreman/foreman/foreman.spec
lzap: Whoops, thanks for testing. I already have a VM with all manual patching this and I'd need to spin new nightly for this as I don't want to break this one :-)

lzap:
So I am removing the extra `tfm-ruby-script` which is not necessary therefore we keep the `rails` script unchanged and we label it correctly (no change in the policy needed I already did that).

```diff
diff --git a/packages/foreman/foreman/foreman.spec b/packages/foreman/foreman/foreman.spec
index 51fcb4a9e..135054cc3 100644
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -693,15 +693,13 @@ plugins required for Foreman to work.
 
 #replace shebangs and binaries in scripts for SCL
 %if 0%{?scl:1}
-  # shebangs for programs without SELinux transition
-  for f in bin/bundle bin/rake script/performance/profiler script/performance/benchmarker script/dynflowd; do
+  # shebangs
+  for f in bin/* script/performance/profiler script/performance/benchmarker script/dynflowd; do
     sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X%{scl_ruby_bin}X' $f
   done
   # script content
   sed -ri 'sX/usr/bin/rakeX%{scl_rake}X' extras/dbmigrate script/foreman-rake
 %endif
-# shebangs for programs with SELinux transition
-sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X%{_libexecdir}/%{name}/ruby-selinuxX' bin/rails
 # sidekiq service SELinux helper path update
 sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|%{_libexecdir}/%{name}/sidekiq-selinux \1|' extras/systemd/%{dynflow_sidekiq_service_name}.service
 
@@ -764,28 +762,17 @@ install -Dp -m0644 extras/systemd/%{name}.socket %{buildroot}%{_unitdir}/%{name}
 
 # SELinux libexec wrappers
 %if 0%{?scl:1}
-  cat > %{buildroot}%{_libexecdir}/%{name}/ruby-selinux <<-EOF
-  #!/bin/bash
-  # Shell wrapper with SELinux transition into foreman_rails_t domain.
-  source scl_source enable %{scl_prefix}
-  exec ruby "\$@"
-EOF
-  cat > %{buildroot}%{_libexecdir}/%{name}/sidekiq-selinux <<-EOF
-  #!/bin/bash
-  # Shell wrapper with SELinux transition into foreman_rails_t domain.
-  source scl_source enable %{scl_prefix}
-  exec sidekiq "\$@"
+cat > %{buildroot}%{_libexecdir}/%{name}/sidekiq-selinux <<EOF
+#!/bin/bash
+# Shell wrapper with SELinux transition into foreman_rails_t domain.
+source scl_source enable %{scl}
+exec sidekiq "\$@"
 EOF
 %else
-  cat > %{buildroot}%{_libexecdir}/%{name}/ruby-selinux <<-EOF
-  #!/bin/bash
-  # Shell wrapper with SELinux transition into foreman_rails_t domain.
-  exec ruby "\$@"
-EOF
-  cat > %{buildroot}%{_libexecdir}/%{name}/sidekiq-selinux <<-EOF
-  #!/bin/bash
-  # Shell wrapper with SELinux transition into foreman_rails_t domain.
-  exec sidekiq "\$@"
+cat > %{buildroot}%{_libexecdir}/%{name}/sidekiq-selinux <<EOF
+#!/bin/bash
+# Shell wrapper with SELinux transition into foreman_rails_t domain.
+exec sidekiq "\$@"
 EOF
 %endif
```

packages/foreman/foreman/foreman.spec
adamruzicka: TIL :)

ekohl: It does [here](https://github.com/ekohl/ownca/blob/2034411c5819ac62a41bcc27ffe82f1e81ac9573/ownca#L5-L78). Tabs vs spaces perhaps?

lzap:
I have resolved the conflict, anything else @ekohl and @adamruzicka ?

packages/foreman/foreman/foreman.spec
```diff
@@ -763,6 +764,22 @@ install -Dp -m0644 %{SOURCE5} %{buildroot}%{_tmpfilesdir}/%{name}.conf
...
+source scl_source enable %{scl}
+exec sidekiq "\$@"
+EOF
+%else
+cat > %{buildroot}%{_libexecdir}/%{name}/sidekiq-selinux <<EOF
+#!/bin/bash
+# Shell wrapper with SELinux transition into foreman_rails_t domain.
+exec sidekiq "\$@"
+EOF
+%endif
```
ekohl, line 781: Wouldn't this also work:

```suggestion
# SELinux libexec wrappers
cat > %{buildroot}%{_libexecdir}/%{name}/sidekiq-selinux <<EOF
#!/bin/bash
# Shell wrapper with SELinux transition into foreman_rails_t domain.
%if 0%{?scl:1}
source scl_source enable %{scl}
%endif
exec sidekiq "\$@"
EOF
```

packages/foreman/foreman/foreman.spec
```diff
@@ -698,14 +698,14 @@ plugins required for Foreman to work.
 #replace shebangs and binaries in scripts for SCL
 %if 0%{?scl:1}
   # shebangs
-  for f in bin/* script/performance/profiler script/performance/benchmarker script/dynflowd ; do
+  for f in bin/* script/performance/profiler script/performance/benchmarker script/dynflowd; do
```
ekohl, line 701: Please keep the diff small and leave this out

packages/foreman/foreman/foreman.spec
```diff
@@ -698,14 +698,14 @@ plugins required for Foreman to work.
...
     sed -ri '1sX(/usr/bin/ruby|/usr/bin/env ruby)X%{scl_ruby_bin}X' $f
   done
   # script content
   sed -ri 'sX/usr/bin/rakeX%{scl_rake}X' extras/dbmigrate script/foreman-rake
-
-  sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|/usr/bin/scl enable tfm "sidekiq \1"|' extras/systemd/%{dynflow_sidekiq_service_name}.service
 %endif
+# sidekiq service SELinux helper path update
+sed -i '/^ExecStart/ s|/usr/bin/sidekiq \(.\+\)$|%{_libexecdir}/%{name}/sidekiq-selinux \1|' extras/systemd/%{dynflow_sidekiq_service_name}.service
```
ekohl, line 708: Debian also patches this:

https://github.com/theforeman/foreman-packaging/blob/21e0bef50d59a25e8c4b7c48934c3ffb0da770f9/debian/bionic/foreman/rules#L28

I'm wondering if we should make the change in `foreman.git` (https://github.com/theforeman/foreman/blob/9b487e87b29e9a2ec5dfaa91b25ff2c9e73cc566/extras/systemd/dynflow-sidekiq%40.service#L18)

@evgeni thoughts?

packages/foreman/foreman/foreman.spec
lzap: Yeah constant refactoring of a specific code makes you blind.

packages/foreman/foreman/foreman.spec
lzap: Sure.

packages/foreman/foreman/foreman.spec
lzap: Whatever, but keep in mind I am trying to put this into 2.1, this is a critical feature for ReX. If you insist on this, I'd ask you to shorten the review loop so I can get this through.

Don't have a strong opinion on this other than `/usr/bin/sidekiq` looks like a reasonable default :)

lzap:
Corrected the two things, resolved conflicts and rebased.

packages/foreman/foreman/foreman.spec
evgeni: there is no reasonable catch all default that would work for debian and centos, so having a generic path and fix that up seems OK to me.

ekohl:
That's https://github.com/theforeman/foreman-selinux/pull/100

evgeni:
Great, I have no idea about it's contents, but that should not be a concern here ;)

lzap:
Thank you guys for a good review!

PR #5261, opened 2020-05-20
Katello 3.16.0 RC1

evgeni:


package_manifest.yaml
```diff
@@ -801,22 +795,8 @@ katello_packages:
     rubygem-fx: {}
     rubygem-hammer_cli_csv: {}
     rubygem-hammer_cli_foreman_virt_who_configure: {}
-    rubygem-hammer_cli_katello:
-      releasers:
```
evgeni, line 805: IMHO this should have happened during branching…

evgeni:
[test rpm]

PR #5270, opened 2020-05-21
Release tracer-0.7.3-1

jturel:


jturel:
@evgeni so upstream Tracer is at 0.7.2-3 but Jenkins want me to reset the release to 1. Should I listen to him?

evgeni:
EL8 is missing deps:

```
DEBUG util.py:600:  No matching package to install: 'python3-beautifulsoup4'
DEBUG util.py:600:  No matching package to install: 'python3-future'
DEBUG util.py:600:  No matching package to install: 'python3-psutil'
DEBUG util.py:600:  Not all dependencies satisfied
```

evgeni:
@pcreech @ehelms @ekohl the missing packages are in EPEL, and that's where we consume them from for EL7, but for EL8 we've so far have not been adding EPEL and trying to avoid it. Shall we just go and package those ourself (and then also drop the EPEL requirement for EL7 clients?)

ehelms:
If that's not too much work, I lean towards yes. This will simplify things for our users a smidge and keep us aligned with EL8.

jturel:
Very unfortunate because I wanted to get 0.7.3 into this week's downstream build. Is there any appetite for the simpler option: removing the BuildRequires for those missing packages and comment out the unit tests execution from the RPM build? I think that would fix it temporarily.

ekohl:
> If that's not too much work, I lean towards yes. This will simplify things for our users a smidge and keep us aligned with EL8.

I feel like over time we find more and more packages that already exist in EPEL. Maintaining it in our repos can cause conflicts and EPEL use is very common.

> Very unfortunate because I wanted to get 0.7.3 into this week's downstream build. Is there any appetite for the simpler option: removing the BuildRequires for those missing packages and comment out the unit tests execution from the RPM build? I think that would fix it temporarily.

They're still runtime dependencies. I looked into the source, but it has configs in XML so you need beautifulsoup for example. Disabling tests only masks the issue.

jturel:
> > Very unfortunate because I wanted to get 0.7.3 into this week's downstream build. Is there any appetite for the simpler option: removing the BuildRequires for those missing packages and comment out the unit tests execution from the RPM build? I think that would fix it temporarily.
> 
> They're still runtime dependencies. I looked into the source, but it has configs in XML so you need beautifulsoup for example. Disabling tests only masks the issue.

D'oh. Good point

evgeni:
I've packaged bs4 and psutil for el7 and el8 in https://github.com/theforeman/foreman-packaging/pull/5322
I've omitted future as we have it the pulpcore set already, so we can "steal" from there.

evgeni:
[test rpm]

evgeni:
```
nothing provides python3-lxml needed by python3-beautifulsoup4-4.6.3-1.el8.noarch
```

bleh.

evgeni:
@pcreech halp.

pcreech:
*shakes fist at sky*  I'll look into it

pcreech:
[test rpm]

pcreech:
Not sure how it was fixed, but it seems to be fixed now.  I guess it just needed to be threatened by getting outside help to fix it.

jturel:
@pcreech @evgeni thank you guys! this is mergeable :)

evgeni:
[test rpm]

PR #5271, opened 2020-05-22
Release 1.24.3

evgeni:


evgeni:
```
08:26:51 ovirt-engine-sdk-4.4.0 requires ruby version >= 2.5, which is incompatible with
08:26:51 the current version, ruby 2.3.3p222
```

@mmoll I am lost, do we need some pin here?

mmoll:
@evgeni I'm super-puzzled that this didn't reproduce in the tests and also for me locally.

Anyway, I added a pin and another commit for handling nulldb. Please check :gift_heart:.

PR #5272, opened 2020-05-22
Foreman 1.24.3

evgeni:


evgeni:
@tbrisker tis fails to build as it tries to load (not use) nulldb, as you picked https://github.com/theforeman/foreman/commit/073651eef7fc896edcd2d634ebf49b9e14093b86 -- was that intentional?

evgeni:
lets see if that builds now

evgeni:
it does, repoclosure expected

PR #5276, opened 2020-05-22
add activerecord nulldb for 1.24

evgeni:
it's not used, but referenced in 1.24.3


evgeni:
damn, this is not as easy as I thought, as this crosses the rails merge boundary.

@tbrisker is the nulldb change really needed? or can we avoid it?

tbrisker:
It's not required, feel free to drop the `bundler.d/nulldb.rb` file in packaging. I think maybe we pulled it in so upgrade tests on prs with nulldb work, but tbh i don't remember.

PR #5278, opened 2020-05-23
Update rubygem-katello to 3.16.0 RC1.1

evgeni:


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -244,6 +244,9 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/webpack
 
 %changelog
+* Sat May 23 2020 Evgeni Golov - 3.16.0-0.13.rc1.1
+- Release rubygem-katello 3.16.0
```
tbrisker, line 248: not crucial but this doesn't sound like the correct message

evgeni: It isn't. It never was, our tooling generates it incorrectly ;)

evgeni: (I have a todo somewhere to fix, but lazy ;))

PR #5279, opened 2020-05-25
move pin of ovirt-engine-sdk

mmoll:


evgeni:
Do we need to bump the version here? As 1.24.3-1 is already built.

mmoll:
Yes, I just want to wait for DEBs to test locally.

mmoll:
fixes the problem, building now

PR #5286, opened 2020-05-26
Fix tfm-ror52 Obsoletes

evgeni:

### DIFF
--- a/packages/foreman/rubygem-multi_json/rubygem-multi_json.spec
+++ b/packages/foreman/rubygem-multi_json/rubygem-multi_json.spec
@@ -4,8 +4,8 @@
 
 Summary: A gem to provide swappable JSON backends
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 1.12.2
-Release: 6%{?dist}
+Version: 1.14.1
+Release: 1%{?dist}
 Group: Development/Languages
 License: MIT
 URL: https://github.com/intridea/multi_json
@@ -31,7 +31,7 @@ BuildRequires: %{?scl_prefix_ruby}ruby(rubygems)
 BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 
-Obsoletes: tfm-ror52-rubygem-%{gem_name} <= 1.12.2
+Obsoletes: tfm-ror52-rubygem-%{gem_name} <= 1.13.1
 
 # OkJson is allowed to be bundled:
 # https://fedorahosted.org/fpc/ticket/113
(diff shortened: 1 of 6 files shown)


packages/foreman/rubygem-actioncable/rubygem-actioncable.spec
```diff
@@ -28,7 +28,7 @@ BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 # end specfile generated dependencies
 
-Obsoletes: tfm-ror52-rubygem-%{gem_name} <= 5.2.1
+Obsoletes: tfm-ror52-rubygem-%{gem_name} < %{version}-%{release}
```
ehelms, line 31: This essentially will, for all time, Obsolete any tfm-ror52 version-release for this package as this will change everytime the package get's release or version updated. If that's the case, what is the benefit then to including version-release specification?

evgeni: Technically, we could re-create "tfm-ror52-something" with a newer release.
Effectively: make rpmlint shut up, as I don't see us having a rails 5.2 scl ever again.

Either way, the old way was not working as it was not matching the last released ror52 version (as it was lacking the release bit).

ekohl: I think the old worked because it was `<=`. On my system most packages were removed automatically, save for those where I made a PR.

packages/foreman/rubygem-multi_json/rubygem-multi_json.spec
```diff
@@ -5,7 +5,7 @@
 Summary: A gem to provide swappable JSON backends
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 1.12.2
```
ekohl, line 7: the tfm-ror52 SCL contained version 13.x so this still won't obsolete it.

evgeni: Yepp, that's still to be fixed.

packages/foreman/rubygem-actioncable/rubygem-actioncable.spec
evgeni: So I retested this, and it's indeed sufficient. So while I don't fully understand it, I'll back out that change, as it's not really needed.

evgeni:
Okay, with the current set it works for me.

The only question is: why is multi_json built for nonscl-el7? Nothing seem to require that.

evgeni:
> The only question is: why is multi_json built for nonscl-el7? Nothing seem to require that.

I think that's an oversight from the proxy-SCL effort, but I'll clean that up another day.

ekohl:
> I think that's an oversight from the proxy-SCL effort, but I'll clean that up another day.

It could also be in kafo so an oversight of the installer-SCL effort.

PR #5289, opened 2020-05-26
Update rubygem-hammer_cli_foreman_remote_execution to 0.1.2

evgeni:


packages/plugins/rubygem-hammer_cli_foreman_remote_execution/rubygem-hammer_cli_foreman_remote_execution.spec
```diff
@@ -1,34 +1,40 @@
...
-Summary: Foreman Remote Execution commands for Hammer CLI
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.1.1
+Version: 0.1.2
 Release: 1%{?foremandist}%{?dist}
-Group: Applications/System
-License: GPLv3
+Summary: CLI for the Foreman remote execution plugin
+Group: Development/Languages
+License: GPL-3.0-or-later
```
ekohl, line 16: This is the SPDX license tag, not the RPM one.

evgeni: tell `licenses.json` (like I did now ;P)

licenses.json
```diff
@@ -2,5 +2,6 @@
   "GPL-2.0-or-later": "GPLv2+",
   "GPL-3.0": "GPLv3",
   "GPL-3.0+": "GPLv3+",
-  "GPLv3-or-later": "GPLv3+"
+  "GPLv3-or-later": "GPLv3+",
```
ekohl, line 5: Perhaps this should be a separate commit.

evgeni: it can be, if you want to. but still noting that my addition is a valid spdx.

ekohl: Yes. I think it should be a separate commit. I'd be OK with just adding a line but let's keep it sorted.

evgeni: updated

PR #5291, opened 2020-05-26
Update rubygem-hammer_cli_foreman_remote_execution to 0.1.2

evgeni:
(cherry picked from commit 32c880db0f7548e0307411fb3bcecf41c873acc4)


packages/plugins/rubygem-hammer_cli_foreman_remote_execution/rubygem-hammer_cli_foreman_remote_execution.spec
```diff
@@ -1,34 +1,40 @@
...
-Summary: Foreman Remote Execution commands for Hammer CLI
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.1.1
+Version: 0.1.2
 Release: 1%{?foremandist}%{?dist}
-Group: Applications/System
-License: GPLv3
+Summary: CLI for the Foreman remote execution plugin
+Group: Development/Languages
+License: GPL-3.0-or-later
```
ekohl, line 16: This isn't a correct RPM license tag

licenses.json
```diff
@@ -2,5 +2,6 @@
   "GPL-2.0-or-later": "GPLv2+",
   "GPL-3.0": "GPLv3",
   "GPL-3.0+": "GPLv3+",
-  "GPLv3-or-later": "GPLv3+"
+  "GPLv3-or-later": "GPLv3+",
```
ekohl, line 5: AFAIK this is not a valid SPDX license code. Should we keep it out of the cherry pick commit?

evgeni: GPL-3.0-or-later is valid, and that's what I added. GPLv3-or-later is not, but that was here before.

ekohl: Yes, I later realized this was the cherry pick PR.

PR #5296, opened 2020-05-27
Update rubygem-fog-aws to 3.6.5

apuntamb:
Update fog-aws to 3.6.5.


evgeni:
Seems bin/setup and bin/console are new files, could you please exclude them?

apuntamb:
> Seems bin/setup and bin/console are new files, could you please exclude them?

You mean something like: `%exclude %{gem_instdir}/bin/setup` and same for `bin/console`?

But just a query - the new files in a new version should be included, right?

evgeni:
Yes, exactly.

And yeah, normally we would include new files, but console and setup are generic things rubygems adds and is usually without value to us.

evgeni:
See https://github.com/fog/fog-aws/commit/6fb9c3c80793a0532595df5b3eec0614feaa0aaf#diff-c1111bd512b29e821b120b86446026b8 for details, those files are only useful for development.

PR #5299, opened 2020-05-27
fix tfm-ror52 obsoletes for 2.1

evgeni:


evgeni:
@ekohl should we also backport that to 2.0?

ekohl:
@evgeni I think so yes.

PR #5312, opened 2020-05-28
Let tfm-runtime obsolete tfm-ror52-runtime

evgeni:
(cherry picked from commit 24b26be4f0370a02233cdc1fe539cc0204f3f0dd)

slightly odd versioning as we don't want *all* the changes between 6.0-4 and 6.1-1 (esp not the rackspace one), but 6.0-4.1 also felt weird for that change.

### DIFF
--- a/packages/foreman/tfm/tfm.spec
+++ b/packages/foreman/tfm/tfm.spec
@@ -18,8 +18,8 @@
 
 Summary: Package that installs %scl
 Name: %scl_name
-Version: 6.0
-Release: 4%{?dist}
+Version: 6.1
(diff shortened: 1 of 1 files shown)


evgeni:
> I think this is the result of us not bumping the version often enough.

Maybe, but then again, what does the version of this package represent anyways? :)

ekohl:
Well, 4f810d719e9a9e1c391b2d351187bda5c8245b8d is a good example of something that is *not* a release. Same for 6252d2f4594f0f654724b247dbb6f61c8a883580

PR #5313, opened 2020-05-28
Only obsolete the ror52 SCL version in the tfm SCL build

evgeni:
I have untagged the non-scl build from 2.0 and 2.1, but wanted to get it right in nightly at least.


evgeni:
@ekohl well, yes-ish. This still bleeds into the non-scl EL8 builds, so I'm leaning on "fix at some point"

PR #5315, opened 2020-05-28
proxy picks for 2.1

evgeni:


evgeni:
@lzap @adamruzicka would you mind having a brief look that this should go to 2.1 as is?

adamruzicka:
LGTM

evgeni:
"Version updated but release was not reset back to 1 for rubygem-smart_proxy_dns_infoblox" because this touches it twice. Meh.

ekohl:
Sounds like one we should disable for stable branches where we might pick multiple changes (like now).

evgeni:
dropped the infoblox bump, lzap said it's not needed

PR #5316, opened 2020-05-28
Add release version to the repo ids

evgeni:
this removes the need to call `yum clean` on upgrades as the repo id is
used as the cache key buy yum, and if that changes it will download the
metadata again


evgeni:
> The downside is that our Puppet modules can modify these. If they don't reflect the same thing, it may add additional repos. Perhaps the cleanest solution there is to ensure `foreman-release` is absent?

"there" = "in puppet"?

pcreech:
[test rpm]

evgeni:
I'll reopen when I have time to address the puppet part too.

PR #5322, opened 2020-05-29
Tracer deps

evgeni:


evgeni:
@pcreech do you know why it can't find python3-lxml?

pcreech:
The simple answer is, it does not exist in the flattened repo.  Digging into why

pcreech:
[test]

pcreech:
[test rpm]

pcreech:
So, currently, I am not sure.  it exists under the "non_modular" folder, it exists in the repodata, and I can install the package on a centos:8 container by removing all the repos and adding the url to the flat repo.

evgeni:
thankfully it's not really a build-dep, so I just dropped it :disappointed_relieved:

PR #5326, opened 2020-05-29
Update rubygem-hammer_cli to 2.1.1

ofedoren:


mmoll:
the dependency change onto amazing_print needs to get added, too

ofedoren:
Thanks, @mmoll, good point! Updated.

mmoll:
Needs GH-5332 first

PR #5330, opened 2020-05-29
Update rubygem-apipie-dsl to 2.2.5

ofedoren:


tbrisker:
@ofedoren is this the change needed to get examples to work properly? if so we'll want to also cherry-pick to rpm/2.1

ofedoren:
@tbrisker, unfortunately it's not. This change has some fixes regarding other stuff (e.g. broken language links). Probably this also needs to be CPed...

PR #5337, opened 2020-06-01
Update foreman_remote_execution to 3.3.0

adamruzicka:


mmoll:
@adamruzicka With 2.x versions is this also for?

adamruzicka:
This needs to get into at least 2.1, should be safe to go to 2.0

mmoll:
merged, díky @adamruzicka!

PR #5350, opened 2020-06-03
regenerate hammer, hammer-foreman, hammer-katello

evgeni:

### DIFF
--- a/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
+++ b/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
@@ -1,64 +1,60 @@
+# template: hammer_plugin
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
 %global gem_name hammer_cli
-%global confdir hammer
 
-%global release 2
+%global release 3
 %global prereleasesource pre.develop
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 
 %{!?_root_bindir:%global _root_bindir %{_bindir}}
 %{!?_root_mandir:%global _root_mandir %{_mandir}}
 %{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
+%global hammer_confdir %{_root_sysconfdir}/hammer
 
-Summary: Universal command-line interface for Foreman
 Name: %{?scl_prefix}rubygem-%{gem_name}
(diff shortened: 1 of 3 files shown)


packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
```diff
@@ -1,64 +1,60 @@
...
 %{!?_root_bindir:%global _root_bindir %{_bindir}}
 %{!?_root_mandir:%global _root_mandir %{_mandir}}
 %{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
+%global hammer_confdir %{_root_sysconfdir}/hammer
 
-Summary: Universal command-line interface for Foreman
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 2.2.0
 Release: %{?prerelease:0.}%{release}%{?prerelease}%{?nightly}%{?dist}
+Summary: Universal command-line interface
```
ekohl, line 19: the `for Foreman` part was nice

packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
```diff
@@ -102,33 +97,36 @@ mkdir -p %{buildroot}%{_root_mandir}/man1
...
+%license %{gem_instdir}/LICENSE
 %{gem_instdir}/bin
-%{gem_instdir}/lib
+%{gem_libdir}
 %{gem_instdir}/locale
-%license %{gem_instdir}/LICENSE
 %exclude %{gem_cache}
 %{gem_spec}
+%{hammer_confdir}/cli.modules.d
+%config %{hammer_confdir}/cli_config.yml
```
ekohl, line 117: Looks like this is true for all, but does the `(noreplace)` part make sense to keep and include in the template? Are we consistent with all hammer (plugins) now?

evgeni: When this is merged, we're consistent with all plugins not having "noreplace"

packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
evgeni: *technically* `hammer_cli` is not for Foreman, `hammer_cli_foreman` is :)

I don't mind changing it back, but we should also maybe update the gem descriptions then?

PR #5357, opened 2020-06-04
Update foreman-tasks to 2.0.0

adamruzicka:
Dependency of foreman was bumped earlier, as part of introducing nulldb.


mmoll:
merged, díky @adamruzicka!

PR #5359, opened 2020-06-04
pulpcore 3.4.1 and plugins

evgeni:


evgeni:
explicitly setting review on Justin, as this needs changes in Katello/pulp gems to come along.

packages/pulpcore/python-pulpcore/python-pulpcore.spec
```diff
@@ -20,8 +20,6 @@ BuildRequires:  python3-aiodns
 BuildRequires:  python3-aiofiles
 BuildRequires:  python3-aiohttp
 BuildRequires:  python3-backoff
-Requires:       python3-coreapi >= 2.3.3
```
ekohl, line 23: Will you create a separate PR to drop coreapi?

evgeni: no, because it's still pulled in via python-drf-yasg :(

PR #5362, opened 2020-06-04
Make libmodulemd dependency versioned

evgeni:
libmodulemd1 provide libmodulemd = 1.8.0, which satisfies the dependency
otherwise


evgeni:
@ehelms @wbclark let's see if that'll be the last patch for https://github.com/theforeman/puppet-pulpcore/pull/89

PR #5363, opened 2020-06-04
Adding python3-subscription-manager-rhsm for pulp builds

jlsherrill:
The goal of this is to build python3-subscription-manager-rhsm
in order to use certguard with rhsm cert protection.  I think this is okay given:
* this is only needed until we move to el8
* it wont need to stay in lock-step with the subscription-manager version installed
* its unlikely to need updates


package_manifest.yaml
```diff
@@ -1351,6 +1351,7 @@ pulpcore_packages:
     python-xlwt: {}
     python-yarl: {}
     python-zipp: {}
+    subscription-manager: {}
```
ehelms, line 1354: Add

```
    koji_tags:
      - name: katello-pulpcore-nightly-el7
        dist: '.el7'
```

ehelms:
You are gonna have to introduce a blacklist for https://github.com/theforeman/foreman-packaging/pull/5363/files#diff-150be58e1a8579150d5a3803c4f60d6cR1246 to stop the EL8 build as well.

jlsherrill:
This is now green!  @evgeni i did have to hack up the spec files quite a bit, mostly to get repoclsoure to pass since we only care about one of the subpackages of rhsm.  Any objections?  If this was a longer term solution i'd not go this route, but this buys us time to get to el8

jlsherrill:
and i think that the python3-subscrption-manager-rhsm package should be installed via the installer, and not via a dependency of the certguard plugin

evgeni:
Fwiw, py-iniparse in el8 does not build py2, so it might be cleaner to use that spec as a base? https://koji.mbox.centos.org/koji/buildinfo?buildID=430

That said, I don't care too much as long as these hacks are self contained and vanish at some point where we stop supporting el7 installs of pulpcore (right @ehelms, that will be soon?!).

packages/pulpcore/subscription-manager/subscription-manager.spec
```diff
@@ -0,0 +1,5935 @@
...
+Group: Development/Libraries
+%endif
+
+%if %use_m2crypto
+Requires: %{?suse_version:python-m2crypto} %{!?suse_version:m2crypto}
+%endif
+%if 0%{?suse_version} >= 1500
+Requires:  %{py_package_prefix}-python-dateutil
+%else
+Requires: python36-dateutil
```
evgeni, line 558: That's coming from EPEL, is that correct?

packages/pulpcore/subscription-manager/subscription-manager.spec
```diff
@@ -0,0 +1,5935 @@
...
+%endif
+Requires: python36-iniparse
```
evgeni, line 560: Technically, this should be `python3-iniparse`, but your package also provides the 36 name.

packages/pulpcore/subscription-manager/subscription-manager.spec
```diff
@@ -0,0 +1,5935 @@
...
+Requires: %{py_package_prefix}-six
+Requires: subscription-manager-rhsm-certificates
+# Required by Fedora packaging guidelines
+%{?python_provide:%python_provide %{py_package_prefix}-rhsm}
+%if %{with python3}
+Requires: python3-rpm
+Provides: python3-rhsm = %{version}-%{release}
+Obsoletes: python3-rhsm <= 1.20.3-1
+Provides: python-rhsm = %{version}-%{release}
+Obsoletes: python-rhsm <= 1.20.3-1
```
evgeni, line 571: I don't think we should provide/obsolete these, given that on EL7 that are py2 versions.

packages/pulpcore/subscription-manager/subscription-manager.spec
jlsherrill: correct!

packages/pulpcore/subscription-manager/subscription-manager.spec
jlsherrill: fixed!

packages/pulpcore/subscription-manager/subscription-manager.spec
jlsherrill: fixed

jlsherrill:
i think i've addressed all the comments/questions

PR #5368, opened 2020-06-09
update katello gem deps

evgeni:

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -9,7 +9,7 @@
 %global prereleasesource pre.master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global mainver 4.0.0
-%global release 3
+%global release 4
 
 Name:    %{?scl_prefix}rubygem-%{gem_name}
 Summary: Content and Subscription Management plugin for Foreman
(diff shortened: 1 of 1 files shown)


snecklifter:
```Gem loading error: Could not find 'pulp_2to3_migration_client' (< 0.3.0, > 0.02b2) - did find: [pulp_2to3_migration_client-0.2.0b2]

**Comment by evgeni** [comment]
[test rpm]

File: packages/katello/rubygem-katello/rubygem-katello.spec [hunk 1]
```diff
@@ -22,9 +22,10 @@ URL:     https://theforeman.org/plugins/katello
 Source0: https://rubygems.org/downloads/%{gem_name}-%{version}%{?prerelease}.gem
 
 Requires: foreman-postgresql
+Requires: foreman < %{foreman_max_version}
+BuildRequires: %{?scl_prefix}rubygem(pg)
```
evgeni, line 26: @jlsherrill seems this was introduced by https://github.com/Katello/katello/commit/514770541975012ea85a5780fd67c237cb535e2d but never added a dependency to the gemspec

jlsherrill: ahh, did not realize this was a build time dep now, but it makes sense

evgeni: https://github.com/Katello/katello/pull/8757

evgeni:
repoclosure because I didn't mash :see_no_evil:

PR #5379, opened 2020-06-10
Release 2.0.1

ehelms:


evgeni:
Heh. nulldb here too. Thanks @mmoll 

I will never remember that "just because debian builds, doesn't mean the result is usable" :)

ehelms:
What do I need to do here to resolve?

evgeni:
> What do I need to do here to resolve?

lean back and drink tea, Mister Moll did all the necessary.

PR #5388, opened 2020-06-10
Ensure foreman socket is restarted on upgrade

ehelms:
See https://community.theforeman.org/t/foreman-nightly-rpm-pipeline-608-failed/19154/2?u=ehelms


ehelms:
When I tested the scratch RPM, I got successful tests. Testing with the repositories and this fix, it still fails.

PR #5393, opened 2020-06-11
Add update_vendor.sh

sharvit:
Add a small script to update tfm-vendor and tfm-builder together


sharvit:
https://github.com/theforeman/foreman-packaging/pull/5394 created using the script

PR #5404, opened 2020-06-15
Update sprockets to 4.0.2

mmoll:


evgeni:
The diff is fine and the repoclosure error unrelated, but needs to go in together with the core change (and a corresponding update of `foreman.spec` after that).

mmoll:
@evgeni I'd merge the core PR and this RPM PR after 2.1 RC3 is out, ACK?

evgeni:
Sounds good to me

PR #5405, opened 2020-06-16
Katello 3.16 RC3

evgeni:


packages/katello/rubygem-pulp_certguard_client/ruybgem-pulp_cerguard_client.spec
```diff
@@ -0,0 +1,91 @@
+# template: scl
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+
+%global gem_name pulp_certguard_client
+%global gem_require_name %{gem_name}
+
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 0.1.0rc5
```
evgeni, line 9: Yes, the versioning is off, but that's also what we have in `nightly`, and I just picked/tagged from there.

evgeni:
[test rpm]

PR #5407, opened 2020-06-16
bump supported openscap version

evgeni:
CentOS 8.2 contains openscap 1.3.2 which still has soname 25
1.3.3 (latest upstream) also has 25
so lets break on 1.3.4 now

(cherry picked from commit c7554323c7814cb80cac1c6e74954928d79226ec)


evgeni:
merging as this is CP only change

PR #5371, opened 2020-06-09
Bump version of smart proxy for salt

sbernhard:

PR #5413, opened 2020-06-16
Bump version of smart proxy for salt (foreman-1.24)

m-bucher:
Co-Authored-By: @adamruzicka
(cherry picked from commit 480cae4300a0400c6f775d87260f634402edd767)
Cherry-pick of #5371

@adamruzicka does the cherry-pick of your changes from #5229 make sense here (for foreman-1.24) and if so did I do it correctly?


adamruzicka:
Well, since that macro is not being used anywhere in the spec, then it isn't really necessary, but it brings the spec closer to the original template.

m-bucher:
Not sure why the repoclosure ever worked, but I guess to fix this we have to switch `rubygem-smart_proxy_salt` to `tfm-rubygem-smart_proxy_salt`!?

m-bucher:
Included changes from #5416

m-bucher:
So packages/plugins/rubygem-smart_proxy_salt_core/rubygem-smart_proxy_salt_core.spec was wrong all along?
So I guess I will have to de-SCL both spec-files then :laughing:

evgeni:
No, the spec was OK, the SCL things in the spec are conditional, you just don't need the changes to tito.props and comps.

m-bucher:
OK, but is `tfm-rubygem-smart_proxy_dynflow_core` then just as wrong in `comps/comps-foreman-plugins-rhel7.xml`(just being curious here :wink: ).

evgeni:
_core is in SCL, the rest isn't. 1.24 and before was wild ;)

m-bucher:
OK, back to the beginning:
```
package: tfm-rubygem-smart_proxy_salt_core-0.0.3-1.el7.noarch from downloaded_rpms
  unresolved deps:
    rh-ruby25-ruby(release)
    rh-ruby25-ruby(rubygems)
    rh-ruby25-ruby
```
This means in order to check the repoclosure for salt_core, it needs the `rh-ruby` SCL repositories.
How do I fix this?

evgeni:
You don't. It's a bug with the repoclosure test when you modify scl and non scl packages in the same pr, that's fine

m-bucher:
OK, so looks like I still have to fix something, because the replacement of `%{__python2}` does not work, yet.

m-bucher:
Hashbangs are fixed here and I opened a PR to fix it in develop as well (#5466).
@evgeni it's done from my side.

sbernhard:
LGTM. Would be good to merge this soon :)

PR #5416, opened 2020-06-16
Put local state to /var/lib instead of /usr/com for smart proxy salt

adamruzicka:
s/_sharedstatedir/_localstatedir/g


adamruzicka:
smart proxy salt got separated from the other changes around statedir because of (at the time) missing salt-master dependency. Because of this I forgot to update it when the majority was updated, this PR addresses that

PR #5418, opened 2020-06-17
regenerate GEM file dependencies

evgeni:


evgeni:
[test rpm]

evgeni:
repoclosure because I didn't mash

PR #5426, opened 2020-06-17
Add the candlepin artemis broker.xml to foreman-debug

barnabycourt:
Add the broker.xml that is used for configuring the Artemis instance that is embedded inside candlepin to the list of files collected during foreman-debug.

### DIFF
--- a/packages/katello/katello/katello-debug.sh
+++ b/packages/katello/katello/katello-debug.sh
@@ -52,7 +52,11 @@ add_files /var/log/tomcat*/host-manager*.log*
 add_files /var/log/tomcat*/localhost*.log*
 add_files /var/log/tomcat*/manager*.log*
 add_files /etc/candlepin/candlepin.conf
+add_files /etc/candlepin/broker.xml
 add_files /etc/tomcat*/server.xml
+add_files /etc/tomcat/cert-roles.properties
+add_files /etc/tomcat/cert-users.properties
+add_files /etc/tomcat/login.config
 
 # RHSM
 add_files /var/log/rhsm/*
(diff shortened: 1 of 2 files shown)


evgeni:
ok to test

evgeni:
@barnabycourt can you please add a changelog entry and increase the rpm release number?

packages/katello/katello/katello-debug.sh
```diff
@@ -52,6 +52,7 @@ add_files /var/log/tomcat*/host-manager*.log*
 add_files /var/log/tomcat*/localhost*.log*
 add_files /var/log/tomcat*/manager*.log*
 add_files /etc/candlepin/candlepin.conf
+add_files /etc/candlepin/broker.xml
```
ehelms, line 55: Do you also want the other files associated with auth?

https://github.com/theforeman/puppet-candlepin/blob/master/manifests/artemis.pp#L15-L39

@jturel

jturel: Couldn't hurt

```
/etc/tomcat/cert-roles.properties
/etc/tomcat/cert-users.properties
/etc/tomcat/login.config
```

barnabycourt:
@evgeni Are you referring to hand editing the spec template in packages/katello/katello/katello.spec? If so, is that file usually edited by hand in this project or is there a tool that you recommend using?

packages/katello/katello/katello-debug.sh
barnabycourt: @ehelms Done

evgeni:
@barnabycourt there is https://GitHub.com/theforeman/obal that can do this for you, but doing it by hand is also fine.

barnabycourt:
@evgeni based on comments in IRC, I squashed down to a single commit. I don't know what is causing the repoclosure issue in the CI tests but it appears to be unrelated to these changes.

evgeni:
Yeah, repoclosure error is unrelated

barnabycourt:
@jturel @ehelms Anything else that you would like to see from this?

jturel:
[test rpm]

PR #5446, opened 2020-06-19
Switch to pyOpenSSL naming for pyopenssl to match official packages

ehelms:
I am guessing the py2rpm template did this automatically and that we should follow the convention for this package to ensure it does not clash:

https://community.theforeman.org/t/katello-el8-failure/19329

Gonna give the scratch RPM a test.


ehelms:
Looks like not -- https://koji.fedoraproject.org/koji/buildinfo?buildID=1523906

PR #5449, opened 2020-06-21
Fixes #30176 - Set foreman-proxy file owner + mode

ekohl:
Currently the installer sets settings.yml and settings.d/*.yml to foreman_proxy:foreman_proxy with mode 0640 since some files contain secrets. Packaging sets them to root:root with 0644.

This causes inconsistency creates problems. rpm -qV now has mismatches on these properties. It also causes the installer to change them and restart the foreman-proxy service, even if nothing changed.

### DIFF
--- a/packages/foreman/foreman-proxy/foreman-proxy.spec
+++ b/packages/foreman/foreman-proxy/foreman-proxy.spec
@@ -6,7 +6,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 2
+%global release 3
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


evgeni:
Why should the proxy be able to write it's own configuration? IMHO this should be `440` or `root:foreman_proxy` with `640`.

ekohl:
I had indeed intended to match with the installer which sets it to `root:foreman_proxy` with `640`:
https://github.com/theforeman/puppet-foreman_proxy/blob/21f50bebae877eac73d9d405962a7e818630faad/manifests/settings_file.pp#L28-L30

packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -143,7 +143,10 @@ ln -sv %{_tmppath} %{buildroot}%{_datadir}/%{name}/tmp
 %license LICENSE
 %{_datadir}/%{name}
 %{_libexecdir}/%{name}/puppet_sign.rb
-%config(noreplace) %{_sysconfdir}/%{name}
+%dir %{_sysconfdir}/%{name}
```
ekohl, line 146: This is my attempt at avoid the warning that it listed the file twice. Not sure if this is needed though if we list files in that directory.

ekohl:
I verified the files are correctly owned by installing on a container and don't see any more warnings in the build log.

PR #5394, opened 2020-06-11
Update tfm-vendor and tfm-builder to v4.7.0

sharvit:

PR #5450, opened 2020-06-21
Bump foreman-js to 4.8.0

tbrisker:
Replaces GH-5394


tbrisker:
Looking deeper into the spec file, it looks like we are pulling in 2 versions of the pinned modules (e.g. both `babel-plugin-dynamic-import-node@2.3.0` and `babel-plugin-dynamic-import-node@2.3.3`). Hopefully we will still be able to build properly.
I would suggest to keep the pinning in core as well for now just to make sure that we keep 2.3.0 at the top level, and work towards fixing the code in a way that will allow us to remove the problematic part that requires this pinning.

packages/foreman/nodejs-theforeman-builder/nodejs-theforeman-builder.spec
```diff
@@ -4,287 +4,295 @@
...
+Source93: https://registry.npmjs.org/@babel/preset-modules/-/preset-modules-0.1.3.tgz
+Source94: https://registry.npmjs.org/@babel/preset-react/-/preset-react-7.10.1.tgz
+Source95: https://registry.npmjs.org/@babel/runtime/-/runtime-7.10.3.tgz
+Source96: https://registry.npmjs.org/@babel/template/-/template-7.10.3.tgz
+Source97: https://registry.npmjs.org/@babel/traverse/-/traverse-7.10.3.tgz
+Source98: https://registry.npmjs.org/@babel/types/-/types-7.10.3.tgz
+Source99: https://registry.npmjs.org/@theforeman/builder/-/builder-4.8.0.tgz
+Source100: https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz
+Source101: https://registry.npmjs.org/babel-plugin-dynamic-import-node/-/babel-plugin-dynamic-import-node-2.3.0.tgz
+Source102: https://registry.npmjs.org/babel-plugin-dynamic-import-node/-/babel-plugin-dynamic-import-node-2.3.3.tgz
```
tbrisker, line 115: Notice how we have both versions here

tbrisker:
[test rpm]

tbrisker:
[test rpm]

tbrisker:
looks like i messed up something in the registry cache while digging in this, closing

PR #5453, opened 2020-06-22
Update npm requirements for vendor v4.8.0

sharvit:


packages/foreman/foreman/foreman.spec
```diff
@@ -1016,6 +1016,9 @@ exit 0
 %systemd_postun_with_restart %{name}.service
 
 %changelog
+* Mon Jun 22 2020 Avi Sharvit <sharvita@gmail.com> - 2.2.0-0.27.develop
+-  - update foreman-js dependencies
```
tbrisker, line 1020: nitpick - looks like there is an extra ` - ` here
```suggestion
- update foreman-js dependencies
```

PR #5454, opened 2020-06-22
Fix dynflow_core bundler.d location on EL8

evgeni:


evgeni:
I've only edited the packages that actually place a file into that folder.

Ping if you want me to update all of them.

ekohl:
I'd prefer to update them all. If they start to place a file there, you'll end up with an annoying bug that's hard to trace.

evgeni:
@ekohl there you go. `sed` + `obal` = ♥

ehelms:
Did scl_utils change? How does this work on EL7 and EL8 now?

evgeni:
@ehelms well, on EL7 we're *inside* the SCL now, so no need for "if not in SCL prepend SCL path for dynflow core", and on EL8 it was always wrong to prepend it :)

ekohl:
This is essentially step 2 of what @ehelms said in https://github.com/theforeman/foreman-packaging/pull/4147#discussion_r328714325. Sadly it took us more than a week to follow through.

PR #5459, opened 2020-06-22
Require python-setuptools on EL7 for k-certs-tools

ekohl:
Since 2.7.1 setuptools is used to create /usr/bin/katello-ssl-tool by generating a wrapper. This needs pkg_resources which on EL8 is included in Python, but isn't on EL7. This is provided by python-setuptools.


ekohl:
Repoclosure is unrelated

PR #5473, opened 2020-06-23
Update rubygem-foreman-tasks to 2.0.1

adamruzicka:


packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -162,6 +162,9 @@ type foreman-selinux-relabel >/dev/null 2>&1 && foreman-selinux-relabel 2>&1 >/d
 %doc %{gem_instdir}/extra/dynflow-executor.example
 
 %changelog
+* Thu Jun 23 2020 Adam Ruzicka <aruzicka@redhat.com> 2.0.1-1
```
tbrisker, line 165: Linter doesn't like the fact that June 23rd was Tuesday and not Thursday...

PR #5476, opened 2020-06-24
Fix rubygem-radcli package building

ekohl:
172820d849aed463fd7d443dbcee1511da8a98a5 changed the spec for SCL but didn't take care of the native library.

Cherry picked from commits 77b16a36a33b4b8ff9dea6804abb34d8c072daff and be5b502953e21392e981d8d5b433044b4f392021.


evgeni:
Now I will have nightmares again. Thanks!

PR #5494, opened 2020-06-30
make acd build on el7 and el8

evgeni:


evgeni:
@ekohl because I fixed the failure and then realized it only built for el8, so it won't retry

evgeni:
we can ofcourse omit the bump and run a manual `obal release` for that one.

pcreech:
It's probably sufficient to manually build it with obal and a specified releaser instead of version bumping.  Since it was my mistake that led to this, I don't mind being the one to obal release.

pcreech:
eh, llooks like this brings in comps updates as well.  Just going ahead and merging

PR #5499, opened 2020-07-01
also use our python2-libcomps

evgeni:
otherwise the version from extras is pulled in and that needs a
different libcomps.so


evgeni:
cc @jlsherrill @pcreech I think this will need to go to 2.1 too, after justins pulpcore picks

PR #5483, opened 2020-06-25
Fixes #30040 - Update foreman-proxy dependencies

ekohl:
This uses the same scripted update mechanism as foreman, katello and other gem packages.

PR #5500, opened 2020-07-01
Fixes #30040 - Update foreman-proxy dependencies

ekohl:
This uses the same scripted update mechanism as foreman, katello and other gem packages.

(cherry picked from commit accae9fd118249fc831d9c8c04e70f4a1a943314)

### DIFF
--- a/packages/foreman/foreman-proxy/foreman-proxy.spec
+++ b/packages/foreman/foreman-proxy/foreman-proxy.spec
@@ -6,7 +6,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 2
+%global release 3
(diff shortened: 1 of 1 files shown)


ehelms:
It was.

ekohl:
Yes, see the cherry pick commit. https://github.com/theforeman/foreman-packaging/pull/5483 was the PR itself.

ekohl:
The bug was reported against 2.0 and I think the cherry pick is low risk. Should I also pick there?

tbrisker:
@ekohl sounds good to me, though i'm guessing it won't propagate until we do another full 2.0 release cycle?

ekohl:
Correct, unless we deem it important enough to do an async release. I think it's not needed though.

tbrisker:
@ekohl makes sense, let's fix it so it gets in to 2.0.2 whenever it is released.

ekohl:
https://github.com/theforeman/foreman-packaging/pull/5503

PR #5502, opened 2020-07-01
Update rubygem-foreman-tasks to 1.1.3

adamruzicka:


tbrisker:
The migration is failing since 2.0 still has rails 5.2:
```
ArgumentError: Unknown migration version "6.0"; expected one of "4.2", "5.0", "5.1", "5.2"
```
you'll want to change the migration version to 5.2 in tasks and re-release :(

PR #5507, opened 2020-07-02
Fixes #30313 - Use case-insensitive sed for hostname change

synkd:

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -411,7 +411,7 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       end
 
       STDOUT.puts "updating hostname in /etc/hostname"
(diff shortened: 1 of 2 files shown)


ehelms:
@synkd Do you mind bumping the release on the katello package? You can do:

```
pip install obal
obal bump-release katello --changelog "Use case-insensitive sed for hostname change"
```

ekohl:
I also maintain RPMs @ https://copr.fedorainfracloud.org/coprs/ekohl/obal/ if you prefer to avoid `pip install`. It has the added benefit that it already pulls in all dependencies rather than requiring you to use `obal setup` and gives you bash completion out of the box.

pcreech:
[test rpm]

wbclark (approved):
LGTM. Thanks @synkd for the collab on this one!

evgeni (approved):

PR #5508, opened 2020-07-02
do not change hashbang on centos < 8

bochi:
This will fix the hashbang for CentOS7.
On CentOS7, /usr/bin/env python exists which will make the script work for python2 and python3. 
Please note that with Salt 3001 python2 support was deprecated, so a script with /usr/bin/python2 is pretty much useless.


evgeni:
ok to test

@sbernhard @m-bucher

m-bucher:
Looks OK to me.
@bochi can you increase the release of the spec-file, so the rpms are built by koji and we can look into them.
```
Version or release needs updating for rubygem-smart_proxy_salt"
```

bochi:
> Looks OK to me.
> @bochi can you increase the release of the spec-file, so the rpms are built by koji and we can look into them.
> 
> ```
> Version or release needs updating for rubygem-smart_proxy_salt"
> ```

Done

sbernhard:
LGTM, thanks @bochi

PR #5509, opened 2020-07-02
Update foreman-release to 2.1.0

pcreech:


pcreech:
This is failing to include teh scl in repoclosure.  We occasionally have this happen with rpms that straddle non-scl and scl.

@ehelms thoughts on merging with repoclosure failure?

PR #5512, opened 2020-07-07
Add sub-man rhsm to comps for pulpcore

jlsherrill:


ekohl:
Should we perhaps pick ca41b39217d0b38e6cd5e2e41b9eee0a7a9a669a? That also pulls in iniparse and I'm guessing that's a dependency. Since you made the PR, you might remember.

jlsherrill:
good catch, cherry picked

rel-eng/tito.props
```diff
@@ -1243,4 +1243,3 @@ disttag = .el8
 blacklist =
   python-iniparse
   subscription-manager
-
```
ekohl, line 1246: Is this needed?

jlsherrill: cherry pick conflict :)

PR #5513, opened 2020-07-07
Ensure foreman.socket is removed on package removal

ehelms:
This reverts commit 5c83debb7411b8f42bd53d197c4febb47533cea5.

Cleanup from broken upgrades that were the result of needing to backport SELinux changes. This change is not present in 2.1 from my investigation so just needs this revert in nightly branch.

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -9,7 +9,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 28
+%global release 29
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -1005,18 +1005,19 @@ fi
 exit 0
 
 %post service
-%systemd_post %{name}.socket
-%systemd_post %{name}.service
+%systemd_post %{name}.socket %{name}.service
 
 %preun service
-%systemd_preun %{name}.socket
-%systemd_preun %{name}.service
+%systemd_preun %{name}.socket %{name}.service
 
 %postun service
-%systemd_postun_with_restart %{name}.socket
 %systemd_postun_with_restart %{name}.service
+%systemd_postun %{name}.socket
 
 %changelog
+* Wed Jul 29 2020 Eric D. Helms <ericdhelms@gmail.com> - 2.2.0-0.29.develop
+- Ensure foreman.socket is removed on package removal
+
 * Mon Jul 20 2020 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> - 2.2.0-0.28.develop
 - Update Gem and NPM dependencies
 


ekohl:
Does that perhaps mean we actually need to cherry pick it and see if it fixes the occasional 502?

ehelms:
I don't think we do. This was originally due to the fact that SELinux issues were causing the service to enter a failed state due to file descriptor on upgrade. One way to fix that was restarting the socket but we shouldn't *need* to restart the socket.

ehelms:
Updated to reflect just the uninstall cleanup

packages/foreman/foreman/foreman.spec
```diff
@@ -1005,18 +1005,20 @@ fi
 exit 0
 
 %post service
-%systemd_post %{name}.socket
```
ekohl, line 1008: This is incorrect. It needs to be there and sets up the initial installation. See https://github.com/systemd/systemd/blob/6b222c4b0227f9914446ac54754aea867f742be4/src/core/macros.systemd.in#L46-L52

I do think `%{?*}` in there means you can combine them to be done in parallel.

packages/foreman/foreman/foreman.spec
```diff
@@ -1005,18 +1005,20 @@ fi
...
 %systemd_post %{name}.service
 
 %preun service
-%systemd_preun %{name}.socket
 %systemd_preun %{name}.service
+%systemd_preun %{name}.socket
```
ekohl, line 1012: Like with `%post` I think you can combine them which means systemd can properly figure out the dependencies and ensure they're both stopped.
```suggestion
%systemd_preun %{name}.service %{name}.socket
```

packages/foreman/foreman/foreman.spec
```diff
@@ -1032,9 +1034,6 @@ exit 0
 * Mon Jun 15 2020 Eric D. Helms <ericdhelms@gmail.com> - 2.2.0-0.24.develop
 - Update Rails to 6.0.3.1
 
-* Wed Jun 10 2020 Eric D. Helms <ericdhelms@gmail.com> - 2.2.0-0.23.develop
-- Ensure socket gets handled with service
```
ekohl, line 1036: This shouldn't be removed

packages/foreman/foreman/foreman.spec
```diff
@@ -1005,18 +1005,20 @@ fi
...
 
 %preun service
-%systemd_preun %{name}.socket
 %systemd_preun %{name}.service
+%systemd_preun %{name}.socket
 
 %postun service
-%systemd_postun_with_restart %{name}.socket
 %systemd_postun_with_restart %{name}.service
+%systemd_postun_with_restart %{name}.socket
```
ekohl, line 1016: Did you intend this? I think you meant
```suggestion
%systemd_postun %{name}.socket
```

packages/foreman/foreman/foreman.spec
```diff
@@ -1005,18 +1005,20 @@ fi
...
 
 %changelog
+* Wed Jul 29 2020 Eric D. Helms <ericdhelms@gmail.com> - 2.2.0-0.29.develop
+- Drop restart of socket
```
ekohl, line 1020:
```suggestion
- Drop restart of socket on upgrade in foreman-service
```

ekohl:
Code looks good, though the commit summary + the changelog entry + PR title all differ.

PR #5518, opened 2020-07-08
Switch to pyOpenSSL naming for pyopenssl to match official packages

evgeni:


evgeni:
@jlsherrill another set you forgot :stuck_out_tongue:

evgeni:
also fixing the repoclosure by fixing comps comps, not that it will become green in this PR

PR #5524, opened 2020-07-09
Fix path of smart_proxy_salt (s)/bin scripts

sbernhard:

### DIFF
--- a/packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
+++ b/packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
@@ -15,6 +15,7 @@
 %{!?scl:%global pkg_name %{name}}
 
 %{!?_root_bindir:%global _root_bindir %{_bindir}}
+%{!?_root_sbindir:%global _root_sbindir %{_sbindir}}
 %{!?_root_datadir:%global _root_datadir %{_datadir}}
 %{!?_root_localstatedir:%global _root_localstatedir %{_localstatedir}}
 %{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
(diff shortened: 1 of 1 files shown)


sbernhard:
@ekohl what do you think about this? As said, other smart proxys are affected, too. Can you confirm this (e.g. for omaha, openscap)?

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -176,16 +178,18 @@ EOF
...
 %exclude %{gem_instdir}/cron
 
-
 %files doc
 %doc %{gem_docdir}
 %doc %{gem_instdir}/README.md
 
 %changelog
+* Thu Jul 09 2020 Bernhard Suttner <suttner@atix.de> - 3.1.2-6
+- Fix upload-salt-report pat
```
ekohl, line 190:
```suggestion
- Fix upload-salt-report path
```

sbernhard:
Have a look at /usr/bin/smart_proxy_openscap_send (https://yum.theforeman.org/plugins/2.1/el7/x86_64/tfm-rubygem-smart_proxy_openscap-0.7.2-6.fm2_1.el7.noarch.rpm) which uses "#!/usr/bin/env ruby". I guess this can not work because of wrong ruby environment.

ekohl:
You're right, that can't work. I think the delivery team should take a look.

sbernhard:
> I wonder why this should be in sbin.

upload-salt-reports loads /etc/salt/master which is only readable for root. Is this a reason to put the upload script in /sbin?

ekohl:
Hmm, fair point. I sort assumed we had fixed shebangs in the template. Looks like we don't:
https://github.com/theforeman/foreman-packaging/blob/8243ef3d8510d743321cb8321ad8a6c282dfd8ed/gem2rpm/smart_proxy_plugin.spec.erb#L124-L130

sbernhard:
Anything I need to change before this can be merged?

evgeni:
Nope, just time for someone to review this :)

evgeni:
@sbernhard will you take care of opening any cherry picks against stable branches that might be needed?

PR #5527, opened 2020-07-13
Fixes #30367 - Add Infoblox dependency to sp_dns_infoblox

evgeni:
f825335074aacaf60b97ff99aa72e8b2bc0c8279 dropped this dependency, but it
should really be there.

(cherry picked from commit 6cc7188f246577dee2a58d99aa83d76a63709b0f)

### DIFF
--- a/packages/plugins/rubygem-smart_proxy_dns_infoblox/rubygem-smart_proxy_dns_infoblox.spec
+++ b/packages/plugins/rubygem-smart_proxy_dns_infoblox/rubygem-smart_proxy_dns_infoblox.spec
@@ -18,7 +18,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 1.0.0
-Release: 3%{?foremandist}%{?dist}
+Release: 3.1%{?foremandist}%{?dist}
 Summary: Infoblox DNS provider plugin for Foreman's smart proxy
 Group: Applications/Internet
 License: GPLv3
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-smart_proxy_dns_infoblox/rubygem-smart_proxy_dns_infoblox.spec
```diff
@@ -18,7 +18,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 1.0.0
-Release: 3%{?foremandist}%{?dist}
+Release: 3.1%{?foremandist}%{?dist}
```
evgeni, line 21: small bump is intentional, as 4 exists already in koji

ehelms: The foremandist should be defined differently and thus end up with a different package that rpm/develop right?

evgeni: Yeah, the foremandist would save you, but as a user, I'd expect `package-1.2.3-4.el7` and `package-1.2.3-4.el8` being the same source, built for different dists. As a -4 already exists, we would pretend that this -4 and the existing -4 are the same, but built for 2.0 vs 2.1, which they won't be.

ekohl: Can't we pick the -4 change and go -5?

evgeni: We could.

evgeni: Well, we could, but then we'd also have to pick some newer changes as that in -4 is broken.

PR #5528, opened 2020-07-13
Update katello to 3.15.3

evgeni:


evgeni:
`Gem loading error: Could not find 'foreman-tasks' (<= 1.1.1) - did find: [foreman-tasks-1.1.3]`

PR #5533, opened 2020-07-14
Patch rubygem-safemode to allow to_sentence for array

dgoetz:
Safemode has not been released in a while but now includes a fix I need, so I opted for bumping to snapshot version.

### DIFF
--- a/packages/foreman/rubygem-safemode/rubygem-safemode.spec
+++ b/packages/foreman/rubygem-safemode/rubygem-safemode.spec
@@ -7,11 +7,13 @@ Summary: A library for safe evaluation of Ruby code
 Name: %{?scl_prefix}rubygem-%{gem_name}
 
 Version: 1.3.5
-Release: 3%{?dist}
+Release: 4%{?dist}
 Group: Development/Ruby
 License: MIT
 URL: https://github.com/svenfuchs/safemode
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+# Patch safemode to allow to_sentence for array
+Patch0: https://patch-diff.githubusercontent.com/raw/svenfuchs/safemode/pull/32.patch
 Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}rubygems
 Requires: %{?scl_prefix}rubygem(ruby2ruby) >= 2.4.0
@@ -45,6 +47,7 @@ This package contains documentation for rubygem-%{gem_name}.
 gem unpack %{SOURCE0}
 %{?scl:"}
 %setup -q -D -T -n  %{gem_name}-%{version}
+%patch0 -p1
 
 %{?scl:scl enable %{scl} "}
 gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
@@ -86,6 +89,9 @@ rm %{buildroot}%{gem_instdir}/{VERSION,.travis.yml}
 %{gem_docdir}
 
 %changelog
+* Tue Jul 14 2020 Dirk Goetz <dirk.goetz@netways.de> - 1.3.5-4
+- Add patch to allow to_sentence for array
+
 * Wed Apr 08 2020 Zach Huntington-Meath <zhunting@redhat.com> - 1.3.5-3
 - Bump to release for EL8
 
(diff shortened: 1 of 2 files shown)


ekohl:
Have you asked for a release? IIRC we also have various people with commit access and https://rubygems.org/gems/safemode suggests 2 former devs also have release access.

dgoetz:
I did not ask explicitly, only implied in my pull request which fixed my problem. Shall I open an issue to have a new release over there?

ekohl:
Please do. If it isn't addressed, we can link to it.

dgoetz:
Ok, lets see if we get a new release: https://github.com/svenfuchs/safemode/issues/33

dgoetz:
No release yet for the safemode gem. 
@ekohl How do you want to progress? Should I add a note to the spec or is the release version good enough as indicator?

ekohl:
Given the lack of response upstream, I'm ok with proceeding here.

On the technical details: where does it get the tarball from? Is the preview tarball on Rubygems?

dgoetz:
The gem is not on rubygems. It is a `gem build` on the date of the commit that is indicated by the snapshot version in the package release, so `20200714gitd9b1630` stands for commit `d9b1630` and build date `14.07.2020`. 

I have no idea if there is a better way for ruby gems, because for most other software I would get the snapshot tarball without a command to run locally.

ekohl:
So normally we get the actual gem via git annex. It uses `Source0` which points to Rubygems. It also doesn't refer to the pre version so I'm wondering if it actually has the patched version. If it doesn't, that may indicate a problem in our build chain.

dgoetz:
I checked the gem in via git annex, so if it uses this it would be the patched version. I locally build the package and have it already installed on a productive environment where the patch originates from. 

The packages are removed from koji output, so I can not verify koji build packages at the moment. Shall I rerun the tests?

Another approach would be using the release gem and then apply patches. Not sure if this actually works, but I could give it a try. Would this be preferred?

ekohl:
[test rpm] since the scratch builds have been rotated.

dgoetz:
Ok, it seems like it is not taking the source from git annex (package is missing the to_sentence from the fix). So I will try to use the original gem and a patch.

dgoetz:
@ekohl Not sure about versioning, commit message and if patch should go to git annex. If I am correct, strictly by the rule is adding a patch only a new package release, or should it be named differently? Are commit message and changelog entry ok? Should the patch also go to git annex like the gems?

ekohl:
My understanding is that git annex is a way of indirection, storing only a checksum of the file. We can derive it from the Source, so know it comes from Rubygems. We do this to prevent storing big binaries in git. I'm not sure how the previous build passed since my understanding is that it shouldn't have.

The patch can live on some HTTP source, but this is small enough to just inline.

For the commit message, you can consider describing the patch (`Patch safemode to allow to_sentence`). You may also want to do the same in the filename, but since we (luckily) have next to no patches, it's not that important.

dgoetz:
Ok, I added a comment to the spec, but left the patch name unchanged as it is available by this name via github url. Furthermore I adjusted the commit message and title of the PR. Tests are green, so everything is ready for a hopefully final review. Thanks!

packages/foreman/rubygem-safemode/rubygem-safemode.spec
```diff
@@ -7,11 +7,13 @@ Summary: A library for safe evaluation of Ruby code
...
 
 Version: 1.3.5
-Release: 3%{?dist}
+Release: 4%{?dist}
 Group: Development/Ruby
 License: MIT
 URL: https://github.com/svenfuchs/safemode
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+# Patch safemode to allow to_sentence for array
+Patch0: https://patch-diff.githubusercontent.com/raw/svenfuchs/safemode/pull/32.patch
```
evgeni, line 16: You're not annexing the patch, so this can be just `32.patch` or even better `allow-to_sentence.patch` ;)

evgeni: not blocking, as it works as it is

dgoetz: I wanted to keep the URL so it is easily recognizable that the patch is upstream.

PR #5572, opened 2020-07-21
Rename skipped npm modules to clearer terms

tbrisker:


evgeni:
I'd have went with "excluded" not "skipped", but fine either way.

tbrisker:
@evgeni you got it, that is indeed a better choice.

PR #5573, opened 2020-07-21
Exclude @theforeman/find-foreman npm package

johnpmitsch:
Package is just used for testing


evgeni:
Why do we need it packaged, if it's testing only?

PR #5574, opened 2020-07-21
Add downshift package

johnpmitsch:


comps/comps-foreman-plugins-el8.xml
```diff
@@ -266,9 +266,9 @@
       <packagereq type="default">rubygem-jquery-matchheight-rails-doc</packagereq>
       <packagereq type="default">rubygem-kubeclient-doc</packagereq>
       <packagereq type="default">rubygem-logify-doc</packagereq>
+      <packagereq type="default">rubygem-msgpack-doc</packagereq>
```
tbrisker, line 269: this change doesn't seem related

comps/comps-foreman-plugins-rhel7.xml
```diff
@@ -270,9 +270,9 @@
       <packagereq type="default">tfm-rubygem-jquery-matchheight-rails-doc</packagereq>
       <packagereq type="default">tfm-rubygem-kubeclient-doc</packagereq>
       <packagereq type="default">tfm-rubygem-logify-doc</packagereq>
+      <packagereq type="default">tfm-rubygem-msgpack-doc</packagereq>
```
tbrisker, line 273: ditto

package_manifest.yaml
```diff
@@ -1329,8 +1330,8 @@ pulpcore_packages:
     python-inflection: {}
     python-iniparse:
       koji_tags:
-      - name: katello-pulpcore-nightly-el7
-        dist: '.el7'
+        - name: katello-pulpcore-nightly-el7
+          dist: '.el7'
```
tbrisker, line 1334: ditto

package_manifest.yaml
```diff
@@ -1391,8 +1392,8 @@ pulpcore_packages:
     python-zipp: {}
     subscription-manager:
       koji_tags:
-      - name: katello-pulpcore-nightly-el7
-        dist: '.el7'
+        - name: katello-pulpcore-nightly-el7
+          dist: '.el7'
```
tbrisker, line 1396: ditto

johnpmitsch:
This was a package update - see the change [here](https://github.com/Katello/katello/commit/8f592663f699f07656cb37cbda847eac74df5116#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L49)

Let me know if this change needs anything else, I followed the instructions [here](https://github.com/theforeman/foreman-packaging/tree/rpm/develop#updating-npm-packages)

johnpmitsch:
@tbrisker what should I do differently?

packages/foreman/nodejs-downshift/nodejs-downshift.spec
```diff
@@ -0,0 +1,92 @@
...
+%global npm_name downshift
+
+Name: %{?scl_prefix}nodejs-downshift
+Version: 5.4.7
+Release: 1%{?dist}
+Summary: A set of primitives to build simple, flexible, WAI-ARIA compliant React autocomplete components
+License: MIT
+Group: Development/Libraries
+URL: https://github.com/downshift-js/downshift#readme
+Source0: https://registry.npmjs.org/@babel/runtime/-/runtime-7.10.5.tgz
```
ekohl, line 13: I think all these dependencies except `compute-scoll-into-view` are already present in our packaging. To reduce duplication I'd suggest to add that one and this package both using the `single` strategy. This keeps our packages smaller

comps/comps-foreman-rhel7.xml
```diff
@@ -34,6 +34,7 @@
       <packagereq type="default">foreman-service</packagereq>
       <packagereq type="default">foreman-telemetry</packagereq>
       <packagereq type="default">foreman-vmware</packagereq>
+      <packagereq type="default">nodejs-downshift</packagereq>
```
ekohl, line 37: You need to do the same thing for EL8. The script doesn't handle that. Same thing for the package manifest.

tbrisker:
> @tbrisker what should I do differently?

Ideally, find out why the script caused these changes and fix it :) barring that, manually make the needed changes afterwards

comps/comps-foreman-rhel7.xml
johnpmitsch: updated, but not sure where I should manually update the package manifest?

johnpmitsch:
@ekohl did the downshift package exist in foreman-packaging before? I didn't add it, only updated it, it seems like everything is new here

ekohl:
It exists as https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/katello/nodejs-downshift

johnpmitsch:
@ekohl what is the best way to update that? I'm not really sure how to proceed

ekohl:
Use the undocumented 4th option:
https://github.com/theforeman/foreman-packaging/blob/3fd847ea0a3f573d955d8acb55875fb9761401d3/add_npm_package.sh#L8

Bonus points for the person who updates the docs:
https://github.com/theforeman/foreman-packaging/blob/3fd847ea0a3f573d955d8acb55875fb9761401d3/add_npm_package.sh#L117-L123

johnpmitsch:
@ekohl thanks, updated!

rel-eng/tito.props
```diff
@@ -64,6 +64,7 @@ whitelist = foreman
   nodejs-change-emitter
   nodejs-classnames
   nodejs-compression-webpack-plugin
+  nodejs-compute-scroll-into-view
```
tbrisker, line 67: This should be on the katello list

package_manifest.yaml
```diff
@@ -366,6 +366,7 @@ foreman_nodejs_packages:
     nodejs-classnames: {}
     nodejs-compression-webpack-plugin:
       strategy: bundle
+    nodejs-compute-scroll-into-view: {}
```
tbrisker, line 369: this should be in `katello_packages` section

comps/comps-foreman-rhel7.xml
```diff
@@ -34,6 +34,7 @@
       <packagereq type="default">foreman-service</packagereq>
       <packagereq type="default">foreman-telemetry</packagereq>
       <packagereq type="default">foreman-vmware</packagereq>
+      <packagereq type="default">nodejs-compute-scroll-into-view</packagereq>
```
tbrisker, line 37: this should be added to `comps-katello-server-rhel7.xml` as `tfm-nodejs-compute-scroll-into-view` and to `comps-katello-el8.xml` as is and removed from here

packages/katello/nodejs-downshift/nodejs-downshift.spec
```diff
@@ -4,18 +4,24 @@
...
+Version: 5.4.7
+Release: 1%{?dist}
 Summary: A set of primitives to build simple, flexible, WAI-ARIA compliant React autocomplete components
 License: MIT
 Group: Development/Libraries
-URL: https://github.com/paypal/downshift#readme
-Source0: https://registry.npmjs.org/%{npm_name}/-/%{npm_name}-%{version}.tgz
-%if 0%{?scl:1}
-BuildRequires: %{?scl_prefix_nodejs}npm
-%else
```
ekohl, line 16: This part was needed. I think you may have an old npm2rpm version. It's also needed in the other package.

rel-eng/tito.props
ekohl: Also needs to be in EL8.

comps/comps-foreman-rhel7.xml
tbrisker: this comment is still outstanding

johnpmitsch: was still updating! It should be good now

johnpmitsch:
@ekohl @tbrisker updated, but I'm not sure why the tests are failing, is something not correct in the spec files?

tbrisker:
This is a new one, i don't think i've seen it before:

```
+ /usr/lib/rpm/nodejs-symlink-deps /usr/lib/node_modules
ERROR: the path for dependency "@babel/runtime" already exists
This could mean that bundled modules are being installed.  Bundled libraries are
forbidden in Fedora. For more information, see:
    <https://fedoraproject.org/wiki/Packaging:No_Bundled_Libraries>
    
It is generally reccomended to remove the entire "node_modules" directory in
%prep when it exists. For more information, see:
    <https://fedoraproject.org/wiki/Packaging:Node.js#Removing_bundled_modules>
    
If you have obtained permission from the Fedora Packaging Committee to bundle
libraries, please use `%nodejs_fixdep -r` in %prep to remove the dependency on
the bundled module. This will prevent an unnecessary dependency on the system
version of the module and eliminate this error.
error: Bad exit status from /var/tmp/rpm-tmp.a6ahNg (%install)
RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.a6ahNg (%install)
Child return code was: 1
EXCEPTION: [Error()]
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.7/site-packages/mockbuild/util.py", line 755, in do_with_status
    raise exception.Error("Command failed: \n # %s\n%s" % (command, output), child.returncode)
mockbuild.exception.Error: Command failed: 
```

tbrisker:
Probably not really a runtime dependency, it was initially added in https://github.com/downshift-js/downshift/commit/da3863e2b96287bc88914c706b6fdaac614c390a and looks like is only used in testing afaict.
it might be safe to try dropping it from the specfile and see if that resolves the issue, though i think we have other npm modules that require `@bla/bla` style modules?

tbrisker:
or rather - require `babel-runtime` instead of `@babel/runtime`

packages/katello/nodejs-downshift/nodejs-downshift.spec
```diff
@@ -4,18 +4,26 @@
...
 BuildRequires: nodejs-packaging
 %endif
+Requires: npm(@babel/runtime) >= 7.10.2
+Requires: npm(@babel/runtime) < 8.0.0
+Requires: npm(compute-scroll-into-view) >= 1.0.14
+Requires: npm(compute-scroll-into-view) < 2.0.0
+Requires: npm(prop-types) >= 15.7.2
+Requires: npm(prop-types) < 16.0.0
+Requires: npm(react-is) >= 16.13.1
+Requires: npm(react-is) < 17.0.0
```
tbrisker, line 26: I think this should fix it:
```suggestion
Requires: %{?scl_prefix}npm(babel-runtime) >= 7.10.2
Requires: %{?scl_prefix}npm(babel-runtime) < 8.0.0
Requires: %{?scl_prefix}npm(compute-scroll-into-view) >= 1.0.14
Requires: %{?scl_prefix}npm(compute-scroll-into-view) < 2.0.0
Requires: %{?scl_prefix}npm(prop-types) >= 15.7.2
Requires: %{?scl_prefix}npm(prop-types) < 16.0.0
Requires: %{?scl_prefix}npm(react-is) >= 16.13.1
Requires: %{?scl_prefix}npm(react-is) < 17.0.0
```

packages/katello/nodejs-compute-scroll-into-view/nodejs-compute-scroll-into-view.spec
```diff
@@ -0,0 +1,50 @@
...
+Release: 1%{?dist}
+Summary: The engine that powers scroll-into-view-if-needed
+License: MIT
+Group: Development/Libraries
+URL: https://scroll-into-view-if-needed.netlify.com
+Source0: https://registry.npmjs.org/%{npm_name}/-/%{npm_name}-%{version}.tgz
+%if 0%{?scl:1}
+BuildRequires: %{?scl_prefix_nodejs}npm
+%else
+BuildRequires: nodejs-packaging
```
tbrisker, line 17: This will likely be needed for el8 to work: (also in the other package)
```suggestion
BuildRequires: nodejs-packaging
BuildRequires: npm
```

packages/katello/nodejs-downshift/nodejs-downshift.spec
ekohl: No it doesn't fix it, but it's a good point: those SCL prefixes are needed.

ekohl:
If it shouldn' be a runtime dependency, you can use `%nodejs_fixdep -r` to fix `package.json`. I'd also open an issue in upstream and add that link as a comment.

packages/katello/nodejs-downshift/nodejs-downshift.spec
```diff
@@ -16,14 +16,14 @@ BuildRequires: %{?scl_prefix_nodejs}npm
...
-Requires: npm(@babel/runtime) >= 7.10.2
-Requires: npm(@babel/runtime) < 8.0.0
-Requires: npm(compute-scroll-into-view) >= 1.0.14
-Requires: npm(compute-scroll-into-view) < 2.0.0
-Requires: npm(prop-types) >= 15.7.2
-Requires: npm(prop-types) < 16.0.0
-Requires: npm(react-is) >= 16.13.1
-Requires: npm(react-is) < 17.0.0
+Requires: %{?scl_prefix}npm(babel-runtime) >= 7.10.2
+Requires: %{?scl_prefix}npm(babel-runtime) < 8.0.0
```
ekohl, line 20: This part was a wrong suggestion.
```suggestion
Requires: %{?scl_prefix}npm(babel/runtime) >= 7.10.2
Requires: %{?scl_prefix}npm(babel/runtime) < 8.0.0
```

tbrisker: are you sure? https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/nodejs-babel-runtime/nodejs-babel-runtime.spec#L4

ekohl: That's not how it's supposed to work and a bug in the babel runtime packaging ...

tbrisker: in any case it doesn't seem to fix the issue. `nodejs-symlink-deps` looks at the `package.json`, not the spec file.

ekohl: Yes `Requires` are unrelated in the build process (unlike `BuildRequires`). They are only added after the fact and are used by other packages.

tbrisker: where would it get `@babel/runtime` installed from though? It's not part of the tarball, could it be installed by the npm `BuildRequires`? 
also, it seems to be linking to `/opt/theforeman/tfm/root/usr/lib/node_modules` and not `${BUILD_ROOT}/opt/theforeman/tfm/root/usr/lib/node_modules`, not sure if that's expected? could it be some leftovers from some other build?

ekohl: Yes, that may be an issue. The NodeJS packaging is painful. The NPM packages are quite often **very** bad themselves and then add to that that the RPM layer on top of it is also mediocre. Like this example, we've found that it appears we're the only ones who actually create NPM packages in SCL.

evgeni:
You can't `BuildRequire` a package from the same PR. Can you please split the PR in two?

ekohl:
@evgeni it shouldn't be a `BuildRequire` since it's mostly just placing the right files in the right directories. I think I already suggested to patch out the dependency:

> If it shouldn' be a runtime dependency, you can use `%nodejs_fixdep -r` to fix `package.json`. I'd also open an issue in upstream and add that link as a comment.

Another problem can be that it essentially packages the same thing twice: once in the root and once in the `preact` directory (with its own `package.json`). That may also be the cause.

packages/katello/nodejs-downshift/nodejs-downshift.spec
```diff
@@ -29,6 +38,7 @@ Provides: %{?scl_prefix}npm(%{npm_name}) = %{version}
 %install
 mkdir -p %{buildroot}%{nodejs_sitelib}/%{npm_name}
 cp -pfr dist %{buildroot}%{nodejs_sitelib}/%{npm_name}
+cp -pfr flow-typed %{buildroot}%{nodejs_sitelib}/%{npm_name}
 cp -pfr package.json %{buildroot}%{nodejs_sitelib}/%{npm_name}
 cp -pfr preact %{buildroot}%{nodejs_sitelib}/%{npm_name}
```
ekohl, line 43: I'd suggest to drop this and try again.
```suggestion
```

tbrisker: let's try that and see if it helps

tbrisker: Looks like that didn't work

ekohl:
Taking a stab at this myself, I can confirm that `nodejs_symlink_deps` can't handle dependencies with a slash in them. A workaround is to add a mkdir:
```
mkdir -p %{buildroot}%{nodejs_sitelib}/%{npm_name}/node_modules/@babel
%nodejs_symlink_deps
```

However, if `@babel/runtime` is not needed you can also remove it using `%nodejs_fixdep -r @babel/runtime`.

tbrisker:
let's try with `%nodejs_fixdep`

tbrisker:
looks like that solved the issue, but now we're failing on mismatching dependency versions

ekohl:
```
[2020-07-27T17:14:20.717Z]         tfm-npm(prop-types) >= 15.7.2
```

We're on 15.6.2 now

```
[2020-07-27T17:14:20.717Z]         tfm-npm(react-is) >= 16.13.1
```

Now on 16.5.2

```
[2020-07-27T17:14:20.717Z]         tfm-npm(babel/runtime) >= 7.10.2
[2020-07-27T17:14:20.717Z]         tfm-npm(babel/runtime) < 8.0.0
```

Now on 6.26.0. Looks like this probalby means a whole bunch of packages will need to be updated.

tbrisker:
Looks like this is a case where bundle might actually be better, trying to see if that builds.

tbrisker:
[test rpm]

johnpmitsch:
Hey all, I was out yesterday but catching up now, thanks for the patience.

What is the latest blocking issue? This is getting a bit past my knowledge of foreman-packaging.

tbrisker:
TBH i'm not sure why it's failing now. i tried switching it back to bundled mode due to the mismatching dependency versions to what we are currently using, and now it tries to download some files from npmjs instead of using the sources in the srpm for some reason. 
Maybe @ehelms has any idea? IIUC @ekohl and @evgeni are stumped as well, but feel free to correct me as well.

ehelms:
I just attempted a clean take on this here: https://github.com/theforeman/foreman-packaging/pull/5593

tbrisker:
Looks like it worked! I guess i messed up the specfile somehow when trying to switch it to a bundle.

PR #5440, opened 2020-06-19
Disable autorequires to prevent rich dependency auto-creation on EL8 …

ehelms:
…for rubygem-katello

PR #5575, opened 2020-07-21
Update katello to 3.16.0

zjhuntin:


evgeni:
this should build fine if you merge #5576, pick https://github.com/theforeman/foreman-packaging/pull/5440 here so that EL8 generates the right deps, and re-generate the rpm dependencies using https://github.com/theforeman/foreman-packaging/blob/rpm/develop/update-requirements.

zjhuntin:
Merging as the issue was fixed with https://github.com/theforeman/foreman-packaging/pull/5576

PR #5585, opened 2020-07-27
Fix pulp-ansible pyyaml depdency

evgeni:


evgeni:
Otherwise it pulls in `python36-PyYAML` from EPEL and that conflicts with our stuff.

PR #5555, opened 2020-07-17
Galaxy NG packaging

evgeni:

PR #5586, opened 2020-07-27
use python3-dateutil for sub-man

evgeni:
when https://github.com/theforeman/foreman-packaging/pull/5555 is merged, we'll have our own dateutil, so we don't need to depend on EPEL anymore (and it won't work if we try to install both anyways).


evgeni:
Judging by the failed build: nope :)

evgeni:
[test rpm]

evgeni:
[test rpm]

PR #5587, opened 2020-07-27
Patch out Ansible and ansible-lint dependencies from galaxy importer

evgeni:
We use system Ansible, which uses on EL7 Python2 and thus the
requirement is never met.

### DIFF
--- a/packages/pulpcore/python-galaxy-importer/python-galaxy-importer.spec
+++ b/packages/pulpcore/python-galaxy-importer/python-galaxy-importer.spec
@@ -3,7 +3,7 @@
 
 Name:           python-%{pypi_name}
 Version:        0.2.5
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Galaxy content importer
 
 License:        Apache-2.0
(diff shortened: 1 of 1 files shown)


packages/pulpcore/python-galaxy-importer/python-galaxy-importer.spec
```diff
@@ -52,6 +52,8 @@ Requires:       python3-semantic-version >= 2.8.4
 # Remove bundled egg-info
 rm -rf %{pypi_name}.egg-info
 
+sed -i -E '/\s+ansible($|-lint)/d' setup.cfg
```
ekohl, line 55: Isn't this stricter?
```suggestion
sed -i -E '/\s+ansible(-lint)?$/d' setup.cfg
```

evgeni: Lint has a versioned dependency, and I didn't want to include the specific version.

And there are not many packages that match `^ansible-lint` anyways.

evgeni: We could do `/\s+ansible(-lint=.*)?/d`

PR #5591, opened 2020-07-28
Drop some selinux requires on installer and katello-selinux

ehelms:

### DIFF
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,7 +1,7 @@
 %{?scl:%global scl_prefix %{scl}-}
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 4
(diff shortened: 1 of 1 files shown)


packages/foreman/katello-selinux/katello-selinux.spec
```diff
@@ -41,7 +41,6 @@ BuildRequires:  /usr/bin/pod2man
 BuildArch:      noarch
 
 Requires:           foreman-selinux
-Requires:           candlepin-selinux >= 3.1.10
```
ekohl, line 44: Can `katello-selinux` work without `candlepin-selinux`? I thought there was a SELinux technical reason it was added here.

ehelms: You are correct -- https://github.com/Katello/katello-selinux/pull/22#discussion_r419471890

ehelms:
Kept only the installer package change and updated the commit message.

PR #5599, opened 2020-07-29
Add crane-selinux sub-package

ehelms:
Depends on and will need updating with https://github.com/Katello/katello-selinux/pull/30


ehelms:
Does foreman-proxy-selinux get installed on a Foreman? I could call it pulp-selinux but I will find that even more confusing as to where it lives, who owns it.

This is kind of a crappy bridge gap. I could name it crane-selinux? Still a bit confusing as to where it comes from, but its a bit more on the nose naming wise.

ekohl:
> Does foreman-proxy-selinux get installed on a Foreman?

AFAIK we never install that and the proxy runs unconfined. I know @lzap wants to finish the work, but hasn't happened yet.

> This is kind of a crappy bridge gap. I could name it crane-selinux? Still a bit confusing as to where it comes from, but its a bit more on the nose naming wise.

I also thought about that and I don't mind it that name.

My concern comes from the fact we have meta packages `katello` and `foreman-proxy-content`. You install one or the other, not both, right?

ehelms:
The intent is to let the installer handle pulling in the new crane-selinux package rather than RPM dependencies given it's need on Foreman server with Pulp and Foreman Proxy with Pulp -- https://github.com/theforeman/foreman-installer/pull/547

packages/foreman/katello-selinux/katello-selinux.spec
```diff
@@ -127,7 +127,66 @@ fi
...
+    # uninstall only
+    if [ $1 -eq 0 ]; then
+        %{_sbindir}/crane-selinux-disable
+    fi
+    # upgrade and uninstall
+    %{_sbindir}/crane-selinux-relabel
+fi
+
+%files -n crane-selinux
+%doc LICENSE crane.fc crane.if crane.te
```
ekohl, line 176: Also not optimal in the main package, but I like to use `%license`
```suggestion
%license LICENSE
%doc crane.fc crane.if crane.te
```

ehelms: :+1: fixed it in both places

ehelms:
Now with fixed comps and moving the katello selinux packages to their proper home given the installer no longer requires them.

PR #5606, opened 2020-07-30
Cherry picks for 2.1

ehelms:
Goes hand in hand with https://github.com/theforeman/foreman-installer/pull/548


ehelms:
This is only used by Katello which is not GA yet. And this would make things cleaner for those who upgrade to 3.16. I'm not clear on what the pain point would be for a Foreman user.

tbrisker:
What is the status here?

ekohl:
@ehelms I'm not worried about a Foreman user, but the fact that we don't remove RPMs from our repos. That makes moving things around more error prone. There may be a risk that the user set up the foreman repo with a higher priority so they don't get updates. While the chances are probably low, not making the move eliminates the risk.

Not moving it does mean there is a little bit of knowledge required when doing release engineering, but we've had the same thing in 3.14 and IIRC 3.13 so nothing new.

PR #5610, opened 2020-07-31
Update rubygem-pulp_rpm_client to 3.5.0

sjha4:
Supported Versions:

- Nightly

Working on associated pulp_rpm python package.


jlsherrill:
same comment as the other gem pr, its showing up as a binary, probably need to git annex add

PR #5614, opened 2020-08-03
update pulp_2to3_migration to 0.2.0b7

evgeni:

### DIFF
--- a/packages/pulpcore/python-pulp_2to3_migration/python-pulp_2to3_migration.spec
+++ b/packages/pulpcore/python-pulp_2to3_migration/python-pulp_2to3_migration.spec
@@ -5,12 +5,12 @@
 # See https://github.com/dgoodwin/tito/pull/333
 Name:           python-pulp_2to3_migration
 Version:        0.2.0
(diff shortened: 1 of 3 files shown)


packages/pulpcore/python-pulp_2to3_migration/python-pulp_2to3_migration.spec
```diff
@@ -23,8 +23,8 @@ BuildRequires:  python3-setuptools
 Summary:        %{summary}
 %{?python_provide:%python_provide python3-%{pypi_name}}
 
-Requires:       python3-pulpcore >= 3.3
-Requires:       python3-jsonschema
+Requires:       python3-pulpcore >= 3.4
+Requires:       python3-jsonschema >=3.0
```
jlsherrill, line 27: need a space after the '=' i believe :)

evgeni: don't tell me how to creatively break RPM spec files! ;-) (fixed!)

PR #5622, opened 2020-08-03
Update rubygem-foreman_maintain to 0.6.9

upadhyeammit:


evgeni:
@tbrisker did you tag this in koji? That's a stable build that won't build otherwise.

tbrisker:
@evgeni I did not, i thought we fixed auto-tagging in stable branches?

evgeni:
No. We fixed building. But this package won't build as the same NVR already exists in koji.

evgeni:
Tagged now.

PR #5631, opened 2020-08-04
Rpm/2.1 cherry pick

sjha4:
Supported Versions:
 * 2.1


jlsherrill:
[test rpm]

404 from centos

jlsherrill:
@evgeni repoclosure for el8 seems pretty messed up on 3.16, any objection to this as is?

evgeni:
the python things were indeed missing, so I tagged them.

faraday should be there, this is probably because it's testing different tags (pulpcore vs ruby)

evgeni:
Given we won't release 3.16 on EL8 anyways (right?), we could also completely disable building and closure there?

@ehelms ?

sjha4:
Might need to hold off for a bit on this one..I am checking on this PR to cherry-pick a change we need in pulp_rpm 3.5..https://github.com/pulp/pulp_rpm/pull/1796 ..

jlsherrill:
@sjha4 lets go forward with this and then backport https://github.com/theforeman/foreman-packaging/pull/5634  when its merged

jlsherrill:
i think we should disable building and repoclosure for 3.16 el8

PR #5633, opened 2020-08-04
Change katello version to 3.17

jlsherrill:

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,10 +5,10 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 3
+%global release 1
 
 Name:       katello
-Version:    4.0.0
+Version:    3.17.0
 Release:    %{?prerelease:0.}%{release}%{?prerelease}%{?dist}
 Summary:    A package for managing application life-cycle for Linux systems
 BuildArch:  noarch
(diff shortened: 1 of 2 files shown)


jlsherrill:
these packages were never in an actual release, should we bother bumping epoch?

jlsherrill:
related katello pr: https://github.com/Katello/katello/pull/8874/

ekohl:
I'm not sure we need to bump epoch since there are no support guarantees for nightly. I would clearly document it since going back in the changelog is a bit odd. It also means untagging in koji. Probably also worth a notice on discourse for users/developers who have such a box.

jlsherrill:
@ekohl you mean a comment in the change log?

ekohl:
Yes, or squash the changelog from 4.0. My main concern is that the changelogs are visible to users and it's surprising to go from 4.0 to 3.17 (and in the future back to 4.0). Wondering what @theforeman/packaging thinks about it.

ehelms:
Yea, squash it down!

ehelms:
I think there are a few dependencies you will need to update with this PR to get it to build.

ehelms:
[test rpm]

ehelms:
```
DEBUG util.py:598:  Error: No Package found for tfm-rubygem(pulp_rpm_client) < 3.5.0
```

jlsherrill:
Packaging actually got ahead of katello in this case:  This pr has the changes we need  https://github.com/Katello/katello/pull/8869/

and then we can update this spec file

packages/katello/katello/katello.spec
```diff
@@ -180,6 +180,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Tue Aug 04 2020 Justin Sherrill <jsherril@redhat.com> 3.17.0-0.1.master
+- change version to 3.17.0
+
 * Thu Jun 18 2020 Barnaby Court <bcourt@redhat.com> - 4.0.0-0.3.master
```
ekohl, line 186: Can you squash this up too?

packages/katello/katello/katello.spec
```diff
@@ -180,6 +180,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Tue Aug 04 2020 Justin Sherrill <jsherril@redhat.com> 3.17.0-0.1.master
```
ehelms, line 183: Can you squash this changelog as well

packages/katello/katello/katello.spec
jlsherrill: hah, wasn't on latest rpm/develop

jlsherrill:
I think this just needs to wait now on whatever magic generates a new gem from katello

ehelms:
[test rpm]

PR #5641, opened 2020-08-06
Update python-psutil to 5.7.2

ehelms:

### DIFF
--- a/packages/client/python-psutil/python-psutil.spec
+++ b/packages/client/python-psutil/python-psutil.spec
@@ -5,18 +5,14 @@
 %global __provides_exclude_from ^(%{python2_sitearch}|%{python3_sitearch})/.*\\.so$
 
 Name:           python-%{srcname}
-Version:        5.6.3
-Release:        5%{?dist}
+Version:        5.7.2
+Release:        1%{?dist}
 Summary:        %{sum}
 
 License:        BSD
 URL:            https://github.com/giampaolo/psutil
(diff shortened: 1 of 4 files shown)


ehelms:
@evgeni Do you know if the above is true? And/or have any suggestions given your recent playing with python packaging?

evgeni:
Setuptools on EL7 could be a Snakeasaurus. So yeah, pretty much possible.

ekohl:
It's just a helper for pip to know how to select it, but that's not very relevant for RPMs since that can also ensure the correct Python version so it's safe to strip it out. See https://github.com/giampaolo/psutil/commit/f7171c45d7cfc1ce68baa7cd0afdaa94e28305a5 as well.

ehelms:
This ended up being something with the format of `extras_requires` which is only used to specify test requires. I went the route of patching this to an empty hash similar to how it was in past releases so that `setup.py` could proceed on EL7.

packages/client/python-psutil/python-psutil.spec
```diff
@@ -57,7 +53,8 @@ ifconfig, nice, ionice, iostat, iotop, uptime, pidof, tty, who, taskset, pmap.
 
 
 %prep
-%autosetup -p0 -n %{srcname}-release-%{version}
+%setup -q -n %{srcname}-release-%{version}
+%patch0 -p1
```
evgeni, line 57: Technically you could have left the %autosetup here, with -p1. But no hard preference either way from me.

ehelms: Sold!

evgeni (approved):

PR #5642, opened 2020-08-06
Refs #30578: Only generate en_US Apipie docs

ehelms:

### DIFF
--- a/plugins/ruby-puppetdb-foreman/debian/rules
+++ b/plugins/ruby-puppetdb-foreman/debian/rules
@@ -16,10 +16,10 @@ build:
 	cp $(PLUGIN).rb /usr/share/foreman/bundler.d/
 	cd /usr/share/foreman && ( \
 		/usr/bin/foreman-ruby /usr/bin/bundle install --local && \
-		/usr/bin/foreman-ruby /usr/bin/bundle exec rake db:migrate RAILS_ENV=development && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
-		OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
-		)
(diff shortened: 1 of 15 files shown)


plugins/ruby-puppetdb-foreman/debian/rules
```diff
@@ -19,6 +19,7 @@ build:
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake db:migrate RAILS_ENV=development && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
 		OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
```
tbrisker, line 21: Looks like we're missing the nulldb change here,
```suggestion
		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
		OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development DATABASE_URL=nulldb://nohost \
```

plugins/ruby-puppetdb-foreman/debian/rules
```diff
@@ -18,8 +18,9 @@ build:
 		/usr/bin/foreman-ruby /usr/bin/bundle install --local && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake db:migrate RAILS_ENV=development && \
```
tbrisker, line 19: This line will fail, no need to migrate with nulldb

ekohl:
> I'm not 100% sure, but i think you need your branch to be called `deb/develop-something` to pass the `if [[ x$repo =~ ^xdevelop ]]; then` condition that decides whether to try and download a tarball or use the source release version

This is indeed true. I tried looking into this, but the Jenkins jobs for this are quite complicated.

PR #5644, opened 2020-08-06
Fixes #30584: fix NoMethodError in change-hostname after foreman-installer fails

jeremylenz:
### To see the issue:

1. Take a snapshot of your satellite
2. In `/sbin/foreman-installer`, comment out the line
```
@result = Kafo::KafoConfigure.run
```
and add the following below it:
```
exit(1)
```
3. Run change-hostname:
```
[root@sat-6-8-qa-rhel7 vagrant]# satellite-change-hostname sat2.example.com -u admin -p changeme
```
4.  See the error:
```
foreman-installer --scenario satellite -v --disable-system-checks --certs-regenerate=true --foreman-proxy-register-in-foreman true
restoring last_scenario.yaml
/usr/share/katello/hostname-change.rb:382:in `restore_last_scenario_yaml': undefined method `path' for nil:NilClass (NoMethodError)
	from /usr/share/katello/hostname-change.rb:533:in `block in run'
	from /usr/share/katello/helper.rb:59:in `run_cmd'
	from /usr/share/katello/hostname-change.rb:530:in `run'
	from /sbin/satellite-change-hostname:23:in `<main>'
```

### To test:

1. Revert/restore your snapshot
2. Repeat step 2 above
3. Apply this patch to `/usr/share/katello/hostname-change.rb`
4. Run change-hostname again
5. Verify that the error does not appear.  You should now be free to fix the issue with foreman-installer and run it again.

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,7 +5,7 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 1
+%global release 2
 
 Name:       katello
 Version:    3.18.0
@@ -180,6 +180,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Tue Aug 11 2020 Jeremy Lenz <jlenz@redhat.com> - 3.18.0-0.2.master
+- Fixes #30584: fix NoMethodError in change-hostname after foreman-installer fails
+
 * Tue Aug 11 2020 Eric D. Helms <ericdhelms@gmail.com> - 3.18.0-0.1.master
 - Bump to 3.18.0
 
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -15,8 +15,6 @@ module KatelloUtilities
   class HostnameChange
     include ::KatelloUtilities::Helper
 
-    attr_accessor :temp_last_scenario_yaml
-
     def initialize(init_options)
       @default_program = self.get_default_program
       @proxy = init_options.fetch(:proxy)
@@ -375,11 +373,11 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
     def restore_last_scenario_yaml
       STDOUT.puts 'restoring last_scenario.yaml'
       if File.exist?(last_scenario_yaml)
-        run_cmd("cp #{temp_last_scenario_yaml.path} #{last_scenario_yaml}")
+        run_cmd("cp #{@temp_last_scenario_yaml.path} #{last_scenario_yaml}")
       else
         # if the installer failed early the last_scenario symlink won't exist
         scenario_path = "#{scenarios_path}/#{@options[:scenario]}.yaml"
-        run_cmd("cp #{temp_last_scenario_yaml.path} #{scenario_path}")
+        run_cmd("cp #{@temp_last_scenario_yaml.path} #{scenario_path}")
         File.symlink(scenario_path, last_scenario_yaml)
       end
     end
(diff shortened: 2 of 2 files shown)


ekohl:
Just reading the code: how does this fix the issue? I'm failing to see how this is a difference.

jeremylenz:
@ekohl I'm not quite sure either, but I think the issue started after `restore_last_scenario_yaml` was extracted to a method.  (It used to be part of `run`.)  You'd think that since `temp_last_scenario_yaml` had an `attr_accessor` it should work fine as is, but it seems that something sets it to `nil` by the time that method is run, and I'm not sure how.   I've verified that it works with this change and doesn't work without it.

jeremylenz:
update: It's because `temp_last_scenario_yaml` is treated as a local variable by Ruby, since it's inside the method now.  Using `self.temp_last_scenario_yaml` would have fixed it as well.

ehelms:
@jeremylenz Would you mind including a bump to the release so it gets built?

```
obal bump-release katello --changelog "Fixes #30584: fix NoMethodError in change-hostname after foreman-installer fails "
```

jeremylenz:
@ehelms I downloaded podman to try to get obal working, but I get authentication errors when trying 
```
podman run -v `pwd`:/opt/packaging -v ~/.koji:/root/.koji obal:latest scratch katello`
```
I think it's because I don't have Koji credentials.  :cry:

jeremylenz:
@ehelms the `add_changelog.sh` script added the entry, but I had to change the version number manually and fix a merge conflict.  please let me know if it looks ok..

packages/katello/katello/katello.spec
```diff
@@ -180,6 +180,12 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+
+* Tue Aug 11 2020 Jeremy Lenz <jlenz@redhat.com> - 4.0.0-0.4.master 
+- Fixes #30584: fix NoMethodError in change-hostname after foreman-installer fails
+
+* Thu Jun 18 2020 Barnaby Court <bcourt@redhat.com> - 4.0.0-0.3.master
```
ehelms, line 187: You can drop this line.

packages/katello/katello/katello.spec
```diff
@@ -180,6 +180,12 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+
```
ehelms, line 183: No empty line here.

packages/katello/katello/katello.spec
```diff
@@ -180,6 +180,12 @@ Provides a federation of katello services
...
+* Tue Aug 11 2020 Jeremy Lenz <jlenz@redhat.com> - 4.0.0-0.4.master
```
ehelms, line 184: Almost there. Can you change (https://github.com/theforeman/foreman-packaging/pull/5644/files#diff-524a56c89a7dd0e9dbe2443042774384R8) to be a `2` and then update this to be `3.17.0-0.2.master`

chris1984:
Starting to test

packages/katello/katello/katello.spec
jeremylenz: @ehelms Change what to a `2`?  (The link goes to this commit)

jeremylenz:
@ekohl @ehelms Rebased and updated

packages/katello/katello/katello.spec
jeremylenz: Assuming you meant https://github.com/theforeman/foreman-packaging/blob/f0dc385aa362ad2bbe0e74c49b4495b3fb11267f/packages/katello/katello/katello.spec#L8

chris1984:
[test rpm]

jeremylenz:
@ehelms Rebased and fixed merge conflict in `katello.spec`.  please let me know what else this needs, if anything!

packages/katello/katello/katello.spec
```diff
@@ -183,6 +183,9 @@ Provides a federation of katello services
 * Tue Aug 11 2020 Eric D. Helms <ericdhelms@gmail.com> - 3.18.0-0.1.master
 - Bump to 3.18.0
 
+* Tue Aug 11 2020 Jeremy Lenz <jlenz@redhat.com> - 3.17.0-0.2.master
```
ehelms, line 186: Bump this to `3.18.0-0.2.master` and move this entry to just below `%changelog` and you'll be good to go.

jeremylenz:
@ehelms @ekohl Thanks, updated

PR #5646, opened 2020-08-07
Rpm/develop update rubygem kafo parsers 1.1.1

ehelms:


ehelms:
Follow up -- https://github.com/theforeman/foreman-packaging/pull/5647

PR #5653, opened 2020-08-10
add ansible-lint to el8 comps

evgeni:


ehelms:
This is only needed on EL8?

evgeni:
Correct, we don't build it on EL7 (that'd be too complicated) as discussed with the Ansible Galaxy team.

PR #5654, opened 2020-08-10
Bump galaxy ng

evgeni:


evgeni:
draft because this needs pulpcore 3.5

PR #5660, opened 2020-08-11
Update foreman_monitoring to 2.0.0

laugmanuel:
(cherry picked from commit c9b6dcbd91815435d1dde92ac559c78e2ececfb8)


laugmanuel:
@ekohl , for whatever reason the build seems to fail for Foreman 2.0:
```
[...]
 WARN  app : Creating scope :path_of. Overwriting existing method Hostgroup.path_of.
2020-08-11T11:44:21 [W|app|] Creating scope :path_of. Overwriting existing method Hostgroup.path_of.
** Execute db:load_config
** Invoke plugin:refresh_migrations (first_time)
** Invoke environment 
** Execute plugin:refresh_migrations
** Execute db:migrate
== 201910180900 RenameDowntimeHostPermission: migrating =======================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:
Could not find table 'permissions'
[...]
```

I can verify this locally as well. Do you have any idea why it works with 2.1 and not with 2.0?

ekohl:
No, I'm quite unfamiliar with the migration internals. Maybe @tbrisker knows more, or at least who would know.

laugmanuel:
@tbrisker do you have any idea for this behaviour on Foreman 2.0?

tbrisker:
in Foreman 2.1 we moved to using `nulldb` for the builds so that migrations don't actually do anything during the build. I assume this means that there might be an issue with the migration itself, or more likely that there is an issue with the order in which migrations run during the build with the plugin migrations running before the core migrations.
[test rpm] so we can take another look at the logs which expired since than.

laugmanuel:
That would explain the different behavior of the two builds.
I will have a look at the migration next week and see if I can figure out the cause for the problem.
We’ve been using the patch for some time now, but applied it onto of a working installation, which did not trigger the problem (because core migrations were already applied)

tbrisker:
Looking at the log, it seems that indeed core migrations are not executed. This is because https://github.com/theforeman/foreman-packaging/pull/4607 dropped them from core `posttrans` section, so when the builder installs foreman they are not executed. I'm not quite sure what part of the spec file triggers the `db:migrate` during the build, but we likely need to change it to skip this step in the 2.0 branch

laugmanuel:
@dgoetz can you have a look at this? ;)

dgoetz:
Sorry, for stable branches I am not sure if I have all privileges/knowledge, so it is better if someone else can have a look.

dgoetz:
[test rpm] to see if something changed and how we can process here

dgoetz:
Ok, here are my findings: The macros `foreman_precompile_plugin` with `-a` calls the rake tasks `db:create` and `db:migrate` which fails. For 2.1 this got removed. 
@theforeman/packaging Would removing this from the macro in foreman-plugin and releasing a new version to 2.0 be a valid option? As it is created from the Foreman's spec itself it would cause a complete new release also if a bump for the package release would be enough and no minor version would be required.

ekohl:
The removal in 2.1 was accompanied by actual changes in the installer. It is not something that should be cherry picked to 2.0. The precompile macro is still needed to compile assets. However, it is weird that it would fail.

The only thing I can imagine is that `db:migrate` fails on sqlite which is used in 2.0 packaging.

evgeni:
given 2.0 is EOL and this didn't build anyways, I am going to close the PR now.

PR #5661, opened 2020-08-11
Update foreman_monitoring to 2.0.0

laugmanuel:
(cherry picked from commit c9b6dcbd91815435d1dde92ac559c78e2ececfb8)


laugmanuel:
Maybe we should wait for https://github.com/theforeman/foreman-packaging/pull/5660 as well as there seems to be a problem during the RPM build for Foreman 2.0..

tbrisker (approved):
Actually we don't need to wait for the other one, the issue was only for 2.0 branch which dropped the migrations but didn't switch to nulldb yet. This one is fine.

PR #5664, opened 2020-08-11
Update nightly to 2.2

ehelms:


package_manifest.yaml
```diff
@@ -42,12 +42,12 @@ foreman_scl_packages:
     scl: 'tfm'
     package_base_dir: 'packages/foreman/'
     koji_tags:
-      - name: foreman-nightly-rhel7
+      - name: foreman-2.2-rhel7
```
ekohl, line 45: Makes me wonder if we can set a global variable for the Foreman and Katello versions to prevent this in the future to make this easier. Similar to how we have `foremandist`.

ehelms:
I'll double check and cleanup Fedora 29 across the board.

package_manifest.yaml
ehelms: I similarly pondered while doing it.

ekohl: And it also shows why using Foreman versions (2.2) instead of Katello (3.17) versions would make branching easier since there's just a trivial replacement of nightly.

PR #5668, opened 2020-08-12
Release 2.2.0-rc1

ehelms:


ehelms:
[test rpm]

ehelms:
Good catch! This is the culprit -- https://github.com/theforeman/foreman-packaging/pull/5669

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -146,6 +146,9 @@ done
 %{_sbindir}/foreman-proxy-certs-generate
 
 %changelog
+* Wed Aug 12 2020 Eric D. Helms <ericdhelms@gmail.com> - 1:2.2.0-0.6.rc1
+- Release foreman-installer 2.2.0
```
tbrisker, line 150: Not sure if it's important, but i think the message should be 
```suggestion
- Release foreman-installer 2.2.0-rc1
```
(also for the other packages)

ehelms: Seems to be how we've done it in the past, based on the script. I'll have to add some functionality to do this in the future.

PR #5680, opened 2020-08-17
update ansible collection to 1.1.0

evgeni:
this has been established in Fedora, let's follow suit


evgeni:
Fedora link: https://src.fedoraproject.org/rpms/ansible/blob/master/f/ansible-generator

evgeni:
[test rpm]

PR #5695, opened 2020-08-19
update python-box to 5.1.0

jlsherrill:


ekohl:
Should we have some dependency on this version? Dynaconf now depends on this but if it needs 5.1.0 then it should specify so.

PR #5696, opened 2020-08-19
Add all group so version variables are available to packages and repo…

ehelms:
Found by https://ci.theforeman.org/job/katello-nightly-rpm-pipeline/641/console

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -1,10 +1,16 @@
 ---
-packages:
+all:
   vars:
     foremandist: .fm2_3
-    rpmlint_rc_file: "{{ inventory_dir }}/.rpmlintrc"
     foreman_version: 'nightly'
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -1,8 +1,16 @@
...
+  children:
+    packages: {}
+    repoclosures: {}
+
 packages:
   vars:
     foremandist: .fm2_3
-    rpmlint_rc_file: "{{ inventory_dir }}/.rpmlintrc"
     foreman_version: 'nightly'
     katello_version: 'nightly'
```
ekohl, line 15: This looks the wrong way around: rpmlintrc should stay, but remove the moved ones?

ehelms: Yep, good catch.

PR #5698, opened 2020-08-20
Obsolete python36-dateutil to allow upgrades

ehelms:


packages/pulpcore/python-dateutil/python-dateutil.spec
```diff
@@ -17,6 +17,8 @@ BuildRequires:  python3-setuptools
 BuildRequires:  python3-setuptools-scm
 BuildRequires:  python3-six >= 1.5
 
+Obsoletes: python36-dateutil <= 1:2.4.2-5
```
evgeni, line 20: I think this is one of the cases where we want a non-versioned obsoletes? Otherwise if EPEL gets an update and that's installed before our repo is enabled, it'll break again?

evgeni: We're obsoleting a package from EPEL, so epoch/version won't match

ehelms: Hrmm, does that imply we also need a Conflicts?

evgeni: Why?

PR #5697, opened 2020-08-20
Update rubygem-facter to 4.0.44

lzap:
Is the discovery image the only reason why carry Facter in our repo? It
looks like 2.x series have problems with IPv6 and I would like to know
how feasible is to make the switch to 4.x.

* Nightly

EDIT:

Dependant PRs:

* https://github.com/theforeman/foreman-packaging/pull/5699

Related PRs:

* https://github.com/theforeman/foreman-packaging/pull/5700
* https://github.com/theforeman/foreman-packaging/pull/5721

PR #5699, opened 2020-08-20
Add rubygem-hocon package

lzap:
New dependency for facter 4.x - https://github.com/theforeman/foreman-packaging/pull/5697

* Nightly

Please carefully review, I am not sure what I am doing here. I want
facter (thus this package) to be in Foreman core repo in BOTH SCL and
non-SCL. The latter is required by discovery.


lzap:
Aded also to non-SCL version I hope.

ekohl:
> Aded also to non-SCL version I hope.

This isn't needed. non-SCL is now limited to foreman-maintain.

packages/foreman/rubygem-hocon/rubygem-hocon.spec
```diff
@@ -0,0 +1,95 @@
...
+BuildRequires: %{?scl_prefix_ruby}ruby(release)
+BuildRequires: %{?scl_prefix_ruby}ruby >= 1.9.0
+BuildRequires: %{?scl_prefix_ruby}rubygems-devel
+BuildArch: noarch
+Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
+# end specfile generated dependencies
+
+%description
+== A port of the Java {Typesafe Config}[https://github.com/typesafehub/config]
+library to Ruby.
```
ekohl, line 32: Mind fixing this up a bit?

packages/foreman/rubygem-hocon/rubygem-hocon.spec
```diff
@@ -0,0 +1,95 @@
...
+
+%global gem_name hocon
+%global gem_require_name %{gem_name}
+
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 1.3.1
+Release: 1%{?dist}
+Summary: HOCON Config Library
+Group: Development/Languages
+License: Apache License, v2
```
ekohl, line 13: This isn't a valid license code
```suggestion
License: ASL 2.0
```

ekohl (changes requested):
The script to add hasn't been modified to be able to add both EL7 and EL8. Mind fixing that up manually?

lzap:
Tested:

```
# scl enable tfm -- irb
irb(main):001:0> require "hocon"
=> true
```

lzap:
> This isn't needed. non-SCL is now limited to foreman-maintain.

Not sure what you mean. I also need the non-SCL package to be published. Guide me what to do if I do this wrong please.

ekohl:
Why do you need the non-SCL package? Smart Proxy runs in SCL too nowadays.

lzap:
> Why do you need the non-SCL package? Smart Proxy runs in SCL too nowadays.

But on the discovery image we have a ton of Ruby 2.0 (system Ruby) code and we also need "facter" command itself. I'd like to refactor this but there is no time at the moment.

ekohl:
AFAIK we don't build facter itself as non-SCL now. Do you intend to change that?

lzap:
> AFAIK we don't build facter itself as non-SCL now. Do you intend to change that?

Yeah I know. This is what I was stating from the first line, yes. It's a small package, should not be big deal.

lzap:
For the record, the plan with discovery is to completely get rid of Smart Proxy and base it on SSH protocol. I am thinking even going far and removing Ruby and Facter with uFacter to minimize the image even further because it is huge and has been growing last couple of years. But that's the future, not on my agenda for this year.

ekohl:
> Yeah I know. This is what I was stating from the first line, yes. It's a small package, should not be big deal.

Well, it is a big deal because Facter 4 requires Ruby ~> 2.3 so you can't use the system Ruby 2.0.

lzap:
> Well, it is a big deal because Facter 4 requires Ruby ~> 2.3 so you can't use the system Ruby 2.0.

F^&%^&^%&#%#$@*{k. Thanks for letting me know, you saved me some time. I guess I will need to SCL everything. Damn.

lzap:
I amended the changes, hold on with merging for now. I need to take a look.

ekohl:
> F^&%^&^%&#%#$@*{k. Thanks for letting me know, you saved me some time. I guess I will need to SCL everything. Damn.

From the top of my head, the system ruby scripts are just placed in the directories rather than a proper gem. That means it probably doesn't work out of the box on EL8 either. However, you may want to consider that option as well.

lzap:
Ok moving these packages: https://github.com/theforeman/foreman-packaging/pull/5699

ekohl:
You're linking to this PR. Did you intend #5700?

lzap:
Yes that was my intention.

Anyway, I am going to ugprade Facter (and whole FDI) to SCL Ruby. This is ready for review.

lzap:
Almost there, this is the last dependency before I can upgrade facter to 4.0 on SCL ruby and remove Ruby 2.0 from FDI.

lzap:
This PR was closed by GH for no reason, this is the 2nd time and I am sure I haven't closed this by myself. I use 2FA and my account is secure, not sure what happened.

lzap:
Rebased, are we good?

ekohl (changes requested):
You still need to add it to the EL8 comps file and in the EL8 tag in tito.props.

lzap:
Rebased. I could merge this into https://github.com/theforeman/foreman-packaging/pull/5697 however since there are already references from that PR I'd probably prefer to keep this one separate.

I noticed that branch name does not start with `rpm/`, I think this is an issue at some point. Maybe merging it with https://github.com/theforeman/foreman-packaging/pull/5697 is a good idea.

ekohl:
> I noticed that branch name does not start with `rpm/`, I think this is an issue at some point.

That's only something for deb builds.

comps/comps-foreman-el8.xml
```diff
@@ -238,6 +238,7 @@
       <packagereq type="default">rubygem-hirb</packagereq>
       <packagereq type="default">rubygem-hirb-unicode-steakknife</packagereq>
       <packagereq type="default">rubygem-hoe</packagereq>
+      <packagereq type="default">tfm-rubygem-hocon</packagereq>
```
ekohl, line 241: EL8 doesn't use tfm prefixes.

ekohl (changes requested):

comps/comps-foreman-el8.xml
lzap: Jeez how could have I missed this one? Fixed.

lzap:
Finally green.

lzap:
Just to align expectations - I know you mentioned that I should probably create a single PR for both facter and this dependency, @ekohl but would you mind merging separate since these PRs already exist? Thanks.

ekohl (approved):
Technically you should really ping @theforeman/packaging. Then you'll also reach the build team who is responsible for the repo. I'm trying to reduce my involvement in packaging to focus on the platform.

ekohl:
Btw, the reason I recommended a single PR is that https://github.com/theforeman/foreman-packaging/pull/5697 will fail repoclosure until the RPM pipeline has ran. That's a nightly operation. If you combine it in a single PR, you reduce the effort needed by the packaging team since the PR can be green and just merged without looking at the build failure.

Right now I don't have my koji cert at hand to add the package so not merging now. However, the package looks good now.

evgeni:
> Technically you should really ping @theforeman/packaging. Then you'll also reach the build team who is responsible for the repo. I'm trying to reduce my involvement in packaging to focus on the platform.

Can non members ping that team? Last time I looked they couldn't

tbrisker:
> > Technically you should really ping @theforeman/packaging. Then you'll also reach the build team who is responsible for the repo. I'm trying to reduce my involvement in packaging to focus on the platform.
> 
> Can non members ping that team? Last time I looked they couldn't

Yes, I've made the team public recently for that purpose

evgeni (approved):

PR #5700, opened 2020-08-20
Moved newt and fdi to SCL

lzap:
Moving FDI packages to SCL, moving away from system Ruby 2.0. Can I move
two packages in one PR?

* Nightly

### DIFF
--- a/packages/plugins/rubygem-newt/rubygem-newt.spec
+++ b/packages/plugins/rubygem-newt/rubygem-newt.spec
@@ -2,18 +2,12 @@
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
-%if 0%{?rhel} == 7
-%global gem_extdir_mri_lib %{gem_extdir_mri}/lib
-%else
-%global gem_extdir_mri_lib %{gem_extdir_mri}
-%endif
-
 %global gem_name newt
 %global gem_require_name %{gem_name}
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.9.7
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: Ruby bindings for newt
 Group: Development/Languages
 License: MIT
@@ -74,8 +68,8 @@ mkdir -p %{buildroot}%{gem_dir}
 cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
-mkdir -p %{buildroot}%{gem_extdir_mri_lib}/ruby_newt
-cp -a .%{gem_instdir}/ext/ruby_newt/*.so %{buildroot}%{gem_extdir_mri_lib}/ruby_newt/
+mkdir -p %{buildroot}%{gem_extdir_mri}/ruby_newt
+cp -a .%{gem_instdir}/ext/ruby_newt/*.so %{buildroot}%{gem_extdir_mri}/ruby_newt/
 touch %{buildroot}%{gem_extdir_mri}/gem.build_complete
 
 # Prevent dangling symlink in -debuginfo (rhbz#878863).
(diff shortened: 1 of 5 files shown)


ekohl (commented):
I think it also needs an updated comps and probably also tito.props.

lzap:
Anything else I can do?

evgeni:
Wait until someone from packaging has time to review? ;-)

lzap:
Yeah sorry, I am trying to rush some changes (SCLing FDI) to find if I am actually able to hit my milestone. But no big deal if I won't make it.

packages/plugins/rubygem-newt/rubygem-newt.spec
```diff
@@ -74,24 +68,28 @@ mkdir -p %{buildroot}%{gem_dir}
...
-cp -a .%{gem_instdir}/ext/ruby_newt/*.so %{buildroot}%{gem_extdir_mri_lib}/ruby_newt/
+mkdir -p %{buildroot}%{gem_extdir_mri}/ruby_newt
+cp -a .%{gem_instdir}/ext/ruby_newt/*.so %{buildroot}%{gem_extdir_mri}/ruby_newt/
 touch %{buildroot}%{gem_extdir_mri}/gem.build_complete
 
 # Prevent dangling symlink in -debuginfo (rhbz#878863).
 rm -rf %{buildroot}%{gem_instdir}/{ext,tmp,.require_paths}
 
 %check
+find %{buildroot}
```
evgeni, line 79: this does not belong here I guess? :)

packages/plugins/rubygem-newt/rubygem-newt.spec
```diff
@@ -74,24 +68,28 @@ mkdir -p %{buildroot}%{gem_dir}
...
 %{?scl:scl enable %{scl} - << \EOF}
-# Ideally, this would be something like this:
-# GEM_PATH="%{buildroot}%{gem_dir}:$GEM_PATH" ruby -e "require '%{gem_name}'"
-# But that fails to find native extensions on EL8, so we fake the structure that ruby expects
+%if 0%{?rhel} == 8
```
evgeni, line 81: I don't particularly like the split, but it works (both, the faked %check here and the resulting RPMs)…
So I'm leaning to ack this and move on.

@ekohl opinions?

lzap: I spent quite some time trying to find a clean way but honestly - I already have a prototype based on `whiptail` and `bash` any my plan is to get rid of newt and ruby completely in the future. I want to minimize the image as little as possible.

packages/plugins/rubygem-newt/rubygem-newt.spec
lzap: Helps debugging a lot, you want this out? Won't harm...

packages/plugins/rubygem-newt/rubygem-newt.spec
evgeni: We have the "problem" (that the tests are weird) in all gems that build native extensions, not only `newt`.

ekohl: My only suggestion would be to consider what's more likely to work in the future. Which of the 2 is more likely to work on EL9? So in short: should it be reversed and be `== 7`? To be clear: I don't know the answer, but I have a feeling that EL9 will be closer to 8 than to 7.

evgeni: While I agree that 9 will probably be closer to 8 than to 7, the current flow reads like this:
```
if rhel8
  apply this terrible hack
else
  do this almost sensible thing
end
```

And I think scoping the terrible hack (which it is, really) as tight as possible seems like a good idea? If it'll fail when we rebuild for EL9, we can widen it then.

packages/plugins/rubygem-newt/rubygem-newt.spec
evgeni: I'd prefer not to have it here.

packages/plugins/rubygem-newt/rubygem-newt.spec
ekohl: Looks like it isn't available in Fedora we can't easily check, but I agree the EL7 version looks like the sensible thing. It'd be great if it worked on EL8 too.

packages/plugins/rubygem-newt/rubygem-newt.spec
lzap: Sure, removed. Rebased.

evgeni (approved):
ACK

Not merging as this needs `koji add-pkg` and I don't have my koji credentials here.

PR #5706, opened 2020-08-24
fix python-dateutil obsoletes by placing them to the binary package

evgeni:


evgeni:
otherwise the *S*RPM ends up with the obsoletes :roll_eyes:

ekohl (approved):

PR #5709, opened 2020-08-24
Update rubygem-foreman_fog_proxmox to 0.12.1

tristanrobert:
* Nightly


tristanrobert:
I do not see errors in root.log from 366155 and 366156 jobs :-(

evgeni:
Yeah, koji is sometimes lying and the real error is in build.log:

```
ActiveModel::UnknownAttributeError: unknown attribute 'name' for Permission.
```

tristanrobert:
> Yeah, koji is sometimes lying and the real error is in build.log:
> 
> ```
> ActiveModel::UnknownAttributeError: unknown attribute 'name' for Permission.
> ```

I don't understand why because I didn't change the syntax. Permission on controller is declared as usual: https://github.com/theforeman/foreman/blob/9a251d43967605dbef0bfec79d4ec8e1a7f5e919/test/unit/plugin_test.rb#L580

pcreech:
[test rpm]

ekohl:
This is probably because there is no active record model loaded. However, it is odd that this triggers. You would think that `pending_migrations` returns `true` and returns false, but it doesn't. Otherwise it wouldn't reach that line.

The reason behind that is your plugin has no migrations, so it isn't triggered. This was changed in https://github.com/theforeman/foreman/commit/3a5d4162c0d0cbfaaf690d7b48b273bd097731ea to not look at the base migration but rather the plugin migrations. I guess this really should be looking at both.

That indicates a bug in Foreman itself that should be fixed.

tristanrobert:
> This is probably because there is no active record model loaded. However, it is odd that this triggers. You would think that `pending_migrations` returns `true` and returns false, but it doesn't. Otherwise it wouldn't reach that line.
> 
> The reason behind that is your plugin has no migrations, so it isn't triggered. This was changed in [theforeman/foreman@3a5d416](https://github.com/theforeman/foreman/commit/3a5d4162c0d0cbfaaf690d7b48b273bd097731ea) to not look at the base migration but rather the plugin migrations. I guess this really should be looking at both.
> 
> That indicates a bug in Foreman itself that should be fixed.

Thank you @ekohl for your diagnostic. I have posted an [issue](https://projects.theforeman.org/issues/30821).

tristanrobert:
Fixed in : https://projects.theforeman.org/projects/foreman/repository/revisions/a4f4c1307e97a0c0dc2f633a0307c411844de011
Could you reload the checks ?

evgeni:
@tristanrobert this PR is obsoleted by https://github.com/theforeman/foreman-packaging/pull/5716, right?

evgeni:
I am going to say yes, feel free to yell at me if not ;)

tristanrobert:
> I am going to say yes, feel free to yell at me if not ;)

Yes you are right. Thanx!

PR #5685, opened 2020-08-17
Update rubygem-kafo to 5.1.0

ehelms:

PR #5712, opened 2020-08-25
Update rubygem-kafo to 5.1.0

wbclark:


ekohl (changes requested):
In our current build system you can't update a build dependency in the same PR as where the build dependency is used. That means you need to send a PR that updates kafo and then a separate one that updates the installer.

evgeni:
dupe of https://github.com/theforeman/foreman-packaging/pull/5685?

PR #5719, opened 2020-08-26
Update to kafo 5.1.0

ehelms:


ekohl (approved):
Technically the build dependency is much lower. I'm not sure if the build will pass when they're in the same PR, similar to the problem we have in koji with RPMs. However, I'm not sure foreman-installer is even built after a push. I think it may skip it and rely on the actual nightly package.

PR #5725, opened 2020-08-28
also add the changelog using script_root

evgeni:


ekohl (approved):
That's what I get for trusting you and not properly reviewing the whole script.

PR #5687, opened 2020-08-18
Rpm/vendor v4.14.0

sharvit:
Updating vendor to the latest version, required for: https://github.com/theforeman/foreman/pull/7896

PR #5735, opened 2020-08-31

ekohl:


packages/foreman/foreman/foreman.spec
```diff
@@ -263,7 +263,7 @@ BuildRequires: systemd
 # start package.json devDependencies BuildRequires
 BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
 BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
-BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 4.11.1
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 4.12.0
```
ekohl, line 266: https://github.com/theforeman/foreman/pull/7896 bumps it to 4.14 so I'm going to update this ahead of time.

ehelms: Could you revert it back to 4.12.0 to reflect current state? This is mixing two update concerns that will block builds unless we know that the referenced PR is going in really soon.

ekohl: This doesn't block the build since 4.14.0 is available (https://github.com/theforeman/foreman-packaging/pull/5687). Repoclosure is only complaining about `sd_notify` which is because we didn't mash.

PR #5739, opened 2020-08-31
use new pulpcore repos in katello-repos

evgeni:

### DIFF
--- a/packages/katello/katello-repos/katello-repos.spec
+++ b/packages/katello/katello-repos/katello-repos.spec
@@ -2,12 +2,14 @@
 %global pulp_version 2.21
 %global use_pulp_nightly false
 
+%global pulpcore_version 3.6
+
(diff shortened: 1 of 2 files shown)


packages/katello/katello-repos/katello-repos.spec
```diff
@@ -2,12 +2,14 @@
 %global pulp_version 2.21
 %global use_pulp_nightly false
 
+%global pulpcore_version 3.4
```
evgeni, line 5: @jlsherrill or should we directly set 3.6 here? :)

evgeni:
please wait on Justin with the merge :)

packages/katello/katello-repos/katello-repos.spec
jlsherrill: yeah, go straight to 3.6!

PR #5743, opened 2020-09-01
correct script root definition

evgeni:
otherwise it uses the name of the script, not the folder it is in as the
root, oups


evgeni:
@ekohl this is what you get for trusting me, again ;)

ekohl (approved):

PR #5745, opened 2020-09-01
Update foreman-tasks to 3.0.0

adamruzicka:
CPs:
- [ ] 2.2

### DIFF
--- a/plugins/ruby-foreman-tasks/debian/changelog
+++ b/plugins/ruby-foreman-tasks/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-tasks (3.0.0-1) stable; urgency=low
+
+  * 3.0.0 released
+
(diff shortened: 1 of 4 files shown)


evgeni (approved):

evgeni:
and why exactly did that fail NOW?!

```
10:10:44 Bundler could not find compatible versions for gem "rack-protection":
10:10:44   In Gemfile:
10:10:44     foreman-tasks (= 3.0.0) was resolved to 3.0.0, which depends on
10:10:44       sinatra was resolved to 2.0.8.1, which depends on
10:10:44         rack-protection (= 2.0.8.1)
10:10:44 
10:10:44 Could not find gem 'rack-protection (= 2.0.8.1)', which is required by gem
10:10:44 'sinatra', in any of the sources.
```

evgeni:
fixed and built

PR #5747, opened 2020-09-01
Update foreman_ansible to 6.0.0

adamruzicka:
CPs:
- [ ] 2.2

### DIFF
--- a/plugins/ruby-foreman-ansible/debian/changelog
+++ b/plugins/ruby-foreman-ansible/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-ansible (6.0.0-1) stable; urgency=low
+
+  * 6.0.0 released
+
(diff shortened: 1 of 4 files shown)


tbrisker (changes requested):
I think this also needs an update to `plugins/ruby-foreman-ansible/debian/changelog`

adamruzicka:
Done

evgeni (approved):

evgeni:
and that fails on

```
10:11:52  pbuilder-satisfydepends-dummy : Depends: ruby-foreman-deface (< 2.0.0) which is a virtual package and is not provided by any available package
10:11:52 
10:11:52                                  Depends: ruby-foreman-remote-execution (>= 4.0.0-1) which is a virtual package and is not provided by any available package
```

evgeni:
fixed and built

PR #5751, opened 2020-09-02
Bump Hammer Katello to version 0.24.premaster

chris1984:


chris1984:
@jturel here is the hammer pr for nightly, will make another pr for the 2.2 branch

jturel:
Looking at the changelog gives me the impression that we need to bump this to 0.24 and make that the new version here based on hammer-cli-katello-master, with a separate change for rpm/2.2 to bump the 0.23 version?

chris1984:
> Looking at the changelog gives me the impression that we need to bump this to 0.24 and make that the new version here based on hammer-cli-katello-master, with a separate change for rpm/2.2 to bump the 0.23 version?

ok will update, pinged you on the 2.2 one

chris1984:
@jturel updated

chris1984:
@evgeni

chris1984:
@tbrisker how does this look?

tbrisker (approved):
Thanks @chris1984 !

PR #5754, opened 2020-09-03
Add ruby-sd-notify

ekohl:
This is currently for buster only. This allows me to see if it builds and then iterate on it when I see failures. If it all passes, I'll copy to other distros.


dependencies/bionic/sd_notify/changelog
```diff
@@ -0,0 +1,5 @@
+ruby-sd-notify (0.1.0-1) UNRELEASED; urgency=medium
```
evgeni, line 1:
```suggestion
ruby-sd-notify (0.1.0-1) stable; urgency=medium
```

or unstable, but not unreleased ;)

evgeni:
LGTM otherwise

evgeni (approved):

PR #5787, opened 2020-09-10
Bump foreman-maintain to 0.6.11

upadhyeammit:


upadhyeammit:
The version bump is only required for stable branch hence not created any pr in rpm/develop

PR #5788, opened 2020-09-10
Tidy up cron

ezr-ondrej:


debian/bionic/foreman/changelog
```diff
@@ -1,3 +1,9 @@
+foreman (2.3.0-2) stable; urgency=low
+
+  * Clean up unecessary cron entries
+
+ -- Ondřej Ezr <oezr@redhat.com>  Thu, 10 Sep 2020 14:28:42 +0200
+
```
ekohl, line 6: With Debian packages we don't keep the changelog. This is mostly because we don't really keep any pre tags. I'm not sure if our nightly packages can handle this.

ezr-ondrej: Ah, then I missunderstood you on IRC, I though we are talking about changelog is missing to get the statistics plugin bump in.

ekohl: Well, we keep changelogs for the Foreman RPM, not Foreman deb.

ekohl (approved):

PR #5789, opened 2020-09-10
Update rubygem-foreman_leapp to 0.1.5

stejskalleos:


pcreech (approved):
ACK

looks like this should fix the repoclosure issues.

PR #5782, opened 2020-09-09
Update rubygem-foreman_ansible to 6.0.0

adamruzicka:

PR #5802, opened 2020-09-15
Update rubygem-foreman_leapp to 0.1.5

evgeni:


evgeni:
@ares @stejskalleos can you please ack that 0.1.5 is what should be in 2.2? the current 0.1.3 is not compatible with the rex in 2.2

evgeni:
it also is not compatible with ansible that is slated in https://github.com/theforeman/foreman-packaging/pull/5782

evgeni:
closure only complains the package *in* the repo is broken, which we know ;)

evgeni:
confirmed on IRC, 0.1.5 is the right version.

PR #5807, opened 2020-09-16
load pulpcore gpg keys from the server

evgeni:

### DIFF
--- a/packages/katello/katello-repos/katello-repos.spec
+++ b/packages/katello/katello-repos/katello-repos.spec
@@ -9,7 +9,7 @@
 
 %global prereleasesource nightly
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 2
+%global release 3
 
 Name:           katello-repos
 Version:        3.18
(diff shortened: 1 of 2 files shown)


jlsherrill (approved):

ekohl (commented):
I still think it's good to have GPG keys as part of the RPM since they're signed.

I'd like to see a `pulpcore-release` package that includes the `pulpcore` repo, including the GPG key. Then `katello-repos` should depend on that. However, that's a bit harder to consume.

I've been thinking of having `foreman-release` and `katello-release` both in the main foreman repo (and I guess `pulpcore-release` could live there too) so the instructions become `yum install https://yum.theforeman.org/releases/latest/el7/x86_64/foreman-release.rpm && yum install katello-release`. The benefit is that you don't need to think about which katello release matches with foreman.

Thoughts?

evgeni:
Long term this sounds like a good idea, but sounds like more (design) work to get it right for 3.17 than we have time?

ekohl:
Yes, it is indeed the long term solution.

PR #5809, opened 2020-09-17
Update rubygem-foreman_discovery to 16.2.0 - RPM

lzap:

* Nightly
* 2.2


packages/plugins/rubygem-foreman_discovery/rubygem-foreman_discovery.spec
```diff
@@ -17,7 +17,7 @@
 
 Summary:    MaaS Discovery Plugin for Foreman
 Name:       %{?scl_prefix}rubygem-%{gem_name}
-Version:    %{mainver}
+Version:    16.2.0
```
evgeni, line 20: Can you please either update the `mainver` macro, or get rid of it (and `release` and `prever`)? ;-)

evgeni (changes requested):

packages/plugins/rubygem-foreman_discovery/rubygem-foreman_discovery.spec
lzap: For the record, I used the script. Sure I can do that.

lzap:
Had to change source branch: https://github.com/theforeman/foreman-packaging/pull/5820

PR #5811, opened 2020-09-17
Fix license for method_source

pcreech:


evgeni (approved):

ekohl (approved):
Looks like this was fixed in 0.9.1 (https://github.com/banister/method_source/commit/71c04053970e5803e5cb8161e13a7b3fc3414d11#diff-414dba437e0d5e68677b1789b9edc4dc) but the spec wasn't regenerated so it didn't benefit from this.

PR #5812, opened 2020-09-17
Use -fno-strict-aliasing for nio4r

pcreech:

### DIFF
--- a/packages/foreman/rubygem-nio4r/rubygem-nio4r.spec
+++ b/packages/foreman/rubygem-nio4r/rubygem-nio4r.spec
@@ -7,7 +7,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 2.5.2
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: New IO for Ruby
 Group: Development/Languages
 License: MIT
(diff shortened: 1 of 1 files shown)


packages/foreman/rubygem-nio4r/rubygem-nio4r.spec
```diff
@@ -52,6 +52,9 @@ gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
 %{?scl:EOF}
 
 %build
+# Use -fno-strict-alisaing, due to various "warning: dereferencing type-punned pointer will break strict-aliasing rules"
+export CONFIGURE_ARGS="--with-cflags='%{optflags} -fno-strict-aliasing'"
```
ekohl, line 56: This is what Fedora uses:
```suggestion
# Update %%optflags used in %%gem_install to avoid strict-aliasing warnings.
# https://github.com/socketry/nio4r/pull/130
# http://pkgs.fedoraproject.org/cgit/rpms/rubygems.git/tree/macros.rubygems
%global optflags %{?optflags} -fno-strict-aliasing
```

ekohl (commented):
https://github.com/socketry/nio4r/blob/40de869376178b113de3bc3452002d0f7f314650/CHANGES.md#210-2017-05-28 suggests that this should be done already since 2.1.0. However, it's still present in the Fedora spec as well so that surprises me a bit.

packages/foreman/rubygem-nio4r/rubygem-nio4r.spec
pcreech: I'll test fedora's syntax in a scratch build real quick.

My fix was based off of:
https://fedoraproject.org/wiki/Archive:PackagingDrafts/RubyGem_with_C_code#.5BSOLVED.5D_To_pass_CFLAGS_correctly

And pre-existing examples in packaging.

pcreech: confirmed fedora syntax worked as intended, updated to that syntax.

ekohl (approved):

PR #5835, opened 2020-09-22
don't try to build scl-only gems for non-scl

evgeni:
2c624dcdbe4f91b7a59142db1d4b07043f184eed removed those from the comps
and the package_manifest, but they were still in tito.props


evgeni:
updated to re-add them to the manifest in the right place too.

ekohl (approved):

PR #5842, opened 2020-09-23
Drop x-editable-rails

ezr-ondrej:
https://github.com/theforeman/foreman/pull/7775 is dropping the necessity for having this gem.


tbrisker:
@ezr-ondrej I think you will first need to remove the package from the foreman requires block and only then drop it from packaging completely, but @theforeman/packaging will know better

evgeni:
Nah, that can happen in the same PR, you just need to properly add a changelog etc to foreman so that Jenkins builds it :)

Or just take https://github.com/theforeman/foreman-packaging/pull/5848 :D

ezr-ondrej:
So just like this? :)

evgeni:
that's also a solution, but will need to be merged after #5848 ;)

ezr-ondrej:
> that's also a solution, but will need to be merged after #5848 ;)

I guess that shouldn't be a problem, right? There is no rush for this.

evgeni (approved):

evgeni:
removed the pkg from koji nightly

PR #5845, opened 2020-09-23
ignore eslint-plugin-spellcheck

laviro:

### DIFF
--- a/update-requirements
+++ b/update-requirements
@@ -23,6 +23,7 @@ EXCLUDE_NPM_PACKAGES = [
     'babel-plugin-dynamic-import-node',
     'coveralls',
     'cross-env',
+    'eslint-plugin-spellcheck',
     'highlight.js',
     'jest',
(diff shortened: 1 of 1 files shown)


update-requirements
```diff
@@ -23,6 +23,7 @@ EXCLUDE_NPM_PACKAGES = [
     'babel-plugin-dynamic-import-node',
     'coveralls',
     'cross-env',
+    'eslint-plugin-spellcheck',
```
ekohl, line 26: This isn't really needed since `eslint` is in `EXCLUDE_NPM_PREFIXES`. That was easier than keeping track of all the plugins.

laviro: cc @evgeni

evgeni: oh, then I shut up what I said in the foreman PR :D

evgeni:
not needed as pointed out by ewoud

PR #5848, opened 2020-09-24
update foreman dependencies

evgeni:


evgeni:
`LoadError: cannot load such file -- x/editable/rails` seems it's not been source built yet?

evgeni:
[test rpm]

ehelms (approved):

PR #5862, opened 2020-10-01
Update nodejs-patternfly-react to 2.39.17

xprazak2:


xprazak2:
I have no idea why the build fails as it does on a version bump...

tbrisker:
Do we actually use this package directly? isn't it bundled inside foreman-js vendor bundle?

xprazak2:
Right, this should be consumed via foreman-js. Thanks, closing.

PR #5867, opened 2020-10-05
Update rubygem-katello to 3.16.1.2

ianballou:
I noticed the 3.16.1.1 version bump didn't contain any other gem version updates in the spec file, should those have stayed the same?  They match what I have in katello.gemspec now.


ianballou:
Hm there are errors, maybe I should keep the build packages the same as they were for 3.16.1.1.  katello.gemspec didn't change at all between 3.16.1.1 and 3.16.1.2

zjhuntin (approved):
Thanks Ian, this looks good!

Generally when updating requirements I let scripts handle that.

The particular script for handling those is: https://github.com/theforeman/foreman-packaging/blob/rpm/2.1/update-requirements

You can run `./update-requirements /path/to/your/katello/package.json ./path/to/rubygem-katello/rubygem-katello.spec` to see it run for yourself. For Katello it does add vendor-core (which it should not), but the rest looks good to go.

PR #5870, opened 2020-10-07
Update pulp_2to3_migration_client to 0.4.0

zjhuntin:


evgeni:
Updating this will break Katello:
https://github.com/theforeman/foreman-packaging/blob/4e874616b7d79de1b97be66f48269030e60e2c26/packages/katello/rubygem-katello/rubygem-katello.spec#L61-L62

https://github.com/Katello/katello/blob/master/katello.gemspec#L58 needs an update

@ianballou can you take care of that

ianballou:
@evgeni yep, I am working on updating the bindings.

ianballou:
PR for Katello update is here: https://github.com/Katello/katello/pull/8982

Currently blocked on a bug with  pulp_2to3_migration_client 0.4.0.

ianballou:
Since 0.4.0 had a bug and 0.4.1 is out now, can we update this PR for that?

ianballou:
Actually let's hold off on 0.4.1, I might be dealing with another pulp 3 bug.

ianballou:
Looks like we'll need to wait for 0.4.2, now blocked on https://pulp.plan.io/issues/7693

(it appears to be the last issue)

evgeni:
> Looks like we'll need to wait for 0.4.2, now blocked on https://pulp.plan.io/issues/7693
> 
> (it appears to be the last issue)

Isn't that an issue with the unit file and thus the installer? cc @ekohl @theforeman/installer

ianballou:
> > Looks like we'll need to wait for 0.4.2, now blocked on https://pulp.plan.io/issues/7693
> > (it appears to be the last issue)
> 
> Isn't that an issue with the unit file and thus the installer? cc @ekohl @theforeman/installer

It's a bug with the migration plugin because the Pulp team doesn't want the pulpcore workers to be using the service file's `WorkingDirectory`.  In Pulp, they have their own way to recall the worker's working directory.

That being said, I'm not sure if the information above changes what we should be putting as the pulpcore worker's systemd working directory.

zjhuntin:
Closing till we get 0.4.2

ianballou:
We should be good to go now: https://pypi.org/project/pulp-2to3-migration-client/0.5.0/

I'm running the tests for Katello to update the bindings gems.  Should be merged soon.

PR #5875, opened 2020-10-15
Update Pulp 3 bindings for 3.7

ianballou:


ianballou:
Need to update other pulp 3 bindings too.

jlsherrill (approved):
ACK, i think we need an update for rubygem-katello too ( can be another pr)

PR #5891, opened 2020-10-20
Update foreman-release to 2.2.0

zjhuntin:


tbrisker (approved):
repoclosure is expected

PR #5893, opened 2020-10-21
Update rubygem-pulp_2to3_migration_client to 0.5.0

jlsherrill:


chris1984 (approved):

chris1984:
@tbrisker this is needed for 3.17

packages/katello/rubygem-pulp_2to3_migration_client/rubygem-pulp_2to3_migration_client.spec
```diff
@@ -21,13 +21,14 @@ Source0: https://rubygems.org/gems/%{gem_name}-%{version}%{?prerelease}.gem
 # start specfile generated dependencies
 Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}ruby >= 1.9
-Requires: %{?scl_prefix_ruby}ruby(rubygems) > 1.3.1
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
 Requires: %{?scl_prefix}rubygem(faraday) >= 0.14.0
+Requires: %{?scl_prefix_ruby}rubygem(json) >= 2.1.0
```
tbrisker, line 26: This seems redundant with the next line, was the specfile generated by obal or manually? (i.e. is this a mistake in the gemspec?)

jlsherrill: it was auto-generated using 'bump_rpm.sh'.  Looks like this is a quirk of the spec file: https://rubygems.org/gems/pulp_2to3_migration_client

tbrisker: huh, odd. I tried to find where the gemspec is but seems like it's generated in some automated process as part of the version release. i couldn't find where it got these requirements from.

tbrisker (approved):
spec is odd but looks like it's true to the gemspec.

packages/katello/rubygem-pulp_2to3_migration_client/rubygem-pulp_2to3_migration_client.spec
evgeni: Oh yeah, if the gem has both >= and ~> in the spec, this looks like what you'd get.

Not nice, but also not hurting.

evgeni (approved):
Will need the koji 💃, so not merging right now.

PR #5899, opened 2020-10-22
fix katello requires on pulp_2to3_migration_client

evgeni:


evgeni:
repoclosure is failing due to the package that's currently in the repo, and we know that's broken ;)

ekohl (approved):

PR #5900, opened 2020-10-22
Do not start foreman.service or foreman.socket on package install

ehelms:
Warning: I do not know exactly if I am doing this right to start, or how to test this without trying to get a test build from the PR process like I would get a scratch build with RPM.

### DIFF
--- a/debian/bionic/foreman/rules
+++ b/debian/bionic/foreman/rules
@@ -33,9 +33,16 @@ build:
 	dh $@ --with=systemd
 
 override_dh_installinit:
-	dh_installinit
+	dh_installinit --only-scripts
+
+override_dh_systemd_start:
+	dh_systemd_start --no-start --name=foreman.service --package=foreman-service
+	dh_systemd_start --no-start --name=foreman.socket --package=foreman-service
+	dh_systemd_start --no-start --name=dynflow-sidekiq@orchestrator --package=foreman-dynflow-sidekiq
+	dh_systemd_start --no-start --name=dynflow-sidekiq@worker --package=foreman-dynflow-sidekiq
 
 override_dh_systemd_enable:
-	dh_systemd_enable --package=foreman-service
+	dh_systemd_enable --name=foreman.service --package=foreman-service
+	dh_systemd_enable --name=foreman.socket --package=foreman-service
 	dh_systemd_enable --name=dynflow-sidekiq@orchestrator --package=foreman-dynflow-sidekiq
 	dh_systemd_enable --name=dynflow-sidekiq@worker --package=foreman-dynflow-sidekiq
--- a/debian/buster/foreman/rules
+++ b/debian/buster/foreman/rules
@@ -30,9 +30,16 @@ build:
 	dh $@ --with=systemd
 
 override_dh_installinit:
-	dh_installinit
+	dh_installinit --only-scripts
+
+override_dh_systemd_start:
+	dh_systemd_start --no-start --name=foreman.service --package=foreman-service
+	dh_systemd_start --no-start --name=foreman.socket --package=foreman-service
+	dh_systemd_start --no-start --name=dynflow-sidekiq@orchestrator --package=foreman-dynflow-sidekiq
+	dh_systemd_start --no-start --name=dynflow-sidekiq@worker --package=foreman-dynflow-sidekiq
 
 override_dh_systemd_enable:
-	dh_systemd_enable --package=foreman-service
+	dh_systemd_enable --name=foreman.service --package=foreman-service
+	dh_systemd_enable --name=foreman.socket --package=foreman-service
 	dh_systemd_enable --name=dynflow-sidekiq@orchestrator --package=foreman-dynflow-sidekiq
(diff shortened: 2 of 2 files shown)


ekohl:
After the build has finished, you add your repo from https://stagingdeb.theforeman.org/ to the system. This is on top of the regular nightly builds. The README on top of it should be fairly self-explanatory, but they live in `/etc/apt/sources.list`. Note you need to run `apt update` after changing that file since (unlike yum) apt doesn't auto refresh repo metadata.

debian/bionic/foreman/rules
```diff
@@ -33,9 +33,10 @@ build:
 	dh $@ --with=systemd
 
 override_dh_installinit:
-	dh_installinit
+	dh_installinit --only-scripts
```
ehelms, line 36: I did this to stop foreman.service from being created during `apt-get install foreman`, however, I wondered more broadly if this should be dropped all together in favor of systemd?

debian/bionic/foreman/rules
```diff
@@ -33,9 +33,10 @@ build:
...
 
 override_dh_systemd_enable:
-	dh_systemd_enable --package=foreman-service
+	dh_systemd_enable --name=foreman.service --package=foreman-service
+	dh_systemd_enable --name=foreman.socket --package=foreman-service
```
ehelms, line 40: So far in testing, this has the effect of not creating these services during `apt-get install foreman` and requires a user to install `foreman-service` which is in line with RPM. However, this still seems to start the services so working to figure out how to prevent that (going to try the --no-start).

debian/bionic/foreman/rules
evgeni: Well, technically, Debian still supports non-systemd systems, so if that package would go into Debian proper, that'd be a bug.

Given we don't aim for that, we could say we only support systemd systems.

debian/bionic/foreman/rules
```diff
@@ -33,9 +33,14 @@ build:
 	dh $@ --with=systemd
 
 override_dh_installinit:
-	dh_installinit
+	dh_installinit --only-scripts
 
 override_dh_systemd_enable:
-	dh_systemd_enable --package=foreman-service
+	dh_systemd_enable --name=foreman.service --package=foreman-service
+	dh_systemd_start --no-start --name=foreman.service --package=foreman-service
```
evgeni, line 40: these need to go into a `override_dh_systemd_start` target, otherwise you end up calling 'em twice.

debian/bionic/foreman/rules
ekohl: I think that may have been the true bug. We did do it right for the dynflow part, but omitted it for the actual service.

ehelms:
Working now, after `apt-get install foreman-service`:

```
root@pipeline-foreman-server-nightly-ubuntu1804:/home/vagrant# systemctl status foreman.service
● foreman.service - Foreman
   Loaded: loaded (/lib/systemd/system/foreman.service; enabled; vendor preset: enabled)
   Active: inactive (dead)
     Docs: https://theforeman.org
root@pipeline-foreman-server-nightly-ubuntu1804:/home/vagrant# systemctl status foreman.socket
● foreman.socket - Foreman HTTP Server Accept Sockets
   Loaded: loaded (/lib/systemd/system/foreman.socket; enabled; vendor preset: enabled)
   Active: inactive (dead)
   Listen: 0.0.0.0:3000 (Stream)
```

ekohl:
Note you can also use `systemctl status foreman.service foreman.socket` and get the output from both at once.

evgeni:
Is this PR then the right time to drop https://github.com/theforeman/foreman-packaging/blob/deb/develop/debian/buster/foreman/foreman.init (and https://github.com/theforeman/foreman-packaging/blob/deb/develop/debian/buster/foreman/foreman.default)?

would require some further cleanup like we did for dynflow:
https://github.com/theforeman/foreman-packaging/blob/a1da82fdfaeebe72a1879eb52be65d9b07945db5/debian/buster/foreman/foreman.preinst#L15-L18

but at least it would mean we don't actually ship an init script that we don't support at all anymore

evgeni:
packaging wise this looks correct to me

ehelms:
Can I do the drop of initscript separately to reduce potential issues? I can do a PR for it now. I'd like to, if this change is approved, get this in so I can continue finishing up https://github.com/theforeman/puppet-foreman/pull/883

evgeni:
Oh yeah, certainly!

debian/bionic/foreman/rules
evgeni: FWIW, `dh_installinit` is still responsible for also *copying* the `foreman.service` files etc, so this line can't be dropped completely yet. But at some point we will be able to. (Needs to bump some other places to make `dh_systemd_*` take over the copying)

evgeni (approved):
from a debian packaging perspective, this looks correct so ACK

I assume you tested this that it actually does what you need in your setup.

ehelms:
To the best of my ability I tested this, this at least brings this into alignment with RPM which is a win

PR #5910, opened 2020-10-25

ekohl:


tbrisker (approved):
repoclosure is failing, for some reason it still tries with foreman-gce-2.3.0-0.6.develop.20201023144446git5e3bea2.el7.noarch from el7-foreman-nightly
 - I guess that's expected until the new package is built?

ekohl:
Yes I think so. Merging to proceed.

PR #5913, opened 2020-10-26
Update Rails to 6.0.3.4

evgeni:


tbrisker (approved):
Changes are just CVE fixes, nothing should break. Thanks @evgeni !

PR #5930, opened 2020-11-02
Update deb/develop to 2.4

pcreech:


tbrisker:
[test packaging]

tbrisker:
This is failing due to wrong source branch naming (why do we have that check in the first place??). I'll open a new PR with the correct naming

tbrisker:
closing for #5934

ekohl:
> This is failing due to wrong source branch naming (why do we have that check in the first place??)

It's not really a check but rather logic that attempts to figure out whether it's a nightly build. It's a bit complicated, but it could be modified to use the target branch (as it should). I made an attempt once, but that broke things so we reverted it. However, there's no reason it couldn't be done.

PR #5939, opened 2020-11-03
Remove el5 from package_manifest.yaml

pcreech:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -722,8 +722,6 @@ foreman_client_packages:
         dist: '.el7'
       - name: "foreman-client-{{ foreman_version }}-rhel6"
         dist: '.el6'
-      - name: "foreman-client-{{ foreman_version }}-rhel5"
-        dist: '.el5'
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -722,8 +722,6 @@ foreman_client_packages:
         dist: '.el7'
       - name: "foreman-client-{{ foreman_version }}-rhel6"
         dist: '.el6'
-      - name: "foreman-client-{{ foreman_version }}-rhel5"
-        dist: '.el5'
       - name: "foreman-client-{{ foreman_version }}-fedora29"
         dist: '.fc29'
```
ekohl, line 726: We also dropped Fedora 29 from 2.2. I think it's safe to also drop it in 2.3 since it's EOL.

ekohl (approved):

package_manifest.yaml
pcreech: Will go ahead and drop fc29, I think it just bit me on 2.3 rc1

PR #5941, opened 2020-11-03
Remove fedora 29 from develop

pcreech:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -757,8 +757,6 @@ foreman_client_packages:
         dist: '.el7'
       - name: "foreman-client-{{ foreman_version }}-rhel6"
         dist: '.el6'
-      - name: "foreman-client-{{ foreman_version }}-fedora29"
-        dist: '.fc29'
     repoclosure_target_repos:
       el8:
         - "el8-foreman-client-{{ foreman_version }}"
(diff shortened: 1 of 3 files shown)


ekohl (commented):
In the past we've said we want to keep Fedora supported for client tools and kept it in develop so we can easily update it to Fedora latest (33 at the time of writing). Thoughts on doing that instead?

pcreech:
So, based on the comment here, https://github.com/theforeman/foreman-packaging/pull/5939#discussion_r516762778 it sounded like fc29 is at least dropped in 2.2+.  Since this has bitten me for 2.3, i figured it was at least best to go ahead and remove this so it doesn't cause more pain in the future at time of branching.

I'm not against updating fedora build tags.  The question here is do we want to hold up removing f29 from develop when we've already dropped it in releases, or do we need to add in steps to remove fedora 29 in the branching docs?

ekohl:
With 2.2 Fedora 29 was already EOL so we decided not to ship it there, but also said we should update it in develop. That never happened which is why it's a problem now. It would be my preference to update the tags to Fedora 33 before 2.4 branches.

pcreech:
@ekohl While looking into some things fedora 33 related, it looks like they've changed a few things around with policies w/r to modularity an a manner that would be benificial.  in January 2021 we should be able to add fedora 33 as external repos and build out from there

ekohl:
Sounds good. Do you agree it's easier to keep f29 in until then and replace it when we get to it?

pcreech:
Given timeframes, and likeleyhood we will need to branch before we add f33 is low, a "removal, then add later" is moot.  Not sure yet if I'll close this PR or modify it.

pcreech:
This pr is rather outdated.  closing

PR #5954, opened 2020-11-05
Fixes #27239: Add timestamps to cron actions in Foreman

ehelms:


evgeni:
If only EL would ship [moreutils](http://joeyh.name/code/moreutils/)…

This needs a rebase and I wonder if we should add an explicit dependency on `gawk`?

ehelms:
I lost track of this one but I have rebased and addressed the explicit requires on gawk request.

evgeni (approved):

PR #5959, opened 2020-11-06
2.0 salt closure

evgeni:


evgeni:
yes, this is 2.0, yes that's unsupported, but I wanted to do a final plugin publish round, before closing that repo

ekohl (approved):

PR #5964, opened 2020-11-09
obsolete x-editable-rails

evgeni:


ekohl (approved):
I guess this also needs a cherry pick to rpm/2.3

evgeni:
> I guess this also needs a cherry pick to rpm/2.3

Yes indeed.

PR #5303, opened 2020-05-27
Update discovery to 16.1

lzap:
For plugin updates, please indicate which repos this should be built into:

* Nightly
* 2.1

PR #5977, opened 2020-11-10
Update bump_rpm to handle global version macro

lzap:
This is not a package, not sure what the process is.


ekohl (changes requested):
:-1: discovery is the only one. Last time I suggested to get rid of it, but this hasn't been implemented. I don't see a good reason to keep it.

lzap:
I don't understand, I was asked by @evgeni to keep the macro. What's the correct approach then?

evgeni:
No, I asked you to either use it properly or get rid of it (preferred):
https://github.com/theforeman/foreman-packaging/pull/5809#discussion_r490765192
https://github.com/theforeman/foreman-packaging/pull/5303#issuecomment-635204901

lzap:
I falled into thinking the macro is something we want. I can remove it easily.

PR #5980, opened 2020-11-11
Add nodejs-optimize-css-assets-webpack-plugin package

amirfefer:


comps/comps-foreman-rhel7.xml
```diff
@@ -34,6 +34,7 @@
       <packagereq type="default">foreman-service</packagereq>
       <packagereq type="default">foreman-telemetry</packagereq>
       <packagereq type="default">foreman-vmware</packagereq>
+      <packagereq type="default">nodejs-optimize-css-assets-webpack-plugin</packagereq>
```
evgeni, line 37:
```suggestion
      <packagereq type="default">tfm-nodejs-optimize-css-assets-webpack-plugin</packagereq>
```

and then this needs proper sorting :)

evgeni:
could you also please omit the ordering changes around google-cloud-env?

comps/comps-foreman-rhel7.xml
evgeni: and `nodejs-optimize-css-assets-webpack-plugin` (without the `tfm-` prefix) needs adding to `comps-foreman-el8.xml` (the script currently fails to do that for two comps files :()

rel-eng/tito.props
```diff
@@ -89,6 +89,7 @@ whitelist = foreman
   nodejs-node-gyp
   nodejs-node-sass
   nodejs-object-assign
+  nodejs-optimize-css-assets-webpack-plugin
```
evgeni, line 92: same needs to be added to `foreman-nightly-el8`, our scripts fail to do so :(

evgeni (changes requested):
comments inline

amirfefer:
Thanks @evgeni !
Updated

amirfefer:
Rebased

PR #5981, opened 2020-11-12
Let cockpit subpackage depend on exact version of the main package

evgeni:

### DIFF
--- a/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
+++ b/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
@@ -11,7 +11,7 @@
 Summary:    Plugin that brings remote execution capabilities to Foreman
 Name:       %{?scl_prefix}rubygem-%{gem_name}
 Version:    4.1.0
-Release:    1%{?foremandist}%{?dist}
+Release:    2%{?foremandist}%{?dist}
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -60,7 +60,8 @@ management functionality with remote management functionality.
 
 %package cockpit
 BuildArch:  noarch
-Requires:   cockpit
+Requires: cockpit
+Requires: %{?scl_prefix}%{pkg_name} = %{version}-%{release}
```
ekohl, line 64: Isn't `%name` sufficient?

evgeni: honestly? I just copied what's in the `doc` package ;)

evgeni: and the template:

https://github.com/theforeman/foreman-packaging/blob/8363a28d884d5ff4361bb9bc3e1eaf6d8729cfc4/gem2rpm/scl.spec.erb#L73

ekohl (approved):

pcreech (approved):
ACK

PR #5990, opened 2020-11-17
foreman_datacenter 2.2.0

Zenya:


evgeni:
can you add a `DATABASE_URL=nulldb://nohost` to the `assets:precompile` line so it becomes:

```
/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production DATABASE_URL=nulldb://nohost
```

evgeni (approved):

PR #5995, opened 2020-11-17
Fixes #31324 - Katello 3.17 should use Pulpcore 3.7

ianballou:
Katello 3.17 should use Pulpcore 3.7.


chris1984 (approved):

pcreech (approved):
ACK

evgeni:
Is there a reason this was done without a changelog entry and revision bump? This results in no build being made.

ianballou:
@evgeni that was a mistake. Here's the PR that fixes it https://github.com/theforeman/foreman-packaging/pull/6013

PR #5999, opened 2020-11-18
Update smart_proxy_remote_execution_ssh to 0.3.1

adamruzicka:
(cherry picked from commit b63a8b089a8815d3b9ccfe66b92e8e1232069b6f)


evgeni:
hum, I think we forgot that in nightly too:

https://github.com/theforeman/foreman-packaging/blob/c043d4377168849757e722a0750107f6adb8f1c2/plugins/smart_proxy_remote_execution_ssh/control#L15

needs an update

adamruzicka:
That's being addressed in https://github.com/theforeman/foreman-packaging/pull/6003 . Will CP it to this PR once it goes it

evgeni (approved):

PR #6013, opened 2020-11-18
Refs #31324 - Katello 3.17 should use Pulpcore 3.7

ianballou:
[This PR](https://github.com/theforeman/foreman-packaging/pull/5995) was missing the changelog entry and release version bump.

### DIFF
--- a/packages/katello/katello-repos/katello-repos.spec
+++ b/packages/katello/katello-repos/katello-repos.spec
@@ -7,7 +7,7 @@
 %define repo_dir %{_sysconfdir}/yum.repos.d
 %define repo_dist %{dist}
 
-%global release 1
+%global release 2
 
 Name:           katello-repos
 Version:        3.17.0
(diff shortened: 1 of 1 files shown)


evgeni:
Thanks! Should we push that out immediately, or wait until the next RC?

ianballou:
@evgeni I think we should build it as soon we we're able since 3.17.0 is already out with the older Pulpcore. We figured we'd hurry up 3.17.1 so that people can upgrade and start using the new release of katello-repos.

evgeni (approved):

evgeni:
@ianballou does the user has to do anything when the packages get updated? I'd bet they need to run migrations etc.

ianballou:
@evgeni we were thinking of having it be available for 3.17.1 since people should be running the foreman-installer. That should run the Pulp 3 migrations, if there are any.

evgeni:
Ah, okay, then I'll just let the build where it is until we ship .1

PR #6019, opened 2020-11-20
Add logging-journald dependency to dynlfow-core

lzap:
Dynflow rubygem does not ship with this dependency because by default it uses a file logger. But for CentOS production deployment we configure with journald by default for enahanced logging capabilities. This fails to start and breaks installation.

* Nightly
* 2.3

### DIFF
--- a/packages/plugins/rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
+++ b/packages/plugins/rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
@@ -15,7 +15,7 @@
 Summary: Core Smart Proxy Dynflow Service
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.3.1
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Group: Development/Libraries
 License: GPLv3
 URL: https://github.com/theforeman/smart_proxy_dynflow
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-smart_proxy_dynflow_core/rubygem-smart_proxy_dynflow_core.spec
```diff
@@ -33,6 +33,7 @@ Requires: %{?scl_prefix}rubygem(sinatra)
 Requires: %{?scl_prefix}rubygem(rack)
 Requires: %{?scl_prefix}rubygem(sqlite3)
 Requires: %{?scl_prefix}rubygem(logging)
+Requires: %{?scl_prefix}rubygem(logging-journald)
```
evgeni, line 36:
```suggestion
Requires: %{?scl_prefix}rubygem(logging-journald) >= 2.0
Requires: %{?scl_prefix}rubygem(logging-journald) < 3.0
```

See https://github.com/theforeman/smart_proxy_dynflow/blob/60eadab42a18a2ec5387c1bb85e1a70e270e077c/Gemfile#L18

lzap:
Rebased.

evgeni (approved):

evgeni:
After IRC discussion: the correct way would be to install the `foreman(-proxy)?-journald` package when one configures the journald logging, which this RPM does not (by default).

lzap:
To add more context, foreman and smart proxy has bundler groups. This allows running them without the dependency, to do that for dynflow core we'd need to implement the same behavior.

lzap:
Here is a patch we need in dynflow first, then I will rebase this PR to create the subpackage and we need someone to do the installer change: https://github.com/theforeman/smart_proxy_dynflow/pull/75 @adamruzicka

evgeni:
obsoleted by #6024

PR #6035, opened 2020-11-23
Update Katello @theforeman/builder dependency

evgeni:
see https://github.com/Katello/katello/pull/9044

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -9,7 +9,7 @@
 %global prereleasesource pre.master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global mainver 4.0.0
-%global release 1
+%global release 2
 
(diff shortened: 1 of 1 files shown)


tbrisker (approved):

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -235,6 +233,9 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/webpack
 
 %changelog
+* Mon Nov 23 2020 Evgeni Golov - 4.0.0-0.2.pre.master
```
tbrisker, line 236: not sure why but linter seems unhappy about this

tbrisker: ignore me, you were too fast again

evgeni: not anymore ;)

PR #6038, opened 2020-11-24
add python-qpid

evgeni:

### DIFF
--- a/comps/comps-katello-server-rhel7.xml
+++ b/comps/comps-katello-server-rhel7.xml
@@ -23,6 +23,7 @@
       <packagereq type="default">katello-utils</packagereq>
       <packagereq type="default">python-gofer</packagereq>
       <packagereq type="default">python-gofer-qpid</packagereq>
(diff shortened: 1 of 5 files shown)


evgeni:
the spec is taken straight from EPEL, the only change was the Source0 url, so it points at an actual tarball and not the download page

comps/comps-katello-server-rhel7.xml
```diff
@@ -23,6 +23,7 @@
       <packagereq type="default">katello-utils</packagereq>
       <packagereq type="default">python-gofer</packagereq>
       <packagereq type="default">python-gofer-qpid</packagereq>
+      <packagereq type="default">python-qpid</packagereq>
```
ekohl, line 26: Isn't it `python2-qpid`?

evgeni: it is, you're right.

ekohl (approved):

PR #6043, opened 2020-11-25
use qpid copr instead of building it ourself

evgeni:


ekohl (approved):
I suppose this needs a similar change for forklift.

PR #6044, opened 2020-11-25
add qpid copr repo

evgeni:
this contains the python2-qpid that got removed from EPEL

(cherry picked from commit 49694ccac9bc3bf8fc5a70d4d2eff748dd04c446)

### DIFF
--- a/packages/katello/katello-repos/katello.repo
+++ b/packages/katello/katello-repos/katello.repo
@@ -16,6 +16,13 @@ enabled=@PULP_ENABLED@
 gpgcheck=@PULP_GPG_CHECK@
 module_hotfixes=1
 
+[qpid]
+name=Copr repo for qpid owned by @qpid
+baseurl=https://download.copr.fedorainfracloud.org/results/@qpid/qpid/epel-7-$basearch/
+gpgcheck=@REPO_GPGCHECK@
+gpgkey=https://download.copr.fedorainfracloud.org/results/@qpid/qpid/pubkey.gpg
+enabled=@PULP_ENABLED@
+
 # Candlepin RPMs as supported by Katello - This is a coordinated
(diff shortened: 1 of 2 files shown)


packages/katello/katello-repos/katello.repo
```diff
@@ -16,6 +16,13 @@ enabled=@PULP_ENABLED@
 gpgcheck=@PULP_GPG_CHECK@
 module_hotfixes=1
 
+[qpid]
+name=Copr repo for qpid owned by @qpid
+baseurl=https://download.copr.fedorainfracloud.org/results/@qpid/qpid/epel-7-$basearch/
+gpgcheck=@REPO_GPGCHECK@
+gpgkey=https://download.copr.fedorainfracloud.org/results/@qpid/qpid/pubkey.gpg
+enabled=@PULP_ENABLED@
```
ekohl, line 24: Should this be conditional on EL7 so we don't break EL8?

ekohl (changes requested):

packages/katello/katello-repos/katello.repo
evgeni: it is -- `@PULP_ENDABLED@` is only true on EL7:

https://github.com/theforeman/foreman-packaging/blob/49694ccac9bc3bf8fc5a70d4d2eff748dd04c446/packages/katello/katello-repos/katello-repos.spec#L4-L8

https://github.com/theforeman/foreman-packaging/blob/49694ccac9bc3bf8fc5a70d4d2eff748dd04c446/packages/katello/katello-repos/katello-repos.spec#L71

ekohl: Good point.

ekohl (approved):

PR #6046, opened 2020-11-26
Bump to ACD 0.3.0

sbernhard:


evgeni:
You need to update the "@theforeman/builder" dependency, as we now have v6 in develop. For Katello we just dropped the upper bound for now.

sbernhard:
Thanks @evgeni for your help!

evgeni (approved):

evgeni:
thanks @sbernhard!
when picking back to 2.3 and older: don't bump the builder dependency there :)

PR #6049, opened 2020-11-26
Fix mv-error if non-empty dir is in rex-ssh dir

m-bucher:
Not yet sure if this makes sense, but we had people experiencing the following error during upgrade of `tfm-rubygem-smart_proxy_remote_execution_ssh`:
```
mv: inter-device move failed: ‘/usr/share/foreman-proxy/.ssh’ to ‘/var/lib/foreman-proxy/ssh/.ssh’; unable to remove target: Directory not empty
error: %pre(tfm-rubygem-smart_proxy_remote_execution_ssh-0.3.0-4.fm2_1.el7.noarch) scriptlet failed, exit status 1
```
Reason was there was a non-empty `.ssh`-directory existing in `/var/lib/foreman-proxy/ssh`

### DIFF
--- a/packages/plugins/rubygem-smart_proxy_remote_execution_ssh/rubygem-smart_proxy_remote_execution_ssh.spec
+++ b/packages/plugins/rubygem-smart_proxy_remote_execution_ssh/rubygem-smart_proxy_remote_execution_ssh.spec
@@ -18,7 +18,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.3.1
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Summary: Ssh remote execution provider for Foreman Smart-Proxy
 Group: Applications/Internet
 License: GPLv3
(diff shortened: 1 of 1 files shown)


dgoetz:
By removing the directory data could potentially be lost. Would be testing if the target directory exists and if it does only copy the content be better?

m-bucher:
> By removing the directory data could potentially be lost.

Yes, but is this not the expected behaviour of the `mv` anyway?
My problem is that I do not know exactly what the expected behaviour is :thinking:

m-bucher:
So I think I wrapped my head around it now.
The problem is that the switchover from ssh-keys within `/usr/share/foreman-proxy/.ssh` to `/var/lib/foreman-proxy/ssh` did not work and we ended up with both directories instead of one being a symbolic-link to the other.
To fix this without loosing data I added an additional step to move the existing `ssh`-dir out of the way before running the `mv`.

packages/plugins/rubygem-smart_proxy_remote_execution_ssh/rubygem-smart_proxy_remote_execution_ssh.spec
```diff
@@ -91,6 +91,9 @@ gem build %{gem_name}.gemspec
 
 %pre
 if [ -d %{foreman_proxy_dir}/.ssh ] && [ ! -L %{foreman_proxy_dir}/.ssh ] ; then
+  if [ -d %{foreman_proxy_statedir}/ssh ] ; then
+    mv %{foreman_proxy_statedir}/ssh %{foreman_proxy_statedir}/ssh.save$(date '+%%y%%m%%d%%H%%M')
```
m-bucher, line 95: Ideally this will run only once, so technically adding the date to the backup-directory should not be necessary.
I hope I did escape the `%` in the `date`-command correctly :crossed_fingers:

ekohl:
This has been there for a while. Perhaps you should also figure out why it's there. Otherwise you risk getting a huge number of save directories. Perhaps some tool (Ansible?) that tries to create the directory even if it's already a symlink?

ehelms:
Given the age of this PR I am opting to close it. Please re-open if this is an issue you wish to persue.

PR #6050, opened 2020-11-26
Bump rubygem-foreman_rh_cloud to 3.0.14-1

laviro:


evgeni:
Why the 2 to 3 bump?

You probably need to update the theforeman/vendor (build)dependency.

laviro:
> Why the 2 to 3 bump?

Our plugins branch against foreman 2.1 will remain with v2 and our master branch is now updated to v3, no major changes.

I will look over the npm dependency changes in foreman/ foreman-js, thanks

evgeni (approved):

PR #6052, opened 2020-11-27
reexec systemd after policy changes to make socket labeling work

evgeni:

### DIFF
--- a/packages/foreman/foreman-selinux/foreman-selinux.spec
+++ b/packages/foreman/foreman-selinux/foreman-selinux.spec
@@ -22,7 +22,7 @@
 
 %define moduletype apps
 
-%global release 1
+%global release 2
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


evgeni:
not sure about the `&>/dev/null || :` part (copied partially from `systemd` itself) and whether the same should also happen in the proxy package (it's not needed right now, but may be in the future?)

dgoetz:
Looks good to me, also if I am not sure about what `daemon re-exec` does. (This is way I do not a review and approve)

The `&>/dev/null || :` is always good if failing is an option a package can live with and better than preventing installation because of a failing script. It will simply suppress messages and always exit with 0.

evgeni:
a bit background why this is needed can be found in https://community.theforeman.org/t/foreman-nightly-rpm-pipeline-836-failed/21418

TL;DR:
* we introduced a unix socket for foreman to listen on, and this is handled by systemd
* systemd (at least in el7/el8) will label that socket with the filecontext that was valid when systemd was started
* but we first define the filecontext when we install/update foreman-selinux
* thus systemd doesn't know the changed context and will miss-label the socket, leading to problems
* this restarts the daemon, making it reload the selinux policy and learn the new rules

dgoetz (approved):

dgoetz:
Thanks for the explanation. Going to merge it.

PR #6057, opened 2020-11-27
introduce a shared foreman-plugin.mk that can be used to build plugins

evgeni:

### DIFF
--- a/debian/bionic/foreman/foreman.install
+++ b/debian/bionic/foreman/foreman.install
@@ -5,6 +5,7 @@ config/settings.yaml                     etc/foreman
 config/database.yml                      etc/foreman
 script/foreman-rake                      usr/sbin
 script/foreman-tail                      usr/sbin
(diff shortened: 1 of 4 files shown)


evgeni:
let's try against 2.3 for now: https://github.com/theforeman/foreman-packaging/pull/6058

debian/bionic/foreman/foreman.install
```diff
@@ -5,6 +5,7 @@ config/settings.yaml                     etc/foreman
 config/database.yml                      etc/foreman
 script/foreman-rake                      usr/sbin
 script/foreman-tail                      usr/sbin
+foreman-plugin.mk                        usr/share/foreman
```
evgeni, line 8: this needs to be debian/…

evgeni:
https://github.com/theforeman/foreman-packaging/pull/6060

PR #6060, opened 2020-11-27
introduce a shared foreman-plugin.mk that can be used to build plugins

evgeni:


ehelms (approved):
Not as familiar with Debian, but looks good from a basic review.

PR #6063, opened 2020-11-30
don't use self-built yard

evgeni:


dependencies/buster/puppet-strings/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-puppet-strings (1.1.0-2) stable; urgency=low
```
ekohl, line 1: I don't see any change here (same for focal)

evgeni: that's correct, it was already using the "real" package there, but I wanted the versions to align.

evgeni: my plan is to get rid of most contents in the focal and bionic folders, unless the builds *really* differ from buster, but for that they need to match first ;)

ekohl: Fair enough.

ekohl (approved):

PR #6065, opened 2020-11-30
add redfish support for 2.3

evgeni:


evgeni:
[test packaging]

(redfish needs to be built *after* server-sent-events)

lzap (approved):
Great, all green. Thank you.

PR #6083, opened 2020-12-02
Update Puma to 5.1

ekohl:
Needed for https://github.com/theforeman/foreman/pull/8171


evgeni:
Foreman needs a versionbump to make jenkins happy ;)

ekohl:
At least lint now passes :)

evgeni:
But not much more than that. WTH is that Rdoc error?!

ehelms:
Fun times. You will need to include the patch file below into the RPM (add an ole Patch0 line and a `patch0 -p1` after the `%setup` call). See https://bugs.ruby-lang.org/issues/14617

Once we move to rh-ruby27, I believe we can drop this patch as the version of RDoc is newer (http://mirror.centos.org/centos/7/sclo/x86_64/rh/Packages/r/rh-ruby27-rubygem-rdoc-6.2.1-128.el7.noarch.rpm). We could try to patch this in Puma directly and wait for a release. We would just not have a really compelling reason to.

```
Refs: [HEAD], v5.1.0-1-g15196eb5
Author:     Eric D. Helms <ericdhelms@gmail.com>
AuthorDate: Mon Dec 14 14:52:36 2020 -0500
Commit:     Eric D. Helms <ericdhelms@gmail.com>
CommitDate: Mon Dec 14 14:52:36 2020 -0500

    Drop inline trailing comment due to RDoc bug https://bugs.ruby-lang.org/issues/14617
---
 lib/puma/control_cli.rb | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/lib/puma/control_cli.rb b/lib/puma/control_cli.rb
index df749ff6..4b015c6f 100644
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -171,7 +171,7 @@ module Puma
           OpenSSL::SSL::SSLSocket.new(
             TCPSocket.new(uri.host, uri.port),
             OpenSSL::SSL::SSLContext.new)
-            .tap { |ssl| ssl.sync_close = true }  # default is false
+            .tap { |ssl| ssl.sync_close = true }
             .tap(&:connect)
         when 'tcp'
           TCPSocket.new uri.host, uri.port
```

ekohl:
Thanks! I think it does make sense to at least submit it upstream, see if they're interested in the patch.

ekohl:
It's green now. I chose to use `%autosetup` instead. Perhaps we should use this in our base templates

ehelms:
For upgrades, I think something (probably puma?) should obsolete the puma-plugin-systemd package otherwise it will error out trying to yum update.

ehelms:
@ekohl could you rebase this PR and tackle that obsoletes?

ekohl:
Rebased with Obsoletes line. Also updated to Puma 5.1.1 while I was at it.

ehelms (approved):

PR #6088, opened 2020-12-03
Update rubygem-foreman_ansible to 6.0.1

xprazak2:


evgeni:
You'll need to allow newer `@theforeman/builder` versions, we have 6.x in develop now.
For most other plugins, we just dropped the upper bound and it was fine.

xprazak2:
I dropped the upper bound for `@theforeman/builder`.

evgeni (approved):

PR #6091, opened 2020-12-04
use vault.centos.org for el6 repoclosure -- CentOS 6 is done

evgeni:

### DIFF
--- a/repoclosure/yum.conf
+++ b/repoclosure/yum.conf
@@ -163,7 +163,7 @@ baseurl=https://vault.centos.org/6.10/extras/x86_64/
 
 [el6-epel]
 name=epel
-mirrorlist=http://mirrors.fedoraproject.org/mirrorlist?repo=epel-6&arch=x86_64
+baseurl=https://archives.fedoraproject.org/pub/archive/epel/6/x86_64/
 failovermethod=priority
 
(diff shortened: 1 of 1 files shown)


repoclosure/yum.conf
```diff
@@ -148,18 +148,18 @@ baseurl=http://koji.katello.org/releases/yum/katello-nightly/candlepin/el8/x86_6
...
-mirrorlist=http://mirrorlist.centos.org/?release=6&arch=x86_64&repo=updates
+baseurl=https://vault.centos.org/6.10/updates/x86_64/
 failovermethod=priority
 
 [el6-extras]
 name=extras
-mirrorlist=http://mirrorlist.centos.org/?release=6&arch=x86_64&repo=extras
+baseurl=https://vault.centos.org/6.10/extras/x86_64/
 
 [el6-epel]
```
ekohl, line 164: This is also empty now and should be removed or at least replaced with something that does work.

ekohl (changes requested):

repoclosure/yum.conf
evgeni: replaced (can't be removed as the openscap client depends on rubygem-json that's in epel)

ekohl (approved):

PR #6095, opened 2020-12-07
remove el5 remains

evgeni:


evgeni:
repoclosure errors unrelated

ekohl (commented):
Did you keep the EL5 bits in https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/client/katello-host-tools/katello-host-tools.spec on purpose?

evgeni:
> Did you keep the EL5 bits in https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/client/katello-host-tools/katello-host-tools.spec on purpose?

nope, I must admit I just grepped for `el5`, not things like `%{rhel} == 5` :)

ekohl (commented):
I think this can also be dropped:
https://github.com/theforeman/foreman-packaging/blob/a475b3131edda206b4de188813c391ca423633bf/packages/client/katello-host-tools/katello-host-tools.spec#L26

pcreech (approved):
ACK

PR #6126, opened 2020-12-10
Bump nodejs-node-sass to 4.14.1

zjhuntin:


evgeni:
@zjhuntin was this PR opened because Foreman didn't build on EL8? That was due to bad modules that we imported from EL8.3, I've fixed that now (also see https://github.com/theforeman/foreman-infra/pull/1533), so this PR might not be needed anymore?

PR #6134, opened 2020-12-11
Add wrapper-script to select python-exe for salt

m-bucher:
Since the python-version that needs to be used for `upload-salt-reports` does primarily depend on whether the python2 or python3 version of the Saltstack python-lib is installed, it might be better to have a wrapper script to dynamically decide, which python needs to be used.
The script prefers python3.

It might be better placed in https://github.com/theforeman/smart_proxy_salt repository instead of Adding it through the spec-file, but this was the fastest solution to test this :grin:


pcreech:
Closing due to inactivity.  If you feel it needs to be reopened, please reopen it!

PR #6137, opened 2020-12-14
Add rubygem-hammer_cli_foreman_scc_manager package

nadjaheitmann:
New hammer plugin for the foreman scc manager.


sbernhard:
[test rpm]

sbernhard (approved):

quba42:
Should we also add a dependency from the SCC manager package to this new Hammer CLI package?
(I think that is what other plugins are doing.)
It someone with more foreman-packaging experience than myself could confirm or deny, that would be great!

dgoetz:
No, there is no dependency required between the Foreman plugin and the Hammer plugin as you do not need the CLI on the server and on the other hand the CLI can be used from any client.

quba42:
@dgoetz 

> No, there is no dependency required between the Foreman plugin and the Hammer plugin as you do not need the CLI on the server and on the other hand the CLI can be used from any client.

This sounds reasonable to me, but then I don't understand why Katello has a dependency like this: https://github.com/theforeman/foreman-packaging/blob/c96ff37e0a734d5514ae1aa4ee70e026c4450e04/packages/katello/katello/katello.spec#L39

Do you know if this is just a matter of Katello _choosing_ to do things this way? (Or am I misunderstanding what that dependency is doing?)

ekohl:
It is indeed Katello choosing to. Note that katello is mostly a meta package. The actual plugin is rubygem-katello.

comps/comps-foreman-el8.xml
```diff
@@ -461,8 +461,8 @@
       <packagereq type="default">rubygem-hirb-unicode-steakknife-doc</packagereq>
       <packagereq type="default">rubygem-hocon-doc</packagereq>
       <packagereq type="default">rubygem-hoe-doc</packagereq>
-      <packagereq type="default">rubygem-httpclient-doc</packagereq>
```
evgeni, line 464: I think none of the changes in this file is actually relevant, mind reverting it?

comps/comps-foreman-plugins-el8.xml
```diff
@@ -195,9 +195,11 @@
       <packagereq type="default">rubygem-fast_gettext-doc</packagereq>
       <packagereq type="default">rubygem-fog-kubevirt-doc</packagereq>
       <packagereq type="default">rubygem-fog-proxmox-doc</packagereq>
+      <packagereq type="default">rubygem-foreman-tasks-core-doc</packagereq>
```
evgeni, line 198: this file is missing the addition of your new package (and it's -doc subpackage)

comps/comps-foreman-rhel7.xml
```diff
@@ -480,8 +480,8 @@
       <packagereq type="default">tfm-rubygem-hirb-unicode-steakknife-doc</packagereq>
       <packagereq type="default">tfm-rubygem-hocon-doc</packagereq>
       <packagereq type="default">tfm-rubygem-hoe-doc</packagereq>
-      <packagereq type="default">tfm-rubygem-httpclient-doc</packagereq>
```
evgeni, line 483: I think none of the changes in this file is actually relevant, mind reverting it?

rel-eng/tito.props
```diff
@@ -946,13 +946,14 @@ whitelist = foreman-discovery-image-service
   rubygem-hammer_cli_foreman_openscap
   rubygem-hammer_cli_foreman_remote_execution
   rubygem-hammer_cli_foreman_salt
+  rubygem-hammer_cli_foreman_scc_manager
```
evgeni, line 949: you need to add that to the el8 whitelist too

evgeni (changes requested):
Hi @nadjaheitmann, thanks for the PR and sorry for the longish delay in review!

I've added a few comments inline, please have a look and feel free to ping me if I was too short and you need help making the changes.

One thing that jumped at me, but I couldn't immediately say is wrong or right: SCC manager itself lives in the "katello" package group, as it depends on Katello, but you added the hammer bits to plain "plugins" (especially as it does not depend on `hammer_cli_katello`). I *think* it still would be more correct to put the hammer plugin also in the "katello" group (like we do with rubygem-hammer_cli_foreman_virt_who_configure for example). If you agree, you'd have to move a few things around (comps, tito, package_manifest) - I can show you if you want.

nadjaheitmann:
Hi @evgeni , thank you for your feedback, it is highly appreciated! 

What I did to add the package was running the following code, that is why it ended up in the plugins section, I guess:
'./add_gem_package.sh hammer_cli_foreman_scc_manager hammer_plugin'
I am still new to the topic, so it is difficult for me to judge where the hammer scc plugin actually belongs to. I am wondering why certain hammer plugins are in katello group and others are just "plain' plugins - especially given the explanation given by @dgoetz and given the `hammer_plugin` flag of the `add_gem_package` flag. 

If I move the package to Katello group, maybe I can rerun the script `add_gem_package` script with another type of plugin. Is there a preferred type? I can then fix the issues you pointed out on top of that.

evgeni:
the `add_gem_package` script doesn't know the special case of "hammer plugin that technically belongs to katello", so it can only serve as a start here with some manual changes afterwards.
it also doesn't know that we build packages for both: el7 and el8 today, so one ends up adding that by hand too :(

as to how things are split up today: we differentiate packaging in 3 types: foreman (core), foreman plugins and katello, mostly to allow releases of the different parts to happen at different times.
initially, the katello category was just that, katello (and it's dependencies), but over the time we gained plugins that themself require katello to be present, so it felt wrong to put them into the generic "plugin" category (they wouldn't be installable from there). scc manager is such a plugin, as it uses katello and its repository management functionality. and then the hammer plugins for those plugins also live in katello.

nadjaheitmann:
Thank you for the explanation! Let me try to move things and fix the issues and see how far I get.

ekohl:
> the `add_gem_package` script doesn't know the special case of "hammer plugin that technically belongs to katello", so it can only serve as a start here with some manual changes afterwards.

This can be done by adding a 4th parameter:
```
./add_gem_package.sh hammer_cli_foreman_scc_manager hammer_plugin katello
```

nadjaheitmann:
I have updated the PR. Actually, I had to call the script with five parameters, I chose `katello-nightly-rhel7` as `TITO-TAG`.
`./add_gem_package.sh GEM_NAME [TEMPLATE [TITO_TAG [PACKAGE_SUBDIR]]]`

@evgeni Do you mind checking whether everything is in the right location now and I am not missing any files?

evgeni (approved):
perfect! thanks again @nadjaheitmann!

PR #6143, opened 2020-12-15
Katello 3.18

evgeni:

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -3,8 +3,6 @@
 
 %global homedir %{_datarootdir}/%{name}
 %global confdir common
-%global prereleasesource rc2
-%global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global release 1
 
 Name:       katello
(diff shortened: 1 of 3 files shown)


evgeni:
@zjhuntin you forgot to bump katello and katello-repos in your last PR

also dropping the qpid workaround, now that epel is fixed

packages/katello/katello/katello.spec
```diff
@@ -180,6 +178,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Tue Dec 15 2020 Evgeni Golov - 3.18.0-1
+- Release katello 3.18.0
+
 * Thu Nov 19 2020 Evgeni Golov - 3.18.0-0.1.rc2
```
ekohl, line 184: This is a confusing changelog with 3 times `Release Katello 3.18.0`

ekohl (approved):

packages/katello/katello/katello.spec
evgeni: *whistle* we'd need to fix obal to add the prerelease to the message

PR #6152, opened 2020-12-17
Rewrite compare-with-tags in Python

ekohl:
The Ruby implementation relies on the configparser gem which is no longer in Fedora. This Python without any additional packages.


rel-eng/compare-with-tags
```diff
@@ -0,0 +1,64 @@
...
+
+def get_koji_packages(tag: str) -> Set[str]:
+    command = ['koji', 'list-pkgs', f'--tag={tag}', '--quiet']
+    output = check_output(command, universal_newlines=True)
+    return {line.split(None, 1)[0] for line in output.splitlines()}
+
+
+def main(filename: Path) -> None:
+    cfg = ConfigParser()
+    cfg.read(filename)
```
evgeni, line 27: this silently fails if `failname` doesn't exist, which is rather weird. not that this should happen in our setup (unless Evgeni copies the file to the wrong folder for testing), but still weird

evgeni (approved):
the code works fine when I'm not too stupid, so ACK with a nit

rel-eng/compare-with-tags
ekohl: Updated with code to verify it was read.

evgeni (approved):

PR #6153, opened 2020-12-17
Clean up tito.props and comps

ekohl:
2c624dcdbe4f91b7a59142db1d4b07043f184eed removed nodejs-node-gyp from non-SCL, but left one entry in.

162257c13fb80076465f01c055c55a762e032412 removed nodejs-css-element-queries but left a tito entry.

eb6690230abb3ca5bae80d67a2936b20c2a5601d added an entry for rubygem-activepack, rubygem-activeview and rubygem-turbolinks but the actual packages don't exist.

### DIFF
--- a/comps/comps-foreman-el8.xml
+++ b/comps/comps-foreman-el8.xml
@@ -123,13 +123,11 @@
       <packagereq type="default">rubygem-actionview</packagereq>
       <packagereq type="default">rubygem-activejob</packagereq>
       <packagereq type="default">rubygem-activemodel</packagereq>
-      <packagereq type="default">rubygem-activepack</packagereq>
       <packagereq type="default">rubygem-activerecord</packagereq>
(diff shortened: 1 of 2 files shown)


comps/comps-foreman-el8.xml
```diff
@@ -249,7 +247,6 @@
       <packagereq type="default">rubygem-ovirt-engine-sdk</packagereq>
       <packagereq type="default">rubygem-paint</packagereq>
       <packagereq type="default">rubygem-parallel</packagereq>
-      <packagereq type="default">rubygem-passenger</packagereq>
```
evgeni, line 252: there is also `rubygem-passenger-doc` in this comps.xml that can be removed.

evgeni (changes requested):

comps/comps-foreman-el8.xml
ekohl: Updated

evgeni (approved):

PR #6154, opened 2020-12-17
Rpm/develop update rubygem sinatra 2.1.0

ehelms:

PR #6155, opened 2020-12-17
Update rubygem-rack-protection to 2.1.0

ehelms:


ehelms:
This will fail repoclosure for current Sinatra as the two are tied closely together. Should go with -- https://github.com/theforeman/foreman-packaging/pull/6154

ehelms:
I put this over here -- https://github.com/theforeman/foreman-packaging/pull/6154

PR #6165, opened 2021-01-05
Fix deployment of bundler.d files for ansible-core

adamruzicka:
For remote execution the `gem 'foreman_remote_execution_core'` gets deployed when installing `ruby-smart-proxy-remote-execution-ssh`, which feels less clean than this approach.

CPs:
- [ ] 2.3
- [ ] 2.2

### DIFF
--- a/dependencies/bionic/foreman_ansible_core/changelog
+++ b/dependencies/bionic/foreman_ansible_core/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-ansible-core (4.0.0-1) stable; urgency=low
+
+  * 4.0.0 released
+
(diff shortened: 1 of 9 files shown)


ekohl (commented):
Is it intentional that it's only done for buster?

adamruzicka:
Eh, of course not, fixed.

evgeni:
@Baboon92 want to have a look here too?

evgeni (approved):
this still doesn't make ansible work on debian out of the box, but at least the gem is loaded now

I'll open a redmine with all the steps needed to finish this one off.

PR #6166, opened 2021-01-05
Add generated bundled information for f-installer

ekohl:
It's very useful to know which versions of modules are shipped by the installer. It's tedious to maintain this by hand so the dependency generator is used to generate it on the fly.

It's a draft since I don't even know if this is valid. https://rpm.org/user_doc/dependency_generators.html only describes the case where it's common, but for this use case it's mostly just painful to ship a macros package as a build dependency. This first tries out if it can directly use a Source.


ekohl:
This doesn't work. Keeping it open for now to try some other things out. If anyone has a good idea, I've love to hear it.

PR #6169, opened 2021-01-06
Update rubygem-highline to 2.0.3

ehelms:


ehelms:
Planning to use these scratch builds for some testing first

ehelms:
I tested a few actions from foreman-maintain, change-hostname and hammer to verify this version of highline would work. I ran into no issues. There may be some still possibly, but I think we should take in the change and see if we encounter them in nightly.

evgeni (approved):

packages/foreman/rubygem-kafo/rubygem-kafo.spec
```diff
@@ -25,7 +25,7 @@ Requires: %{?scl_prefix}rubygem(logging) < 3.0.0
 Requires: %{?scl_prefix}rubygem(clamp) >= 0.6.2
 Requires: %{?scl_prefix}rubygem(clamp) < 1.3.1
 Requires: %{?scl_prefix}rubygem(highline) >= 1.6.21
-Requires: %{?scl_prefix}rubygem(highline) < 2.0
+Requires: %{?scl_prefix}rubygem(highline) < 2.1.0
```
ekohl, line 28: Looks like the proper update procedure wasn't followed. Updating to 6.2.0 should have updated this
```suggestion
Requires: %{?scl_prefix}rubygem(highline) < 3.0.0
```

It has the correct markings so `bump_rpm.sh` should have caught this. If you used obal to bump and it doesn't do this, IMHO it means obal should not be used to bump gem RPMs until it's fixed to update dependencies.

ekohl (changes requested):

ekohl (approved):
Given currently nightly is red, I'm wondering if we should first make it green to keep the number of possibly breaking changes small. Other than that :+1:

evgeni:
Foreman nightlies have been green for a while, and with 2.4 stabilization approaching, we should get that either in, or wait longer, so I vote for in.

PR #6171, opened 2021-01-07
Update foreman-tasks to 3.0.3

adamruzicka:
CPs:
- [ ] 2.3


evgeni (changes requested):
you need to bump https://github.com/theforeman/foreman-packaging/blob/2aeda709ccf24ff20358af96b68a167dc5fc4d8a/plugins/ruby-foreman-tasks/debian/gem.list#L2

evgeni (approved):

PR #6172, opened 2021-01-07
bump_rpm.sh dependency handling

adamruzicka:
Having a somewhat fresh machine I ran into a couple of issues with `bump_rpm.sh`


ekohl (approved):
Thanks!

PR #6176, opened 2021-01-08
Fixes #31598 - Don't recreate AAAA record as A

jturel:
Evidently we aren't managing IPv6 downstream (see bugzilla), so I've made that the case here.

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -334,7 +334,7 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       begin
         aaaa_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::AAAA)
         commands << "update delete #{@old_hostname} AAAA"
-        commands << "update add #{@new_hostname} #{aaaa_record.ttl} A #{aaaa_record.address}"
+        commands << "update add #{@new_hostname} #{aaaa_record.ttl} AAAA #{aaaa_record.address}"
       rescue Resolv::ResolvError => e
         # This is fine
       end


packages/katello/katello/hostname-change.rb
```diff
@@ -331,12 +332,14 @@ def update_zone(zone, resolver)
...
-        aaaa_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::AAAA)
-        commands << "update delete #{@old_hostname} AAAA"
-        commands << "update add #{@new_hostname} #{aaaa_record.ttl} A #{aaaa_record.address}"
-      rescue Resolv::ResolvError => e
-        # This is fine
+      unless @options[:scenario] == SATELLITE_SCENARIO
+        begin
+          aaaa_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::AAAA)
+          commands << "update delete #{@old_hostname} AAAA"
+          commands << "update add #{@new_hostname} #{aaaa_record.ttl} AAAA #{aaaa_record.address}"
```
jturel, line 339: Note this was previously 'A' but I changed it to 'AAAA'. Let me know if that is not correct.

jturel:
@chris1984 @jeremylenz here's what I came up with - lemme know if it matches what we talked about earlier this week

packages/katello/katello/hostname-change.rb
jeremylenz: I think it should have been AAAA all along.  Why would we delete an AAAA but add back an A... This is what was causing the bug.

chris1984:
> @chris1984 @jeremylenz here's what I came up with - lemme know if it matches what we talked about earlier this week

Looks like what you suggested. +1 on @jeremylenz comment, that is prob what was causing it.

packages/katello/katello/hostname-change.rb
ekohl: Like @jeremylenz says, it should be AAAA. I think that we shouldn't make scenario specific rules for this since it has nothing to do with a scenario. The fact that this code was triggered by someone indicates it can be useful in some scenarios.

ekohl (approved):

packages/katello/katello/hostname-change.rb
jturel: That makes sense to me! Updated

packages/katello/katello/hostname-change.rb
```diff
@@ -334,7 +334,7 @@ def update_zone(zone, resolver)
       begin
         aaaa_record = resolver.getresource(@old_hostname, Resolv::DNS::Resource::IN::AAAA)
         commands << "update delete #{@old_hostname} AAAA"
-        commands << "update add #{@new_hostname} #{aaaa_record.ttl} A #{aaaa_record.address}"
+        commands << "update add #{@new_hostname} #{aaaa_record.ttl} AAAA #{aaaa_record.address}"
```
evgeni, line 337: While this fixes the "original" issue mentioned in the BZ, I think the whole v6 block is not exactly correct.

When we look at the BZ, it tried to add `FE80::5054:FF:FE6C:2A3B` which is a link-local address and should never be in DNS…

evgeni: But then again, this is what the "remote" server answers, so maybe it's not our fault

ekohl: You can detect those `IPAddr.new(aaaa_record.address).link_local?` but I agree that it's out of scope for this PR.

ekohl:
I'm merging this since it's clearly a bug. If there's follow up work, a new PR can be opened. Of course I just realized that this doesn't revbump the katello package so it won't be released until then ...

PR #6188, opened 2021-01-13
Update rubygem-foreman_remote_execution to 4.2.2

adamruzicka:
CPs:
- [ ] 2.3


pcreech:
@adamruzicka looks like it's depending on an older version of foreman-builder 

``` DEBUG util.py:600:  No matching package to install: 'npm(@theforeman/builder) < 5.0.0' ```

adamruzicka:
The upper bound was added by bump_rpm. I can remove it again, but I have no idea if and how it might affect the build

PR #6202, opened 2021-01-20
Update rubygem-fog-ovirt to 2.0.1

shiramax:


tbrisker (commented):
not crucial, but on line 90 you can remove the exclude for the Gemfile.local.rb file which was removed in https://github.com/fog/fog-ovirt/commit/9a894e9a60743fa9c540c714217fe3682401c9e9

tbrisker (approved):
Thanks @shiramax ! I think you can also open a seperate PR to remove rbovirt package, iirc it isn't used anywhere else.

PR #6208, opened 2021-01-21
Update rubygem-foreman_remote_execution to 4.2.2

adamruzicka:
(cherry picked from commit 16da588bfeda861117d8cfc051b6fa93fc76b13b)


packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -48,6 +48,7 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
 BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
 BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
 BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 4.14.0
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) < 5.0.0
```
evgeni, line 51: This is technically not necessary, and not present in the rpm/develop commit -- mind removing it?

evgeni (changes requested):

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
adamruzicka: Not at all, done.

evgeni (approved):

PR #6219, opened 2021-01-25
foreman-bootloaders-redhat: Add efinet module to Grub2 modules.

olifre:
This module is needed for network connectivity on UEFI systems,
and used e.g. for loading the grub.cfg and additional Grub2 modules
when booting via UEFI PXE.

### DIFF
--- a/packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
+++ b/packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
@@ -1,5 +1,5 @@
 Name: foreman-bootloaders-redhat
-Version: 202005201200
+Version: 202102211600
 Release: 1%{?dist}
 Summary: Metapackage with Grub2 and Shim TFTP bootloaders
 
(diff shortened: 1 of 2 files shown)


evgeni:
ok to test

@lzap any opinion here?

lzap (approved):
Yes, this is correct. On Red Hat system we copy bootloader, but on Debians we generate them.

I guess package version needs to be bumped in order to release a new packag, can you do that please in the `SPEC` file @olifre ?

olifre:
> I guess package version needs to be bumped in order to release a new packag, can you do that please in the `SPEC` file @olifre ?

Sure, I can. Do you want this "per PR", i.e. both here and in #6220, or one bump in the PR you merge as the last one?

evgeni:
@olifre if you fold both PRs into one (with two commits, preferably), then you can have one release/changelog bump, would that work?

olifre:
> @olifre if you fold both PRs into one (with two commits, preferably), then you can have one release/changelog bump, would that work?

Of course, I am mainly asking on your preference here (most projects prefer separate PRs for separate features to keep reviews and tests separate). ;-)
One final question: Do you prefer adding both commits to one of the existing PRs (+ `SPEC` bump), or a third PR with both changes plus the `SPEC` bump? I'll then proceed in whichever way you prefer :-).

evgeni:
"whatever" ;)

and I do agree on the "one change per PR" practice, but the way we build packages, it would mean we build it twice, for nothing.

olifre:
> "whatever" ;)

That's easy ;-). 
So I just took this PR here and cherry-picked #6220 over, and bumped the `spec` in another commit, also adding a changelog entry for both changes. 


> and I do agree on the "one change per PR" practice, but the way we build packages, it would mean we build it twice, for nothing.

Ahh, thanks for the clarification! Given the way packages are built, I do then fully understand.

lzap (approved):
Thanks.

PR #6220, opened 2021-01-25
foreman-bootloaders-redhat: Set prefix "grub2" in generated Grub2.

olifre:
This is the actual prefix used in the tftpboot layout,
and embedding it into the image increases compatibility.

An example case are DHCP servers sending the boot_file in
DHCP option 67 instead of using the classic BOOTP field
to conserve space.
This is not yet handled in Grub2 < 2.0.4, which then fails
to find its configuration unless the prefix is known.


evgeni:
ok to test

@lzap any opinion here?

lzap (approved):
Yeah, makes sense.

Can you also bump the SPEC version?

If you want to fix the same for Debian systems, I think the same command is called from foreman-installer (Puppet) there. For Red Hats we do have a special RPM package to support non-intel architectures.

olifre:
> If you want to fix the same for Debian systems, I think the same command is called from foreman-installer (Puppet) there. For Red Hats we do have a special RPM package to support non-intel architectures.

I see. For now, we don't have any Debian TFTP smart proxies to test this on, so in our case, the fix here is sufficient. 
I've cherry-picked this change into #6219 including the `spec` bump so all can be merged and built in one go.

lzap:
Thank you, closing this one then.

PR #6221, opened 2021-01-25
Add rubygem-qpid_proton gem package

jturel:


jturel:
Looks like koji node doesn't have access to qpid-proton-cpp-devel: http://koji.katello.org/kojifiles/work/tasks/4230/394230/mock_output.log

pcreech:
[test]

pcreech:
[test rpm]

evgeni:
#6231 contained a rubygem-qpid_proton, so is this obsolete then?

ehelms:
Ha, I didnt even notice. I think we still need this in order to produce an SCL version of it for EL7. We just need to restrict this to only build on EL7.

jturel:
@ehelms updated!

ehelms (approved):

PR #6226, opened 2021-01-28
Prepare add_gem_pkg for non-EL7

ekohl:
This rewrites the add_gem_to_comps function to support multiple distro versions. To really make use of this, the tito tag logic should also be updated to support it.


ekohl:
By now we've merged a solution.

PR #6228, opened 2021-01-28
Update foreman_expire_hosts to 7.0.4

laugmanuel:
FYI @timogoebel


evgeni (changes requested):
```
   /usr/share/gems/gems/foreman_expire_hosts-7.0.2/.github/workflows/ci.yml
```

You want to `%exclude` that file I guess. (Well, probably the whole `.github` directory)

laugmanuel:
> You want to `%exclude` that file I guess. (Well, probably the whole `.github` directory)

Actually not. This is caused by the use of `git ls-files` in the Gemspec. 
I've already opened a PR here: https://github.com/theforeman/foreman_expire_hosts/pull/65

After this is merged and a new Gem is released, I will modify this PR to match the new version.

evgeni (approved):

PR #6229, opened 2021-01-28
Add jsoncpp

pcreech:

### DIFF
--- a/rel-eng/tito.props
+++ b/rel-eng/tito.props
@@ -1045,7 +1045,8 @@ whitelist = katello
 
 [katello-nightly-el8]
 disttag = .el8
-whitelist = katello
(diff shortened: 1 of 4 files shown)


rel-eng/tito.props
```diff
@@ -963,6 +963,7 @@ disttag = .el8
 # don't whitelist foreman-release here, it's already in foreman-nightly-el8
 whitelist =
   ansible-collection-theforeman-foreman
+  jsoncpp
```
ehelms, line 966: I think this should be in the katello-nightly-el8 tag since it will be used to build qpid_proton packages ?

ehelms (approved):
Assuming this is a build time only dependency and not needed in comps

PR #6235, opened 2021-01-29
depend on ansible-core

evgeni:

### DIFF
--- a/plugins/smart_proxy_ansible/changelog
+++ b/plugins/smart_proxy_ansible/changelog
@@ -1,3 +1,10 @@
+ruby-smart-proxy-ansible (3.0.0-3) stable; urgency=low
+
+  * Depend on ansible_core
+  * Don't ship json_inventory.sh - it's in core.
+
+ -- Evgeni Golov <evgeni@debian.org>  Fri, 29 Jan 2021 16:01:00 +0100
+
 ruby-smart-proxy-ansible (3.0.0-2) stable; urgency=low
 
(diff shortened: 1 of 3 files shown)


ekohl (commented):
I think it should be removed from `smart_proxy_ansible`. AFAIK nothing in that repo refers to it.

evgeni:
https://github.com/theforeman/smart_proxy_ansible/blob/9fe3018710216f99c0ef6aea85cb7c05772cbe76/lib/smart_proxy_ansible/plugin.rb#L16-L17

And without -core being installed, I can't do run ansible on the proxy.

evgeni:
Oh and RPM does the same: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec#L37

Not saying that's the perfect deployment strategy, but it at least works.

evgeni:
Or do you mean `inventory.sh`?

ekohl:
Yes, "it" indeed referred to `inventory.sh`. So rather than removing it in packaging, let's remove it from the source.

evgeni:
I don't disagree, but this route is quicker to get working foreman_ansible on Debian in the meantime :)

evgeni:
(https://github.com/theforeman/smart_proxy_ansible/pull/27 drops it from the gem)

ekohl (approved):
I'm ok with a workaround.

PR #6237, opened 2021-01-29
Only install pulp-selinux on Katello server for now

ehelms:


ehelms:
Can drop this fully once we drop Pulp 2 for katello-agent.

ekohl (approved):

PR #6239, opened 2021-01-30
Bump rubygem-qpid_proton to 0.33.0

jturel:


ehelms (approved):
Expected repoclosure failure, this should fix it

PR #6241, opened 2021-02-01
Fix deployment of bundler.d files for ansible-core

adamruzicka:
(cherry picked from commit 487c6acade5832f4dbb07172365affd38b9b756f)


evgeni:
obsoleted by https://github.com/theforeman/foreman-packaging/pull/6306 which has more picks that we need

PR #6243, opened 2021-02-01
Require container-selinux by Katello

ehelms:
Temporary until https://github.com/pulp/pulpcore-selinux/pull/33 is
released in a Pulp 3 release. Otherwise pulpcore-api.socket cannot
be accessed by Apache.

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,7 +5,7 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 4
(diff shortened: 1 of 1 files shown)


ehelms:
[test rpm]

evgeni:
Is EL8 sufficient? EL7 doesn't fail on the same issues? Then ACK as a temp-workaround

evgeni:
repoclosure unrelated

ehelms:
I do not see it failing on EL7 from my testing. I would have also expected us to have seen a reported failure from someone at this point given Pulp 3 has been in use in previous Katello releases.

PR #6246, opened 2021-02-01
Require qpid and dispatch-router on EL8 and EL7

ehelms:

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,7 +5,7 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 6
+%global release 7
 
 Name:       katello
 Version:    4.0.0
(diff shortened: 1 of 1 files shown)


packages/katello/katello/katello.spec
```diff
@@ -36,6 +36,10 @@ Requires: %{?scl_prefix}rubygem-hammer_cli
 Requires: %{?scl_prefix}rubygem-hammer_cli_foreman
 Requires: %{?scl_prefix}rubygem-hammer_cli_katello
 
+# katello-agent requirements
+Requires: qpid-cpp-server-linearstore
+Requires: qpid-cpp-client-devel
```
jlsherrill, line 41: is client-devel actaully a run time requirement?

packages/katello/katello/katello.spec
```diff
@@ -113,11 +114,17 @@ install -m 644 ./manpages/katello-change-hostname.8.gz %{buildroot}/%{_mandir}/m
...
 Requires:       rubygem-highline
 Requires:       %{name}-debug
 %if 0%{?rhel} == 8
 Requires:       container-selinux
 %endif
+=======
+Requires: rubygem-highline
+Requires: %{name}-debug
+Requires: qpid-dispatch-router
+>>>>>>> 6ddc6d02c (Require qpid and dispatch-router on EL8 and EL7)
```
jlsherrill, line 127: image

packages/katello/katello/katello.spec
ehelms: Thats a good question

jlsherrill: oh i see this is actually down below, is it actually required?

ehelms: One way to find out, removed.

jlsherrill (approved):

PR #6257, opened 2021-02-02
Update rubygem-foreman_rh_cloud to 3.0.15

laviro:


packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -42,18 +44,18 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
 
 # start package.json devDependencies BuildRequires
 BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
-BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
-BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 4.0.2
-BuildRequires: %{?scl_prefix}npm(@theforeman/builder) < 5.0.0
-BuildRequires: %{?scl_prefix}npm(raf) >= 3.4.0
-BuildRequires: %{?scl_prefix}npm(raf) < 4.0.0
+BuildRequires: %{?scl_prefix}npm(@babel/core) < 7.8.0
```
laviro, line 47: pinned in our plugin's package.json to `~7.7.0` which is used in packaging `2.3` to be more specific.

packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -42,18 +44,18 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
...
+BuildRequires: %{?scl_prefix}npm(@redhat-cloud-services/frontend-components) >= 2.5.0
+BuildRequires: %{?scl_prefix}npm(@redhat-cloud-services/frontend-components) < 3.0.0
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 4.14.0
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) < 4.15.0
```
laviro, line 51: pinned in our plugin's package.json to `~4.14.0` which is used in packaging `2.3` to be more specific.

packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -42,18 +44,18 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
 
 # start package.json devDependencies BuildRequires
 BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
-BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
-BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 4.0.2
-BuildRequires: %{?scl_prefix}npm(@theforeman/builder) < 5.0.0
-BuildRequires: %{?scl_prefix}npm(raf) >= 3.4.0
-BuildRequires: %{?scl_prefix}npm(raf) < 4.0.0
```
laviro, line 49: `raf` wasn't used on our side

packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -42,18 +44,18 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
...
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) < 4.15.0
 # end package.json devDependencies BuildRequires
 
 # start package.json dependencies BuildRequires
 BuildRequires: %{?scl_prefix}npm(jed) >= 1.1.1
-BuildRequires: %{?scl_prefix}npm(jed) < 2.0.0
+BuildRequires: %{?scl_prefix}npm(jed) < 1.2.0
 BuildRequires: %{?scl_prefix}npm(react-intl) >= 2.8.0
-BuildRequires: %{?scl_prefix}npm(react-intl) < 3.0.0
+BuildRequires: %{?scl_prefix}npm(react-intl) < 2.9.0
```
laviro, line 58: These two were also pinned to be more specific with what packaging uses

packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -28,10 +28,12 @@ Requires: %{?scl_prefix_ruby}ruby
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
 Requires: %{?scl_prefix}rubygem(katello)
 Requires: %{?scl_prefix}rubygem(redhat_access)
+Requires: %{?scl_prefix}rubygem(foreman_ansible)
 BuildRequires: foreman-assets >= %{foreman_min_version}
 BuildRequires: foreman-plugin >= %{foreman_min_version}
 BuildRequires: %{?scl_prefix}rubygem(katello)
 BuildRequires: %{?scl_prefix}rubygem(redhat_access)
+BuildRequires: %{?scl_prefix}rubygem(foreman_ansible)
```
laviro, line 36: needed by the new cloud connector feature

evgeni:
The diff looks reasonable, but I wonder why it's *only* against `rpm/2.3`?

The pins won't work in 2.4/develop, so do we expect a 4.x version of the plugin there?

laviro:
Thanks @evgeni, we tested these versions with `2.3`, as for forward-compatibility, we will see if we can do some magic to catch also theforeman-js packages which moved to version 6 in 2.4 and 8 in develop, if we will see other things are also breaking we will bump to the next plugin version, `v4.x` as you mentioned.

evgeni:
@laviro my problem with this is that if we merge this, 2.3 users will get the .15 version, while users deploying 2.4 (that just branched) will end up on .14, which sounds wrong to me?

laviro:
Gotcha... I see the only conflicting package is `@theforeman/builder` and its siblings,
looks like we can use `>=` in package.json, see an example in [foreman](https://github.com/theforeman/foreman/blob/develop/package.json#L63)

so in our plugin, we could specify `"@theforeman/builder": ">=4.14.0"`

another option would be to use ranges: `"@theforeman/builder": "~4.14.0 || ~6.0.0"`

though in both options, I am not sure how we could decide which version to use in development. cc @shimshtein @sharvit

we could also remove the top border from line 51: `BuildRequires: %{?scl_prefix}npm(@theforeman/builder) < 4.15.0`
so it would fit in packaging for `2.3` and above

evgeni:
for the record, in other plugins, we've been just removing the upper bound of the vendor/builder dep, as Avi was saying that the recent major bumps were not incompatible in terms of building. (and the results were looking fine)

laviro:
Opened CPs:
- rpm/2.4: #6269 
- rpm/develop: #6270

evgeni (approved):

PR #6268, opened 2021-02-03
only copy webpack if it exists

evgeni:

### DIFF
--- a/debian/bionic/foreman/foreman-plugin.mk
+++ b/debian/bionic/foreman/foreman-plugin.mk
@@ -3,17 +3,19 @@ build:
 	cp $(PLUGIN).rb /usr/share/foreman/bundler.d/
 	cd /usr/share/foreman && ( \
 		/usr/bin/foreman-ruby /usr/bin/bundle install --local && \
-		/usr/bin/npm install --no-audit --no-optional --unsafe-perm && \
-		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production DATABASE_URL=nulldb://nohost && \
-		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
-			OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
-			DATABASE_URL=nulldb://nohost FOREMAN_APIPIE_LANGS=en_US \
+		if [ "$(PLUGIN_HAS_PACKAGE_JSON)" = "true" ]; then /usr/bin/npm install --no-audit --no-optional --unsafe-perm; fi && \
+		if [ "$(PLUGIN_HAS_ASSETS)" = "true" ] || [ "$(PLUGIN_HAS_WEBPACK)" = "true" ]; then /usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production DATABASE_URL=nulldb://nohost; fi && \
+		if [ "$(PLUGIN_HAS_API)" = "true" ]; then \
+			/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
+				OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
+				DATABASE_URL=nulldb://nohost FOREMAN_APIPIE_LANGS=en_US; \
+		fi \
 	)
 	GEM_PATH=$$(cd /usr/share/foreman && /usr/bin/foreman-ruby /usr/bin/bundle show $(PLUGIN)) && \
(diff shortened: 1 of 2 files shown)


debian/bionic/foreman/foreman-plugin.mk
```diff
@@ -11,7 +11,7 @@ build:
 	)
 	GEM_PATH=$$(cd /usr/share/foreman && /usr/bin/foreman-ruby /usr/bin/bundle show $(PLUGIN)) && \
 	cp -rp $${GEM_PATH}/public/assets ./ && \
```
ekohl, line 13: At this point, does `&& \` make sense?

debian/bionic/foreman/foreman-plugin.mk
```diff
@@ -11,7 +11,7 @@ build:
...
-	cp -rp $${GEM_PATH}/public/webpack ./
+	if [ -e $${GEM_PATH}/public/webpack ]; then cp -rp $${GEM_PATH}/public/webpack ./; fi
```
ekohl, line 14: Can we use multiple lines?

ekohl (approved):

debian/bionic/foreman/foreman-plugin.mk
evgeni: the `\` is needed, to get `$GEM_PATH`, otherwise it's a new shell… the `&&` not so much

ekohl: Ah, so it's combining them and `&&` is just the way to run multiple commands in the same shell. Got it.

evgeni: yeah, running subshells in make is wild

evgeni:
So this works now, but is completely opt-in, and I wonder how correct that is?

I briefly looked at the RPM packaging and the macro usage there.

We have 32 packages using the `foreman_precompile_plugin` macro (that's roughly equivalent to the first section of this makefile). Out of these:
* 20 are using `-a -s` aka "API and Rails assets/Webpack"
* 7 are using only `-a` aka "API"
* 5 are using only `-s` aka "Rails assets/Webpack"

We also have macros for the installation part (a mix of the second part of the makefile and `debian/install`), where we have the following split:
* 25 `foreman_assets_plugin` aka classic Rails assets
* 11 `foreman_webpack_plugin` aka Webpack
* 25 `foreman_apipie_cache_plugin` aka API

Looking at that, I'd argue that enabling API and Rails assets by default is a safe bet. Weback not (yet?).

That would mean that packages using webpack will have to set `PLUGIN_HAS_WEBPACK = true` in their `debian/rules` and those who don't want API/Rails assets something like `PLUGIN_HAS_API = false` (and I'd need to adjust this Makefile to accept that).

Would love y'all input @ekohl @baboon92 @pcreech @zjhuntin @ehelms

debian/bionic/foreman/foreman-plugin.mk
```diff
@@ -3,17 +3,19 @@ build:
...
-			DATABASE_URL=nulldb://nohost FOREMAN_APIPIE_LANGS=en_US \
+		if [ -n "$(PLUGIN_HAS_PACKAGE_JSON)" ]; then /usr/bin/npm install --no-audit --no-optional --unsafe-perm; fi && \
+		if [ -n "$(PLUGIN_HAS_ASSETS)" ] || [ -n "$(PLUGIN_HAS_WEBPACK)" ]; then /usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production DATABASE_URL=nulldb://nohost; fi && \
+		if [ -n "$(PLUGIN_HAS_API)" ]; then \
+			/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
+				OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
+				DATABASE_URL=nulldb://nohost FOREMAN_APIPIE_LANGS=en_US; \
+		fi \
 	)
 	GEM_PATH=$$(cd /usr/share/foreman && /usr/bin/foreman-ruby /usr/bin/bundle show $(PLUGIN)) && \
```
ekohl, line 14: I notice this now and unrelated to this PR, but `bundle show` started to output that it's deprecated and `bundle info` should be used. That's not great because `bundle show` is exactly what we need while info has more.

ekohl (approved):
Even if it's not perfect yet, I'd be ok with merging this and iterating on it.

ekohl:
As for defaults: I'm leaning to having a bit of boiler plate for plugin building that explicitly sets all 3. We have seen in the past that people copy other plugins and modify. If all 3 options are there, it's more likely that they'll take a look rather than looking in `foreman-plugin.mk`.

evgeni:
Okay, I've updated it to explicitly check for `true`, as when people copy the example, they might end up setting the value to `false` and that was not accounted for before.

Ideally, tho, people won't have to copy anything, but we'd have a thing that detects features and creates packaging from a template, just like with RPM.

evgeni:
given this is ack'ed, I am going to merge it as is and iterate over it later, thanks

PR #6277, opened 2021-02-04
Bump minimum foreman version for rubygem-katello

ehelms:
This is to get nightly running again, which we will need to see how the changes related to katello-agent are going. I think this shouldn't hurt anything at this point given we are also on the cusp of Katello branching (I think they are waiting on the katello-agent stuff, so a bit of a chicken-egg).

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -2,14 +2,14 @@
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
-%global foreman_min_version 2.4
-%global foreman_max_version 2.5
+%global foreman_min_version 2.5
+%global foreman_max_version 2.6
 %global plugin_name katello
 %global gem_name katello
 %global prereleasesource pre.master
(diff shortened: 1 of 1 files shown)


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -3,13 +3,13 @@
 %{!?scl:%global pkg_name %{name}}
 
 %global foreman_min_version 2.4
```
ekohl, line 5: Should this be 2.5 to match?
```suggestion
%global foreman_min_version 2.5
```

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -3,13 +3,13 @@
...
-%global foreman_max_version 2.5
+%global foreman_max_version 2.6
 %global plugin_name katello
 %global gem_name katello
 %global prereleasesource pre.master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global mainver 4.0.0
```
jlsherrill, line 11: master got bumped to 4.1 last night, so probably need that too

jlsherrill (approved):

PR #6281, opened 2021-02-04
Drop Pulp 2 requires from katello

ehelms:

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,7 +5,7 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 7
+%global release 8
 
 Name:       katello
 Version:    4.0.0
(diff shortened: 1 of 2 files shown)


ekohl (commented):
We also have https://github.com/theforeman/foreman-packaging/blob/e0df5ab38228d8f10b5c1ed8342b9e72de294cf6/packages/katello/katello/katello.cron and I wonder if those are still needed in a Pulp 3 world.

Other than that, it looks good.

packages/katello/katello/katello.spec
```diff
@@ -5,7 +5,7 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 7
+%global release 8
 
 Name:       katello
 Version:    4.0.0
```
ekohl, line 11: Aren't we at 4.1.0 by now?

ehelms: I think folks are working on that, but this PR is not the one to change the version. And this change applies to 4.0 as well.

evgeni:
seems like it's something that exists in pulp3 too: https://github.com/katello/katello/blob/master/app/lib/actions/katello/orphan_cleanup/remove_orphans.rb

(not sure that code will not explode w/o pulp2, but that's a different question)

evgeni (approved):

PR #6284, opened 2021-02-05
Cherry picks katello 4.0 qpid

ehelms:
Became a bit of doozy @evgeni -- but I am pretty sure I got everything up to todays changes.


evgeni:
[test rpm]

packages/katello/katello/katello.spec
```diff
@@ -113,18 +96,18 @@ install -m 644 ./manpages/katello-change-hostname.8.gz %{buildroot}/%{_mandir}/m
 BuildArch:  noarch
 Summary:    Common runtime components of %{name}
 
-Requires:       rubygem-highline
-Requires:       %{name}-debug
+Requires: rubygem-highline
+Requires: %{name}-debug
+Requires: qpid-dispatch-router
+%if 0%{?rhel} == 8
+Requires: container-selinux
```
evgeni, line 103: we have a fixed pulpcore-selinux now, should we yank that requirement in develop, test it again, and pick that change too?

package_manifest.yaml
```diff
@@ -1087,6 +1092,23 @@ repoclosures:
...
+    katello-repoclosure-el8:
+      repoclosure_target_repos:
+        el8:
+          - "el8-katello-{{ katello_version }}"
+      repoclosure_target_dist: el8
+      repoclosure_lookaside_repos:
+        el8:
+          - el8-baseos
+          - el8-appstream
+          - el8-powertools
```
evgeni, line 1104: that means our users need to enable powertools too?

package_manifest.yaml
```diff
@@ -1087,6 +1092,23 @@ repoclosures:
...
+          - el8-devel
+          - el8-extras
+          - el8-powertools
```
evgeni, line 1107: dupe?

evgeni (commented):
I'm fine with the diff, but left two inline comments.

Another thing: this currently fails to build rubygem-katello as there is no nightly builder configured for the branch. Should we remove rubygem-katello here and include the qpid_proton dependency when we build katello rc1?

package_manifest.yaml
ehelms: That is unfortunately, correct. They need it for `libdb_cxx`.

packages/katello/katello/katello.spec
ehelms: We should. Has that pulpcore-selinux fix been released to 3.9?

package_manifest.yaml
ehelms: https://github.com/theforeman/foreman-packaging/pull/6292

packages/katello/katello/katello.spec
evgeni: it did, yes: https://yum.theforeman.org/pulpcore/3.9/el8/x86_64/pulpcore-selinux-1.2.4-1.el8.x86_64.rpm

package_manifest.yaml
evgeni: :+1:

package_manifest.yaml
evgeni: :crying_cat_face: but oh well, I don't want to carry another set of packages for that crazyness. is updating install/upgrade docs on the 4.0 "release table"?

evgeni (commented):
ACK if: rubygem-katello change is split out (so this can actually be built), and it then builds fine ;)

ehelms:
'Tis dropped.

evgeni (approved):
ACK, thanks!

will you do the tag-dance?

ehelms:
> ACK, thanks!
> 
> will you do the tag-dance?

Will do

PR #6291, opened 2021-02-08
Update rubygem-net-ssh to 6.1.0

ezr-ondrej:
https://github.com/theforeman/foreman/pull/8313 unblocks this bump (or we might switch to OS default gem? :thinking: )


ekohl:
> (or we might switch to OS default gem? thinking )

The gem is in EPEL and with EL8 we're trying not to depend on it so bumping it here is fine. We'll also need it for EL7 anyway.

packages/foreman/rubygem-net-scp/rubygem-net-scp.spec
```diff
@@ -5,15 +5,14 @@
...
-Version: 1.2.1
-Release: 4%{?dist}
+Version: 3.0.0
+Release: 1%{?dist}
 Group: Development/Languages
 License: MIT
 URL: http://net-ssh.rubyforge.org/scp
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 # Fix test broken due to Net::SSH 4.0+.
 # https://github.com/net-ssh/net-scp/pull/30
```
ekohl: Shouldn't this be removed as well?

packages/foreman/rubygem-net-scp/rubygem-net-scp.spec
```diff
@@ -5,15 +5,12 @@
...
 Group: Development/Languages
 License: MIT
 URL: http://net-ssh.rubyforge.org/scp
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
-# Fix test broken due to Net::SSH 4.0+.
-# https://github.com/net-ssh/net-scp/pull/30
-Patch0: net-scp-1.2.1-Fix-compatiblity-with-net-ssh-4.0.patch
 Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
 Requires: %{?scl_prefix}rubygem(net-ssh)
```
ekohl, line 16: This now has a version bound. Perhaps the spec should be regenerated.

ezr-ondrej: good idea, thanks, looks better now :+1:

ezr-ondrej:
We can't do this yet. I guess I'll close for now

PR #6293, opened 2021-02-08
Update rubygem-fog-core to 2.2.3

ezr-ondrej:
We need this bump for #6291.


evgeni:
this will break `fog-google`:
https://github.com/theforeman/foreman-packaging/blob/252ac97135a017220a0fc86d23bfd1f05161a116/packages/foreman/rubygem-fog-google/rubygem-fog-google.spec#L22

and core *currently* pins 2.1.0 too, but that is fixed in https://github.com/theforeman/foreman/pull/8313

ekohl:
FYI: https://github.com/fog/fog-google/pull/422 was opened to unpin fog-core in fog-google, but it's been stale for a long time now.

ezr-ondrej:
> FYI: [fog/fog-google#422](https://github.com/fog/fog-google/pull/422) was opened to unpin fog-core in fog-google, but it's been stale for a long time now.

Right, so we will probably need to wait for that to happen, right?

ekohl:
Yes, it's blocked on it.

evgeni (changes requested):
marking as blocked/needs changes as per discussion wrt fog-google above

ekohl:
Also, repoclosure in CI doens't find this because it still considers the old package. However, if we do merge it I'd expect it to fail repoclosure.

ezr-ondrej:
> Also, repoclosure in CI doens't find this because it still considers the old package. However, if we do merge it I'd expect it to fail repoclosure.

Shouldn't it be tested automatically? It is something that's quite hard to test manually, wouldn't be for @evgeni's magic skills I would not notice it myself.
I guess it is because we don't want to fail on foreman dependencies? Could it be separate not required check then?

evgeni:
It is tested automatically, but due to the way how `dnf repoclosure` works it doesn't see this particular case.

Some background can be found in https://community.theforeman.org/t/issues-with-dnf-repoclosure/18495

pcreech:
This pr has some merge conflicts.  Is it still relevant?  Or should we close the PR?

ehelms:
This is old and conflicted, choosing to close and can be revisited in the future.

PR #6297, opened 2021-02-09
create a workdir for ansible on the proxy

evgeni:


ekohl (approved):

evgeni:
(still testing if that's actually fixes one of my problems)

PR #6300, opened 2021-02-09
Bump rubygem-foreman_rh_cloud to 2.0.16-1

ShimShtein:


evgeni:
```
   /opt/theforeman/tfm/root/usr/share/gems/gems/foreman_rh_cloud-2.0.16/public/assets/foreman_rh_cloud/foreman_rh_cloud.json
```

Foreman 2.1 is EOL, so even if we merge this, the package won't ever reach the users. I tend to just close it?

ShimShtein:
Agreed!

PR #6303, opened 2021-02-09
Katello 4.0.0 RC1

evgeni:


ekohl (approved):
I think repoclosure is expected since no mashing has happened yet.

PR #6311, opened 2021-02-12
Katello 3.17.2

evgeni:


evgeni:
seem repoclosure is unhappy as it couldn't reach the pulp2 repos, but those work fine from here, so I'd just ignore it for now…

chris1984 (approved):
ACK

PR #6321, opened 2021-02-17
Update rubygem-katello to 3.18.1.1 for internal testing

ianballou:
We are bumping rubygem-katello to 3.18.1.1 so we can test the Pulp 3 migration reset, which is merging straight into KATELLO-3.18 rather than master.

The only change in 3.18.1.1 is the Pulp 2 to 3 migration bindings update to 0.7.  There is also a PR to bump the bindings version for rpm/2.3: https://github.com/theforeman/foreman-packaging/pull/6320.

Since it's for testing, we aren't making a public release out of it. This is likely to happen again since the master branch has diverged so much from 3.18 Pulp3-wise.


ianballou:
The errors are due to the missing `tfm-rubygem(pulp_2to3_migration_client) >= 0.7.0`

jlsherrill:
[test katello]

jlsherrill:
[test rpm]

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -12,7 +12,7 @@
 Name:    %{?scl_prefix}rubygem-%{gem_name}
 Summary: Content and Subscription Management plugin for Foreman
 
-Version: %{mainver}
+Version: 3.18.1.1
```
jlsherrill, line 15: a script may have done this, but we usually update the mainver variable above

jlsherrill (approved):

PR #6322, opened 2021-02-18
Deploy foreman-puma-status script when installing foreman-service

ehelms:

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -9,7 +9,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 5
+%global release 6
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -672,6 +672,7 @@ Group:  Applications/System
 Requires: %{?scl_prefix}rubygem(puma) >= 5.1
 Requires: %{?scl_prefix}rubygem(puma) < 6.0
 # end specfile service Requires
+Requires: %{?scl_prefix}rubygem(puma-status)
 Requires: %{name} = %{version}-%{release}
 
 %description service
@@ -681,6 +682,7 @@ Meta Package to install requirements for Foreman service
 %{_unitdir}/%{name}.service
 %{_unitdir}/%{name}.socket
 %{_datadir}/%{name}/bundler.d/service.rb
+%{_sbindir}/%{name}-puma-status
 
 %description
 Foreman is aimed to be a Single Address For All Machines Life Cycle Management.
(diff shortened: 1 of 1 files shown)


ehelms:
Needs https://github.com/theforeman/foreman/pull/7735

packages/foreman/foreman/foreman.spec
```diff
@@ -670,6 +670,7 @@ Group:  Applications/System
 # start specfile service Requires
 Requires: %{?scl_prefix}rubygem(puma) >= 5.1
 Requires: %{?scl_prefix}rubygem(puma) < 6.0
+Requires: %{?scl_prefix}rubygem(puma-status)
```
ekohl, line 673: I don't see this reflected in foreman's bundler.d file so on an update it will be removed. If it should be tracked just in packaging it should be outside of `start/end specfile service Requires`.

ehelms: Since this a more optional production helper, we had thought to not including in bundler and keep it outside. So I'll move it.

packages/foreman/foreman/foreman.spec
```diff
@@ -706,6 +708,7 @@ plugins required for Foreman to work.
   done
   # script content
   sed -ri 'sX/usr/bin/rakeX%{scl_rake}X' extras/dbmigrate script/foreman-rake
+  sed -i '$ i\source scl_source enable tfm' script/foreman-puma-status
```
evgeni, line 711: I have no idea what this `sed` expression does.

Is it inserting `source scl …` before the ` puma-status /run/foreman/puma.state` in the script?
Wouldn't it be easier to read if it'd prepend that line with a `scl enable tfm -- `?

ekohl: I think inserting the line at line 2 is more straight forward.
```suggestion
  sed -i '2isource scl_source enable tfm' script/foreman-puma-status
```

I think adding a source line is more future proof. If the script is ever changed to multiple lines, it'll still work. You also know that the first line will remain the shebang so line 2 should be safe.

packages/foreman/foreman/foreman.spec
```diff
@@ -705,6 +707,8 @@ plugins required for Foreman to work.
   done
   # script content
   sed -ri 'sX/usr/bin/rakeX%{scl_rake}X' extras/dbmigrate script/foreman-rake
+  sed -i '2isource scl_source enable tfm' script/foreman-puma-status
+  when: foreman_installer_upgrade == True
```
ekohl, line 711: Copy paste error?
```suggestion
```

ehelms: Huh. I don't know even know how that little bit of magic made it in there.

evgeni:
ACK pending the actual script being in core

packages/foreman/foreman/foreman.spec
```diff
@@ -672,6 +672,7 @@ Group:  Applications/System
 Requires: %{?scl_prefix}rubygem(puma) >= 5.1
 Requires: %{?scl_prefix}rubygem(puma) < 6.0
 # end specfile service Requires
+Requires: %{?scl_prefix}rubygem(puma-status)
```
evgeni, line 675: Should `puma-status` be added to `bundler.d/service.rb` in core?

We'll need to smuggle it in there for Debian packaging anyways, and it feels cleaner to just add it in core instead of packaging?

evgeni:
[test rpm]

packages/foreman/foreman/foreman.spec
evgeni: According to https://github.com/theforeman/foreman/pull/7578#discussion_r405930554 it shall not, and I am not going up any hills today.

evgeni (approved):

PR #6323, opened 2021-02-18
Add support for foreman-puma-status command

ehelms:

### DIFF
--- a/debian/bionic/foreman/control
+++ b/debian/bionic/foreman/control
@@ -210,7 +210,7 @@ Package: foreman-service
 Architecture: all
 Section: web
 Priority: extra
-Depends: foreman (= ${binary:Version}), ${misc:Depends}
+Depends: foreman (= ${binary:Version}), puma-status, ${misc:Depends}
 Description: package providing Foreman services running via puma
  This package provides Foreman systemd service for Foreman,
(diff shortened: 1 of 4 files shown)


evgeni:
can you also add changes for focal, now that we support it?

debian/bionic/foreman/control
```diff
@@ -210,7 +210,7 @@ Package: foreman-service
 Architecture: all
 Section: web
 Priority: extra
-Depends: foreman (= ${binary:Version}), ${misc:Depends}
+Depends: foreman (= ${binary:Version}), puma-status, ${misc:Depends}
```
evgeni, line 213: where is this package coming from? I don't see it in Debian or our repos?

ehelms: Honestly, I don't know what determines when a package comes from one place or the other given we use bundle install for dependencies?

evgeni: We bundle things that are in bundler.d, not from the Debian Depends.

evgeni:
replaced by https://github.com/theforeman/foreman-packaging/pull/6567

DaviidFr (commented):
Do we have to get some @gas

PR #6325, opened 2021-02-19
foreman-bootloaders-redhat: Add efinet only on *-efi platforms.

olifre:
This fixes an issue introduced in 311d91c98ae7b39088d1aaa90777e8751405705f,
since `efinet.mod` is only provided for `*-efi` platforms.

I introduced that bug in #6219 / version `202102211600-1`. While the generation works, it fails if generation is done for a non-efi platform, since `efinet.mod` does not exist there.

### DIFF
--- a/packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
+++ b/packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
@@ -1,5 +1,5 @@
 Name: foreman-bootloaders-redhat
-Version: 202102211600
+Version: 202102220000
 Release: 1%{?dist}
 Summary: Metapackage with Grub2 and Shim TFTP bootloaders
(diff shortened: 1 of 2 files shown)


evgeni:
ok to test

lzap (approved):
Tested this time:

```
[root@sat69 ~]# foreman-generate-bootloaders
grub2-mkimage: error: cannot open `/usr/lib/grub/i386-pc//efinet.mod': No such file or directory.
```

After the patch:

    [root@sat69 ~]# foreman-generate-bootloaders 

Sorry about breaking this, I did not realize that.

olifre:
> Sorry about breaking this, I did not realize that.

I have to be sorry. I could have sworn I tested it, but apparently I did not test it well enough :-(. 

At least I realized before a new Foreman release was cut with it. Thanks a lot for merging this :-).

PR #6313, opened 2021-02-15
Add acts_as_list gem

xprazak2:

PR #6331, opened 2021-02-19
Add acts_as_list gem dependency

ezr-ondrej:
Takes over for #6313 and fixes the tests.


xprazak2:
Nice.

ekohl:
From the other PR:
> @zjhuntin do we need this in deb? this will be a dependency of foreman_ansible so in deb the bunler will install that, right? thinking

You are right that Foreman dependencies and its plugins use bundler. I think you need to add `acts_as_list` to https://github.com/theforeman/foreman-packaging/blob/deb/develop/plugins/ruby-foreman-ansible/debian/gem.list but I'd look at https://github.com/theforeman/foreman-packaging/blob/deb/develop/plugins/ruby-foreman-azure/debian/gem.list for a reference.

evgeni:
Given we vendored it in #6823, I am going to close here until something outside plugins needs it.

PR #6335, opened 2021-02-22
Bump rubygem-foreman_rh_cloud to 3.0.17-1

laviro:
(cherry picked from commit 420a1cf72b5cb17e1b47709fec499c26c4b21b74)


evgeni:
According to Shimon, this version conflicts with redhat_access 2.2.18, please add that to the RPM metadata. (Something like `Conflicts: %{?scl_prefix}rubygem-redhat_access < 2.2.19`)

laviro:
Thanks @evgeni @shimshtein, updated

ShimShtein:
@evgeni, thanks for the tip! The restriction sounds reasonable.

evgeni:
Just so that I understand the change correctly: the redhat_access plugin is not completely obsolete, you just take over a specific part of the functionality and users shall keep .19 installed for other features, right?

ShimShtein:
Correct. We took one part of its functionality (handling insights-client uploads) into rh_cloud. redhat_access is still responsible for adding "Insights" menu in foreman and all the pages there.

laviro:
After this will get in I will update the 2.4 and 2.3 cherry picks

evgeni:
Who will update the `redhat_access` package so that users can actually use both together?

ShimShtein:
I'll talk to @rexwhite, and we'll figure it out.

ShimShtein:
@evgeni, created PR for redhat access too. Should I consolidate them both?

evgeni (approved):

PR #6345, opened 2021-02-24
Bump rubygem-redhat_access to 2.2.19-1

ShimShtein:
Added dependency on rh_cloud plugin, since insights-client uploads are handled there now.


ShimShtein:
Waits for https://github.com/theforeman/foreman-packaging/pull/6335

packages/katello/rubygem-redhat_access/rubygem-redhat_access.spec
```diff
@@ -25,6 +27,7 @@ Requires: %{?scl_prefix}rubygem(redhat_access_lib) >= 1.1.5
 Requires: %{?scl_prefix}rubygem(angular-rails-templates) >= 0.0.4
 Requires: %{?scl_prefix}rubygem(foreman-tasks)
 Requires: %{?scl_prefix}rubygem(katello)
+Requires: %{?scl_prefix}rubygem(foreman_rh_cloud) >= 3.0.17
```
evgeni, line 30: Why? I don't see that dep in https://github.com/redhataccess/foreman-plugin/blob/master/redhat-access/redhat_access.gemspec

If that's intentional, please either add it to the gemspec, or place it outside the "generated dependencies" block

ShimShtein: Moved it out of the autogenerated block. I have added it manually anyway.

evgeni (approved):

PR #6348, opened 2021-02-24
Update facter to 4.0.51

lzap:
This should fix https://projects.theforeman.org/issues/31943

* Nightly

According to testing of the problem, I might actually file 2.4 PR. But first things first - nightly and I need to test this.


packages/foreman/rubygem-facter/rubygem-facter.spec
```diff
@@ -88,6 +88,9 @@ cp -a .%{_bindir}/* \
 %doc %{gem_docdir}
 
 %changelog
+* Wed Feb 24 2021 Lukas Zapletal <lzap+rpm@redhat.com> 4.0.51-1
+- Update to 5.0.51
```
do3meli, line 92: i guess here is a typo. that should be `4.0.51` and not `5.0.51` i assume.

do3meli (changes requested):

packages/foreman/rubygem-facter/rubygem-facter.spec
ekohl: For what it's worth, `./bump_rpm.sh packages/foreman/rubygem-facter` should generally do the right thing for you already. With gems it even knows how to pull the latest gem version.

lzap: Damn thanks, why I keep forgetting this script. In the past we had nothing like this and I am still used to the old workflow. Rebased.

ekohl: It's been there for 3 years now, time to adjust your habits :)

do3meli (approved):

ehelms (approved):

PR #6367, opened 2021-03-03
Support more version specifiers in update-requirements

ekohl:
>= is used for the NPM package @theforeman/builder by some packages.

### DIFF
--- a/update-requirements
+++ b/update-requirements
@@ -72,6 +72,9 @@ def get_requirement(version):
 
         return ['>= {}'.format(min_version), '< {}'.format(max_version)]
 
+    if version.startswith('>=') or version.startswith('<'):
+        return [version]
+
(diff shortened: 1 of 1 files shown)


update-requirements
```diff
@@ -72,6 +72,9 @@ def get_requirement(version):
 
         return ['>= {}'.format(min_version), '< {}'.format(max_version)]
 
+    if version.startswith('>=') or version.startswith('<'):
```
evgeni, line 75: why not also allow plain `>`? or because `1.0.0-1 > 1.0.0`?

ekohl: I think `>` is weird in package requirements. Exactly because of that example. IMHO you should always use `>= fixed.version` rather than `> broken.version`.

evgeni (approved):

PR #6369, opened 2021-03-03
Drop Pulp 2 repositories from katello-repos

ekohl:
Now that Pulp 2 is removed, there's no reason to ship these anymore.

### DIFF
--- a/packages/katello/katello-repos/katello-repos.spec
+++ b/packages/katello/katello-repos/katello-repos.spec
@@ -1,12 +1,3 @@
-%global pulp_release stable
-%global pulp_version 2.21
-%global use_pulp_nightly false
-%if 0%{?rhel} == 7
-%global pulp_enabled 1
-%else
-%global pulp_enabled 0
-%endif
-
 %global pulpcore_version 3.9
 
 %define repo_dir %{_sysconfdir}/yum.repos.d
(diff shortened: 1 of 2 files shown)


ekohl:
@evgeni is the qpid copr still needed or were the packages introduced in EPEL again?

evgeni:
le copr can be gone, epel is good.

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -1,20 +1,11 @@
...
 %define repo_dir %{_sysconfdir}/yum.repos.d
 %define repo_dist %{dist}
 
 %global prereleasesource nightly
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 3
+%global release 4
 
 Name:           katello-repos
 Version:        4.0
```
evgeni, line 11: shouldn't nightly be at 4.1 now?

evgeni (approved):

packages/katello/katello-repos/katello-repos.spec
ekohl: Can I blame a lack of branching procedures in packaging? Looks like there's been an rc release in the 2.4 branch so I'll first submit a PR to bump to 4.1 and then rebase this.

ekohl:
I opened https://github.com/theforeman/foreman-packaging/pull/6370 which I think is a replacement for this since it includes the change.

PR #6371, opened 2021-03-03
Debian packaging of all azure-rm related gems

eLvErDe:
Hello,

Please find attached packaging for all azure-rm ruby gems for Debian Buster.
Sadly, after installing these package and running foreman-installer again to enable azure-rm plugin it does not show in Foreman interface.

Any help would be appreciated...

Regards, Adam


eLvErDe:
I'm pretty sure the issue is in ruby-foreman-azure-rm package, I probably need to install app/config/db/locale files somewhere but if I go for /usr/share/foreman I get conflicts with already installed files.
I have trouble understanding where the files are installed in the RPM package. Can someone list me content of the RPM package ?

Regards, Adam.

ekohl:
Usually we include the gem instead of all the source files. https://github.com/theforeman/foreman-packaging/tree/deb/develop/plugins/ruby-foreman-fog-proxmox is such an example. Note that it lists a dependency here:
https://github.com/theforeman/foreman-packaging/blob/2a144e46e0bb3f89d81f2c5e75c33dc56286a106/plugins/ruby-foreman-fog-proxmox/foreman_fog_proxmox.rb#L2

And then for the build process here:
https://github.com/theforeman/foreman-packaging/blob/2a144e46e0bb3f89d81f2c5e75c33dc56286a106/plugins/ruby-foreman-fog-proxmox/debian/gem.list#L3

Perhaps that would work?

PR #6374, opened 2021-03-05
Drop Pulp 2 & qpid repositories from katello-repos

ekohl:
Now that Pulp 2 is removed, there's no reason to ship these anymore.

qpid's copr repository is also no longer needed. In
49694ccac9bc3bf8fc5a70d4d2eff748dd04c446 it was added since the
packages were removed from EPEL-7 but they have been readded.

Pulpcore repositories are always GPG signed, unlike Katello nightly.
This means that it should also always be checked.

(cherry picked from commit 1c350d6f9a005b07a785b5c07cccb2eb96a4f132)

### DIFF
--- a/packages/katello/katello-repos/katello-repos.spec
+++ b/packages/katello/katello-repos/katello-repos.spec
@@ -1,12 +1,3 @@
-%global pulp_release stable
-%global pulp_version 2.21
-%global use_pulp_nightly false
-%if 0%{?rhel} == 7
-%global pulp_enabled 1
-%else
-%global pulp_enabled 0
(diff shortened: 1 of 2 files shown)


jlsherrill:
needs a changelog entry for katello-repos, otherwise looks good

ekohl:
Ah, I messed up the version number in the conflict resolution.

ekohl:
This is now green and IMHO should be merged before the next Katello 4.0 RC (just like https://github.com/theforeman/foreman-packaging/pull/6408).

PR #6392, opened 2021-03-08
Bump rubygem-foreman_rh_cloud to 3.0.18-1

laviro:


ehelms:
[test rpm]

ehelms:
Looks like to get a version of this into nightly, you are gonna need to relax https://github.com/theforeman/foreman_rh_cloud/blob/master/foreman_rh_cloud.gemspec#L24

laviro:
Thanks @ehelms, the issue with `foreman-tasks >= 4` is that it requires `foreman > 2.4` while our plugin supports also `foreman 2.3`, we could adjust our CI to clone the version of the correct foreman-task, although if someone would use our plugin it will automatically download tasks gem > 4, so I think the only option is to add another branch which will work with current `develop`...

@shimshtein any other thoughts?

ehelms:
Compounded, is that in nightly, we can only have one version. And as foreman_rh_cloud requires Katello, and Katello in nightly requires foreman-tasks > 4.0, you won't be able to release a new version into nightly until you've either made that change to use `> 4.0` or somehow made it so you can run with either 4.0 or less than 4.0. Katello had to make a code change to support foreman-tasks 4.0. You might be more lucky and can relax that upper bound?

ehelms:
The change Katello had to make appears to all be test related: https://github.com/Katello/katello/pull/9205

laviro:
updated with our plugin's v3.0.18.1 that contains two BZs fixes and no versions limit required for foreman-tasks

ehelms (approved):

PR #6399, opened 2021-03-09
Foreman Azure-RM plugins packaging for Debian based systems

eLvErDe:


evgeni:
ok to test

eLvErDe:
Hi guys,

Any news about this ?

plugins/ruby-foreman-azure-rm/debian/changelog
```diff
@@ -0,0 +1,5 @@
+ruby-foreman-azure-rm (2.2.0) unstable; urgency=medium
```
evgeni, line 1:
```suggestion
ruby-foreman-azure-rm (2.2.0-1) unstable; urgency=medium
```

We've been using 3.0(native) and yet with debian revisions, to indicate changes to packaging.
Would you mind conforming tot that? :)

plugins/ruby-foreman-azure-rm/debian/rules
```diff
@@ -0,0 +1,40 @@
...
+	  . debian/gem.list; \
+	  mkdir cache/; \
+	  cd cache/; \
+	  for GEM_URL in $${GEMS}; do \
+	    echo "INFO: Downloading $${GEM_URL}"; \
+	    wget --no-verbose --no-clobber $${GEM_URL}; \
+	  done; \
+	fi
+
+build: cache
```
evgeni, line 17: am I reading this right, you essentially duplicated what our jenkins does in https://github.com/theforeman/foreman-packaging/blob/deb/develop/plugins/download_gems here as a make target, and it's only really executed if the `cache` folder is not present, which it should be when run on jenkins?

wonder if it would be better to call that script instead? for consistency.

evgeni: hmm, no, looking at https://ci.theforeman.org/job/packaging_build_deb_plugin/720/arch=x86,label=debian,os=bionic/consoleFull it re-ran this as you clean `cache` in `clean`.

if you don't clean it, it'd work I think.

I'd really prefer our jenkins not downloading things twice (and technically you shouldn't have network access from inside the build-root)

eLvErDe: Hello,

Well this was basically needed to be able to run the package creation outside Jenkins.
If think you can safely remove it but I do think it's handy to be able to test the package building locally ;)

evgeni: I know, and that's why I proposed to just remove the cleaning of the cache from the clean step below, so the `cache` target becomes essentially noop on Jenkins, while still working for you locally.

evgeni, line 17:
```suggestion
build:
```

plugins/ruby-foreman-azure-rm/debian/rules
```diff
@@ -0,0 +1,40 @@
...
+		cp -rp $${GEM_PATH}/public/assets/$(PLUGIN) ./
+	[ -e apipie-cache ] || mkdir apipie-cache/
+	cp -rp /var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) ./apipie-cache/
+
+override_dh_auto_clean:
+	dh_auto_clean
+	rm -rf cache/
+	rm -rf apipie-cache/
+	rm -rf $(PLUGIN)/
+
```
evgeni, line 38:
```suggestion
```

evgeni (commented):
As 2.5 branching is near, and we wanted to have this in 2.5, I am going to commit changes to your branch to match our current other packages, and would like to discuss with you after that how we can make all packages better approachable outside our custom build infra.

eLvErDe:
Sure, the reason why I removed the debian revision is the same as the clean stuff. Debian tools wont let me build the package, saying it's forbidden to use a debian revision for a native package and I did not found a way to skip this verification...

Btw, the ruby gem has been bumped, to include my work to allow custom disk size and windows machine deployment, can you jump it as well ?

evgeni:
To version 2.2.2? Sure!

eLvErDe:
2.2.1 for me but I see 2.2.2 has been released now with newer ruby version support

plugins/ruby-foreman-azure-rm/debian/changelog
```diff
@@ -0,0 +1,5 @@
+ruby-foreman-azure-rm (2.2.0-1) unstable; urgency=medium
```
evgeni, line 1:
```suggestion
ruby-foreman-azure-rm (2.2.2-1) unstable; urgency=medium
```

plugins/ruby-foreman-azure-rm/debian/gem.list
```diff
@@ -0,0 +1,11 @@
+# list of gem urls to download via Jenkins, space separated
+GEMS="https://rubygems.org/downloads/foreman_azure_rm-2.2.0.gem"
```
evgeni, line 2:
```suggestion
GEMS="https://rubygems.org/downloads/foreman_azure_rm-2.2.2.gem"
```

plugins/ruby-foreman-azure-rm/foreman_azure_rm.rb
```diff
@@ -0,0 +1 @@
+gem 'foreman_azure_rm', '2.2.0'
```
evgeni, line 1:
```suggestion
gem 'foreman_azure_rm', '2.2.2'
```

evgeni (commented):
2.2.2 update

eLvErDe:
Do you want me to do something, like squashing the PR or it's ok for you ?

evgeni:
Nah, I'll squash on merge, thanks!

PR #6400, opened 2021-03-10
Drop unused packages syntax, hoe and rubyforge

ehelms:
As far as I can tell grepping through the repository and checking installs these 3 packages are not used by anything.


ekohl (approved):
It looks like the need for `syntax` was removed in e20236569946cab2e8e80a2660a8cd14d6053d04 (dated in 2014)

The need for `hoe` was removed in 825d56cbb75b70ad4dad6ec1892b10da81add111. Once you drop `hoe`, you can also drop `rubyforge`. Interesting cyclic dependency there.

Approving here but leaving it open so that the proper removal dance can be done.

PR #6401, opened 2021-03-10

ekohl:


ekohl:
It fails in the same way as nightly.

ehelms:
[test rpm]

ehelms (approved):

ekohl:
Updated to include https://github.com/theforeman/foreman/pull/8387. IMHO that doesn't have to merged before this is. It's ok to be stricter in packaging.

PR #6413, opened 2021-03-11
Bump rubygem-get_process_mem to 0.2.7-1

ehelms:


ehelms:
Can be built as an update now, will break under Ruby 2.7 with the current version.

packages/foreman/rubygem-get_process_mem/rubygem-get_process_mem.spec
```diff
@@ -59,15 +65,16 @@ gem build %{gem_name}.gemspec
...
-cp -pa .%{gem_dir}/* \
+cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
 %files
 %dir %{gem_instdir}
+%{gem_instdir}/.github
+%exclude %{gem_instdir}/.gitignore
+%exclude %{gem_instdir}/.travis.yml
+%{gem_instdir}/appveyor.yml
```
ekohl, line 76: This can still be excluded since it's CI config.

ekohl (changes requested):

packages/foreman/rubygem-get_process_mem/rubygem-get_process_mem.spec
ehelms: To help ensure this systemically -- https://github.com/theforeman/foreman-packaging/pull/6414

pcreech (approved):
ACK

PR #6434, opened 2021-03-15
Bump ACD and Smart Proxy for ACD

sbernhard:


packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
```diff
@@ -127,6 +129,9 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/test
 
 %changelog
+* Mon Mar 15 2021 Bernhard Suttner,,, 0.6.0-1
```
evgeni, line 132:
```suggestion
* Mon Mar 15 2021 Bernhard Suttner <suttner@atix.de> 0.6.0-1
```

evgeni (changes requested):
can you please fix the changelog entry? it's good to go then.

rel-eng/tito.props
```diff
@@ -757,6 +757,8 @@ whitelist = ansiblerole-foreman_scap_client
   rubygem-ruby-nmap
   rubygem-safe_yaml
   rubygem-sentry-raven
+  rubygem-smart_proxy_acd
+  rubygem-smart_proxy_acd_core
```
evgeni, line 761: you also need to add these to foreman-plugins-2.4-rhel7

evgeni (changes requested):
oh, and tito

sbernhard:
I guess, I missed the flight for 2.4.0 :-(

evgeni:
> I guess, I missed the flight for 2.4.0 :-(

Luckily, for you, plugins are released independently, so you only missed being there on day 0.

evgeni (approved):

PR #6435, opened 2021-03-15
Update sequel to 5.42.0

ehelms:


ehelms:
This update can be taken now, it helps with Ruby 2.7 effort as the version we currently have breaks against Ruby 2.7

packages/foreman/rubygem-sequel/rubygem-sequel.spec
```diff
@@ -80,11 +84,15 @@ find %{buildroot}%{gem_instdir}/bin -type f | xargs chmod a+x
...
 %doc %{gem_instdir}/doc
-%{gem_instdir}/spec
 
 %changelog
+* Mon Mar 15 2021 Eric D. Helms <ericdhelms@gmail.com> 5.42.0-1
+- Update to 5.42.0-1
+
+* Mon Mar 15 2021 Eric D. Helms <ericdhelms@gmail.com> 5.42.0-1
+- Update to 5.42.0
+
```
evgeni, line 95: I think just one entry is sufficient

evgeni (approved):

PR #6448, opened 2021-03-18
Update shellhooks to 0.9.2 (RPM)

lzap:
This package was never part of 2.4 but we would like to ship the very first version of webhooks-shellhooks for 2.4.

Thanks.


lzap:
Rebased, I need to cherry pick both commits in order to have the whitelists and other obal stuff.

evgeni (approved):

PR #6447, opened 2021-03-17
Update foreman-fog-proxmox to 0.13.1

tristanrobert:
* Nightly

PR #6454, opened 2021-03-18
Update foreman-fog-proxmox 0.13.2

tristanrobert:
* Nightly
fixes #6447


evgeni:
```
12:30:19 Uglifier::Error: Unexpected token: name (volume_id). To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true).
```

Seems `id` was not the thing it didn't like :(

tristanrobert:
> ```
> 12:30:19 Uglifier::Error: Unexpected token: name (volume_id). To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true).
> ```
> 
> Seems `id` was not the thing it didn't like :(

It is ES6 syntax that is not accepted. I have fixed it in #6457

PR #6459, opened 2021-03-18
Update rubygem-foreman_openscap to 4.0.6

xprazak2:


evgeni:
2.1 is unsupported by now and we're not publishing any updates to it.

PR #6467, opened 2021-03-19
Drop qpid install requires from Katello

ehelms:
This drops qpid package requires that are handled by the puppet
modules when the functionality that requires them is enabled. Drops
the qpid-core-dump script as it was targeted at the Pulp 2 celery
usage of Qpid.


ehelms:
This should not be merged without https://github.com/theforeman/puppet-qpid/pull/149 having been merged first. Otherwise the cyrus-sasl-plain package will be missing and installation will fail.

wbclark (commented):
@ehelms this needs a rebase when you have a moment

ehelms:
> @ehelms this needs a rebase when you have a moment

Done

ekohl (approved):

PR #6469, opened 2021-03-20
Update foreman-release to 2.4.0

pcreech:


packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -47,7 +45,7 @@ Requires: which
 
 # puppet-candlepin enables the pki-core module which needs DNF module support
 # That was introduced in 5.5.20 and 6.15.0
-%if 0%{rhel} == 8
+%if 0%{?rhel} == 8
```
pcreech, line 48: This was causing obal to die during the changelog generation phase of `obal update`.
```
    error: parse error in expression:  0%{rhel} == 8
    error:                              ^
    error: /home/pcreech/devel/theforeman/foreman-packaging/packages/foreman/foreman-installer/foreman-installer.spec:48: bad %if condition:  0%{rhel} == 8
    error: query of specfile /home/pcreech/devel/theforeman/foreman-packaging/packages/foreman/foreman-installer/foreman-installer.spec failed, can't parse
{snip}
subprocess.CalledProcessError: Command 'rpmspec --query --queryformat %{evr} --srpm /home/pcreech/devel/theforeman/foreman-packaging/packages/foreman/foreman-installer/foreman-installer.spec --undefine dist' returned non-zero exit status 1.
```

pcreech: My assumption being that rhel doesn't exist on fedora.

evgeni: correct. can you open a change for that in develop too?

evgeni (approved):

PR #6494, opened 2021-03-29
Fixes #29779: Drop passenger packages

ehelms:


comps/comps-foreman-rhel7.xml
```diff
@@ -240,9 +239,6 @@
       <packagereq type="default">tfm-rubygem-ovirt-engine-sdk</packagereq>
       <packagereq type="default">tfm-rubygem-paint</packagereq>
       <packagereq type="default">tfm-rubygem-parallel</packagereq>
-      <packagereq type="default">tfm-rubygem-passenger</packagereq>
-      <packagereq type="default">tfm-rubygem-passenger-native</packagereq>
-      <packagereq type="default">tfm-rubygem-passenger-native-libs</packagereq>
```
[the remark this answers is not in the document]
ehelms: I've included the addition of these to the Obsoletes in tfm.

packages/foreman/tfm/tfm.spec
```diff
@@ -131,6 +131,12 @@ Obsoletes: tfm-ror52-rubygem-turbolinks-doc < 2.5.4-3
 Obsoletes: tfm-ror52-rubygem-arel <= 9.0.0
 Obsoletes: %{scl_prefix}-rubygem-arel <= 9.0.0
 
+# obsolete all of the tfm Passenger packages since Passenger support is dropped
```
ekohl, line 134: I think it should also do the same with doc packages but not entirely sure.

ehelms: Done.

ekohl (approved):
Code looks good. Not merging yet since that's probably good to coordinate a bit.

PR #6497, opened 2021-03-30
Update rubygem-smart_proxy_pulp to 3.0.0

ehelms:

### DIFF
--- a/gem2rpm/smart_proxy_plugin.spec.erb
+++ b/gem2rpm/smart_proxy_plugin.spec.erb
@@ -5,6 +5,10 @@ root = `git rev-parse --show-toplevel`.strip
 SCL_PREFIXES = JSON.load(File.read(File.join(root, 'scl_prefixes.json')))
 LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
 has_cronjob = spec.files.any? { |f| f =~ %r{\Aextra/.*\.cron\z } }
+
+config.rules[:ignore] << 'bundler.d'
+config.rules[:ignore] << 'settings.d'
+config.rules[:ignore] << 'config'
 -%>
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
--- a/packages/plugins/rubygem-smart_proxy_pulp/rubygem-smart_proxy_pulp.spec
+++ b/packages/plugins/rubygem-smart_proxy_pulp/rubygem-smart_proxy_pulp.spec
@@ -2,6 +2,7 @@
 %{?scl:%scl_package rubygem-%{gem_name}}
 %{!?scl:%global pkg_name %{name}}
 
+%{!?_root_bindir:%global _root_bindir %{_bindir}}
 %{!?_root_datadir:%global _root_datadir %{_datadir}}
 %{!?_root_localstatedir:%global _root_localstatedir %{_localstatedir}}
 %{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
@@ -9,7 +10,7 @@
 %global gem_name smart_proxy_pulp
 %global plugin_name pulp
 
-%global foreman_proxy_min_version 1.25
+%global foreman_proxy_min_version 2.3
 %global foreman_proxy_dir %{_root_datadir}/foreman-proxy
 %global foreman_proxy_statedir %{_root_localstatedir}/lib/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
@@ -17,8 +18,8 @@
 %global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 2.1.0
-Release: 4%{?foremandist}%{?dist}
+Version: 3.0.0
+Release: 1%{?foremandist}%{?dist}
 Summary: Basic Pulp support for Foreman Smart-Proxy
 Group: Applications/Internet
 License: GPLv3
@@ -28,10 +29,12 @@ Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 # start specfile generated dependencies
 Requires: foreman-proxy >= %{foreman_proxy_min_version}
 Requires: %{?scl_prefix_ruby}ruby(release)
-Requires: %{?scl_prefix_ruby}ruby
+Requires: %{?scl_prefix_ruby}ruby >= 2.5
+Requires: %{?scl_prefix_ruby}ruby < 3
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
 BuildRequires: %{?scl_prefix_ruby}ruby(release)
-BuildRequires: %{?scl_prefix_ruby}ruby
+BuildRequires: %{?scl_prefix_ruby}ruby >= 2.5
+BuildRequires: %{?scl_prefix_ruby}ruby < 3
 BuildRequires: %{?scl_prefix_ruby}rubygems-devel
 BuildArch: noarch
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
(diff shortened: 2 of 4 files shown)


ehelms:
Once this goes in https://github.com/theforeman/puppet-foreman_proxy_content/pull/339 and https://github.com/theforeman/puppet-foreman_proxy/pull/638 need to go in afterwards. Running all of them together in a pipeline passed for me.

packages/plugins/rubygem-smart_proxy_pulp/rubygem-smart_proxy_pulp.spec
```diff
@@ -2,23 +2,24 @@
...
+%{!?_root_bindir:%global _root_bindir %{_bindir}}
 %{!?_root_datadir:%global _root_datadir %{_datadir}}
 %{!?_root_localstatedir:%global _root_localstatedir %{_localstatedir}}
 %{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
 
 %global gem_name smart_proxy_pulp
 %global plugin_name pulp
 
-%global foreman_proxy_min_version 1.25
+%global foreman_proxy_min_version 1.24
```
ekohl, line 13: I don't think this is right. It is actually 2.3 now since it uses `rackup_path` introduced there:
```suggestion
%global foreman_proxy_min_version 2.3
```

packages/plugins/rubygem-smart_proxy_pulp/rubygem-smart_proxy_pulp.spec
```diff
@@ -90,20 +93,16 @@ mv %{buildroot}%{gem_instdir}/bundler.d/%{plugin_name}.rb \
...
 
 %files
 %dir %{gem_instdir}
-%config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/pulp.yml
 %config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/pulpcore.yml
-%config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/pulpnode.yml
 %license %{gem_instdir}/LICENSE
+%{gem_instdir}/bundler.d
 %{gem_libdir}
+%{gem_instdir}/settings.d
```
ekohl, line 105: Why is this directory needed? There should be no `settings.d` directory in `%{gem_instdir}`

packages/plugins/rubygem-smart_proxy_pulp/rubygem-smart_proxy_pulp.spec
```diff
@@ -90,20 +93,16 @@ mv %{buildroot}%{gem_instdir}/bundler.d/%{plugin_name}.rb \
...
-mv %{buildroot}%{gem_instdir}/settings.d/pulpnode.yml.example \
-   %{buildroot}%{foreman_proxy_settingsd_dir}/pulpnode.yml
 
 %files
 %dir %{gem_instdir}
-%config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/pulp.yml
 %config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/pulpcore.yml
-%config(noreplace) %attr(0640, root, foreman-proxy) %{foreman_proxy_settingsd_dir}/pulpnode.yml
 %license %{gem_instdir}/LICENSE
+%{gem_instdir}/bundler.d
```
ekohl, line 103: Why is this directory needed? There should be no `bundler.d` directory in `%{gem_instdir}`

ekohl (changes requested):

packages/plugins/rubygem-smart_proxy_pulp/rubygem-smart_proxy_pulp.spec
ehelms: Thats the default in the gem2rpm template, I'll adjust this plugin. (Fixing the template if we want is a separate issue)

packages/plugins/rubygem-smart_proxy_pulp/rubygem-smart_proxy_pulp.spec
ehelms: Good question, gemr2rpm added these from the template.

ehelms: Ahh, cause they exist in the gem itself and are packaged inside the gem e.g. https://github.com/theforeman/smart_proxy_pulp/tree/master/settings.d 

gem2rpm automatically added them

gem2rpm/smart_proxy_plugin.spec.erb
```diff
@@ -5,6 +5,9 @@ root = `git rev-parse --show-toplevel`.strip
 SCL_PREFIXES = JSON.load(File.read(File.join(root, 'scl_prefixes.json')))
 LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
 has_cronjob = spec.files.any? { |f| f =~ %r{\Aextra/.*\.cron\z } }
+
+config.rules[:ignore] << 'bundler.d'
+config.rules[:ignore] << 'settings.d'
```
ekohl, line 10: Rather than ignoring them, how about adding `rmdir` in `%install`. That way we know for sure we don't accidentally miss files when they're added (since the build will fail).

@evgeni?

evgeni: I am not sure that I follow the problem here.

An entry in the ignore list essentially generates an `%exclude`, right?

And `rmdir` would only work if the directory is *empty*, which it isn't, the plugin dropped its example config there?

ehelms: In theory they would be empty at the end of the `%install` section, because we perform a `mv` to copy both locations:

e.g. https://github.com/theforeman/foreman-packaging/pull/6497/files#diff-2fd47368d60a1da78f86557b3824e35a393f46a8ce13b5f596a8367064f93d02R96-R97

ekohl: What @ehelms said indeed. If we're not moving all files, we may be missing something. The `rmdir` should catch that.

evgeni: Okay, the move was what was missing from my view.

Yeah, then `rmdir` sounds nice

ehelms: This gets a bit more complicated. If I add the `rmdir` this does not prevent it from showing in the files list everytime you run gem2rpm. The macro used (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/gem2rpm/smart_proxy_plugin.spec.erb#L165) takes as input the files being reported from the gemspec not what exists in the buildroot. So, I would have to add a rmdir *and* update this chunk of code to explicitly ignore `bundler.d` and `settings.d` directories. That feels like more work than it should be when I can use built in ignores to just exclude directories.

Thoughts?

ekohl: Ironically, Pulp is the only plugin that I know of that used to ship multiple config files. Especially Pulp would benefit from `rmdir`. I still think that both is the right solution.

Also, there's code to support `config/` in plugins in the same way as `settings.d/` so I think it should be
```suggestion
config.rules[:ignore] << 'bundler.d'
config.rules[:ignore] << 'settings.d'
config.rules[:ignore] << 'config'
```

ehelms: I'm a bit confused by what the `rmdir` would accomplish. If you run `rmdir` inside `%install` section you are essentially just removing it from the buildroot which has no effect on the final set of files available and laid down by the RPM itself.

ekohl:
```console
$ mkdir example
$ touch example/file
$ rmdir example
rmdir: failed to remove 'example': Directory not empty
```

So if you add a file to settings.d but don't update the spec file, the mv won't be there and thus rmdir should fail. This means the build fails which is a good thing. Otherwise it would silently not work and only once you start testing you find out something is missing.

That is my reasoning. You get an early failure which makes iterating on changes faster.

ehelms: Now that makes sense, appreciate the explanation.

ehelms: And to complete the round trip of this, the example case that you are looking ahead to is accounted for already via the way the template is written (as I can parse it).

https://github.com/theforeman/foreman-packaging/blob/rpm/develop/gem2rpm/smart_proxy_plugin.spec.erb#L136-L146

This detects all files in `settings.d/` or `config` and generates statements to ensure they get moved, for example, this plugin when it had Pulp 2 support, the following is generated:

```
mkdir -p %{buildroot}%{foreman_proxy_settingsd_dir}
mv %{buildroot}%{gem_instdir}/settings.d/pulp.yml.example \
   %{buildroot}%{foreman_proxy_settingsd_dir}/pulp.yml
mv %{buildroot}%{gem_instdir}/settings.d/pulpcore.yml.example \
   %{buildroot}%{foreman_proxy_settingsd_dir}/pulpcore.yml
mv %{buildroot}%{gem_instdir}/settings.d/pulpnode.yml.example \
   %{buildroot}%{foreman_proxy_settingsd_dir}/pulpnode.yml
```

Two of those lines are removed in this PR because as part of updating it the spec file also got updated to match the new state of the gem.

ehelms: If there is more to do here, could I move this change to an issue that is stand-alone (since in theory we'd want to apply this change to all smart-proxy plugins) to get this particular update of the smart_proxy_pulp plugin built and allow me to additionally merge https://github.com/theforeman/puppet-foreman_proxy/pull/638 as those two things will effectively close out the removal of Pulp 2 from the installer.

evgeni: I am happy with the current changes

evgeni (approved):

ekohl (approved):
That's why I normally separate the changes into two PRs: one can have comments while the other is OK to merge.

ehelms:
> That's why I normally separate the changes into two PRs: one can have comments while the other is OK to merge.

That was just hard to coordinate in this case. I needed the template changes to be able to test and ensure they updated the spec correctly and re-test on the various changes.

PR #6498, opened 2021-03-30
Pulpcore 3.11 client bindings

jlsherrill:


jlsherrill:
needs to wait on https://github.com/Katello/katello/pull/9255

evgeni:
Needs rebase after Erics mass rebuild.

Why does it remove `rubygem-pulp_2to3_migration_client`? Because there will be no more migrations in 4.0→4.1?

evgeni:
Should something (f-m content remove-pulp2) also purge the package from systems then?

jlsherrill:
so the remove rake task will be run on 4.0, but this gem isn't actually removed until 4.1 (unless we backport the removal).  Would obsoleting it in the rubygem-katello rpm for a few releases be a bad idea?

jlsherrill:
also, we will need to keep shipping and updating the server component of the migration plugin until we have a 'removal process' from the pulp team.

evgeni:
Yeah, `Obsoletes` in `rubygem-katello` is probably not a great but a workable solution.

@ehelms @pcreech? What do y'all think?

pcreech:
I can't think of a better process currently.

ekohl:
I think traditionally we've used the `tfm` package to obsolete packages in the SCL. If it's only EL7 that could also be a solution. Wouldn't work on EL8 though.

evgeni:
It is on EL8 with Katello 4.0 ;)

ehelms:
Sounds like `rubyem-katello` it is.

jlsherrill:
Katello PR here: https://github.com/theforeman/foreman-packaging/pull/6517  

i think this is good to go?

evgeni:
With https://github.com/Katello/katello/pull/9255 now merged: yes.

Thanks!

evgeni (approved):

PR #6420, opened 2021-03-11
Rebuild foreman_scl_packages H-Q for Ruby 2.7

ehelms:

PR #6501, opened 2021-04-05
Rebuild rubygem-puma for Ruby 2.7

ehelms:


ehelms:
Pulled from https://github.com/theforeman/foreman-packaging/pull/6420/files since it requires rubygem-nior to be built first

ehelms:
[test rpm]

PR #6516, opened 2021-04-08
Drop katello meta requirement on hammer-cli-katello

ehelms:
Move the requirement and enablement to the installer -- https://github.com/theforeman/foreman-installer/pull/673

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,7 +5,7 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 4
+%global release 5
 
 Name:       katello
 Version:    4.1.0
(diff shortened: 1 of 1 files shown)


ehelms:
The foreman-installer change has been merged so I think it's safe to drop this and rely on the installer now to set this up.

packages/katello/katello/katello.spec
```diff
@@ -136,6 +133,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Thu Apr 08 2021 Eric D. Helms <ericdhelms@gmail.com> - 4.1.0-0.5.master
+- Drop requirement on hammer-cli-katello
```
ekohl, line 137: Shouldn't this be broader?
```suggestion
- Drop requirement on Hammer CLI packages
```

ekohl (approved):

packages/katello/katello/katello.spec
ehelms: Could be, the other hammer packages are redundantly declared given hammer-cli-katello already requires them.

ekohl: Sure, but as a user who reads the changelog I think this would be easier to understand. Before I got Hammer, now I don't.

PR #6525, opened 2021-04-12
build foreman-proxy on focal

evgeni:


debian/focal/foreman-proxy/foreman-proxy.tmpfile
```diff
@@ -0,0 +1 @@
+d /run/foreman-proxy 0755 foreman-proxy foreman-proxy -
```
ekohl, line 1: This makes me wonder if we should ship this file in the repo itself so it's shared between RPMs and debs. Not for this patch, but this is a good time to review those things.

ekohl (approved):

PR #6526, opened 2021-04-12
build installer for focal

evgeni:


debian/focal/foreman-installer/changelog
```diff
@@ -0,0 +1,77 @@
...
+
+foreman-installer (1.20.0-1) stable; urgency=low
+
+  * Bump changelog to 1.20.0 to match VERSION
+
+ -- Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl>  Tue, 17 Jul 2018 18:26:17 +0200
+
+foreman-installer (1.19.0-1) stable; urgency=low
+
+  * Initial release for Ubuntu/bionic
```
ekohl, line 75: Looking at this, we didn't copy it in the past. I think you mentioned that you'd prefer to end up with the same directories where possible. Is this still your intention and why you copied the changelog too?

evgeni: Yeah, one day I want to get rid of the (now three) two folders.

evgeni:
[test packaging]

evgeni:
[test packaging]

evgeni:
Green!

ekohl (approved):

PR #6531, opened 2021-04-12
build foreman for focal

evgeni:


debian/focal/foreman/control
```diff
@@ -0,0 +1,229 @@
+Source: foreman
+Maintainer: Greg Sutcliffe <greg.sutcliffe@gmail.com>
+Section: web
+Priority: extra
+Standards-Version: 3.9.1
+Build-Depends: debhelper (>= 9.20160709), bundler (>= 1.3), ruby2.5-dev, git | git-core,
```
evgeni, line 6: heh…

ekohl (commented):
I think it makes sense to have a Redmine issue for the unversioned Ruby use so it's tracked in the changelog. Then I'm ok with it.

ekohl (approved):

evgeni:
If only it would build…

```
12:09:35 /usr/bin/bundle exec rake locale:pack RAILS_ENV=production
12:09:54 rake aborted!
12:09:54 TypeError: superclass mismatch for class Command
12:09:54 /var/lib/gems/2.7.0/gems/thor-1.1.0/lib/thor/command.rb:2:in `<class:Thor>'
```

ekohl:
That does sound vaguely familiar. The internet shows https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=955405 and https://github.com/erikhuda/thor/issues/721

evgeni:
Yes, that's also what I have open right now.

I think the easiest path for us would be to build thor 1.0.1 or 1.1.0 as a package, so that it matches what we vendor

ekohl:
https://bugs.launchpad.net/ubuntu/+source/ruby-thor/+bug/1885424 suggests that too so :+1: for that.

evgeni:
[test packaging]

evgeni:
nope, still `TypeError: superclass mismatch for class Command`

evgeni:
[test packaging]

evgeni:
This is finally green (or will be once the other targets finish)

ekohl (approved):

PR #6537, opened 2021-04-13
add thor

evgeni:


ekohl (approved):
Looks like at least Bullseye has 1.0.1 (https://packages.debian.org/bullseye/ruby-thor) so it shouldn't be a problem there.

evgeni:
I'll wait with the actual merge until a run of #6531 with this scratch build is done.

evgeni:
FWIW, the real difference is whether `thor` is built into `/usr/lib/ruby/vendor_ruby/` or `/usr/share/rubygems-integration`, the later is triggered by `export DH_RUBY = --gem-install` in `debian/rules` and is the place where it does not end up conflicting with the world.

PR #6539, opened 2021-04-13
Update Katello's pulp_rpm_client requirement to 3.10

ianballou:


jlsherrill:
[test rpm]

jlsherrill:
i don't think this will pass until an updated gem is released?

PR #6546, opened 2021-04-16
rebuild discovery, puppet, statistics plugins for Ruby 2.7

evgeni:


evgeni:
dropped f_templates as it doesn't build for *other* reasons

ehelms (approved):

PR #6548, opened 2021-04-16
Require ansible-collection-theforeman-foreman in smart_proxy_ansible

evgeni:
this should allow it to use the callback we ship in the collection


packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -34,6 +34,7 @@ Requires: python3-requests
 
 Requires: %{?scl_prefix}rubygem(smart_proxy_dynflow_core) >= 0.1.5
 Requires: %{?scl_prefix}rubygem(foreman_ansible_core)
+Requires: ansible-collection-theforeman-foreman
```
ekohl, line 37: Does this mean we can drop the dependency on python(3)-requests and let the collection pull it in?

evgeni: indeed, it does.

ekohl: Can you drop lines 29-33 then?

ekohl (approved):

PR #6549, opened 2021-04-19
Accept python3-requests for the callback too.

evgeni:

### DIFF
--- a/plugins/smart_proxy_ansible/changelog
+++ b/plugins/smart_proxy_ansible/changelog
@@ -1,3 +1,9 @@
+ruby-smart-proxy-ansible (3.0.0-5) stable; urgency=low
+
+  * Install python3-requests for the callback too.
+
+ -- Evgeni Golov <evgeni@debian.org>  Mon, 19 Apr 2021 11:56:00 +0200
+
 ruby-smart-proxy-ansible (3.0.0-4) stable; urgency=low
 
   * Create workdir for Ansible on the Proxy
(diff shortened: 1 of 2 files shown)


ekohl (commented):
Just checking what uses what for Ansible:
* Debian Buster uses Python 3 (https://packages.debian.org/buster/ansible)
* Ubuntu Bionic uses Python 2 (https://packages.ubuntu.com/bionic/ansible)
* Ubuntu Focal uses Python 3 (https://packages.ubuntu.com/focal/ansible)

Given that, should it prefer the 3 version over 2?

evgeni:
Well… That'll pull in the wrong requests on Bionic then, while the current one pulls in the wrong one on Buster…

ekohl:
I also thought about that. Here is an example where a common plugins repository does break down.

evgeni:
Yes.

ekohl:
In control files, can you specify `(ansible >= 2.6 && python3-requests) || (ansible >= 2.2 && python-requests)`? This would rely on the fact that Bionic ships Ansible 2.5 while Buster ships 2.7. Still a hack, but if that works ...

evgeni:
No you can't. You could do `python-requests | ansible (>= 2.6), python3-requests` which is similar-ish. But that'd "break" (as in not force pulling what we want it to) with backports which people probably do for Ansible heavily.

ekohl:
I was afraid of that. This also has the problem that on a Bionic system where `python3-requests` is already installed that the dependency will be missing. Prior to https://github.com/theforeman/puppet-foreman_proxy/commit/d1954c714a912af0c098357c3684d8635fb234a3#diff-490f5830bc142ccd235c22d610cdedc2ac99eef2bc8dde306912f3c5b366ae8f it could have been handled in the installer.

ekohl:
Sounds good to me.

evgeni:
Updated!

ekohl (approved):
Might be nice to explain this in a commit message, but otherwise :+1:

evgeni:
Updated the commit message.

ekohl (approved):
Thanks!

PR #6558, opened 2021-04-23
Update webhooks to 1.0.0

lzap:

* Nightly
* 2.4

Will do CP PR


zjhuntin (approved):
ACK

PR #6563, opened 2021-04-27
Update rubygem-foreman_discovery to 16.3.5

lzap:

* Nightly

Only nightly.

https://bugzilla.redhat.com/show_bug.cgi?id=1950929


packages/plugins/rubygem-foreman_discovery/rubygem-foreman_discovery.spec
```diff
@@ -33,7 +33,8 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
 # start package.json devDependencies BuildRequires
 BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
 BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
-BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 6.0.0
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 4.14.0
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) < 5.0.0
```
evgeni, line 37: that won't work

evgeni (approved):

packages/plugins/rubygem-foreman_discovery/rubygem-foreman_discovery.spec
lzap: I am assuming I should just delete.

This is actually for downstream only, 2.3 branch is the target one. I am going to release new version this week anyway.

lzap:
For the record, I only want to make sure it builds fine into nightly so it can go into Satellite 6.9. The bug is probably not worth fixing in 2.3.

I am going to release Discovery for 2.4 this week (I haven't done that yet :-)

PR #6566, opened 2021-04-27
Update rubygem-pulp_deb_client to 2.11.1

quba42:
The relevant pulp_deb version is already available in pulpcore-packaging:

https://yum.theforeman.org/pulpcore/3.11/el7/x86_64/python3-pulp-deb-2.11.1-1.el7.noarch.rpm

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -9,7 +9,7 @@
 %global prereleasesource pre.master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global mainver 4.1.0
-%global release 8
+%global release 9
 
 Name:    %{?scl_prefix}rubygem-%{gem_name}
 Summary: Content and Subscription Management plugin for Foreman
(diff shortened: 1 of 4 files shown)


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -56,8 +56,8 @@ Requires: %{?scl_prefix}rubygem(pulp_ansible_client) >= 0.7
 Requires: %{?scl_prefix}rubygem(pulp_ansible_client) < 0.8
 Requires: %{?scl_prefix}rubygem(pulp_container_client) >= 2.4.0
 Requires: %{?scl_prefix}rubygem(pulp_container_client) < 2.5.0
-Requires: %{?scl_prefix}rubygem(pulp_deb_client) >= 2.10.0
-Requires: %{?scl_prefix}rubygem(pulp_deb_client) < 2.11.0
+Requires: %{?scl_prefix}rubygem(pulp_deb_client) >= 2.11.0
```
quba42, line 59: I am near certain the API bindings have not actually changed from `2.10.0`, should I nevertheless bump this minimum version to the new version, or not?

evgeni: this is generated from https://github.com/Katello/katello/blob/893cc6dd265754e865c1c93a3a6df73cf9a4a736/katello.gemspec#L59, so you'll need to change it there, and then it will be (semi) automatically updated here.

quba42: So, like this: https://github.com/Katello/katello/pull/9317
What needs to be merged first?

quba42:
As I now understand it, this should be merged *after* this change (https://github.com/Katello/katello/pull/9317) is merged, and then makes it into this repo via some kind of automation?

evgeni:
yeah, usually we update deps based on the gemspecs every once in a while

jlsherrill:
@quba42 mind updating https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello/katello.spec too?  

We can probably do it as part of this PR since you're just raising the ceiling

quba42:
> @quba42 mind updating https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello/katello.spec too?

Not sure what needs changing in `packages/katello/katello/katello.spec`, I am only aware of changes needed in `packages/katello/rubygem-katello/rubygem-katello.spec`. But I am not sure how those are meant to be performed (I understand there is automation for it).

quba42:
Ok, I think the second commit now brings this in line with https://github.com/Katello/katello/pull/9317

jlsherrill (approved):

PR #6567, opened 2021-04-27
Add support for foreman-puma-status command

evgeni:

### DIFF
--- a/debian/bionic/foreman/rules
+++ b/debian/bionic/foreman/rules
@@ -14,6 +14,7 @@ build:
 	/usr/bin/gem install -f minitest -v '5.10.3'
 	/usr/bin/gem install -f thor -v '1.1.0'
 	/bin/rm -f bundler.d/test.rb bundler.d/development.rb
+	/bin/sed -i '2i  gem "puma-status"' bundler.d/service.rb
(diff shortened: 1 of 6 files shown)


ehelms (approved):

debian/bionic/foreman/rules
```diff
@@ -24,6 +25,7 @@ build:
 	/bin/rm -f db/schema.rb
 	/bin/sed -ri '1sXenv rubyXenv foreman-rubyX' bin/*
 	/bin/sed -ri 's~^BUNDLER_CMD=""~BUNDLER_CMD="/usr/bin/foreman-ruby /usr/bin/bundle exec"~' script/foreman-rake
+	/bin/sed -i 's#puma-status#cd /usr/share/foreman \&\& /usr/bin/bundle exec &#' script/foreman-puma-status
```
ekohl, line 28: Why is `&&` escaped? but `&` isn't? Is that some sed magic?

evgeni: `&` means "the match" in a regex, so `s/foo/bar &/` replaces "foo" with "bar foo".

and in the middle I want the literal `&&` for the shell. but also see https://github.com/theforeman/foreman/pull/8472#issuecomment-827548229

PR #6581, opened 2021-04-30
Add rubygem-racc

ehelms:


ekohl (approved):

ehelms:
This now includes some additional exclude and updates the gem2rpm template

packages/foreman/rubygem-racc/rubygem-racc.spec
```diff
@@ -0,0 +1,124 @@
...
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+# start specfile generated dependencies
+Requires: %{?scl_prefix_ruby}ruby(release)
+Requires: %{?scl_prefix_ruby}ruby
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
+BuildRequires: %{?scl_prefix_ruby}ruby(release)
+BuildRequires: %{?scl_prefix_ruby}ruby-devel
+BuildRequires: %{?scl_prefix_ruby}rubygems-devel
+BuildRequires: tree
```
ekohl, line 24: How does it generate this dependency? Feels odd.

packages/foreman/rubygem-racc/rubygem-racc.spec
```diff
@@ -0,0 +1,124 @@
...
+%dir %{gem_instdir}
+%{_bindir}/racc
+%exclude %{gem_instdir}/ext
+%{gem_extdir_mri}
+%license %{gem_instdir}/COPYING
+%exclude %{gem_instdir}/TODO
+%{gem_instdir}/bin
+%{gem_libdir}
+%exclude %{gem_instdir}/sample
+%exclude %{gem_instdir}/web
```
ekohl, line 108: In https://src.fedoraproject.org/rpms/rubygem-racc/blob/rawhide/f/rubygem-racc.spec they ship it in the doc subpackage

gem2rpm/scl.spec.erb
```diff
@@ -12,6 +12,9 @@ config.rules[:ignore] << '.document'
 config.rules[:ignore] << 'Appraisals'
 config.rules[:ignore] << 'gemfiles'
 config.rules[:ignore] << 'appveyor.yml'
+config.rules[:ignore] << 'TODO'
+config.rules[:ignore] << 'sample'
+config.rules[:ignore] << 'web'
```
ekohl, line 17: In https://src.fedoraproject.org/rpms/rubygem-racc/blob/rawhide/f/rubygem-racc.spec they're shipped in the doc pacakge. I'm not sure we should ignore them.

ekohl (commented):
https://src.fedoraproject.org/rpms/rubygem-racc/blob/rawhide/f/rubygem-racc.spec also has a build requires on gcc. I wonder why it's not needed here.

packages/foreman/rubygem-racc/rubygem-racc.spec
ehelms: Oops, that was from my debugging.

gem2rpm/scl.spec.erb
```diff
@@ -6,12 +6,15 @@ SCL_PREFIXES = JSON.load(File.read(File.join(root, 'scl_prefixes.json')))
 LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
 
 config.rules[:doc] << /\/?CHANGES/
+config.rules[:doc] << 'web'
```
ekohl, line 9: I still feel like web (and sample below) are too generic and have a risk of causing problems in the future. @evgeni?

evgeni: I am not versed enough in the standard (or well, in this case less standard) layouts of gems and what can be hidden in these folders.

That said, the content of the `web` folder doesn't look like docs directly -- these are templates that need processing and look kinda useless w/o it. So for `racc`, I lean towards `%exclude`ing it completely.

And then I know no examples of `web` being anything, so maybe it should not be part of the template either?

`sample` on the other hand actually looks like useful examples, which I'd consider documentation? But the `sample` name is indeed a tad off, I'd usually put those in `examples`.

evgeni:
> also has a build requires on gcc. I wonder why it's not needed here.

probably because we add it to the default build group, while Fedora stopped doing that a few releases back?

https://github.com/theforeman/tool_belt/blob/548c363fe61a68438cb3db3b6d0a9673f2ae7e48/configs/foreman/2.5.yaml#L49

packages/foreman/rubygem-racc/rubygem-racc.spec
```diff
@@ -0,0 +1,123 @@
...
+%{gem_spec}
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/ChangeLog
+%doc %{gem_instdir}/README.ja.rdoc
+%doc %{gem_instdir}/README.rdoc
+%{gem_instdir}/Rakefile
+%doc %{gem_instdir}/rdoc
+%{gem_instdir}/web
```
evgeni, line 117: As mentioned in the other comment, I think this is not useful documentation and we should not ship it.

(those are the sources that generate https://i.loveruby.net/en/projects/racc/, which hasn't been updated for the latest releases either)

packages/foreman/rubygem-racc/rubygem-racc.spec
```diff
@@ -0,0 +1,123 @@
...
+%files
+%dir %{gem_instdir}
+%{_bindir}/racc
+%exclude %{gem_instdir}/ext
+%{gem_extdir_mri}
+%license %{gem_instdir}/COPYING
+%exclude %{gem_instdir}/TODO
+%{gem_instdir}/bin
+%{gem_libdir}
+%exclude %{gem_instdir}/sample
```
evgeni, line 106: IMHO examples belong to the doc package

ehelms:
Updated based on last round of comments

evgeni (approved):

evgeni:
@ehelms seems you forgot something ;)

https://ci.theforeman.org/job/foreman-packaging-rpm-nightly-release/931/console

PR #6584, opened 2021-04-30
add script to compare git and debian repo

evgeni:

### DIFF
--- a/scripts/compare_deb_repo.py
+++ b/scripts/compare_deb_repo.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+
+import requests
+from itertools import chain
+from pathlib import Path
+from debian.deb822 import Packages
+from debian.changelog import Changelog
+
+DISTS = [path.name for path in Path('debian/').glob('*') if path.name != 'buildscripts']
+PLUGINS = ['plugins']
+
+
+def get_repo_packages(dist, release='nightly', arch='amd64'):
+    repo_packages = set()
+    remote_packages = requests.get(f'https://deb.theforeman.org/dists/{dist}/{release}/binary-{arch}/Packages')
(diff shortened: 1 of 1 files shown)


scripts/compare_deb_repo.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
+
+import glob
+import os
+import requests
+from debian.deb822 import Packages
+from debian.changelog import Changelog
+
+DISTS = [os.path.basename(d) for d in glob.glob('debian/*') if d != 'debian/buildscripts']
```
ekohl, line 9: I like using `pathlib.Path`
```suggestion
DISTS = [p.name for p in Path('debian').glob('*') if p.name != 'buildscripts']
```

ekohl (approved):
This looks smarter than what I wrote in https://github.com/ekohl/theforeman-plumbing/blob/master/packages/deb/application.py (and I never finished the git part @ https://github.com/ekohl/theforeman-plumbing/blob/master/packages/git/build_cache).

evgeni:
@ekohl I went ahead and fully migrated to pathlib (also the glob parts), mind having a look before I merge?

scripts/compare_deb_repo.py
```diff
@@ -0,0 +1,54 @@
...
+            continue
+        repo_packages.add((source, version))
+    return repo_packages
+
+
+def get_git_packages(dist):
+    if dist == 'plugins':
+        package_folders = Path('plugins/').glob('*/debian')
+    else:
+        package_folders = chain(Path('debian/').glob(f'{dist}/*'), Path('dependencies/').glob(f'{dist}/*'))
```
ekohl, line 29: I normally don't use a trailing slash for directories but it doesn't make a difference.

ekohl (approved):

PR #6597, opened 2021-05-05
Add package to all distros

ekohl:
When adding a package, it is common to use the EL7 tito tag. This makes sure it is added to all distros. This mostly means EL8 but for client distros it includes Fedora.

### DIFF
--- a/add_gem_package.sh
+++ b/add_gem_package.sh
@@ -49,34 +49,74 @@ generate_gem_package() {
 	popd
 }
 
+add_to_all_tito_props() {
+	add_to_tito_props $TITO_TAG
+	if [[ $TITO_TAG == "foreman-nightly-rhel7" ]] || [[ $TITO_TAG == "foreman-nightly-nonscl-rhel7" ]] ; then
+		add_to_tito_props foreman-nightly-el8
+	elif [[ $TITO_TAG == "foreman-plugin-nightly-rhel7" ]] || [[ $TITO_TAG == "foreman-plugin-nightly-nonscl-rhel7" ]] ; then
+		add_to_tito_props foreman-plugin-nightly-el8
+	elif [[ $TITO_TAG == "foreman-client-nightly-rhel7" ]] ; then
+		add_to_tito_props foreman-client-nightly-el8
+		add_to_tito_props foreman-client-nightly-rhel6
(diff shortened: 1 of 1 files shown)


add_gem_package.sh
```diff
@@ -49,34 +49,74 @@ generate_gem_package() {
 	popd
 }
 
+add_to_all_tito_props() {
+	add_to_tito_props $TITO_TAG
+	if [[ $TITO_TAG == "foreman-nightly-rhel7" ]] || [[ $TITO_TAG == "foreman-nightly-nonscl-rhel7" ]] ; then
```
evgeni, line 54: should this instead match `$TITO_TAG == foreman-nightly-*`?

add_gem_package.sh
```diff
@@ -49,34 +49,74 @@ generate_gem_package() {
...
 	local whitelist=$(echo "$current_whitelist $PACKAGE_NAME" | tr " " "\n" | sort -u)
-	crudini --set rel-eng/tito.props $TITO_TAG whitelist "$whitelist"
+	crudini --set rel-eng/tito.props $tag whitelist "$whitelist"
 	export LC_COLLATE=$original_locale
 	git add rel-eng/tito.props
 }
 
+add_gem_to_all_comps() {
+	add_gem_to_comps $TITO_TAG
+	if [[ $TITO_TAG == "foreman-nightly-rhel7" ]] || [[ $TITO_TAG == "foreman-nightly-nonscl-rhel7" ]] ; then
```
evgeni, line 82: same q as above

evgeni (approved):
one q, but nothing that should prevent merging

add_gem_package.sh
ekohl: Doesn't that mean that if you pass EL8 as a tito tag it gets added twice?

evgeni: Yeah, that'd mean it. But shouldn't the underlying "add" code handle that?

Anyways, that's not blocking, just an idea ;)

ekohl: I thought about that, but I'd prefer to merge this now and improve it if people dislike it. I was mostly annoyed at having to add it manually to additional files and we've seen the same thing in reviews.

PR #6599, opened 2021-05-05
25 configs

zjhuntin:


rel-eng/tito.props
```diff
@@ -971,31 +971,23 @@ whitelist =
...
-[foreman-client-nightly-fedora29]
-disttag = .fc29
-whitelist =
-  foreman-release
-  katello-host-tools
-  rubygem-foreman_scap_client
-  tracer
-
-[katello-nightly-rhel7]
+[katello-2.5-rhel7]
```
evgeni, line 990: 4.1 please

evgeni (changes requested):
There are still Fedora 29 related entries in package manifest

rel-eng/tito.props
zjhuntin: Thanks, I think I got em all this time.

evgeni (approved):

PR #6600, opened 2021-05-05
Update rubygem-smart_proxy_dynflow_core to 0.3.3

adamruzicka:
CPs:
- [ ] 2.5
- [ ] 2.4


ekohl (approved):

ekohl:
If this works, I think it should be picked to 2.5. Right?

adamruzicka:
Yes, definitely 2.5 and since it is just a bugfix release I wouldn't mind pushing it into 2.4 as well.

PR #6609, opened 2021-05-07
Update rubygem-kafo to 6.4.0

ehelms:
Supported Versions:

 * Nightly
 * 2.5


packages/foreman/rubygem-kafo/rubygem-kafo.spec
```diff
@@ -17,19 +17,19 @@ Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 # start specfile generated dependencies
 Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}ruby >= 2.4.0
-Requires: %{?scl_prefix_ruby}ruby(rubygems)
-Requires: %{?scl_prefix}rubygem(kafo_wizards)
-Requires: %{?scl_prefix}rubygem(ansi)
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
```
ekohl, line 20: This trailing whitespace is odd.

ekohl (changes requested):

packages/foreman/rubygem-kafo/rubygem-kafo.spec
ehelms: This seems to be a result of this line:

```
Requires: %{?scl_prefix<%= SCL_PREFIXES[d.name] %>}rubygem(<%= d.name %>) <%= req  %>
```

Where, if there is no value for `<%= req  %>` (an unversioned dependency) you still get that space between the name and the requirement. I could try to clean this up with:

```
Requires: %{?scl_prefix<%= SCL_PREFIXES[d.name] %>}rubygem(<%= d.name %>)<%= req == '' ? '' : ' ' + req %>
```

ekohl: Perhaps Github never highlighted trailing whitespace before and that's why I only see it now. I'd prefer to keep the diff as small as possible and not go back and forth on trailing whitespace.

ehelms: See https://github.com/theforeman/foreman-packaging/pull/6611 for solving this systematically.

ekohl (approved):

PR #6611, opened 2021-05-08
Prevent end of line whitespace in scl gem2rpm template

ehelms:
If there is no 'req' to append at the end of a line, the line will
end up with the space before becoming trailing whitespace.

### DIFF
--- a/gem2rpm/scl.spec.erb
+++ b/gem2rpm/scl.spec.erb
@@ -33,29 +33,29 @@ Source0: <%= download_path %>%{gem_name}-%{version}.gem
 # start specfile generated dependencies
 Requires: %{?scl_prefix_ruby}ruby(release)
 <% for req in spec.required_ruby_version -%>
-Requires: %{?scl_prefix_ruby}ruby <%= req %>
+Requires: %{?scl_prefix_ruby}ruby<%= " #{req}" unless req.empty? %>
 <% end -%>
 <% for req in spec.required_rubygems_version -%>
-Requires: %{?scl_prefix_ruby}ruby(rubygems) <%= req %>
+Requires: %{?scl_prefix_ruby}ruby(rubygems)<%= " #{req}" unless req.empty? %>
 <% end -%>
 <% for d in spec.runtime_dependencies -%>
 <% for req in d.requirement -%>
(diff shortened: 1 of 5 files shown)


gem2rpm/scl.spec.erb
```diff
@@ -33,29 +33,29 @@ Source0: <%= download_path %>%{gem_name}-%{version}.gem
 # start specfile generated dependencies
 Requires: %{?scl_prefix_ruby}ruby(release)
 <% for req in spec.required_ruby_version -%>
-Requires: %{?scl_prefix_ruby}ruby <%= req %>
+Requires: %{?scl_prefix_ruby}ruby<%= req != '' ? req : '' %>
```
evgeni, line 36: Isn't this the same as `<%= req %>` as the code says: if req is not an empty string, print req, otherwise print an empty string?

ehelms: It's the little things... ya know, like the space that is missing. Now it's present.

evgeni: Heh.

I wish we could do something like this (we can't):
```erb
Requires: %{?scl_prefix_ruby}ruby <%=- req %>
```

ehelms: Yea, I'm open to some other code solution but as I keep running into this when trying to update packages I need a solution present in the code base.

gem2rpm/scl.spec.erb
```diff
@@ -33,29 +33,29 @@ Source0: <%= download_path %>%{gem_name}-%{version}.gem
 # start specfile generated dependencies
 Requires: %{?scl_prefix_ruby}ruby(release)
 <% for req in spec.required_ruby_version -%>
-Requires: %{?scl_prefix_ruby}ruby <%= req %>
+Requires: %{?scl_prefix_ruby}ruby<%= req != '' ? ' ' + req : '' %>
```
ekohl, line 36: How about

```suggestion
Requires: %{?scl_prefix_ruby}ruby<%= " #{req}" unless req.empty? %>
```

ekohl (commented):
Shouldn't we apply this in all our specs?

ekohl (approved):

evgeni (commented):
post-merge ack just to keep the bill clean :)

PR #6615, opened 2021-05-10
Update rubygem-puma to 5.3.0

evgeni:


packages/foreman/rubygem-puma/rubygem-puma.spec
```diff
@@ -7,14 +7,13 @@
...
-Version: 5.1.1
-Release: 2%{?dist}
+Version: 5.3.0
+Release: 1%{?dist}
 Summary: Puma is a simple, fast, threaded, and highly concurrent HTTP 1.1 server for Ruby/Rack applications
 Group: Development/Languages
 License: BSD-3-Clause
 URL: https://puma.io
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
-Patch0: rdoc-issue-14617-workaround.patch
```
evgeni, line 17: I expect not to need this anymore with Ruby 2.7, we'll see.

evgeni: It did build :fireworks:

PR #6622, opened 2021-05-10
Update rubygem-foreman_remote_execution to 4.5.0

adamruzicka:
CPs:
- [ ] 2.5


packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -47,6 +47,8 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
 # start package.json devDependencies BuildRequires
 BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
 BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
+BuildRequires: %{?scl_prefix}npm(@patternfly/react-catalog-view-extension) >= 4.8.126
+BuildRequires: %{?scl_prefix}npm(@patternfly/react-catalog-view-extension) < 5.0.0
```
evgeni, line 51:
```
DEBUG util.py:600:  No matching package to install: 'npm(@patternfly/react-catalog-view-extension) >= 4.8.126'
DEBUG util.py:600:  No matching package to install: 'npm(@patternfly/react-catalog-view-extension) < 5.0.0'
```

needs packaging

evgeni (changes requested):

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
evgeni: https://github.com/theforeman/foreman-packaging/pull/6671

I wonder tho, if this shouldn't be exposed via `@theforeman/vendor` cc @theforeman/ui-ux

ekohl: I think we went for a model where plugins could still add their own dependencies.

evgeni: Yeah, not saying it *has* to, just wondering if having it there and consumable by everyone wouldn't be easier.

laviro: +1 for adding it to foreman vendor

evgeni:
[test rpm]

evgeni:
```
ModuleNotFoundError: Module not found: Error: Can't resolve '@patternfly/react-icons/dist/js/icons' in '/builddir/build/BUILDROOT/rubygem-foreman_remote_execution-4.4.0-1.fm2_6.el8.noarch/usr/share/gems/gems/foreman_remote_execution-4.4.0/webpack/react_app/components/RecentJobsCard'
```

I think you need to import that icon differently.

In core we recently did smth like this:

```patch
-import { ExclamationCircleIcon } from '@patternfly/react-icons/dist/js/icons/exclamation-circle-icon';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
```

@laviro will remember :)

jeremylenz:
> In core we recently did smth like this:
> 
> ```diff
> -import { ExclamationCircleIcon } from '@patternfly/react-icons/dist/js/icons/exclamation-circle-icon';
> +import { ExclamationCircleIcon } from '@patternfly/react-icons';
> ```
> 
> @laviro will remember :)

yes, this is the way to import modules! The main module takes care of defining its own named exports.  You shouldn't have to go grab them directly like this.  I've never seen this and wouldn't want to do it unless there were some special reason.

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
jeremylenz: +1 to adding it as well.  However, I'm fine keeping it here for the time being and then removing it once it's in @theforeman/vendor, since there's so much work involved in adding new dependencies there.

evgeni:
@jeremylenz if only the docs would agree with us ;)

https://github.com/patternfly/patternfly-react/tree/master/packages/react-icons

jeremylenz:
> @jeremylenz if only the docs would agree with us ;)

:see_no_evil:

MariaAga:
>  if only the docs would agree with us ;)

@evgeni 
Well there are a few docs, and this one says:

> If you're looking to copy React for an icon:
> Use this for 'pficon' icons: `import { [insert-icon-name] } from '@patternfly/react-icons'`

https://www.patternfly.org/2021.02/guidelines/icons

adamruzicka:
So if I'm reading this right we need to patch rex and make a new release because js is being js?

evgeni:
Yeah, let's see if https://github.com/theforeman/foreman_remote_execution/pull/594 helps

evgeni:
@adamruzicka are you planing to release a fixed version? I think this update blocks foreman_ansible too.

adamruzicka:
@evgeni it is finally out

evgeni (approved):

PR #6644, opened 2021-05-14
don't run the elf_tools sequence

evgeni:
it's rather expensive (it scans all files for being dynamic libraries),
but we don't ship any libs, while we ship A LOT OF FILES.


evgeni:
[test packaging]

evgeni:
seems not to work :(

evgeni:
That sequence doesn't exist in older debhelper :woman_facepalming:

PR #6647, opened 2021-05-14
Katello 4.1.0-rc1

pcreech:


packages/katello/katello-repos/katello-repos.spec
```diff
@@ -3,12 +3,10 @@
...
 %define repo_dist %{dist}
 
-%global prereleasesource nightly
-%global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 3
+%global release 1
 
 Name:           katello-repos
-Version:        4.1
+Version:        4.1.0.rc1
```
evgeni, line 9: No. This needs to be a prerelease

pcreech: inded. should be fixed.

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -3,12 +3,11 @@
 %define repo_dir %{_sysconfdir}/yum.repos.d
 %define repo_dist %{dist}
 
-%global prereleasesource nightly
-%global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 3
+%global prerelease .rc1
+%global release 1
```
evgeni, line 7: Don't reset release for a rc please

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -3,12 +3,11 @@
 %define repo_dir %{_sysconfdir}/yum.repos.d
 %define repo_dist %{dist}
 
-%global prereleasesource nightly
-%global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 3
+%global prerelease .rc1
```
evgeni, line 6: Technically correct, but obal would have generated a prerelease source rc1, and left the other line as is.

I don't mind

packages/katello/katello/katello.spec
```diff
@@ -132,6 +131,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Fri May 14 2021 Patrick Creech <pcreech@redhat.com> - 4.1.0.rc1-1
```
evgeni, line 134: I'd argue that's not the EVR of that release ;)

evgeni (changes requested):

packages/katello/katello-repos/katello-repos.spec
pcreech: *shakes fist at obal*  Scrapping and re-running obal.  probably faster at this point.

evgeni: `--version 4.1.0 --prerelease rc1 --release keep` should do what you need

pcreech: local changes to support parallelization of release/scratch builds  temporarily bit me in the but.  But should _really_ be good now.

evgeni (approved):
Hammer is missing, feel free to add here or in another PR

pcreech:
> Hammer is missing, feel free to add here or in another PR

Check again :D .  I made the change and forgot to add it to the commit initially.  Immediately added it and re-pushed.

pcreech:
Repoclosures expected i'm assuming, as it can't sync 4.1 repos (which aren't there yet)

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -3,12 +3,12 @@
...
 %define repo_dist %{dist}
 
-%global prereleasesource nightly
+%global prereleasesource rc1
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global release 3
 
 Name:           katello-repos
-Version:        4.1
+Version:        4.1.0
```
ekohl, line 11: Should nightly also be x.y.z?
https://github.com/theforeman/foreman-packaging/blob/d5b11c61ec7da57060f8dc760f4b644e4bd075eb/packages/katello/katello-repos/katello-repos.spec#L11

PR #6661, opened 2021-05-18
Update azure_rm plugin packaging so it can work both with Jenkins CI …

eLvErDe:
@evgeni 

Hello, can you try this PR ? It should allow azure_rm plugin package to be built either from your CI and from my dev laptop.

Best regards, Adam.


evgeni:
can you re-open with a branch that's called `deb/develop…`? jenkins is unhappy otherwise

eLvErDe:
Sure.

PR #6671, opened 2021-05-19
Add nodejs-patternfly-react-catalog-view-extension package

evgeni:


ekohl (approved):

ekohl (changes requested):
This is only needed by a plugin, right? So it should live in `packages/plugins`.

evgeni:
> This is only needed by a plugin, right? So it should live in `packages/plugins`.

Technically? Yeah. But we have almost all node things for plugins in "foreman".

ekohl:
I still think it's cleaner to add it to plugins but it doesn't matter that much.

evgeni:
moved

PR #6675, opened 2021-05-19
Require puppet-agent 6.15.0+ in foreman-installer

ehelms:

### DIFF
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,7 +1,7 @@
 %{?scl:%global scl_prefix %{scl}-}
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 2
+%global release 3
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -48,7 +48,7 @@ Requires: which
 # puppet-candlepin enables the pki-core module which needs DNF module support
 # That was introduced in 5.5.20 and 6.15.0
 %if 0%{?rhel} == 8
-Requires:   puppet-agent >= 5.5.20
+Requires:   puppet-agent >= 6.1.0
 Conflicts: (puppet-agent >= 6.0.0 with puppet-agent < 6.15.0)
```
evgeni, line 52: Can we fold that into `puppet-agent >= 6.15.0`?

ehelms: I considered that, but I wanted to be consistent with the testing in foreman-installer.

ekohl: The way https://github.com/theforeman/puppet-candlepin/pull/200 is written now, it must be `>= 6.15.0` everywhere. With versions older than that it'll always show an error. That's why I wrote https://github.com/theforeman/foreman-installer/pull/683#discussion_r634208321. So IMHO foreman-installer should align with that.

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -18,13 +18,13 @@ Source0:    https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?prerel
 BuildArch:  noarch
 
 Requires:   curl
-Requires:   puppet-agent >= 5.5.0
+Requires:   puppet-agent >= 6.15.0
```
ehelms, line 21: I've now unified this to 6.15.0 across the board given puppet-candlepin needs 6.15.0 as a minimum due to enabling the pki-core module.

ehelms:
This looks like it will need updates proposed in here (the last commit) to pass https://github.com/theforeman/foreman-infra/pull/1610

ehelms:
[test rpm]

ehelms:
I have applied the changes in Koji for external Puppet repositories. This PR is now green building against Puppet 6.

ekohl (approved):
Thanks! Do we have a Redmine issue to track this?

PR #6676, opened 2021-05-19
Require puppet-agent 6.15.0+ in foreman-installer

ehelms:

### DIFF
--- a/debian/bionic/foreman-installer/changelog
+++ b/debian/bionic/foreman-installer/changelog
@@ -1,3 +1,9 @@
+foreman-installer (2.6.0-2) stable; urgency=low
+
+  * Require puppet 6.15.0+
+
+ -- Eric D. Helms <ericdhelms@gmail.com>  Wed, 19 May 2021 11:58:40 -0400
+
 foreman-installer (2.6.0-1) stable; urgency=low
 
   * Bump changelog to 2.6.0 to match VERSION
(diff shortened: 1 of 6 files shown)


evgeni:
This should be updated to 6.15 to match #6675?

Not that we need modularity support on Debian, but given we don't test anything older…

ekohl:
I'd say the same thing. In fact, it'll fail otherwise:
https://github.com/theforeman/kafo/blob/a45b1ebedda66cdf460c3fe5bfbb3696f2bd198f/lib/kafo/puppet_command.rb#L86-L99
We do ship all modules (even unused ones) to Debian and that glob version check will ensure it fails if any module doesn't match. That's why I suggested to go straight to 6.15.0 in https://github.com/theforeman/foreman-installer/blob/3cbb7bd7201654a491a927cbb3d6867f50523c99/spec/puppet_version_spec.rb#L5.

ehelms:
Yea, I was just waiting for agreement over here https://github.com/theforeman/foreman-packaging/pull/6675 before I updated everywhere.

ehelms:
Updated to 6.15.0

ekohl (changes requested):
Code wise :+1: but the commit message should reflect the version number. Also, is there a Redmine issue?

ehelms:
Updated commit message with a Redmine

ekohl (approved):

PR #6685, opened 2021-05-20
Update rubygem-qpid_proton to 0.34.0

jlsherrill:


packages/katello/rubygem-qpid_proton/rubygem-qpid_proton.spec
```diff
@@ -7,8 +7,8 @@
 %global qpid_version 0.33.0
```
ehelms, line 7: This is the line that needs updating so that the dependencies get the right version

jlsherrill: oh thanks!

ehelms (approved):

PR #6690, opened 2021-05-21
Katello 3.18.3

evgeni:


evgeni:
[test rpm]

this should build, but will end up being built with (partially) Ruby 2.7 which we don't want.

jturel (approved):

evgeni:
@pcreech as you didn't tag the new rpm_client build into 3.18, this failed to build now.

PR #6693, opened 2021-05-21
rebuild pulp-rpm-client against ruby 2.5

evgeni:

### DIFF
--- a/packages/katello/rubygem-pulp_rpm_client/rubygem-pulp_rpm_client.spec
+++ b/packages/katello/rubygem-pulp_rpm_client/rubygem-pulp_rpm_client.spec
@@ -6,7 +6,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 3.10.0
-Release: 1%{?dist}
+Release: 0.1%{?dist}
 Summary: Pulp 3 RPM plugin API Ruby Gem
 Group: Development/Languages
 License: GPLv2
@@ -85,8 +85,10 @@ cp -a .%{gem_dir}/* \
 %{gem_instdir}/spec
 
 %changelog
-* Tue Apr 13 2021 ianballou <ianballou67@gmail.com> 3.10.0-1
+* Fri May 21 2021 Evgeni Golov 3.10.0-0.1
 - Update to 3.10.0
+- Rebuild against Ruby 2.5
+- Force the version to be lower than the one in Katello 4.1
 
 * Wed Feb 24 2021 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> 3.9.0-1
 - Update to 3.9.0


evgeni:
`tfm-rubygem-pulp_rpm_client-3.10.0-1.el7` needs to be untagged, and then the same needs to happen for 3.18/2.3

packages/katello/rubygem-pulp_rpm_client/rubygem-pulp_rpm_client.spec
```diff
@@ -85,8 +85,10 @@ cp -a .%{gem_dir}/* \
 %{gem_instdir}/spec
 
 %changelog
-* Tue Apr 13 2021 ianballou <ianballou67@gmail.com> 3.10.0-1
+* Fri May 21 2021 Evgeni Golov 3.10.0-0.1
```
evgeni, line 88: Otherwise RPMlint will yell at me for a changelog "going backwards" and I am in no mood to rebuild this here and then also in 4.1/nightly for no good reason.

packages/katello/rubygem-pulp_rpm_client/rubygem-pulp_rpm_client.spec
```diff
@@ -6,7 +6,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 3.10.0
-Release: 1%{?dist}
+Release: 0.1%{?dist}
```
ekohl, line 9: Should this use `%{foremandist}`? That's what we've been using to deal with multiple Foreman releases that both tag the same version. We should then do the same in nightly and 2.5.
```suggestion
Release: 0.1%{foremandist}%{?dist}
```

evgeni: It kinda feels odd to use foremandist in RPMs that are not related to Foreman at all. But we could.

pcreech: It's probably worth a broader conversation, since we will probably have these ruby27 <-> ruby25 issues untill 2.4 goes away.  I wouldn't block this PR on it.

ekohl: I think we should have the broader discussion sooner rather than later, but I won't block this PR on it.

pcreech (approved):

packages/katello/rubygem-pulp_rpm_client/rubygem-pulp_rpm_client.spec
pcreech: agreed

pcreech:
> `tfm-rubygem-pulp_rpm_client-3.10.0-1.el7` needs to be untagged, and then the same needs to happen for 3.18/2.3

untagged for katello-4.0-rhel7 and katello-3.18-rhel7

PR #6583, opened 2021-04-30
Update rubygem-nokogiri to 1.11.3

ehelms:

PR #6717, opened 2021-05-26
Only build rubygem-racc on EL7

ehelms:
On EL8 the Ruby package contains racc as a bundled project.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -247,6 +247,10 @@ foreman_scl_packages:
     rubygem-quantile: {}
     rubygem-rabl: {}
     rubygem-racc: {}
+      koji_tags:
+        - name: "foreman-{{ foreman_version }}-rhel7"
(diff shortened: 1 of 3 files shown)


evgeni (approved):

ehelms:
Note you can see the effect of this over here -- https://github.com/theforeman/foreman-packaging/pull/6583/files#diff-41b699811af63453ce63c3b9795bb6cf9889d50e91896df86c271081997cad7fR33

package_manifest.yaml
```diff
@@ -247,6 +247,10 @@ foreman_scl_packages:
     rubygem-quantile: {}
     rubygem-rabl: {}
     rubygem-racc: {}
```
evgeni, line 249: and the syntax was broken :(

evgeni: (fixed with Odilon)

PR #6723, opened 2021-05-31
Update ACD and its components

sbernhard:


packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
```diff
@@ -103,6 +108,8 @@ cp -pa .%{gem_dir}/* \
 %foreman_bundlerd_file
 %foreman_precompile_plugin -a -s
 
+mkdir -p %{buildroot}%{_root_localstatedir}/lib/foreman/%{gem_name}/ansible-playbooks/
```
evgeni, line 111: Shouldn't this happen inside smart_proxy_acd instead? As that's the one actually running the playbooks?

sbernhard: In %{_root_localstatedir}/lib/foreman/%{gem_name}/ansible-playbooks/ are the ansible playbook stored, which are cloned from git (with the new git support feature).

evgeni: OK

packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
```diff
@@ -18,16 +19,20 @@ Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
...
+Requires: %{?scl_prefix}rubygem(git)
 BuildRequires: foreman-assets >= %{foreman_min_version}
 BuildRequires: foreman-plugin >= %{foreman_min_version}
 BuildRequires: %{?scl_prefix}rubygem(foreman_remote_execution) >= 3.3.0
 BuildRequires: %{?scl_prefix}rubygem(foreman-tasks) >= 0.10
+BuildRequires: %{?scl_prefix}rubygem(git)
 BuildRequires: %{?scl_prefix_ruby}ruby(release)
-BuildRequires: %{?scl_prefix_ruby}ruby
+BuildRequires: %{?scl_prefix_ruby}ruby >= 2.5
+BuildRequires: %{?scl_prefix_ruby}ruby < 3
```
ekohl, line 35: I'd like to note that Fedora 34 ships Ruby 3 by default. I'd strongly recommend to make sure the plugin becomes Ruby 3 compatible. Ruby 2.7 (our default version now) is already quite verbose if you do use deprecated things.

sbernhard: Thanks for letting me know. I don't think it will break with ruby 3

ekohl: It's generated from https://github.com/ATIX-AG/foreman_acd/blob/89cdafc2029480dbbf2343b311fd34fe8fce0e2b/foreman_acd.gemspec#L16 which declares it to be incompatible with Ruby 3.

sbernhard: Ah, thank you very much.
Should I change it now or should I do this for the next version?

ekohl: You shouldn't change this PR since it reflects what's actually in the gemspec, but you should plan for it. Today we don't test with Ruby 3, but it's an effort we should start soon (IMHO).

sbernhard: Yes. make sense. Very appreciated - as always @ekohl :-)

evgeni (approved):

PR #6727, opened 2021-05-31
Bump rubygem-foreman_rh_cloud to 3.0.21-1

laviro:


laviro:
@evgeni @shimshtein in the end we didn't need the patch, so v3.0.21 should work

evgeni (approved):

PR #6741, opened 2021-06-02
Update katello to 4.1.0.rc2.2

Odilhao:


packages/katello/katello/katello.spec
```diff
@@ -3,12 +3,10 @@
 
 %global homedir %{_datarootdir}/%{name}
 %global confdir common
-%global prereleasesource rc2
-%global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global release 6
 
 Name:       katello
-Version:    4.1.0
+Version:    4.1.0.rc2.2
```
evgeni, line 9: nope :(

you need to run obal with `--version 4.1.0 --release keep --prerelease rc2.2`, otherwise it will break versioning :(

(4.1.0.rc2 is bigger than 4.1.0 if you ask RPM)

evgeni (changes requested):

packages/katello/katello/katello.spec
Odilhao: The first time I only used `obal update katello --version 4.1.0.rc2.2 --release keep`

Now it was updated with `--version 4.1.0 --release keep --prerelease rc2.2`

evgeni (approved):

PR #6742, opened 2021-06-02
Bump rubygem-foreman_rh_cloud to 3.0.21.1-1

laviro:


laviro:
@evgeni ready for review :)
@shimshtein for the `develop` branch, should we release v4 of the plugin, or can we still use the same version for both?

evgeni:
given `develop` has 3.0.21 right now (via https://github.com/theforeman/foreman-packaging/pull/6727), I'd think it also needs the same update?

evgeni:
develop in #6744

ShimShtein:
@laviro, since we're already branched, let's do the proper 4.0.21 for develop. This way at least the versioning won't be confusing.

laviro:
@evgeni @ShimShtein, I released `v4` properly now and added it to `develop` in https://github.com/theforeman/foreman-packaging/pull/6750
once it goes in I will do the CPs to `2.5` & `2.4`

evgeni:
ACK!

PR #6750, opened 2021-06-06
Bump rubygem-foreman_rh_cloud to 4.0.21.1-1

laviro:


packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -5,10 +5,10 @@
 
 %global gem_name foreman_rh_cloud
 %global plugin_name rh_cloud
-%global foreman_min_version 2.3
+%global foreman_min_version 2.5
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 3.0.21.1
+Version: 4.0.21.1
```
laviro, line 11: The change in code is very minimal, but from now on `v4` will be developed mainly and relevant cherry-picks will be added to `v3`, cc @shimshtein

packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -5,10 +5,10 @@
 
 %global gem_name foreman_rh_cloud
 %global plugin_name rh_cloud
-%global foreman_min_version 2.3
+%global foreman_min_version 2.5
```
laviro, line 8: should we set the min version to `2.4` for easier CPs in the future?

ShimShtein: I think it would be better to set it to 2.5. The 2.4 could use 3.y.z versions.

evgeni (approved):

PR #6753, opened 2021-06-07
Update smart_proxy_dynflow* to 0.4.0

adamruzicka:
- Update rubygem-smart_proxy_dynflow to 0.4.0
- Update rubygem-smart_proxy_dynflow_core to 0.4.0
- Drop smart_proxy_dynflow_core leftovers


evgeni:
Does that need to go in together with the appropriate installer change, dropping "external_core" support?

https://github.com/theforeman/puppet-foreman_proxy/blob/941e934a4a27a40e7963bc59f6ab5c14fafb82dc/manifests/plugin/dynflow.pp

cc @ekohl

evgeni:
Or is https://github.com/theforeman/puppet-foreman_proxy/commit/99d41ad9ade1a322368b1a5b3a53860287ed42aa sufficient?

adamruzicka:
theforeman/puppet-foreman_proxy@99d41ad should be sufficient unless someone explicitly forces the use of external core. In that case, the installer would probably fail when trying to start `smart_proxy_dynflow_core.service` which no longer exists, but could be cleanly fixed with rerun that parameter set to false.

How about I draft up a installer pr which will stop accepting this parameter and will always set it to false so that spdc.service is properly stopped. Core related bits could then be dropped from the installer once we drop spdc completely. Opinions?

evgeni:
As long merging this PR doesn't break default installs, I don't mind too much.


Cleaning up remains of core at some point sounds good, but doesn't need to be coupled.

adamruzicka:
It should not break default installs. Can we proceed with this?

evgeni (approved):

adamruzicka:
Default installation seems to work just fine even with these changes

PR #6759, opened 2021-06-07
Fixes #32743 - Correct scenario answers file

ekohl:
The script used to assume it's always ${scenario}-answers.yaml which can lead to incorrect behavior. The right way to determine the answers file location is to read the scenario file and use the value for :answer_file.

It has some safeguard that it could at least determine the path.

### DIFF
--- a/packages/katello/katello/helper.rb
+++ b/packages/katello/katello/helper.rb
@@ -32,8 +32,16 @@ module KatelloUtilities
       "This utility can't run on a non-katello system."
     end
 
+    def load_scenario_data(scenario)
+      YAML.load_file(File.join(scenarios_path, "#{scenario}.yaml"))
+    end
+
     def load_scenario_answers(scenario)
-      YAML.load_file("/etc/foreman-installer/scenarios.d/#{scenario}-answers.yaml")
+      data = load_scenario_data(scenario)
+      unless data && data[:answer_file]
+        fail_with_message("Could not determine answers file location for scenario '#{scenario}'")
+      end
+      YAML.load_file(data[:answer_file])
     end
 
     def fail_with_message(message, opt_parser=nil)


ekohl:
I don't have a good environment to test this change, which is why it's a draft.

ekohl:
Actually, it can be reviewer. Just know I haven't tested this :)

ianballou:
I tested this on Katello 4.0 and it worked just fine. I also tested it on Satellite 6.7 and got an error relating to `last_scenario.yaml`.  From a glance it looks unrelated to that change but I'll share anyway:
```
[root@sat-6-7-qa-rhel7 ~]# satellite-change-hostname test-hostname.server.example.com -u admin -p changeme

Checking hostname validity

Checking overall health of server

Checking credentials

***WARNING*** This script will modify your system.
You will need to re-register any foreman clients registered to this system after script completion.
Capsules will have to be re-registered and reinstalled. If you are using custom certificates,
you will have to run the foreman-installer again with custom certificate options after this script completes.
 Have you taken the necessary precautions (backups, snapshots, etc...)?
Proceed with changing your hostname? [y/n]
y
Precheck passed
updating hostname in /etc/hostname
setting hostname
checking if hostname was changed

Updating default Capsule
Updating installation media paths
stopping services
removing old cert rpms
deleting old certs
backed up /var/www/html/pub to /var/www/html/pub/sat-6-7-qa-rhel7.cannolo.example.com-20210607153317.backup
updating hostname in /etc/hosts
updating hostname in foreman installer scenarios
updating hostname in hammer configuration
backing up last_scenario.yaml
removing last_scenario.yaml
re-running the installer
foreman-installer --scenario satellite -v --disable-system-checks --certs-regenerate=true --foreman-proxy-register-in-foreman true
Forward DNS points to 192.168.122.104 which is not configured on this server
restoring last_scenario.yaml
/usr/share/katello/hostname-change.rb:382:in `restore_last_scenario_yaml': undefined method `path' for nil:NilClass (NoMethodError)
	from /usr/share/katello/hostname-change.rb:533:in `block in run'
	from /usr/share/katello/helper.rb:67:in `run_cmd'
	from /usr/share/katello/hostname-change.rb:530:in `run'
	from /sbin/satellite-change-hostname:23:in `<main>'

```

ekohl:
That may be d6b61491707e880189ec92c5436c8b6f01b1f615 that's missing.

ianballou (approved):
@ekohl that missing commit fixed it for me on Satellite. All good!

PR #6774, opened 2021-06-11
Fixes #32725 - Make delete_puppet_certs handle all input on rpm/2.5

Odilhao:
If a user runs the installer with --no-enable-puppet the value of
@scenario_answers['puppet'] becomes false. false['ssldir'] raises an
exception. By returning early, it's handled safely.

(cherry picked from commit ee14227ebe203d54e5640c67d7d4925c60b9b27a)


ekohl (approved):

ekohl (commented):
I now do realize that I forgot to bump the katello version so it was never built.

ekohl:
Yes, but we may need to fix that in develop.

PR #6776, opened 2021-06-13
Drop jquery-ui-rails gem

ezr-ondrej:
We moved all widgets used from jquery-ui library to React
implementations, so we can drop this dependency.

This needs to wait for theforeman/foreman#8586 to get in.


ezr-ondrej:
I've tried to follow the guide, if something is not right, please let me know :)

evgeni:
Given https://github.com/theforeman/foreman/pull/8586#issuecomment-860495467, I don't think we can really drop it?

ezr-ondrej:
Yeah :(( unfortunatelly no, I've just got excited.

PR #6785, opened 2021-06-15
Bumping releases to reflect downstream

Odilhao:


packages/katello/rubygem-activerecord-import/rubygem-activerecord-import.spec
```diff
@@ -7,7 +7,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 1.0.0
-Release: 3%{?dist}
+Release: 7%{?dist}
```
ekohl, line 10: There are newer versions on Rubygems. Can't we update instead?

packages/plugins/rubygem-rainbow/rubygem-rainbow.spec
```diff
@@ -5,7 +5,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 2.2.1
-Release: 4%{?dist}
+Release: 6%{?dist}
```
ekohl, line 8: Can we bump to 2.2.2 instead?

packages/katello/rubygem-activerecord-import/rubygem-activerecord-import.spec
Odilhao: Updated the rubygem to 1.1.0

packages/katello/rubygem-anemone/rubygem-anemone.spec
```diff
@@ -58,6 +58,9 @@ cp -a .%{gem_dir}/* \
 %doc %{gem_instdir}/VERSION
 
 %changelog
+* Tue Jun 15 2021 Odilon Sousa <osousa@redhat.com> - 0.7.2-23
+- align version with satellite
```
ehelms, line 62: This is not something we have traditionally done in the past as the project and product are different cadences. What would be the reason to start doing so now?

packages/plugins/rubygem-rainbow/rubygem-rainbow.spec
ekohl: I generally like it if there aren't irrelevant changes in the git log. Perhaps you can clean up the commits a bit.

Another piece of advice is that bumping is pretty safe while bump is its own discussion. If you submit the bumps as a separate PR it can be merged and not be held up by the revbump alignment discussion. Small PRs tend to be merged faster.

packages/katello/rubygem-anemone/rubygem-anemone.spec
evgeni: the problem is that anemone got a few bumps downstream, and when we rebase a downstream release on upstream, we pull in a "too old" version, thus maybe forgetting changes.

by syncing the release numbers, my intent here was to get out of this mess to have a higher version than downstream.

I know we did that in the past, not often, but we did.

evgeni (approved):

evgeni:
[test rpm]

PR #6791, opened 2021-06-16
Update rubygem-foreman_discovery to 17.0.1 (2.5)

lzap:
* 2.5
 * 2.4

Nightly to be updated later, this is an important patch, develop is already on 17.1.


evgeni (approved):

lzap:
In 2.4 we also have 17.0 so I will CP into it: http://yum.theforeman.org/plugins/2.4/el7/x86_64/ TY

PR #6793, opened 2021-06-16
Update rubygem-foreman_discovery to 17.0.1 (2.4)

lzap:
(cherry picked from commit 7007e2ad50c0b392f8d4a71f891cc4ff2088bd62)


evgeni (approved):

lzap:
Green. I just found that debian contains 16.3.x so I will do that version there rather than bumping it that late :-)

PR #6798, opened 2021-06-17
drop rubygem-mimemagic

evgeni:
it's unused since fd6a14311c1f132d541be91fd042c7952b8eec82


ekohl (approved):
This is unused since fd6a14311c1f132d541be91fd042c7952b8eec82.

evgeni:
@ehelms @ekohl do we want to pick that to 2.5 which got that new marcel via https://github.com/theforeman/foreman-packaging/commit/5cee56445860a72f58a2854f229961258ff5119e

ekohl:
It feels a bit odd since the mimemagic gem is in the yum repo but at runtime the user will have the benefit of having fewer files on disk so I'd be open to it.

Perhaps you can include the commit ID that made it obsolete in the commit message.

evgeni:
updated the commit msg

ekohl (approved):

PR #6807, opened 2021-06-21
Pulp 3.13 bindings update

jlsherrill:


jlsherrill:
note that container isn't updated due to a bug (https://pulp.plan.io/issues/8941), however the old bindings appear to work just fine.

I'm not 100% sure if i should update the pulp-python  bindings or not, as thats still a WIP on the katello side

jlsherrill:
requires https://github.com/Katello/katello/pull/9419

PR #6810, opened 2021-06-22
Pulp 3.13 bindings update

jlsherrill:


ehelms:
I think we can close tihs?

evgeni:
Yeah, #6840 obsoleted that.

PR #6818, opened 2021-06-23
Fixes #32859 - fix cron syntax

evgeni:


ekohl (approved):

evgeni:
going to merge this before the build finishes, as its just a file that gets copied in 1:1, so the package can't fail :)

PR #6816, opened 2021-06-23
Update rubygem-smart_proxy_ansible to 3.2.1

adamruzicka:

PR #6819, opened 2021-06-23
Update smart_proxy_ansible to 3.2.0

adamruzicka:


adamruzicka:
Not sure how to do an equivalent of https://github.com/theforeman/foreman-packaging/pull/6816/commits/1af087e2e1995e82d8b7d3e21341811d803e417b and https://github.com/theforeman/foreman-packaging/pull/6816/commits/052eade58924ba8274d0a7d35637599268e04e10 in debian packaging. Any pointers?

evgeni:
See https://wiki.debian.org/PackageTransition, I think this is case 6 "Merge", so you'd do something like:

```
Breaks: ruby-foreman-ansible-core (<<4.1.0)
Replaces: ruby-foreman-ansible-core (<<4.1.0)
Provides: ruby-foreman-ansible-core # optional, only needed if things still depend on -core, but I think you remove the only depending relation in this PR
```

(I picked 4.1.0 as 4.0.0-2 is the last version we have a package for. you could also make it tighter, if you want)

plugins/smart_proxy_ansible/control
```diff
@@ -11,7 +11,8 @@ Package: ruby-smart-proxy-ansible
 Architecture: all
 XB-Ruby-Versions: ${ruby:Versions}
 Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter, foreman-proxy (>= 1.22.0~rc1),
-  ruby-smart-proxy-dynflow (>= 0.1.0), ruby-smart-proxy-dynflow (<< 1.0.0),
-  ruby-foreman-ansible-core,
+  ruby-smart-proxy-dynflow (>= 0.5.0), ruby-smart-proxy-dynflow (<< 1.0.0),
+  ruby-foreman-ansible-core, ruby-net-ssh,
```
ezr-ondrej, line 15: do we want to keep `foreman-ansible-core`?

plugins/smart_proxy_ansible/changelog
```diff
@@ -1,3 +1,15 @@
...
+  * 3.2.1 released
+
+ -- Adam Ruzicka <aruzicka@redhat.com>  Wed, 23 Jun 2021 15:56:38 +0200
+
+ruby-smart-proxy-ansible (3.2.0-1) stable; urgency=low
+
+  * 3.2.0 released
+
+ -- Adam Ruzicka <aruzicka@redhat.com>  Wed, 23 Jun 2021 15:04:21 +0200
+
```
ezr-ondrej, line 12: do we want to include this? it will not be in repository, just in the history.... :thinking:

ezr-ondrej (commented):
The title is missleading and just two nits I could see.

But would be lovely if someone who knows what's happening would also taken a look.

plugins/smart_proxy_ansible/changelog
evgeni: usually we do, yes

plugins/smart_proxy_ansible/control
evgeni: my understanding is, this should go away

adamruzicka: It should go away, I kept it around as I wasn't sure how to proceed with this

plugins/smart_proxy_ansible/changelog
ezr-ondrej: even tho this version was never added in develop branch? just for a brief period in this PR? interesting :)

evgeni (approved):

PR #6821, opened 2021-06-23
Update rubygem-foreman_ansible to 6.4.1

adamruzicka:


ehelms:
This looks to be failing due to a webpack compile problem. Wil have to do some digging. I believe this package is needed to fix the nightly plugin release pipeline which would then fix Katello nightly upgrades.

ehelms:
See https://github.com/theforeman/foreman_ansible/pull/423

adamruzicka:
foreman_ansible_core was dropped in https://github.com/theforeman/foreman-packaging/pull/6816 yet it seems to block repoclosure. Do we need to purge it somehow?

packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -125,6 +125,12 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/test
 
 %changelog
+* Fri Jun 25 2021 Adam Ruzicka <aruzicka@redhat.com> 6.4.1-1
+- Update to 6.4.1
+
+* Wed Jun 23 2021 Adam Ruzicka <aruzicka@redhat.com> 6.4.0-1
+- Update to 6.4.0
```
ehelms, line 132: Since we did not release 6.4.0 as an RPM we can drop this

ehelms: Could you also then squash?

ehelms:
Nah, the repoclosure is an inconsistency in how repoclosure works and should work itself out after we merge and run pipelines.

ehelms (approved):

PR #6823, opened 2021-06-23
Update foreman_ansible to 6.4.0

adamruzicka:


evgeni:
```
15:18:10 Bundler could not find compatible versions for gem "acts_as_list":
15:18:10   In Gemfile:
15:18:10     foreman_ansible (= 6.4.1) was resolved to 6.4.1, which depends on
15:18:10       acts_as_list (~> 1.0.3)
15:18:10 
15:18:10 Could not find gem 'acts_as_list (~> 1.0.3)', which is required by gem
15:18:10 'foreman_ansible (= 6.4.1)', in any of the sources.
```

I think you need to add that to `gem.list`?

evgeni (approved):

PR #6832, opened 2021-06-30
Bump nodejs-theforeman-{builder,vendor} to 8.7.0

ekohl:


ekohl:
Needed for https://github.com/theforeman/foreman/pull/8627. After that, foreman.spec should also be updated to require the newer versions.

ehelms (approved):

PR #6840, opened 2021-07-02
Update katello to user pulpcore 3.14, part 1

jjeffers:
Updates katello to use Pulpcore 3.14.


packages/katello/katello-repos/katello-repos.spec
```diff
@@ -68,6 +68,9 @@ rm -rf %{buildroot}
 %config %{repo_dir}/*.repo
 
 %changelog
+* Thu Jul 2 2021 James Jeffers <jjeffers@redhat.com> - 4.2-0.1.nightly
```
ekohl, line 71: You missed the release bump (line 8). Then this needs to match
```suggestion
* Thu Jul 2 2021 James Jeffers <jjeffers@redhat.com> - 4.2-0.2.nightly
```

ekohl (changes requested):

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -68,6 +68,9 @@ rm -rf %{buildroot}
 %config %{repo_dir}/*.repo
 
 %changelog
+* Fri Jul 2 2021 James Jeffers <jjeffers@redhat.com> - 4.2-0.2.nightly
```
ehelms, line 71: You are missing a bump to release on line 8 above.

evgeni: I took the liberty to commit that, to get the builds further :)

evgeni:
this PR deletes `packages/katello/rubygem-pulp_ansible_client/pulp_ansible_client-0.7.3.gem`, but I don't see a 0.8.0 version added, or the spec changed?

evgeni:
:disappointed: the Ruby bindings seem to (again) contain the Python bindings too.

Justin had a crude fix for that in #6810 -- you might want to copy that over.

evgeni:
in #6810 we also bumped pulpcore_client, deb_client and python_client -- I don't see these here -- intentional?

evgeni:
And depends on something like https://github.com/Katello/katello/pull/9419 -- is there such a PR for 3.14 in the works?

jjeffers:
> in #6810 we also bumped pulpcore_client, deb_client and python_client -- I don't see these here -- intentional?

From a few days ago, you had mentioned in another thread that pulp-python wasn't packaged yet. Would that exclude the python_client here?

evgeni:
> > in #6810 we also bumped pulpcore_client, deb_client and python_client -- I don't see these here -- intentional?
> 
> From a few days ago, you had mentioned in another thread that pulp-python wasn't packaged yet. Would that exclude the python_client here?

The server (pulpcore) bits aren't packaged yet. But I see Katello depending on the client bindings, so I think these should be packaged/bumped.

jjeffers:
> > > in #6810 we also bumped pulpcore_client, deb_client and python_client -- I don't see these here -- intentional?
> > 
> > 
> > From a few days ago, you had mentioned in another thread that pulp-python wasn't packaged yet. Would that exclude the python_client here?
> 
> The server (pulpcore) bits aren't packaged yet. But I see Katello depending on the client bindings, so I think these should be packaged/bumped.
> 
> Edit: But given we're not actually going to release this functionality just yet (due to the missing server bits), it doesn't need to be bumped in this PR.

Are packaging PRs squashed before merging?

evgeni:
> Are packaging PRs squashed before merging?

usually not.

the python bindings have the same issue as the rpm ones: shipped python files :(

jjeffers:
> > Are packaging PRs squashed before merging?
> 
> usually not.
> 
> the python bindings have the same issue as the rpm ones: shipped python files :(

Ok, I copied over the exclusion directives from Justin's abandoned pulp 3.13 update PR. I think I got them all.

ianballou:
pulpcore-client 3.14.1(I'm guessing) will need to be released to deal with https://pulp.plan.io/issues/9008

jjeffers:
> pulpcore-client 3.14.1(I'm guessing) will need to be released to deal with https://pulp.plan.io/issues/9008

Yup, waiting to see what the new gem versions will be before proceeding.

ehelms:
Sounds like we are waiting on one thing before we merge this, in the meantime can you squash all of these commits together? I see early on you broke them out to individual update commits but now with all the changes piled on top that is muddy, so feels like better off squashing it down.

jjeffers:
> Sounds like we are waiting on one thing before we merge this, in the meantime can you squash all of these commits together? I see early on you broke them out to individual update commits but now with all the changes piled on top that is muddy, so feels like better off squashing it down.

I will squash!

jlsherrill:
3.14.1 is released, mind updating to that?  https://rubygems.org/gems/pulpcore_client/versions/3.14.1

jlsherrill:
and now the final piece :)  https://rubygems.org/gems/pulp_rpm_client/versions/3.13.3

evgeni:
Given only pulp_ansible_client and pulp_python_client have the problem of python code being included in the gem (probably because those releases don't have https://github.com/pulp/plugin_template/commit/36a8e7d73ea38bb4bf92ebb84d7e5263114b2fa6 applied yet), I would lean towards only applying the workaround there?

evgeni:
Otherwise, https://github.com/Katello/katello/pull/9446 looks pretty ready (besides a small comment-nit), so I would expect this to go in today, and then we can merge this PR and follow up with a katello update to the deps there (can't be in the same PR)

evgeni (approved):

PR #6841, opened 2021-07-02
Update nodejs-theforeman-builder to 8.7.1

ezr-ondrej:


ekohl (changes requested):
I don't see a change in the source file.

ezr-ondrej:
Can it be because the only package that changed is the builder itself?

ekohl:
No, you're missing the checksum update. This is also why the build fails:
[2021-07-02T15:37:20.424Z]     Command output: ['error: File /tmp/tito/rpmbuild-nodejs-theforeman-builderI06JyG/SOURCES/nodejs-theforeman-builder-8.7.1/nodejs-theforeman-builder-8.7.1-registry.npmjs.org.tgz: No such file or directory']

ezr-ondrej:
I've figured, but I can't seem to make the bump work properly. @ehelms is trying that locally if that's even gonna fix things, I'll figure why the file is not bumping properly if it's the proper fix

ehelms:
Replaced by https://github.com/theforeman/foreman-packaging/pull/6842

PR #6844, opened 2021-07-05
Add nodejs-use-deep-compare-effect package

evgeni:


ekohl:
https://github.com/Katello/katello/pull/9430 has some discussion on whether it should be included in foreman-js. It's not very clear to me if it should.

evgeni:
Given it breaks nightly packages *now*, I'd rather include it now and remove later.

ekohl (approved):

PR #6847, opened 2021-07-06
Update rubygem-foreman_maintain to 0.8.3

Odilhao:
(cherry picked from commit 96c42e65fa8ef525d597240bff1d1463ff92b5da)


evgeni (commented):
ACK

This will need tagging, and I don't think we set you up with koji credentials yet.

Odilhao:
> ACK
> 
> This will need tagging, and I don't think we set you up with koji credentials yet.

It's correct, I don't have the credentials for koji atm.

PR #6862, opened 2021-07-12
Do not obsolete ansible_core from smart_proxy_ansible

evgeni:


evgeni:
Based on the discussion in #6861, this syncs the changes in Ansible too.

ehelms (approved):

PR #6865, opened 2021-07-12
Update foreman_ansible to 6.3.1

ezr-ondrej:
matching foreman-ansible version for deb in 2.5 as we are minor bumping it in [rpm](https://github.com/theforeman/foreman-packaging/pull/6864)

### DIFF
--- a/plugins/ruby-foreman-ansible/debian/changelog
+++ b/plugins/ruby-foreman-ansible/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-ansible (6.3.1-1) stable; urgency=low
+
+  * 6.3.1 released
+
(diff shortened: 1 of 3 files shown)


evgeni (changes requested):
```
13:52:38   In Gemfile:
13:52:38     foreman_ansible (= 6.3.1) was resolved to 6.3.1, which depends on
13:52:38       acts_as_list (~> 1.0.3)
```

In `deb/develop` we have:
https://github.com/theforeman/foreman-packaging/blob/bf35f551015e42a3852afaf7e48cc803bd391036/plugins/ruby-foreman-ansible/debian/gem.list#L3

ezr-ondrej:
> In `deb/develop` we have

Thanks for noticing, now we do here as well ^_^

zjhuntin (approved):

PR #6873, opened 2021-07-13
Bump foreman-js

tbrisker:
Only change is adding redux toolkit library.


ekohl (commented):
It's not needed to bump the builder too?

tbrisker:
@ekohl there have not been any actual changes in the builder, the version bump there is only due to how lerna is configured to bump and release all packages together. I did initially try to update the builder rpm as well, but the koji build failed for some reason and i figured it isn't worth digging into for no actual changes.

ekohl (approved):

PR #6889, opened 2021-07-14
Update katello-selinux to 3.5.0.1

jturel:


ekohl (commented):
I do wonder why you chose 3.5.0.1 instead of just 3.5.1.

jturel:
> 
> 
> I do wonder why you chose 3.5.0.1 instead of just 3.5.1.

The 3.5.1 tag already exists - no reason beyond that

https://github.com/Katello/katello-selinux/releases/tag/3.5.1

ekohl (approved):
There's no way that'll ever be confusing ;)

PR #6896, opened 2021-07-15
Update rubygem-hammer_cli_foreman_admin to 1.1.0

lzap:

* Nightly


lzap:
There is no debian package of this plugin.

Odilhao (approved):

PR #6899, opened 2021-07-15
Update rubygem-foreman_puppet to 1.0.0.rc.1

ezr-ondrej:


ehelms (approved):

packages/plugins/rubygem-foreman_puppet/rubygem-foreman_puppet.spec
```diff
@@ -8,7 +8,7 @@
 %global foreman_min_version 2.4
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 0.1.0
+Version: 1.0.0.rc.1
```
evgeni, line 11: This is bigger (if you ask RPM) than 1.0.0, so we won't be able to have an 1.0.0 without an epoch (🤮) and would need a 1.0.1 ir similar

PR #6902, opened 2021-07-16
Correct rubygem-arel obsolete statement

ekohl:
The dash is already part of %{scl_prefix} and doesn't need to be specified again.

Reported @ https://community.theforeman.org/t/foreman-package-cleanup/24477.

### DIFF
--- a/packages/foreman/tfm/tfm.spec
+++ b/packages/foreman/tfm/tfm.spec
@@ -19,7 +19,7 @@
 Summary: Package that installs %scl
 Name: %scl_name
 Version: 7.0
-Release: 3%{?dist}
+Release: 4%{?dist}
 License: GPLv2+
 Group: Applications/File
 Source0: README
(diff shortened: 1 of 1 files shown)


packages/foreman/tfm/tfm.spec
```diff
@@ -327,6 +327,9 @@ selinuxenabled && load_policy || :
 %{_root_sysconfdir}/rpm/macros.%{scl_name}-scldevel
 
 %changelog
+* Fri Jul 16 2021 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl>
```
ehelms, line 330: Missing the version-release at the end

ekohl: Surprising - I used obal but now I recall that the latest released version doesn't work and you need a git version.

ehelms (approved):

PR #6906, opened 2021-07-16
Add a conflicts on installer-katello with Pulp < 3.14

ekohl:
This was supposed to be done before 2.5.2 was released. This is because the theforeman/pulpcore was upgraded to a version that's only compatible with Pulp 3.14. Without this users can end up in broken situations. The conflict should prevent them from installing an incompatible version.

https://community.theforeman.org/t/issue-after-yum-update-foreman-2-5-2-pulpcore-workers-failing/24480 shows that users are running into this quickly.


ekohl:
Should it also include a conflict on (tfm-)rubygem-katello < 4.1.1 since that version adds support for Pulp 3.14?

Odilhao (approved):

PR #6907, opened 2021-07-16
Add a conflicts on installer-katello with Katello < 4.1.1

ekohl:
Katello 4.1.1 is the first version to add support for Pulp 3.14 (which is installed by 2.5.2+). While it's not clear if 4.1.0 and 3.14 will break, preventing the use of Katello 4.1.0 with Pulp 3.14 avoids the whole scenario.

This was intended to be part of d86ee2104dcc4e607ffa02038b3b074c78eaa949


jlsherrill (approved):
changes make sense to me

PR #6909, opened 2021-07-17
Update foreman_puppet to 1.0.0.rc.1

ezr-ondrej:


plugins/ruby-foreman-puppet/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-foreman-puppet (1.0.0.rc.1-1) stable; urgency=low
```
evgeni, line 1: This needs to be 1.0.0~rc1 or similar, or it won't sort properly!

evgeni (changes requested):

evgeni (approved):

PR #6911, opened 2021-07-17
Refs #33406: Move scenarios to /var/lib/foreman-installer

ehelms:
Supports the move of scenario answers and config to /var/lib (sharedstatedir)
and out of system config directory. On upgrade, before installing the
new version of the package the scenario files are moved out of /etc
and into /var/lib to allow migrations to apply on the new location properly.
The custom-hiera.yaml remains in /etc as users are expected to edit this file.

See https://github.com/theforeman/foreman-installer/pull/698

Also, I am sure this is entirely the wrong approach. I did not want to forget about Debian and this lets others tell me how I should actually do this.

### DIFF
--- a/debian/bionic/foreman-installer/postinst
+++ b/debian/bionic/foreman-installer/postinst
@@ -19,6 +19,9 @@ set -e
 
 case "$1" in
   configure)
+    if [ -d /etc/foreman-installer/scenarios.d ] ; then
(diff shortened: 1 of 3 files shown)


pcreech:
@ehelms Looks like the issue to reference is 33046, not 33406.

PR #6914, opened 2021-07-20
Only generate post scripts when building for upstream Foreman

evgeni:

### DIFF
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,7 +1,7 @@
 %{?scl:%global scl_prefix %{scl}-}
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 3
+%global release 4
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -73,6 +73,7 @@ Various scenarios and tools for the Katello ecosystem
   SYSCONFDIR=%{buildroot}%{_sysconfdir} \
   --trace
 
+%if 0%{?foremandist:1}
 %post
 foreman-installer --scenario foreman --migrations-only > /dev/null
 
@@ -90,6 +91,7 @@ for scenario in foreman-proxy-content katello ; do
 		rm -rf $MIGRATIONS
 	fi
 done
+%endif
 
 %files
 %defattr(-,root,root,-)
@@ -142,6 +144,9 @@ done
 %{_sbindir}/foreman-proxy-certs-generate
 
 %changelog
+* Tue Jul 20 2021 Evgeni Golov - 1:2.6.0-0.4.develop
+- Only generate post scripts when building for upstream Foreman
+
 * Wed May 19 2021 Eric D. Helms <ericdhelms@gmail.com> - 1:2.6.0-0.3.develop
 - Bump puppet-agent requires to 6.15.0
 


evgeni:
@sbernhard would that affect your workflows in any way?

ekohl (commented):
We talked about splitting the packages, creating `foreman-installer-scenario-{foreman,katello,foreman-proxy-content}`. Perhaps that would be a better solution?

evgeni:
Technically? Maybe. But that would require more work ;)

ehelms:
This does certainly help the situation as an easy solution. The core package + scenario per sub-package has some fun implications and niceness to it. I suppose you'd then need macros per sub-package to allow toggling them off and on.

ekohl:
I'd think that every sub package would have its own `%post` for the migration.

sbernhard:
> @sbernhard would that affect your workflows in any way?

first of all, thank you very much that you think about "us". 
No, it doesn't affect us.

evgeni:
What would pull in these new packages? Or would we tell the users "install foo-scenario" as the main entry point?

ekohl:
The latter IMHO. The only question is the migration. One way to solve that is to have:
* `foreman-installer-base` all modules and shared config (hiera)
* `foreman-installer-scenario-$scenario` a package for each scenario (depends on `foreman-installer-base`
* `foreman-installer` a transitional meta package that depends on `foreman-installer-scenario-foreman`
* `foreman-installer-katello` a transitional meta package that depends on `foreman-installer-scenario-foreman` and `foreman-installer-scenario-katello` and `foreman-installer-scenario-content-proxy`.

Then if you want to produce a custom installer, you would create a package, make it depend on `foreman-installer-base` and include your scenario files.

The upside of it is that if you only have a single scenario installed, you don't need to pass `--scenario`. This simplifies instructions for users.

evgeni:
Today, Katello users have both `foreman-installer` and `foreman-installer-katello` installed, so the transitional `foreman-installer` package would pull in "the wrong" `foreman-installer-scenario-foreman`. On the other hand you say `f-i-k` should pull in three scenarios? Then you again need `--scenario`?

So we would make `kafo` (it's kafo who handles scenarios, right?) behave like this:
* if only one scenario is available, `--scenario` becomes optional (but accepted to allow old automation to work)
* if multiple are available (because multiple packages are installed), `--scenario` remains required as it is today?

ekohl:
> Today, Katello users have both `foreman-installer` and `foreman-installer-katello` installed, so the transitional `foreman-installer` package would pull in "the wrong" `foreman-installer-scenario-foreman`. On the other hand you say `f-i-k` should pull in three scenarios? Then you again need `--scenario`?

Correct, but from RPMs we can't decide which is the right scenario so we can't cleanly migrate. This is also the situation today for Katello users.

> So we would make `kafo` (it's kafo who handles scenarios, right?) behave like this:
> 
>     * if only one scenario is available, `--scenario` becomes optional (but accepted to allow old automation to work)
>     * if multiple are available (because multiple packages are installed), `--scenario` remains required as it is today?

Correct. This is how Kafo behaves (and always has):

https://github.com/theforeman/kafo/blob/35a70342e2cc99d8ee9d58129e91efae66e98187/lib/kafo/scenario_manager.rb#L98-L112

This is why you don't see `--scenario` anywhere in the Foreman manual (which only ships a single scenario).

My argument is to also bring this to Katello. For fresh installations an up front selection should be made by choosing the right RPM. For existing installations you will have multiple scenarios available but since there is already a selection, it's already an optional argument.

evgeni:
I can't remember when I deployed a pure Foreman last time without Ansible, so I never noticed that :rofl:

evgeni:
However, I must admit I don't have the time to work on the "split the scenarios in own packages" proposal right now.

ekohl:
Is this only relevant when we branch in downstream or continuous? In other words: is there value if we merge it now or can we postpone to 3.1?

evgeni:
This is only relevant for downstream consumers, so 3.0 won't benefit from this PR (but would from your proposal, as it'd be cleaner overall)

ekohl:
So that was what I was getting at. I'd be in favor of planning and executing my proposal in the 3.1 timeframe.

evgeni:
Closing then.

PR #6916, opened 2021-07-20
Update ruby-foreman-remote-execution{,-core} to 4.5.4/1.4.7

adamruzicka:
- Update ruby-foreman-remote-execution to 4.5.2
- Update ruby-foreman-remote-execution-core to 1.4.5


adamruzicka:
`ruby-foreman-remote-execution-core to-1.4.5` should need `rbnacl` and `rbnacl-libsodium` gems. Should they be packaged separately or can they be bundled somehow? Also, for 2.6+ it won't be in rex core but in smart_proxy_remote_execution_ssh

PR #6913, opened 2021-07-20
Add rubygem-rbnacl package

adamruzicka:
Different versions of net-ssh have different dependencies for ed25519 key
support. Net-ssh 4.2 line needs rbnacl and rbnacl-libsodium.

- Add rubygem-rbnacl package
- Add rubygem-rbnacl-libsodium package

CPs:
- [ ] 2.5

PR #6922, opened 2021-07-21
Add rubygem-rbnacl-libsodium package

adamruzicka:
Requires https://github.com/theforeman/foreman-packaging/pull/6913


evgeni:
rebase please ;)

evgeni:
```
cp: cannot stat './usr/lib64/gems/ruby/rbnacl-libsodium-1.0.16/*.so': No such file or director
```

:(

evgeni:
So just that nobody has to do the same debugging as I did:

`rbnacl-libsodium` is not *really* a gem with native extensions, its more of a gem with a vendored copy of `libsodium`, so there is no direct `rbnacl-libsodium.so` or anything similar.

It's more like a `rbnacl/libsodium.rb` that finds the right `libsodium.so` and loads that:
https://github.com/RubyCrypto/rbnacl-libsodium/blob/efffe38b37077fe63e2323881b47374c3c41935a/lib/rbnacl/libsodium.rb#L6-L22

However, it doesn't, at least not on my Fedora, and also not on Koji (for different reasons, maybe).

One of the reasons is that 1.0.16 is missing https://github.com/RubyCrypto/rbnacl-libsodium/commit/9288ae7bb7e1403e10af96a4dab608090204344d which fixes finding the .so file (and then maybe https://github.com/RubyCrypto/rbnacl-libsodium/commit/795a6987f4586269281c6003a0b8aeeb3d99cd7)

Given https://github.com/RubyCrypto/rbnacl-libsodium is archived and explicitly documented as being obsolete, I think I'd prefer if we would just use libsodium directly, and not some old, vendored copy of it.

@adamruzicka what is needed for us to use a native libsodium (whereever we will get it)?

adamruzicka:
> what is needed for us to use a native libsodium (whereever we will get it)?

I first tried going with only `rbnacl` and system libsodium but had an issue with `rbnacl` not being able to find the right libsodium so and I had to help it by placing a symlink somewhere into /lib64/. Later I removed the symlink and tried `rbnacl-libsodium` which worked out of the box for me. Now I removed `rbnacl-libsodium`, but I'm failing to reproduce the original issue.

So in case the original issue doesn't appear again, all we'd need to do would be to update smart_proxy_remote_execution_ssh to not depend on `rbnacl-libsodium`, make a new release and then make it depend on `libsodium` (or `libsodium23` on debs) in packaging. And somehow deal with the fact that for EL* libsodium is only available in epel

evgeni:
Okay, that doesn't sound tooo weird. Even if I'd prefer to pull things from epel and build ourself.

adamruzicka:
Well, turns out the ancient version of net-ssh we ship requires ancient version of rbnacl, which expects `libsodium.so` to exist. Manually symlinking it to `libsodium.so.23` seems to work. The patch[1] that fixes this is rather small.

We'd either need to do that in our packaging and in every CI that pulls in rex core (or smart_proxy_remote_execution_ssh, depending on the version) or somehow patch rbnacl in packaging and CI. Really hope there's some thrid option that I'm not seeing.

[1] - https://github.com/RubyCrypto/rbnacl/commit/38c1bcd2fe7a3735c52fdd9dd9127814cbcc44e6

adamruzicka:
We plan to get rid of net-ssh https://github.com/theforeman/foreman-packaging/pull/6951 rather soon, so every solution until that happens will be a temporary one. In the end we decided to carry the necessary patch in packaging

pcreech:
[test rpm]

adamruzicka:
Since we'll be shipping patched rbnacl, this won't be necessary anymore. Closing

PR #6923, opened 2021-07-21
add ruby-version and coveralls.yml to gem2rpm ignores

evgeni:
inspired by #6913


ekohl (approved):
I opened https://github.com/fedora-ruby/gem2rpm/pull/107 to upstream some of these changes we have.

PR #6934, opened 2021-07-23
Update Katello requirement on Foreman

evgeni:


evgeni:
repoclosure is complaining about the package already in the archive

ekohl (approved):

PR #6942, opened 2021-07-23
Disable automatic dbgsym packages for foreman

ekohl:
This takes a long time and is generally not needed since Foreman doesn't ship compiled code with debug symbols.

For now this really is just an experiment.


ekohl:
This is an alternative to https://github.com/theforeman/foreman-packaging/pull/6644 but doesn't appear to work either.

PR #6951, opened 2021-07-27
Patch RbNaCl to find system libsodium

adamruzicka:
The patch is taken from this PR
https://github.com/RubyCrypto/rbnacl/pull/186.

CPs:
- [ ] 2.5


adamruzicka:
@pcreech could you take a look if I'm not shooting myself in the leg too much with this?

ekohl (commented):
I think you need to modify the `%setup` step to get it to apply. There are examples elsewhere in foreman-packaging which use %patch.

adamruzicka:
Huh, so rpms are not as magical as I hoped they are. Will fix

pcreech:
They're not as magical at least in the way we do them (with setup vs autosetup). 

Can confirm @adamruzicka that the rpm contains the fix.  And this is exactly how we would do it in any other case.  LGTM

pcreech (approved):

PR #6958, opened 2021-07-29
Add katello-host-tools for debian/ubuntu

sbernhard:


sbernhard:
I think, I will need help to finish this successfully.

pcreech:
[test deb]

sbernhard:
I guess, this needs at least https://github.com/theforeman/jenkins-jobs/pull/152
Additonally, katello-host-tools need to be pushed to pip. What are your thoughts @jturel

jturel:
> Additonally, katello-host-tools need to be pushed to pip. What are your thoughts @jturel

That seems like more work. Why do we need to do that?

evgeni:
We don't *require* it, but it makes things easier ;)
As we fetch gems from rubygems, we fetch python stuff from pypi.

sbernhard:
Somehow the source needs to be provided. Or should the source just be made avilable via git-annex? Currently I dont know how this can be done in deb build env.

jturel:
> We don't _require_ it, but it makes things easier ;) As we fetch gems from rubygems, we fetch python stuff from pypi.

Do we have a "foreman" org over there or anything, or automation to do the pushing? Trying to avoid having to do that by hand - can you blame me?

ekohl:
Take https://github.com/theforeman/obal/blob/master/.github/workflows/release.yml for example. If you push a tag, automation will push it to PyPI.

evgeni:
[test deb]

evgeni:
well, it failed, so at least it did try to build something :see_no_evil:

evgeni:
[test deb]

evgeni:
now the failure is legit, there is no `katello-host-tools` on PyPI right now.

sbernhard:
> now the failure is legit, there is no `katello-host-tools` on PyPI right now.

@jturel How do we want to handle this? The "easiest" solution is, to put katello-host-tools on PyPi. Otherwise we need to find another solution to provide the source.

evgeni:
the other alternative is we implement a kind of "source_url" entry in the build_vars, and make that point at `https://github.com/Katello/katello-host-tools/archive/refs/tags/$VERSION.tar.gz` or similar.

But I would prefer PyPI as it has a rather standardized interface (which we already implement)

jturel:
> > now the failure is legit, there is no `katello-host-tools` on PyPI right now.
> 
> @jturel How do we want to handle this? The "easiest" solution is, to put katello-host-tools on PyPi. Otherwise we need to find another solution to provide the source.

I guess that's OK? I don't have any experience with putting a project there. If one of us does it first, are we able to grant access to others to do so as well? ie I don't want to be the only person who can perform a release to pypi :)

ekohl:
You can use https://github.com/theforeman/obal/blob/master/.github/workflows/release.yml. It may be better if a GH admin sets the secret for you. I'm not a Katello GH admin so I can't do that. @evgeni?

evgeni:
I am not an admin either.

ekohl:
So I added an API token to the project and opened https://github.com/Katello/katello-host-tools/pull/123. While looking at the code I think `bin/` won't be part of the resulting tarball. See [obal](https://github.com/theforeman/obal/blob/a638b614bb4ecb31b490b49114a3aacb2cf0f3d6/setup.py#L74-L78) for an example of how to do this. It maps a command to be generated to a function in the code (in this case [obal.main](https://github.com/theforeman/obal/blob/a638b614bb4ecb31b490b49114a3aacb2cf0f3d6/obal/__init__.py#L66-L70)).

@sbernhard could you pick up this effort?

sbernhard:
> dit: or if you don't need `bin/*` at all, there's no work to be done.

Well, the files in /bin/* are required and the zypper / apt / yum plugins, too. Actually, I don't know the purpose of setup.py regarding creating a simple tarball.

ekohl:
> Well, the files in /bin/* are required and the zypper / apt / yum plugins, too. Actually, I don't know the purpose of setup.py regarding creating a simple tarball.

Try `python setup.py sdist`. You'll end up with a tarball that's uploaded to PyPI. The old way is listing those `bin/* files in `scripts` in `setup.py`. The modern way is to ship the code inside the package itself and using the console entry points. Then an appropriate wrapper is generated during installation.

ekohl:
Since there has been no progress on this, I took a stab at this and opened https://github.com/Katello/katello-host-tools/pull/125. That will likely also have packaging consequences, both RPM and deb side.

ekohl (changes requested):
That has been released and 4.0.1 should be available on PyPI. Please update the PR to reflect that.

ekohl:
I took a look at the RPM and I realized we don't ship all the additional files (etc, extra) in the tarball on PyPI. That's probably something to address.

sbernhard:
Need [https://github.com/Katello/katello-host-tools/pull/130](https://github.com/Katello/katello-host-tools/pull/130) before we can continue.

sbernhard:
[test deb]

evgeni:
I started working on this again in https://github.com/theforeman/foreman-packaging/pull/11027 :)

PR #6963, opened 2021-07-29
Add rubygem-rexml package

jlsherrill:


jlsherrill:
reference https://github.com/theforeman/foreman_maintain/pull/499

comps/comps-foreman-plugins-el8.xml
```diff
@@ -183,6 +183,7 @@
         Do not edit this section manually and use ./comps_doc.sh script
       -->
       <!--RGD-START-->
+      <packagereq type="default">rubygem-acts_as_list-doc</packagereq>
```
jlsherrill, line 186: the tooling added this, i assume it was missed previously, i can remove it if desired

ekohl (commented):
REXML is built into Ruby 2.x. Why is it needed to package this? Do you need a newer version?

PR #6967, opened 2021-08-02
Update rubygem-foreman_salt to 14.0.0

bastian-src:


ekohl:
ok to test

pcreech:
is this still WIP?

bastian-src:
Thanks for the reply @pcreech!
It depends on the following PRs:

* [foreman-packaging: Smart Proxy Salt](https://github.com/theforeman/foreman-packaging/pull/6968)
* [foreman: Provisioning Template change](https://github.com/theforeman/foreman/pull/8489)

We need to adapt `foreman_min_version` depending on the foreman-PR and then, this should be merged together with the foreman-packaging-PR of Smart Proxy Salt.

evgeni:
I am trying to wrap my head around this (also because @sbernhard opened a PR for Debian, and didn't mark it WIP at all: https://github.com/theforeman/foreman-packaging/pull/7422).

The changes in https://github.com/theforeman/foreman/pull/8489 look like they would *benefit* from the new features in `foreman_salt` 14, but should this really be blocking? Shouldn't `host_param('salt_grains')` still work, just without the new and fancy autosigning?

packages/plugins/rubygem-foreman_salt/rubygem-foreman_salt.spec
```diff
@@ -4,12 +4,12 @@
 
 %global gem_name foreman_salt
 %global plugin_name salt
-%global foreman_min_version 1.24.0
+%global foreman_min_version 2.5.2
```
ekohl, line 7: This is a very specific number. Why this version instead of just 2.5?

bastian-src: I just set it to the most up-to-date version when I started the PR, since the Autosign procedure (which is part of the new Foreman Salt version) needed [this](https://github.com/theforeman/foreman/commit/ce2190057571ad3d9c3db3f6a80b9dcf52448426) commit to be part of Foreman in order to work properly. I wanted to update the `foreman_min_version` when the corresponding PR got merged.

It has been merged by now, but there isn't a new Foreman release which includes it. So, I guess we have to wait until 3.2 is released, right?

bastian-src: Alright, thanks! I set it to 2.5 now which is the version I tested it for initially.

ekohl (approved):
Is this still WIP as the PR title suggests?

bastian-src:
Nope, thanks for the reminder!

ekohl:
Ok, can this be merged without the Smart Proxy side or do they need to be merged together?

bastian-src:
They should be merged simultaneously, since their features depend on each other.

ekohl:
Note that we generally try to support an older Smart Proxy for upgrades. This means users update their Foreman first and then all additional Smart Proxies. If you do need to differentiate or provide clear errors then you can expose capabilities on the Smart Proxy. I've written https://theforeman.org/2019/04/smart-proxy-capabilities-explained.html about this. Then on the Foreman side you can use [`smart_proxy.has_capability?`](https://github.com/theforeman/foreman/blob/0a49d3ff5c9833d7263e0fc42e72ae04b87165a9/app/models/smart_proxy.rb#L104-L106). This allows you to show an error message like `"{smart_proxy} lacks capability X, upgrade to a later version"` or similar.

In this case you would probably name it `autosign_key` and introduced it in https://github.com/theforeman/foreman/blob/0a49d3ff5c9833d7263e0fc42e72ae04b87165a9/app/models/smart_proxy.rb#L104-L106.

I'm not pushing for that here, but those are techniques we've developed to provide a smoother user experience. Hopefully you're aware of them for a future scenario.

PR #6968, opened 2021-08-02
Update smart_proxy_salt to 4.0.0

bastian-src:


packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -23,19 +23,23 @@
...
 %global foreman_proxy_dir %{_root_datadir}/foreman-proxy
 %global foreman_proxy_statedir %{_root_localstatedir}/lib/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
 %global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
 
 %global salt_config_dir %{_root_sysconfdir}/salt
+%global salt_config_runners_dir /var/lib/foreman-proxy/salt/runners
+%global salt_config_reactors_dir /var/lib/foreman-proxy/salt/reactors
+%global salt_config_grains_dir /var/lib/foreman-proxy/salt/grains
```
ekohl, line 36: Please reuse %{foreman_proxy_statedir}
```suggestion
%global salt_config_runners_dir %{foreman_proxy_statedir}/salt/runners
%global salt_config_reactors_dir %{foreman_proxy_statedir}/salt/reactors
%global salt_config_grains_dir %{foreman_proxy_statedir}/salt/grains
```

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -97,14 +102,6 @@ gem unpack %{SOURCE0}
 
 %setup -q -D -T -n  %{gem_name}-%{version}
 
-%if %{with python2} && 0%{?rhel} >= 8
-sed -i -e '1s|^#!.*$|#!%{__python2}|' sbin/upload-salt-reports
-%endif
-
-%if %{with python3} && 0%{?rhel} >= 8
-sed -i -e '1s|^#!.*$|#!%{__python3}|' sbin/upload-salt-reports
-%endif
```
ekohl, line 106: Why was the old code replaced with a shell wrapper? IMHO this is a huge step back. In packaging we already know what the Python version should be.

ekohl (changes requested):

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
bastian-src: With the shell wrapper, we can distinguish between the Python version when the package is installed already. This makes Py2 to Py3 migrations easier: the user won't need to reinstall Smart Proxy Salt since the wrapper just uses the version which works.
With the current solution, the package must be reinstalled to enable Python3.

What do you think about it?

ekohl: That the previous approach was the correct one. You already know what the platform uses. On EL7 you will use Python 2 regardless. On EL8 it will always be Python 3. Trying to determine it on every single request is just overhead. So for packaged installations I'd stick with what's there since it's the right way to do it.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -23,19 +23,23 @@
...
 %global foreman_proxy_statedir %{_root_localstatedir}/lib/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
 %global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
 
 %global salt_config_dir %{_root_sysconfdir}/salt
+%global salt_config_runners_dir %{foreman_proxy_statedir}/salt/runners
+%global salt_config_reactors_dir %{foreman_proxy_statedir}/salt/reactors
+%global salt_config_grains_dir %{foreman_proxy_statedir}/salt/grains
+%global salt_minion_auth_install_dir %{buildroot}%{gem_instdir}/salt/minion_auth
```
ekohl, line 37: You only use this in the build step. I'm not sure a separate macro makes that much sense.

bastian-src: File permissions are set [here](https://github.com/theforeman/foreman-packaging/pull/6968/files#diff-7f2b45922f25a10c95613ac59ae09ca985b02817e57bc926206b8815a454747aR194) as well, so its used at least two times. IMHO macros give a better overview of the paths used and reduce redundancy. Is there any downside of using them?

ekohl: I was just talking about this single line. For the others (salt_config_*) I agree they do make sense.

bastian-src: Implemented as suggested.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
bastian-src: Implemented as suggested.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -177,18 +190,26 @@ EOF
...
 %exclude %{gem_cache}
 %{gem_spec}
-%{foreman_proxy_bundlerd_dir}/smart_proxy_salt_core.rb
 %{smart_proxy_dynflow_bundlerd_dir}/smart_proxy_salt_core.rb
 %config(noreplace) %{salt_config_dir}/foreman.yaml
 %config %{_root_sysconfdir}/cron.d/%{gem_name}
 %exclude %{gem_instdir}/etc
 %exclude %{gem_instdir}/cron
+%attr(-,foreman-proxy,foreman-proxy) %{salt_config_runners_dir}/foreman_file.py*
+%attr(-,foreman-proxy,foreman-proxy) %{salt_config_runners_dir}/foreman_https.py*
```
ekohl, line 199: It's odd to ship some .py file in the runners dir (which is in `/var/lib`). https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html summarizes it:
> This hierarchy holds state information pertaining to an application or the system. State information is data that programs modify while they run, and that pertains to one specific host.
So the runners directory should really be application state only. RPMs are by definition static. So what you typically see is that RPMs ship the state directory itself and permissions, but leave it empty for the application to fill.

It also means the application itself can modify its execution script. While unlikely, it opens the door to security exploits. Overall something I'd strongly discourage.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -155,9 +170,7 @@ mkdir -p %{buildroot}%{_root_sbindir}
 mv %{buildroot}/%{gem_instdir}/sbin/upload-salt-reports %{buildroot}%{_root_sbindir}/upload-salt-reports
 mv .%{gem_instdir}/cron/smart_proxy_salt %{buildroot}%{_root_sysconfdir}/cron.d/%{gem_name}
 mkdir -p %{buildroot}%{smart_proxy_dynflow_bundlerd_dir}
-cat <<EOF | tee %{buildroot}%{smart_proxy_dynflow_bundlerd_dir}/smart_proxy_salt_core.rb \
-                %{buildroot}%{foreman_proxy_bundlerd_dir}/smart_proxy_salt_core.rb \
-                >/dev/null
+cat <<EOF > %{buildroot}%{smart_proxy_dynflow_bundlerd_dir}/smart_proxy_salt_core.rb
```
ekohl, line 173: Why is this still here if `smart_proxy_salt_core` was dropped?

ekohl (changes requested):
Shouldn't this also drop the dependency on `smart_proxy_dynflow_core` around line 58?

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
bastian-src: It's not the _final_ runners directory which is used during runtime. According to the manual provided [here](https://github.com/theforeman/smart_proxy_salt/tree/master/salt/minion_auth), the Python runners have to be moved to the appropriate location which depends heavily on the Salt configuration.

The `/var/lib` is just used as a temporal storage directory when it comes to the Python runners. What would be a better directory for this?  
I thought it would be more consistent to have everything in one place like `/var/lib/foreman-proxy/salt/[grains|runners|reactors]`.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
bastian-src: Fixed as suggested.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
ekohl: This is wild. It looks like you're mixing configuration, code and state.

If an admin is supposed to modify a config file, it should live in `/etc`. That means I'd expect `foreman_minion_auth.sls` to live there.

Runners look like code (and the application shouldn't change them, right?) so I'd expect them to live somewhere in `/usr`. It depends on how the files are used. If it's executed, `/usr/libexec` might be a good place. If it's a Python library, the Python site-packages is probably the right place. If it's a custom format, perhaps `/usr/share` is a better place.

Then the state should live in `/var/lib`. A good way to think about it is that you can run `rm -rf /var/lib/foreman-proxy/* && systemctl restar foreman-proxy`. Then it should still be able to work. It'd have lost all its state so essentially be a clean installation again but what you are doing now implies they also need to run `yum reinstall $package`. That's wrong.

bastian-src: Okay, thanks for the state/`var/lib` example. I got it now and I agree, the reactor file (`foreman_minion_auth.sls`) and the Python Salt runners should not be placed there.

Still, I would refer to both of them as _code_. The reactor is not a configuration file - it is called by the Salt event system and actually interpreted/executed. Since a webhook is performed in there, it has to be configured in terms of setting the right hostname beforehand. It would be nice if `foreman-installer` could take care of this instead of some administrator doing it manually. Do you think this would be good/possible?

The Python Salt Runners have to be placed in a specific directory depending on the Salt Master configuration. This has to be done by the administrator manually, since it depends on the existing Salt setup. When they've been copied to the appropriate location by hand & the `salt-run saltutil.sync_all` command is executed, the runners are copied to the corresponding Salt Python `site-packages` directory.  
According to this, placing them during the rpm installation under some `/usr/libexec` so that the administrator can move the runners to the right Salt configuration dir & having them copied by Salt itself back into some `/usr/lib/python/site-packages/` folder doesn't sound like a good design to me. We would have the runners lying in `/usr/libexec` where they aren't executed plus they would exist in `/usr/lib/python/site-packages` where they are used indeed.  
IMHO, we just need some kind of static place where they can be stored, such that the administrator can copy/move them into the right directory so they are recognized by `salt-run saltutil.sync_all`. Even if `/tmp` could fit the needs, I would like to avoid it since it would break the post-installation setup process if the machine restarts.

bastian-src: Coming back to this, it seems like `/var/lib` is not the appropriate place to store Salt reactors and runners.

To summarize:  
Salt reactors are interpreted by the Salt runtime and are used to react on events (for example: the installation of some package has been finished). One of the reactors we want to use can be found [here](https://github.com/theforeman/smart_proxy_salt/blob/master/salt/report_upload/foreman_report_upload.sls). It reacts on the Salt _highstate_ event and triggers the execution of the `foreman_report_upload` runner which can be found [here](https://github.com/theforeman/smart_proxy_salt/blob/master/salt/report_upload/srv/salt/_runners/foreman_report_upload.py). It is a custom Salt runner which we want to ship into the Salt runtime environment.

In order to enable the usage of this (custom) runner within the reactor execution, the runner must be included into the Salt runtime beforehand. Therefore, the runner has to be placed in the appropriate folder (`/srv/salt/_runners/` by default, but it depends on the corresponding configuration) and the `salt-run saltutil.sync_all` command has to be executed. This step has to be done manually and must be part of the Salt installation manual - at least I don't see a way to automate this nicely.

@ekohl what do you think about creating the following folders during the rpm installation and placing the files there:
```
/usr/share/foreman-proxy/salt/runners/
/usr/share/foreman-proxy/salt/reactors/
```

After the installation, the admin has to copy the runners from `/usr/share/foreman-proxy/salt/runners/` to the corresponding runners directory depending on their configuration & run the salt command as described previously.

pcreech:
@ekohl are you still evaluating this one?

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
evgeni: `/usr/share/foreman-proxy/salt` is probably a good compromise.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -177,18 +179,25 @@ EOF
...
 %{gem_spec}
-%{foreman_proxy_bundlerd_dir}/smart_proxy_salt_core.rb
-%{smart_proxy_dynflow_bundlerd_dir}/smart_proxy_salt_core.rb
 %config(noreplace) %{salt_config_dir}/foreman.yaml
 %config %{_root_sysconfdir}/cron.d/%{gem_name}
 %exclude %{gem_instdir}/etc
 %exclude %{gem_instdir}/cron
+%attr(-,foreman-proxy,foreman-proxy) %{salt_proxy_runners_dir}/foreman_file.py*
+%attr(-,foreman-proxy,foreman-proxy) %{salt_proxy_runners_dir}/foreman_https.py*
+%attr(-,foreman-proxy,foreman-proxy) %{salt_proxy_reactors_dir}/foreman_minion_auth.sls
```
evgeni, line 188: why do these files need to be owned by `foreman-proxy`?

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -177,18 +179,25 @@ EOF
...
+%attr(-,foreman-proxy,foreman-proxy) %{salt_state_grains_dir}/autosign_key
```
evgeni, line 189: does this file need to be writable by the proxy, because it's used as a mechanism to share data between `smart_proxy_salt` and the salt master?

evgeni:
ok to test

evgeni:
```
Checking for unpackaged file(s): /usr/lib/rpm/check-files /builddir/build/BUILDROOT/rubygem-smart_proxy_salt-4.0.0-1.fm3_2.el8.noarch
   /usr/bin/salt_python_wrapper
   /usr/bin/salt_python_wrapper
```

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
bastian-src: Exactly, `smart_proxy_salt` adds entries to the file and salt master reads the file during minion authentication.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
bastian-src: I checked it and it's actually not necessary. Should these lines be left out completely, then?

evgeni: yes please

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -23,19 +23,22 @@
 %global gem_name smart_proxy_salt
 %global plugin_name salt
 
-%global foreman_proxy_min_version 1.25
+%global foreman_proxy_min_version 2.5
 %global foreman_proxy_dir %{_root_datadir}/foreman-proxy
 %global foreman_proxy_statedir %{_root_localstatedir}/lib/foreman-proxy
 %global foreman_proxy_bundlerd_dir %{foreman_proxy_dir}/bundler.d
 %global foreman_proxy_settingsd_dir %{_root_sysconfdir}/foreman-proxy/settings.d
 %global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d
```
evgeni, line 31: you can drop this line, it's unused now.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
evgeni: @bastian-src you *do* need to include those dirs, just no need to set the permissions

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
bastian-src: So, [this](https://github.com/theforeman/foreman-packaging/pull/6968/files#diff-7f2b45922f25a10c95613ac59ae09ca985b02817e57bc926206b8815a454747aR59) line can be dropped too?

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -155,11 +162,6 @@ mkdir -p %{buildroot}%{_root_sbindir}
...
 mv .%{gem_instdir}/cron/smart_proxy_salt %{buildroot}%{_root_sysconfdir}/cron.d/%{gem_name}
 mkdir -p %{buildroot}%{smart_proxy_dynflow_bundlerd_dir}
-cat <<EOF | tee %{buildroot}%{smart_proxy_dynflow_bundlerd_dir}/smart_proxy_salt_core.rb \
-                %{buildroot}%{foreman_proxy_bundlerd_dir}/smart_proxy_salt_core.rb \
-                >/dev/null
-gem 'smart_proxy_salt_core'
-EOF
 
 %files
 %dir %{gem_instdir}
```
evgeni, line 167:
```suggestion
%dir %{gem_instdir}
%{_root_bindir}/salt_python_wrapper
```

or probably a line later (after `foreman-node`) so that it's sorted properly.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
evgeni: No?

just the ` %global smart_proxy_dynflow_bundlerd_dir %{_datadir}/smart_proxy_dynflow_core/bundler.d` line is not necesary anymore, given that you don't use dynflow_core anymore.

bastian-src: Sure, sry! The highlighting of my own change irritated me. It would've wondered me if the `min_version` wasn't necessary anymore.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
bastian-src: I've just created a [PR](https://github.com/theforeman/smart_proxy_salt/pull/73/files) to replace the `salt_python_wrapper` with a single line, given by an example from your blog. But I guess we'll need a new release then.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -177,18 +179,22 @@ EOF
 %{foreman_proxy_bundlerd_dir}/%{plugin_name}.rb
 %exclude %{gem_cache}
 %{gem_spec}
-%{foreman_proxy_bundlerd_dir}/smart_proxy_salt_core.rb
-%{smart_proxy_dynflow_bundlerd_dir}/smart_proxy_salt_core.rb
 %config(noreplace) %{salt_config_dir}/foreman.yaml
 %config %{_root_sysconfdir}/cron.d/%{gem_name}
 %exclude %{gem_instdir}/etc
 %exclude %{gem_instdir}/cron
+%attr(-,foreman-proxy,foreman-proxy) %{salt_state_grains_dir}/autosign_key
```
evgeni, line 186:
```suggestion
%{salt_proxy_runners_dir}
%{salt_proxy_reactors_dir}
%attr(-,foreman-proxy,foreman-proxy) %{salt_state_grains_dir}/autosign_key
```

otherwise the runners and reactors in `/usr/share` are not picked up.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
ekohl: I fear that as it is now a `yum reinstall` or `yum update` will clear this file back to its packaged state. Wouldn't it be better to have it create the directory (with the correct permissions) and expect the Smart Proxy to create the file at runtime?
```suggestion
%dir %attr(-,foreman-proxy,foreman-proxy) %{salt_state_grains_dir}
```

You may also need to add the `autosign_key` file as a `%ghost` to ensure it's properly owned.

I'm also looking at @evgeni here for advice on what's best, but I think the replacement on upgrades is a real issue.

evgeni: oh right, this is outside of `/etc`, so RPM will replace it, good catch!
Yeah, I think just creating the dir and `%ghost`ing the file is the correct approach here from a packaging perspective.

But! ;-)

the proxy plugin explicitly doesn't create this file, but expects it to be present: https://github.com/theforeman/smart_proxy_salt/blob/ddaef0df69c850015a6ec98e425dd58b9339c760/lib/smart_proxy_salt/cli.rb#L67

I don't see a good reason for it not to create it, but maybe I don't see the whole picture with salt…

ekohl: I'm not sure why it fails on that: `open(filepath, File::RDWR)` should be able to create it, right?

@bastian-src Some unrelated (to this PR) comments that I saw while reading through that file. I'd recommend this pattern to guarantee the file descriptor is closed:
```ruby
open(filepath, File::RDWR) do |file|
  found = false
  file.each_line { |line| found = true if line.chomp == value }
  file.puts value unless found
end
```

I also like to use `File.open` instead of plain `open`. This can avoid accidentally exposing much more. For example, `open()` can also be used to open a subshell (by starting it with `|`). That can escalate a vulnerability from writing to arbitrary files to a remote code execution.

Also, in https://github.com/theforeman/smart_proxy_salt/blob/ddaef0df69c850015a6ec98e425dd58b9339c760/lib/smart_proxy_salt/cli.rb#L89 you're leaking a file descriptor since it's never closed. You can use `File.read()` as well if you just want to read the file completely.

And I think https://github.com/theforeman/smart_proxy_salt/blob/ddaef0df69c850015a6ec98e425dd58b9339c760/lib/smart_proxy_salt/cli.rb#L101-L103 could be replaced with `File.write(filepath, entries.join())`.

bastian-src: > I don't see a good reason for it not to create it, but maybe I don't see the whole picture with salt…

The `autosign_key` file contains an authentication key during the host creation which is used to authenticate the Salt Minion against the Salt Master initially. Afterwards, the key is removed from the file.
Accordingly, we would only run into problems if a package update/reinstall is performed during the deployment of a host. Then, the Salt Minion wouldn't be authenticated automatically and would've to be accepted manually.

I don't think this is a scenario we should actually care about too much/support at all.
Still, do you think it would be better to let Smart Proxy Salt create the `autosign_key` file instead of foreman-packaging?

@ekohl thanks for the advice! I'll have a look at the file.

ekohl: I think it would be better if Smart Proxy Salt could create it. The major benefit is that a user can remove the file and it can recover from it. Right now it will just "crash" requiring manual admin intervention.

And I do think it's important to deal with this. It would feel unreliable to the user. Losing data is always bad and it makes the user question what else it will just throw away.

bastian-src: Alright, I agree that it enhances the user experience in case the file doesn't exist for some reason.

I've created [this](https://github.com/theforeman/smart_proxy_salt/pull/74) PR and tested it right away.
Unfortunately, I'm getting a `Permission denied @ rb_sysop` error. Could this be related to SELinux?

bastian-src:
Can you tell me where the actual error occurs now? The processes in [root.log](http://koji.katello.org/kojifiles/work/tasks/3108/473108/root.log) finish with `0`.

ekohl (approved):
I think this is now correct, but I don't have the environment to correctly test this.

sbernhard:
@ekohl I guess, we need to put everything together so that we can test everything. Like https://github.com/theforeman/puppet-foreman_proxy/pull/696

ekohl:
Looks like the build was rotated but I want to inspect it so rebuilding.
[test rpm]

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -177,18 +179,25 @@ EOF
...
-%{foreman_proxy_bundlerd_dir}/smart_proxy_salt_core.rb
-%{smart_proxy_dynflow_bundlerd_dir}/smart_proxy_salt_core.rb
 %config(noreplace) %{salt_config_dir}/foreman.yaml
 %config %{_root_sysconfdir}/cron.d/%{gem_name}
 %exclude %{gem_instdir}/etc
 %exclude %{gem_instdir}/cron
+%{salt_proxy_runners_dir}
+%{salt_proxy_reactors_dir}
+%dir %attr(-,foreman-proxy,foreman-proxy) %{salt_state_grains_dir}
+%ghost %{salt_state_grains_dir}/autosign_key
```
ekohl, line 189: By marking it as %ghost it looks like it's not included in the final RPM. 4.0.0 can't deal with that, so this is still something that should be addressed.

bastian-src: So, do we include it without ghosting the file in 4.0.0 and make another release 4.0.1? That 4.0.1 release ghosts the file, does not include it in the RPM direclty, and creates it programmatically with the Smart Proxy instead.

ekohl:
On a related note: https://github.com/theforeman/smart_proxy_salt/tags only lists 2.1.9 as the latest tag. Please try to see if you still have those tags locally. I can recommend [--follow-tags](https://git-scm.com/docs/git-push#Documentation/git-push.txt---follow-tags) and have configured my git to always do that.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
ekohl: The alternative is to add a `%post` section:
```spec
%post
if [ ! -f %{salt_state_grains_dir}/autosign_key ] ; then
  touch %{salt_state_grains_dir}/autosign_key
  chmod 644 %{salt_state_grains_dir}/autosign_key
  chown foreman-proxy:foreman-proxy %{salt_state_grains_dir}/autosign_key
fi
```

I'm not sure the chmod is needed, though for security I'd expect it to not be world readable. key implies there are secrets in there.

I also think right now you don't have the correct ownership on the file and it ends up being unwriteable for the `foreman-proxy` user.

bastian-src: LGTM, especially to add the permissions with chown and manage access. Previously, the user access was set when the folder has been created.

bastian-src:
Regarding the tags: I don't have local tags as well, I guess they've been just forgotten after that 2.1.9 release. I'll create new tags for upcoming versions.

ekohl:
Have any your colleagues made the tags by any chance? Perhaps @sbernhard? It would be nice to have them for a correct history.

ekohl (approved):
This looks ok to me. There is a merge conflict now, but once that's resolved I think it can be merged.

@evgeni could you also have a last look?

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -135,6 +130,17 @@ cp -a .%{_bindir}/* \
...
+# autosign runner, reactor, key file
+mkdir -p %{buildroot}%{salt_proxy_runners_dir}
+mv %{buildroot}%{gem_instdir}/salt/minion_auth/srv/salt/_runners/* \
+   %{buildroot}%{salt_proxy_runners_dir}
+mkdir -p %{buildroot}%{salt_proxy_reactors_dir}
+mv %{buildroot}%{gem_instdir}/salt/minion_auth/foreman_minion_auth.sls \
+   %{buildroot}%{salt_proxy_reactors_dir}/
+mkdir -p %{buildroot}%{salt_state_grains_dir}
+touch %{buildroot}%{salt_state_grains_dir}/autosign_key
+chmod 644 %{buildroot}%{salt_state_grains_dir}/autosign_key
```
evgeni, line 142: technically, the `chmod` is not required, as the perms won't be in the package anyways, thanks to `%ghost` (the `touch` *is* required, tho)

ekohl: As I said: https://github.com/theforeman/foreman-packaging/pull/6968#discussion_r780249542 I'm not sure it would be needed. However, it would deal with any umask. That said, I still wonder if it should 644 or 640. If there are secrets in there, it should not be world readable.

bastian-src: I agree, 640 should be preferred. I've removed the referenced `chmod` and adapted it in the post section [here](https://github.com/theforeman/foreman-packaging/pull/6968/commits/eee49308522161dcd9a34bf1155f894524f680aa).

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
```diff
@@ -23,18 +23,20 @@
...
 %global salt_config_dir %{_root_sysconfdir}/salt
+%global salt_proxy_runners_dir %{foreman_proxy_dir}/salt/runners
+%global salt_proxy_reactors_dir %{foreman_proxy_dir}/salt/reactors
+%global salt_state_grains_dir %{foreman_proxy_statedir}/salt/grains
 
 Summary: SaltStack support for Foreman Smart-Proxy
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 3.1.2
+Version: 4.0.0
 Release: 9%{?foremandist}%{?dist}
```
ekohl, line 40:
```suggestion
Release: 1%{?foremandist}%{?dist}
```

ekohl (commented):
I want to verify what `rpm -qV rubygem-smart_proxy_salt` does with regards to `autosign_key`. If it tracks the permissions at all. However, now it doesn't build.

packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec
bastian-src: Can you explain to me, why this change caused the error? Was it just incorrect to introduce the change in the first place?

ekohl: The logic is that version should always follow the source (in this case smart_proxy_salt) so you have no control over that. The release is there to track the "version" of the spec/package. This then results in `%{version}-%{release}`, here `4.0.0-1` vs `4.0.0-9`. This is taken into account when comparing. As you can see, increasing the release makes it newer even though the version stays the same. However, `4.0.0-1` is already newer than `3.1.2-9` and we don't want to keep increasing release. You would end with a value that's insanely high and doesn't say anything anymore. While technically it wouldn't matter (computers understand it just fine), it's harder for the humans. That's why we always reset the release to 1 when we bump the version.

bastian-src: Alright, thanks for the explanation! I'm gonna squash the commits and then, we are ready to merge, I guess?

ekohl:
I ran on an EL8 container:
```console
# dnf -y install wget https://yum.theforeman.org/releases/nightly/el8/x86_64/foreman-release.rpm
# dnf module enable ruby:2.7
# wget https://repo.saltproject.io/py3/redhat/8/x86_64/latest.repo -O /etc/yum.repos.d/salt.repo
# dnf -y install http://koji.katello.org/kojifiles/work/tasks/5762/475762/rubygem-smart_proxy_salt-4.0.0-1.fm3_2.el8.noarch.rpm
# rpm -ql rubygem-smart_proxy_salt
/etc/cron.d/smart_proxy_salt
/etc/foreman-proxy/settings.d/salt.saltfile
/etc/foreman-proxy/settings.d/salt.yml
/etc/salt/foreman.yaml
/usr/bin/foreman-node
/usr/bin/salt_python_wrapper
/usr/sbin/upload-salt-reports
/usr/share/foreman-proxy/bundler.d/salt.rb
/usr/share/foreman-proxy/salt/reactors
/usr/share/foreman-proxy/salt/reactors/foreman_minion_auth.sls
/usr/share/foreman-proxy/salt/runners
/usr/share/foreman-proxy/salt/runners/foreman_file.py
/usr/share/foreman-proxy/salt/runners/foreman_https.py
/usr/share/gems/gems/smart_proxy_salt-4.0.0
/usr/share/gems/gems/smart_proxy_salt-4.0.0/LICENSE
/usr/share/gems/gems/smart_proxy_salt-4.0.0/bin
/usr/share/gems/gems/smart_proxy_salt-4.0.0/bin/foreman-node
/usr/share/gems/gems/smart_proxy_salt-4.0.0/bin/salt_python_wrapper
/usr/share/gems/gems/smart_proxy_salt-4.0.0/bundler.d
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt.rb
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt/api_request.rb
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt/cli.rb
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt/rest.rb
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt/salt.rb
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt/salt_api.rb
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt/salt_http_config.ru
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt/salt_runner.rb
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt/salt_task_launcher.rb
/usr/share/gems/gems/smart_proxy_salt-4.0.0/lib/smart_proxy_salt/version.rb
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/minion_auth
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/minion_auth/README.md
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/minion_auth/master.snippet
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/minion_auth/srv
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/minion_auth/srv/salt
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/minion_auth/srv/salt/_runners
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/report_upload
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/report_upload/README.md
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/report_upload/master.snippet
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/report_upload/srv
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/report_upload/srv/salt
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/report_upload/srv/salt/_runners
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/report_upload/srv/salt/_runners/foreman_report_upload.py
/usr/share/gems/gems/smart_proxy_salt-4.0.0/salt/report_upload/srv/salt/foreman_report_upload.sls
/usr/share/gems/gems/smart_proxy_salt-4.0.0/settings.d
/usr/share/gems/specifications/smart_proxy_salt-4.0.0.gemspec
/var/lib/foreman-proxy/salt/grains
/var/lib/foreman-proxy/salt/grains/autosign_key
# ls -l /var/lib/foreman-proxy/salt/grains/autosign_key
-rw-r-----. 1 foreman-proxy foreman-proxy 0 Jan 14 18:32 /var/lib/foreman-proxy/salt/grains/autosign_key
# echo test > /var/lib/foreman-proxy/salt/grains/autosign_key
# dnf -y reinstall http://koji.katello.org/kojifiles/work/tasks/5762/475762/rubygem-smart_proxy_salt-4.0.0-1.fm3_2.el8.noarch.rpm
# cat /var/lib/foreman-proxy/salt/grains/autosign_key
test
# dnf -y remove rubygem-smart_proxy_salt
Error: 
 Problem: The operation would result in removing the following protected packages: sudo
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

The last bit is weird and I don't understand it, but IMHO it shouldn't hold up.

What does look odd is:
```
# rpm -q --requires rubygem-smart_proxy_salt
/bin/sh
/bin/sh
/etc/cron.d
/usr/bin/ruby
/usr/libexec/platform-python
config(rubygem-smart_proxy_salt) = 4.0.0-1.fm3_2.el8
foreman-proxy >= 2.5
python3
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
ruby
ruby(release)
ruby(rubygems)
ruby(rubygems)
rubygem(smart_proxy_dynflow) >= 0.5.0
rubygem(smart_proxy_dynflow) >= 0.5.0
salt-master
```

I believe this is due to the invalid shebangs in `/usr/share/foreman-proxy/salt/runners/foreman_file.py` and `/usr/share/foreman-proxy/salt/runners/foreman_https.py`. These are both `/usr/bin/env python` which is generally not allowed.

This also still has `/usr/share/gems/gems/smart_proxy_salt-4.0.0/bin/salt_python_wrapper` which I still think is a bad idea. In our packaging we don't need to probe: we explicitly know which Python should be used and can modify that.

IMHO we should drop that wrapper file and patch `sbin/upload-salt-reports` to use `#!/usr/bin/python2` (EL7) or `#!/usr/bin/python3` (EL8).

@evgeni what do you think?

evgeni:
Well, in reality, it's Python3 everywhere. See my investigation in https://github.com/theforeman/smart_proxy_salt/pull/73#issuecomment-994501133

And as to "knowing" - no we "don't", the idea is that the user can have a Py2-Salt or a Py3-Salt, *regardless* of the default Python of the OS. Well, or rather was, there is no more supported Py2-Salt at all.

evgeni:
I can explain the "The operation would result in removing the following protected packages: sudo" tho.

`sudo` is not installed by default in CentOS 8 containers, but `sudo` is defined as a protected package. Now your `dnf install rubygem-smart_proxy_salt` installed `sudo` as a depdency (via `salt` somewhere, I didn't check too closely). And `dnf remove rubygem-smart_proxy_salt` tries to remove all packages that were installed *as a dependency* of `rubygem-smart_proxy_salt`, well, poor `sudo` is on that list and freaks out `dnf`'s "protected" packages check.

ekohl:
I figured it was something like that. I tried to manually install sudo to get it in the list of installed packages so it wouldn't be removed, but that didn't work. It's probably not a problem on non-minimal installs. However, later I'll do some more testing to see if it properly cleans up the autosign_key file.

evgeni:
Yeah, dnf (contrary to e.g. apt) doesn't seem to mark packages as manually installed if you ask it to install them while they are already present because they are a dependency.

Anyways, not too important right now.

As for the python wrapper: I'd take it as-is right now and fix it cleaner in sp_salt directly.

bastian-src:
The shebang line of the runner scripts have been fixed lately (as part of [this](https://github.com/theforeman/smart_proxy_salt/pull/75) PR). It turns out that they should not contain a shebang at all (compare official runners [here](https://github.com/saltstack/salt/tree/develop/salt/runners)).
In order to exclude `salt_python_wrapper` for future releases, I've created a PR to remove it (see [here](https://github.com/theforeman/smart_proxy_salt/pull/76)). Is it appropriate to replace it with `/usr/bin/env python3` or should `/usr/bin/python3` be preferred for some reason?
I'm not sure since you mentioned that `/usr/bin/env python` is not allowed. @ekohl did you mean we shouldn't make use of `env` at all or just not with `python` as parameter?

evgeni:
`/usr/bin/env` shouldn't be used in *packaged* scripts, but as long as the Python after it is explicitly `python2` or `python3` RPM will translate this properly to `/usr/bin/pythonX`.

bastian-src:
Alright, thanks for the advice.
Is there anything else left for this PR?

evgeni (approved):

ekohl (approved):
Merging now but I'd like to encourage a 4.1.0 release that builds on this. Looks like there are already good fixes in smart_proxy_salt git master.

PR #6971, opened 2021-08-02
Add rubygem-hammer_cli_foreman_puppet package

amirfefer:
Supported Versions:
 * Nightly


packages/plugins/rubygem-hammer_cli_foreman_puppet/rubygem-hammer_cli_foreman_puppet.spec
```diff
@@ -0,0 +1,95 @@
...
+Group: Development/Languages
+License: GPLv3
+URL: https://github.com/theforeman/hammer-cli-foreman-puppet
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+# start specfile generated dependencies
+Requires: %{?scl_prefix_ruby}ruby(release)
+Requires: %{?scl_prefix_ruby}ruby
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
+Requires: %{?scl_prefix}rubygem(hammer_cli_foreman) >= 3.0.0.pre.develop
```
ekohl, line 24: This is now how we version the RPM. I think this should work:
```suggestion
Requires: %{?scl_prefix}rubygem(hammer_cli_foreman) >= 3.0.0
```

packages/plugins/rubygem-hammer_cli_foreman_puppet/rubygem-hammer_cli_foreman_puppet.spec
```diff
@@ -0,0 +1,95 @@
...
+%config %{hammer_confdir}/cli.modules.d/%{plugin_name}.yml
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/README.md
+%doc %{gem_instdir}/config
+%{gem_instdir}/test
+
+%changelog
+* Mon Aug 02 2021 vagrant 0.0.2-1
```
ekohl, line 93: Please fix this

ekohl (changes requested):
You also need to add it to EL8. I think you are on an old version of foreman-packaging.git because the latest version of `add_gem.sh` should do that for you.

packages/foreman/rubygem-hammer_cli_foreman_puppet/rubygem-hammer_cli_foreman_puppet.spec
```diff
@@ -0,0 +1,95 @@
+# template: hammer_plugin
```
ekohl, line 1: This file should be removed

packages/foreman/rubygem-hammer_cli_foreman_puppet/hammer_cli_foreman_puppet-0.0.2.gem
```diff
@@ -0,0 +1 @@
+../../../.git/annex/objects/04/6J/SHA256E-s131072--f15b5d424ade345fdafe7f6393f56bde30e54bd56a45654f0bf04df6b3f33399.2.gem/SHA256E-s131072--f15b5d424ade345fdafe7f6393f56bde30e54bd56a45654f0bf04df6b3f33399.2.gem
```
ekohl, line 1: This file should be removed

rel-eng/tito.props
```diff
@@ -1105,3 +1105,13 @@ whitelist = jsoncpp
...
+
+
+[foreman-plugin-nightly-rhel7]
+whitelist = 
+        rubygem-hammer_cli_foreman_puppet
+
+
+[foreman-plugin-nightly-el8]
+whitelist = 
+        rubygem-hammer_cli_foreman_puppet
```
ekohl, line 1117: This isn't right

ekohl (changes requested):
Looks like you added it to `foreman` instead of `plugins`.

rel-eng/tito.props
amirfefer: the `add_gem` script is broken, https://github.com/theforeman/foreman-packaging/pull/6972 should fix it

tbrisker:
[test rpm] for some reason repoclosure fails on rubygem(hammer_cli_foreman) >= 3.0.0.pre.develop even though the dependency was updated

ekohl:
> for some reason repoclosure fails on rubygem(hammer_cli_foreman) >= 3.0.0.pre.develop even though the dependency was updated

Did it fail on EL8? Because then it might be due to the automatic dependency generator which translates the dependencies in the gemspec to RPM ones.

tbrisker:
> > for some reason repoclosure fails on rubygem(hammer_cli_foreman) >= 3.0.0.pre.develop even though the dependency was updated
> 
> Did it fail on EL8? Because then it might be due to the automatic dependency generator which translates the dependencies in the gemspec to RPM ones.

It did indeed. Does EL8 ignore what is in the rpm spec and rely on the gemspec instead?

packages/plugins/rubygem-hammer_cli_foreman_puppet/rubygem-hammer_cli_foreman_puppet.spec
```diff
@@ -0,0 +1,95 @@
...
+
+%prep
+%{?scl:scl enable %{scl} - << \EOF}
+gem unpack %{SOURCE0}
+%{?scl:EOF}
+
+%setup -q -D -T -n  %{gem_name}-%{version}
+
+%{?scl:scl enable %{scl} - << \EOF}
+gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
```
ekohl, line 54: You can probably patch the gemspec like this:
```suggestion
gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
sed -i 's/.pre.develop//' %{gem_name}.gemspec
```
```

**Review by ekohl — COMMENTED** [review]
This is indeed the autogenerated one.
```
[2021-08-03T10:55:35.682Z]     package: rubygem-hammer_cli_foreman_puppet-0.0.2-1.fm3_0.el8.noarch from downloaded_rpms
[2021-08-03T10:55:35.682Z]       unresolved deps:
[2021-08-03T10:55:35.682Z]         (rubygem(hammer_cli_foreman) >= 3.0.0.pre.develop with rubygem(hammer_cli_foreman) < 4.0.0)
```

ekohl:
> It did indeed. Does EL8 ignore what is in the rpm spec and rely on the gemspec instead?

No, there it's a case of both for EL8. This means one can be stricter than the other (as we see here). Once we drop EL7 we can probably move completely to autogenerators and save a lot of effort in updating spec files in packaging.

ekohl (approved):

PR #6974, opened 2021-08-03
Update foreman_remote_execution to 4.7.0

adamruzicka:


ekohl (commented):
Looks like the dependency on tasks (and possibly foreman itself) are out of date. Would you mind updating those as well?

pcreech:
ACK

PR #6975, opened 2021-08-03
Add hammer-cli-foreman-puppet

ofedoren:
@amirfefer

Supported Versions:

 * Nightly


dependencies/bionic/hammer_cli_foreman_puppet/control
```diff
@@ -0,0 +1,26 @@
+Source: ruby-hammer-cli-foreman-puppet
+Section: ruby
+Priority: optional
+Maintainer: Debian Ruby Extras Maintainers <pkg-ruby-extras-maintainers@lists.alioth.debian.org>
+Uploaders: <>
+Build-Depends: debhelper (>= 11~),
+               gem2deb,
+               rake,
+               ruby-hammer-cli-foreman (<< 4.0.0),
+               ruby-hammer-cli-foreman (>= 3.0.0.pre.develop)
```
ekohl, line 10: I'm not sure if this is valid Debian versioning, but also a bit unsure what to recommend. Right now it's not installable anyway since there is no nightly Hammer package on Debian.

dependencies/bionic/hammer_cli_foreman_puppet/changelog
```diff
@@ -0,0 +1,5 @@
+ruby-hammer-cli-foreman-puppet (0.0.2-1) UNRELEASED; urgency=medium
```
ekohl, line 1: I'm wondering about the UNRELEASED part. Is that needed?

ofedoren: I have no clue, just used the script :(

ofedoren: It seems like it should be `stable` as in other plugins...

dependencies/bionic/hammer_cli_foreman_puppet/control
ekohl:
```
15:35:51  pbuilder-satisfydepends-dummy : Depends: ruby-hammer-cli-foreman (>= 3.0.0.pre.develop) but 2.1.1-1 is to be installed
```

That's very weird because nightly should have a much newer package than 2.1.

tbrisker:
replaced by #6988

PR #6980, opened 2021-08-04
Release hammer_cli and hammer_cli_foreman 3.0.0

ofedoren:
@ekohl

Supported Versions:

 * Nightly


packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
```diff
@@ -5,8 +5,6 @@
 %global gem_name hammer_cli
 
 %global release 1
-%global prereleasesource pre.develop
```
ekohl, line 8: You can comment this

packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
```diff
@@ -5,8 +5,6 @@
...
-%global prerelease %{?prereleasesource:.}%{?prereleasesource}
```
ekohl, line 9: Please keep this

ekohl (changes requested):
Same comments for the other spec.

ofedoren:
@ekohl, updated. Did I understand you right?..

ekohl (commented):
Yes, you understood me right, but I think this is actually an edge case. For nightly packages we don't release the GA build to nightly. What I think is right is if we get to the point where we branch, then update `rpm/develop` to `3.1.0-develop` (including hammer) and then update to 3.0.0 GA in `rpm/3.0`. Does that make sense?

upadhyeammit:
> Yes, you understood me right, but I think this is actually an edge case. For nightly packages we don't release the GA build to nightly. What I think is right is if we get to the point where we branch, then update `rpm/develop` to `3.1.0-develop` (including hammer) and then update to 3.0.0 GA in `rpm/3.0`. Does that make sense?

Agree, can't see better way to handle this.

tbrisker:
IIUC we agreed to stick with nightly hammer builds in rpm/develop, closing. If I misunderstood feel free to reopen.

PR #6988, opened 2021-08-05
Add hammer-cli-foreman-puppet

ekohl:


tbrisker:
Looks like it failed with: 15:36:20 File does not exist: simplecov

ekohl:
Updated so it doesn't run the testsuite.

tbrisker:
hmm, looks like now it isn't finding 3.0.0 for some reason: 
```
16:58:02 /usr/lib/ruby/2.7.0/rubygems/dependency.rb:311:in `to_specs': Could not find 'hammer_cli_foreman' (< 4.0.0, > 2.6.0) among 56 total gem(s) (Gem::MissingSpecError)

ekohl:
Oh, we probably didn't promote it yet? It may be stuck in staging.

tbrisker:
@ekohl could it be because you dropped it from the build deps? maybe it needs to be present in both?

pcreech:
ACK

PR #7020, opened 2021-08-20
another try to speedup debian builds

evgeni:


debian/bionic/foreman/rules
```diff
@@ -5,6 +5,8 @@
 # DEB_VERBOSE_ALL=1
 # DH_VERBOSE=1
 
+DEB_BUILD_OPTIONS=noautodbgsym
```
evgeni, line 8: while this doesn't speed up the build, the dbgsym package is not generated and thus doesn't take up space on the mirror

debian/bionic/foreman/filter-package-json.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+
+import json
+
+# We need a list of packages that are devDependencies but not needed to build
```
evgeni, line 5: if you think you've seen this before, it's true, it's stolen from https://github.com/theforeman/foreman-packaging/blob/8d1663bca7d647e652ca0803772bb21586bff2e4/update-requirements#L9-L53

I wonder if this should move into `foreman.git`, at least the definitions of what to exclude.

ekohl: That actually feels like a good idea. Then it can be done in the same commit as when a dependency is added. It'll also work with reverts etc. :+1:

debian/bionic/foreman/filter-package-json.py
```diff
@@ -0,0 +1,63 @@
...
+def filter_section(packages, section):
+    deps = packages.get(section, {})
+    for package, version in sorted(deps.items()):
+        if not is_excluded(package):
+            yield (package, version)
+
+with open('package.json') as package_json:
+    data = json.load(package_json)
+for section in ('devDependencies', 'dependencies'):
+    data[section] = {package: version for (package, version) in filter_section(data, section)}
```
ekohl, line 60: Can be
```suggestion
    data[section] = dict(filter_section(data, section))
```

debian/bionic/foreman/filter-package-json.py
evgeni: The drawback is that today this list is applied to plugins too. If it lives in `foreman.git`, this wouldn't happen anymore.

evgeni:
[test packaging]

ehelms:
29 minute build duration aint half bad

debian/bionic/foreman/filter-package-json.py
ekohl: Can we ship it in `foreman-assets` then?

evgeni: Oh, I meant for `update-requirements` to use it when we package up plugins in `rpm/develop`.

ekohl: So there's 2 approaches that I can see:
* Ship a JSON file in foreman.git with the 2 lists
* Ship the entire script in foreman.git

Which one are you talking about?

evgeni: To roundup what we talked on IRC:
* put a JSON file into `foreman.git`
* rpm packaging `update-requirements` would gain a way to get that file via HTTPS from GitHub or alternatively a local checkout
* I'd probably put this script (`filter-package-json.py`) also into `foreman.git` as it allows deduplication vs having it here three times :woman_facepalming:

evgeni:
all the learnings from this PR have been merged into various places:
* https://github.com/theforeman/foreman/pull/8729
* https://github.com/theforeman/foreman/pull/8730
* https://github.com/theforeman/foreman/pull/8731
* https://github.com/theforeman/foreman-packaging/pull/7024

the only remaining parts are:
* https://github.com/theforeman/foreman-packaging/pull/7022
* the `DEB_BUILD_OPTIONS=noautodbgsym` addition

I'll wait for #7022 to be merged, and then open a dedicated PR for the `noautodbgsym` change (which is mostly cosmetical)

PR #7022, opened 2021-08-20
drop node dependencies which are unused for building

evgeni:


ehelms:
[test packaging]

ekohl:
Depends on https://github.com/theforeman/foreman/pull/8731

evgeni:
all my changes to core have been merged, so this is now just a simple "call a script and carry on" PR :joy_cat:

evgeni:
one thing this also should help with, is plugin building.
when you build a plugin, it needs to install `foreman-assets`, which takes ~2min to unpack all the content -- less content in there now = faster to unpack \o/

debian/bionic/foreman/rules
```diff
@@ -16,6 +16,7 @@ build:
 	/bin/rm -f bundler.d/test.rb bundler.d/development.rb
 	/bin/sed -i '2i  gem "puma-status"' bundler.d/service.rb
 	BUNDLE_BUILD__SASSC=--disable-march-tune-native BUNDLE_RETRY=5 BUNDLE_JOBS=4 /usr/bin/bundle package
+	/usr/bin/python script/filter-package-json.py
```
ekohl, line 19: It's not executable with a shebang?

evgeni: It's not (and the shebang is using py3, which we don't install by default -- the script works with py2 and 3 identically)

ekohl (approved):

PR #7023, opened 2021-08-23
don't run scripts during "npm install"

evgeni:


ekohl:
With https://github.com/theforeman/foreman/pull/8730, is this still needed?

evgeni:
my *guess* is: no

evgeni:
[test packaging]

I still want to see if that works

PR #7024, opened 2021-08-23
load excludes from foreman.git

evgeni:

### DIFF
--- a/update-requirements
+++ b/update-requirements
@@ -4,53 +4,15 @@
 import argparse
 import json
 
+import requests
 import semver
 
-# We need a list of packages that are devDependencies but not needed to build
-EXCLUDE_NPM_PACKAGES = [
-    '@sheerun/mutationobserver-shim',
-    '@testing-library/jest-dom',
-    '@testing-library/react',
-    '@theforeman/env',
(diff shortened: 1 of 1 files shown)


update-requirements
```diff
@@ -146,11 +108,18 @@ def main():
...
                         type=argparse.FileType('r'))
     parser.add_argument('specfile', help='path to the specfile', type=argparse.FileType('r'))
+    parser.add_argument('--excludefile', help='Path/URL to the package-exclude.json (default: %(default)s)', default=PACKAGE_EXCLUDE_JSON_URL)
     args = parser.parse_args()
 
+    if args.excludefile.startswith('https://'):
+        excludes = requests.get(args.excludefile).json()
+    else:
+        with open(args.excludefile) as excludesfile:
+            excludes = json.load(excludesfile)
```
ekohl, line 118: I think it should be moved into the `if args.requirementstype == 'npm':` section so we don't reach out over the network for non-NPM.

ekohl (changes requested):

update-requirements
evgeni: fair

ekohl (approved):

PR #7026, opened 2021-08-23
Use tracer upload cron job on SLES, too

sbernhard:


packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -379,6 +371,7 @@ exit 0
 %{plugins_dir}/tracer_upload.py*
 %{katello_libdir}/tracer.py*
 %{plugins_confdir}/tracer_upload.conf
+%config(noreplace) %attr(0644, root, root) %{_sysconfdir}/cron.d/%{name}
```
evgeni, line 374: Moving the cron file into the tracer sub-package feels wrong, as the cron file is called `katello-agent-send.cron`.

It does - today - only contain one call to `katello-tracer-upload`, and is totally unrelated to `katello-agent`.

My preferred way would be to rename that file to `katello-tracer-upload.cron` upstream (so that nobody accidentally adds non-tracer commands there), and *then* place it here.

@jturel, thoughts?

jturel: Agreed. I think that is accounted for in an open PR at this time (see the first change) => https://github.com/Katello/katello-host-tools/pull/119/files

evgeni: Ah, there is even work for that. Awesome!

sbernhard: Actually, the name in the PR is already set to "katello-tracer-upload.cron". I guess, we are fine then.

evgeni: Which is interesting, as that shouldn't work: above you do `cp extra/katello-agent-send.cron %{buildroot}%{_sysconfdir}/cron.d/%{name}` and `%{name}` should expand to `katello-host-tools`.

sbernhard: yes, sure. First we need to merge the PR in katello-host-tools. then we can fix this PR and set the correct name.

PR #7030, opened 2021-08-24
speedup installation of foreman by multi-threading bundle install

evgeni:

### DIFF
--- a/debian/bionic/foreman/foreman-assets.postinst
+++ b/debian/bionic/foreman/foreman-assets.postinst
@@ -21,9 +21,9 @@ BUNDLE=bundle
 export HOME=/usr/share/foreman
 cd /usr/share/foreman
 if [ ! -z "${DEBUG}" ]; then
-  $BUNDLE update --local
+  $BUNDLE update --jobs 4 --local
 else
-  $BUNDLE update --local 2>&1 >> $LOGFILE
+  $BUNDLE update --jobs 4 --local 2>&1 >> $LOGFILE
 fi
 
 # Own all the core files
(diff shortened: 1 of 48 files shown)


evgeni:
the (unscientific) tests on my laptop (in a buster container) say:
* current `foreman` nightly installs in 50-52 seconds
* `foreman` from this PR installs in 43-45 seconds

So not *a lot*, but I'd expect more if we add things like `postgresql` or `libvirt` which pulls in native extensions, will re-test

ekohl (approved):
I was reminded of https://projects.theforeman.org/issues/7988 by this.

evgeni:
> I was reminded of https://projects.theforeman.org/issues/7988 by this.

it's on my pile of "Evgeni is angry enough to look at our Debian packaging" things

evgeni:
* `apt install foreman foreman-postgresql foreman-libvirt` takes ~2:30 from nightly
* and ~2:20 from this PR :(

Meh, I hoped it's more. But I think I'd still try that and see what e.g. happens on our CI?

ekohl:
Does it actually compile those on a real system or are there "wheels" with precompiled cache files? If it does compile, it wouldn't surprise me if there's no concurrency within a build. https://nokogiri.org/tutorials/installing_nokogiri.html suggests there is such a thing.

That said, there's https://nokogiri.org/tutorials/installing_nokogiri.html#installing-using-standard-system-libraries which suggests `bundle config build.nokogiri --use-system-libraries`. Perhaps there's more like that?

evgeni:
It does compile on the target system. But my expectation was that it can compile e.g. nokogiri and libvirt in parallel

ekohl:
Probably, but I suspect it's still tied to the longest one and that's likely nokogiri.

ekohl (approved):
I don't see a good place to set env vars, but otherwise we could `export BUNDLE_JOBS=4` for the same effect. Though I don't know how well that would work with wrappers and it'd be hard to verify since it would be silently ignored. The difference would be speed but we don't have any test on that.

I also think 4 is probably safe. While we may have VMs with just 1 or 2 cores, we don't have that high concurrency that it'd overload machines. Maximum memory consumption is probably also not a problem.

PR #7032, opened 2021-08-24
drop node dependencies which are unused for building

evgeni:
(cherry picked from commit aca4fa59749f98984693e8ba976e001bce959672)

this will fail until rc2 is released, so it needs to be either part of the RC2 PR or merged now failing and hope it won't break RC2 (I am confident it wont)


evgeni:
cc @pcreech @zjhuntin whatever you prefer ;)

ekohl (approved):
I'm leaning to merging it now given the risk we'll forget.

pcreech:
Agreed on merging now, as i will definitely forget to do this when I start 3.0

PR #7046, opened 2021-08-27
Update rubygem-pulp_ansible_client to 0.9.1

jlsherrill:


jlsherrill:
@evgeni we should probably update to 0.9.1 for any fixes, but this update was made in katello to get us from the 0.8.z client to the 0.9.z range.  I didn't realize it leapfrogged what we had in the 3.14 repos.  The client side bits were seemingly request identical to the 0.8.z. client we had, so we don't have to update for this

evgeni (approved):

PR #7047, opened 2021-08-30
Add foreman_ygg_worker

adamruzicka:
I failed crafting a rpm spec which would actually compile the resulting binary as part of the build process. Because of this, I opted to build the binaries and attach them to a release when a tag is pushed to the source repository[1]. In here I just pull the build artifacts and put them into the right location.

Opening ahead of time since there are still some unknowns, such as how will yggdrasil itself be distributed.

[1] - https://github.com/ezr-ondrej/foreman_ygg_worker/pull/1


packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -0,0 +1,37 @@
...
+%global repo_name foreman_ygg_worker
+%global yggdrasil_libexecdir %{_libexecdir}/yggdrasil
+
+Name: foreman_ygg_worker
+Summary: Worker service for yggdrasil that can act as pull client for Foreman Remote Execution
+Version: 0.0.1
+Release: 1%{?dist}
+License: MIT
+
+Source0: https://github.com/%{repo_orgname}/%{repo_name}/releases/download/v%{version}/%{repo_name}-linux-amd64.tar.gz
```
ehelms, line 11: Does this contain all dependencies? I was comparing this spec file to what `go2pm` generates, and I noticed that it generates explicit dependencies as BuildRequires:

```
BuildRequires:  golang(git.sr.ht/~spc/go-log)
BuildRequires:  golang(github.com/google/uuid)
BuildRequires:  golang(github.com/redhatinsights/yggdrasil/protocol)
BuildRequires:  golang(google.golang.org/grpc)
```

Since it seems to do that build in the spec file:

```
%description
%{common_description}

%gopkg

%prep
%goprep

%build
for cmd in src; do
  %gobuild -o %{gobuilddir}/bin/$(basename $cmd) %{goipath}/$cmd
done

%install
%gopkginstall
install -m 0755 -vd                     %{buildroot}%{_bindir}
install -m 0755 -vp %{gobuilddir}/bin/* %{buildroot}%{_bindir}/
```

ehelms:
Fedora's guidelines seem to indicate bundling dependencies can be done if justified but it is preferred not to: https://docs.fedoraproject.org/en-US/packaging-guidelines/Golang/#_bundled_or_unbundled

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -0,0 +1,37 @@
...
+%description
+Worker service for yggdrasil that can act as pull client for Foreman Remote Execution.
+
+%prep
+%setup -n %{repo_name}-%{version}-linux-amd64.tar.gz -c
+
+%install
+# TODO: License?
+# TODO: Readme?
+install -D -m755 ./%{name}-%{version} %{buildroot}%{yggdrasil_libexecdir}/%{name}
```
ehelms, line 30: I realize this is a unique location so the use of `%gopkginstall` may not work.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
adamruzicka: TIL `go2rpm` exists. I tried putting together a spec by hand which didn't go exactly great. I'll give it another shot.

adamruzicka: The worker itself requires `git.sr.ht/~spc/go-log`, `github.com/google/uuid`, `github.com/redhatinsights/yggdrasil` and `google.golang.org/grpc`.

`git.sr.ht/~spc/go-log` and `github.com/google/uuid` have no dependencies, `google.golang.org/grpc` seems to be packaged (or at least in fedora's repos), however yggdrasil itself has quite a few dependencies (see https://github.com/RedHatInsights/yggdrasil/blob/main/go.mod). There might be overlap between yggdrassil's and grpc's dependencies, but it would still mean we would have to carry a lot of random build dependencies in our repos. Is that really what we want?

ehelms: Not if I can help it, however, I am trying to make sure we really think through it since that has bit us with nodejs sometimes. And, building binaries outside of the RPM can get weird sometimes.

Is it possible to bundle the dependencies but still build it inside the RPM environment? Similar question if providing a binary as the source makes things easier or harder for Debian. @evgeni thoughts?

evgeni: I always thought that in Go land, it's usual to vendor stuff inside your app. We do have one Go think already packaged: [pcp-mmvstatsd](https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/foreman/pcp-mmvstatsd) and it uses a `vendor/` folder with its deps, so we only fetch one tarball and then go `go build` (SCNR):

https://github.com/theforeman/foreman-packaging/blob/331d9e9058e8da9c5504d6f52aede35fc346242d/packages/foreman/pcp-mmvstatsd/pcp-mmvstatsd.spec#L30-L38

I have no idea about the Debian side of things.

adamruzicka:
Builds here will still be broken until https://github.com/ezr-ondrej/foreman_ygg_worker/pull/1 goes in

adamruzicka:
[test rpm]

adamruzicka:
[test rpm]

adamruzicka:
Rebuilt the distribution tarball [test rpm]

adamruzicka:
> Error: No Package found for golang

Is there a way to make epel available?

evgeni:
> Is there a way to make epel available?

for EL7 yes, and I thought we dropped el6 support for clients fir 3.0?

jturel:
[test rpm]

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -0,0 +1,62 @@
...
+Release: 1%{?dist}
+License: MIT
+
+Source0: https://github.com/%{repo_orgname}/%{repo_name}/releases/download/v%{version}/%{repo_name}-%{version}.tar.gz
+Url: https://github.com/%{repo_orgname}/%{repo_name}/
+
+ExclusiveArch: x86_64
+
+BuildRequires: tar
+BuildRequires: gzip
```
evgeni, line 32:
```suggestion
```

I think those two are not necessary.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -0,0 +1,62 @@
...
+%setup -q
+
+%build
+mkdir -p _gopath/src
+ln -fs $(pwd)/src _gopath/src/%{name}-%{version}
+ln -fs $(pwd)/vendor _gopath/src/%{name}-%{version}/vendor
+export GOPATH=$(pwd)/_gopath
+pushd _gopath/src/%{name}-%{version}
+%{gobuild}
+strip %{name}-%{version}
```
evgeni, line 49: Shouldn't `strip` be handled by the RPM tooling?

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -0,0 +1,62 @@
...
+%global goipath         github.com/%{repo_orgname}/%{repo_name}
+
+%if 0%{?rhel} > 7 && ! 0%{?fedora}
+%define gobuild(o:) \
+        go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -compressdwarf=false -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v %{?**};
+%else
+%if ! 0%{?gobuild:1}
+%define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '" -a -v %{?**};
+%endif
+%endif
```
evgeni, line 18: what is this for?

adamruzicka: IIRC el8 and newer ship a `gobuild` macro which can be used to build hardened go binaries. EL7 doesn't have this macro and this chunk is a reimplementation for it, borrowed from here http://pkgs.devel.redhat.com/cgit/rpms/podman/tree/podman.spec?h=stream-container-tools-rhel8

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -0,0 +1,62 @@
...
+
+Source0: https://github.com/%{repo_orgname}/%{repo_name}/releases/download/v%{version}/%{repo_name}-%{version}.tar.gz
+Url: https://github.com/%{repo_orgname}/%{repo_name}/
+
+ExclusiveArch: x86_64
+
+BuildRequires: tar
+BuildRequires: gzip
+BuildRequires: golang
+Requires: yggdrasil
```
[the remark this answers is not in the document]
adamruzicka: To be honest I have no idea. I have not written neither yggdrasil nor the worker. CC @ezr-ondrej

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
adamruzicka: Maybe it should, but I don't see it doing it.

adamruzicka:
Uhmm, what?

evgeni:
I think `go_arches` is not defined on our Koji (and that's why you had this static list in `yggdrasil.spec`).

adamruzicka:
On EL7 it should be provided by `golang` package itself. Or does it need to be defined before the build deps are pulled?

evgeni:
chicken egg, I guess. we regenerate the SRPM on Koji, and during that time, no `BuildRequires` are present, only what koji considers necessary for srpm creation, and that doesn't include golang.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
evgeni: fascinating (in a bad way, but you're just a messenger here)

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -0,0 +1,66 @@
...
+Url: https://github.com/%{repo_orgname}/%{repo_name}/
+
+%if ! 0%{?go_arches:1}
+%define go_arches %{ix86} x86_64 %{arm} aarch64 ppc64le
+%endif
+ExclusiveArch: %{go_arches}
+
+%if 0%{?rhel} >= 8
+BuildRequires: go-srpm-macros
+%endif
```
evgeni, line 36: I think you can drop this, the if/define in the lines above should handle everything for the places where we don't have it, and the correct macros will be in the brew buildroots.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -0,0 +1,63 @@
...
+Name: foreman_ygg_worker
+Version: 0.0.3
+Summary: Worker service for yggdrasil that can act as pull client for Foreman Remote Execution
+Release: 1%{?dist}
+License: MIT
+
+Source0: https://github.com/%{repo_orgname}/%{repo_name}/releases/download/v%{version}/%{repo_name}-%{version}.tar.gz
+Url: https://github.com/%{repo_orgname}/%{repo_name}/
+
+%if ! 0%{?go_arches:1}
```
ekohl, line 29: Just a slight note: I always like to mention why this is. In this particular case it's on EL7 compat.
```suggestion
# EL7 doesn't define go_arches
%if ! 0%{?go_arches:1}
```

ekohl (commented):
Not a real review, just a small note.

evgeni (approved):

evgeni:
```console
% koji add-pkg --owner kojiadm foreman-client-nightly-rhel7 foreman_ygg_worker
Adding 1 packages to tag foreman-client-nightly-rhel7
% koji add-pkg --owner kojiadm foreman-client-nightly-el8 foreman_ygg_worker
Adding 1 packages to tag foreman-client-nightly-el8
```

PR #7049, opened 2021-08-30
just a test, nothing to see here

evgeni:

### DIFF
--- a/plugins/ansible-collection-theforeman-foreman/debian/changelog
+++ b/plugins/ansible-collection-theforeman-foreman/debian/changelog
@@ -1,6 +1,6 @@
 ansible-collection-theforeman-foreman (2.2.0-1) stable; urgency=low
 
-  * 2.2.0 released
+  * 2.2.0 released - DON@T TOUCH
 
  -- Evgeni Golov <evgeni@debian.org>  Wed, 25 Aug 2021 09:33:54 +0200
 


ekohl:
image

evgeni:
[test packaging]

evgeni:
```
cp: cannot create regular file '/var/cache/pbuilder/bionic64/result/ansible-collection-theforeman-foreman_9999-plugin+scratchbuild+20210830145932_all.deb': Permission denied
```

Meh, that's what the second sudo did? Baaaaah

evgeni:
@ekohl  the cheap solution would be to do env_keep for everyone, not just Jenkins?

ekohl:
Just an idea (and much bigger) but would https://tauware.blogspot.com/2020/04/building-packages-with-buildah-in-debian.html be a good idea? Benefits would be that it's rootless (no more sudo) and no longer requires anything Debian-specific (no more need for a Debian node which also means we have a lot more capacity for Debian packaging). Obvious downside is that it's rewriting our entire Debian package build process.

evgeni:
Long term? Yeah. But I'd prefer to switch one thing at a time right now ;)

ekohl:
This looks like something we need: https://unix.stackexchange.com/questions/13240/etc-sudoers-specify-env-keep-for-one-command-only/13298#13298

evgeni:
You mean so that the env_keep is anchored on the command, not the user?

evgeni:
Or we can tell pdebuild to do the right thing: https://github.com/theforeman/foreman-infra/pull/1647

evgeni:
[test packaging]

evgeni:
cool, seems to work again \o/

PR #7052, opened 2021-08-31
install the bundle into a local path

evgeni:

### DIFF
--- a/debian/bionic/foreman/rules
+++ b/debian/bionic/foreman/rules
@@ -7,6 +7,9 @@
 
 DEB_BUILD_OPTIONS=noautodbgsym
 
+RUBY_VERSION := $(shell ruby -e 'print RbConfig::CONFIG["RUBY_API_VERSION"]')
+BUNDLE := $(firstword $(wildcard /usr/bin/bundle$(RUBY_VERSION) /usr/bin/bundle))
+
 build:
 	/bin/cp config/settings.yaml.example config/settings.yaml
 	/bin/cp config/database.yml.example config/database.yml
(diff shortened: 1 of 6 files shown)


ekohl (commented):
Does this mean we start shipping these gems in `vendor`? I'm not sure if `vendor` is really shipped as part of `foreman.deb` or if it's only a cache here.

evgeni:
[test packaging]

I shouldn't have had fiddled around with the builddir locally on the 

vendor/cache is shipped, and that's what I need, and where things end up in. it's just called out explicitly now

evgeni:
Huh, wonder why it complains about bundler on focal... I thought I wiped everything from the node. Will debug tomorrow.

evgeni:
[test packaging]

debian/bionic/foreman/rules
```diff
@@ -12,10 +12,11 @@ build:
 	/bin/cp config/database.yml.example config/database.yml
 	/bin/cp config/dynflow/orchestrator.yml.example config/dynflow/orchestrator.yml
 	/bin/cp config/dynflow/worker.yml.example config/dynflow/worker.yml
-	/usr/bin/gem install -f thor -v '1.1.0'
+	/bin/mkdir -p ./vendor/cache && cd ./vendor/cache && /usr/bin/gem fetch thor -v '1.1.0'
```
ekohl, line 15: Just wondering: is this something we should patch into a `Gemfile` or some `bundler.d` file instead?

evgeni: It's worse, I can explain on IRC tomorrow

evgeni:
wow, it built, amazing.

ekohl (commented):
For my understanding: why do you use the bundled bundler on Focal?

evgeni:
> For my understanding: why do you use the bundled bundler on Focal?

So the general problem is that Bundler uses Thor, but to allow bundles to use a different version of Thor than Bundler needs, they vendor Thor internally, so that when the bundle needs Thor, the one from Bundler is not seen and you get whatever Thor the bundle needs. (pip does similar things with its dependencies, for similar reasons.)
Now the Bundler package in Debian (before Bundler 2.2.z) used to unvendor Thor, and use the packaged version from `ruby-thor`, which then gets seen by Bundler and fucks up bundles that need a different version of Thor as the one packaged. (Gory details in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=955405)

Bundler 1.y (as we have in Buster and Bionic) is happy if we just place `thor-X.Y.Z.gem` in the right spot, regardless whether it installs the bundle system-wide or in a custom path.
Bundler 2.y (as we have in Focal) is happy for the system-wide case, but breaks in funny ways if a local bundle is used.

Luckily, starting with Focal, the *Ruby* package also has Bundler, so we don't *need* the separately packaged Bundler from the Bundler package.

So, TL;DR: we need either the bundled Bundler from Ruby 2.7+ *OR* the standalone Bundler 2.2+ (as that doesn't unvendor Thor anymore)

evgeni:
No, it's a problem now, because we want to use `--path` to install to a local bundle, instead to /usr/lib or wherever Bundler would install it to with `--system` (which is - god knows why - the default)

ekohl:
Oh yes, that's why I have this gem (pun intended) in my .bashrc:
```bash
# https://bugzilla.redhat.com/show_bug.cgi?id=1574594
export GEM_HOME="$(ruby -e 'print Gem.user_dir')"
```

ekohl (approved):

evgeni:
please don't merge just yet, I need to actually *test* the builds

debian/focal/foreman/rules
```diff
@@ -7,31 +7,32 @@
 
 DEB_BUILD_OPTIONS=noautodbgsym
 
+RUBY_VERSION := $(shell ruby -e 'print RbConfig::CONFIG["RUBY_API_VERSION"]')
+BUNDLE := $(firstword $(wildcard /usr/bin/bundle$(RUBY_VERSION) /usr/bin/bundle))
```
evgeni, line 11: I wonder if we should apply the same changes to the other `rules` files, even if it'll pick up `/usr/bin/bundle` there.

I'd probably also keep the (now useless) `gem fetch thor` here, just for the sake to keep all `rules` files identical. Makes edits so much easier.

What do you think?

ekohl: If it's safe and keeps our scripts the same, I can see the benefit for that.

evgeni: Yeah, there is no `/usr/bin/bundle2.5` on Buster and Bionic, so it will fall-through to `/usr/bin/bundle` and keep doing what it does today.

ekohl: Looking ahead at Bullseye, it looks like there is `/usr/bin/bundle2.7`. Would that behave like focal?

evgeni: Yes. Ideally, `cp -a focal bullseye` and done.

ekohl: Speaking of which, I think you looked into merging them. Right now it looks like they are all the same except for a changelog entry in `foreman-installer`. Are you still looking into merging them?
```console
$ diff -Nutr bionic/ buster/
diff -Nutr bionic/foreman-installer/changelog buster/foreman-installer/changelog
--- bionic/foreman-installer/changelog	2021-09-06 12:15:00.682999880 +0200
+++ buster/foreman-installer/changelog	2021-09-06 12:15:00.960003684 +0200
@@ -14,7 +14,7 @@
 
   * Require puppet 6.15.0+
 
- -- Eric D. Helms <ericdhelms@gmail.com>  Wed, 19 May 2021 11:58:40 -0400
+ -- Eric D. Helms <ericdhelms@gmail.com>  Wed, 19 May 2021 11:58:45 -0400
 
 foreman-installer (2.6.0-1) stable; urgency=low
 
$ diff -Nutr bionic/ focal/
diff -Nutr bionic/foreman-installer/changelog focal/foreman-installer/changelog
--- bionic/foreman-installer/changelog	2021-09-06 12:15:00.682999880 +0200
+++ focal/foreman-installer/changelog	2021-09-06 12:15:01.229007377 +0200
@@ -14,7 +14,7 @@
 
   * Require puppet 6.15.0+
 
- -- Eric D. Helms <ericdhelms@gmail.com>  Wed, 19 May 2021 11:58:40 -0400
+ -- Eric D. Helms <ericdhelms@gmail.com>  Wed, 19 May 2021 11:58:18 -0400
 
 foreman-installer (2.6.0-1) stable; urgency=low
 
```

evgeni: Yes, still my plan, even if the current PR actually makes the `control` file diverge (for now).

ekohl: To me that would be a good reason to copy the bundle changes into bionic & buster.

evgeni: I did copy over the `debian/rules` changes, for `debian/control`, that would break the builds, as `d/control` on Focal now drops the `bundler` dependency. On the other hand, we don't *need* to drop that dep... mhhh :thinking:

debian/bionic/foreman/rules
```diff
@@ -16,22 +19,22 @@ build:
 	cd ./vendor/cache && /usr/bin/gem fetch thor -v '1.1.0'
 	/bin/rm -f bundler.d/test.rb bundler.d/development.rb
 	/bin/sed -i '2i  gem "puma-status"' bundler.d/service.rb
-	BUNDLE_BUILD__SASSC=--disable-march-tune-native BUNDLE_RETRY=5 BUNDLE_JOBS=4 /usr/bin/bundle install
-	/usr/bin/bundle package --no-install
+	BUNDLE_BUILD__SASSC=--disable-march-tune-native BUNDLE_RETRY=5 BUNDLE_JOBS=4 $(BUNDLE) install --path ./install-bundle/
```
ekohl, line 22: I do think `--path` is deprecated. It's now preferred to use `bundle config` or the `BUNDLE_PATH` env var. https://bundler.io/man/bundle-config.1.html#REMEMBERING-OPTIONS says it will be changed in Bundler 3.

ekohl (approved):

evgeni:
well, OF COURSE this failed now, because we call bundle in NPM too. will fix tomorrow.

ekohl:
I asked this before, but can we set the `BUNDLE_*` env vars at the top?

evgeni:
> I asked this before, but can we set the `BUNDLE_*` env vars at the top?

this is precisely what I was going to try ;)

evgeni:
:green_apple:

ekohl (approved):

PR #7063, opened 2021-09-02
make bundle packaging more explicit

evgeni:
this applies the learnings from #7052 without trying to *install* the bundle without root (which freaks out focal for now)


evgeni:
> Can you help me understand what's the benefit to splitting bundle install and package actions?

Sure. I want to get rid of the need of running those builds as root. To do so, one would usually do `bundle install --path <something_user_can_write>`. And while you *can* pass `--path` (heh) to `bundle package`, it behaves slightly differently and also creates the `vendor/cache` *inside* this path, but for the sake of not touching too much, I'd prefer to keep it at `./vendor/cache`, not `./somepath/vendor/cache`.

And yes, we probably can export those `BUNDLE_` vars at the top, but at the same time there is probably no real benefit to any step but the `install` one?

PR #7064, opened 2021-09-02
don't ship JS and CSS maps in the package

evgeni:
they are *huge*

### DIFF
--- a/debian/bionic/foreman/rules
+++ b/debian/bionic/foreman/rules
@@ -25,6 +25,7 @@ build:
 	/usr/bin/bundle exec rake assets:precompile RAILS_ENV=production DATABASE_URL=nulldb://nohost
 	/usr/bin/bundle exec rake webpack:compile DATABASE_URL=nulldb://nohost
 	/usr/bin/bundle exec rake apipie:cache cache_part=resources --trace FOREMAN_APIPIE_LANGS=en_US RAILS_ENV=production DATABASE_URL=nulldb://nohost
+	/bin/rm -f public/webpack/*.js.map public/webpack/*.js.map.gz public/webpack/*.css.map public/webpack/*.css.map.gz
 	/bin/rm -f db/schema.rb
 	/bin/sed -ri '1sXenv rubyXenv foreman-rubyX' bin/*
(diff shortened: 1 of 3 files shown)


evgeni:
```console
% du -h *.map* | sort -h -r 
33M	foreman-vendor.bundle-v8.12.0-production-c73ae7d5c13c5a6ac6f0.js.map
7.0M	foreman-vendor.bundle-v8.12.0-production-c73ae7d5c13c5a6ac6f0.js.map.gz
4.7M	foreman-vendor.bundle-v8.12.0-production-c73ae7d5c13c5a6ac6f0.css.map
2.0M	bundle-b7d074c253583e9e0d43.js.map
444K	vendor-4674ef9c29ba15042016.js.map
440K	foreman-vendor.bundle-v8.12.0-production-c73ae7d5c13c5a6ac6f0.css.map.gz
380K	bundle-b7d074c253583e9e0d43.js.map.gz
104K	vendor-4674ef9c29ba15042016.js.map.gz
```

so ~48M (todays nightly DEB).

For RPM we drop these from plugins, but not from core (whyever).

Shipping just `.gz` should probably work with modern browsers, but I don't 100% know.

ekohl:
That does explain why foreman.deb became so HUGE.

> For RPM we drop these from plugins, but not from core (whyever).

I think because we couldn't make them work. It found wrong paths, pointing to `/BUILD` or something which isn't allowed by RPM. It's also weird for users because those files don't exist.

laviro (approved):
ACK, no need for that in packaging I believe

evgeni:
FWIW, the resulting .deb shrinks from 74M to 61M, which is roughly double the size of `foreman-vendor.bundle-v8.12.0-production-c73ae7d5c13c5a6ac6f0.js.map.gz`, which makes sense, as that's the biggest compressed file (so it can't be further compressed) and the non-compressed version gets compressed by `dpkg`.

tbrisker (approved):
fine by me, i don't think there are any developers who use debian for debugging production issues, and js code *should* be identical between deb and rpm (although it probably isn't...).

ekohl (approved):

PR #7067, opened 2021-09-02
Update katello to 4.2.0.1

pcreech:


pcreech:
@evgeni mind taking a look to make sure bump_rpm did what it needed to, in case i missed something?

evgeni (approved):

PR #7077, opened 2021-09-06
make deb compare packages more robust

evgeni:

### DIFF
--- a/scripts/compare_deb_repo.py
+++ b/scripts/compare_deb_repo.py
@@ -9,6 +9,7 @@ from debian.changelog import Changelog
 
 DISTS = [path.name for path in Path('debian/').glob('*') if path.name != 'buildscripts']
 PLUGINS = ['plugins']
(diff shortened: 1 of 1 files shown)


scripts/compare_deb_repo.py
```diff
@@ -23,15 +24,17 @@ def get_repo_packages(dist, release='nightly', arch='amd64'):
     return repo_packages
 
 
-def get_git_packages(dist):
+def get_git_packages(dist, release='nightly'):
     if dist == 'plugins':
-        package_folders = Path('plugins/').glob('*/debian')
+        package_changelogs = Path('plugins/').glob('**/changelog')
```
ekohl, line 29: Thinking out loud:  is it good to be more specific?
```suggestion
        package_changelogs = Path('plugins/').glob('*/debian/changelog')
```

ekohl (approved):

scripts/compare_deb_repo.py
evgeni: no, this was the point of the change, not all plugins have a debian subfolder :)

PR #7079, opened 2021-09-07
document how to remove packages from freight

evgeni:


ekohl (approved):
Should it mention that this happens on the webserver or is that obvious from `/var/www`? It is to me, but it may not be to everyone.

evgeni:
Technically, "on the machine hosting the repository", which happens to be web01.

I refrained from it, as it might get out of sync.

upadhyeammit (approved):
I am not sure if we should explicitly add that these steps are not required for stable repos, otherwise looks good!

evgeni:
> I am not sure if we should explicitly add that these steps are not required for stable repos, otherwise looks good!

They would work identically for stable repos, just doing such changes to a stable repo requires more reason than "cleanup".

PR #7086, opened 2021-09-10
Update rubygem-foreman_puppet to 1.0.0

ezr-ondrej:


ehelms (approved):

packages/plugins/rubygem-foreman_puppet/rubygem-foreman_puppet.spec
```diff
@@ -8,7 +8,7 @@
 %global foreman_min_version 2.4
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 1.0.0.rc.2
+Version: 1.0.0
```
evgeni, line 11: @ezr-ondrej @ehelms this version is smaller than 1.0.0.rc2 if you ask RPM...

ehelms: Oh cause of the whole we publish all the versions of plugins thing? Fair. @ezr-ondrej the easiest solution is release a 1.0.1 :D

evgeni: Yeah. And because we didn't catch that when rc1 want out.

ezr-ondrej: might that be an issue?

PR #7101, opened 2021-09-14
Update rubygem-foreman_virt_who_configure to 0.5.7

Odilhao:
(cherry picked from commit ef6043de316acc7984510747f91e6876446a0306)


evgeni (approved):
we fixed qpid in https://github.com/theforeman/foreman-packaging/pull/7090, but its not published yet, so that's why closure is failing

PR #7108, opened 2021-09-16
Update rubygem-foreman_puppet to 1.0.3

ezr-ondrej:
Latest bug fixes

### DIFF
--- a/packages/plugins/rubygem-foreman_puppet/rubygem-foreman_puppet.spec
+++ b/packages/plugins/rubygem-foreman_puppet/rubygem-foreman_puppet.spec
@@ -8,7 +8,7 @@
 %global foreman_min_version 2.4
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 1.0.1
+Version: 1.0.3
 Release: 1%{?foremandist}%{?dist}
 Summary: Adds puppet ENC features
 Group: Applications/Systems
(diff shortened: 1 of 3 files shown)


packages/plugins/rubygem-foreman_puppet/rubygem-foreman_puppet.spec
```diff
@@ -8,7 +8,7 @@
 %global foreman_min_version 2.4
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 1.0.1
+Version: 1.0.2
 Release: 1%{?foremandist}%{?dist}
 Summary: Adds puppet ENC features
```
ekohl, line 13: I think this isn't entirely accurate anymore since it does more.

ekohl (commented):
Looks good, but I think you're already working on a next release. Should we wait for that?

packages/plugins/rubygem-foreman_puppet/rubygem-foreman_puppet.spec
ezr-ondrej: Happy to change that in the gem, but now it is the gem's summary :shrug:

ezr-ondrej:
I've updated this to the 1.0.3, since it contains few more bugfixes

PR #7124, opened 2021-09-23
Update bootstrap-select to 1.13.18 for katello

jlsherrill:


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -127,7 +127,7 @@ BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 6.0.0
 # end package.json devDependencies BuildRequires
 # start package.json dependencies BuildRequires
 BuildRequires: %{?scl_prefix}npm(angular) = 1.8.2
-BuildRequires: %{?scl_prefix}npm(bootstrap-select) = 1.13.6
+BuildRequires: %{?scl_prefix}npm(bootstrap-select) = 1.13.18
```
evgeni, line 130: this will make this PR fail, sorry

jlsherrill: @evgeni should i update it here or as a 2nd pr?  I can never keep it straight :)

evgeni: Another PR, as it's a BR which needs to exist before Katello can be built

PR #7132, opened 2021-09-27
Stop calling gem compare in bump_rpm.sh

ekohl:
This command was very useful to detect new dependency updates, but most gems now have code where update-requirements can actually update it automatically. That means it's mostly useless. The file changes are generally not that interesting. In addition to that, it's also broken on Ruby 3 and has old outdated dependencies.


evgeni (approved):

ekohl:
And https://github.com/theforeman/forklift/pull/1416 removes it from the packaging box.

PR #7134, opened 2021-09-27
Bump bootstrap-select dep for katello

jlsherrill:


evgeni (commented):
Shouldn't this also bump the build requires of Katello itself?

evgeni:
Oh, this already happened in https://github.com/theforeman/foreman-packaging/pull/7124 ?!

evgeni (approved):

PR #7140, opened 2021-09-28
Fixes #33580 - Suggest new --puppet-server-foreman-url with hostname …

wbclark:
…change

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,7 +5,7 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 1
+%global release 2
 
 Name:       katello
 Version:    4.3.0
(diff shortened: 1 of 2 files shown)


wbclark:
I think I see why this is breaking:

https://github.com/theforeman/obal/blob/ead56f57e7e3da5763482e3ede3143b458974190/obal/data/modules/rpmspec_changelog_check.py#L45-L47

It expects a specifically formatted changelog entry for the release bump

packages/katello/katello/katello.spec
```diff
@@ -132,6 +132,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Tue Sep 28 2021 William Bradford Clark <wclark@redhat.com> 4.3.0-0.2master
```
ekohl, line 135:
```suggestion
* Tue Sep 28 2021 William Bradford Clark <wclark@redhat.com> 4.3.0-0.2.master
```

ekohl (changes requested):
:+1: for the change in general

wbclark:
> +1 for the change in general

Fixed the typo... Thank you!

ekohl (approved):

PR #7141, opened 2021-09-29
Update rubygem-foreman_puppet to 1.0.3

ezr-ondrej:
(cherry picked from commit ef3c0e3c69c6beb63177e249b78fea58fd1cf5e9)


ezr-ondrej:
Ping, can we get this in? it have few crucial bugfixes we should get in.

lzap (approved):
I am taking the action, 3.0 puppet is broken. The patch appears to be correct.

PR #7146, opened 2021-09-29
Add rubygem-msgpack package

jlsherrill:


evgeni:
We already have msgpack in plugins, it just needs moving to core packaging.

jlsherrill:
i've not moved a package before, but i think this would do it  (aside from the add-pkg and untagging from plugins)

pcreech (approved):

PR #7148, opened 2021-10-04
Update rubygem-rubyipmi to 0.11.0

ekohl:


packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
```diff
@@ -19,6 +19,7 @@ Requires: ipmitool
 Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}ruby
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
+Requires: %{?scl_prefix}rubygem(highline)
```
ekohl, line 22: This was added but it looks like this is only in Rakefile. I opened https://github.com/logicminds/rubyipmi/issues/54 for this.

ekohl:
I opened https://github.com/logicminds/rubyipmi/pull/55 for the hidden files. Will update the spec to exclude them for now.

lzap:
I merged the linked PR, but I guess 0.11 is out already so excluding is the only option.

Tests failed.

packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
lzap: Is there any concern in this regard? I mean, we have that dependency already so whats the deal?

packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
```diff
@@ -74,6 +75,11 @@ cp -a .%{gem_dir}/* \
 %{gem_libdir}
 %exclude %{gem_cache}
 %{gem_spec}
+%exclude .document
+%exclude .gitignore
+%exclude .rspec
+%exclude .rubocop.yml
+%exclude .travis.yml
```
evgeni, line 82:
```suggestion
%exclude %{gem_instdir}/.document
%exclude %{gem_instdir}/.gitignore
%exclude %{gem_instdir}/.rspec
%exclude %{gem_instdir}/.rubocop.yml
%exclude %{gem_instdir}/.travis.yml
```

packages/foreman/rubygem-rubyipmi/rubygem-rubyipmi.spec
ekohl: Not a huge amount, but fewer dependencies is generally speaking easier. Remember that the Smart Proxy pulls in this gem as a dependency and then in turn another. That means a container image would also become larger and even on my Smart Proxy only RPM installation I don't have it installed. Luckily highline itself has no (runtime) dependencies.

PR #7150, opened 2021-10-04
Fixes #33623 - Add cron rule for audits cleanup

domitea:
Adding cron rule for audits:expire rake task for removing old audits in automate way.


domitea:
Accidental closing...

ekohl (commented):
I'm not that happy with this because it's really painful to modify the number of days. Adding this to an upgrade also brings a great risk to delete audits that users wanted to maintain. Silently deleting data is a bad experience for users. They may even be liable, depending on some laws.

I think that with reports we have a setting for the amount of time to keep. Wouldn't that be a better way?

Also, this runs at midnight of the new year. That means at some point you will have 2 years of audits stored. Is that a good idea?

Thinking about other solutions: can we actually implement this as some recurring logic in foreman-tasks? While tasks is not installed everywhere right now, I'd like to get rid of all of these cronjobs and implement them as tasks.

Also, there needs to be a similar PR to deb/develop.

lzap:
I agree that amount of days to keep should be required option and by default if you just happen to execute it it should print an error.

But I do not think we need foreman-tasks for this, just introduce `/etc/sysconfig/foreman` or something like that (shell format) so users/installer can redefine. Eval the file from here and that should be sufficient.

domitea:
Thanks for the comments, but I'm little confused. 

I understand that silently deleting data is not great idea but on the flip side having DB full of audits is also not a good option. I see that have some way to configure amount of persisted days is a good idea. Ideal place for this setting would be a Settings because it's already known and it not introducing another config file somewhere at file system. 

This only the one piece, the second one is cron and the time of invocation. From my point of view, when I change the value a persisted days, I should also change the time of invocation. However @ekohl gives me a good point about two-year persistence when I remove old audits only once a year. I think that the time of invocation of this could be every day so that means I will only remove an old data for one day after the configured amount of days passed. I'm not sure if it's a good idea to do that this way. What do you think @lzap or @ekohl?

Also, if I'm not mistaken, the reports for expiring uses environmental variable (`ENV['days']`) that's also an option.
Also, after this PR will be just before merging, I'll do another PRs.

lzap:
Please follow what report expiration does (there is a default value hardcoded in the rake task). We can merge this to unblock you and then let's focus on improving this alltogether.

lzap:
I think it is hard to find a sweet spot, therefore I suggest that the deletion process would actually perform some data export to CSV or something via fast PostgreSQL bulk SQL operations prior deletion so customers could easily load the data back.

ekohl:
Generally :+1: on that approach, with one note:
> discuss the default configuration with users prior the branching, as that's something trivial to change even a day before the branching, we can default to no cleanup to keep things as they are for now

We can also use a default value that reflects no cleanups happen. That's the least invasive patch.

ares:
I tried to say that by "we can default to no cleanup to keep things as they are for now". If @lzap is also ok with that, let's proceed that way then.

domitea:
Checks are green so I think it's ready to merge. I will also prepare another PR for the Debian like OSes.

PR #7170, opened 2021-10-18
Update rubygem-foreman_discovery to 18.0.4

lzap:

* Nightly
* 3.0

Another small release in regard to Puppet.


itszakko:
Hi, when can we expect this fix in the 3.0 repository. We're being hit by the bug this fixes.

I see it in 3.1:
https://yum.theforeman.org/plugins/3.1/el8/x86_64/rubygem-foreman_discovery-18.0.4-1.fm3_1.el8.noarch.rpm

but nothing in 3.0:
https://yum.theforeman.org/plugins/3.0/el8/x86_64/

evgeni:
it now got updated to 18.0.5

PR #7172, opened 2021-10-18
Add yggdrasil package

adamruzicka:


adamruzicka:
Not sure what is going on here. Do we need to build it multiple times for different targets even if the result is an rpm containing a static binary able to run on anything of the same arch?

adamruzicka:
[test rpm]

adamruzicka:
On EL7 there's make-3.82-24, on EL8 there's make-4.2.1. Yggdrasil's Makefile does some nonstandard things, it all boils down to it setting `.SHELLFLAGS := -eu -o pipefail -c`. On EL7 make passes the whole variable as a single argument, while EL8's make splits it on whitespace.

ezr-ondrej (approved):
I needed to manually install `subscription-manager` (which is fixed in latest commit). Apart of it works as expected on Centos7 :+1:

adamruzicka:
@evgeni Could you take a look? I'd like to squash the commits before merging, do you want me to do that before final review or after?

packages/client/yggdrasil/yggdrasil.spec
```diff
@@ -0,0 +1,97 @@
+%define debug_package %{nil}
+
+%global go_arches x86_64 s390x ppc64le
+
+Name:    yggdrasil
+Version: 0.2.0
+Release: 3%{?dist}
+Epoch:   1
```
evgeni, line 8: this is so that our package always wins against which other package?

evgeni:
you'll have to add this to https://github.com/theforeman/foreman-packaging/blob/rpm/develop/comps/comps-foreman-client-rhel7.xml and https://github.com/theforeman/foreman-packaging/blob/rpm/develop/comps/comps-foreman-client-el8.xml, so the RPMs are actually shipped, but otherwise it looks good to me.

packages/client/yggdrasil/yggdrasil.spec
evgeni: I don't see an `yggdrasil` package in CentOS or EPEL, so not sure why this is set.

adamruzicka: https://github.com/theforeman/foreman-packaging/pull/7172#issuecomment-950464199

evgeni: No, this explains why you had to add the epoch to the changelog entry, but not why the epoch is used to begin with.

adamruzicka: Ah, I misread the question. No idea to be honest, it was already there when I pulled the rpm spec from rhc

evgeni: Hmm. So I *think* this is because there existed a `rhc` in the past that was totally unrelated to this new package, and downstream wanted to start versioning "fresh".

This doesn't apply to the `yggdrasil` name we're using here.

So I'd be in favor of dropping the Epoch (I *hate* Epochs), but also wouldn't block on this.

@ekohl @zjhuntin @Odilhao -- what do y'all think?

evgeni:
The Epoch question is the last open one, otherwise this can go in!

packages/client/yggdrasil/yggdrasil.spec
ekohl: If there's no need for an epoch, I'd avoid it. Epoch should only be used as a workaround.

adamruzicka: Dropped it

packages/client/yggdrasil/yggdrasil.spec
```diff
@@ -0,0 +1,96 @@
+%define debug_package %{nil}
+
+%global go_arches x86_64 s390x ppc64le
+
+Name:    yggdrasil
+Version: 0.2.0
+Release: 3%{?dist}
```
ekohl, line 7: Why do we start on release 3 for a new package?

packages/client/yggdrasil/yggdrasil.spec
```diff
@@ -0,0 +1,96 @@
+%define debug_package %{nil}
+
+%global go_arches x86_64 s390x ppc64le
```
ekohl, line 3: This is already defined, at least on EL7 (if golang is installed):
```console
# rpm --showrc |grep go_arc
-14: go_arches	%{ix86} x86_64 %{arm} aarch64 ppc64le
# grep go_arches /usr/lib/rpm/macros.d/macros.*
/usr/lib/rpm/macros.d/macros.golang:# ExclusiveArch:  %{go_arches}
/usr/lib/rpm/macros.d/macros.golang:%go_arches       %{ix86} x86_64 %{arm} aarch64 ppc64le
```

On EL you need to depend on `go-srpm-macros` for it. It's odd that https://github.com/RedHatInsights/yggdrasil/blob/main/yggdrasil.spec.in doesn't list this though.

Is this something we'd like to address?

evgeni: I'd prefer to stay as close to upstream spec as possible.

If it builds, it builds.

ekohl: Upstream does not declare it, which also means it can't build on EL8 as I understand it.

adamruzicka: I'd go with if it builds, it builds. I'd like to avoid the spec from getting more complex by pulling different deps on different platforms.

adamruzicka:
@ekohl could we get this moving in any direction?

ekohl (approved):
I'm OK with this. Shame we can't really rely on all the Go building infrastructure improvements in EL8 and beyond but here we are, stuck with EL7.

ekohl:
Note there are merge conflicts so you'll have to rebase to resolve them.

adamruzicka:
> Note there are merge conflicts so you'll have to rebase to resolve them.

I didn't see any, but rebased anyway

evgeni:
```console
% koji add-pkg --owner kojiadm foreman-client-nightly-el8 yggdrasil              
Adding 1 packages to tag foreman-client-nightly-el8
% koji add-pkg --owner kojiadm foreman-client-nightly-rhel7 yggdrasil
Adding 1 packages to tag foreman-client-nightly-rhel7
```

evgeni (approved):

PR #7175, opened 2021-10-19
Update rubygem-foreman_remote_execution to 4.8.0

adamruzicka:
(cherry picked from commit 458f51fa1569179660eb80065a65790b3faf3bd8)


packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -47,7 +47,7 @@ BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
 BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
 BuildRequires: %{?scl_prefix}npm(@patternfly/react-catalog-view-extension) >= 4.8.126
 BuildRequires: %{?scl_prefix}npm(@patternfly/react-catalog-view-extension) < 5.0.0
-BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 4.14.0
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 8.10.0
```
tbrisker, line 50: Is this change actually needed? afaict the builder hasn't changed since 8.7.2, and rpm/3.0 already has version 8.8.0 - so this should build just fine with the existing version (although that might require tweaking the package.json to get it building)

evgeni: no, nothing evaluates `package.json` at build time ;)

evgeni:
@adamruzicka is this still something we want to get into 3.0 and if so, shall we try with a lower `@theforeman/builder` requirement?

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
evgeni, line 50:
```suggestion
BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 8.7.2
```

evgeni:
> If possible, yes

pushed a change to use older builder, let's see

evgeni:
it built. :shipit: ?

evgeni (approved):

adamruzicka:
The rpm from here seem to work fine on EL7. :ship:

PR #7181, opened 2021-10-20
Fixes #33743: Mention Puppet options separately in change hostname script

ehelms:


ekohl:
Can we have a Redmine issue for this to track the work?

PR #7183, opened 2021-10-20
Remove unused smart_proxy_dynflow_bundlerd_dir macro

ekohl:
smart_proxy_dynflow_core is no more. Everything now runs in the smart-proxy process itself.

There are still 3 cases where it's used:
* rubygem-smart-proxy-probing
* rubygem-smart_proxy_acd
* rubygem-smart_proxy_salt

These need an actual update and is non-trivial.

For now I haven't added changelogs and haven't bumped the releases because it shouldn't make a difference in the final builds. I'm not sure what's wise here.


ekohl:
> rubygem-smart-proxy-probing

https://github.com/theforeman/foreman-packaging/pull/7190 cleans that up.

evgeni:
I'm good with the change.

But please add changelogs/revbumps, otherwise merging this will result in an endless fail-to-build on jenkins.

ehelms:
Fixed by https://github.com/theforeman/foreman-packaging/pull/7876

PR #7206, opened 2021-10-27
Bump nodejs-theforeman-{builder,vendor} to 8.15.0

pcreech:


tbrisker:
Probably a bad idea to pick this, the chance of all other plugins to break is high (as they were built against an older version). Is this really required for the rex bump? looking at the spec change in #7175 it looks like they only bumped the builder version, but there weren't any actual changes to the builder afaict since 8.7.2, so rex should build just fine with what is already in rpm/3.0.

evgeni:
closing, given the recent feedback

PR #7210, opened 2021-10-28
Add katello-pull-transport-migrate rpm

jeremylenz:
~~`katello-pull-provider-migrate`~~ `katello-pull-transport-migrate` is a new RPM whose purpose is solely to migrate a client currently using `katello-agent` to the new pull transport. This RPM will be installed via katello-agent and has these behaviors:

* The RPM will rely on the existing configuration in rhsm.conf to know which Satellite or Capsule it should communicate with
* ~~The RPM install will fail if client is not registered to a Satellite~~
* The RPM will have a hard dependency on the new pull provider client RPM
* The RPM will understand the configuration options needed to configure the new pull provider client
* The RPM will perform the configuration and start the pull provider agent

The benefits here are:

* no changes to Katello server (new APIs)
* no changes to katello-agent / katello-host-tools
* ~~If any of the actions executed by the RPM install script fail, the katello-agent action (dynflow task) used to install the package will fail providing actionable feedback to the user~~
* RPM not specific to a particular installation like katello-ca-consumer is. It could live in the satellite-tools repo and be synced just as katello-agent would be

### DIFF
--- a/comps/comps-foreman-client-el8.xml
+++ b/comps/comps-foreman-client-el8.xml
@@ -15,6 +15,7 @@
       <packagereq type="default">katello-agent</packagereq>
       <packagereq type="default">katello-host-tools</packagereq>
       <packagereq type="default">katello-host-tools-tracer</packagereq>
+      <packagereq type="default">katello-pull-provider-migrate</packagereq>
       <packagereq type="default">python3-gofer</packagereq>
       <packagereq type="default">python3-gofer-proton</packagereq>
       <packagereq type="default">python3-qpid-proton</packagereq>
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -777,7 +777,6 @@ foreman_client_packages:
     gofer: {}
     katello-host-tools: {}
     katello-pull-provider-migrate: {}
-    python-beautifulsoup4: {}
     python-psutil: {}
     rubygem-foreman_scap_client: {}
     tracer: {}
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -776,6 +776,8 @@ foreman_client_packages:
   hosts:
     gofer: {}
     katello-host-tools: {}
+    katello-pull-provider-migrate: {}
+    python-beautifulsoup4: {}
     python-psutil: {}
     rubygem-foreman_scap_client: {}
     tracer: {}
--- a/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate
+++ b/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate
@@ -1,14 +1,19 @@
 #!/bin/sh
 
+# set SBINDIR to /sbin if it is not set
+if [ -z "$SBINDIR" ]; then
+    SBINDIR=/usr/sbin
+fi
+
 # check if system is registered
 subscription-manager identity > /dev/null
 if [ $? -ne 0 ]; then
-    echo "Please register with subscription-manager first, then re-run `katello-pull-provider-migrate.sh`."
+    echo "Please register with subscription-manager first, then re-run `$SBINDIR/katello-pull-provider-migrate`."
     exit 1
 fi
 
 echo "Getting configuration from subscription-manager..."
-KPPTEMPFILE=/tmp/kpp_tempfile
+KPPTEMPFILE=$(mktemp kpp_tempfile_XXXXXXXX)
 subscription-manager config --list > $KPPTEMPFILE
 CONSUMER_CERT_DIR=$(grep 'consumercertdir' $KPPTEMPFILE | cut -d= -f2 | xargs | sed 's/[][]//g')
 CERT_FILE=$CONSUMER_CERT_DIR/cert.pem
@@ -66,7 +71,7 @@ systemctl start yggdrasild
 # and broker must be running on the server
 yggdrasil status
 
-if [ $(systemctl is-active yggdrasild) != "active" ]; then
+if ! systemctl is-active --quiet yggdrasild; then
     echo ""
     echo "yggdrasild failed to start! Check configuration in $CONFIGTOML and make sure the broker is running on the server."
     exit $?
--- a/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
+++ b/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
@@ -57,6 +57,7 @@ fi
 
 # start the yggdrasild service
 echo "Starting yggdrasild..."
+systemctl enable yggdrasild
 systemctl start yggdrasild
 
 # check status of yggdrasild and fail if it is not running
--- a/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
+++ b/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
@@ -1,0 +1,74 @@
+#!/bin/sh
+
+# check if system is registered
+subscription-manager identity > /dev/null
+if [ $? -ne 0 ]; then
+    echo "Please register with subscription-manager first, then re-run `katello-pull-provider-migrate.sh`."
+    exit 1
+fi
+
+echo "Getting configuration from subscription-manager..."
+KPPTEMPFILE=/tmp/kpp_tempfile
+subscription-manager config --list > $KPPTEMPFILE
+CONSUMER_CERT_DIR=$(grep 'consumercertdir' $KPPTEMPFILE | cut -d= -f2 | xargs | sed 's/[][]//g')
+CERT_FILE=$CONSUMER_CERT_DIR/cert.pem
+KEY_FILE=$CONSUMER_CERT_DIR/key.pem
+CA_FILE=$(grep 'repo_ca_cert = ' $KPPTEMPFILE | cut -d= -f2 | xargs | cut -d ' ' -f1)
+SERVER_NAME=$(grep 'hostname = ' $KPPTEMPFILE | cut -d= -f2 | xargs | cut -d ' ' -f1)
+
+# fail if no server name, cert dir, or ca file
+if [ -z "$SERVER_NAME" ] || [ -z "$CONSUMER_CERT_DIR" ] || [ -z "$CA_FILE" ]; then
+    echo "Unable to determine config from `subscription-manager config --list`; exiting"
+    exit 1
+fi
+
+# fail if client is not registered to a Katello
+if ! grep -q 'prefix = \/rhsm' $KPPTEMPFILE; then
+    echo "Client is not registered to Katello; exiting"
+    exit 2
+fi
+
+rm -f $KPPTEMPFILE
+
+# set SYSCONFDIR to /etc if it is not set
+if [ -z "$SYSCONFDIR" ]; then
+    SYSCONFDIR=/etc
+fi
+
+# see if /etc/yggdrasil/config.toml exists
+CONFIGTOML=$SYSCONFDIR/yggdrasil/config.toml
+if [ -f $CONFIGTOML ]; then
+    # make a backup of CONFIGTOML
+    cp $CONFIGTOML $CONFIGTOML.bak
+    cat <<EOF > $CONFIGTOML
+# yggdrasil global configuration settings written by katello-pull-provider-migrate
+
+broker = ["tcp://$SERVER_NAME:1883"]
+cert-file = "$CERT_FILE"
+key-file = "$KEY_FILE"
+ca-root = ["$CA_FILE"]
+log-level = "error"
+EOF
+
+else
+    echo "$SYSCONFDIR/yggdrasil/config.toml not found! Did `yum install yggdrasil` succeed?"
+    exit 1
+fi
+
+# start the yggdrasild service
+echo "Starting yggdrasild..."
+systemctl start yggdrasild
+
+# check status of yggdrasild and fail if it is not running
+# possible failure reason: incorrect protocol (should be tcp:// or mqtt://) or port (should be 1883)
+# also, cert-file and key-file must be valid
+# and broker must be running on the server
+yggdrasil status
+
+if [ $(systemctl is-active yggdrasild) != "active" ]; then
+    echo ""
+    echo "yggdrasild failed to start! Check configuration in $CONFIGTOML and make sure the broker is running on the server."
+    exit $?
+fi
+
+echo "Complete!"
--- a/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
+++ b/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
@@ -5,8 +5,8 @@ Summary:        An RPM that migrates katello-agent users to the new pull provide
 BuildArch:      noarch
 
 License:        LGPLv2
-URL:            https://github.com/theforeman/foreman-packaging/packages/client/katello-pull-provider-migrate
-Source0:        %{name}.sh
+URL:            https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/client/katello-pull-provider-migrate
+Source0:        %{name}
 
 Requires:       bash
 Requires:       yggdrasil
@@ -15,9 +15,9 @@ Requires:       foreman_ygg_worker
 %description
 An RPM that migrates katello-agent users to the new pull provider.
 
-Relies on the existing configuration in rhsm.conf to know which Katello server it should communicate with.
-The RPM install will fail if client is not registered to a Katello server.
-This RPM obtains the relevant values from the existing rhsm.conf, writes the configuration to yggdrasil's config.toml,
+Relies on the existing Red Hat Subscription Manager configuration to know which Katello server it should communicate with.
+The scriptlet will fail if the client is not registered to a Katello server.
+This RPM obtains the relevant values from the existing RHSM configuration, writes the configuration to yggdrasil's config.toml,
 and starts the pull provider agent.
 
 %prep
@@ -26,17 +26,17 @@ and starts the pull provider agent.
 
 %install
 rm -rf %{buildroot}
-mkdir -p  %{buildroot}%{_bindir}
-cp %{SOURCE0} %{buildroot}%{_bindir}
+mkdir -p  %{buildroot}%{_sbindir}
+cp %{SOURCE0} %{buildroot}%{_sbindir}
 
 %clean
 rm -rf %{buildroot}
 
 %post
-SYSCONFDIR=%{_sysconfdir} %{_bindir}/%{name}.sh
+SYSCONFDIR=%{_sysconfdir} SBINDIR=%{_sbindir} %{_sbindir}/%{name}
 
 %files
-%{_bindir}/%{name}.sh
+%{_sbindir}/%{name}
 
 %changelog
 * Thu Oct 28 2021 Jeremy Lenz <jlenz@redhat.com> 1.0.0-1
--- a/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
+++ b/packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
@@ -1,0 +1,43 @@
+Name:           katello-pull-provider-migrate
+Version:        1.0.0
+Release:        1%{?dist}
+Summary:        An RPM that migrates katello-agent users to the new pull provider
+BuildArch:      noarch
+
+License:        LGPLv2
+URL:            https://github.com/theforeman/foreman-packaging/packages/client/katello-pull-provider-migrate
+Source0:        %{name}.sh
+
+Requires:       bash
+Requires:       yggdrasil
+Requires:       foreman_ygg_worker
+
+%description
+An RPM that migrates katello-agent users to the new pull provider.
+
+Relies on the existing configuration in rhsm.conf to know which Katello server it should communicate with.
+The RPM install will fail if client is not registered to a Katello server.
+This RPM obtains the relevant values from the existing rhsm.conf, writes the configuration to yggdrasil's config.toml,
+and starts the pull provider agent.
+
+%prep
+
+%build
+
+%install
+rm -rf %{buildroot}
+mkdir -p  %{buildroot}%{_bindir}
+cp %{SOURCE0} %{buildroot}%{_bindir}
+
+%clean
+rm -rf %{buildroot}
+
+%post
+SYSCONFDIR=%{_sysconfdir} %{_bindir}/%{name}.sh
+
+%files
+%{_bindir}/%{name}.sh
+
+%changelog
+* Thu Oct 28 2021 Jeremy Lenz <jlenz@redhat.com> 1.0.0-1
+ update values in packaging repo
--- a/rel-eng/tito.props
+++ b/rel-eng/tito.props
@@ -976,7 +976,6 @@ whitelist =
   jsoncpp
   katello-host-tools
   katello-pull-provider-migrate
-  python-beautifulsoup4
   python-psutil
   qpid-proton
   rubygem-foreman_scap_client
@@ -988,7 +987,6 @@ disttag = .el7
 whitelist =
   katello-host-tools
   katello-pull-provider-migrate
-  python-beautifulsoup4
   python-psutil
   rubygem-foreman_scap_client
   tracer
--- a/rel-eng/tito.props
+++ b/rel-eng/tito.props
@@ -975,6 +975,8 @@ whitelist =
   gofer
   jsoncpp
   katello-host-tools
+  katello-pull-provider-migrate
+  python-beautifulsoup4
   python-psutil
   qpid-proton
   rubygem-foreman_scap_client
@@ -985,6 +987,8 @@ disttag = .el7
 # don't whitelist foreman-release here, it's already in foreman-nightly-nonscl-rhel7
 whitelist =
   katello-host-tools
+  katello-pull-provider-migrate
+  python-beautifulsoup4
   python-psutil
   rubygem-foreman_scap_client
   tracer
--- a/comps/comps-foreman-client-rhel7.xml
+++ b/comps/comps-foreman-client-rhel7.xml
@@ -16,6 +16,7 @@
       <packagereq type="default">katello-host-tools</packagereq>
       <packagereq type="default">katello-host-tools-fact-plugin</packagereq>
       <packagereq type="default">katello-host-tools-tracer</packagereq>
+      <packagereq type="default">katello-pull-provider-migrate</packagereq>
       <packagereq type="default">python-gofer</packagereq>
       <packagereq type="default">python-gofer-proton</packagereq>
       <packagereq type="default">python2-psutil</packagereq>
--- a/packages/client/katello-pull-transport-migrate/katello-pull-transport-migrate
+++ b/packages/client/katello-pull-transport-migrate/katello-pull-transport-migrate
@@ -8,7 +8,7 @@ fi
 # check if system is registered
 subscription-manager identity > /dev/null
 if [ $? -ne 0 ]; then
-    echo "Please register with subscription-manager first, then re-run `$SBINDIR/katello-pull-provider-migrate`."
+    echo "Please register with subscription-manager first, then re-run `$SBINDIR/katello-pull-transport-migrate`."
     exit 1
 fi
 
@@ -29,7 +29,7 @@ fi
 
 # fail if client is not registered to a Katello
 if ! grep -q 'prefix = \/rhsm' $KPPTEMPFILE; then
-    echo "Client is not registered to Katello; exiting"
+    echo "Client is registered to RHSM; exiting"
     exit 2
 fi
 
@@ -46,7 +46,7 @@ if [ -f $CONFIGTOML ]; then
     # make a backup of CONFIGTOML
     cp $CONFIGTOML $CONFIGTOML.bak
     cat <<EOF > $CONFIGTOML
-# yggdrasil global configuration settings written by katello-pull-provider-migrate
+# yggdrasil global configuration settings written by katello-pull-transport-migrate
 
 broker = ["tcp://$SERVER_NAME:1883"]
 cert-file = "$CERT_FILE"
--- a/packages/client/katello-pull-transport-migrate/katello-pull-transport-migrate.spec
+++ b/packages/client/katello-pull-transport-migrate/katello-pull-transport-migrate.spec
@@ -16,7 +16,7 @@ Requires:       foreman_ygg_worker
 An RPM that migrates katello-agent users to the new pull provider.
 
 Relies on the existing Red Hat Subscription Manager configuration to know which Katello server it should communicate with.
-The scriptlet will fail if the client is not registered to a Katello server.
+The scriptlet will fail if the client is not registered to a Foreman server with the Katello plugin.
 This RPM obtains the relevant values from the existing RHSM configuration, writes the configuration to yggdrasil's config.toml,
 and starts the pull provider agent.
 


packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,82 @@
...
+fi
+
+# see if /etc/yggdrasil/config.toml exists
+CONFIGTOML=$SYSCONFDIR/yggdrasil/config.toml
+if [ -f $CONFIGTOML ]; then
+    # make a backup of CONFIGTOML
+    cp $CONFIGTOML $CONFIGTOML.bak
+    echo "Writing configuration to $CONFIGTOML..."
+    # find the line starting with 'broker = ' and overwrite it with the broker url
+    sed -i "s/broker = .*/broker = [\"tcp:\/\/$SERVER_NAME:1883\"]/" $CONFIGTOML
```
adamruzicka, line 42: Instead of modifying an existing line or appending it to the file if there is none, wouldn't it be easier to just write the whole file from scratch? Currently the only untouched line from the upstream config is log level

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,67 @@
...
+    echo "Writing configuration to $CONFIGTOML..."
+    echo "# yggdrasil global configuration settings" >> $CONFIGTOML
+    echo "" >> $CONFIGTOML
+    echo "broker = [\"tcp://$SERVER_NAME:1883\"]" >> $CONFIGTOML
+    echo "cert-file = \"$CERT_FILE\"" >> $CONFIGTOML
+    echo "key-file = \"$KEY_FILE\"" >> $CONFIGTOML
+    echo "log-level = \"error\"" >> $CONFIGTOML
+
+else
+    echo "$SYSCONFDIR/yggdrasil/config.toml not found! Did `yum install yggdrasil` succeed?"
```
jturel, line 50: Would someone ever get here realistically? I think the RPM would have failed to install. Just curious - not suggesting a change

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# fail if client is not registered to a Satellite
+if [ $(grep -c "prefix = /rhsm" $RHSMCONF_PATH) -eq 0 ]; then
+    echo "Client is not registered to Katello; exiting"
+    exit 2
+fi
+
+# see if /etc/yggdrasil/config.toml exists
+CONFIGTOML=$SYSCONFDIR/yggdrasil/config.toml
+if [ -f $CONFIGTOML ]; then
```
jturel, line 37: I think it's safe to assume that by this point yggdrasil is installed, so the presence of the config file may not be a problem since you can just create it

adamruzicka: But if we just create it, then we should also handle permissions and possibly selinux.

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
jeremylenz: I suppose you could get here if you had installed yggdrasil and then removed the config file yourself :laughing:

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
jeremylenz: ooh I don't like that :(

Now I'm realizing that adding `rm $CONFIGTOML` to the script may necessitate that as well..

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,67 @@
...
+    # make a backup of CONFIGTOML
+    cp $CONFIGTOML $CONFIGTOML.bak
+    rm -f $CONFIGTOML
+    echo "Writing configuration to $CONFIGTOML..."
+    echo "# yggdrasil global configuration settings" >> $CONFIGTOML
+    echo "" >> $CONFIGTOML
+    echo "broker = [\"tcp://$SERVER_NAME:1883\"]" >> $CONFIGTOML
```
jturel, line 44: If you change `>>` to '>` here that should overwrite the contents of the config completely so the prior `rm -f` and `echo ""` should be unnecessary

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
jturel: Seems edge-casey indeed but maybe worth accounting for?

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
adamruzicka: Or we could just heredoc it
```shell
cat <<EOF > "$CONFIGTOML"
# yggdrasil global configuration settings
broker = ["tcp://$SERVER_NAME:1883"]
cert-file = "$CERT_FILE"
key-file = "$KEY_FILE"
log-level = "error"
EOF
```

jeremylenz:
Updated to edit `config.toml` with heredoc and single `>`

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,67 @@
...
+if [ -z "$SYSCONFDIR" ]; then
+    SYSCONFDIR=/etc
+fi
+
+# see if /etc/rhsm/rhsm.conf exists
+RHSMCONF_PATH=$SYSCONFDIR/rhsm/rhsm.conf
+if [ -f $RHSMCONF_PATH ]; then
+    echo "Getting configuration from $RHSMCONF_PATH..."
+    # get cert-file and key-file from rhsm.conf
+    CONSUMER_CERT_DIR=$(cat ${RHSMCONF_PATH} | grep consumerCertDir | cut -f3 -d ' ')
```
evgeni, line 13:
```suggestion
    CONSUMER_CERT_DIR=$(grep consumerCertDir ${RHSMCONF_PATH} | cut -f3 -d ' ')
```

no need to `cat` ;)

`rhsm.conf` is an ini-style file, which can have comments, and someone could have added a `# consumerCertDir is for consumer certs` before the actual assignment line, brealing your `grep|cut` logic.

The following *should* be safer: `grep '^consumerCertDir' /etc/rhsm/rhsm.conf | cut -d= -f2 | xargs` (only look for lines *starting* with `consumerCertDir`, split at `=`, and the `xargs` removes any extra whitespace.

This still would fail if there is an inline comment though…

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,67 @@
...
+    CERT_FILE=$CONSUMER_CERT_DIR/cert.pem
+    KEY_FILE=$CONSUMER_CERT_DIR/key.pem
+    # get server name from rhsm.conf
+    SERVER_NAME=$(cat $RHSMCONF_PATH | grep -m1 'hostname = ' | cut -f3 -d ' ')
```
evgeni, line 17: same comments as with the certdir :)

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+# fail if no server name, cert file or key file
+if [ -z "$SERVER_NAME" ] || [ -z "$CONSUMER_CERT_DIR" ]; then
+    echo "Unable to determine hostname or consumerCertDir from $SYSCONFDIR/rhsm/rhsm.conf; exiting"
+    exit 1
+fi
+
+# fail if client is not registered to a Satellite
+if [ $(grep -c "prefix = /rhsm" $RHSMCONF_PATH) -eq 0 ]; then
```
evgeni, line 30: `prefix=/rhsm` is valid syntax too…

and you probably want
```
if ! grep -q '^prefix\s*=\s*/rhsm'; then
```

jeremylenz:
Thanks @evgeni 

Addressed your comments and also added a sleep because I realized it takes a bit for the daemon to enter a failed state if the broker isn't running on the server.

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
jeremylenz: you forgot to add $RHSMCONF_PATH here and I was going nuts trying to figure out why grep wasn't exiting.. ha

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
jeremylenz: I love overengineering! :smile:

jturel: Why stop now?

If we're concerned about the file having been touched, subscription-manager can output a cleaned up version: `subscription-manager config --list`

jeremylenz: Are you saying you want to overwrite `rhsm.conf` with this output? But then we'd lose any comments, a lot of which are useful explanations of the settings.

jturel: I'm only saying it's possible to read a version of rhsm.conf in a seemingly standard format unaffected by comments, spacing, etc that the user has done

jeremylenz: oohh, gotcha!  :thinking:

evgeni:
Is there a better way than sleep? The sleep length will vary by system and might end up in unreproducible errors.

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
evgeni: Hah, never said it's valid syntax 🙈

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
jeremylenz: I think I'm going to go this route rather than rewriting the entire thing in Python.  Seems easier.

jeremylenz:
Seems that switching the order of `systemctl is-active yggdrasild` and `yggdrasil status` works.  Removed the sleep and updated to use `subscription-manager config --list`.

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
+
+# set SYSCONFDIR to /etc if it is not set
+if [ -z "$SYSCONFDIR" ]; then
+    SYSCONFDIR=/etc
+fi
+
+# see if /etc/rhsm/rhsm.conf exists
+RHSMCONF_PATH=$SYSCONFDIR/rhsm/rhsm.conf
+if [ -f $RHSMCONF_PATH ]; then
```
jturel, line 10: What if we removed RHSMCONF_PATH and just relied on the output of subscription-manager now that reliance on rhsm.conf itself was removed?

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,73 @@
...
+    echo "Getting configuration from subscription-manager..."
+    subscription-manager config --list > kpptempfile
```
jturel, line 12: Not sure if there's a best practice here like putting this file in /tmp ?

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
+
```
jturel, line 2: Maybe a good first check in this file would be to see if the system is registered? `subscription-manager identity` might handle that for ya

jturel (commented):
Can you point me to the bits I need to test this? (ygg, etc)

jeremylenz:
> Can you point me to the bits I need to test this?

Install on the client: ping @adamruzicka for the builds of `yggdrasil` and `foreman_ygg_worker` 
Install on the server: `yum install mosquitto` then run `mosquitto -v`

jeremylenz:
@jturel You'll also need to set up smart_proxy_remote_execution_ssh (I emailed you Adam's instructions for that.)

jeremylenz:
Updated :)

adamruzicka:
> Install on the client: ping @adamruzicka for the builds of yggdrasil and foreman_ygg_worker

You should be able to grab the rpms from https://github.com/theforeman/foreman-packaging/pull/7172 and https://github.com/theforeman/foreman-packaging/pull/7047 . Yggdrasil doesn't currently build for EL7, but the rpm for el8 should install and run there just fine

jeremylenz:
I made a custom product in my Katello and installed the RPM with its dependencies :)

```
====================================================================================================================================================
 Package                                  Arch              Version                   Repository                                               Size
====================================================================================================================================================
Installing:
 katello-pull-provider-migrate            noarch            1.0.0-1.el7               Default_Organization_Pull_Provider_yggdrasil            3.9 k
Installing for dependencies:
 foreman_ygg_worker                       x86_64            0.0.3-1.el7               Default_Organization_Pull_Provider_yggdrasil            2.8 M
 yggdrasil                                x86_64            1:0.2.0-3.el8             Default_Organization_Pull_Provider_yggdrasil            6.3 M

Transaction Summary
====================================================================================================================================================
Install  1 Package (+2 Dependent packages)

Total download size: 9.1 M
Installed size: 33 M
Is this ok [y/d/N]: y
Downloading packages:
(1/3): katello-pull-provider-migrate-1.0.0-1.el7.noarch.rpm                                                                  | 3.9 kB  00:00:00     
(2/3): foreman_ygg_worker-0.0.3-1.el7.x86_64.rpm                                                                             | 2.8 MB  00:00:00     
(3/3): yggdrasil-0.2.0-3.el8.x86_64.rpm                                                                                      | 6.3 MB  00:00:00     
----------------------------------------------------------------------------------------------------------------------------------------------------
Total                                                                                                                42 MB/s | 9.1 MB  00:00:00     
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
Warning: RPMDB altered outside of yum.
  Installing : 1:yggdrasil-0.2.0-3.el8.x86_64                                                                                                   1/3 
  Installing : foreman_ygg_worker-0.0.3-1.el7.x86_64                                                                                            2/3 
  Installing : katello-pull-provider-migrate-1.0.0-1.el7.noarch                                                                                 3/3 
Getting configuration from subscription-manager...
Starting yggdrasild...
Connection status for centos7.fedora.example.com:

● Connected to Red Hat Subscription Management
● The yggdrasil daemon is active

Manage your Red Hat connector systems: https://red.ht/connector
Complete!
Loaded plugins: fastestmirror, product-id, subscription-manager
  Verifying  : katello-pull-provider-migrate-1.0.0-1.el7.noarch                                                                                 1/3 
  Verifying  : foreman_ygg_worker-0.0.3-1.el7.x86_64                                                                                            2/3 
  Verifying  : 1:yggdrasil-0.2.0-3.el8.x86_64                                                                                                   3/3 

Installed:
  katello-pull-provider-migrate.noarch 0:1.0.0-1.el7                                                                                                

Dependency Installed:
  foreman_ygg_worker.x86_64 0:0.0.3-1.el7                                       yggdrasil.x86_64 1:0.2.0-3.el8                                      

Complete!
```

adamruzicka:
What would happen to already registered/set up clients if we rolled out a new version of this package for any reason? Would it just rerun the configuration script, effectively resetting the configuration to whatever it deems fit?

jeremylenz:
If you updated the package I think that would happen, yes.   (But I also see how it could be useful if we've changed broker settings in a future version, etc.)

jeremylenz:
We realized that systems registered with global registration won't have the Katello CA `/etc/rhsm/ca/katello-server-ca.pem` in the system's truststore. So I updated the script to get the `repo_ca_cert` from RHSM config and use it to configure `ca-root` in `config.toml`.

jeremylenz:
```
# yggdrasil global configuration settings written by katello-pull-provider-migrate

broker = ["tcp://centos7-katello-devel-stable.example.com:1883"]
cert-file = "/etc/pki/consumer/cert.pem"
key-file = "/etc/pki/consumer/key.pem"
ca-root = ["/etc/rhsm/ca/katello-server-ca.pem"]
log-level = "error"
```

jturel:
Gave this a good test and it works great. checked the added functionality around ca-cert and that works as well:


```
rm /etc/pki/ca-trust/source/anchors/katello-server-ca.pem && update-ca-trust

openssl s_client centos7-katello-devel.jturel.example.com:443
 ->
    Verify return code: 19 (self signed certificate in certificate chain)
```

Tested some package actions via REX and they worked perfectly!

jturel:
@jeremylenz can you add a card to our overflow to add some upstream docs on this? there's a section on migrating to REX from katello-agent but that only captures the SSH case right now

jturel (approved):

jeremylenz:
@evgeni or someone with merge permissions, any more concerns / can this be merged?

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,74 @@
...
+EOF
+
+else
+    echo "$SYSCONFDIR/yggdrasil/config.toml not found! Did `yum install yggdrasil` succeed?"
+    exit 1
+fi
+
+# start the yggdrasild service
+echo "Starting yggdrasild..."
+systemctl start yggdrasild
```
jeremylenz, line 60: @adamruzicka thinking about your email reply, should this be `systemctl enable yggdrasild`?

adamruzicka: More like `systemctl enable --now yggdrasild`, although I'm not sure if that's available on all supported platforms

```
systemctl enable yggdrasild
systemctl start yggdrasild
```
would be the safe bet

jeremylenz: thanks, updated! :+1:

jeremylenz:
Squashed & rebased

jeremylenz:
@theforeman/packaging Any further concerns / can this be merged? :slightly_smiling_face:

rel-eng/tito.props
```diff
@@ -985,6 +987,8 @@ disttag = .el7
 # don't whitelist foreman-release here, it's already in foreman-nightly-nonscl-rhel7
 whitelist =
   katello-host-tools
+  katello-pull-provider-migrate
+  python-beautifulsoup4
```
jlsherrill, line 991: what are these references to python-beautifulsoup4 ?  are they still needed?

jeremylenz: @jlsherrill must have been a rebase artifact.  Thought I did it right but guess not.  Removed.

jlsherrill:
i think this needs to wait until         yggdrasil & foreman_ygg_worker are packaged right?

Otherwise we break repoclosure for everything

package_manifest.yaml
```diff
@@ -776,6 +776,8 @@ foreman_client_packages:
   hosts:
     gofer: {}
     katello-host-tools: {}
+    katello-pull-provider-migrate: {}
+    python-beautifulsoup4: {}
```
evgeni, line 780: this seems not needed in this PR? (it was a tracer dep, but we dropped it, iirc)

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
```diff
@@ -0,0 +1,43 @@
+Name:           katello-pull-provider-migrate
+Version:        1.0.0
+Release:        1%{?dist}
+Summary:        An RPM that migrates katello-agent users to the new pull provider
+BuildArch:      noarch
+
+License:        LGPLv2
+URL:            https://github.com/theforeman/foreman-packaging/packages/client/katello-pull-provider-migrate
```
evgeni, line 8: this URL is 404. :)

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
```diff
@@ -0,0 +1,43 @@
...
+cp %{SOURCE0} %{buildroot}%{_bindir}
+
+%clean
+rm -rf %{buildroot}
+
+%post
+SYSCONFDIR=%{_sysconfdir} %{_bindir}/%{name}.sh
+
+%files
+%{_bindir}/%{name}.sh
```
evgeni, line 39: while this is (currently) implemented in shell, I don't see a compelling reason to call the file in `/usr/bin` `.sh`, as we might decide to reimplement it in Python or Go tomorrow.

Also, I think this should be in `sbin`, given the commands need to run as root.

Or maybe, not in any *user* visible dir at all, or are users supposed to run this script on their own too?
If they aren't, `/usr/libexec` is a good place for things that are only executed by tools.

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,75 @@
...
+systemctl enable yggdrasild
+systemctl start yggdrasild
+
+# check status of yggdrasild and fail if it is not running
+# possible failure reason: incorrect protocol (should be tcp:// or mqtt://) or port (should be 1883)
+# also, cert-file and key-file must be valid
+# and broker must be running on the server
+yggdrasil status
+
+if [ $(systemctl is-active yggdrasild) != "active" ]; then
```
evgeni, line 69:
```suggestion
if ! systemctl is-active --quiet yggdrasild; then
```

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# check if system is registered
+subscription-manager identity > /dev/null
+if [ $? -ne 0 ]; then
+    echo "Please register with subscription-manager first, then re-run `katello-pull-provider-migrate.sh`."
+    exit 1
+fi
+
+echo "Getting configuration from subscription-manager..."
+KPPTEMPFILE=/tmp/kpp_tempfile
```
evgeni, line 11:
```suggestion
KPPTEMPFILE=$(mktemp)
```

Otherwise you allow attackers to overwrite arbitrary files by pre-creating `/tmp/kpp_tempfile`

evgeni:
I've left a few comments (most are not strictly packaging related tho), but would raise a few more generic questions/concerns:

1. Why should this file live in packaging? This effectively makes it impossible to have any kind of CI for it, which I think is wrong for a rather crucial piece of the migration workflow.
2. If a user migrates to ygg using this package, and then decides to change anything in the config and we also decide to push out a new version, the users config is gone, which I find bad. (Yes, it will be in the `.bak` file, but only until the user accidentally installs the package again for some reason or runs the script in `/usr/bin`)
3. If the user already migrated, and we push out a new version, the on disk config is changed, but yggd is not actually *re*started, so that the running config and the on disk config don't match anymore, surprising the user on next boot

So, packaging wise (after the comments were addressed) "ACK"
But platform/architecture wise "NACK" :stuck_out_tongue_closed_eyes:

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
jeremylenz: It may be nice to leave it user-accessible in the case where the user has messed up their config and wants to reset it to the correct one.

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.sh
jeremylenz: PR reviews are the best Linux crash course :smile:

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
jeremylenz: I think I meant
```
https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/client/katello-pull-provider-migrate
```
which is a 404 now but won't be when this is merged. I'm not sure what other URL would be correct..

evgeni: https://theforeman.org is a fine URL too ;)

or https://github.com/katello/katello-pull-provider-migrate ;)

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
jeremylenz: Removed the `.sh` and put it in `/usr/sbin`. :+1:

jeremylenz:
@evgeni updated

package_manifest.yaml
jeremylenz: Missed that one.  removed.

ehelms:
> 1. Why should this file live in packaging? This effectively makes it impossible to have any kind of CI for it, which I think is wrong for a rather crucial piece of the migration workflow.

I tend to agree with this sentiment. We have put code in this packaging repository in the past and it has not aged well, and been hard to develop. From taking a look at the workflow and proposal here, there is a script doing all the heavy lifting and the RPM is simply the delivery mechanism for said script dude to how katello-agent functionality works.

While it may seem silly to have a dedicated repository for a single tiny script, I would concur that its the right direction to break apart the abstractions. It will provide potential for testing, but it also keeps the concerns separate, releases and cherry picking if needed easier. If a small repository seems overkill, I'd suggest putting this in `katello-host-tools` repository and creating this as a sub-package of either katello-host-tools or katello-agent or keep it as a stand-alone RPM.

comps/comps-foreman-client-el8.xml
```diff
@@ -15,6 +15,7 @@
       <packagereq type="default">katello-agent</packagereq>
       <packagereq type="default">katello-host-tools</packagereq>
       <packagereq type="default">katello-host-tools-tracer</packagereq>
+      <packagereq type="default">katello-pull-provider-migrate</packagereq>
```
ehelms, line 18: Shaving some yaks, IIRC we landed that it is a pull transport rather than provider. Where we are clarifying that "provider" should be used to indicate the user-facing mechanism by which actions are performed (e.g. Script, Ansible) and transport should indicate how the action is delivered. I'd like us to try to be consistent with our use of terminology so I think that would modify this to `katello-pull-transport-migrate` or tie it into REX `katello-remote-execution-pull-migrate` or some such.

@adamruzicka @ezr-ondrej if y'all had a thought

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate
```diff
@@ -0,0 +1,80 @@
...
+
+# fail if no server name, cert dir, or ca file
+if [ -z "$SERVER_NAME" ] || [ -z "$CONSUMER_CERT_DIR" ] || [ -z "$CA_FILE" ]; then
+    echo "Unable to determine config from `subscription-manager config --list`; exiting"
+    exit 1
+fi
+
+# fail if client is not registered to a Katello
+if ! grep -q 'prefix = \/rhsm' $KPPTEMPFILE; then
+    echo "Client is not registered to Katello; exiting"
```
ehelms, line 32: You may consider not using `Katello` here to reduce branding scope.

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate
```diff
@@ -0,0 +1,80 @@
...
+fi
+
+rm -f $KPPTEMPFILE
+
+# set SYSCONFDIR to /etc if it is not set
+if [ -z "$SYSCONFDIR" ]; then
+    SYSCONFDIR=/etc
+fi
+
+# see if /etc/yggdrasil/config.toml exists
```
ehelms, line 43: This will be a fun one, and is viable as a phase 2 addition (but should be tracked if it is). On RHEL, this won't be `yggdrasil` but rather `rhc` and you'll need to either include paths for both cases or add some conditional logic to determine which is present (rhc or yggdrasil) and set the paths based on this information.

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
```diff
@@ -0,0 +1,43 @@
...
+Release:        1%{?dist}
+Summary:        An RPM that migrates katello-agent users to the new pull provider
+BuildArch:      noarch
+
+License:        LGPLv2
+URL:            https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/client/katello-pull-provider-migrate
+Source0:        %{name}
+
+Requires:       bash
+Requires:       yggdrasil
```
ehelms, line 12: See https://github.com/theforeman/foreman-packaging/pull/7210/files#r766697887

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate.spec
```diff
@@ -0,0 +1,43 @@
...
+Requires:       foreman_ygg_worker
+
+%description
+An RPM that migrates katello-agent users to the new pull provider.
+
+Relies on the existing Red Hat Subscription Manager configuration to know which Katello server it should communicate with.
+The scriptlet will fail if the client is not registered to a Katello server.
```
ehelms, line 19: I would either drop `Katello`, use `Foreman` instead. Or `to a Foreman server with the Katello plugin.` I generally want us all to get away from calling it a Katello server and think of it as a Foreman server with X functionality (in this case Katello).

jturel:
> > Why should this file live in packaging? This effectively makes it impossible to have any kind of CI for it, which I think is wrong for a rather crucial piece of the migration workflow

Are we unable to write a bats test for this package if the code lives in this repo? My guidance to @jeremylenz was to put the code here primarily because the important part at the time was figure out the logic and how pull provider works, etc. Where the code lives is relatively minor to all of that and I'm not against it living elsewhere whatsoever.

evgeni:
bats is way too late. bats run on released RPMs. what I am aiming at is direct feedback when someone opens a PR. *that* is not possible with the current setup (and is exactly what has bitten us with katello-change-hostname recently)

jturel:
> bats is way too late. bats run on released RPMs. what I am aiming at is direct feedback when someone opens a PR. _that_ is not possible with the current setup (and is exactly what has bitten us with katello-change-hostname recently)

Comparing this script to change-hostname is one of those apples vs oranges things. That's a script that has a LOT of complexity (execution paths thru the code) while this script has virtually one path. My point being that a reviewer testing a change to this script would inherently be validating the whole thing. I'm not making an argument to keep it in this repo - just elaborating on the comparison to k-h-c.

evgeni:
And having scripts not in packaging, helps us not to play redmine pingpong: https://projects.theforeman.org/issues/34099

comps/comps-foreman-client-el8.xml
jeremylenz: I like `katello-pull-transport` and `katello-pull-transport-migrate`! 
KPT™
:+1:

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate
jeremylenz: But if I delete `Katello` it won't be clear that it will exit if the client is registered to RHSM but not to a Katello / Satellite. :thinking:

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate
jeremylenz: > viable as a phase 2 addition

Does this mean I don't have to worry about it just yet?

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate
jeremylenz: What about
```
Client is registered to RHSM; exiting
```

jeremylenz:
I'll close this PR soon once we decide for sure where the code will live. Converting to draft for now to prevent accidental merges :P 

Until then, @ehelms I addressed some of your comments and left some questions.

jeremylenz:
> While it may seem silly to have a dedicated repository for a single tiny script, I would concur that its the right direction to break apart the abstractions. It will provide potential for testing, but it also keeps the concerns separate, releases and cherry picking if needed easier. If a small repository seems overkill, I'd suggest putting this in katello-host-tools repository and creating this as a sub-package of either katello-host-tools or katello-agent or keep it as a stand-alone RPM.

I think we're leaning towards a separate repo rather than `katello-host-tools`.

But that got me thinking, When users raise bugs around this, they will probably choose the REX category over anything else. 
 Since `katello-pull-transport` is a REX transport method rather than something tightly coupled to Katello functionality, would it be better to have this new repo live in `theforeman` org alongside `foreman_remote_execution` and `smart_proxy_remote_execution_ssh`? @ehelms @evgeni thoughts?

evgeni:
I'm happy with either Org hosting this.

ehelms:
> But that got me thinking, When users raise bugs around this, they will probably choose the REX category over anything else.
> Since `katello-pull-transport` is a REX transport method rather than something tightly coupled to Katello functionality, would it be better to have this new repo live in `theforeman` org alongside `foreman_remote_execution` and `smart_proxy_remote_execution_ssh`? @ehelms @evgeni thoughts?

Absolutely let's put it in theforeman org.

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate
ehelms: I could see how RHSM is confusing as well. So I might opt for 'Foreman' that at least is the application that is running.

Is checking the configuration of `rhsm.conf` an actual teller of registration though? If I configure `rhsm.conf` to point at a Foreman or smart-proxy, I am not actually registered at that point. Could we use subscription-manager to report registration?

e.g, if not registered I see:

```
# subscription-manager status
+-------------------------------------------+
   System Status Details
+-------------------------------------------+
Overall Status: Unknown

System Purpose Status: Unknown
```

packages/client/katello-pull-provider-migrate/katello-pull-provider-migrate
ehelms: Yea, I think you could focus on CentOS support to start with and move to RHEL support later.

jeremylenz:
> Absolutely let's put it in theforeman org

sounds like a plan! @evgeni can you please do the honors of creating `theforeman/katello-pull-transport-migrate` since Eric is out? :smile:

evgeni:
> katello-pull-transport-migrate

done, you're admin of that repo :)

jeremylenz:
:eyes: https://github.com/theforeman/katello-pull-transport-migrate/pull/1

jeremylenz:
This work is continued in https://github.com/theforeman/katello-pull-transport-migrate/pull/2, so closing here.

jturel:
@jeremylenz  we'll still need a packaging PR with your spec file

PR #7212, opened 2021-10-28
Bump Hammer Foreman Azure_RM to 0.2.2

chris1984:
@ekohl here is the packaging pr


chris1984:
@ekohl fixed the spec file, i swear i used the `bump_rpm` script so that git annex did the correct things with the file, idk.

ekohl (approved):
Thanks!

PR #7213, opened 2021-10-31
Update rubygem-foreman_statistics to 2.0.0

ezr-ondrej:
DEB: #7214

PR #7214, opened 2021-10-31
Update ruby-foreman-statistics to 2.0.0

ezr-ondrej:
RPM: #7213


ekohl (changes requested):
Could you also change the minimum Foreman version in debian/control here?

ezr-ondrej:
Is this ok, or do I need to have the `~rc` part?

ekohl (commented):
I think you need the rc part. Otherwise you can't install it with RCs. `3.1.0~rc1 < 3.1.0`.

ezr-ondrej:
Ah, right :)

ekohl (approved):

PR #7220, opened 2021-11-02
Updated hammer_cli to 3.0.1

ofedoren:
(cherry picked from commit 82678dd87cd4e29d918b5b97b5ff4e7433bb3890)


evgeni:
for 3.0, you'll also need to bump bionic, which we dropped in develop

evgeni (approved):

PR #7224, opened 2021-11-04
fix debian db migrations and service restarts

evgeni:

### DIFF
--- a/debian/buster/foreman/rules
+++ b/debian/buster/foreman/rules
@@ -47,13 +47,9 @@ build:
 	dh $@ --with=systemd
 
 override_dh_systemd_start:
-	dh_systemd_start --no-start --restart-after-upgrade --name=foreman.service --package=foreman-service
-	dh_systemd_start --no-start --restart-after-upgrade --name=foreman.socket --package=foreman-service
-	dh_systemd_start --no-start --name=dynflow-sidekiq@orchestrator --package=foreman-dynflow-sidekiq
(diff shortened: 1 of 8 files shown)


evgeni:
our plugins still call `db:migrate`, so if users have plugins installed, this might still fail…

debian/buster/foreman/rules
```diff
@@ -47,13 +47,9 @@ build:
 	dh $@ --with=systemd
 
 override_dh_systemd_start:
-	dh_systemd_start --no-start --restart-after-upgrade --name=foreman.service --package=foreman-service
-	dh_systemd_start --no-start --restart-after-upgrade --name=foreman.socket --package=foreman-service
-	dh_systemd_start --no-start --name=dynflow-sidekiq@orchestrator --package=foreman-dynflow-sidekiq
-	dh_systemd_start --no-start --name=dynflow-sidekiq@worker --package=foreman-dynflow-sidekiq
+	dh_systemd_start --no-start --restart-after-upgrade --package=foreman-service
+	dh_systemd_start --no-start --restart-after-upgrade --package=foreman-dynflow-sidekiq
```
evgeni, line 51: this is technically a noop, as the package doesn't contain any restartable services (the template isn't restartable), so we *could* drop it…

ehelms (approved):
Seems sane and brings closer alignment with RPM environments. I can take this for a test drive today or we can merge and see what those pipes say.

evgeni:
let's go for a pipe ;)

PR #7225, opened 2021-11-04
foreman-puppet: don't run db:migrate in postinst

evgeni:

### DIFF
--- a/plugins/ruby-foreman-puppet/debian/postinst
+++ b/plugins/ruby-foreman-puppet/debian/postinst
@@ -38,17 +38,10 @@ else
   fi
 fi
 
-# DB migrate/seed after install
-if [ -f /usr/share/foreman/config/database.yml ]; then
-  if [ ! -z "${DEBUG}" ]; then
-    /usr/sbin/foreman-rake db:migrate || true
-    /usr/sbin/foreman-rake db:seed || true
-    /usr/sbin/foreman-rake apipie:cache:index || true
-  else
-    /usr/sbin/foreman-rake db:migrate >> $LOGFILE 2>&1 || true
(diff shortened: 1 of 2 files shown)


plugins/ruby-foreman-puppet/debian/postinst
```diff
@@ -38,19 +38,6 @@ else
...
-# DB migrate/seed after install
-if [ -f /usr/share/foreman/config/database.yml ]; then
-  if [ ! -z "${DEBUG}" ]; then
-    /usr/sbin/foreman-rake db:migrate || true
-    /usr/sbin/foreman-rake db:seed || true
-    /usr/sbin/foreman-rake apipie:cache:index || true
-  else
-    /usr/sbin/foreman-rake db:migrate >> $LOGFILE 2>&1 || true
-    /usr/sbin/foreman-rake db:seed >> $LOGFILE 2>&1 || true
-    /usr/sbin/foreman-rake apipie:cache:index >> $LOGFILE 2>&1 || true
```
evgeni, line 50: huh, I guess *this* should still happen?

ehelms: I do think we should keep, I wondered if this should be the plugin version of it like RPM?

https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/foreman/foreman.spec#L911

evgeni: yeah (and no), the line you're linking to is the compile step, not the index step

evgeni:
yepp, this is def needed to unbreak upgrades

ehelms (approved):

PR #7226, opened 2021-11-04
put plugin postinstall into a central file, which can be loaded

evgeni:

### DIFF
--- a/debian/buster/foreman/foreman-plugin.postinst-lib
+++ b/debian/buster/foreman/foreman-plugin.postinst-lib
@@ -1,0 +1,50 @@
+foreman_plugin_install() {
+  PLUGIN="${1}"
+  PREVIOUS_VERSION="${2}"
+  PLUGIN_HAS_APIPIE="${3:-true}"
+  PLUGIN_HAS_WEBPACK="${4:-true}"
+  LOGFILE="${5:-/var/log/foreman-install.log}"
+
+  BUNDLE=bundle
+  [ -h /usr/bin/foreman-ruby ] && BUNDLE="/usr/bin/foreman-ruby /usr/bin/bundle"
+
+  # Update gems
+  export HOME=/usr/share/foreman
+  cd /usr/share/foreman
+  if [ ! -z "${DEBUG}" ]; then
+    if [ -n "${PREVIOUS_VERSION}" ] ; then
+      $BUNDLE update $PLUGIN --local
+    else
+      $BUNDLE install --local --no-prune
+    fi
+  else
+    if [ -n "${PREVIOUS_VERSION}" ] ; then
+      $BUNDLE update $PLUGIN --local 2>&1 >> $LOGFILE
+    else
+      $BUNDLE install --local --no-prune 2>&1 >> $LOGFILE
+    fi
+  fi
+
+  if [ "${PLUGIN_HAS_APIPIE}" = "true" ]; then
+    if [ ! -z "${DEBUG}" ]; then
+      /usr/sbin/foreman-rake apipie:cache:index || true
+    else
+      /usr/sbin/foreman-rake apipie:cache:index >> $LOGFILE 2>&1 || true
+    fi
+  fi
+
+  if [ "${PLUGIN_HAS_WEBPACK}" = "true" ]; then
+    # The output sometimes contain additional data like the following line:
+    #    Found no changes, using resolution from the lockfile
+    # But we really just want the path, so lets grep
+    PLUGIN_ROOT="$(${BUNDLE} show ${PLUGIN} | grep '/usr/share/foreman')"
+    # Make sure we got a path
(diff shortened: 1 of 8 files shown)


evgeni:
this (after 7 years), fixes https://projects.theforeman.org/issues/7988 :tada:

ehelms (approved):

debian/buster/foreman/foreman-plugin.postinst-lib
```diff
@@ -0,0 +1,50 @@
+foreman_plugin_install() {
+  PLUGIN="${1}"
+  PREVIOUS_VERSION="${2}"
+  PLUGIN_HAS_APIPIE="${3:-true}"
+  PLUGIN_HAS_WEBPACK="${4:-true}"
```
ekohl, line 5: Can we be smart? In RPMs we have:
https://github.com/theforeman/foreman-packaging/blob/461c8df52c17de853a8a0c0657a7d18135a73229/gem2rpm/foreman_plugin.spec.erb#L20-L24

Should we look for files in `$PLUGIN_ROOT` and determine the values dynamically if the parameter wasn't passed?

ekohl (commented):
I like the direction. Wouldn't this actually resolve https://projects.theforeman.org/issues/7988?

evgeni:
To truly solve https://projects.theforeman.org/issues/7988 this needs to be used by all plugins, and then we'd need to drop the duplication in the core packages too ;)

debian/buster/foreman/foreman-plugin.postinst-lib
evgeni: I guess we don't need the parameter at all in that case ;)

evgeni: So, something like this:

```shell
  # The output sometimes contain additional data like the following line:
  #    Found no changes, using resolution from the lockfile
  # But we really just want the path, so lets grep
  PLUGIN_ROOT="$(${BUNDLE} show ${PLUGIN} | grep '/usr/share/foreman')"
  # Make sure we got a path
  if [ -n "${PLUGIN_ROOT}" ]; then
    if [ -d "${PLUGIN_ROOT}/webpack" ]; then
      mkdir -p $PLUGIN_ROOT/public
      ln -snf /var/lib/foreman/public/webpack/ $PLUGIN_ROOT/public/webpack
    fi
  fi
```

evgeni: Even tho, looking at the actual plugins, those are all plugins that *also* have a `package.json`

evgeni: And calling `apipie:cache:index` doesn't technically hurt for plugins without API, just takes a bit of time.

evgeni: And because I am already talking to myself: I don't think creating this symlink for plugins that don't have package.json/webpack will actually hurt.

ekohl: > I guess we don't need the parameter at all in that case ;)

I thought it could be useful to override. It's better to have that ability now than to need it at some point and having to go through the dance with updating Foreman and then the plugin.

> Even tho, looking at the actual plugins, those are all plugins that _also_ have a `package.json`

Yes. The only odd one used to be somewhere in Katello where there was a package.json but didn't use webpack. For Debian I think we can ignore that edge case.

ehelms:
I am good with giving this a go, I think we should do so now or wait till RC1 is out given we are aiming to branch next week Tuesday.

debian/buster/foreman/foreman-plugin.postinst-lib
evgeni: Do you absolutely want the "smart" part?

The current layout works and I don't have the time to play around with the smartness for RC1. :(

evgeni:
I prefer now, as otherwise it means if someone tries a test-upgrade to RC1, they might end up in an error (recoverable, but still).

debian/buster/foreman/foreman-plugin.postinst-lib
ekohl: I'm OK with postponing smartness.

ekohl (approved):
I'd also prefer this sooner rather than later. My biggest concern here is that we end up building ruby-foreman-puppet before foreman. Would that still work?

evgeni:
@ekohl this is a install-time thing (postinst is just a shell script that runs when dpkg installs the package), so build order doesn't matter.

debian/buster/foreman/foreman-plugin.postinst-lib
evgeni: I have added it to https://projects.theforeman.org/issues/7988, as it's still "in progress"

PR #7232, opened 2021-11-08
drop deprecated plugins: azure, digitalocean, xen

evgeni:


evgeni:
I am not 100% sure xen is deprecated, but I think it is?

cc @tbrisker

ehelms (approved):
Looks like the conversation was started with no one really stepping up. Does not look like we have officially dropped it per say but if it has troubles building or working with 3.0 then I say drop it here and pick it up later.

https://community.theforeman.org/t/foreman-xen-plugin-is-unmaintained-what-should-be-done/21688

evgeni:
Oh yeah, forgot about this one. We started discussing it in the release meeting today, and I've opened https://projects.theforeman.org/issues/33867 to track this and add a list of dropped plugins to the release notes.

PR #7240, opened 2021-11-09
foreman-webhooks: use foreman-plugin.postinst-lib

evgeni:

### DIFF
--- a/plugins/ruby-foreman-webhooks/debian/rules
+++ b/plugins/ruby-foreman-webhooks/debian/rules
@@ -19,7 +19,7 @@ build:
 		/usr/bin/npm install --no-audit --no-optional --unsafe-perm && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production DATABASE_URL=nulldb://nohost && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
-			OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
+			OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=production \
 			DATABASE_URL=nulldb://nohost FOREMAN_APIPIE_LANGS=en_US \
 	)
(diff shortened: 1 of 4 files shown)


evgeni:
`Setting 'administrator' is already defined, please avoid collisions`

@lzap is webhooks not compatible with Foreman 3.1?

lzap:
Webhooks has no settings, what is the error about?

lzap:
So on IRC @ofedoren found out this is caused by `eager_load!` we do in development Rails environment to get to all models and classes. It appears we call `apipie:cache` with development environment set.

Webhooks work fine on 3.1 on my dev setup for the record.

evgeni:
don't merge just yet, need to compare the results first

plugins/ruby-foreman-webhooks/debian/rules
```diff
@@ -19,7 +19,7 @@ build:
 		/usr/bin/npm install --no-audit --no-optional --unsafe-perm && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:assets:precompile[$(PLUGIN)] RAILS_ENV=production DATABASE_URL=nulldb://nohost && \
 		/usr/bin/foreman-ruby /usr/bin/bundle exec rake plugin:apipie:cache[$(PLUGIN)] cache_part=resources \
-			OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=development \
+			OUT=/var/lib/foreman/public/apipie-cache/plugin/$(PLUGIN) RAILS_ENV=production \
```
ekohl, line 22: I do wonder if this needs to be done for all plugins, but I'm fine merging this for now. I'm just going to copy what @ofedoren said on IRC here to make it easier to read back:
> it happens only in development environment probably due to https://github.com/theforeman/foreman/blob/bca1b576776197dcaff1d29b6e23e66463d4dd30/lib/foreman/event_subscribers.rb#L6-L9. all_observable_events are used in API docs for listing events that user can specify as a parameter...

ekohl (approved):

plugins/ruby-foreman-webhooks/debian/rules
evgeni: Probably. But I think this should happen via the central makefile I've been working on ;)

Other plugins don't fail as they don't use events.

ekohl: :+1: for solving it centrally.

PR #7243, opened 2021-11-09
foreman-omaha: use foreman-plugin.postinst-lib

evgeni:


evgeni:
the old `debian/rules` was broken, so I flipped to the shared one and that seems to build :woman_shrugging:

ekohl (approved):

PR #7246, opened 2021-11-09
foreman-rescue: use foreman-plugin.postinst-lib

evgeni:

### DIFF
--- a/plugins/ruby-foreman-rescue/debian/changelog
+++ b/plugins/ruby-foreman-rescue/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-rescue (2.0.1-2) stable; urgency=low
+
+  * use foreman-plugin.postinst-lib from foreman
+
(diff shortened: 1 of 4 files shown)


ekohl (approved):
Tests are still running (and we don't have automerge enabled on this repo).

evgeni:
tests are actually failing, old debian/rules is old.

ekohl (changes requested):
Looks like they just failed after I submitted the build. Looks like there's a gem issue. Should it depend on foreman-postgresql?

evgeni:
> Looks like they just failed after I submitted the build. Looks like there's a gem issue. Should it depend on foreman-postgresql?

No, the actual error is further up. But TL;DR: this needs to use the nulldb setup, which it doesn't yet.

ekohl (approved):
Yay even more lines removed.

PR #7257, opened 2021-11-09
Swap wget for curl dependency on foreman-proxy

ekohl:
This reflects that the Smart Proxy now [uses curl instead of wget](https://github.com/theforeman/smart-proxy/commit/3d87c6feaa8caaf9b8c040f96061c3b8c33a39ef).


lzap (approved):
Ah damn sorry I forgot! Good catch. Green!

ehelms (approved):

PR #7269, opened 2021-11-11
Bump Hammer Katello to 1.4.0

chris1984:


chris1984:
@Odilhao here is the one for nightly and the new hammer release

Odilhao (approved):

PR #7275, opened 2021-11-12
Update foreman-packaging deb/develop to ensure nightlies build

Odilhao:


evgeni:
ACK pending develop has been bumped to 3.2 and thus this actually builds ;)

PR #7291, opened 2021-11-16
Update rubygem-foreman_remote_execution to 5.0.1

adamruzicka:
CPs:
- 3.1


Odilhao:
[test rpm]

Odilhao:
Hi @adamruzicka looks like we will need to add BuildRequires `tfm-npm(graphql-tag)`

```
ModuleNotFoundError: Module not found: Error: Can't resolve 'graphql-tag/loader' in '/builddir/build/BUILD/foreman_remote_execution-5.0.0/usr/share/foreman'
```

We are not seeing the graphql-tag being required here as well https://github.com/theforeman/foreman_remote_execution/blob/29d7c2036893a800a75ab5d26c4c808a9de7e721/package.json

adamruzicka:
@mariaaga is it possible `graphql-tag` was getting pulled as a transitive dependency of something so we didn't need to declare it explicitly?

evgeni:
We had this with `foreman_openscap` too. [It's not in the `package.json`](https://github.com/theforeman/foreman_openscap/blob/master/package.json), but the RPM spec has it listed:

https://github.com/theforeman/foreman-packaging/blob/d6215a4c4119f8ebba56cb71697b7684a076bc2b/packages/plugins/rubygem-foreman_openscap/rubygem-foreman_openscap.spec#L39-L42

adamruzicka:
Yeah, that did the trick

Odilhao (approved):

PR #7300, opened 2021-11-18
Update katello-certs-tools to 2.8.1

ehelms:


ehelms:
Repoclosure failure is unrelated and from local testing appears to either be something cached on the nodes or just network issues talking to the AppStream repository.

evgeni:
Yeah, it is definitely not the change to the devel repo. I'd merge and see what happens in the pipeline

evgeni:
it'll get green if you include the changes from https://github.com/theforeman/foreman-packaging/pull/7304

evgeni (approved):

PR #7302, opened 2021-11-18
Update katello-certs-tools to 2.8.1

ehelms:


ehelms:
Repoclosure failure is unrelated and from local testing appears to either be something cached on the nodes or just network issues talking to the AppStream repository.

evgeni:
Is this safe to pick to 2.5? Given it had 2.7.z before?
Also, if it is, can you include the old changelog entries? It will be tagged, not built, so the rpm wouldn't match git?

ehelms:
> Is this safe to pick to 2.5? Given it had 2.7.z before? Also, if it is, can you include the old changelog entries? It will be tagged, not built, so the rpm wouldn't match git?

I feel OK with it. Before I opened this I looked at the diff and nothing stuck out to me.

ehelms:
Here ya go -- added changelog entry and cherry picked the repoclosure fix

evgeni (approved):

evgeni:
@ehelms I think you forgot to actually tag the build into 2.5 ;)

PR #7304, opened 2021-11-19
exclude ruby 3 packages from repoclosure

evgeni:
we're currently building with ruby 2.7, but the way "dnf repoclosure"
works it ends up detecting 3.0 as "latest" and then breaks our packages
that need 2.7

this will never happen in the real world, as there one needs to enable
the right version of the module explicitly


evgeni:
I am going to merge this w/o review, as I want to get nighty green again.

@ehelms @ekohl @Odilhao @pcreech you can yell at me later

PR #7312, opened 2021-11-23
bump supported openscap version

evgeni:


evgeni:
this needs to be backported to all supported branches

Odilhao (approved):

PR #7349, opened 2021-11-29
Update rubygem-smart_proxy_remote_execution_ssh to 0.5.0

adamruzicka:
(cherry picked from commit e0d143ffa5dd85d1fd5f160762b029677b8c80a1)

Requires https://github.com/theforeman/foreman-packaging/pull/7350

PR #7327, opened 2021-11-24
Add rubygem-mqtt package

Odilhao:

PR #7355, opened 2021-11-29
Update smart_proxy_remote_execution_ssh to 0.5.0

adamruzicka:
(cherry picked from commit 5985511c0e446839ee52971ad18dfa02a3a62387)

### DIFF
--- a/plugins/smart_proxy_remote_execution_ssh/changelog
+++ b/plugins/smart_proxy_remote_execution_ssh/changelog
@@ -1,3 +1,9 @@
+ruby-smart-proxy-remote-execution-ssh (0.5.0-1) stable; urgency=low
+
+  * 0.5.0 released
+
+ -- Adam Ruzicka <aruzicka@redhat.com>  Tue, 16 Nov 2021 10:15:22 +0100
+
 ruby-smart-proxy-remote-execution-ssh (0.4.1-1) stable; urgency=low
 
   * 0.4.1 released
--- a/plugins/smart_proxy_remote_execution_ssh/control
+++ b/plugins/smart_proxy_remote_execution_ssh/control
@@ -11,7 +11,7 @@ Package: ruby-smart-proxy-remote-execution-ssh
 Architecture: all
 XB-Ruby-Versions: ${ruby:Versions}
 Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter, foreman-proxy (>= 1.11.0~rc3),
-  ruby-smart-proxy-dynflow (>= 0.5.0), ruby-smart-proxy-dynflow (<< 1.0.0), ruby-net-ssh
+  ruby-smart-proxy-dynflow (>= 0.5.0), ruby-smart-proxy-dynflow (<< 1.0.0), ruby-net-ssh (>= 4.2.0)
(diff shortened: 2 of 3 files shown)


Odilhao (approved):

shr-project:
Shouldn't this depend on ruby-mqtt as in rpm packaging?

it was added there with https://github.com/theforeman/foreman-packaging/pull/7349, but ruby-mqtt isn't packaged for debian only for rpm as in:
https://github.com/theforeman/foreman-packaging/pull/7327

On Ubuntu 20.04 with foreman 3.1 I'm seeing:
```
Message
Couldn't enable 'ssh'
Backtrace
cannot load such file -- mqtt: /usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
/usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
/usr/lib/ruby/vendor_ruby/smart_proxy_remote_execution_ssh/actions/pull_script.rb:1:in `<top (required)>'
/usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
/usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
/usr/lib/ruby/vendor_ruby/smart_proxy_remote_execution_ssh/actions.rb:4:in `<module:Actions>'
/usr/lib/ruby/vendor_ruby/smart_proxy_remote_execution_ssh/actions.rb:2:in `<module:Ssh>'
/usr/lib/ruby/vendor_ruby/smart_proxy_remote_execution_ssh/actions.rb:1:in `<top (required)>'
```

adamruzicka:
> Shouldn't this depend on ruby-mqtt as in rpm packaging?

Yes, it should. Thank you for pointing that out

evgeni:
This should only fail when using the MQTT backend, right?

Because our tests call `hammer job-invocation create --job-template 'Run Command - SSH Default' --inputs 'command=uptime' --search-query "name = $HOSTNAME"` and that seems to still work fine?

adamruzicka:
This is interesting. It attempts to load to load `mqtt` on startup without any failsafes, however if an exception is raised doing loading, smart proxy merrily ignores it and still marks the feature as available. I think this is not the first time we're running into this (cc @ekohl). Although it seems to be up and apparently working in our tests, I wouldn't vouch for it being 100% ready

```
$ bundle exec ruby bin/smart-proxy | grep mqtt &
2021-12-10T17:23:30  [W] Error details for Couldn't enable 'ssh': <LoadError>: cannot load such file -- mqtt

$ curl http://localhost:8000/features
["ansible","dynflow","logs","probing","salt","shellhooks","ssh"]
```

evgeni:
@adamruzicka packaging fixes in https://github.com/theforeman/foreman-packaging/pull/7416 -- mind having a quick look?

the smart-proxy behavior is disturbing :(

ekohl:
I think https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/57 would have caught that. Right now it's loading the plugin in `after_activation` (which is deprecated). The `load_classes` method is ran before the plugin is marked as `:running`.

PR #7369, opened 2021-12-01
Add GH workflows to automatically update plugins

ezr-ondrej:
Example run:
https://github.com/ezr-ondrej/foreman-packaging/actions/runs/1607974007

Examples for resulting PRs in my repo: https://github.com/ezr-ondrej/foreman-packaging/pulls


ehelms:
I had a couple questions around the mechanics of it.

 1) How does this handle not opening multiple of the same PR?
 2) For this type of GH workflow versus Jenkins, how does one monitor the job and get awareness of failures?

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,109 @@
+name: Update plugins
+on:
+  workflow_dispatch:
```
ezr-ondrej, line 3: This mean we need to trigger it manually, which I'd prefer for the time being, once it's stable enough and the list of plugins is finalized, we can switch to schedule

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,109 @@
...
+    env:
+      PLUGIN_DIR: "packages/plugins/rubygem-${{ matrix.plugin }}"
+      SPEC_FILE: "packages/plugins/rubygem-${{ matrix.plugin }}/rubygem-${{ matrix.plugin }}.spec"
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get install -y rpm curl jq git-annex python3-pip
+        sudo gem install gem2rpm
+        sudo apt-get install -y libcurl4-openssl-dev
+        sudo gem install gem-compare
```
ekohl, line 37: In https://github.com/theforeman/foreman-packaging/commit/71c11e00084045b718ffdef9b17c5ac9e727c86b we dropped this
```suggestion
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,109 @@
...
+    - name: Initialize git annex
+      run: git annex init
+    - name: Try to bump plugin
+      run: ./bump_rpm.sh $PLUGIN_DIR -a "$CHANGE_AUTHOR" --no-commit
+      env:
+        CHANGE_AUTHOR: Github Actions <noreply@github.com>
+    - name: Open a PR
+      uses: peter-evans/create-pull-request@v3
+      with:
+        commit-message: "Updates ${{ matrix.plugin }} to ${{ steps.latest_gem.outputs.version }}"
```
ekohl, line 62: I'd like to avoid duplicating the logic from `bump_rpm.sh` but if you do, please copy its syntax. It uses the convention as specified in https://cbea.ms/git-commit/#imperative.
```suggestion
        commit-message: "Update ${{ matrix.plugin }} to ${{ steps.latest_gem.outputs.version }}"
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,109 @@
...
+      PLUGIN_DIR: "packages/plugins/rubygem-${{ matrix.plugin }}"
+      SPEC_FILE: "packages/plugins/rubygem-${{ matrix.plugin }}/rubygem-${{ matrix.plugin }}.spec"
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get install -y rpm curl jq git-annex python3-pip
+        sudo gem install gem2rpm
+        sudo apt-get install -y libcurl4-openssl-dev
+        sudo gem install gem-compare
+        sudo pip3 install semver
```
ekohl, line 38: Why not
```suggestion
        sudo apt-get install -y python3-semver
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,109 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      plugins: ${{ steps.set-list.outputs.plugins }}
+    steps:
+    - name: Write the plugin list
+      run: echo '["foreman_ansible"]' > plugin_list.json
+    - name: Set the plugins list
+      id: set-list
+      run: |
+        PLUGINS=$(echo $(cat plugin_list.json))
```
ekohl, line 17: Why the echo? Isn't this sufficient?
```suggestion
        PLUGINS=$(cat plugin_list.json)
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,109 @@
...
+        echo "::set-output name=plugins::$PLUGINS"
+
+  # Bumpt the plugins on RPM
```
ekohl, line 20:
```suggestion
  # Bump the plugins on RPM
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,109 @@
...
+    - name: Try to bump plugin
+      run: ./bump_rpm.sh $PLUGIN_DIR -a "$CHANGE_AUTHOR" --no-commit
+      env:
+        CHANGE_AUTHOR: Github Actions <noreply@github.com>
+    - name: Open a PR
+      uses: peter-evans/create-pull-request@v3
+      with:
+        commit-message: "Updates ${{ matrix.plugin }} to ${{ steps.latest_gem.outputs.version }}"
+        branch: "bump_rpm/${{ matrix.plugin }}"
+        title: "Updates ${{ matrix.plugin }} to ${{ steps.latest_gem.outputs.version }}"
```
ekohl, line 64:
```suggestion
        title: "Update ${{ matrix.plugin }} to ${{ steps.latest_gem.outputs.version }}"
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,109 @@
...
+        ./scripts/update_package.rb --name ruby-$PACKAGE_NAME --version $NEW_VERSION
+      env:
+        NEW_VERSION: ${{ steps.latest_gem.outputs.version }}
+        PACKAGE_NAME: ${{ steps.package_info.outputs.name }}
+        AUTHOR_NAME: ${{ github.actor }}
+        AUTHOR_MAIL: ${{ github.actor }}@users.noreply.github.com
+    - name: Open a PR
+      uses: peter-evans/create-pull-request@v3
+      with:
+        commit-message: "Updates ${{ matrix.plugin }} to ${{ steps.latest_gem.outputs.version }}"
```
ekohl, line 106:
```suggestion
        commit-message: "Update ${{ matrix.plugin }} to ${{ steps.latest_gem.outputs.version }}"
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,109 @@
...
+        branch: "bump_deb/develop-${{ matrix.plugin }}"
+        title: "Updates ${{ matrix.plugin }} to ${{ steps.latest_gem.outputs.version }}"
```
ekohl, line 108:
```suggestion
        title: "Update ${{ matrix.plugin }} to ${{ steps.latest_gem.outputs.version }}"
```

ekohl (changes requested):
I appreciate the effort. I have been thinking about this for a long time as well but never came up with something that would work well. I've thought about implementing it as part of a publish of a gem (using GitHub actions) but never really like that either. Implementing it in Jenkins also felt painful due to credentials.

My biggest concern here is that we can easily blow up the already large number of branches in this repository. That also affects everyone who clones it. I wonder if we should a fork somewhere and send PRs from that.

ehelms:
Could any of this logic be either encapsulated into a script or into a container that is ran inside the GH action environment? Debugging CI jobs is hard, and one philosophy that has proven to be effective is when we encapsulate what is happening into tooling that can be run in any environment and we treat the CI system, as much as possible, as a simple runner of well-defined actions.

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,107 @@
...
+
+jobs:
+  # ugly way to share the plugin list
+  plugin_list:
+    runs-on: ubuntu-latest
+    outputs:
+      plugins: ${{ steps.set-list.outputs.plugins }}
+    steps:
+    - name: Write the plugin list
+      run: echo '["foreman_ansible"]' > plugin_list.json
```
ekohl, line 13: How will you deal with plugins which are only packaged on RPMs or (less common) only on deb? I think it may be needed to write two workflows: one for RPMs and one for deb.

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,107 @@
...
+
+    env:
+      PLUGIN_DIR: "packages/plugins/rubygem-${{ matrix.plugin }}"
+      SPEC_FILE: "packages/plugins/rubygem-${{ matrix.plugin }}/rubygem-${{ matrix.plugin }}.spec"
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get install -y rpm curl jq git-annex python3-semver
+        sudo gem install gem2rpm
+        sudo apt-get install -y libcurl4-openssl-dev
```
ekohl, line 38: Is this really still needed?

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,107 @@
...
+        sudo curl --create-dirs -o /usr/local/bin/spectool https://pagure.io/rpmdevtools/raw/26a8abc746fba9c0b32eb899b96c92841a37855a/f/spectool.in
+        sudo chmod +x /usr/local/bin/spectool
+        alias rpmdev-packager="echo $RPM_PACKER"
+    - name: Fetch latest plugin gem info
+      id: latest_gem
+      run: |
+        curl_output=$(curl -s https://rubygems.org/api/v1/gems/${GEM_NAME}.json)
+        echo "::set-output name=version::$(jq -r .version <<<$curl_output)"
+        echo "::set-output name=uri::$(jq -r .gem_uri <<<$curl_output)"
```
ekohl, line 46: I don't see this used anywhere. Could this whole block be simplified? (same with deb)

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,107 @@
...
+      env:
+        GEM_NAME: ${{ matrix.plugin }}
+    - name: Checkout RPM
+      uses: actions/checkout@v2
+      with:
+        ref: rpm/develop
+    - name: Initialize git annex
+      run: git annex init
+    - name: Try to bump plugin
+      run: ./bump_rpm.sh $PLUGIN_DIR
```
ekohl: This should still pass in the version to avoid mismatches where `bump_rpm.sh` would pick a different version than what's in the commit/pr message.

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,107 @@
...
+        curl_output=$(curl -s https://rubygems.org/api/v1/gems/${GEM_NAME}.json)
+        echo "::set-output name=version::$(jq -r .version <<<$curl_output)"
+        echo "::set-output name=uri::$(jq -r .gem_uri <<<$curl_output)"
+      env:
+        GEM_NAME: ${{ matrix.plugin }}
+    - name: Update package
+      run: ./scripts/update_package.rb --name ruby-$PACKAGE_NAME --version $NEW_VERSION
+      env:
+        NEW_VERSION: ${{ steps.latest_gem.outputs.version }}
+        PACKAGE_NAME: ${{ steps.package_info.outputs.name }}
```
ekohl, line 100: I'd make this the complete package name rather than just a part of it.
```suggestion
        PACKAGE_NAME: ruby-${{ steps.package_info.outputs.name }}
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,107 @@
...
+    needs: plugin_list
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        plugin:
+          - ${{ fromJson(needs.plugin_list.outputs.plugins) }}
+
+    env:
+      PLUGIN_DIR: "packages/plugins/rubygem-${{ matrix.plugin }}"
+      SPEC_FILE: "packages/plugins/rubygem-${{ matrix.plugin }}/rubygem-${{ matrix.plugin }}.spec"
```
ekohl, line 31: This is unused now
```suggestion
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,107 @@
...
+  bump_plugin_rpm:
+    needs: plugin_list
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        plugin:
+          - ${{ fromJson(needs.plugin_list.outputs.plugins) }}
+
+    env:
+      PLUGIN_DIR: "packages/plugins/rubygem-${{ matrix.plugin }}"
```
ekohl, line 30: This is only used once. Perhaps just inline it?

On the other hand, I can also imagine we end up with a script that collects directories to update and make that the test matrix. That seems likely given that the list of plugins depends on what we package for.

.github/workflows/bump_plugins.yml
ezr-ondrej: Because there is an old version in the repositories, that doesn't have `parse_version_info` (which reading the docs is also deprecated in the newest)

ekohl: Sounds like it's running on your `rpm/develop` and that is not up to date.

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,103 @@
...
+        echo "::set-output name=uri::$(jq -r .gem_uri <<<$curl_output)"
+    - name: Checkout RPM
+      uses: actions/checkout@v2
+      with:
+        ref: rpm/develop
+    - name: Initialize git annex
+      run: git annex init
+    - name: Try to bump plugin
+      run: |
+        alias rpmdev-packager="echo $RPM_PACKER"
```
ekohl, line 55: Can we use a global env var for author and email so RPM and deb both have the same author?

.github/workflows/bump_plugins.yml
ezr-ondrej: the plugin list gets feeded into two jobs, there are adjustments possible for each of them.

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,103 @@
...
+    needs: plugin_list
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        plugin:
+          - ${{ fromJson(needs.plugin_list.outputs.plugins) }}
+    steps:
+    - name: Set plugin package name
+      id: package_info
+      run: echo "::set-output name=name::$(echo ${{ matrix.plugin }} | tr _ -)"
```
ekohl, line 80: Looking at it further, I think this is the right place for the real package name.
```suggestion
      run: echo "::set-output name=name::ruby-$(echo ${{ matrix.plugin }} | tr _ -)"
```

.github/workflows/bump_plugins.yml
```diff
@@ -0,0 +1,103 @@
...
+        body: ''
+        delete-branch: true
+
+  # Bump the plugins on DEB
+  bump_plugin_deb:
+    needs: plugin_list
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        plugin:
```
ekohl, line 75: Naming wise: should this be `gem`?
```suggestion
        gem:
```

.github/workflows/bump_plugins.yml
ekohl: My point is that the list of plugins packaged on one is different than the other. I don't think we'll see a completely matching list on both.

Another thing I didn't think about before: on RPMs you also have plugins in `packages/katello`. Just 3 now (since we don't bump Katello itself), but `rubygem-foreman_rh_cloud`, `rubygem-foreman_scc_manager` and `rubygem-foreman_virt_who_configure` do live there because they depend on Katello.

.github/workflows/bump_plugins.yml
ezr-ondrej: I've inlined it, if we use lets refactor once we would actually use it twice.

.github/workflows/bump_plugins.yml
ezr-ondrej: curl had failed without it, but let me try :)

.github/workflows/bump_plugins.yml
ezr-ondrej: the difference between DEB and RPM can be addressed easily, this would be just the common ones, and for each job respectively we would add additional entries.

The katello packages would be a bit of an issue, but it just adds a bit of complexity to determinig the directory, right?

ekohl: It could exit early if it isn't found, that would be one solution, but you'll probably find this issue quickly.

For Katello you can probably work around it by calling `bump_rpm.sh` with `packages/*/$package` and let shell globbing solve it.

ekohl:
I've opened https://github.com/ezr-ondrej/foreman-packaging/pull/5 to offload a lot of the logic into a script, see https://github.com/theforeman/foreman-packaging/pull/7412. This makes the workflow easier to read and allows reuse. If people like this I can also take a stab at writing an equivalent for Debian.

ezr-ondrej:
> I've opened [ezr-ondrej#5](https://github.com/ezr-ondrej/foreman-packaging/pull/5) to offload a lot of the logic into a script, see #7412. This makes the workflow easier to read and allows reuse. If people like this I can also take a stab at writing an equivalent for Debian.

I've merged that and thus this PR is IMHO ready. I'm still keeping this to be triggered manually. I'd first try to run it few times manually and only then scheduled it, if others agree.

.github/workflows/bump_packages.yml
```diff
@@ -0,0 +1,95 @@
+name: Update packages
+on:
+  workflow_dispatch:
+env:
+  GIT_AUTHOR_NAME: ${{ github.actor }}
+  GIT_AUTHOR_EMAIL: ${{ github.actor }}@users.noreply.github.com
```
ekohl, line 6: I still think we should consider using static values here
```suggestion
  GIT_AUTHOR_NAME: Foreman Packaging Automation
  GIT_AUTHOR_EMAIL: packaging@theforeman.org
```

I'm not sure that email address goes anywhere, but it feels cleaner to indicate it's really automated.

ekohl (approved):
This looks good to me, but I wrote a significant part of it so I'd appreciate another person looking at this. @evgeni?

.github/workflows/bump_packages.yml
```diff
@@ -0,0 +1,95 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-list.outputs.matrix }}
+    steps:
+    - name: Checkout RPM
+      uses: actions/checkout@v2
+      with:
+        ref: rpm/develop
+    - name: Set the list
+      id: set-list
```
evgeni, line 20:
```suggestion
      id: set_list
```

I always get goose bumps when I see dashes in things that equally could be computations ;)

.github/workflows/bump_packages.yml
```diff
@@ -0,0 +1,95 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-list.outputs.matrix }}
+    steps:
+    - name: Checkout deb
+      uses: actions/checkout@v2
+      with:
+        ref: deb/develop
+    - name: Set the list
+      id: set-list
```
evgeni, line 70: same story as above ;)

evgeni (approved):

ezr-ondrej:
I've fixed your concerns @evgeni :)
Can we try this out? :) I'll be happy to fix any issues we encounter while running it manually :)

ekohl (approved):

ekohl:
I do wonder if we could squash the commits a bit.

evgeni (approved):
I'd squash on merge, unless @ezr-ondrej wants a more detailed history (I don't think it's needed)

ezr-ondrej:
Please squash :) I do not feel strong about the history of this, I just wanted ease up reviews :)

PR #7370, opened 2021-12-01
Allows to skip commiting 

ezr-ondrej:
bump_rpm can opt out from commiting and can get author for changelog as argument
add_changelog can have author of the change as argument

### DIFF
--- a/bump_rpm.sh
+++ b/bump_rpm.sh
@@ -104,7 +104,9 @@ if [[ $CURRENT_VERSION != $NEW_VERSION ]] ; then
 		rm -rf "$UNPACKED_GEM_DIR"
 	fi
 
-	git commit -m "Update $PACKAGE_NAME to $NEW_VERSION"
+	if [[ "$SKIP_GIT_COMMIT" != "1" ]] ; then
+		git commit -m "Update $PACKAGE_NAME to $NEW_VERSION"
+	fi
 else
 	echo "${PACKAGE_NAME}: $CURRENT_VERSION == $NEW_VERSION ; skipping"
 fi


add_changelog.sh
```diff
@@ -1,18 +1,32 @@
...
+    shift
+done
+set -- "${POSITIONAL[@]}" # restore positional parameters
+
 SPEC_FILE=$1
 VERSION=$2
 
 if [[ -z $SPEC_FILE ]] || [[ ! -e $SPEC_FILE ]] ; then
-	echo "Usage: $0 SPEC_FILE [VERSION]"
+	echo "Usage: $0 SPEC_FILE VERSION [-a AUTHOR]"
```
ekohl, line 17: This isn't correct. `VERSION` is optional and this help suggests it's required.

ekohl (changes requested):
Why go through all the trouble when you can set `RPM_PACKAGER` as an environment variable for `add_changelog.sh`? I'd prefer to use that instead.

ezr-ondrej:
Do you mean getting the information from environmental variable? We can not set it and leave it up to `rpmdev-packer` to pick it up as that tool is not present on ubuntu, on which we run the GH action (unless we download it as we download `spectool` what I like even less than adding arugment honestly)

ekohl:
> Do you mean getting the information from environmental variable? We can not set it and leave it up to `rpmdev-packer` to pick it up as that tool is not present on ubuntu, on which we run the GH action (unless we download it as we download `spectool` what I like even less than adding arugment honestly)

Then I'd recommend to use a container to just run the right tools. See [foreman-selinux](https://github.com/theforeman/foreman-selinux/blob/develop/.github/workflows/compile.yaml) for an example. Perhaps you can also use it just for a single step.

ezr-ondrej:
> Then I'd recommend to use a container to just run the right tools. See [foreman-selinux](https://github.com/theforeman/foreman-selinux/blob/develop/.github/workflows/compile.yaml) for an example. Perhaps you can also use it just for a single step.

Yeah I didn't want to take that route, but sure, container is better then hacks. let's do that.

bump_rpm.sh
```diff
@@ -8,8 +8,22 @@
...
+        *) POSITIONAL+=("$1") ;;
+    esac
+    shift
+done
+set -- "${POSITIONAL[@]}" # restore positional parameters
+
 if [[ -z $1 ]] ; then
 	echo "Usage: $0 directory [version]"
+	echo " -n|--no-commit dont commit"
+	echo " -a|--author Name <mail> author of the change for changelog"
```
ekohl, line 26: The author is no longer an argument.

bump_rpm.sh
```diff
@@ -8,8 +8,22 @@
 # rpmdevtools
 # rubygem-gem2rpm
 
+NO_COMMIT=false
+POSITIONAL=()
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        -n|--no-commit) NO_COMMIT=true ;;
+        -a|--author) RPM_PACKAGER="$2"; shift ;;
```
ekohl, line 16: The author is no longer an argument.

ekohl (changes requested):
If I'm honest, I'm :-1: on adding this. This introduces yet another convention and a lot of complexity in the script.

bump_rpm.sh
```diff
@@ -104,7 +104,7 @@ if [[ $CURRENT_VERSION != $NEW_VERSION ]] ; then
 		rm -rf "$UNPACKED_GEM_DIR"
 	fi
 
-	git commit -m "Update $PACKAGE_NAME to $NEW_VERSION"
+	[[ "$SKIP_GIT_COMMIT" == "1" ]] || git commit -m "Update $PACKAGE_NAME to $NEW_VERSION"
```
ekohl, line 107: I'd prefer an `if` statement

ezr-ondrej: That would allow any non empty value, right? don't we want to force it to be `1`?
I've used test != 1, but if you feel non-empty value is more readable, I'll use yours :shrug:

ekohl: Well, I'm fine with either check. I would like to avoid `[[ ]] || command` since I think it's a bad pattern.

ekohl (approved):

PR #7292, opened 2021-11-16
Update foreman_remote_execution to 5.0.1

adamruzicka:
CPs:
- 3.1

PR #7375, opened 2021-12-03
REX 5.0.1 with node stuff

evgeni:


evgeni:
Whoever will be reviewing this.

This PR is an "alternative" to #7292 which builds, even tho the way we did it in #7292 should also work.

I'll leave it to @adamruzicka to say if the build this PR produced actually works (I believe it does) and whether we should take this shortcut while we figure out WTH is wrong with the "real" way to build things.

Odilhao (approved):

Odilhao (approved):

adamruzicka:
3.1 CP https://github.com/theforeman/foreman-packaging/pull/7592

PR #7386, opened 2021-12-07
Update rubygem-foreman_ansible to 7.0.1

ezr-ondrej:
Should get into 3.1

RPM: #7387

PR #7387, opened 2021-12-07
Update rubygem-foreman_ansible to 7.0.0

ezr-ondrej:
Should get into 3.1

DEB: #7386

### DIFF
--- a/packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
+++ b/packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
@@ -8,7 +8,7 @@
 %global foreman_min_version 2.2.0
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 6.4.1
+Version: 7.0.1
 Release: 1%{?foremandist}%{?dist}
 Summary: Ansible integration with Foreman (theforeman.org)
 Group: Applications/Systems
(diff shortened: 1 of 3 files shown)


packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -46,6 +44,8 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
 # start package.json devDependencies BuildRequires
 BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
 BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
+BuildRequires: %{?scl_prefix}npm(@testing-library/user-event) >= 13.2.1
+BuildRequires: %{?scl_prefix}npm(@testing-library/user-event) < 14.0.0
```
evgeni, line 48:
```
DEBUG util.py:600:  No matching package to install: 'npm(@testing-library/user-event) >= 13.2.1'
DEBUG util.py:600:  No matching package to install: 'npm(@testing-library/user-event) < 14.0.0'
```

is this something we actually need? or is this for testing only?

aka: should we add `@testing-library` to excluded prefixes in https://github.com/theforeman/foreman/blob/develop/package-exclude.json?

evgeni: looking at https://testing-library.com/, excluding all these packages seems sane?

ezr-ondrej: yeah, I think it make sense :+1: Will do

evgeni:
This is now failing on the graphql loader issue.

You can workaround it like we did in https://github.com/theforeman/foreman-packaging/pull/7291 until we properly fix https://projects.theforeman.org/issues/34086

evgeni:
```
ModuleNotFoundError: Module not found: Error: Can't resolve '@patternfly/react-icons/dist/js/icons/exclamation-circle-icon' in '/builddir/build/BUILDROOT/rubygem-foreman_ansible-7.0.0-1.fm3_2.el8.noarch/usr/share/gems/gems/foreman_ansible-7.0.0/webpack'
```

Hmmm.... I recall we had this issue in the past. Using `dist` is the wrong way here.

Need to dig foreman core git history for the fix.

evgeni:
https://github.com/theforeman/foreman_ansible/pull/492

ezr-ondrej:
Leapp fails, it is probably not an issue of this PR, right?

evgeni (approved):

ezr-ondrej:
CP 3.1: #7397

PR #7388, opened 2021-12-07
Fixes #34098 - correctly print proxy type in Puppet message

evgeni:

### DIFF
--- a/packages/katello/katello/helper.rb
+++ b/packages/katello/katello/helper.rb
@@ -24,6 +24,10 @@ module KatelloUtilities
       "#{scenarios_path}/last_scenario.yaml"
     end
 
+    def scenario_yaml(scenario)
+      File.join(scenarios_path, "#{scenario}.yaml")
+    end
+
     def accepted_scenarios
       @accepted_scenarios || ["katello", "foreman-proxy-content"]
     end
@@ -33,7 +37,9 @@ module KatelloUtilities
     end
 
     def load_scenario_data(scenario)
-      YAML.load_file(File.join(scenarios_path, "#{scenario}.yaml"))
+      if File.exist?(scenario_yaml(scenario))
+        YAML.load_file(scenario_yaml(scenario))
+      end
     end
 
     def load_scenario_answers(scenario)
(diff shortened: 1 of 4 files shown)


packages/katello/katello/hostname-change.rb
```diff
@@ -156,7 +156,7 @@ def next_steps_message
   #{@options[:program]}-installer --foreman-proxy-foreman-base-url https://#{@new_hostname} \\
                                   --foreman-proxy-trusted-hosts #{@new_hostname}
 
-  If Puppet is enabled on the #{proxy}, add the following to the installer command:
+  If Puppet is enabled on the #{@proxy}, add the following to the installer command:
```
evgeni, line 159: this is the real fix, the rest is "oh god, why did I look"

packages/katello/katello/helper.rb
```diff
@@ -32,8 +32,16 @@ def error_message
       "This utility can't run on a non-katello system."
     end
 
+    def scenario_yaml(scenario)
+      File.join(scenarios_path, "#{scenario}.yaml")
+    end
```
ekohl, line 37: Nit: Above are a bunch of paths. Should this be together with `last_scenario_yaml` (which should also use `File.join()` IMHO, but we're getting deep into to "why did I look" territory).?

packages/katello/katello/helper.rb
```diff
@@ -32,8 +32,16 @@ def error_message
...
+
     def load_scenario_data(scenario)
-      YAML.load_file(File.join(scenarios_path, "#{scenario}.yaml"))
+      if File.exist?(scenario_yaml(scenario))
+        YAML.load_file(scenario_yaml(scenario))
+      else
+        {}
```
ekohl, line 43: Looks like `load_scenario_answers` can also deal with `nil` so it can be
```suggestion
```

packages/katello/katello/katello.spec
```diff
@@ -132,6 +132,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Tue Dec 07 2021 Evgeni Golov - 4.4.0-0.2.master
+- Fixes #34098 - correctly print proxy type in Puppet message
```
ekohl, line 136: Should this be squashed into another commit?

ekohl (commented):
Thanks!

packages/katello/katello/helper.rb
evgeni: Can certainly move it up there.

ekohl (approved):

PR #7389, opened 2021-12-07
Fixes #34098 - correctly print proxy type in Puppet message

evgeni:


ekohl (approved):

evgeni:
Waiting to also get https://github.com/theforeman/foreman-packaging/pull/7390 in the same round, I think

evgeni:
> Waiting to also get #7390 in the same round, I think

let's not do this.

PR #7390, opened 2021-12-07
Fixes #34099 - only restart puppetserver if it is present

evgeni:

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -550,8 +550,10 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
         if success
           STDOUT.puts result
           STDOUT.puts 'Restarting puppet services'
-          run_cmd('/bin/systemctl try-restart puppet')
-          run_cmd('foreman-maintain service restart --only puppetserver')
+          puppet_services = ['puppet', 'puppetserver'].select do |service|
+            system("systemctl is-active #{service} --quiet")
+          end
+          run_cmd("/bin/systemctl try-restart #{puppet_services.join(' ')}") unless puppet_services.empty?
 
           STDOUT.puts "**** Hostname change complete! ****".green
           STDOUT.puts "IMPORTANT:"
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,7 +5,7 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 2
+%global release 3
 
 Name:       katello
 Version:    4.4.0
@@ -132,6 +132,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Tue Dec 07 2021 Evgeni Golov - 4.4.0-0.3.master
+- Fixes #34099 - only restart puppetserver if it is present
+
 * Tue Dec 07 2021 Evgeni Golov - 4.4.0-0.2.master
 - Fixes #34098 - correctly print proxy type in Puppet message
 


packages/katello/katello/hostname-change.rb
```diff
@@ -551,7 +551,9 @@ def run
           STDOUT.puts result
           STDOUT.puts 'Restarting puppet services'
           run_cmd('/bin/systemctl try-restart puppet')
-          run_cmd('foreman-maintain service restart --only puppetserver')
+          if File.exist?('/usr/lib/systemd/system/puppetserver.service')
```
evgeni, line 554: This felt cleaner than shelling out to `rpm -q puppetserver`, but still not ideal.

We also could parse answers, but…

ekohl: How about `systemctl is-active puppetserver --quiet` and checking the exit code? From the man page:
> Check whether any of the specified units are active (i.e. running). Returns an exit code 0 if at least one is active, or non-zero otherwise.

packages/katello/katello/hostname-change.rb
```diff
@@ -551,7 +551,9 @@ def run
...
+            run_cmd('foreman-maintain service restart --only puppetserver')
```
ekohl, line 555: Why do we shell out to `foreman-maintain`? My reasoning when reviewing it back then is that I thought it would not restart it if it wasn't enabled. If it doesn't provide this guarantee then this is really the same as:
```suggestion
            run_cmd('systemctl restart puppetserver')
```

And again, getting into the territory of "why am I looking at this": Shouldn't we really restart the server before the client?

evgeni: @upadhyeammit says this should exist 0 and carry on in this case, so probably this patch is not needed when we fix the issue in FM, nice

evgeni: so the FM fix is in https://projects.theforeman.org/issues/34100, but after applying this, FM does *not* exit 0 in the case where we ask it to restart a service that isn't there.

ekohl: Since there was some confusion, I explained what my original reasoning was. In https://github.com/theforeman/foreman-packaging/pull/4249 when I last looked at this I don't see comments, but I recall having offline conversations about why we need a wrapper at all. Also, back then `puppetserver` wasn't really easily optional.

> And again, getting into the territory of "why am I looking at this": Shouldn't we really restart the server before the client?

Looks like I am repeating myself: https://github.com/theforeman/foreman-packaging/pull/4249#discussion_r340711477

In short: If we do use `foreman-maintain`, then it should be powerful enough to not need any manual workarounds but I prefer to purely use `systemctl` and not `foreman-maintain` here.

packages/katello/katello/hostname-change.rb
evgeni: done

packages/katello/katello/hostname-change.rb
evgeni: flipped to systemctl

packages/katello/katello/hostname-change.rb
```diff
@@ -551,7 +551,10 @@ def run
           STDOUT.puts result
           STDOUT.puts 'Restarting puppet services'
           run_cmd('/bin/systemctl try-restart puppet')
-          run_cmd('foreman-maintain service restart --only puppetserver')
+          `systemctl is-active puppetserver --quiet`
+          if $?.success?
+            run_cmd('systemctl try-restart puppetserver')
+          end
```
ekohl, line 557: Thoughts on parallel restarts?
```ruby
services = ['puppet']
`systemctl is-active puppetserver --quiet`
services << 'puppetserver' if $?.success?
run_cmd('/bin/systemctl try-restart ' + ' '.join(services))

Comment (evgeni) [reply packages/katello/katello/hostname-change.rb:557]: No strong feelings either way.

@ehelms?

Comment (ehelms) [reply packages/katello/katello/hostname-change.rb:557]: Sounds fine to me for puppet services. Now let me ask this question. In a world where Puppet is disabled, like Katello's default currently, would `puppet` be enabled? i.e. does this code need to also do the same sort of check on `puppet` ?

Comment (evgeni) [reply packages/katello/katello/hostname-change.rb:557]: No. The puppet service is always there (as the installer is using Puppet Agent), but a "try-restart" will only restart it if it's also *running*.

Comment (evgeni) [reply packages/katello/katello/hostname-change.rb]: well, look at this

File: packages/katello/katello/hostname-change.rb [hunk 4]
```diff
@@ -550,8 +550,12 @@ def run
         if success
           STDOUT.puts result
           STDOUT.puts 'Restarting puppet services'
-          run_cmd('/bin/systemctl try-restart puppet')
-          run_cmd('foreman-maintain service restart --only puppetserver')
+          puppet_services = []
+          ['puppet', 'puppetserver'].each do |service|
+            `systemctl is-active #{service} --quiet`
+            puppet_services << service if $?.success?
+          end
```
ekohl, line 557: Nit: this is the same, but shorter:
```suggestion
          puppet_services = ['puppet', 'puppetserver'].select do |service|
            system("systemctl is-active #{service} --quiet")
          end
```

ekohl (approved):

packages/katello/katello/hostname-change.rb
evgeni: indeed.

packages/katello/katello/hostname-change.rb
```diff
@@ -550,8 +550,10 @@ def run
         if success
           STDOUT.puts result
           STDOUT.puts 'Restarting puppet services'
-          run_cmd('/bin/systemctl try-restart puppet')
-          run_cmd('foreman-maintain service restart --only puppetserver')
+          puppet_services = ['puppet', 'puppetserver'].select do |service|
+            system("systemctl is-active #{service} --quiet")
+          end
+          run_cmd("/bin/systemctl try-restart #{puppet_services.join(' ')}") unless puppet_services.empty?
```
ekohl, line 556: Of course I only see this post merge. Perhaps it should have been:
```ruby
if puppet_services.any?
  STDOUT.puts 'Restarting puppet services'
  run_cmd("/bin/systemctl try-restart #{puppet_services.join(' ')}")
end
```

evgeni: Probably. Can follow up later/Monday, unless you want to?

evgeni: https://github.com/theforeman/foreman-packaging/pull/7417

PR #7391, opened 2021-12-07
Update rubygem-foreman_remote_execution to 5.0.1

adamruzicka:
- Update rubygem-foreman_remote_execution to 5.0.0
- Add graphql related deps
- Update rubygem-foreman_remote_execution to 5.0.1


evgeni:
this breaks leapp (which depends on rex < 5), so I'd prefer if we could get a compatible foreman_leapp release and get both in together.

evgeni:
the leapp update is in https://github.com/theforeman/foreman-packaging/pull/7405 :fireworks:

evgeni (approved):

PR #7402, opened 2021-12-09
Update foreman to 3.1.0 GA

Odilhao:


evgeni (approved):

evgeni:
repoclosure is confused if we build non-scl and scl at the same time, good to go!

PR #7406, opened 2021-12-10
add debcheck script for repoclosure like analyzis

evgeni:

### DIFF
--- a/scripts/debcheck
+++ b/scripts/debcheck
@@ -1,0 +1,58 @@
+#!/bin/bash
+#
+# use dose-debcheck/dose-distcheck to do "repoclosure" for debian
+# Debian package dose-distcheck
+# Fedora package dose3-tools
+
+set -e
+set -o pipefail
+
+DEBIAN_RELEASE=${1:-bullseye}
+FOREMAN_VERSION=${2:-nightly}
+INCLUDE_PLUGINS=${3:-true}
+USE_STAGING=${4:-false}
+PUPPET_VERSION=${PUPPET_VERSION:-6}
+# user repos TBD
+
+case ${DEBIAN_RELEASE} in
+  focal)
+    DEBIAN_PACKAGES_SOURCE=http://archive.ubuntu.com/ubuntu/dists/${DEBIAN_RELEASE}/@DEBIAN_SUITE@/binary-amd64/Packages.xz
+    DEBIAN_SUITES="main universe"
+    ;;
+  *)
+    DEBIAN_PACKAGES_SOURCE=https://deb.debian.org/debian/dists/${DEBIAN_RELEASE}/@DEBIAN_SUITE@/binary-amd64/Packages.xz
+    DEBIAN_SUITES="main"
+    ;;
+esac
+
+PUPPET_PACKAGES_SOURCE=https://apt.puppet.com/dists/${DEBIAN_RELEASE}/puppet${PUPPET_VERSION}/binary-amd64/Packages
+
+if [[ ${USE_STAGING} == true ]]; then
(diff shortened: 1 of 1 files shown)


scripts/debcheck.sh
```diff
@@ -0,0 +1,38 @@
...
+  focal)
+    # this is broken, as ubuntu needs multiple archives to be enabled to work
+    DEBIAN_PACKAGES_SOURCE=http://archive.ubuntu.com/ubuntu/dists/${DEBIAN_RELEASE}/main/binary-amd64/Packages.xz
+    ;;
+  *)
+    DEBIAN_PACKAGES_SOURCE=https://deb.debian.org/debian/dists/${DEBIAN_RELEASE}/main/binary-amd64/Packages.xz
+    ;;
+esac
+
+PUPPET_PACKAGES_SOURCE=https://apt.puppet.com/dists/${DEBIAN_RELEASE}/puppet6/binary-amd64/Packages
```
ekohl, line 21: Should we default to Puppet 7?

ekohl (approved):

scripts/debcheck.sh
```diff
@@ -0,0 +1,38 @@
...
+FOREMAN_PACKAGES_SOURCE=https://deb.theforeman.org/dists/${DEBIAN_RELEASE}/${FOREMAN_VERSION}/binary-amd64/Packages
+PLUGINS_PACKAGES_SOURCE=https://deb.theforeman.org/dists/plugins/${FOREMAN_VERSION}/binary-amd64/Packages
+
+WORKDIR=$(mktemp -d)
+PACKAGES_DEBIAN=${WORKDIR}/Packages_debian
+PACKAGES_PUPPET=${WORKDIR}/Packages_puppet
+PACKAGES_FOREMAN=${WORKDIR}/Packages_foreman
+PACKAGES_PLUGINS=${WORKDIR}/Packages_plugins
```
ekohl, line 29: Should plugins be optional so you can also check core? Any plans to integrate this into CI pipelines?

scripts/debcheck.sh
evgeni: technically, we still support 6, so testing it as the lower boundary is more correct?

scripts/debcheck.sh
evgeni: slow down, will ya? :)

yes, and yes, and I still need to figure out how to properly do this for ubuntu

scripts/debcheck.sh
ekohl: I guess. I also considered a an env var for just the version that allows you to easily test with both. Maybe it's overkill for now.

evgeni: that's actually a good idea, done

scripts/debcheck.sh
evgeni: 1. done
2. will be somewhere else ;)

ekohl (approved):
Looks like a good start. I was thinking about reducing some duplication by using more arrays but declarative array may just complicate things. We can always refactor it later on.

The only thing I'd question is if we need an extension. We have a shebang and if calling this script becomes the "interface", dropping the extension would allow rewriting it in a different language.

scripts/debcheck.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
ekohl, line 1: Should it use `set -e` to fail early?
```suggestion
#!/bin/bash

set -e
```

evgeni: and also pipefail, yes!

evgeni:
> The only thing I'd question is if we need an extension. We have a shebang and if calling this script becomes the "interface", dropping the extension would allow rewriting it in a different language.

Stop using my arguments against me!

PR #7412, opened 2021-12-10
Add a script to list updatable packages

ekohl:
This script lists all directories which can be updated using bump_rpm.sh and done so safely in an automated manner.

If GitHub actions are detected it also outputs the data in a way that integrates. Then the output can be registered and used in other steps.

The goal is to provide an integration point for https://github.com/theforeman/foreman-packaging/pull/7369.

I've debated only outputting gem names or returning a more complicated structure. For now I settled on a simple first version and have the discussion.

Something else I also considered is already checking which versions can be updated in this loop. Another discussion point.

### DIFF
--- a/list_updatable_packages
+++ b/list_updatable_packages
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+
+import json
+import os
+import subprocess
+import sys
+from pathlib import Path
+from typing import Generator, Iterable, Mapping, Optional
+
+import requests
+
+SUPPORTED_TEMPLATES = ('foreman_plugin', 'hammer_plugin', 'smart_proxy_plugin')
+SESSION = requests.Session()
+
+
+class Spec:
+    def __init__(self, path: Path):
+        self.path = path
+        spec = path.read_text()
+        self.lines = spec.splitlines()
+        self._globals = None
+
+    @property
+    def directory(self) -> str:
+        return self.path.parent.as_posix()
+
+    @property
+    def package_name(self) -> str:
+        return self.path.name[:-len('.spec')]
+
+    @property
+    def template(self) -> Optional[str]:
+        for line in self.lines:
+            if line.startswith('# template:'):
+                _, _, template = line.split(None, 3)
+                return template
+
+        return None
+
+    @property
+    def globals(self) -> Mapping[str, str]:
+        """
+        Perform very naive spec parsing to extract the globals
+        """
+        if self._globals is None:
+            self._globals = {}
+            for line in self.lines:
+                if line.startswith('%global'):
+                    definition, name, value = line.split(None, 3)
+                    if definition == '%global':
+                        self._globals[name] = value
+
+        return self._globals
+
+    @property
+    def is_nightly(self) -> bool:
+        return 'prereleasesource' in self.globals
+
+    @property
+    def gem_name(self) -> str:
+        return self.globals.get('gem_name')
+
+    @property
+    def version(self) -> str:
+        cmd = ['rpmspec', '--srpm', '--query', '--queryformat=%{version}', self.path.as_posix()]
+        return subprocess.check_output(cmd, stderr=subprocess.DEVNULL, universal_newlines=True)
+
+
+def find_specs() -> Generator[Spec, None, None]:
+    for path in Path('packages').glob('*/*/*.spec'):
+        spec = Spec(path)
+        if spec.template in SUPPORTED_TEMPLATES and not spec.is_nightly:
+            yield spec
+
+
+def latest_version(spec: Spec) -> str:
+    if spec.gem_name:
+        url = f'https://rubygems.org/api/v1/versions/{spec.gem_name}/latest.json'
+        response = SESSION.get(url)
+        response.raise_for_status()
+        return response.json()['version']
+    raise ValueError('Unable to determine latest version', spec)
+
+
+def build_matrix(specs: Iterable[Spec]) -> Generator[dict, None, None]:
+    for spec in specs:
+        try:
+            current_version = spec.version
+        except subprocess.CalledProcessError as e:  # pylint: disable=invalid-name
+            print('Failed to determine version for', spec.path.as_posix(), str(e), file=sys.stderr)
+        else:
+            new_version = latest_version(spec)
+            if current_version != new_version:
+                entry = {
+                    'directory': spec.directory,
+                    'package_name': spec.package_name,
+                    'gem_name': spec.gem_name,
+                    'current_version': current_version,
+                    'new_version': new_version,
+                }
+                yield entry
+
+
+def main() -> None:
+    matrix = list(build_matrix(find_specs()))
+
+    if 'GITHUB_ACTION' in os.environ:
+        directories = [entry['directory'] for entry in matrix]
+        print(f'::set-output name=directories::{json.dumps(directories)}')
+        print(f'::set-output name=matrix::{json.dumps(matrix)}')
+    for entry in matrix:
+        print(entry['directory'], entry['new_version'])
+
+
+if __name__ == '__main__':
+    main()


list_updatable_packages
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+DIRECTORIES=( $(grep --files-with-matches '^# template: foreman_plugin' packages/*/*/*.spec | xargs --max-args=1 dirname) )
```
ekohl, line 3: For now this is just `foreman_plugin` as a safe start but there's little reason to stick to it. We could easily also add others:
```suggestion
DIRECTORIES=( $(grep --files-with-matches '^# template: (foreman|smart_proxy|hammer)_plugin' packages/*/*/*.spec | xargs --max-args=1 dirname) )
```

ekohl (commented):
@ezr-ondrej you may not be watching all PRs so here's an explicit ping. Could you have a look?

evgeni:
What's the benefit of doing this in a rather manual way, vs getting the info from `package_manifest.yaml`?

ekohl:
I don't want to parse YAML. That would also have meant I probably would need to write it in Python and import Ansible. This feels much easier and much more flexible to me. Now I didn't need half an hour to look up how an Ansible inventory is supposed to work again.

For example, a naive implementation may find  `packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec` which doesn't have a template so it can't be updated automatically (the dependencies would be out of date). It may also catch `packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec`. By looking at the `template: ` line you get exactly what really works.

ekohl:
This has now been rewritten entirely into a Python script. It can now do a lot of things `bump_rpm.sh` can do. It's very tempting to include that functionality too but I'm resisting.

Running `rpmspec` does make the script slow (~10 seconds on my system for just `foreman_plugin` packages), but it saves running an entire matrix where every run takes ~30 seconds.

ekohl:
https://github.com/ezr-ondrej/foreman-packaging/pull/5 makes use of this new script.

ezr-ondrej:
I'm sure there is plenty of ways to do this, but this is roughly what I would imagine to have and script in the rpm/deb to get the packages is what I quite like as you want to have the repo anyway.

The only thing I'm not sure is if we want to keep the script in the root dir.
Otherwise I would love to get a working version (which I think this is) in and we can later on improve it if it misbehave.

So :+1:  on this it just works and we will still review the PRs until we are sure, so I'd iterate here, the PR in my repo can't be merged without this in.

ezr-ondrej:
Can we push this forward? :)

bump_rpm.sh
```diff
@@ -55,6 +55,15 @@ else
...
 fi
 
+if [[ -n $GITHUB_ACTION ]] ; then
+	if [[ -n $GEM_NAME ]] ; then
+		echo "::set-output name=gem_name::$GEM_NAME"
+	fi
+	echo "::set-output name=package_name::$PACKAGE_NAME"
+	echo "::set-output name=current_version::$CURRENT_VERSION"
+	echo "::set-output name=new_version::$NEW_VERSION"
+fi
```
ekohl, line 65: In the end I didn't need this. Should I remove this?

list_updatable_packages
```diff
@@ -0,0 +1,116 @@
...
+    def globals(self) -> Mapping[str, str]:
+        """
+        Perform very naive spec parsing to extract the globals
+        """
+        if self._globals is None:
+            self._globals = {}
+            for line in self.lines:
+                if line.startswith('%global'):
+                    definition, name, value = line.split(None, 3)
+                    if definition == '%global':
```
ekohl, line 50: I couldn't find a good way for this with just one comparison (besides using regexes).

list_updatable_packages
```diff
@@ -0,0 +1,116 @@
...
+        return 'prereleasesource' in self.globals
+
+    @property
+    def gem_name(self) -> str:
+        return self.globals.get('gem_name')
+
+    @property
+    def version(self) -> str:
+        command = ['rpmspec', '--srpm', '-q', '--queryformat=%{version}', self.path.as_posix()]
+        return subprocess.check_output(command, stderr=subprocess.DEVNULL, universal_newlines=True)
```
ekohl, line 66: This part is slow but I'm not sure if there's a faster way (that's still easily maintainable). AFAIK rpm isn't really a nice library to use though it would probably be faster.

list_updatable_packages
```diff
@@ -0,0 +1,116 @@
...
+    def is_nightly(self) -> bool:
+        return 'prereleasesource' in self.globals
+
+    @property
+    def gem_name(self) -> str:
+        return self.globals.get('gem_name')
+
+    @property
+    def version(self) -> str:
+        command = ['rpmspec', '--srpm', '-q', '--queryformat=%{version}', self.path.as_posix()]
```
evgeni, line 65:
```suggestion
        command = ['rpmspec', '--srpm', '--query', '--queryformat=%{version}', self.path.as_posix()]
```

list_updatable_packages
evgeni: `rpmspec`: `./list_updatable_packages  2.24s user 0.48s system 8% cpu 32.325 total`
`import rpm`: `./list_updatable_packages  1.93s user 0.08s system 9% cpu 20.501 total`

I wouldn't bother too much ;)

evgeni (approved):
small nit, otherwise I'm good with this!

list_updatable_packages
evgeni:
```diff
diff --git list_updatable_packages list_updatable_packages
index fd637781a..33ef64fb0 100755
--- list_updatable_packages
+++ list_updatable_packages
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Generator, Iterable, Mapping, Optional
 
 import requests
+import rpm
 
 SUPPORTED_TEMPLATES = ('foreman_plugin', 'hammer_plugin', 'smart_proxy_plugin')
 SESSION = requests.Session()
@@ -62,8 +63,9 @@ class Spec:
 
     @property
     def version(self) -> str:
-        command = ['rpmspec', '--srpm', '-q', '--queryformat=%{version}', self.path.as_posix()]
-        return subprocess.check_output(command, stderr=subprocess.DEVNULL, universal_newlines=True)
+        ts=rpm.TransactionSet()
+        s = ts.parseSpec(self.path.as_posix())
+        return s.sourceHeader.format('%{version}')
 
 
 def find_specs() -> Generator[Spec, None, None]:
```

if you're trully interested.

ekohl: You'd only take the `import rpm` overhead once instead of once per package. I'll see about the entire runtime.

evgeni: `rpmspec` is not Python, it's C.

Toying with the Debian side of things, the slow part of this all is talking to `rubygems.org`, not parsing specs.

bump_rpm.sh
ezr-ondrej: weeeell I guess it might be usefull, but we get the same info in the matrix, right?
Tho I believe that if you have an output to share, you should and this might be interesting in the future :shrug:

ekohl:
(trivial rebase, no actual changes)

list_updatable_packages
ekohl: with calling rpmspec:
```console
$ time ./list_updatable_packages 
packages/plugins/rubygem-smart_proxy_reports 0.0.7
packages/plugins/rubygem-smart_proxy_salt 4.0.0
packages/plugins/rubygem-smart_proxy_dhcp_dnsmasq 1.0
packages/plugins/rubygem-foreman_salt 14.1.0
packages/plugins/rubygem-smart_proxy_dhcp_bluecat 0.1.6

real	0m29.179s
user	0m3.772s
sys	0m0.964s
```

With using RPM as a library:
```console
$ time ./list_updatable_packages 
warning: line 21: It's not recommended to have unversioned Obsoletes: Obsoletes: rubygem-redhat_access
warning: line 22: It's not recommended to have unversioned Obsoletes: Obsoletes: rubygem-redhat_access_lib
warning: line 24: It's not recommended to have unversioned Obsoletes: Obsoletes: rubygem-foreman_inventory_upload
warning: line 25: It's not recommended to have unversioned Obsoletes: Obsoletes: rubygem-foreman_inventory_upload-doc
warning: line 21: It's not recommended to have unversioned Obsoletes: Obsoletes: rubygem-redhat_access
warning: line 22: It's not recommended to have unversioned Obsoletes: Obsoletes: rubygem-redhat_access_lib
warning: line 24: It's not recommended to have unversioned Obsoletes: Obsoletes: rubygem-foreman_inventory_upload
warning: line 25: It's not recommended to have unversioned Obsoletes: Obsoletes: rubygem-foreman_inventory_upload-doc
packages/plugins/rubygem-smart_proxy_reports 0.0.7
packages/plugins/rubygem-smart_proxy_salt 4.0.0
packages/plugins/rubygem-smart_proxy_dhcp_dnsmasq 1.0
packages/plugins/rubygem-foreman_salt 14.1.0
packages/plugins/rubygem-smart_proxy_dhcp_bluecat 0.1.6

real	0m24.668s
user	0m3.658s
sys	0m0.175s
```

So about 15% faster to use it as a library, but I'd need to look at silencing the warnings. I'd be in favor of keeping it as it is now and only look at it if we really need to.

evgeni: Same. Please merge ;)

ekohl:
Updated:
* Removed changes in `bump_rpm.sh`
* changed `-q` to `--query` (and `command` to `cmd` to keep the line below 100 chars)
* Added pylint ignore for invalid-name

PR #7418, opened 2021-12-13
drop foreman-tasks-core and smart_proxy_dynflow_core

evgeni:


ekohl (approved):
Do we need anything that obsoletes these so they're removed?

evgeni:
We don't have anything right now (and doing this from the non-core packages is wrong), so let's keep it as is for now.

ekohl:
Should we add it as an upgrade warning in the release notes?

PR #7419, opened 2021-12-13
drop smart_proxy_salt_core

evgeni:


evgeni:
cc @bastian-src @sbernhard

feel free to revert the `smart_proxy_salt` commit once there is a version that works without `*_core` packages, until then this blocks packaging automation/cleanup.

sbernhard:
Thanks @evgeni. This would not be necessary if we solve https://github.com/theforeman/foreman-packaging/pull/6968 (open since Aug 2)

evgeni:
Sure, please solve it ;-)

(but it then still would require cloning the work to the Debian side of things)

sbernhard:
Lets try this one: https://github.com/theforeman/foreman-packaging/pull/7422

evgeni:
cool, dropped the "drop salt" commit, we can continue in #7429 on that, droping salt-core now, as that's really "gone"

PR #7422, opened 2021-12-13
Update foreman_salt to 14.1.0

sbernhard:

### DIFF
--- a/plugins/ruby-foreman-salt/debian/changelog
+++ b/plugins/ruby-foreman-salt/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-salt (14.1.0-1) stable; urgency=low
+
+  * 14.1.0 released
+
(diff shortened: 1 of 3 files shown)


evgeni:
Would you mind splitting this up in three PRs, one for each: hammer, proxy, foreman? Thanks!

Looking at https://github.com/theforeman/foreman-packaging/pull/6967 the foreman plugin seems (or maybe not) to depend on a change in foreman, which hasn't been merged yet?!

Looking at https://github.com/theforeman/foreman-packaging/pull/6968 the proxy plugin gained new files where we're not quite sure where to install them to, but this PR doesn't touch these at all. Intentionally?

evgeni:
Thanks for splitting!

The question whether this will work without https://github.com/theforeman/foreman/pull/8489 being merged is still open.

evgeni:
My understanding from IRC is: this should work without, merging.

evgeni (approved):

PR #7426, opened 2021-12-14
Update ruby-foreman-tasks to 5.3.0

adamruzicka:


plugins/ruby-foreman-tasks/debian/gem.list
```diff
@@ -1,5 +1,5 @@
 # list of gem urls to download via Jenkins, space separated
-GEMS="https://rubygems.org/downloads/foreman-tasks-5.2.0.gem"
+GEMS="https://rubygems.org/downloads/foreman-tasks-5.3.0.gem"
 GEMS="$GEMS https://rubygems.org/downloads/foreman-tasks-core-0.4.0.gem"
```
evgeni, line 3: I think this can be dropped now?

evgeni (approved):

plugins/ruby-foreman-tasks/debian/gem.list
adamruzicka: salt core depends on it. Can I drop it before salt core goes away?

evgeni: to hell with salt core, it's not installable right now anyways (it depends on rex-core, which is gone)

PR #7429, opened 2021-12-14
Bump smart_proxy_salt

sbernhard:


plugins/smart_proxy_salt/control
```diff
@@ -10,6 +10,6 @@ XS-Ruby-Versions: all
 Package: ruby-smart-proxy-salt
 Architecture: all
 XB-Ruby-Versions: ${ruby:Versions}
-Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter, foreman-proxy (>= 1.8.0~rc1), salt-master, python, ruby-smart-proxy-salt-core (>= 0.0.2)
+Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter, foreman-proxy (>= 1.8.0~rc1), salt-master, python
```
evgeni, line 13: in https://github.com/theforeman/foreman-packaging/pull/6968 the proxy dependency was bumped to 2.5, not sure if intentionally, but would be nice to keep DEB and RPM in sync (you decide which version is correct, I have no idea)

evgeni: also we should add a dependency on `ruby-smart-proxy-dynflow`, as declared in https://github.com/theforeman/smart_proxy_salt/commit/970d39d081ef40c8a3a9960ce18904621e4d89fc

evgeni (approved):

PR #7434, opened 2021-12-17
Add a script to list updatable packages

evgeni:

### DIFF
--- a/scripts/list_updatable_packages
+++ b/scripts/list_updatable_packages
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+import json
+import os
+from collections import namedtuple
+from pathlib import Path
+from typing import Generator, Iterable
+
+from debian.changelog import Changelog  # type: ignore
+from debian.debian_support import NativeVersion  # type: ignore
+
+import requests
+
+SESSION = requests.Session()
+
+
+Package = namedtuple('Package', ['package_name', 'version', 'gem_name', 'directory'])
+
+
+def find_packages() -> Generator[Package, None, None]:
+    for plugin_rb in Path('plugins/').glob('**/*.rb'):
+        gem_name = plugin_rb.read_text().strip().split()[1].strip("',\"")
+        package_changelog = next(plugin_rb.parent.glob('**/changelog'))
+        with package_changelog.open() as debchangelog:
+            changelog = Changelog(debchangelog)
+            version = NativeVersion(changelog.version).upstream_version
+            yield Package(str(changelog.package), version, gem_name, plugin_rb.parent.as_posix())
+
+
+def latest_version(spec: Package) -> str:
+    if spec.gem_name:
+        url = f'https://rubygems.org/api/v1/versions/{spec.gem_name}/latest.json'
+        response = SESSION.get(url)
+        response.raise_for_status()
+        return response.json()['version']
+    raise ValueError('Unable to determine latest version', spec)
+
+
+def build_matrix(specs: Iterable[Package]) -> Generator[dict, None, None]:
+    for spec in specs:
+        current_version = spec.version
+        new_version = latest_version(spec)
+        if current_version != new_version:
+            entry = {
+                'directory': spec.directory,
+                'package_name': spec.package_name,
+                'gem_name': spec.gem_name,
+                'current_version': current_version,
+                'new_version': new_version,
+            }
+            yield entry
+
+
+def main() -> None:
+    matrix = list(build_matrix(find_packages()))
+
+    if 'GITHUB_ACTION' in os.environ:
+        directories = [entry['directory'] for entry in matrix]
+        print(f'::set-output name=directories::{json.dumps(directories)}')
+        print(f'::set-output name=matrix::{json.dumps(matrix)}')
+    for entry in matrix:
+        print(entry['directory'], entry['new_version'])
+
+
+if __name__ == '__main__':
+    main()


evgeni:
@ekohl you might say this is inspired ;)

scripts/list_updatable_packages
```diff
@@ -0,0 +1,67 @@
...
+
+def main() -> None:
+    matrix = list(build_matrix(find_packages()))
+
+    if 'GITHUB_ACTION' in os.environ:
+        directories = [entry['directory'] for entry in matrix]
+        print(f'::set-output name=directories::{json.dumps(directories)}')
+        print(f'::set-output name=matrix::{json.dumps(matrix)}')
+    for entry in matrix:
+        print(entry['directory'], entry['new_version'])
```
ekohl, line 63: This should match what's called. In this case it's `./scripts/update_package.rb --name $package_name --version $version`. So `directory` here doesn't make a lot of sense?

evgeni: I didn't look too closely at that part (which is copied from you 1:1), but yeah, probably

evgeni: looking closer, the directory names in `plugins/` are *not* **always** package names (yay):
* `ruby-foreman-azure-rm` is a package name
* `smart_proxy_chef` is a gem name

So, really, one would need to call `./scripts/update_package.rb --name $(basename $directory) --version $version`

ekohl: Looks like that works

scripts/list_updatable_packages
```diff
@@ -0,0 +1,67 @@
...
+
+SESSION = requests.Session()
+
+
+Package = namedtuple('Package', ['package_name', 'version', 'gem_name', 'directory'])
+
+
+def find_packages() -> Generator[Package, None, None]:
+    for plugin_rb in Path('plugins/').glob('**/*.rb'):
+        gem_name = plugin_rb.read_text().strip().split()[1].strip("',\"")
```
ekohl, line 23: Technically gem_name isn't really used anymore in the action so I wonder if we should drop this. In #7412 we could make the same argument. But I'm fine either way.

scripts/list_updatable_packages
```diff
@@ -0,0 +1,67 @@
...
+
+from debian.changelog import Changelog
+from debian.debian_support import NativeVersion
+
+import requests
+
+SESSION = requests.Session()
+
+
+Package = namedtuple('Package', ['package_name', 'version', 'gem_name', 'directory'])
```
ekohl, line 18: Recently I've taken to using [dataclasses](https://docs.python.org/3/library/dataclasses.html) but that requires Python 3.7.

scripts/list_updatable_packages
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+
+import json
+import os
+import sys
+from collections import namedtuple
+from pathlib import Path
+from typing import Generator, Iterable, Mapping, Optional
```
ekohl, line 8: `Mapping` and `Optional` are unused.

scripts/list_updatable_packages
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+
+import json
+import os
+import sys
```
ekohl, line 5: Unused import

ekohl (approved):
Thanks! This does work.

scripts/list_updatable_packages
evgeni: It certainly is needed to ask `rubygems.org` about the version? :)

scripts/list_updatable_packages
evgeni: What would be the benefit? (besides type hints)

ezr-ondrej (approved):
Nice :+1:
Can we get that in so I can update my PR to use both? (basically merge @ekohl's additions :))

scripts/list_updatable_packages
ekohl: The syntax feels more natural and you can have defaults. For example:
```python
@dataclasses.dataclass
class Package:
    package_name: str
    count: int = 0
    mylist: list[int] = dataclasses.field(default_factory=list)
```

The [field()](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) method can also determine if the field is used in the generated methods like `__repr__` and __eq__`.

I quite like it over `namedtuple`, especially if you want to add some additional methods which is always awkward with `namedtuple`.

scripts/list_updatable_packages
```diff
@@ -0,0 +1,66 @@
...
+    for plugin_rb in Path('plugins/').glob('**/*.rb'):
+        gem_name = plugin_rb.read_text().strip().split()[1].strip("',\"")
+        package_changelog = next(plugin_rb.parent.glob('**/changelog'))
+        with package_changelog.open() as debchangelog:
+            changelog = Changelog(debchangelog)
+            version = NativeVersion(changelog.version).upstream_version
+            yield Package(str(changelog.package), version, gem_name, plugin_rb.parent.as_posix())
+
+
+def latest_version(spec: Package) -> str:
```
ekohl, line 30: Is `spec` the right name?

evgeni: Copy paste from your script ;)

ekohl: Well, RPMs have specs. Do Debian packages have specs? ;)

evgeni: No. Honestly, I would have used "package" in both cases. The fact we're pulling data from a spec is IMHO irrelevant ;)

evgeni:
Thanks!

PR #7435, opened 2021-12-17
Respect GIT_AUTHOR_{NAME,EMAIL} env vars

ekohl:
Git itself prefers these environment variables over what's in the config. I couldn't find a better way in git itself to get the name and email.

This makes it easier to use in scripting such as GitHub Actions.

### DIFF
--- a/scripts/changelog.rb
+++ b/scripts/changelog.rb
@@ -5,8 +5,8 @@ require 'time'
 
 options = {:message => []}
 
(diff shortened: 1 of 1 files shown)


ekohl:
Confirmed in https://github.com/ekohl/foreman-packaging/pull/35 that this works.

evgeni (approved):

PR #7436, opened 2021-12-17
Add rubygem-rh_cloud_rhc_worker package and dependencies

ehelms:
This aims to package up https://github.com/ShimShtein/rhc-cloud-connector-worker and it's dependencies. This is a draft to start as I need to update it to the latest structure for the gem and we'll need to have an actual release of it. The dependent packages can be reviewed now as they are packaging up existing gRPC gems.

- [ ] Update to latest version of gem code and naming
- [ ] Get first release of the gem
- [ ] Add yggdrasil/rhc dependencies in the correct places


ehelms:
This is now just the RH cloud worker plugin, the dependencies have been split out to https://github.com/theforeman/foreman-packaging/pull/7438/

PR #7437, opened 2021-12-20
Add base modulemd for Foreman and Katello

ehelms:
This adds base modulemd configurations for Foreman and Katello repositories that would be used as inputs to mashing (see https://github.com/theforeman/foreman-infra/pull/1698). See modulemd reference (https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_stream_v2.yaml).

The base module metadata would be dynamically altered during mashing to add all the actual RPM builds when creating the new repository. This base metadata defines the module dependencies that allow for automatic module dependency resolution and ease of installation for users. Since the metadata and dependencies are per repository, we would have modules and streams each for Foreman and Katello. One open question might be how to handle something like PostgreSQL which has a module, and has a module stream we'd like to require but we do not require PostgreSQL in all cases (e.g. external database).

You can see some example output from playing around with this for the Satellite ecosystem:

```
# dnf module enable satellite
Last metadata expiration check: 0:00:26 ago on Wed Dec  8 02:11:41 2021.
Problems in request:
Modular dependency problems with Defaults:
 Problem: module satellite:7:0:-0.noarch requires module(ruby:2.7), but none of the providers can be installed
  - module ruby:2.5:8040020210629164822:9f9e2e7e-0.x86_64 conflicts with module(ruby:2.7) provided by ruby:2.7:8040020210805145045:9f9e2e7e-0.x86_64
  - module ruby:2.7:8040020210805145045:9f9e2e7e-0.x86_64 conflicts with module(ruby:2.5) provided by ruby:2.5:8040020210629164822:9f9e2e7e-0.x86_64
  - conflicting requests
Dependencies resolved.
==========================================================================================================================================================================================================================================================================================
 Package                                                              Architecture                                                        Version                                                              Repository                                                            Size
==========================================================================================================================================================================================================================================================================================
Enabling module streams:
 pki-core                                                                                                                                 10.6                                                                                                                                           
 pki-deps                                                                                                                                 10.6                                                                                                                                           
 ruby                                                                                                                                     2.7                                                                                                                                            
 satellite                                                                                                                                7                                                                                                                                              
Transaction Summary
==========================================================================================================================================================================================================================================================================================
```

### DIFF
--- a/modulemd/modulemd-foreman-el8.yaml
+++ b/modulemd/modulemd-foreman-el8.yaml
@@ -1,0 +1,28 @@
+---
+document: modulemd
+version: 2
+data:
+  name: foreman
+  stream: latest
+  version: 0
+  arch: x86_64
+  summary: Foreman module
+  description: >-
+    A module to make Foreman easier to install.
+  license:
+    module:
+      - GPLv3
+  profiles:
+    installer:
+      rpms:
+        - foreman-installer
+  dependencies:
+    - buildrequires:
+        platform: ['el8']
+        nodejs: ['12']
+        ruby: ['2.7']
+        postgresql: ['12']
+      requires:
+        platform: ['el8']
+        ruby: ['2.7']
+        postgresql: ['12']
--- a/modulemd/modulemd-katello-el8.yaml
+++ b/modulemd/modulemd-katello-el8.yaml
@@ -1,0 +1,25 @@
+---
+document: modulemd
+version: 2
+data:
+  name: katello
+  stream: latest
+  version: 0
+  arch: x86_64
+  summary: Katello module
+  description: >-
+    A module to make Katello easier to install.
+  license:
+    module:
+      - GPLv3
+  profiles:
+    installer:
+      rpms:
+        - foreman-installer-katello
+  dependencies:
+    - buildrequires:
+        platform: ['el8']
+        foreman: ['latest']
+      requires:
+        foreman: ['latest']
+        pki-core: []


modulemd/foreman_modulemd.yaml
```diff
@@ -0,0 +1,18 @@
+---
+document: modulemd
+version: 2
+data:
+    name: foreman
+    stream: latest
```
ekohl, line 6: Thinking about versioning: does it make sense to introduce the version number?

modulemd/katello_modulemd.yaml
```diff
@@ -0,0 +1,19 @@
...
+data:
+    name: katello
+    stream: latest
+    summary: Katello module
+    description: >-
+        A module to make Katello easier to install.
+    license:
+        module:
+            - GPLv3
+    dependencies:
```
ekohl, line 13: Should it also depend on the Foreman module?

modulemd/foreman_modulemd.yaml
ehelms: That depends on the user workflow we want. As I understand it, keeping the stream the same means you can have more seamless like experience. I would honestly need to play around a bit more with this by creating two versions with two versions of Foreman and see.

modulemd/katello_modulemd.yaml
ehelms: That's a good call.

modulemd/foreman_modulemd.yaml
ekohl: Mostly thinking out loud here.

I suppose that if you use `latest`, the upgrade workflow would be essentially the same as we have today: you change the yum repository and then run `dnf upgrade` and all the additional steps.

With a versioned approach you can keep the multiple repositories but it also becomes possible to have a single repository. With both you would still need an explicit instruction to tell dnf to use the new module.

Having a versioned approach also means the Katello module can properly depend on the correct Foreman version.

I also wonder what happens if the module would switch from Ruby 2.7 to 3.0. Can that happen within `latest` or are separate versions better?

A single repository has the benefit of no longer needing to manage the `yum.repos.d` file. However, it would also mean our metadata grows forever. Our current pipelines are also not at all ready for a single repository.

Another thought: can we have a `latest` and `latest-staging` module? I'm not sure if it makes sense.

So in short:
1. current yum.theforeman.org layout
  1.1. `latest` only
  1.2. explicitly state the version
2. Introduce a single modular repository with multiple module versions

Option 2 is unrealistic in the short term and it's questionable if it's desirable at all in the long term.

modulemd/katello_modulemd.yaml
ekohl: And nodejs is also a build requirement

modulemd/foreman_modulemd.yaml
```diff
@@ -0,0 +1,18 @@
...
+        A module to make Foreman easier to install.
+    license:
+        module:
+            - GPLv3
+    dependencies:
+      - buildrequires:
+            platform: [el8]
+        requires:
+            platform: [el8]
+            ruby: [2.7]
```
evgeni, line 18: + `postgresql: [12]`?

modulemd/katello_modulemd.yaml
evgeni: While it is a build requirement, I don't think *users* need this, so I'd keep the number of modules as short as possible.

evgeni:
> we do not require PostgreSQL in all cases (e.g. external database).

That is not exactly true. We do not require the PostgreSQL *server*, but we still need the client bits, and while the wire-protocol of PostgreSQL is rather stable (you can trivially use a 9.2 PostgreSQL client against a PostgreSQL 12), tools like `pg_dump` want a matching version (and we use that for e.g. backups).

modulemd/katello_modulemd.yaml
ekohl: But you literally have a `buildrequires` section. IMHO it would be nice to be explicit about what we do.

evgeni: indeed. that slightly slipped my mind.

modulemd/foreman_modulemd.yaml
evgeni: Explicitly stating the version means the user will have to perform wild `module reset` & `module enable` dances when upgrading between versions. At no benefit (as our repo will never offer multiple streams from the same repo).

evgeni: as for switching `ruby:2.7` to `ruby:3.0`, you can change that in the `requires`, but the user will need to perform a manual `dnf module reset` before this takes action.

evgeni: before that (`dnf module reset`) happens, no packages from that module are available to install, so the installer can't "fix" it for us :/

ekohl:
> ```
> # dnf module enable foreman
> # dnf install foreman-installer
> ```

If we make foreman a default module, would the enable still be needed? Are we allowed to do this?

> Given this makes the _install_ experience better, allows us to drop two hacks (hotfixes and the installer stuff), and doesn't _really_ worsen the upgrade experience, I think we should go for this for nightly and 3.2.

:+1:

evgeni:
I think we can create a default profile, yes, and that would maybe solve the enable step. I can try tomorrow.

evgeni:
So to make a module default, we'd need to add another document to the YAML file, according to the [modulemd-defaults](https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_defaults_v1.yaml) spec.

certainly possible, but maybe let's keep this to step 1.5, after we got the normal modular metadata in?

evgeni:
One think that confuses me about the [modulemd spec](https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_stream_v2.yaml): many fields are marked as `AUTOMATIC` which is supposed to mean "Filled in by the buildsystem." (aka MBS), but we don't really have a build system, and nothing is filling them in.

`dnf` doesn't seem to reject the metadata based on this (yet?), but many of those fields have a comment "Mandatory for module metadata in a yum/dnf repository.", so I would expect something, somewhere, to explode if it's not present.

Hunting for a linter as we speak.

modulemd/foreman_modulemd.yaml
ekohl: This is very helpful in understanding the implications for users. I appreciate it.

modulemd/foreman_modulemd.yaml
```diff
@@ -0,0 +1,28 @@
...
+    summary: Foreman module
+    description: >-
+        A module to make Foreman easier to install.
+    license:
+        module:
+            - GPLv3
+    profiles:
+        installer:
+            rpms:
+                - foreman-installer
```
ekohl, line 18: What does this mean at runtime? Is this when you run `dnf module install foreman`?

evgeni: `dnf module install foreman/installer` (as it is not the default profile, yet) but yes

evgeni:
changes since Eric is out:
- added a `profiles` section with a `installer` profile, so users can now do `dnf module install foreman/installer` or `dnf module install katello/installer`, which will do enable+install in one step
- added nodejs to buildrequires
- made katello just require foreman, not duplicate the modules foreman needs 
- added `version` and `architecture` fields (`version` to be updated by the generator playbook)

modulemd/katello_modulemd.yaml
```diff
@@ -0,0 +1,25 @@
...
+    profiles:
+        installer:
+            rpms:
+                - foreman-installer-katello
+    dependencies:
+      - buildrequires:
+            platform: ['el8']
+            foreman: ['latest']
+        requires:
+            foreman: ['latest']
```
ekohl, line 24: I guess versioning here would be nice since you can match the right Foreman to the Katello release, but given we don't have this today it's not a regression.

ekohl (approved):
Looks like a good start we can build on later.

PR #7438, opened 2021-12-22
Add rubygem-grpc and dependencies

ehelms:
Dependencies for the RH cloud RHC worker that are being split from https://github.com/theforeman/foreman-packaging/pull/7436 to get in ahead of time while we iron out the naming and conventions for the worker package.


evgeni:
[test rpm]

evgeni:
Dropped the odd x86_86 gems that were accidentally in the commit and confusing git annex.

evgeni:
so grpc fails to build on EL7 because it embeds a copy of abseil-cpp which can't be compiled with the GCC in EL7. :sob:

evgeni:
I am going to close this, as we're probably won't need the Ruby grpc stack at all.

Feel free to reopen if that changes ;)

PR #7441, opened 2022-01-04
Update rubygem-gettext_i18n_rails_js to 1.3.1

evgeni:
The dependency was bumped in https://github.com/theforeman/foreman/pull/8965, but packaging wasn't updated.


ekohl (approved):
Thanks!

PR #7446, opened 2022-01-06
drop dynflow_core and salt_core packages

evgeni:


ekohl (approved):
Remind me, do we want to add this the `tfm` RPM as something to obsolete? I recall previously it led to some problems with removing too much but by now we've updated all packages, right?

evgeni:
No, obsoletes means yum will install the obsoleting package as a replacement, and we don't strictly want that.

We want a `foreman-obsoletes` package, that does only obsoletes, but nobody implemented that yet.

ekohl:
Fair enough. Feel free to merge and perform the actual Koji cleanup (untagging etc).

evgeni:
(waiting on a installer build that actually includes the changes to puppet-foreman_proxy, as otherwise this would explode)

evgeni:
untagged dynflow_core, salt_core was already untagged.

PR #7450, opened 2022-01-10
use versioned obsoletes for proxy plugins

evgeni:
otherwise this confuses RPM when upgrading from EL7 (SCL) to EL8 (non-SCL) and both packages are available for a brief moment of time

### DIFF
--- a/packages/plugins/rubygem-smart_proxy_acd/rubygem-smart_proxy_acd.spec
+++ b/packages/plugins/rubygem-smart_proxy_acd/rubygem-smart_proxy_acd.spec
@@ -18,7 +18,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.3.0
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Summary: Application Centric Deployment smart proxy plugin
 Group: Applications/Internet
 License: GPLv3
@@ -48,8 +48,6 @@ Summary: Documentation for %{name}
 Group: Documentation
 Requires: %{name} = %{version}-%{release}
 BuildArch: noarch
-
-%{?scl:Obsoletes: rubygem-%{gem_name}-doc}
 
 %description doc
 Documentation for %{name}.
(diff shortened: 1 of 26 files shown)


packages/plugins/rubygem-smart_proxy_acd/rubygem-smart_proxy_acd.spec
```diff
@@ -49,7 +49,7 @@ Group: Documentation
 Requires: %{name} = %{version}-%{release}
 BuildArch: noarch
 
-%{?scl:Obsoletes: rubygem-%{gem_name}-doc}
+%{?scl:Obsoletes: rubygem-%{gem_name}-doc < %{version}-%{release}}
```
ekohl, line 52: It's surprising to me that this has an obsoletes for the -doc package (if SCL) but not for the main package. It feels to me that we should have both or neither. To be consistent, I'd say we should have both.

ekohl (approved):

packages/plugins/rubygem-smart_proxy_acd/rubygem-smart_proxy_acd.spec
evgeni: I think it was added *after* the SCL-ization of the proxy, and for some reason we removed the non-doc, but kept the doc one…

evgeni: proof: c8588a0e8487cb2010be21974fcbc98c418294d7

ekohl: That makes sense.

evgeni: so, cleaning this up, I'd just drop the obsoletes from ACD completely? as it really not required there?

evgeni:
Repoclosure is because of Salt, which is expected (it's currently untagged exactly because of that, and I'll untag again)

packages/plugins/rubygem-smart_proxy_acd/rubygem-smart_proxy_acd.spec
ekohl: Yes, just drop it indeed.

ekohl (approved):

PR #7457, opened 2022-01-14
Update bootdisk to 19.0.1

lzap:
(cherry picked from commit acacc5a3e306c43c3323fa7cbb90424d38bc4d7c)


lzap:
I incorrectly tested this package, I migrated DB without the plugin which worked, but the migration from a clean DB did not work yet. New version 19.0.2 is on the way to fix this.

PR #7487, opened 2022-01-18
Bump nodejs-theforeman-{vendor,builder} to 10.0.0

MariaAga:


ezr-ondrej:
https://github.com/theforeman/foreman/pull/9008 needs this and is ready to merge

lzap:
Can you guys start bumping versions in our official plugins too? This is out of control, I gotta catchup and I always run into this when doing package updates where we have strict versioning rules.

Maybe we can write some tool/script that would update `package.json` with versions from core or something? I don't know, I don't write UI yet I struggle with getting dependencies back to reasonable state. Or maybe there can be other more clever solution to this, I don't know.

lzap:
Opened https://community.theforeman.org/t/whats-up-with-package-json-in-plugins/27145

PR #7511, opened 2022-01-25
Update Foreman Gem and NPM dependencies

evgeni:

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -9,7 +9,7 @@
 %global scl_ruby_bin /usr/bin/%{?scl:%{scl_prefix}}ruby
 %global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
 
-%global release 2
+%global release 3
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman/foreman.spec
```diff
@@ -216,8 +214,6 @@ BuildRequires: %{?scl_prefix}rubygem(sprockets) >= 4.0
 BuildRequires: %{?scl_prefix}rubygem(sprockets) < 5.0
 BuildRequires: %{?scl_prefix}rubygem(sprockets-rails) >= 3.0
 BuildRequires: %{?scl_prefix}rubygem(sprockets-rails) < 4.0
-BuildRequires: %{?scl_prefix}rubygem(record_tag_helper) >= 1.0
-BuildRequires: %{?scl_prefix}rubygem(record_tag_helper) < 2.0
```
ekohl, line 220: We shouldn't forget to drop this, preferably before branching. @zjhuntin @pcreech @Odilhao.

evgeni: ?

ekohl: The `rubygem-record_tag_helper` is now no longer used so we can retire the package.

evgeni: ah! this = the package, not this = this hunk. gotcha.

ekohl: Yes, early morning reviews aren't the best. I opened https://github.com/theforeman/foreman-packaging/pull/7513 which should do this.

PR #7512, opened 2022-01-25
Stop building packages for RHEL 6

ekohl:
Since Foreman 3.0 this is no longer mashed so building and maintainig it is a waste of resources.

It does update foreman-release, but does not modify other specs since there is no benefit.


pcreech (approved):

ekohl:
@pcreech what do we do with the old branches/tags in Koji? Will you clean those up?

evgeni:
this broke client releases: https://github.com/theforeman/jenkins-jobs/pull/177

PR #7522, opened 2022-01-26
rename stream to el8 to avoid problems with puppet

evgeni:
for puppet `latest` has a special meaning, so if you instruct it to
enable the `foreman` module, `latest` stream, it actually looks for the
latest version available, not for the stream with the literal name
`latest`.

looking at the output of `dnf module list` on an EL8 machine, most
streams use the version of the software here, or `rhel8`.
we didn't want to use the version, as that would mean our users need to
explicitly `dnf module reset` and `dnf module enable` the newer version
on upgrades, which seems like a burden given the fact that our repo will
never contain multiple release streams at once anyways.

as Foreman also works fine on non-RH EL8, I propose the name `el8`

possible alternatives: current, released?

### DIFF
--- a/modulemd/modulemd-foreman-el8.yaml
+++ b/modulemd/modulemd-foreman-el8.yaml
@@ -3,7 +3,7 @@ document: modulemd
 version: 2
 data:
   name: foreman
-  stream: latest
+  stream: el8
   version: 0
   arch: x86_64
   summary: Foreman module
--- a/modulemd/modulemd-katello-el8.yaml
+++ b/modulemd/modulemd-katello-el8.yaml
@@ -3,7 +3,7 @@ document: modulemd
 version: 2
 data:
   name: katello
-  stream: latest
+  stream: el8
   version: 0
   arch: x86_64
   summary: Katello module


evgeni:
cc @ekohl @pcreech @Odilhao for your opinions

ekohl (commented):
Using `el8` in the puppet module and install instructions becomes more difficult when we add `el9` to the mix. I proposed `current` on IRC. `stable` is another option, but that would confusing since we also use it in nightly and even the staging builds on Koji.

I'm really stuck on describing the best thing. `any` would really be an accurate description but not really what a user would want to see. Something `foreman` or `from-repo` would be vague as well.

Naming is hard ...

evgeni:
```puppet
      package { 'foreman':
        ensure      => "el${facts['os']['release']['major']}",
        enable_only => true,
        provider    => 'dnfmodule',
        require     => Foreman::Repos['foreman'],
      }
```

doesn't seem too complicated? ;)

Right now the code is using `if $facts['os']['release']['major'] == '8'` anyways and would require adjustment for EL9.

ekohl:
True and I guess we already have per OS instructions anyway due to the differing `foreman-release.rpm`. I wouldn't mind `el$VERSION`.

pcreech:
Given what's written here, and the fact that we wouldn't be actually versioning the modules based on foreman x.y's, "el8" is the best I can think of as well.  +1

evgeni:
So, will anyone add a :green_apple: here? ;)

ekohl (approved):
There is no yellow apple so :yellow_circle: then?

evgeni:
:banana: is almost a yellow apple?

PR #7531, opened 2022-01-28
Update rubygem-foreman_rh_cloud to 4.0.31

Odilhao:


evgeni:
2.5 is EOL since December 2021 when 3.0 was released.

PR #7556, opened 2022-02-08
Drop centos-release-scl-rh from EL7 comps & foreman-release compare fix

ekohl:
See individual commit for details. My goal is for `./rel-eng/compare-with-tags` to be clean. Also the mashing process should ideally be free of warnings, but we're not quite there yet.

### DIFF
--- a/comps/comps-foreman-rhel7.xml
+++ b/comps/comps-foreman-rhel7.xml
@@ -8,7 +8,6 @@
     <uservisible>true</uservisible>
     <packagelist>
       <packagereq type="default">dynflow-utils</packagereq>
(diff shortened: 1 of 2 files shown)


ekohl:
To be complete, the current output of compare is:
 * [x] gofer
```shell
```

# Packages not expected in katello-nightly-rhel7
 * [x] python-qpid
 * [x] tfm-rubygem-qpid_messaging
```shell
koji remove-pkg katello-nightly-rhel7 python-qpid tfm-rubygem-qpid_messaging
```

I'm unsure about these, I've cleaned up the others.

evgeni (approved):

PR #7557, opened 2022-02-08
Update rubygem-smart_proxy_ansible to 3.3.1

ezr-ondrej:


adamruzicka:
Should we pick this to 3.2 and possibly 3.1?

ezr-ondrej:
It got into 3.2, I'm not sure about 3.1 (read: I do not know if there is something we need in 3.1)

adamruzicka:
There was the pty patch which might be helpful to people

PR #7561, opened 2022-02-09
Update rubygem-sequel to 5.53.0

evgeni:


evgeni:
sadly doesn't magically fix the dynflow issue, but hey…

PR #7566, opened 2022-02-10
Add in epoch to ensure priority over distro packages

pcreech:
Distros with rubygem-sinatra have "Epoch: 1", to ensure ours is installed we should have it too


ekohl:
Oh that's a good point. That sounds even better.

zjhuntin (approved):

PR #7589, opened 2022-02-16
Bump katello to 4.5

jlsherrill:


jlsherrill:
i believe the repoclosure failure will go away  after build, merging!

PR #7590, opened 2022-02-17
make installer explicitly depend on hostname

evgeni:
it's used in checks/hostname.rb, but the package is sometimes absent


ekohl (approved):
Is this worth a Redmine issue to properly cherry pick it?

evgeni:
I can create one, sure

PR #7587, opened 2022-02-16
Update smart_proxy_reports to 1.0.0

lzap:
* Nightly
* 3.2

Can do CP PR.

WARNING: The file `reports.rb` was missing, I have probably messed up when I initially added the package, please review carefully :)

PR #7604, opened 2022-02-21
Added Host Reports 1.0.0

lzap:
This is a new package.

* Nightly
* 3.2


lzap:
This is suffering from the same JS problem which I will fix in 1.0.1:

But on top of that there is this:

```
[](https://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-deb-pr-test/detail/foreman-packaging-deb-pr-test/351/pipeline/#step-69-log-1002)[](https://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-deb-pr-test/detail/foreman-packaging-deb-pr-test/351/pipeline/#step-69-log-1003)[](https://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-deb-pr-test/detail/foreman-packaging-deb-pr-test/351/pipeline/#step-69-log-1004)[2022-02-21T17:31:47.670Z] `/usr/share/foreman` is not writable.

[2022-02-21T17:31:47.670Z] Bundler will use `/tmp/bundler-20220221-3427713-16idmh8' as your home directory temporarily.

[2022-02-21T17:31:47.670Z] rake aborted!

[2022-02-21T17:31:47.670Z] LoadError: Error loading the 'postgresql' Active Record adapter. Missing a gem it depends on? pg is not part of the bundle. Add it to your Gemfile.
```

Any advice?

plugins/ruby-foreman-host-reports/debian/changelog
```diff
@@ -0,0 +1,5 @@
+ruby-foreman-webhooks (1.0.0) unstable; urgency=low
```
ekohl, line 1: This is not the plugin you're looking for.

ekohl (changes requested):
Please use the shared file in `debian/rules` since I believe that will fix your problems. Take a look at the bootdisk plugin for an example.

Also, please follow https://cbea.ms/git-commit/#imperative. So `Add Host Reports 1.0.0` as the commit message, not `Added`.

ekohl:
We have a shared rules file and include that. If you look at bootdisk, it should be fairly obvious:
https://github.com/theforeman/foreman-packaging/blob/240448f3033ef466d2cf128a4195f84589d150cc/plugins/ruby-foreman-bootdisk/debian/rules#L12-L18

lzap:
Oh now it makes sense, I was searching for the shared file in the git repo. Amended the changes.

plugins/ruby-foreman-host-reports/debian/rules
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/make -f
+# -*- makefile -*-
+# Sample debian/rules that uses debhelper.
+# This file was originally written by Joey Hess and Craig Small.
+# As a special exception, when this file is copied by dh-make into a
+# dh-make output file, you may use that output file without restriction.
+# This special exception was added by Craig Small in version 0.37 of dh-make.
```
ekohl, line 7: I think we should drop these from our rules files. These are nothing like the original anymore.

ekohl (approved):
Small nit, otherwise :+1: if the build is green.

plugins/ruby-foreman-host-reports/debian/rules
lzap: Yeah sure, dropped.

lzap:
Feel free to merge any/all of my host-reports or smart-proxy-reports packaging PRs (I have 4) I plan to test them against nightly this week and only after then issue backport PRs for 3.2.

ekohl:
I'm not sure why it's failing. It should be using `nulldb` instead of PG.

lzap:
I just tested the proxy plugin, works fine: https://github.com/theforeman/foreman-packaging/pull/7587

Yeah, I have the exactly the same conclusion - I would expect NullDB to be used. Trying to read logs if I spot something.

lzap:
So one issue is `npm ERR! syscall spawn git` - it looks like JS wants to list git files for some reason, I am adding git as a build dependency but I assume this should be now added as a foreman dependency?!

The other problem when Rails wants to load pg gem is a mystery for me.

lzap:
Weird it builds fine now, probably it was caused by the missing git?

Ron-Lavi:
@theforeman/packaging can you take another look? thanks

plugins/ruby-foreman-host-reports/debian/copyright
```diff
@@ -0,0 +1,33 @@
...
+ You should have received a copy of the GNU General Public License
+ along with this program. If not, see <http://www.gnu.org/licenses/>.
+ .
+ On Debian systems, the complete text of the GNU General
+ Public License version 3 can be found in "/usr/share/common-licenses/GPL-3".
+
+# Please also look if there are files or directories which have a
+# different copyright/license attached and list them here.
+# Please avoid to pick license terms that are more restrictive than the
+# packaged work, as it may make Debian's contributions unacceptable upstream.
```
evgeni, line 33:
```suggestion
```

plugins/ruby-foreman-host-reports/debian/changelog
```diff
@@ -0,0 +1,5 @@
+ruby-foreman-host-reports (1.0.1) unstable; urgency=low
```
evgeni, line 1:
```suggestion
ruby-foreman-host-reports (1.0.1-1) unstable; urgency=low
```

evgeni:
LGTM! Two minor nitpicks, good to go afterwards

evgeni:
[test deb]

lzap:
Pushed, please do not backport, we will be releasing another update this week for 3.2 with an important feature. 3.2 needs to go out without the plugin.

PR #7614, opened 2022-02-23
Require ansible or ansible-core on EL8+

evgeni:


ekohl (approved):
I suppose we'll need a cherry pick of this to stable branches.

evgeni:
> I suppose we'll need a cherry pick of this to stable branches.

yes, this is true.

PR #7615, opened 2022-02-23
Move repoclosure for 3.1 from vault to stream-8

Odilhao:

### DIFF
--- a/repoclosure/yum.conf
+++ b/repoclosure/yum.conf
@@ -17,25 +17,18 @@ module_platform_id=platform:el8
 [el8-baseos]
 name=BaseOS
 enabled=1
-#mirrorlist=http://mirrorlist.centos.org/?release=8&arch=x86_64&repo=BaseOS
-baseurl=http://vault.centos.org/centos/8/BaseOS/$arch/os/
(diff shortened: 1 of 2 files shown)


repoclosure/yum.conf
```diff
@@ -17,34 +17,27 @@ module_platform_id=platform:el8
 [el8-baseos]
```
upadhyeammit, line 17: Should we also update the labels to `stream8-baseos` and likewise for other repositories? I dont know if we will also need to update it somewhere else to not break things.

ekohl: I wouldn't do so. If you do, start with it in nightly and then see if it works.

Odilhao: We probably want to wait for 3.1 to pass after we merge, after this one runs and get green, we can/should change Develop and 3.2, tagging @ekohl to see what he thinks

zjhuntin (approved):

repoclosure/yum.conf
ekohl: I think I'd stick to el8. It's good enough here.

ekohl:
Are you planning to give nightly and 3.2 the same treatment? Those have also become very unstable due to vault.

PR #7620, opened 2022-02-23
Add hammer-cli-foreman-host-reports

ofedoren:
Supported Versions:

 * Nightly
 * 3.2


zjhuntin:
[test deb]

zjhuntin:
[test deb]

dependencies/bullseye/hammer_cli_foreman_host_reports/control
```diff
@@ -0,0 +1,25 @@
...
+Maintainer: Debian Ruby Extras Maintainers <pkg-ruby-extras-maintainers@lists.alioth.debian.org>
+Uploaders: <>
+Build-Depends: debhelper-compat (= 12),
+               gem2deb (>= 1),
+               rake,
+               ruby-hammer-cli-foreman (<< 4.0.0),
+               ruby-hammer-cli-foreman (>= 3.0.0)
+Standards-Version: 4.5.0
+Vcs-Git: https://salsa.debian.org/ruby-team/ruby-hammer-cli-foreman-host-reports.git
+Vcs-Browser: https://salsa.debian.org/ruby-team/ruby-hammer-cli-foreman-host-reports
```
evgeni, line 13: this isn't true, so let's not set those options :)

dependencies/bullseye/hammer_cli_foreman_host_reports/control
```diff
@@ -0,0 +1,25 @@
+Source: ruby-hammer-cli-foreman-host-reports
+Section: ruby
+Priority: optional
+Maintainer: Debian Ruby Extras Maintainers <pkg-ruby-extras-maintainers@lists.alioth.debian.org>
+Uploaders: <>
```
evgeni, line 5: this isn't true, as this package is not in debian, please set to yourself, or some generic foreman address.

dependencies/bullseye/hammer_cli_foreman_host_reports/salsa-ci.yml
```diff
@@ -0,0 +1,4 @@
+---
```
evgeni, line 1: you can drop this file, it;s unused in our case

ofedoren:
> tests fail with `File does not exist: minitest-spec-context` -- missing dependency?

Well, my assumption was that since it's required in test files only and this gem is installed with hammer-cli-foreman's Gemfile (https://github.com/theforeman/hammer-cli-foreman/blob/master/Gemfile#L11) I could simply require it in h-c-f-host-reports as well.

Please don't tell me that now I need to update Gemfile/gemspec and release a new version just because of this...

evgeni:
> > tests fail with `File does not exist: minitest-spec-context` -- missing dependency?
> 
> Well, my assumption was that since it's required in test files only and this gem is installed with hammer-cli-foreman's Gemfile (https://github.com/theforeman/hammer-cli-foreman/blob/master/Gemfile#L11) I could simply require it in h-c-f-host-reports as well.
> 
> Please don't tell me that now I need to update Gemfile/gemspec and release a new version just because of this...

That I didn't say.

The packaging of `hammer-cli-foreman` doesn't depend on `minitest-spec-context` (and, it seems, doesn't execute any tests at all). Given `m-s-c` isn't packaged in Debian, I think the easiest would be to not execute tests here too, but I am not exactly sure what triggers them.

dependencies/bullseye/hammer_cli_foreman_host_reports/ruby-tests.rake
```diff
@@ -0,0 +1,6 @@
+require 'gem2deb/rake/testtask'
```
evgeni, line 1: can you try dropping this file and see what happens? I think that's the one that executes the tests for us.

ofedoren:
ok to test

dependencies/bullseye/hammer_cli_foreman_host_reports/control
```diff
@@ -0,0 +1,23 @@
+Source: ruby-hammer-cli-foreman-host-reports
+Section: ruby
+Priority: optional
+Maintainer: Oleh Fedorenko <ofedoren@redhat.com>
+Uploaders: <>
+Build-Depends: debhelper-compat (= 12),
+               gem2deb (>= 1),
```
evgeni, line 7:
```suggestion
               gem2deb,
```

we don't need >=1 and it's not available in buster

dependencies/bullseye/hammer_cli_foreman_host_reports/control
```diff
@@ -0,0 +1,23 @@
+Source: ruby-hammer-cli-foreman-host-reports
+Section: ruby
+Priority: optional
+Maintainer: Oleh Fedorenko <ofedoren@redhat.com>
+Uploaders: <>
```
evgeni, line 5:
```suggestion
```

dependencies/bullseye/hammer_cli_foreman_host_reports/upstream/metadata
```diff
@@ -0,0 +1,9 @@
+---
+Archive: GitHub (FIXME)
```
ekohl, line 2: What's the FIXME part here?

evgeni: I have no idea what this file is, anyways ;)

ofedoren: https://wiki.debian.org/UpstreamMetadata#Fields doesn't tell much and I didn't find an example there :(

ofedoren:
@evgeni, all seems to be OK now :)

evgeni (approved):

PR #7626, opened 2022-02-24
Cleaning Vault mirror and moving to Centos 8 Stream

Odilhao:
(cherry picked from commit 8ab2cd05a655a3c99de3dfb7f7250a09d8518b7a)


Odilhao:
> Not against this, but wouldn't it be better to start in nightly and then do regular cherry picks?

Sure, ignore this for now, I'll go to develop and come back with cherry-picks

PR #7629, opened 2022-02-25
Updated hammer_cli_foreman to 3.1.0

ofedoren:
Supported Versions:

 * 3.1


ekohl (approved):

ekohl:
Note that for Debian we also need to update nightly since there are no automatic nightly packages like with RPM. That also means 3.2 needs an updated version.

ofedoren:
3.2 looks fine: https://github.com/theforeman/foreman-packaging/blob/deb/3.2/dependencies/bullseye/hammer_cli_foreman/changelog

PR #7641, opened 2022-02-28
Remove module_hotfixes=1 from repo files

ekohl:
There is now modular metadata. Removing this means that must be used, but it can properly enforce that the correct dependent modules are also enabled.

Right now I'm starting this, but I'm not sure yet what else needs to be modified so it's still a draft.


ehelms:
Replaced by https://github.com/theforeman/foreman-packaging/pull/7926

PR #7648, opened 2022-03-01
add mosquitto

evgeni:

### DIFF
--- a/packages/plugins/mosquitto/mosquitto.spec
+++ b/packages/plugins/mosquitto/mosquitto.spec
@@ -1,4 +1,5 @@
 %global with_tests 0
+%bcond_with websockets
 
 Name:           mosquitto
 Version:        2.0.14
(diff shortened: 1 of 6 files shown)


packages/plugins/mosquitto/mosquitto.spec
```diff
@@ -0,0 +1,471 @@
+%global with_tests 0
+%bcond_with websockets
+%bcond_with cjson
```
evgeni, line 3: I think we should have cjson support enabled (it is required for `mosquitto_ctrl` tool and the dynamic security plugin), but that needs cjson to be built, which I didn't do yet.

zjhuntin:
@evgeni this seems to be missing `libcjson.so.1()(64bit)` is that expected or is there something that needs to be changed?

evgeni:
I think when this repoclosure ran, the repo wasn't yet re-mashed. Just to be sure:

[test rpm]

PR #7676, opened 2022-03-08
fix thor bundling

evgeni:


evgeni:
@pcreech this was only one of three packages changed…

PR #7677, opened 2022-03-08
fix thor bundling also for bionic and focal

evgeni:

### DIFF
--- a/debian/bionic/foreman/changelog
+++ b/debian/bionic/foreman/changelog
@@ -1,3 +1,9 @@
+foreman (3.0.2-2) stable; urgency=low
+
+  * Fix thor bundling
+
+ -- Evgeni Golov <evgeni@debian.org>  Tue, 08 Mar 2022 20:57:39 +0100
+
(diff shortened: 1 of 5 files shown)


pcreech:
Looks like focal is still borked.

evgeni:
really? ;-)

I have no clue why tho.

evgeni:
:woman_shrugging:

evgeni:
welp, that built

@pcreech @ekohl @ehelms @upadhyeammit this essentially syncs the changes we did in 3.1+ for packaging into 3.0. I am rather confident that this "just works" (as it does in 3.1+) and wonder if we should do this for focal only (where we *need* it) or for all 3.0 supported releases (buster, bionic, focal)? 3.1 doesn't support bionic, so that's the only "risky" part.

ekohl:
If we would keep the branch alive longer, consistency would be important but given 3.0.2 is the last 3.0 release, I'd say apply it only for focal. That keeps the potential impact small.

pcreech (approved):

PR #7714, opened 2022-03-15
adjust dependencies to pull in right requests and pyyaml

evgeni:

### DIFF
--- a/packages/plugins/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
+++ b/packages/plugins/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
@@ -4,15 +4,7 @@
 %global collection_name foreman
 %global collection_directory %{_datadir}/ansible/collections/ansible_collections/%{collection_namespace}/%{collection_name}
 
-%if 0%{?fedora} || 0%{?rhel} >= 8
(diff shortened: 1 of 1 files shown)


packages/plugins/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
```diff
@@ -25,18 +17,19 @@ Source0:    https://galaxy.ansible.com/download/%{collection_namespace}-%{collec
 BuildArch:  noarch
 
 Provides: ansible-collection(%{collection_namespace}.%{collection_name}) = %{version}
-Provides: bundled(%{ansible_python}-apypie) = 0.3.2
+Provides: bundled(python-apypie) = 0.3.2
```
ekohl, line 20: Shouldn't this by python 2 vs 3?

ekohl (approved):

packages/plugins/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
evgeni: or python38? or python39…

technically, the code we vendor is version agnostic, and doesn't live in any python-specific (loadable) path, so no idea

PR #7730, opened 2022-03-18
Update rubygem-sidekiq to 5.2.10

ehelms:
Specifically updating because it includes a fix for a CVE. See https://github.com/mperham/sidekiq/commit/3f963816202940f36809561aef8e22f0cc1aea22


ekohl (changes requested):
When you do update, please use `bump_rpm.sh` since it includes code to update the requirements. I opened https://github.com/theforeman/foreman-packaging/pull/7731 which does that, matching the gemspec.

Once we drop EL7, we can also move to autorequires and drop it. For now it's good to keep it in.

PR #7732, opened 2022-03-18
Update Rails stack to 6.0.4.7

ehelms:
Catches latest bug releases and at least one CVE.


evgeni:
We explicitly pin to `~> 6.0.3.1` in the gemfile:
https://github.com/theforeman/foreman/blob/7a157e0156db8ff527dafe84cf78f5bc9dae2324/Gemfile#L10

So this should not work?

ehelms:
> We explicitly pin to `~> 6.0.3.1` in the gemfile: https://github.com/theforeman/foreman/blob/7a157e0156db8ff527dafe84cf78f5bc9dae2324/Gemfile#L10
> 
> So this should not work?

*This* will work. But Foreman won't. :fireworks:

ehelms:
Foreman PR: https://github.com/theforeman/foreman/pull/9157

ehelms:
Given that `bundler_ext` is designed to ignore version constraints in Gemfile, I think we can safely merge this change and then merge https://github.com/theforeman/foreman/pull/9157 to produce the least amount of potential breakage time?

evgeni:
Nope:
https://github.com/theforeman/foreman-packaging/blob/25d14bace424811f2d4e3ffec95214efd5de0889/packages/foreman/foreman/foreman.spec#L63

ehelms:
> Nope:
> 
> https://github.com/theforeman/foreman-packaging/blob/25d14bace424811f2d4e3ffec95214efd5de0889/packages/foreman/foreman/foreman.spec#L63

Touche. There are times where bounded dependencies are nice, and times when they are really annoying.

evgeni (approved):

evgeni:
https://github.com/theforeman/foreman-packaging/pull/7738

PR #7745, opened 2022-03-25
Update puppet-agent-oauth to 0.5.5

antaflos:


ekohl:
ok to test

antaflos:
@ekohl I just updated the commit message to include the line you specified.

ekohl:
I meant with `$current` the current message. So write `Fixes #34681 - Update puppet-agent-oauth to 0.5.5`

ekohl (approved):
Thanks! I'll try to be more explicit next time. You probably already noticed https://projects.theforeman.org/issues/34681#note-2 which tells you that linking worked. Once I merge it you should see more, like setting the Fixed in version field and setting the state to closed. The latter is on a cronjob so there's some delay in that.

PR #7747, opened 2022-03-25
Update puppet-agent-oauth to 0.5.5

antaflos:


antaflos:
@ekohl does this PR require anything more from me? I think you fixed the branch that was somehow wrongly set to master?

ekohl:
Yes, then CI ran and then I was out for a few days. Looks green so thanks!

Normally we prefer to first merge to `deb/develop` and then open cherry picks with `git cherry-pick -x`. That way you have a reference to the original and tracking can be a bit easier. However, since others were already merged I guess think can we merge this one too.

PR #7758, opened 2022-03-31
Add yggdrasil-worker-forwarder package

ehelms:


package_manifest.yaml
```diff
@@ -1084,6 +1084,7 @@ plugin_nonscl_packages:
     mosquitto: {}
     puppet-foreman_scap_client: {}
     puppetlabs-stdlib: {}
+    yggdrasil-worker-forwarder: {}
```
evgeni, line 1087: one thing that I realized: `yggdrasil` itself is in client, but it really needs to be in both, right?

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
```diff
@@ -0,0 +1,57 @@
...
+%else
+%{gobuild}
+%endif
+
+%install
+mkdir -p %{buildroot}%{_bindir}
+%{__install} -m 755 %{_builddir}/%{name}-%{version}/%{name} %{buildroot}/%{_bindir}
+
+%files
+%{_bindir}/%{name}
```
evgeni, line 50: why `%_bindir`? I would have expected something like `/usr/libexec/yggdrasil`?

ehelms: In ygg/rhc 0.2.0+, a worker is no longer defined by an exec in the libexec directory but rather by a configuration file that points at the exec. That allows both re-use of the exec for multiple workers and putting the exec in a more natural location. Our goal is to plan ahead for this by installing it in a general location. Then it can either be symlinked or a worker file can be deployed to libexec that wraps and calls this workers code.

package_manifest.yaml
ehelms: Truth.

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
evgeni: Right, but those files are not user-executable (unless they prepare the environment to be exactly as yggd dictates), so I wouldn't expect them in `%_bindir`/`%_sbindir`.

ehelms: > Right, but those files are not user-executable (unless they prepare the environment to be exactly as yggd dictates), so I wouldn't expect them in `%_bindir`/`%_sbindir`.

You can pass all the right options to it and it work, and if yggd is running. Where would you expect them to live then? I was trying not to bank on libexec existing in the future *and* breakaway from having to figure out which piece of software we have.

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
```diff
@@ -0,0 +1,57 @@
...
+Version: 0.0.1
+Summary: Worker service for Yggdrasil that can forward requests to an API endpoint
+Release: 1%{?dist}
+License: GPLv3
+
+Source0: https://github.com/ShimShtein/%{name}/releases/download/v%{version}/%{name}-%{version}.tar.gz
+Url: https://github.com/%{repo_orgname}/%{name}/
+
+# EL7 doesn't define go_arches
+%if ! 0%{?go_arches:1}
```
ekohl, line 17: This is only true if `go-srpm-macros.noarch` is installed. Should it be listed as `BuildRequires` on EL8+?

evgeni: BuildRequires aren't evaluated when SRPM is generated, which is when Arch is needed :woman_shrugging: 

It needs to be in the srpm group for the build tag.

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
evgeni: If we're to follow FHS, it should be either `/usr/libexec` or `/usr/lib/<tool>`.

I won't block on this, but bin seems wrong.

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
ekohl: Ok, but we also have gobuild at line 3. The fact it's defined here suggests to me something is wrong in the buildroot.

I'd still be in favor of specifying it with a comment it really belongs in the buildroot. Then also submit a change to https://github.com/theforeman/tool_belt/blob/master/configs/foreman/nightly.yaml so Koji has it.

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
ehelms: Aight - let me know what you think about this version.

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
ehelms: I am not really following the problem or the ask here. I mostly followed the precedence set by https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/client/foreman_ygg_worker/foreman_ygg_worker.spec

ekohl: The problem that I see is that there is a package `go-srpm-macros` (on EL8) that defines `go_arches` and `gobuild`. This spec (and the others) which makes me wonder if we're doing the right thing.

For what it's worth, that has:
```
# Define arches for PA and SA
%golang_arches   x86_64 %{arm} aarch64 ppc64le s390x
%gccgo_arches    %{mips}
%go_arches       %{golang_arches} %{gccgo_arches}

# ... much more ...

# Define commands for building
# BUILD_ID can be generated for golang build no matter of debuginfo
%gobuild(o:) \
CGO_CPPFLAGS="-D_FORTIFY_SOURCE=2 -fstack-protector-all" go build -compiler gc -buildmode pie '-tags=rpm_crashtraceback libtrust_openssl ' -ldflags "-linkmode=external -compressdwarf=false ${LDFLAGS:-} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v -x %{?**};\
```

ehelms: My understanding is that package is only on EL8, and is why we only define `go_build` and `go_arches` if they are not already defined so that EL7 builds will work.

ekohl: I get that, but my point is that currently **nothing** pulls in the macros on EL8. Not a `BuildRequires` (which, as @evgeni noted is insufficient for SRPM building) nor Koji. This means you essentially declare the macros in all situations. And I don't like that ambiguity. The comments and source in this file imply the macros are used. For example:

> `# EL7 doesn't define go_arches`

This isn't really a complete picture.

What I am worried about is that at some point we do have a package that does require `go-srpm-macros` and it's added to Koji. Suddenly this build is changed and it may break in unexpected ways.

So IMHO we should either include `go-srpm-macros` in our buildroot on EL8 or declare the `gobuild` macro unconditionally.

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
```diff
@@ -0,0 +1,57 @@
...
+%endif
+
+Name: yggdrasil-worker-forwarder
+Version: 0.0.1
+Summary: Worker service for Yggdrasil that can forward requests to an API endpoint
+Release: 1%{?dist}
+License: GPLv3
+
+Source0: https://github.com/ShimShtein/%{name}/releases/download/v%{version}/%{name}-%{version}.tar.gz
+Url: https://github.com/%{repo_orgname}/%{name}/
```
ekohl, line 14: I don't see `repo_orgname` defined

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
evgeni: I like that :)

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
ehelms: That's why I kept getting lost in the conversation, Koji does pull them in:

```
DEBUG util.py:636:   go-srpm-macros                noarch  2-17.el8                                       build   13 k
```
http://koji.katello.org/kojifiles/work/tasks/3069/493069/root.log

ehelms:
@evgeni @ekohl can y'all take another look?

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
evgeni: > So IMHO we should either include `go-srpm-macros` in our buildroot on EL8 or declare the `gobuild` macro unconditionally.

This is probably correct, but we do have other packages with these hacks in the repo today. So we can also fix them all at once when we figure out how to do it.

ekohl: I'd be OK with that in a separate step.

evgeni (approved):

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
ehelms: I am still not clear what the ask is. All I can think do is wrap those macros in `if rhel 7` checks to make it clearer that they are only invoked for EL7 builds. EL8 uses the macros defined by the `go-srpm-macros` as far as I can tell:


```
-%if ! 0%{?gobuild:1}
-%define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '" -a -v %{?**};
-%endif
-
 Name: yggdrasil-worker-forwarder
 Version: 0.0.1
 Summary: Worker service for Yggdrasil that can forward requests to an API endpoint
@@ -13,10 +9,6 @@ License: GPLv3
 Source0: https://github.com/ShimShtein/%{name}/releases/download/v%{version}/%{name}-%{version}.tar.gz
 Url: https://github.com/ShimShtein/%{name}/
 
-# EL7 doesn't define go_arches
-%if ! 0%{?go_arches:1}
-%define go_arches %{ix86} x86_64 %{arm} aarch64 ppc64le
-%endif
```

Scratch builds:
 - Failing EL7 -- http://koji.katello.org/koji/taskinfo?taskID=493592
 - Failing El8 - http://koji.katello.org/koji/taskinfo?taskID=493593

ekohl: Somehow I completely missed that. My bad.

packages/plugins/yggdrasil-worker-forwarder/yggdrasil-worker-forwarder.spec
```diff
@@ -0,0 +1,57 @@
+%define debug_package %{nil}
+
+%if ! 0%{?gobuild:1}
+%define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '" -a -v %{?**};
+%endif
+
```
ekohl, line 6: Considering the discussion below about `go-srpm-macros` I believe this is redundant
```suggestion
```

evgeni: `gobuild` is not defined in EPEL7's `go-srpm-macros`, while it is in CentOS8 ones.

evgeni:
```console
% koji add-pkg --owner kojiadm foreman-plugins-nightly-nonscl-rhel7 yggdrasil-worker-forwarder            
Adding 1 packages to tag foreman-plugins-nightly-nonscl-rhel7
% koji remove-pkg foreman-plugins-nightly-rhel7 yggdrasil-worker-forwarder
```

:smirk:

PR #7759, opened 2022-03-31
Update rubygem-foreman_remote_execution to 6.1.0

adamruzicka:


evgeni (approved):

evgeni:
```
   /usr/share/gems/gems/foreman_remote_execution-6.1.0/jsconfig.json
```

I guess it could be excluded?

adamruzicka:
@afeferku Is that file necessary or can we exclude it?

amirfefer:
can be excluded, uses for dev env.

PR #7767, opened 2022-04-01
Add rubygem-google-cloud-core package

stejskalleos:
For [google-cloud-compute package](https://github.com/theforeman/foreman-packaging/pull/7763)


package_manifest.yaml
```diff
@@ -555,6 +555,7 @@ foreman_nonscl_packages:
     keycloak-httpd-client-install: {}
     pcp-mmvstatsd: {}
     rubygem-foreman_maintain: {}
+    rubygem-google-cloud-core: {}
```
adamruzicka, line 558: I think this should go into `plugin_scl_packages`

PR #7777, opened 2022-04-04
RPM rubygem-foreman_google package

stejskalleos:
Waiting for https://github.com/theforeman/foreman-packaging/pull/8248


stejskalleos:
@theforeman/packaging Hi can anybody point me how to fix this error:

```
fatal: [rubygem-foreman_google]: FAILED! => changed=false 
changelog_epoch_version_release: ''
msg: Could not find specfile
specfile_epoch_version_release: ''
```
Spec file `rubygem-foreman_google.spec` is there, not sure what's wrong with it

zjhuntin:
[test rpm]

ekohl:
Marking as draft until the Katello side has been merged. Please mark it as ready for review once it's ready.

ehelms:
@stejskalleos we are at the point of being able to really test and get this plugin packaged. Testing this locally I think some of the dependencies are woefully out of date and will need updating and updated in a new release of the plugin, for example the version of foreman-js that the plugin depends upon.

BTW -- does the foreman_google plugin actually have UI to it? If it does, does it actually have webpack based UI components it needs to ship?

stejskalleos:
@ehelms 
> Testing this locally I think some of the dependencies are woefully out of date
True, updated in https://issues.redhat.com/browse/SAT-13306

> BTW -- does the foreman_google plugin actually have UI to it? If it does, does it actually have webpack based UI components it needs to ship?

No React, just rails `erb` files

packages/plugins/rubygem-foreman_google/rubygem-foreman_google.spec
```diff
@@ -0,0 +1,107 @@
...
+BuildRequires: npm(@babel/core) >= 7.7.0
+BuildRequires: npm(@babel/core) < 8.0.0
+BuildRequires: npm(@theforeman/builder) >= 10.0.0
+BuildRequires: npm(@theforeman/builder) < 11.0.0
+# end package.json devDependencies BuildRequires
+
+# start package.json dependencies BuildRequires
+BuildRequires: npm(react-intl) >= 2.8.0
+BuildRequires: npm(react-intl) < 3.0.0
+# end package.json dependencies BuildRequires
```
ehelms, line 41: @stejskalleos If foreman_google has no React code, then all of these npm requires should not be needed. I think they are being generated from the plugin having a package.json -- which should be superfluous if there is no React UI as part of the plugin? Should we first trim up the plugin to have only what is needed before packaging it?

packages/plugins/rubygem-foreman_google/rubygem-foreman_google.spec
```diff
@@ -0,0 +1,107 @@
...
+# by default, so that we can move it into the buildroot in %%install
+%gem_install
+
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+%foreman_bundlerd_file
+%foreman_precompile_plugin -s
```
ehelms, line 72: You mentioned there is no React UI, is there javascript at the ERB layer? If not, this step should not be needed.

packages/plugins/rubygem-foreman_google/rubygem-foreman_google.spec
stejskalleos: I've been thinking about it, right now there is no React code, but that will change in the future as there are some things in UI that are let's say not ideal and would require some changes. But when (or if) it will happen, hard to say.

I lean towards to remove all the `*.js` stuff right now, we can add it later when necessary.

stejskalleos: So I found that we actually have some JS in the plugin - [webpack/legacy.js](https://github.com/theforeman/foreman_google/blob/develop/webpack/legacy.js). It's just one file, but it's there.

ehelms (approved):

PR #7807, opened 2022-04-20
Require nftables for newer than EL7 operating systems

upadhyeammit:


ehelms (approved):
nftables is required by maintenance mode

PR #7818, opened 2022-04-21
Update rubygem-foreman_remote_execution to 5.0.6

adamruzicka:


Odilhao:
[test rpm]

adamruzicka:
> No matching package to install: 'npm(@theforeman/builder) >= 10.1.0'

Can I remove the version bump? Will it build? Will it build but break on runtime?

pcreech:
Looks like this was the commit that changed the dep:  https://github.com/theforeman/foreman_remote_execution/commit/1b3db1bb4b678c9994b7afb98110dc37633caba3

I am fairly certain just removing the requires from the rpm spec will still fail due to nodejs version resolution, since ```"^10.1.0"``` will only match any 10.y version.

Options are to back down that version in the gem or update foreman-vendor and friends in 3.1.  I'm not sure the latter is advisable.

adamruzicka:
But wait, the change was already in 5.0.5 which got in just fine? And possibly even works?

pcreech:
[test rpm]

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -47,7 +47,7 @@ BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
 BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
 BuildRequires: %{?scl_prefix}npm(@patternfly/react-catalog-view-extension) >= 4.8.126
 BuildRequires: %{?scl_prefix}npm(@patternfly/react-catalog-view-extension) < 5.0.0
-BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 8.16.0
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 10.1.0
```
evgeni, line 50:
```suggestion
BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 8.16.0
```

evgeni (approved):

PR #7819, opened 2022-04-22
Fixes #34801 - update foreman-protector file paths

upadhyeammit:


upadhyeammit:
> if I am correct with what i see then compilation is happening,
> 
> ```
>  bytecode_compilation_path=./usr/lib/yum-plugins/foreman-protector.py
> + failure=0
> ```
> 
> However I dont see **pycache** in /usr/lib/yum-plugins/ even if I try to install old package on EL7 system, not sure why it did not complain earlier?

I misread the code, originally __pycache__ was not expected to be on EL7,
~~~
%if 0%{?rhel} > 7
%{yumplugindir}/__pycache__/foreman-protector.cpython*
%endif
~~~

packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec
```diff
@@ -4,13 +4,14 @@
 %global gem_name foreman_maintain
 %global directory_name foreman-maintain
 %global yumplugindir %{_prefix}/lib%{nil}/yum-plugins
+%global dnfplugindir %{_prefix}/lib%{nil}/python3.6/site-packages/dnf-plugins
```
evgeni, line 7:
```suggestion
%global dnfplugindir %{python3_sitelib}/dnf-plugins

evgeni (changes requested):
one nitpick, otherwise good to go

evgeni (approved):

PR #7831, opened 2022-04-26
Bump rubygem-foreman_rh_cloud to 5.0.34-1

ShimShtein:


packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -115,6 +115,8 @@ cp -a .%{gem_dir}/* \
 %exclude %{gem_cache}
 %{gem_spec}
 %{foreman_bundlerd_plugin}
+%{foreman_apipie_cache_foreman}
+%{foreman_apipie_cache_plugin}
```
ehelms, line 119: These two need to be re-dropped. They were removed with the drop of apipie:cache support.

pcreech (approved):

PR #7849, opened 2022-04-28
katello: update gem deps

evgeni:


evgeni:
needs https://ci.theforeman.org/job/katello-master-source-release/ to actually finish :(

PR #7862, opened 2022-05-03
Add dracut config as discovery subpackage

lzap:
We are unable to regenerate dracut image via livecd-creator due to ordering issues. A BZ was raised that keyboard driver is missing for MS Hyper-V VMs. The missing drivers can be only configured before kernel installs, unless I am wrong doing this in %pre should do the trick.

More info at https://issues.redhat.com/browse/SPMM-9540

This is nightly only change, upstream we do not build FDI this way (we use lorax nowdays). Therefore this new subpackage does not need to be whitelisted at all upstream, we need to pull it downstream to fix the problem in brew.

### DIFF
--- a/packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
+++ b/packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
@@ -1,14 +1,14 @@
 Name: foreman-discovery-image-service
 Version: 1.0.0
-Release: 4%{?dist}
+Release: 5%{?dist}
 Summary: Metapackage with dependencies for FDI
 
 Group: Applications/System
 License: GPLv2+
 URL: https://github.com/theforeman/foreman-discovery-image
 
-# explicitly define, as we build on top of an scl, not inside with scl_package 
-%{?scl:%global scl_prefix %{scl}-} 
+# explicitly define, as we build on top of an scl, not inside with scl_package
+%{?scl:%global scl_prefix %{scl}-}
 
 Requires:	foreman-proxy
 Requires:	%{?scl_prefix}rubygem-smart_proxy_discovery_image
@@ -30,6 +30,13 @@ BuildRequires: %{?scl_prefix_ruby}rubygems-devel
 %description tui
 Metapackage with dependencies for FDI text-user interface
 
+%package dracut
+Summary: Metapackage with dracut config
+Requires: dracut
+
+%description dracut
+Metapackage with dracut configuration
+
 %prep
 
 %build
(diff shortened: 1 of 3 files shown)


lzap:
Build failed, let me rebase and see why.

packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
```diff
@@ -40,7 +47,20 @@ Metapackage with dependencies for FDI text-user interface
...
 
+%files dracut
+
+%pre dracut
+# Because livecd-creator executes %post scriptlets after it copies initramdisk out of the
+# image, it is not possible to rebuild kernel ramdisk with extra drivers for VMWare and
+# MS Hyper-V. Therefore the configuration must be dropped before the kernel package
+# executes its %post scriptlet. Doing it in %pre ensures the configuration is created
+# beforehand.
+echo 'add_drivers="mptbase mptscsih mptspi hv_storvsc hid_hyperv hv_netvsc hv_vmbus"' > %{_sysconfdir}/dracut.conf.d/99-discovery.conf
```
evgeni, line 58: Why is this a post script vs just a file shipped in the package?

evgeni (changes requested):
The package needs to be added to comps if you want to ship it in the repos (which I guess you do).

No whitelisting required.

packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
lzap: So correct me if I am wrong, but the goal is to have that configuration in-place before kernel's %post scriptlet is executed. Since %post scriptlets are executed one by one as packages are installed I cannot guarantee the order (because kernel can/will likely install before this package).

This actually makes me think if I should actually change this to `%pretrans` that is guaranteed to be executed before "transaction", whatever that actually is: https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/

evgeni: Wait, what, the result is dependant on the order dnf decides to install the packages in? What the actual fuck?!

No, I am not going to think about this at this time.

If you're going to edit the file in any scriptlet, please add a `%ghost` entry to the files section, so rpm at least knows who owns the file.

lzap: I have found a solution, %triggerprein is exactly what I need. I am gonna try to test that.

BTW do I miss buildroot on this line? I guess this is correct:

```
echo 'xxx' > %{buildroot}%{_sysconfdir}/dracut.conf.d/99-discovery.conf
```

lzap:
Changing to %triggerprein, added %ghost, fixed the missing buildroot macro, whitelisted the subpackage.

Do not merge yet, I am going to test the scratchbuild manually by creating a repository.

packages/plugins/foreman-discovery-image-service/foreman-discovery-image-service.spec
lzap: Damn this must NOT have a buildroot macro of course! :) Jeez and its Monday.

https://pbs.twimg.com/media/FN9eS68X0AIVaqv.jpg

lzap:
Created repo at https://people.redhat.com/~lzapleta/temp/fdi-dracut/

PR #7874, opened 2022-05-09
log plugin installation in posttrans

evgeni:


evgeni:
Job aborted during repoclosure 🙄
But it did build!

Odilhao (approved):

PR #7878, opened 2022-05-10
Update rubygem-apipie-bindings to 0.5.0

ofedoren:
Supported Versions:

 * Nightly
 * 3.3


evgeni:
replacing this with https://github.com/theforeman/foreman-packaging/pull/7885 -- which also now uses the template so that we catch such updates easier in the future.

PR #7883, opened 2022-05-10
Changing Repoclosure, Rel-eng and Manifests from nigthly to Foreman 3.3

Odilhao:


upadhyeammit (approved):
Looks good with updated mock config changes and gpg key

PR #7888, opened 2022-05-11
Fixes #34896 - properly detect default_program in k-c-h

evgeni:
`get_default_program` needs `@proxy_hyphenated` to be properly defined
to work, so it needs to be called *after* the option parse block


ehelms:
@evgeni needs a rebase

PR #7942, opened 2022-05-19
Update rubygem-fog-google to 1.15.0, rubygem-google-api-client to 0.50.0

evgeni:


packages/foreman/rubygem-fog-google/rubygem-fog-google.spec
```diff
@@ -110,6 +108,9 @@ cp -pa .%{gem_dir}/* \
 %doc %{gem_instdir}/MIGRATING.md
 
 %changelog
+* Thu May 19 2022 Evgeni Golov 1.15.0-1
```
evgeni, line 111: Not updating to 1.16+ yet, as that needs a whole bunch new deps.

evgeni:
needs to be merged once https://github.com/theforeman/foreman/pull/9226 is in

ehelms (approved):

ehelms:
That current EL7 failure looks unpleasant.

evgeni:
> That current EL7 failure looks unpleasant.

rexml is a built-in gem before 3.0, so for now we can just drop it ;)

PR #7776, opened 2022-04-04
[WIP] foreman-google plugin [v0.0.1]

stejskalleos:
For https://github.com/stejskalleos/foreman_google

PR #7961, opened 2022-05-25
DEB dependencies for foreman_google

stejskalleos:
For https://github.com/theforeman/foreman-packaging/pull/7776


evgeni:
you don't need to package debian plugin deps

PR #7970, opened 2022-05-25
Add katello-pull-transport-migrate RPM

ehelms:
The intended workflow for this RPM is to take a katello-agent connected host and turn it into a pull transport connected host. This can be done either manually or through a katello-agent triggered package installation. This does not currently turn off or cleanup katello-agent.

However, I wonder if it should? If the intent is to convert, why leave a dangling katello-agent around? If so, I think this RPM could obsolete katello-agent?

### DIFF
--- a/packages/client/katello-pull-transport-migrate/katello-pull-transport-migrate.spec
+++ b/packages/client/katello-pull-transport-migrate/katello-pull-transport-migrate.spec
@@ -1,0 +1,40 @@
+Name:           katello-pull-transport-migrate
+Version:        1.0.1
+Release:        1%{?dist}
+Summary:        An RPM that migrates katello-agent users to the new pull transport
+BuildArch:      noarch
+
+License:        LGPLv2
+URL:            https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/client/katello-pull-transport-migrate
+Source0:        https://github.com/theforeman/%{name}/archive/refs/tags/%{version}.tar.gz#/%{name}-%{version}.tar.gz
+
+Requires:       yggdrasil
+Requires:       foreman_ygg_worker
+
+%description
+An RPM that migrates katello-agent users to the new pull transport.
+
+Relies on the existing Red Hat Subscription Manager configuration to know which Katello server it should communicate with.
+The scriptlet will fail if the client is not registered to a Foreman server with the Katello plugin.
+This RPM obtains the relevant values from the existing RHSM configuration, writes the configuration to yggdrasil's config.toml,
+and starts the pull transport agent.
+
+%prep
+%setup -q
+
+%build
+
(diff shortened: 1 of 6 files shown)


ehelms:
@adamruzicka @jeremylenz

jeremylenz:
as @jlsherrill once said, removing katello-agent via REX is a great way to test out pull provider :smile:

jeremylenz:
If k-p-t-m fails for some reason, we wouldn't want to remove katello-agent. Can the RPM ensure that it leaves the host with at least one working communication method?

packages/client/katello-pull-transport-migrate/katello-pull-transport-migrate.spec
```diff
@@ -0,0 +1,43 @@
...
+This RPM obtains the relevant values from the existing RHSM configuration, writes the configuration to yggdrasil's config.toml,
+and starts the pull transport agent.
+
+%prep
+%setup -q -c
+
+%build
+
+%install
+rm -rf %{buildroot}
```
evgeni, line 28: ain't nobody needs that in the year 2022

packages/client/katello-pull-transport-migrate/katello-pull-transport-migrate.spec
```diff
@@ -0,0 +1,43 @@
...
+mkdir -p  %{buildroot}%{_sbindir}
+cp %{name}-%{version}/%{name} %{buildroot}%{_sbindir}
+
+%clean
+rm -rf %{buildroot}
```
evgeni, line 33: ain't nobody needs that in the year 2022

packages/client/katello-pull-transport-migrate/katello-pull-transport-migrate.spec
ehelms: Sounds like a Conan O'Brien skit :smile:

adamruzicka:
Just tested the happy path, works like a charm

ehelms:
This is updated with 1.0.1 version

comps/comps-foreman-client-rhel6.xml
```diff
@@ -16,6 +16,7 @@
       <packagereq type="default">katello-client-repos</packagereq>
       <packagereq type="default">katello-host-tools</packagereq>
       <packagereq type="default">katello-host-tools-fact-plugin</packagereq>
+      <packagereq type="default">katello-pull-transport-migrate</packagereq>
```
evgeni, line 19: we don't build for EL6 anymore, neither does yggd work on EL6…

evgeni (approved):
ACK with a small EL6 nit ;)

ehelms:
[test rpm]

PR #7983, opened 2022-05-26
Send email if RPM or Deb package update fails

ehelms:
This needs a few problems solved first regarding what email to use to send and how to handle secrets. I figure this is also a good way to see what is required to handle this.

### DIFF
--- a/.github/workflows/bump_packages.yml
+++ b/.github/workflows/bump_packages.yml
@@ -55,6 +55,20 @@ jobs:
         title: "Update ${{ matrix.package_name }} to ${{ matrix.new_version }}"
         body: ''
         delete-branch: true
+    - name: Send email to Discourse on failure
+      if: failure()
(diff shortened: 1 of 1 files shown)


.github/workflows/bump_packages.yml
```diff
@@ -55,6 +55,22 @@ jobs:
...
         body: ''
         delete-branch: true
+    - name: Send email to Discourse on failure
+      if: failure()
+      uses: dawidd6/action-send-mail@v3
+      with:
+        server_address:
+        server_port:
+        username: ${{secrets.MAIL_USERNAME}}
+        password: ${{secrets.MAIL_PASSWORD}}
```
evgeni, line 65:
```suggestion
        server_address: community.theforeman.org
        server_port: 25
```

no need no auth if we talk to our machine directly ;)

evgeni (approved):

PR #7992, opened 2022-05-30
Revert "Set default value to empty list if no matrix is calculated"

evgeni:
Reverts theforeman/foreman-packaging#7984

Errors out with:
```
The workflow is not valid. .github/workflows/bump_packages.yml (Line: 13, Col: 15): Unexpected symbol: '['. Located at position 34 within expression: steps.set_list.outputs.matrix || [] .github/workflows/bump_packages.yml (Line: 63, Col: 15): Unexpected symbol: '['. Located at position 34 within expression: steps.set_list.outputs.matrix || []
```


evgeni:
cc @ehelms FYI!

PR #8001, opened 2022-06-01
add el9 client config

evgeni:


evgeni:
here is a successful build using this: http://koji.katello.org/koji/taskinfo?taskID=516153

Odilhao (approved):

PR #7994, opened 2022-05-31
Rpm/foreman netbox dependency

dgoetz:
All dependencies for foreman_netbox, will create a second PR for foreman_netbox after merge.

PR #8003, opened 2022-06-01
Update rubygem-ipaddress to 0.8.3

dgoetz:
Separated from #7994


packages/foreman/rubygem-ipaddress/rubygem-ipaddress.spec
```diff
@@ -47,10 +46,8 @@ Documentation for %{pkg_name}
 # Remove un-needed file
 # See https://github.com/bluemonk/ipaddress/issues/23
 rm .%{gem_instdir}/.document
-
-pushd .%{gem_instdir}
-%patch0 -p1
-popd
+rm .%{gem_instdir}/.gitignore
+rm .%{gem_instdir}/.rock.yml
```
evgeni, line 50: you can put those via `%exclude` in the `%files` section, which I'd prefer (but not block on)

evgeni (approved):

PR #8004, opened 2022-06-01
Update rubygem-foreman_webhooks to 3.0.3

ofedoren:
Supported Versions:

 * Nightly
 * 3.3
 * 3.2


packages/plugins/rubygem-foreman_webhooks/rubygem-foreman_webhooks.spec
```diff
@@ -34,6 +34,9 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
 BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
 BuildRequires: %{?scl_prefix}npm(@babel/core) < 8.0.0
 BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 8.15.0
+BuildRequires: %{?scl_prefix}npm(@theforeman/builder) < 9.0.0
```
dgoetz, line 37: This upper boundary is preventing the build as the current version of it is 10.1.0. This probably needs to be lifted in the gems package.json as the dependencies are created from there.

dgoetz (changes requested):

ofedoren:
@dgoetz, thanks, updated.

dgoetz (approved):
Not sure if the dependency should be relaxed so much, but from a packaging perspective it looks fine and the tests are green, so going to merge.

PR #8031, opened 2022-06-03
Update rubygem-ipaddress to 0.8.3 (3.2)

dgoetz:
(cherry picked from commit b18be70bbe2b63853e151d02be995df79d452884)


ekohl (approved):
For what it's worth, I've been trying to get rid of `ipaddress` because it does appear to be unmaintained, but I guess this is still needed.

PR #8048, opened 2022-06-07
Fixes #35028 - drop the infrastructure facet when renaming

evgeni:
otherwise we end up with two machines having
`infrastructure_facet.foreman=true` and that will result in wrong
behaviour down the road

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -204,6 +204,10 @@ If not done, all hosts will lose connection to #{@options[:scenario]} and discov
       run_cmd("LANG=en_US.UTF-8 hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
     end
 
+    def curl_cmd(method, url, exit_codes=[0], message=nil)
+      run_cmd("curl --silent --user #{@options[:username].shellescape}:#{@options[:password].shellescape} --request #{method} https://#{@old_hostname}/#{url}", exit_codes, message)
+    end
+
     def get_default_proxy_id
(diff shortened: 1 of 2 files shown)


evgeni:
cc @adamruzicka @shimshtein

ShimShtein:
While the fix looks good, I wonder if we should add a single instance validation to the facet to prevent it in future use cases.
@adamruzicka, you opinion?

adamruzicka:
I considered this, but decided not to do so. There might be scenarios where having multiple hosts marked as Foreman is completely valid, although not currently supported, such as HA.

adamruzicka:
When the installer runs in a scenario with a smart proxy, it places two markers, one to mark a host as a foreman and the other which links the host against a smart proxy object. Should we remove the smart proxy link on hostname change?

adamruzicka:
Yes and yes

evgeni:
Thanks Adam, updated the code!

packages/katello/katello/hostname-change.rb
```diff
@@ -438,6 +442,8 @@ def run
       unless @foreman_proxy_content
         STDOUT.puts "\nUpdating default #{@proxy}"
         proxy_id = self.get_default_proxy_id
+        self.curl_cmd('DELETE', "/api/v2/instance_hosts/#{@old_hostname}")
+        self.curl_cmd('DELETE', "/api/smart_proxies/#{proxy_id}/hosts/#{@old_hostname}")
```
adamruzicka, line 446: Shouldn't we be explicit about api version? Or at least consistent?

evgeni: probably, I think the one I copied from you from IRC, the other from the API docs.

PR #8071, opened 2022-06-27
Fixes #34898 - Add ACS refresh cron job

chris1984:


ianballou:
We'll have to merge this first: https://github.com/Katello/katello/pull/10177

packages/katello/katello/katello.cron
```diff
@@ -3,3 +3,6 @@ PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 
 # Script for initiating removal of orphaned content
 00 22 * * 0	root	foreman-rake katello:delete_orphaned_content RAILS_ENV=production >/dev/null 2>&1
+
+# Script to refresh all alternate content sources
+00 18 * * 0     root    foreman-rake katello:refresh_alternate_content_sources >/dev/null 2>&1
```
ianballou, line 8: The tabbing isn't lining up with the entry above

ianballou: Should we include RAILS_ENV=production like the one above?

chris1984: Yeah, let me fix it and push up

chris1984: Added production, and will push up. Here is my file, it looks like they are matching up:

```
SHELL=/bin/sh
PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin

# Script for initiating removal of orphaned content
00 22 * * 0     root    foreman-rake katello:delete_orphaned_content RAILS_ENV=production >/dev/null 2>&1

# Script to refresh all alternate content sources
00 18 * * 0     root    foreman-rake katello:refresh_alternate_content_sources RAILS_ENV=production >/dev/null 2>&1
```

ianballou: I think I see the issue, on the above line it has single tabs for white space, but yours are spaces.

chris1984: @ianballou updated to use tabs

chris1984:
@ehelms can this get a merge? Didn't know if you wanted me to bump the spec file, since Partha will be doing GA in a few days.

ianballou (approved):
Looks good to me!

ekohl (commented):
From an architectural standpoint: why is this a cronjob? Any reason this can't be done in a recurring task?

From a platform point of view I think we should get rid of all cronjobs (except perhaps the cronjob that cleans tasks itself). Mostly because if you move to different architectures (like containers or Highly Available setup) dealing with cron becomes a pain.

And yes, this is also true for deleting orphaned content.

I'm not saying we can't merge this, but it is a discussion that should happen.

chris1984:
I will leave that response up to @ianballou as he is the feature lead and decided on cron.

ianballou:
> From an architectural standpoint: why is this a cronjob? Any reason this can't be done in a recurring task?

We chose cron because it comes turned-on out of the box and is easily configurable.  We also picked it because orphan cleanup already uses it.

I don't know enough about recurring tasks to say if it would be a good idea for these two tasks. Can recurring tasks come preconfigured out-of-the-box?  Would they require new API / UI changes for a complete user experience?  We'd want to make sure that users don't have to think about setting these items.

ianballou:
Moving forward, if we make the ACS refresh task a recurring logic we'd also want to move orphan cleanup at the same time. If we merge this now as-is, I can create an issue and get it on our near-term backlog to migrate to recurring logic.

ekohl (approved):
I don't mind merging this, but it should be noted that the cron file is marked as a config file. That means if a user has made any modifications the new file will be installed as `/etc/cron.d/katello.rpmnew`. I should also admit that I don't know if cron filters out `.rpmnew` files. I'd say there is either no guarantee that the cron actually runs or things may run twice.

Perhaps @evgeni knows what would happen.

evgeni:
> I don't mind merging this, but it should be noted that the cron file is marked as a config file. That means if a user has made any modifications the new file will be installed as `/etc/cron.d/katello.rpmnew`. I should also admit that I don't know if cron filters out `.rpmnew` files. I'd say there is either no guarantee that the cron actually runs or things may run twice.
> 
> Perhaps @evgeni knows what would happen.

I'd hope it is smart…

```console
% git grep 'config.*cron'
gem2rpm/foreman_plugin.spec.erb:%config(noreplace) %{_root_sysconfdir}/cron.d/%{gem_name}
gem2rpm/smart_proxy_plugin.spec.erb:%config(noreplace) %{_root_sysconfdir}/cron.d/%{gem_name}
packages/client/katello-host-tools/katello-host-tools.spec:%config(noreplace) %attr(0644, root, root) %{_sysconfdir}/cron.d/katello-tracer-upload
packages/foreman/foreman/foreman.spec:%config %{_sysconfdir}/cron.d/%{name}
packages/katello/katello/katello.spec:%config(missingok) %{_sysconfdir}/cron.d/katello
packages/plugins/rubygem-foreman_expire_hosts/rubygem-foreman_expire_hosts.spec:%config(noreplace) %{_root_sysconfdir}/cron.d/%{gem_name}
packages/plugins/rubygem-smart_proxy_omaha/rubygem-smart_proxy_omaha.spec:%config(noreplace) %attr(0644, root, root) %{_root_sysconfdir}/cron.d/%{name}
packages/plugins/rubygem-smart_proxy_openscap/rubygem-smart_proxy_openscap.spec:%config(noreplace) %attr(0644, root, root) %{_root_sysconfdir}/cron.d/%{name}
packages/plugins/rubygem-smart_proxy_salt/rubygem-smart_proxy_salt.spec:%config %{_root_sysconfdir}/cron.d/%{gem_name}
```

;-)

evgeni:
That said, why is this a cron job vs a Foreman task?

evgeni:
Ooooh, I did only see the last comments. Will re-read!

evgeni (changes requested):
This needs a rev-bump and a changelog entry (so that it actually gets built), but otherwise LGTM!

evgeni (changes requested):

packages/katello/katello/katello.spec
```diff
@@ -132,6 +132,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Thu Jul 14 2022 Chris Roberts - 4.6.0-0.4.master
```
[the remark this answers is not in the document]
chris1984: Updated :) Copy and paste failure

chris1984:
Looks like there is issues with faraday and pulp on repoclosure.

evgeni (approved):

chris1984:
[test rpm]

chris1984:
@evgeni all green, can this get a merge?

PR #8074, opened 2022-06-28
Remove EL7 comps files

ehelms:
This is the first in a series of PRs targeting the removal of EL7 assets. I am opting to open each "area" as a separate dedicated PR rather than as one giant PR with multiple commits to make it easier on reviewers to examine specific areas where these changes take affect.


ekohl (commented):
We also have scripts like https://github.com/theforeman/foreman-packaging/blob/rpm/develop/add_gem_package.sh https://github.com/theforeman/foreman-packaging/blob/rpm/develop/add_npm_package.sh and https://github.com/theforeman/foreman-packaging/blob/rpm/develop/add_pypi_package.sh which add packages to comps. I'd appreciate it if we also update that so they continue working. I think historically they mostly worked on the EL7 comps so we can update them to el8 now.

ehelms:
Updated the references I could find about comps.

add_gem_package.sh
```diff
@@ -78,14 +78,9 @@ add_to_tito_props() {
 
 add_gem_to_all_comps() {
 	add_gem_to_comps $TITO_TAG
-	if [[ $TITO_TAG == "foreman-nightly-rhel7" ]] || [[ $TITO_TAG == "foreman-nightly-nonscl-rhel7" ]] ; then
-		add_gem_to_comps foreman-nightly-el8
-	elif [[ $TITO_TAG == "foreman-plugins-nightly-rhel7" ]] || [[ $TITO_TAG == "foreman-plugins-nightly-nonscl-rhel7" ]] ; then
-		add_gem_to_comps foreman-plugins-nightly-el8
-	elif [[ $TITO_TAG == "foreman-client-nightly-rhel7" ]] ; then
+	if [[ $TITO_TAG == "foreman-client-nightly-rhel7" ]] ; then
```
ekohl, line 81: Thinking out loud: should we recommend always passing EL8 as a base and make the scripts work with that?

ekohl (approved):
Note to self: This won't be an issue with mashing because that was already dropped.

PR #8077, opened 2022-06-28
Drop EL7 koji tags and EL7 only packages

ehelms:

PR #8075, opened 2022-06-28
Drop EL7 repositories from repoclosure config

ehelms:
See https://github.com/theforeman/foreman-packaging/pull/8074#issue-1287819268

PR #8078, opened 2022-06-28
Drop el7 refactor manifest

ehelms:
Requires:

 - https://github.com/theforeman/foreman-packaging/pull/8077
 - https://github.com/theforeman/foreman-packaging/pull/8075


ekohl (commented):
This now needs a rebase due to conflicts. Also, similar to https://github.com/theforeman/foreman-packaging/pull/8074#pullrequestreview-1028665888 it would be nice if the tools that use these headers are updated.

PR #8082, opened 2022-06-30
Bump hammer katello to 1.5.2

chris1984:


ekohl (changes requested):
Place remove the preleasesource on line 9. That should get the build to pass.

ekohl (approved):
Let's see if it builds now.

packages/katello/rubygem-hammer_cli_katello/rubygem-hammer_cli_katello.spec
```diff
@@ -6,14 +6,13 @@
 %global plugin_name katello
 
 %global release 1
-%global prereleasesource pre.master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
```
ekohl, line 9: Looks like this line should also be dropped.

parthaa:
Opened here -> https://github.com/theforeman/foreman-packaging/pull/8164

PR #8088, opened 2022-07-01
Update rubygem-fx to 0.7.0

ehelms:


packages/katello/rubygem-fx/rubygem-fx.spec
```diff
@@ -3,10 +3,11 @@
 %{!?scl:%global pkg_name %{name}}
 
 %global gem_name fx
+%global gem_require_name %{gem_name}
```
ekohl, line 6: What is this new macro?

ehelms: Appears that you added it about two years ago (https://github.com/theforeman/foreman-packaging/pull/4921/files#diff-6192aa78a351560738c958b59214bab5d795aaf4c53870c0e9a56446a92a833dR8) and that's how long it has been since this gem had a spec update.

PR #8092, opened 2022-07-05
Enable building directly through Koji CLI

ehelms:
This drops the need for use of tito to handle building scratch or
release builds of RPMs and uses Koji directly through the CLI in Obal
commands. This will allow dropping the tito functionality and relying
on the package manifest as the source for all configuration and
for doing invidualized builds based on tag. This also paves the way
to make tagging of existing builds a part of the workflow.

Requires:

- [x] https://github.com/theforeman/obal/pull/303
- [x] https://github.com/theforeman/obal/pull/272

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -11,6 +11,7 @@ all:
 packages:
   vars:
     rpmlint_rc_file: "{{ inventory_dir }}/.rpmlintrc"
+    build_package_use_koji_build: true
   children:
     foreman_packages: {}
     plugin_packages: {}


ehelms:
The relevant PRs have been merged and we check out Obal from source for our CI. With this change, all actions - scratch, and release will skip the tito layer all together and use Koji and RPM directly. Additionally, with this flag present there is new functionality that checks if a build already exists and tags it into the target tag.

If we want to use this for Foreman 3.4 and work out any bugs that might exist we should consider merging it prior to branching (~3 weeks away at the time of this comment).

ekohl (approved):
I suppose a revert is quite easy.

ehelms:
> I suppose a revert is quite easy.

That was one of my goals! Deterministically enable-able and easy to change!

ekohl:
https://github.com/theforeman/foreman-packaging/pull/8250#issuecomment-1187708602

PR #8165, opened 2022-07-06
Allow bump_rpm to update specfile

ehelms:

### DIFF
--- a/gem2rpm/default.spec.erb
+++ b/gem2rpm/default.spec.erb
@@ -18,6 +18,7 @@ config.rules[:ignore] << 'web'
 config.rules[:ignore] << '.github'
 config.rules[:ignore] << '.ruby-version'
 config.rules[:ignore] << '.coveralls.yml'
+config.rules[:ignore] << 'Gemfile.lock'
 -%>
 %global gem_name <%= spec.name %>
 <% unless spec.extensions.empty? -%>
--- a/gem2rpm/default.spec.erb
+++ b/gem2rpm/default.spec.erb
@@ -28,7 +28,6 @@ Name: rubygem-%{gem_name}
 Version: <%= spec.version %>
 Release: 1%{?dist}
 Summary: <%= spec.summary.gsub(/\.$/, "") %>
-Group: Development/Languages
 License: <%= spec.licenses.empty? ? 'FIXME' : spec.licenses.map { |l| LICENSES.fetch(l, l) }.join(" and ") %>
 <% if spec.homepage -%>
 URL: <%= spec.homepage.gsub('http://', 'https://') %>
@@ -66,7 +65,6 @@ BuildArch: noarch
 <% if doc_subpackage -%>
 %package doc
 Summary: Documentation for %{name}
-Group: Documentation
 Requires: %{name} = %{version}-%{release}
 BuildArch: noarch
 
--- a/gem2rpm/default.spec.erb
+++ b/gem2rpm/default.spec.erb
@@ -74,15 +74,11 @@ Documentation for %{name}.
 <% end # if doc_subpackage -%>
 
 %prep
-gem unpack %{SOURCE0}
-
-%setup -q -D -T -n  %{gem_name}-%{version}
-
-gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
+%setup -q -n  %{gem_name}-%{version}
 
 %build
 # Create the gem as gem install only works on a gem file
-gem build %{gem_name}.gemspec
+gem build ../%{gem_name}-%{version}.gemspec
 
 # %%gem_install compiles any C extensions and installs the gem into ./%%gem_dir
 # by default, so that we can move it into the buildroot in %%install
@@ -108,9 +104,10 @@ rm -rf %{buildroot}%{gem_instdir}/<%= ext.split(File::SEPARATOR).first %>/
 mkdir -p %{buildroot}%{_bindir}
 cp -a .%{_bindir}/* \
         %{buildroot}%{_bindir}/
-find %{buildroot}%{gem_instdir}/<%= spec.bindir %> -type f | xargs chmod a+x
 
+find %{buildroot}%{gem_instdir}/<%= spec.bindir %> -type f | xargs chmod a+x
 <% end -%>
+
 <% unless spec.extensions.empty? -%>
 %check
 # Ideally, this would be something like this:
@@ -122,14 +119,12 @@ mkdir -p gem_ext_test/gems/extensions/%{_arch}-%{_target_os}/$(ruby -r rbconfig 
 cp -a %{buildroot}%{gem_extdir_mri} gem_ext_test/gems/extensions/%{_arch}-%{_target_os}/$(ruby -r rbconfig -e 'print RbConfig::CONFIG["ruby_version"]')/
 GEM_PATH="./gem_ext_test/gems:$GEM_PATH" ruby -e "require '%{gem_require_name}'"
 rm -rf gem_ext_test
-
 <% end -%>
+
 %files
 %dir %{gem_instdir}
-<% unless spec.executables.nil? or spec.executables.empty? -%>
 <% for f in spec.executables -%>
 %{_bindir}/<%= f %>
-<% end -%>
 <% end -%>
 <% unless spec.extensions.empty? -%>
 %{gem_extdir_mri}
--- a/gem2rpm/foreman_plugin.spec.erb
+++ b/gem2rpm/foreman_plugin.spec.erb
@@ -39,7 +39,6 @@ Name: rubygem-%{gem_name}
 Version: <%= spec.version %>
 Release: 1%{?foremandist}%{?dist}
 Summary: <%= spec.summary.gsub(/\.$/, "") %>
-Group: Applications/Systems
 License: <%= spec.licenses.empty? ? 'FIXME' : spec.licenses.map { |l| LICENSES.fetch(l, l) }.join(" and ") %>
 <% if spec.homepage -%>
 URL: <%= spec.homepage.gsub('http://', 'https://') %>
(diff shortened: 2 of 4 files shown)


bump_rpm.sh
```diff
@@ -79,13 +79,22 @@ if [[ $CURRENT_VERSION != $NEW_VERSION ]] ; then
 	git add $SPEC_FILE
 
 	if [[ $PACKAGE_NAME == *rubygem-* ]]; then
+		CHANGELOG=$(mktemp)
+		sed -n '/%changelog/,$p' $SPEC_FILE > $CHANGELOG
+
 		TEMPLATE="$(awk '/^# template: / { print $3 }' $SPEC_FILE)"
-		if [[ $TEMPLATE == 'scl' || $TEMPLATE == 'nonscl' ]] ; then
+		if [[ $TEMPLATE == 'scl' || $TEMPLATE == 'nonscl' || $TEMPLATE == '' ]] ; then
```
ekohl, line 86:
```suggestion
		if [[ $TEMPLATE == 'scl' || $TEMPLATE == 'nonscl' || -z $TEMPLATE ]] ; then
```

bump_rpm.sh
```diff
@@ -79,13 +79,22 @@ if [[ $CURRENT_VERSION != $NEW_VERSION ]] ; then
...
+
 		if [[ -n $TEMPLATE ]] ; then
-			echo "* Updating requirements"
-			gem2rpm -t $ROOT/gem2rpm/$TEMPLATE.spec.erb *.gem | $SCRIPT_DIR/update-requirements specfile - $SPEC_FILE
+			sed -i 's/\s\+$//' $SPEC_FILE
+			gem2rpm -t $ROOT/gem2rpm/$TEMPLATE.spec.erb -o $SPEC_FILE *.gem
+
+			cat $CHANGELOG >> $SPEC_FILE
+			sed -i '/^%changelog/,/^%changelog/{0,//!d}' $SPEC_FILE
+			rm $CHANGELOG
```
ekohl, line 96: This unconditionally always rewrites the entire spec, but that wipes out local changes. I think this should be hidden behind an env var.

A safer way could be that it only rewrites if the template is `scl`, `nonscl` or missing, otherwise falls back to the safer replacement.

ekohl (changes requested):

bump_rpm.sh
ehelms: What scenario do you forsee where you don't want the spec file getting updated? or that you have local changes where you want to bump the RPM?

ekohl: It is mostly about the spec files where the template is insufficient (https://github.com/theforeman/foreman-packaging/pull/8166 comes to mind). I don't think we'll ever capture everything and we do have a bot now that submits PRs using the script. So I'd stay on the safe side.

packages/foreman/rubygem-deep_cloneable/rubygem-deep_cloneable.spec
```diff
@@ -1,77 +1,78 @@
...
+%global gem_name deep_cloneable
 
+Name: rubygem-%{gem_name}
+Version: 3.2.0
+Release: 1%{?dist}
 Summary: This gem gives every ActiveRecord::Base object the possibility to do a deep clone
-Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 3.0.0
-Release: 4%{?dist}
 Group: Development/Languages
```
ekohl, line 8: We could drop the group from the template. Fedora has done that already and it doesn't add anything.

packages/foreman/rubygem-deep_cloneable/rubygem-deep_cloneable.spec
```diff
@@ -1,77 +1,78 @@
...
+BuildRequires: rubygems-devel
 BuildArch: noarch
-Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
-%{?scl:Obsoletes: ruby193-rubygem-%{gem_name}}
+# end specfile generated dependencies
 
 %description
 Extends the functionality of ActiveRecord::Base#dup to perform a deep clone
-that includes user specified associations.
+that includes user specified associations. .
```
ekohl, line 22: This is an example of where the template is insufficient. I've opened https://github.com/moiristo/deep_cloneable/pull/142 but for now:
```suggestion
that includes user specified associations.
```

packages/foreman/rubygem-deep_cloneable/rubygem-deep_cloneable.spec
```diff
@@ -1,77 +1,78 @@
...
-%exclude %{gem_instdir}/.document
 %{gem_spec}
 
 %files doc
 %doc %{gem_docdir}
-%doc %{gem_instdir}/LICENSE
-%doc %{gem_instdir}/readme.md
 %doc %{gem_instdir}/CHANGELOG.md
+%{gem_instdir}/Gemfile
+%{gem_instdir}/Gemfile.lock
```
ekohl, line 67: Should we exclude this file?

packages/foreman/rubygem-deep_cloneable/rubygem-deep_cloneable.spec
```diff
@@ -1,77 +1,78 @@
...
+%{gem_instdir}/Rakefile
+%{gem_instdir}/deep_cloneable.gemspec
```
ekohl, line 69: I wonder why this shows up. We have `%{gem_spec} already so why is it duplicated?

ekohl (commented):
Did you intend to include the deep_clonable update in the same commit?

packages/foreman/rubygem-deep_cloneable/rubygem-deep_cloneable.spec
ehelms: So far, I have not been able to track down why the logic is grabbing this as a `doc_files`

ehelms:
I added a drop of Group and ignore on Gemfile.lock. This also includes a commit that updates a few of the pre/build/setup sections to match Fedora template as I think our steps were a bit out dated.

bump_rpm.sh
ekohl: > What is the end goal there?

I'd say that it allows a conservative update by default but you can force to fully regenerate specs (on the next update) by increasing the template's version.

> Sorry, I know you said this earlier but I realize I do not know what you mean by this.

Essentially:
```sh
if [[ $TEMPLATE == 'scl' || $TEMPLATE == 'nonscl' || -z $TEMPLATE ]] ; then
  regenerate_spec_file
elif [[ -n $TEMPLATE ]] ; then
  update_spec_file
fi
```

ehelms: What is the difference between regenerate and update in this context? Are you implying that we would have something like:

```
gem2rpm/default-latest.spec.erb
gem2rpm/default-2.spec.erb
gem2rpm/default-1.spec.erb
```

And if a spec file had `# template: default 1` the regenerate action would use this template and the update action would use the latest template?

I am getting a bit lost in the nuance here and maybe that is a sign that we are overengineering things a bit. Whenever we update the version of a gem we know that there are things that may change:

 * requirements (build or runtime)
 * new files that may need to be ignored

Those things are captured by doing a specfile update.Then we have some special case scenarios where some gems may have extracurricular additions to their spec file that break the norm. My argument is this is already a problem that we run into anyway and we should incrementally start adapting to this by first accepting that we should also run a specfile update anytime we do a version bump. The script and bot do the hard work of detecting updates and opening PRs, maintainers still should review and look for anomalies and then take over the PR with updates as needed. That feels sane to me while we engineer a way to preserve the extracurricular's. That's how I looked at how we might incrementally and iteratively automate this more and more to reduce packaging burden.

ekohl: My biggest worry is that we can't rely on gem2rpm to generate a 100% correct spec, so we must make some (small) modifications. If every time the bot tries to update it those get reverted we actually get a lot of additional work. This is why I initially wrote the `update-requirements` script the way I did: allow it to replace specific parts. If we decide that files should also be replaced we can also add similar markers to `%files` to replace those bits automatically.

Then we also have cases where we do want to regenerate the entire spec file. Getting rid of the SCL macros is an excellent example: it doesn't happen often, but when it does the automation helps us a lot.

That's why I'm leaning to a cautious approach: essentially opt-in full replacement. Right now we have a case where we know it's probably the right thing to do ((non)scl to default). My versioning thoughts were about how we can do a similar thing in the future.

So short term I'd suggest:
```bash
                if [[ $TEMPLATE == 'scl' ]] || [[ $TEMPLATE == 'nonscl' ]] || [[ -z $TEMPLATE ]]; then
                        CHANGELOG=$(mktemp)
                        trap "rm -f $CHANGELOG" EXIT
                        sed -e '1,/%changelog/ d' $SPEC_FILE > $CHANGELOG
                        gem2rpm -t $ROOT/gem2rpm/default.spec.erb -o $SPEC_FILE *.gem
                        cat $CHANGELOG >> $SPEC_FILE
                        git add $SPEC_FILE
                elif [[ -n $TEMPLATE ]] ; then
                        echo "* Updating requirements"
                        gem2rpm -t $ROOT/gem2rpm/$TEMPLATE.spec.erb *.gem | $SCRIPT_DIR/update-requirements specfile - $SPEC_FILE
                        git add $SPEC_FILE
                fi
```

Then later, when we do want to force regeneration of the default template, we can think about another trigger to force regeneration.

To improve on regeneration we can change the files section to:
```erb
%files
# start specfile generated files main
%dir %{gem_instdir}
....
%{gem_spec}
# end specfile generated files main
```

And the same with docs:
```erb
<% if doc_subpackage -%>
%files doc
# start specfile generated files doc
%doc %{gem_docdir}
<%= doc_files.to_rpm %>
# end specfile generated files doc
<% end # if doc_subpackage -%>
```

This way `update-requirements` can replace files too.

ehelms: A thought I had been noodling on -- https://github.com/theforeman/foreman-packaging/pull/8169

ekohl: How about we start with https://github.com/theforeman/foreman-packaging/pull/8206 and then see where it goes from there?

ekohl:
Could you rebase this on top of `rpm/develop`? I do like the spec file cleanups which I didn't include in my PR.

ekohl (approved):
I didn't check if packages build with it, but that should be easy enough to figure out.

ehelms:
[test rpm]

PR #8167, opened 2022-07-07
Update rubygem-foreman_ansible to 8.0.1

stejskalleos:


ekohl:
The bot also opened https://github.com/theforeman/foreman-packaging/pull/8190 but it looks like the upper bound of `@theforeman/builder` is quite old. Still, I prefer that PR over this and I'll get that one updated.

PR #8169, opened 2022-07-08
Introduce an exceptions and default template for naming

ehelms:
This is a bit of a thought experiment better shown in code. Where we could use the template naming property to identify if a package is generated with the default and thereby safe to regenerate or if a package has exceptions and is generated with the exceptions template. That could then enable our code to choose whether to regenerate or not based upon the template name.

I'll note, as I played with this I also started thinking about using this approach for common fragments across the templates to align them. For example, across the templates the `%prep`, `%build`, `%install` are the same, the doc sub-package definition as well. Thus we could do things like:

```
<%= File.read(File.join(__dir__, '/gem2rpm/section/build.erb')) %>
<%= File.read(File.join(__dir__, '/gem2rpm/section/doc_subpackage.erb')) %>
```

### DIFF
--- a/gem2rpm/base.spec.erb
+++ b/gem2rpm/base.spec.erb
@@ -1,0 +1,153 @@
+<%-
+require 'json'
+root = `git rev-parse --show-toplevel`.strip
+LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
+
+config.rules[:doc] << /\/?CHANGES/
+config.rules[:doc] << 'sample'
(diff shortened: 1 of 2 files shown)


ekohl:
While splitting that could reduce duplication, I wonder if we won't be overengineering things. Today the templates are fairly straight forward, but splitting it into multiple files makes it quite hard to follow. We already have two layers of templating (RPM itself is also a templating language) and this would introduce a third layer.

PR #8206, opened 2022-07-11
Regenerate outdated rubygem spec files

ekohl:
Prior to this bumping the RPM version for Rubygems used a conservative approach in updating its spec file. The reason was to persist local changes.

However, many spec files are old in what they do. This changes the bumping strategy to regenerate the spec if the template is scl, nonscl or missing altogether. For other templates it maintains the old strategy.


ehelms (approved):
I am ok taking this as an incremental approach.

PR #7954, opened 2022-05-24
RPM dependencies for foreman_google package

stejskalleos:
Dependencies for https://github.com/theforeman/foreman-packaging/pull/7777

PR #8217, opened 2022-07-13
Update rubygem-fog-google

ekohl:
This may overlap with https://github.com/theforeman/foreman-packaging/pull/7954 but these are generated using the new template and I'd like to fix rpm/develop.


ekohl:
I first need to bump faraday: https://github.com/theforeman/foreman-packaging/pull/8218

ekohl:
Scope updated to just wholesale update rubygem-fog-google, including dependencies. This will likely give a repoclosure failure on rubygem-google-api-client but that whole package is unused after this (and should be removed).

ekohl:
It fails because webrick is a bundled gem.
```
[2022-07-13T20:21:05.278Z]     package: rubygem-google-apis-core-0.7.0-1.el8.noarch from downloaded_rpms
[2022-07-13T20:21:05.278Z]       unresolved deps:
[2022-07-13T20:21:05.278Z]         rubygem(webrick)
```

packages/plugins/rubygem-google-apis-core/rubygem-google-apis-core.spec
```diff
@@ -0,0 +1,71 @@
...
+%prep
+%setup -q -n  %{gem_name}-%{version}
+
+# With Ruby < 3 rexml is included
+# May also be not required: https://github.com/googleapis/google-api-ruby-client/pull/11024
+%gemspec_remove_dep -g rexml
+
+# Should not be a hard requirement
+# https://github.com/googleapis/google-api-ruby-client/issues/9560
+%gemspec_remove_dep -g webrick
```
ekohl, line 40: This is not pretty, but it's the easiest way to move forward now.

PR #8220, opened 2022-07-13
Update rubygem-faraday_middleware to 1.2.0

ekohl:
This is needed to get a version that's compatible with faraday 1.x, but currently breaks rubygem-netbox-client-ruby: https://github.com/ninech/netbox-client-ruby/pull/58#discussion_r920532108


ekohl:
> This is needed to get a version that's compatible with faraday 1.x, but currently breaks rubygem-netbox-client-ruby: [ninech/netbox-client-ruby#58 (comment)](https://github.com/ninech/netbox-client-ruby/pull/58#discussion_r920532108)

https://github.com/theforeman/foreman-packaging/pull/8223 was merged to solve that. That's now building.

ekohl:
Repoclosure is on the old versions.

PR #8224, opened 2022-07-14
Update gem2rpm templates

ekohl:
This aligns all templates on the modern standard, which allows using gemspec_{add,remove}_dep macros without modification.

### DIFF
--- a/gem2rpm/default.spec.erb
+++ b/gem2rpm/default.spec.erb
@@ -4,8 +4,12 @@ require 'json'
 root = `git rev-parse --show-toplevel`.strip
 LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
 
+config.rules[:misc] << /\/?CHANGES/
+config.rules[:misc] << /samples?/
+config.rules[:misc] << /examples?/
+
+config.rules[:doc] << /^.+\.md$/
 config.rules[:doc] << /\/?CHANGES/
-config.rules[:doc] << 'sample'
 config.rules[:doc] << /\/?SECURITY/
 
 config.rules[:ignore] << 'Guardfile'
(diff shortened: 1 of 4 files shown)


ekohl:
There are more gem2rpm fixes needed. I'll collect them here.

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -101,15 +107,11 @@ Documentation for %{name}.
...
-
-%setup -q -D -T -n  %{gem_name}-%{version}
-
-gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
+%setup -q -n  %{gem_name}-%{version}
 
 %build
 # Create the gem as gem install only works on a gem file
-gem build %{gem_name}.gemspec
+gem build ../%{gem_name}-%{version}.gemspec
```
ehelms, line 114: Curious about this change, why is the location at a different level for plugins?

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -77,6 +76,13 @@ BuildRequires: gcc
 BuildArch: noarch
 <% end -%>
 Provides: foreman-plugin-%{plugin_name} = %{version}
+<% if has_assets || has_webpack -%>
+<% for d in spec.runtime_dependencies -%>
+<% for req in d.requirement -%>
+BuildRequires: rubygem(<%= d.name %>)<%= " #{req}" unless req.empty? %>
+<% end -%>
+<% end -%>
+<% end -%>
```
ehelms, line 85: Sad :/ What step of the build related to assets is it that ends up requiring all the runtime dependencies present? (my memory fails me)

gem2rpm/foreman_plugin.spec.erb
ekohl: This is aligning it to `default.spec.erb` where we made the same change.

gem2rpm/foreman_plugin.spec.erb
ekohl: `foreman-rake assets:precompile[$plugin]`

gem2rpm/foreman_plugin.spec.erb
ehelms: Oh right, the `../` threw me.

ehelms (approved):

gem2rpm/default.spec.erb
```diff
@@ -6,6 +6,7 @@ LICENSES = JSON.load(File.read(File.join(root, 'licenses.json')))
 
 config.rules[:doc] << /\/?CHANGES/
 config.rules[:doc] << 'sample'
+config.rules[:doc] << 'example'
 config.rules[:doc] << /\/?SECURITY/
```
ekohl, line 10: So I found out why this didn't appear to work anymore: https://github.com/fedora-ruby/gem2rpm/issues/115

ehelms:
I'd be OK taking in these round of changes as it will unblock a set of packages to get built.

PR #8231, opened 2022-07-15
Use modularity in repoclosure

ekohl:
This is a test. It's incomplete, but the principle should be testable.

### DIFF
--- a/repoclosure/repoclosure.sh
+++ b/repoclosure/repoclosure.sh
@@ -12,10 +12,4 @@ config=$1 ; shift
 TEMPDIR=$(mktemp -d)
 trap "rm -rf $TEMPDIR" EXIT
 
-repoclosure -c $config -t $* 2>&1 | tee $TEMPDIR/repoclosure.log
-
-if tail -n1 $TEMPDIR/repoclosure.log | grep -q "Num Packages"; then
-  exit 0
-else
-  exit 1
-fi
+dnf repoclosure -c $config "$@"
--- a/repoclosure/repoclosure.sh
+++ b/repoclosure/repoclosure.sh
@@ -3,13 +3,43 @@
 set -e
 
 if [ $# -lt 2 ]; then
-	echo "Usage: $0 <yum.conf>"
+	echo "Usage: $0 <dnf.conf> <releasever>"
 	exit 1
 fi
 
 config=$1 ; shift
+releasever=$1 ; shift
 
-TEMPDIR=$(mktemp -d)
+TEMPDIR=$(mktemp -d -t repoclosure-XXXXXX)
 trap "rm -rf $TEMPDIR" EXIT
 
-dnf repoclosure -c $config "$@"
+if [[ $releasever == 8 ]] ; then
+	# Enable modularity on EL8
+	MODULESDIR=$TEMPDIR/etc/dnf/modules.d
+	mkdir -p $MODULESDIR
+	cat > $MODULESDIR/ruby.module <<-EOF
+	[ruby]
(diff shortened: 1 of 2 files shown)


ekohl:
It looks like it still reads the modules.d from `/etc/dnf`, since we only specify `yum.conf`. I'll dig further.

ekohl:
This doesn't work because we use repoclosure on EL7 nodes, which don't understand modularity.

repoclosure/repoclosure.sh
```diff
@@ -13,13 +13,15 @@ shift; shift
...
 trap "rm -rf $TEMPDIR" EXIT
+MODULESDIR=$TEMPDIR/modules.d
 
 # repo names must be unique, or yum will get confused between different OSes and URLs
 reponame=undertest-$(basename $yumorig .conf)-$(echo $url | cksum | sed 's/ /-/g')
 
 yumconf=$TEMPDIR/yum.conf
 cat $yumorig > $yumconf
 cat >> $yumconf << EOF
+modulesdir=$MODULESDIR
```
ekohl, line 24: I saw this in https://github.com/rpm-software-management/libdnf/blob/eff7e688e2f1ad5d10a5ed05948c08141083f316/data/tests/modules/etc/dnf/dnf.conf.in#L6 but looking at https://github.com/rpm-software-management/libdnf/blob/dnf-4-master/libdnf/module/ModulePackageContainer.cpp it is hardcoded to `/etc/dnf/modules.d`. I think we have to switch to containers for proper modularity repoclosure.

evgeni: Looking at https://github.com/rpm-software-management/libdnf/blob/eff7e688e2f1ad5d10a5ed05948c08141083f316/libdnf/module/ModulePackageContainer.cpp#L297-L298, you should be able to pass `--installroot` to the DNF call and point it at a different "root" that has an `/etc/dnf/modules.d`

ekohl: Yep, that works. In that case you must also pass `--releasever`. However, that info is also quite useful to have anyway.

Another complication is that obal doesn't use this script anyway and has duplicated the logic.

repoclosure/repoclosure.sh
```diff
@@ -3,35 +3,43 @@
...
+	state=enabled
+	EOF
+
+	cat > $MODULESDIR/foreman.module <<-EOF
+	[foreman]
+	name=foreman
+	stream=el$releasever
+	profiles=
+	state=enabled
+	EOF
```
ehelms, line 42: I take it from this that if the foreman module was the only one in the config, it would not cause its dependencies to get enabled and thus we must hard-code the ruby and postgresql module configs in?

ekohl: Correct - this is also what `dnf module enable foreman:el8` does. If you don't, you get conflicts. I've tried, but sadly we'll have this duplication.

I'm not sure yet how we should deal with Katello, which we don't want to enable unless it should.

ekohl: It may be possible to run `dnf --installroot $TEMPDIR --releasever $releasever module enable foreman` instead.

ekohl: It isn't possible:
```console
$ dnf --installroot /tmp/repoclosure-RSRkCQ --releasever 8 module enable -y foreman
Error: This command has to be run with superuser privileges (under the root user on most systems).
```

ekohl:
Closing in favor of https://github.com/theforeman/foreman-packaging/pull/10514

PR #8245, opened 2022-07-18
send one notification if bumping failed

evgeni:
also correct the link to actually render properly

based on the example in https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-needs-context

### DIFF
--- a/.github/workflows/bump_packages.yml
+++ b/.github/workflows/bump_packages.yml
@@ -60,19 +60,6 @@ jobs:
         title: "Update ${{ matrix.package_name }} to ${{ matrix.new_version }}"
         body: ''
         delete-branch: true
-    - name: Send email to Discourse on failure
-      if: failure()
-      uses: dawidd6/action-send-mail@v3
-      with:
-        server_address: community.theforeman.org
-        server_port: 25
-        subject: "foreman-packaging update RPM packages Github Action failed"
-        to: ci@community.theforeman.org
-        from: Foreman Github Actions <github@theforeman.org>
-        body: |
-          foreman-packaging update RPM packages Github Action failed
-
-          $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID
 
   deb_list:
     name: 'Gather debs'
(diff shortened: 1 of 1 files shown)


.github/workflows/bump_packages.yml
```diff
@@ -112,16 +99,24 @@ jobs:
         title: "Update ${{ matrix.package_name }} to ${{ matrix.new_version }}"
         body: ''
         delete-branch: true
+
+  report:
+    name: Report failure to Discourse
+    needs:
+      - bump_plugin_deb
+      - bump_rpm
```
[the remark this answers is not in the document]
ekohl: This stuff is weird. The person who thought we should do flow control in YAML made the world just a little bit worse.

ekohl (approved):
Let's try this.

.github/workflows/bump_packages.yml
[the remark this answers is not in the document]
evgeni: Yes, GHA syntax and logic is completely backwards, we can agree on that.

.github/workflows/bump_packages.yml
```diff
@@ -112,16 +99,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: failure()
+    steps:
     - name: Send email to Discourse on failure
-      if: failure()
       uses: dawidd6/action-send-mail@v3
       with:
         server_address: community.theforeman.org
         server_port: 25
```
ekohl, line 115: I think it was here before, but why do we need this? Can't it route itself using DNS?

.github/workflows/bump_packages.yml
```diff
@@ -112,16 +99,24 @@ jobs:
...
-        subject: "foreman-packaging update Debian packages Github Action failed"
+        subject: "foreman-packaging update packages Github Action failed"
         to: ci@community.theforeman.org
```
ekohl, line 117: Should we store this in a secret to avoid spam? It may already be out now, but hiding it in past revisions is probably safe enough.

ekohl (commented):
I know this was here before, but I didn't get a chance to review when it was merged.

.github/workflows/bump_packages.yml
evgeni: The action defaults to using port 465 and auth. But we didn't want to create an actual user to submit this data, so we forced this to be port 25 here.

.github/workflows/bump_packages.yml
evgeni: Would then also need hiding here? https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/lib/foreman_infra.groovy#L132

(We can, I just question the usefulness)

ekohl: I suppose. Perhaps I'm just a bit cautious with publicly visible email addresses.

.github/workflows/bump_packages.yml
```diff
@@ -112,16 +99,24 @@ jobs:
...
         from: Foreman Github Actions <github@theforeman.org>
         body: |
-          foreman-packaging update Debian packages Github Action failed
+          foreman-packaging update packages Github Action failed
 
-          $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID
+          ${{ env.GITHUB_SERVER_URL }}/${{ env.GITHUB_REPOSITORY }}/actions/runs/${{ env.GITHUB_RUN_ID }}
```
ehelms, line 122: pfffft, I did read the docs before I crafted the original. Thanks for fixing!

PR #8246, opened 2022-07-18
use the correct bundle binary in plugin.mk

evgeni:
in some situations, we do not have a bundle but a bundle2.7 binary, use
that accordingly


debian/bullseye/foreman/foreman-plugin.mk
```diff
@@ -1,12 +1,15 @@
+RUBY_VERSION := $(shell ruby -e 'print RbConfig::CONFIG["RUBY_API_VERSION"]')
+BUNDLE := $(firstword $(wildcard /usr/bin/bundle$(RUBY_VERSION) /usr/bin/bundle))
```
ekohl, line 2: In the code we use `/usr/bin/foreman-ruby $(BUNDLE)`. Is that still needed after this? If so, should we combine it here in this definition?

evgeni: Those two lines are copied verbatim from `debian/rules` of foreman -- I'd keep it as is for consistency and cleanup afterwards.

evgeni: I decided to go the templating route, which seems more appropriate and leaves us with a single place that decides who and what is to be used

ekohl (approved):

PR #8252, opened 2022-07-19
use foreman postinst library for deface too

evgeni:
while deface is not a plugin as such, it is handled as one when it comes
to installation/post installation and so we can just re-use our plugin
postinstall library for it too.


plugins/ruby-foreman-deface/debian/control
```diff
@@ -8,6 +8,6 @@ Homepage: https://github.com/DefaceCommunity/deface
 
 Package: ruby-foreman-deface
 Architecture: all
-Depends: ${misc:Depends}, bundler, foreman
+Depends: ${misc:Depends}, foreman (>= 3.1.0~)
```
evgeni, line 11: 3.1 is the version introducing this helper and `foreman` already depends on `bundler` if needed.

evgeni:
> I wonder how many of the plugins actually need deface. For example, foreman_azure_rm doesn't need it and dropped it from its gemspec since 2.2.1, but it's still in the control file here.

One day we will have a way to update deps automagically.

Today is not that day ;)

PR #8254, opened 2022-07-19
Turn on auto-requires for rubygem-nokogiri

ekohl:
In the new packaging template we rely on auto-requires to generate the runtime dependencies. Turning this off but not stating the dependencies breaks at runtime.

It hacks around it by removing racc as a dependency. The explicit versions are stated so the build fails if they are ever changed. This means they shouldn't go out of sync.

Fixes: d521dd04e9f4bd0b132aad86935c1fb9b6ce81bb

### DIFF
--- a/packages/foreman/rubygem-nokogiri/rubygem-nokogiri.spec
+++ b/packages/foreman/rubygem-nokogiri/rubygem-nokogiri.spec
@@ -4,15 +4,11 @@
 
 Name: rubygem-%{gem_name}
 Version: 1.13.7
-Release: 1%{?dist}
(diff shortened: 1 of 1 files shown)


evgeni (approved):
```
% rpm -q --requires -p ~/Downloads/rubygem-nokogiri-1.13.7-2.el8.x86_64.rpm |grep rubygem
(rubygem(mini_portile2) >= 2.8.0 with rubygem(mini_portile2) < 2.9)
bundled(rubygem-racc) >= 1.4
bundled(rubygem-racc) < 2
ruby(rubygems)
```

LGTM!

PR #8255, opened 2022-07-19
migrate remaining plugins to foreman-plugin.mk

evgeni:

### DIFF
--- a/plugins/ruby-foreman-templates/debian/changelog
+++ b/plugins/ruby-foreman-templates/debian/changelog
@@ -1,4 +1,10 @@
-ruby-foreman-templates (9.3.0) stable; urgency=low
+ruby-foreman-templates (9.3.0-2) stable; urgency=low
+
+  * Use foreman-plugin.mk
+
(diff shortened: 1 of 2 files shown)


plugins/ruby-foreman-azure-rm/debian/rules
```diff
@@ -1,17 +1,9 @@
 #!/usr/bin/make -f
 
 PLUGIN = foreman_azure_rm
+PLUGIN_HAS_ASSETS = true
+PLUGIN_HAS_WEBPACK = false
+PLUGIN_HAS_PACKAGE_JSON = false
+PLUGIN_HAS_API = true
```
ekohl, line 7: This is redundant since c72f55a4ebe6f346fb69ad9e0bc5061e7810d664 but it looks like the repo wasn't entirely purged of it.

evgeni: Correct, I have it already removed in my local branch, but need https://github.com/theforeman/foreman-packaging/pull/8256 fixed first :see_no_evil:

evgeni:
okay, this should now contain only plugins that actually build fine now :D

ekohl (approved):

PR #8279, opened 2022-08-01
Update rubygem-sinatra to 2.2.2

ekohl:
Needed since https://github.com/theforeman/foreman-packaging/pull/8275 was merged.


ehelms:
Looks like this bug fix included a major update to mustermann. I am working on an update to the dependency resolution script to include smart-proxy and foreman-installer. We can update mustermann directly or wait and see if the script update catches it.

ekohl:
For now I want to move forward and I added mustermann. Looks like it now doesn't see rack-protection because we didn't mash.

PR #8320, opened 2022-08-10
Foreman 3.4 build updates

pcreech:


rel-eng/tito.props
```diff
@@ -546,9 +546,9 @@ whitelist =
   tracer
   yggdrasil
 
-[foreman-client-nightly-rhel7]
+[foreman-client-3.4-rhel7]
 disttag = .el7
-# don't whitelist foreman-release here, it's already in foreman-nightly-nonscl-rhel7
+# don't whitelist foreman-release here, it's already in foreman-3.4-nonscl-rhel7
```
evgeni, line 551: now that you say it, there is no 3.4-nonsc-7, so you NEED to whitelist it here!

same should happen for nightly, I geuss

repoclosure/yum.conf
```diff
@@ -153,6 +153,6 @@ failovermethod=priority
 name=f29-puppet-6
 baseurl=http://yum.puppetlabs.com/puppet6/fedora/29/x86_64/
 
-[f29-foreman-client-nightly]
-name=Foreman Client nightly F29
-baseurl=http://koji.katello.org/releases/yum/foreman-client-nightly/fc29/x86_64/
+[f29-foreman-client-3.4]
+name=Foreman Client 3.4 F29
+baseurl=http://koji.katello.org/releases/yum/foreman-client-3.4/fc29/x86_64/
```
evgeni, line 158: there is no f29, just zuul

evgeni (changes requested):
doop doop doop

repoclosure/yum.conf
pcreech: f29 is gone

rel-eng/tito.props
pcreech: done

PR #8348, opened 2022-08-15
Skip nightly packages in listing updatable pkgs

ekohl:
Fixes: df4b891110357252746f91afe8f150b4ef7f0631


evgeni:
We already do this?

https://github.com/theforeman/foreman-packaging/blob/61cdf829ea481294d8d00dc6162e3524875ebb2d/list_updatable_packages#L77

ekohl:
That is `find_specs`, which finds specs in the repo to update. Currently that only supports plugins. The others read a lockfile and update from there.

evgeni:
I was about to say "but why is Katello in the 'broken-set' right now, it is a plugin?!", but of course, katello's gemfile.lock contains… katello.

evgeni (approved):

PR #8349, opened 2022-08-15
Fixes #35366 - make foreman-protector work as non-root

evgeni:
deploy code and config as readable by mere mortals, there are no secrets in there


ehelms (approved):

upadhyeammit (approved):
Makes sense, thank you!

PR #8351, opened 2022-08-15
Warn about conflicts in updating spec files

ekohl:
This rewrites the code to deal with lock files in order to handle all sources equally. This allows merging them and warning about conflicts. Currently that's:

```
Found conflict for rubygem-jwt: 2.2.3 != 2.4.1
Found conflict for rubygem-rack-test: 1.1.0 != 2.0.2
Found conflict for rubygem-unicode-display_width: 1.8.0 != 1.6.1
```

### DIFF
--- a/list_updatable_packages
+++ b/list_updatable_packages
@@ -22,7 +22,7 @@ class Spec:
         self.path = path
         spec = path.read_text()
         self.lines = spec.splitlines()
-        self._globals = None
+        self._globals: Optional[dict] = None
 
     @property
     def directory(self) -> str:
@@ -61,7 +61,7 @@ class Spec:
         return 'prereleasesource' in self.globals
 
     @property
-    def gem_name(self) -> str:
+    def gem_name(self) -> Optional[str]:
         return self.globals.get('gem_name')
 
     @property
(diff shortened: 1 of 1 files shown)


ekohl:
> Found conflict for rubygem-rack-test: 1.1.0 != 2.0.2

This is a test dependency for Foreman and Foreman Proxy, but a runtime dependency for `rubygem-actionpack`. In https://github.com/theforeman/foreman/commit/a439e1e92e9f98f3dbc8ed2f204bd53cc861565e it was pinned to `< 2`. I wonder if that should actually be a runtime pin.

list_updatable_packages
```diff
@@ -117,15 +130,33 @@ def latest_version(spec: Spec) -> str:
...
-def build_plugin_matrix(specs: Iterable[Spec]) -> Generator[dict, None, None]:
-    if DEBUG: print('Building plugin matrix')
-    for spec in specs:
+def merge_specs(specs: Iterable[Tuple[Spec, Optional[str]]]) -> dict[Spec, Optional[str]]:
+    """
+    Merge specs
+
+    This can process the same spec file multiple times. If a spec is already listed but without a
+    version, it will take the more specific constraint. The last entry is used if a constraint was
+    already given while a warning is printed.
```
evgeni, line 139: this ("last entry is used") sounds dangerous as it will potentially update packaging to an incompatible version, while we don't notice the warning?

I'd prefer not touching a package if a conflict was found.

ekohl: I didn't think about that, but skipping it is a good one.

ekohl: I now set the version to `CONFLICT`. Then I expect `./bump_rpm.sh` will fail in the individual job, causing the whole job to fail so we'll find it.

evgeni: smart.

now I wonder if one can upload arbitrary version strings to rubygems, but I think one cannot ;)

evgeni (approved):

PR #8360, opened 2022-08-17
Add a script to remove packages & clean old cruft from add_*

ekohl:


remove_from_comps
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env ruby
+require 'nokogiri'
+
+def print_usage
+  usage = <<-USAGE
+add_to_comps - adds packagereqs for a package to a comps xml file
```
evgeni, line 6: no? ;-)

remove_from_comps
```diff
@@ -0,0 +1,36 @@
...
+
+Usage #{$0} comps-file package-name [package-name]
+
+Examples:
+  #{$0} comps/comps-foreman-el8.xml nodejs-example
+  #{$0} comps/comps-foreman-el8.xml nodejs-example nodejs-other-example
+
+Return codes:
+
+  0 - successfully added
```
evgeni, line 16: no :)

evgeni (approved):
doc-nitpick, otherwise reads correct to me

remove_package
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
evgeni, line 1: scope creep: does this script have sufficient knowledge to generate the `koji` command for us?

ekohl: I think it should be done in 2 steps. This is to create a PR. Once it's merged, it does indeed need to be cleaned.

What I do today is run `./rel-eng/compare-with-tags` and let that tell me which packages are unexpected and modify that.

Right now it outputs:
````
# Packages not expected in foreman-plugins-nightly-el8
 * [x] yggdrasil
```shell
koji remove-pkg foreman-plugins-nightly-el8 yggdrasil
```

# Packages missing in tag foreman-client-nightly-el9
 * [x] jsoncpp
 * [x] qpid-proton
```shell
koji add-pkg --owner=kojiadm foreman-client-nightly-el9 jsoncpp qpid-proton
```

 * [x] gofer
```shell
```
````

Ideally we would clean this up but I have no idea what the status is for those packages.

Right now I run `koji untag-build --all && koji-block-pkg && koji remove-pkg` but I feel that's redundant. Is `koji remove-pkg` sufficient? If so, the script you're asking for already exists.

evgeni: perfect, I totally forgot about the compare thing

PR #8366, opened 2022-08-18
Update rubygem-foreman_vault to 1.2.0

laugmanuel:

PR #8367, opened 2022-08-18
Update rubygem-foreman_vault to 1.2.0 (#8366)

laugmanuel:
(cherry picked from commit 7fbafda627c84c804865da61fb38a7cf49e7e27c)


dgoetz (approved):
Merging this so the fix is in all supported versions.

PR #8370, opened 2022-08-18
explicitly vendor sinatra

evgeni:
rack-protection and sinatra have to be the same version
foreman pulls in rack-protection via sidekiq, f-tasks needs sinatra
let foreman.deb provide both gems in the cache, so that they always match

### DIFF
--- a/debian/bullseye/foreman/changelog
+++ b/debian/bullseye/foreman/changelog
@@ -1,3 +1,9 @@
+foreman (3.5.0-2) stable; urgency=low
+
+  * explicitly vendor sinatra
+
+ -- Evgeni Golov <evgeni@debian.org>  Thu, 18 Aug 2022 14:14:39 +0200
+
 foreman (3.5.0-1) stable; urgency=low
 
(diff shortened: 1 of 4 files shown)


ekohl (commented):
The alternative to this is create a package for it, similar to how we have `ruby-foreman-deface` which places the files in the cache. It's more overhead to maintain, but leads to a smaller deb. I'd be ok with either solution, just want to make sure you considered this.

evgeni:
> The alternative to this is create a package for it, similar to how we have `ruby-foreman-deface` which places the files in the cache. It's more overhead to maintain, but leads to a smaller deb. I'd be ok with either solution, just want to make sure you considered this.

Then Foreman needs to depend on it too, and the way it's currently built we just let Foreman dictate the exact version and done :)

(read: considered and ignored ;))

ekohl (approved):
You could have `ruby-foreman-rack-protection` and `ruby-foreman-sinatra`, but that may be too much overhead.

PR #8378, opened 2022-08-23
Refs #35409 - Include legacy assets in plugin macros

evgeni:


ekohl (approved):
We are sure Apache is allowed to read the symlink (Option +FollowSymlinks becomes mandatory) and gem files, right? If not, we should copy like the Debian plugins do.

Edit: I'd also refer to this as the sprockets based assets. Legacy is always a bit ambiguous over time.

evgeni:
Reworded!

We're getting a default of `FollowSymlinks` via https://github.com/puppetlabs/puppetlabs-apache/blob/92c47e144fc07a98b706a00d9dc41654b681794b/manifests/params.pp#L32-L33

ehelms (approved):

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -169,6 +169,7 @@ find %{buildroot}%{gem_instdir}/<%= spec.bindir %> -type f | xargs chmod a+x
 %{foreman_bundlerd_plugin}
 <% if has_assets -%>
 %{foreman_assets_plugin}
+%{foreman_assets_foreman}
```
ekohl, line 172: Won't this mess up cherry picks? Should we somehow require a certain foreman version or make the macro optional?

evgeni: The same applies to all "remove SCL code" changes we did recently, and we didn't bump stuff there?

evgeni: We could set that macro to nil in 3.4?
Non SCLed packages are not pickable further anyways

ekohl: I'd also be happy to pick the change back to 3.4. Then the risk is that the build fails with an unpackaged file, which is fine because CI shows it. Existing packages don't need a rebuild, but would benefit from it.

evgeni: That also works!

evgeni: Does this clear your concern about pickability?

ekohl: Yes it does. There's just a merge conflict now.

evgeni: How dare you!? ;-)
(rebased)

ekohl (approved):
Tests are running, but :+1:

PR #8380, opened 2022-08-23
Update rubygem-net-ssh to 7.0.1

ekohl:


packages/foreman/rubygem-net-ssh/rubygem-net-ssh.spec
```diff
@@ -1,93 +1,85 @@
...
+# %%gem_install compiles any C extensions and installs the gem into ./%%gem_dir
+# by default, so that we can move it into the buildroot in %%install
+%gem_install
 
 %install
 mkdir -p %{buildroot}%{gem_dir}
 cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
-# file-not-utf8 correction
```
ekohl, line 54: I'm not sure if this is still needed. I didn't check, but it's now in the -doc package.

packages/foreman/rubygem-net-ssh/rubygem-net-ssh.spec
```diff
@@ -1,93 +1,85 @@
...
-pushd %{buildroot}%{gem_instdir}
-iconv --from=ISO-8859-1 --to=UTF-8 THANKS.txt > THANKS.txt.new && \
-touch -r THANKS.txt THANKS.txt.new && \
-mv THANKS.txt.new THANKS.txt
-popd
-
-# replace shebangs to prevent SCL packages depending on non-SCL Ruby
```
ekohl, line 61: I instead excluded the support script.

packages/foreman/rubygem-net-ssh/rubygem-net-ssh.spec
```diff
@@ -1,93 +1,85 @@
...
+%exclude %{gem_instdir}/Dockerfile
+%exclude %{gem_instdir}/Dockerfile.openssl3
+%exclude %{gem_instdir}/ISSUE_TEMPLATE.md
+%license %{gem_instdir}/LICENSE.txt
+%exclude %{gem_instdir}/Manifest
+%exclude %{gem_instdir}/appveyor.yml
+%exclude %{gem_instdir}/docker-compose.yml
 %{gem_libdir}
-%doc %{gem_instdir}/LICENSE.txt
+%exclude %{gem_instdir}/net-ssh-public_cert.pem
```
ekohl, line 65: This is the cert that signed the gem. We sign the RPM so this isn't needed. If anything, we should check it while building the RPM.

PR #8383, opened 2022-08-24
Refs #35409 - Include legacy assets in foreman_remote_execution

evgeni:


evgeni:
With this rebuild, visiting https://foreman.example.com/job_templates/new does not complain about inaccesible https://foreman.example.com/assets/foreman_remote_execution/template_input-HASH.js anymore, when `/assets` is served by apache :fireworks:

ekohl (approved):

PR #8385, opened 2022-08-24
drop rubygem-foreman_host_reports

evgeni:
it's broken and unmaintained right now


evgeni:
:eyes: @lzap

lzap (approved):
Thanks I should have done than when we decided not to continue development. It was a bit clear if we are going to pick it up at some point, but I would say it's not the case. @ares ?

evgeni:
If you ever pick it back up, we reintroduce the package. Right now it's just harmful to users to be able to install it, as it is incompatible with current Foreman.

ekohl (commented):
Is there a removal procedure? In the past with other places we've used a rake task to purge it, but I don't think we have one.

evgeni:
> Is there a removal procedure? In the past with other places we've used a rake task to purge it, but I don't think we have one.

the discourse post has one: https://community.theforeman.org/t/migrating-reports-to-the-new-format-in-foreman-3-2/25846

ekohl (approved):
I'd be OK with this. Do we want to do this for 3.4 too? Should we include an upgrade warning, even though it's a plugin?

evgeni:
> I'd be OK with this. Do we want to do this for 3.4 too? Should we include an upgrade warning, even though it's a plugin?

Yes, I'll open a pick for 3.4.

Nah, it should be mostly unused (and is incompatible with "recent" Foreman versions anyways).

evgeni:
```console
% koji untag-build --all foreman-plugins-nightly-el8 rubygem-foreman_host_reports rubygem-hammer_cli_foreman_host_reports rubygem-smart_proxy_reports
% koji block-pkg foreman-plugins-nightly-el8 rubygem-foreman_host_reports rubygem-hammer_cli_foreman_host_reports rubygem-smart_proxy_reports
```

PR #8391, opened 2022-08-25
Refs #35409 - Include sprockets assets in rubygem-foreman_chef

evgeni:


ekohl:
I opened https://github.com/theforeman/foreman_chef/pull/99 to trigger CI, but there is no CI ...

evgeni:
```
NameError: undefined local variable or method `belongs_to_host' for ForemanChef::CachedRunList(Table doesn't exist):Class
```

evgeni:
I think https://github.com/theforeman/foreman_chef/pull/100 should fix it, but yeah, Ci would be cool.

evgeni:
as far as I could gather, nobody is developing or using the Chef plugin anymore, I am going to close this PR and see that the plugin is instead removed completely.

PR #8396, opened 2022-08-25
Refs #35429 - Update rubygem-oauth to 1.0.0

ekohl:


packages/foreman/rubygem-oauth/rubygem-oauth.spec
```diff
@@ -2,16 +2,16 @@
...
 Summary: OAuth Core Ruby implementation
 License: MIT
 URL: https://github.com/oauth-xx/oauth-ruby
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
 # start specfile generated dependencies
-Requires: ruby >= 2.0
-BuildRequires: ruby >= 2.0
+Requires: ruby >= 2.7
+BuildRequires: ruby >= 2.7
```
evgeni, line 14: I think you'll have to package `version_gem` first: https://github.com/oauth-xx/oauth-ruby/blob/main/oauth.gemspec#L6

evgeni (approved):

PR #8398, opened 2022-08-25
Update rubygem-pulpcore_client to 3.18.8

sjha4:


ekohl (changes requested):
The gemspec patching in `%prep` currently fails the build. That should be removed now that it's properly fixed.

ekohl (approved):
Thanks!

PR #8401, opened 2022-08-25
Upgrade pulp client gems

sjha4:


sjha4:
@ekohl : Does this looks good? I have a separate PR for pulpcore_client https://github.com/theforeman/foreman-packaging/pull/8398 and this one has all the pulp plugins except ostree.

ekohl (changes requested):
Thanks for submitting these. I was about to.

All the gems you updated here have patches to change the faraday dependency (in `%prep`). Those are now redundant and they fail because it no longer matches. Those need to be removed to get the builds to pass.

It'd also be nice if the changelog entry had a correct name.

ekohl (approved):
This is indeed what I meant, but I prefer to keep changes in single commits. I've squashed the last commit into the various other commits so they can be cherry picked individually. I'll just wait for CI to complete on this.

PR #8413, opened 2022-08-27
Use rich dependencies when updating dependencies

ekohl:
Now that EL 7 has been dropped, rich dependencies can be used. This not only makes specs easier to read, they are also more accurate.

Also adds a commit that updates foreman, foreman-proxy and rubygem-katello. The SCL compatibility is also dropped since the new dependencies don't have SCL macros either. Partial SCL compatibility can only lead to confusion and bugs.

This is still a draft since the gem2rpm template doesn't generate rich dependencies yet.

### DIFF
--- a/get-gemfile-deps
+++ b/get-gemfile-deps
@@ -10,8 +10,29 @@ EXCLUDED_REQUIRES = {
   'Requires' => ['activerecord-nulldb-adapter']
 }
 
-def package(gem)
-  return "rubygem(#{gem})"
+def build_requirement(dependency)
+  name = "rubygem(#{dependency.name})"
+
+  requirements = []
+
+  dependency.requirement.requirements.each do |op, version|
+    if op == '~>'
+      requirements << "#{name} >= #{version}" << "#{name} < #{version.bump}.0"
+    elsif op == '>=' && version.to_s == '0'
+      # handled later
+    elsif op != '!=' # != is not supported in RPM
+      requirements << "#{name} #{op} #{version}"
+    end
+  end
+
+  case requirements.length
+  when 0
+    name
+  when 1
+    requirements.first
+  else
+    "(#{requirements.join(' with ')})"
+  end
 end
 
 unless File.file?(filename)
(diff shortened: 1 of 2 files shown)


ekohl:
And this fails badly because our builder is still EL7 which doesn't understand rich dependencies. That means rpmlint fails.

ekohl:
I've rebased this now that we have an EL8 builder. Let's see if this builds.

packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -40,8 +40,8 @@ Requires:       rubygem(logging) < 3.0.0
 Requires:       rubygem(sinatra)
 
 # start specfile default Requires
-Requires: rubygem(concurrent-ruby) >= 1.0
-Requires: rubygem(concurrent-ruby) < 2.0
+Requires: rubygem(smart_proxy)
```
ekohl, line 43: Hmm, not sure where this came from but this is not intended. Looks like the script needs some tweaking :(

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -41,7 +41,6 @@ BuildRequires: rubygem(foreman-tasks) >= 5.0
 BuildRequires: rubygem(foreman_remote_execution) >= 7.1.0
 BuildRequires: rubygem(dynflow) >= 1.6.1
 BuildRequires: rubygem(activerecord-import)
-BuildRequires: rubygem(qpid_proton)
```
ekohl, line 44: I believe this is one we should also include in 4.10.

ekohl:
[test rpm]

packages/foreman/foreman/foreman.spec
```diff
@@ -44,83 +44,50 @@ Requires(preun): systemd-units
...
-Requires: rubygem(rack-cors) < 2.0
-Requires: rubygem(jwt) >= 2.2.2
-Requires: rubygem(jwt) < 3.0
-Requires: rubygem(graphql) >= 1.13.0
-Requires: rubygem(graphql) < 1.14.0
+Requires: (rubygem(rack-cors) >= 1.1 with rubygem(rack-cors) < 2.0)
+Requires: (rubygem(jwt) >= 2.2.2 with rubygem(jwt) < 3.0)
+Requires: (rubygem(graphql) >= 1.13.0 with rubygem(graphql) < 1.14.0)
 Requires: rubygem(graphql-batch)
+Requires: rubygem(rss)
```
ekohl, line 90: This is from my local Gemfile and breaks things.
```
DEBUG util.py:634:  No matching package to install: 'rubygem(rss)'
DEBUG util.py:634:  Not all dependencies satisfied
DEBUG util.py:634:  Error: Some packages could not be found.
```

ekohl:
Updated to only the template. Individual package update create too many conflicts so I need to keep rebasing.

@evgeni can we merge this now?

evgeni (approved):

PR #8414, opened 2022-08-27
Showcase dynamic build requires

ekohl:
Since Fedora 31 it's possible to use [Dynamic BuildRequires](https://fedoraproject.org/wiki/Changes/DynamicBuildRequires). This removes the need to change the spec file all the time to keep in sync with the Foreman repo itself.

Once EL 8 support is dropped, this can be used. Since that's far in the future, I haven't done the same for NPM yet but obviously that's possible too.

In a proper implementation we would have tooling that's usable in all package, like a foreman-packaging RPM with reusable scripts and macros.


ekohl:
Rebased to resolve a conflict. Since we now have EL 9 builds, it should be possible to verify it works with COPR. Until we drop EL 8, we can't avoid commits like 212285ea5a2416849f67a2d684133cfb1c9f3290 but at least with packit we'll have correct build requires for EL 9.

evgeni:
> I don't quite understand why this is failing.

Because nodejs-packaging is in a module, and we don't enable that in closure.
But also, because it's looking at the SRPM, and we actually don't care about SRPM closure in prod.

packages/foreman/foreman/foreman.spec
```diff
@@ -391,6 +399,7 @@ Meta package with support for building RPMs in the Foreman release cycle.
 
 %files build
 %{_sysconfdir}/rpm/macros.%{name}-dist
+# TODO: ship SOURCE1
```
evgeni, line 402: do we have to? it's the only project that uses a `Gemfile`, others would need `gemspec` support.

packages/foreman/foreman/foreman.spec
```diff
@@ -210,7 +214,11 @@ BuildRequires: (rubygem(coffee-rails) >= 5.0.0 with rubygem(coffee-rails) < 5.1.
 # start specfile facter BuildRequires
 BuildRequires: rubygem(facter)
 # end specfile facter BuildRequires
-
+%else
+%generate_buildrequires
+# TODO: Generate npm
+%{SOURCE1}
```
evgeni, line 220: This syntax is super weird, but I guess it's not something we have influence on.
(For others: you start the `%generate_buildrequires` section(macro? whatever) and then every line that is printed on stdout becomes a `BuildRequires: <the line>`.)

Maybe
```suggestion
%generate_buildrequires
# TODO: Generate npm
./%{SOURCE1}
```
so its more obvious we're executing a script here?

evgeni: Indeed
```
Executing(%generate_buildrequires): /bin/sh -e /var/tmp/rpm-tmp.edESHh
+ umask 022
+ cd /builddir/build/BUILD
+ cd foreman-3.11.0-develop
+ /builddir/build/SOURCES/gen-gem-buildreqs
+ RPM_EC=0
++ jobs -p
+ exit 0
Wrote: /builddir/build/SRPMS/foreman-3.11.0-0.3.develop.el9.buildreqs.nosrc.rpm
```

packages/foreman/foreman/gen-gem-buildreqs
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env ruby
+require 'bundler'
```
evgeni, line 2: This needs `rubygem(bundler)`, but I don't think it's part of our `BuildRequires` today?

```
BuildRequires: rubygem(bigdecimal)
BuildRequires: /usr/bin/ruby
BuildRequires: rubygems
BuildRequires: rubygem(rake) >= 0.8.3
BuildRequires: rubygem(bundler_ext)
```

(rubygems probably pulls it in, or bundler_ext, but I prefer being explicit)

ekohl: I think it's a good point that we become explicit about build requires. I also considered expanding the script with generating a correct Ruby requirement.

Also, I've been wanting to get rid of `bundler_ext`. Its point was to remove the bundler version requirements, but these days we convert them literally. With this it's even more.

ekohl: Added as an explicit build requires.

ekohl:
I'd like to move forward with this, even if it's incomplete.

packages/foreman/foreman/foreman.spec
```diff
@@ -107,61 +108,15 @@ BuildRequires: rubygem(bigdecimal)
 BuildRequires: gettext
 BuildRequires: /usr/bin/ruby
 BuildRequires: rubygems
+BuildRequires: rubygems(bundler)
```
evgeni, line 111:
```suggestion
BuildRequires: rubygem(bundler)
```

packages/foreman/foreman/foreman.spec
```diff
@@ -107,61 +108,15 @@ BuildRequires: rubygem(bigdecimal)
...
 BuildRequires: rubygems
+BuildRequires: rubygem(bundler)
 BuildRequires: rubygem(rake) >= 0.8.3
 BuildRequires: rubygem(bundler_ext)
 BuildRequires: /usr/bin/npx
 BuildRequires: make
-BuildRequires: (rubygem(rss) or ruby-default-gems < 3.0)
 Requires: (rubygem(rss) or ruby-default-gems < 3.0)
 BuildRequires: (rubygem(rexml) or ruby-default-gems < 3.0)
 Requires: (rubygem(rexml) or ruby-default-gems < 3.0)
```
evgeni, line 118: Why do we need rexml, https://github.com/theforeman/foreman-packaging/pull/10393 does not say it :crying_cat_face:

packages/foreman/foreman/foreman.spec
```diff
@@ -198,19 +153,9 @@ BuildRequires: (npm(os-browserify) >= 0.3.0 with npm(os-browserify) < 1.0.0)
...
-BuildRequires: (rubygem(execjs) >= 1.4.0 with rubygem(execjs) < 3.0)
-BuildRequires: (rubygem(terser) >= 1.1 with rubygem(terser) < 2.0)
-BuildRequires: (rubygem(sass-rails) >= 6.0 with rubygem(sass-rails) < 7.0)
-# end specfile assets BuildRequires
-
-# start specfile facter BuildRequires
-BuildRequires: rubygem(facter)
-# end specfile facter BuildRequires
-
+%generate_buildrequires
```
evgeni, line 156: This can also be used on normal gems, right?
The script is only needed because core is not a gem?

packages/foreman/foreman/foreman.spec
ekohl: It also confused me

packages/foreman/foreman/foreman.spec
ekohl: There is no generic implementation for gems because there's no good way to determine what's needed. But we can implement this ourselves.

I'd like to have something for Foreman plugins as well but thought to first focus on Foreman itself since I think it's where we get the most benefits.

Also converting nodejs is a good next step but I thought it was already complex enough now

evgeni (approved):

ekohl:
Rebased to resolve the merge conflict.

I suspect we need to update Foreman's packit config so I opened https://github.com/theforeman/foreman/pull/10592.

@pcreech @zjhuntin can you also take a look if this will cause problems downstream?

Odilhao:
I've been looking into Dynamic Build Requires since 2023 when we build the PEP-517 packages for pulp related stuff.

My concern since them is the fact that we don't control most of the packages, we already have rich dependencies for packages. Can we limit the scope of dynamic build in the beginning for just packages that we own on Foreman? 

On the downstream side of it, since we are already on EL9 I don't see any issue with this, all the required packages are already present in the buildroot.

ekohl:
> My concern since them is the fact that we don't control most of the packages, we already have rich dependencies for packages. Can we limit the scope of dynamic build in the beginning for just packages that we own on Foreman?

This is **just** on `foreman.spec` and even then only the gems, no other packages. In the future we can expand it to Foreman plugins too. Those are the ones we absolutely own.

I'd also like to expand it to runtime gem dependencies and cover NPM as well.

> On the downstream side of it, since we are already on EL9 I don't see any issue with this, all the required packages are already present in the buildroot.

If there are no downstream concerns and there is a prior approval from @evgeni then I'd like to ask you to merge it so we can see how it works in practice.

ShimShtein:
I suppose the next step will be to create a similar script for the npm packages too?

ekohl:
Rebased.

> I suppose the next step will be to create a similar script for the npm packages too?

Yes, I wrote that in https://github.com/theforeman/foreman-packaging/pull/8414#issuecomment-3032439633:

> I'd also like to expand it to runtime gem dependencies and cover NPM as well.

PR #8422, opened 2022-08-29
Refs #34647 - Update rubygem-apipie-rails to 0.8.1

ekohl:


ehelms (approved):
Once the spec file is fixed to deal with the Gemfile not found

PR #8429, opened 2022-08-30
Update Pulp gems for 3.4

ekohl:
These versions allow Faraday < 2 instead of < 1.9, which is needed to properly package foreman_google.


ekohl:
I realize we haven't ran a pipeline with this yet. https://ci.theforeman.org/job/katello-nightly-rpm-pipeline/1421/ is still running.

ekohl:
It was green, so all good.

PR #8431, opened 2022-08-30
Remove jsoncpp and qpid-proton from f-client-el9

ekohl:
These packages are allowed to by built in koji so this change aligns the git repo with koji.

It still leaves 2 issues I'm unsure how to deal with:

# Packages not expected in foreman-plugins-nightly-el8
 * [x] yggdrasil
```shell
koji remove-pkg foreman-plugins-nightly-el8 yggdrasil
```

This package is also in the client tools. Is it intentional that it's in both plugins and client?

 * [x] gofer
```shell
```

I don't know the status of this either.

### DIFF
--- a/rel-eng/tito.props
+++ b/rel-eng/tito.props
@@ -507,11 +507,9 @@ whitelist =
   foreman-release
   foreman_ygg_worker
   gofer
-  jsoncpp
   katello-host-tools
   katello-pull-transport-migrate
   python-psutil
-  qpid-proton
   rubygem-foreman_scap_client
   tracer
   yggdrasil


ekohl:
@ehelms do you happen to know what's up with yggdrasil and gofer?

ehelms:
I don't know why it is reporting those, I do not get those same results when using the Obal functionality. Perhaps we have a discrepancy if this was calculated off of the tito data?

ekohl:
> I don't know why it is reporting those, I do not get those same results when using the Obal functionality. Perhaps we have a discrepancy if this was calculated off of the tito data?

That can be explained by the fact that right now gofer is not in the tito whitelist, but the koji tags in `package_manifest.yaml` do say it should be built. If we look at http://koji.katello.org/koji/buildinfo?buildID=39828 it looks like it was originally built for `pulp-nightly-rhel7`.

It was added for EL8 in e6aa20bb74bdf2abfee4146c9ca254029611f1f8 so perhaps someone did some manual work? I get the impression that gofer should actually be there and it was an omission. It is also in comps.

Then for yggdrasil: http://koji.katello.org/koji/packageinfo?packageID=2854. It is in foreman-plugins. Essentially all tags. That looks like someone manually messed up. However, in 097f17b3900d393eeff7bdffdf90a90da4efc502 it was added to comps. So now I do wonder how it's supposed to work. Should it be released and tagged into both plugins and client?

ehelms:
> It was added for EL8 in [e6aa20b](https://github.com/theforeman/foreman-packaging/commit/e6aa20bb74bdf2abfee4146c9ca254029611f1f8) so perhaps someone did some manual work? I get the impression that gofer should actually be there and it was an omission. It is also in comps.

Gofer used to be built into Pulp tags as it came from them, and since it receives so few updates it was likely tagged over as that was easy and the right build existed and it has since needed no updates.



> Then for yggdrasil: http://koji.katello.org/koji/packageinfo?packageID=2854. It is in foreman-plugins. Essentially all tags. That looks like someone manually messed up. However, in [097f17b](https://github.com/theforeman/foreman-packaging/commit/097f17b3900d393eeff7bdffdf90a90da4efc502) it was added to comps. So now I do wonder how it's supposed to work. Should it be released and tagged into both plugins and client?

yggdrasil is required in the client and plugin repositories

PR #8432, opened 2022-08-30
don't try to install extras/noVNC

evgeni:


ekohl:
Bleh, the one time I think "I might as well merge it" ...

PR #8424, opened 2022-08-29
Refs #29539 - Unbundle websockify

ekohl:
This reflects the removal of a bundled websockify in favor of a system package. While it is in EPEL8 (and the spec is copied from that), Foreman doesn't support running with EPEL8 enabled.

Draft until the core PR is merged: https://github.com/theforeman/foreman/pull/9392

PR #8434, opened 2022-08-31
Add ansible-runner 2+ package

ehelms:


packages/plugins/ansible-runner/ansible-runner.spec
```diff
@@ -0,0 +1,110 @@
...
+
+License:        ASL 2.0
+URL:            https://github.com/ansible/ansible-runner
+Source0:        https://github.com/ansible/%{pypi_name}/archive/%{version}/%{pypi_name}-%{version}.tar.gz
+
+BuildArch:      noarch
+
+BuildRequires: python%{python3_pkgversion}-devel
+BuildRequires: python%{python3_pkgversion}-setuptools
+BuildRequires: python%{python3_pkgversion}-pbr
```
ehelms, line 16: This is the only trouble some build requires from my local testing. Pulpcore packaging carries this so I think we have two options:

 * Carry python-pbr in foreman-packaging
 * Tag over python-pbr from pulpcore-packaging tags into the plugin tags for build time

ehelms:
@adamruzicka for your awareness, I tested a locally built version of this and did not encounter any errors

I think this change also means we would drop all of https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/plugin/ansible/runner.pp#L11

packages/plugins/ansible-runner/ansible-runner.spec
evgeni: Which python version should this runner be built with? 3.6? 3.8? 3.9?

pulpcore uses 3.9 right now

adamruzicka:
I've been running ansible-runner 2.1.3 with python 3.10 locally for some time and haven't encountered any issues.

packages/plugins/ansible-runner/ansible-runner.spec
ekohl: We recently built websockify (https://github.com/theforeman/foreman-packaging/pull/8424). It feels to me that we should use the same Python version in multiple places. I don't strongly care, but it'd be good to reduce the number of Python versions required.
* 3.6: it's the default and others (like EPEL) also build against it, but we don't support EPEL
* 3.8: Ansible (at least in EPEL) is built against it
* 3.9: Katello users have this because of Pulpcore

That doesn't help narrow it down, does it?

dgoetz: It does not as the modules are python36, python38 and python39, so parallel usage would be possible. But I would still prefer having the same version so if some dependencies are needed by both, it would not be installed twice for different versions.

ekohl: I know they're parallel installable, but having fewer packages installed is a good thing IMHO.

evgeni: (This is the reason I deleted half of my comment above and just pointed out the version pulpcore uses :hand_over_mouth:)

Ansible uses 3.8 in EL8 (both EPEL and the one in AppStream), so as long we're using that Ansible build, I think we also should use the "matching" Python for the rest of the Ansible ecosystem (part of the hacks we have today are because we're using a Runner on 3.6 with an Ansible on 3.8).

ekohl: Pretty good argument for 3.8. Does that mean we need to start building all our (Python) packages against 3.8? Is this something we configure in our buildroot in Koji?

ehelms: Step 1 for 3.8 -- https://github.com/theforeman/tool_belt/pull/478

ehelms:
[test rpm]

PR #8437, opened 2022-08-31
Fixes #35461 - Require /usr/sbin/sendmail to be available

evgeni:
Foreman uses sendmail as the default mail handler, but modern systems
might not have it available by default.

Only depend on the path, not an explicit package, as there might be
different packages providing a sendmail binary (e.g. on EL8 there are at
least Postfix and Sendmail doing so).

(cherry picked from commit 0869b371150b168693470a9cc96a038ed657d8db)


ekohl (approved):
Tests are still running, but :+1:

evgeni:
repoclosure unrelated :see_no_evil:

PR #8439, opened 2022-08-31
Add python pbr

ehelms:


ehelms:
[test]

ehelms:
[test rpm]

ehelms:
[test rpm]

ehelms:
This is now building against Python 3.9 and passing.

evgeni:
The same package is also built in `pulpcore-packaging` and due to same NEVR we'll have to cross-tag it over. Is that how we want to handle that? Not against it, just wanted to point out that this will require syncing between the repos every now and then.

ehelms:
> The same package is also built in `pulpcore-packaging` and due to same NEVR we'll have to cross-tag it over. Is that how we want to handle that? Not against it, just wanted to point out that this will require syncing between the repos every now and then.

That works for me, I am not sure how else to handle it given this is a limitation of the build system.

PR #8441, opened 2022-08-31
Add python-docutils for ansible-runner

ehelms:

### DIFF
--- a/packages/plugins/python-docutils/python-docutils.spec
+++ b/packages/plugins/python-docutils/python-docutils.spec
@@ -1,0 +1,80 @@
+%{?scl:%scl_package python-%{pypi_name}}
+%{!?scl:%global pkg_name %{name}}
+
+# Created by pyp2rpm-3.3.8
+%global pypi_name docutils
+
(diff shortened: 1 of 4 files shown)


evgeni:
[test rpm]

packages/plugins/python-docutils/python-docutils.spec
```diff
@@ -0,0 +1,80 @@
...
+%{_bindir}/rst2html5.py
+%{_bindir}/rst2latex.py
+%{_bindir}/rst2man.py
+%{_bindir}/rst2odt.py
+%{_bindir}/rst2odt_prepstyles.py
+%{_bindir}/rst2pseudoxml.py
+%{_bindir}/rst2s5.py
+%{_bindir}/rst2xetex.py
+%{_bindir}/rst2xml.py
+%{_bindir}/rstpep2html.py
```
evgeni, line 73: this will (possibly) clash with the binaries from `python3-docutils` that comes from the BaseOS, and as `python-daemon` doesn't use any of these, I'd prefer to `%exclude them.

PR #8446, opened 2022-09-01
Bump rubygem-foreman_rh_cloud to 6.0.42-1

ShimShtein:
New version of foreman_rh_cloud and changed the spec according to the new format


packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -1,60 +1,44 @@
...
+BuildRequires: npm(@babel/core) >= 7.7.0
+BuildRequires: npm(@babel/core) < 7.8.0
+BuildRequires: npm(@theforeman/builder) >= 8.16.0
+BuildRequires: npm(jed) >= 1.1.1
+BuildRequires: npm(jed) < 1.2.0
 # end package.json devDependencies BuildRequires
 
+# additional dependencies
+BuildRequires: npm(@redhat-cloud-services/frontend-components) >= 2.5.0
+BuildRequires: npm(@redhat-cloud-services/frontend-components) < 3.0.0
```
ekohl, line 39: If these are needed, why aren't they in `package.json`? It now looks like you're installing them in CI after you run `npm install`, but isn't that missing the point of a package manager? I'd consider this a bug.

ShimShtein: They are served by foreman-js, but for some reason we had to install them directly.
@ron-lavi do you know what's the status here for the latest Foreman?

ekohl: Given the build failed on it, perhaps the lower bound of foreman-js is not good enough? It may depend on https://github.com/theforeman/foreman-js/commit/8fe8afedecb8f02d8e1c61b1c485d92fda34bae3 which is part of `@theforeman/vendor` 10.1.1 but in packaging we have 10.1.0.

This is also why I hate that all plugins simply specify `>= 0` rather than a proper lower bound. We'd catch things more easily.

Ron-Lavi (commented):
@ShimShtein now that foreman-js are updated to v10.1.7,
you can rebase and remove the `@redhat-cloud-services/frontend-components` dependency which is part of that version

ekohl (changes requested):

packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -1,58 +1,37 @@
...
-BuildRequires: %{?scl_prefix}npm(@babel/core) >= 7.7.0
-BuildRequires: %{?scl_prefix}npm(@babel/core) < 7.8.0
-BuildRequires: %{?scl_prefix}npm(@redhat-cloud-services/frontend-components) >= 2.5.0
-BuildRequires: %{?scl_prefix}npm(@redhat-cloud-services/frontend-components) < 3.0.0
-BuildRequires: %{?scl_prefix}npm(@theforeman/builder) >= 8.16.0
-BuildRequires: %{?scl_prefix}npm(jed) >= 1.1.1
-BuildRequires: %{?scl_prefix}npm(jed) < 1.2.0
+BuildRequires: npm(@babel/core) >= 7.7.0
+BuildRequires: npm(@babel/core) < 7.8.0
+BuildRequires: npm(@theforeman/builder) >= 8.16.0
```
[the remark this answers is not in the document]
ShimShtein: @ekohl updated this manually to see if it solves the problem, apparently it is not.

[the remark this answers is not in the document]
ekohl: I wonder if it needs to be imported via some vendor path, but the whole frontend is a bit of a mystery to me.

PR #8447, opened 2022-09-01
Rebuild python-websockify against Python 3.9 and updated spec

ehelms:

### DIFF
--- a/packages/foreman/python-websockify/python-websockify.spec
+++ b/packages/foreman/python-websockify/python-websockify.spec
@@ -1,38 +1,35 @@
-%global pkgname websockify
+# Created by pyp2rpm-3.3.8
+%global pypi_name websockify
 %global summary WSGI based adapter for the Websockets protocol
-Name:           python-%{pkgname}
+
+Name:           python-%{pypi_name}
 Version:        0.10.0
-Release:        3%{?dist}
+Release:        1%{?dist}
 Summary:        %{summary}
 
 License:        LGPLv3
 URL:            https://github.com/novnc/websockify
-Source0:        %{url}/archive/v%{version}/websockify-%{version}.tar.gz
+Source0:        https://files.pythonhosted.org/packages/source/w/%{pypi_name}/%{pypi_name}-%{version}.tar.gz
 BuildArch:      noarch
 
+BuildRequires:  python%{python3_pkgversion}-devel
+BuildRequires:  python%{python3_pkgversion}-setuptools
+
 %description
-Python WSGI based adapter for the Websockets protocol
+%{summary}
 
-%package -n python3-%{pkgname}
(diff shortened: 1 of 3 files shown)


packages/foreman/python-websockify/python-websockify.spec
```diff
@@ -1,38 +1,35 @@
-%global pkgname websockify
+# Created by pyp2rpm-3.3.8
```
ekohl, line 1: I explicitly copied it from EPEL to keep it as close as possible.

packages/foreman/python-websockify/python-websockify.spec
```diff
@@ -44,22 +41,16 @@ sed -i '/setup_requires/d; /install_requires/d; /dependency_links/d' setup.py
 %install
 %py3_install
 
-rm -Rf %{buildroot}/usr/share/websockify
-mkdir -p %{buildroot}%{_mandir}/man1/
-install -m 444 docs/websockify.1 %{buildroot}%{_mandir}/man1/
```
ekohl, line 49: Can we keep the manpage?

packages/foreman/python-websockify/python-websockify.spec
```diff
@@ -1,38 +1,35 @@
-%global pkgname websockify
+# Created by pyp2rpm-3.3.8
+%global pypi_name websockify
 %global summary WSGI based adapter for the Websockets protocol
-Name:           python-%{pkgname}
+
+Name:           python-%{pypi_name}
 Version:        0.10.0
-Release:        3%{?dist}
+Release:        1%{?dist}
```
ekohl, line 7: This doesn't match the changelog.

ekohl (changes requested):
You'll need to change the dependency in Foreman too. Perhaps that should instead depend on `/usr/bin/websockify`?

packages/foreman/python-websockify/python-websockify.spec
ehelms: The benefit I saw here is have defined a fairly standard way within our packaging of how Python packages should look for EL8+ with our Pulpcore packaging paradigms. Thus, having our python packages match this is more clear to developers where to look and what the patterns of packaging are. That this came from Fedora and is intended to match it "1:1" rather than our own is not obvious. 

However, I can put it back if there is a strong feeling in that direction and I'll add a comment at the top to make it clear.

ehelms: @ekohl could you weigh in one more time?

ekohl: I do feel that a manpage is good for something that places in `/usr/bin` and that isn't in the PyPI tarball, but is in the current RPM.

PR #8448, opened 2022-09-01
Update rubygem-foreman_puppet to 4.0.3

Ron-Lavi:


packages/plugins/rubygem-foreman_puppet/rubygem-foreman_puppet.spec
```diff
@@ -9,7 +9,7 @@
 %global release 2
```
ehelms, line 9: Needs to be set back to 1

Ron-Lavi: Updated 👍🏼 btw what does `release 1/2` mean?

ekohl: Don't forget the changelog entry. That must match.

The release is a version number for the packaging. It allows packagers to make incremental changes on the same original release. https://github.com/theforeman/foreman-packaging/commit/3e4c0f1924a8760df55b68a0fe826f48f34c3b32 is an example of that.

Ron-Lavi: Thanks for the explanation, done.

LiorKGOW:
Correct me if I am wrong :angel:, I think the needs testing label could be removed

LiorKGOW (approved):

ekohl (approved):

ekohl:
> Correct me if I am wrong angel, I think the needs testing label could be removed

I did ask about that in the past (https://community.theforeman.org/t/drop-needs-testing-label-in-prs/22504) but never moved forward with it.

PR #8463, opened 2022-09-05
Only run package bumping on theforeman

ekohl:
Otherwise forks also run the same, which is wasting a lot of resources and sending failure messages to Discourse.


.github/workflows/bump_packages.yml
```diff
@@ -13,6 +13,7 @@ jobs:
   rpm_list:
     name: 'Gather RPMs'
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'theforeman'
```
evgeni, line 16: same for debian?

PR #8512, opened 2022-09-20
Bump rubygem-foreman_ansible to 10.0.0

nofaralfasi:
Testing if the new foreman_ansible release works.


packages/plugins/rubygem-foreman_ansible/rubygem-foreman_ansible.spec
```diff
@@ -8,7 +8,7 @@
 %global foreman_min_version 3.4.0
```
ekohl, line 8: I see the minimum version was bumped to 3.5 (https://github.com/theforeman/foreman_ansible/blob/630f8d734cd57775e22a609cbacdab113cefe703/lib/foreman_ansible/register.rb#L4). We don't have any automation to update this, so you'll need to do this manually.
```suggestion
%global foreman_min_version 3.5.0
```

PR #8514, opened 2022-09-20
explicitly vendor sinatra

ekohl:
rack-protection and sinatra have to be the same version foreman pulls in rack-protection via sidekiq, f-tasks needs sinatra let foreman.deb provide both gems in the cache, so that they always match

(cherry picked from commit a195bc354b34f22a05a2e688aaa7471af2b08d3e)


ekohl:
I kicked off https://ci.theforeman.org/job/foreman-3.4-release-pipeline/5/ to make it available.

PR #8535, opened 2022-09-27
Drop requires on psutil, runner doesn't need it anymore

evgeni:
This was an workaround introduced in fbae4775a45c122123e475c574df981a1ba0a174, but in the meantime we did get a fixed runner package (ansible-runner 2.x doesn't need psutil at all), so we can drop the workaround.

You could argue this should add a `Requires: ansible-runner >= 2`, and I would agree ;)

### DIFF
--- a/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
+++ b/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
@@ -17,7 +17,7 @@
 
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 3.4.1
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Summary: Smart-Proxy Ansible plugin
 Group: Applications/Internet
 License: GPLv3
@@ -28,7 +28,6 @@ Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 Requires: ansible
 %else
 Requires: (ansible or ansible-core)
-Requires: (python38-psutil if ansible-core)
 %endif
 
 Requires: ansible-collection-theforeman-foreman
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -28,8 +28,8 @@ Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 Requires: ansible
 %else
 Requires: (ansible or ansible-core)
```
ekohl, line 30: Does it still require ansible directory or is it always invoking ansible-runner? It looks to me like the latter and we can simply rely on ansible-runner to pull in ansible itself.

evgeni: It will always invoke runner, but the runner pkg has no dependency on Ansible…

ekohl: Is that something we should fix?

evgeni: It's complicated. *Technically* runner doesn't need a packaged Ansible (it can utilize a container instead, for example). However, for our usecases, it does need it, as we don't use any of the container functionality.

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -148,6 +148,9 @@ ln -sv %{_root_sysconfdir}/foreman-proxy/ansible.cfg %{buildroot}%{foreman_proxy
 %doc %{gem_instdir}/README.md
 
 %changelog
+* Tue Sep 27 2022 Evgeni Golov - 3.4.1-2
+- Drop requires on psutil, runner doesn't need it anymore
```
ekohl, line 152:
```suggestion
- Drop requires on psutil, runner doesn't need it anymore
- Add dependency on ansible-runner
```

ekohl (approved):

PR #8248, opened 2022-07-18
RPM dependencies for foreman_google package

stejskalleos:
For https://github.com/theforeman/foreman-packaging/pull/7777

Adds:
* `google-cloud-compute`
* `google-cloud-core`
* `google-cloud-compute-v1`
* `google-cloud-common`
* `google-cloud-errors`
* `gapic-common`
* `google-protobuf`
* `googleapis-common-protos`
* `googleapis-common-protos-types`

PR #8536, opened 2022-09-27
Add rubygem-google-protobuf package

stejskalleos:
For https://github.com/theforeman/foreman-packaging/pull/8248


ekohl (changes requested):
Please add this to `plugins`, not `foreman`. So
```
./add_gem_package.sh google-protobuf default plugins foreman-plugins-el8
```

This is because the package should end up in the plugins repository, not the main repository.

stejskalleos:
> Please add this to `plugins`, not `foreman`. So
> 
> ```
> ./add_gem_package.sh google-protobuf default plugins foreman-plugins-el8
> ```
> 
> This is because the package should end up in the plugins repository, not the main repository.

Should I do that also with packages in https://github.com/theforeman/foreman-packaging/pull/8248? Right now they are in `packages/foreman`, not in `packages/plugins` folder

stejskalleos:
Linter is failing on `msg: Could not find specfile` even when it's there.

@theforeman/packaging What are proper steps for adding new package?
Looks like running `./add_gem_package.sh google-protobuf default foreman-plugins-el8` is not enough and additional steps must be done, just don't know where to look for more information

package_manifest.yaml
```diff
@@ -194,6 +194,7 @@ foreman_core_packages:
     rubygem-google-apis-sqladmin_v1beta4: {}
     rubygem-google-apis-storage_v1: {}
     rubygem-google-cloud-env: {}
+    rubygem-google-protobuf: {}
```
ehelms, line 197: This needs to be moved under the plugins section around like 674/675

packages/plugins/rubygem-google-protobuf/rubygem-google-protobuf.spec
```diff
@@ -0,0 +1,83 @@
...
+%{gem_libdir}
+%exclude %{gem_cache}
+%{gem_spec}
+
+%files doc
+%doc %{gem_docdir}
+%{gem_instdir}/tests
+
+%changelog
+* Tue Sep 27 2022 Leos Stejskal <lstejska@redhat.com> 3.21.6-1
```
ehelms, line 81:
```suggestion
* Tue Sep 27 2022 Leos Stejskal <lstejska@redhat.com> 1:3.21.6-1
```

stejskalleos:
rpm job is failing on error:
```
package: rubygem-google-protobuf-doc-1:3.21.6-1.el8.noarch from downloaded_rpms
unresolved deps:
rubygem-google-protobuf = 3.21.6-1.el8
```
I'm not sure why it downloads `rubygem-google-protobuf-doc`, that gem doesn't exists, and [google-protobuf](https://github.com/protocolbuffers/protobuf/blob/main/ruby/google-protobuf.gemspec) doesn't have it as dependency

dgoetz:
It is not a gem, but a subpackage created because of the section `%package doc` in the spec file. This is normal and looking at the Koji task it is build properly. So just an explanation for the -doc, not the problem.

stejskalleos:
> It is not a gem, but a subpackage created because of the section `%package doc` in the spec file. This is normal and looking at the Koji task it is build properly. So just an explanation for the -doc, not the problem.

Okay, how do I fix the rpm job? Should I remove the %doc section from spec file?

ehelms:
The failure is a legit issue if you look at the provides on the RPM due to the Epoc:

```
$ rpm -qp rubygem-google-protobuf-3.21.6-1.el8.x86_64.rpm --provides
rubygem(google-protobuf) = 3.21.6
rubygem-google-protobuf = 1:3.21.6-1.el8
rubygem-google-protobuf(x86-64) = 1:3.21.6-1.el8
```

packages/plugins/rubygem-google-protobuf/rubygem-google-protobuf.spec
```diff
@@ -0,0 +1,83 @@
...
+BuildRequires: gcc
+# end specfile generated dependencies
+
+%description
+Protocol Buffers are Google data interchange format.
+
+
+%package doc
+Summary: Documentation for %{name}
+Requires: %{name} = %{version}-%{release}
```
ehelms, line 29:
```suggestion
Requires: %{name} = %{epoch}:%{version}-%{release}
```

packages/plugins/rubygem-google-protobuf/rubygem-google-protobuf.spec
```diff
@@ -0,0 +1,83 @@
+# template: default
+%global gem_name google-protobuf
+%global gem_require_name %{gem_name}
+
+Name: rubygem-%{gem_name}
+Version: 3.21.6
+Release: 1%{?dist}
+Epoch: 1
```
ekohl, line 8: I may have missed this, but why is there an epoch?

ehelms: That's a great question, I cannot find a reason for having one as I do not see any competing RPMs that exist and we have not released one ourselves. I think then you can just drop this line and revert the changelog @stejskalleos

ehelms:
I did some local build and testing and opened up https://github.com/theforeman/foreman-packaging/pull/8546

ehelms:
Replaced by https://github.com/theforeman/foreman-packaging/pull/8546

PR #8538, opened 2022-09-27
Obsolete python3-ansible-runner and update bindir location

ehelms:

### DIFF
--- a/packages/plugins/ansible-runner/ansible-runner.spec
+++ b/packages/plugins/ansible-runner/ansible-runner.spec
@@ -2,7 +2,7 @@
 
 Name:           %{pypi_name}
 Version:        2.2.1
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        A tool and python library to interface with Ansible
 
 License:        ASL 2.0
@@ -16,6 +16,8 @@ BuildRequires: python%{python3_pkgversion}-setuptools
 BuildRequires: python%{python3_pkgversion}-pbr
 BuildRequires: python%{python3_pkgversion}-rpm-macros
 Requires:      python%{python3_pkgversion}-%{pypi_name} = %{version}-%{release}
+Obsoletes:     python3-%{pypi_name} < %{version}
+Obsoletes:     python38-%{pypi_name} < %{version}
 
 %description
 Ansible Runner is a tool and python library that helps when interfacing with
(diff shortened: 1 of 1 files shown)


packages/plugins/ansible-runner/ansible-runner.spec
```diff
@@ -57,15 +58,19 @@ ln -s %{_bindir}/ansible-runner-%{python3_version} %{buildroot}/%{_bindir}/ansib
...
-
-%files -n python%{python3_pkgversion}-%{pypi_name}
-%{python3_sitelib}/*
 %{_bindir}/ansible-runner
 %{_bindir}/ansible-runner-3
 %{_bindir}/ansible-runner-%{python3_version}
+
+%files -n python%{python3_pkgversion}-%{pypi_name}
+%{python3_sitelib}/*
 %{_datadir}/ansible-runner/*
```
evgeni, line 67: Shouldn't the `datadir` be also be in the "normal" (non-`python%{python3_pkgversion}`) package then, as it is also not-python-specific?

packages/plugins/ansible-runner/ansible-runner.spec
```diff
@@ -26,6 +26,7 @@ standalone tool, or imported into a python project.
 Summary:        %{summary}
 %{?python_provide:%python_provide python%{python3_pkgversion}-%{pypi_name}}
 Obsoletes:      python38-%{pypi_name} < %{version}
+Obsoletes:      python3-%{pypi_name} < %{version}
```
evgeni, line 29: IMHO the obsoletes now should be on the `ansible-runner` package, as that is the one providing the binaries (and thus the same functionality) now.

packages/plugins/ansible-runner/ansible-runner.spec
ehelms: I wondered about that, I wasn't sure what exactly the datadir was used for or what is put in there.

packages/plugins/ansible-runner/ansible-runner.spec
```diff
@@ -16,6 +16,7 @@ BuildRequires: python%{python3_pkgversion}-setuptools
 BuildRequires: python%{python3_pkgversion}-pbr
 BuildRequires: python%{python3_pkgversion}-rpm-macros
 Requires:      python%{python3_pkgversion}-%{pypi_name} = %{version}-%{release}
+Obsoletes:      python3-%{pypi_name} < %{version}
```
evgeni, line 19:
```suggestion
Obsoletes:     python3-%{pypi_name} < %{version}
```

And I think we should move the `python38` obsolete here too.

evgeni:
Other than the nitpick, I did test upgrades from both 1.4.7 "upstream" and our "old" 2.2.1 packages, and both worked fine.

evgeni (approved):

PR #8541, opened 2022-09-28
support non-semver range definitions

evgeni:
while node packages need to me semver-versioned, the version range definition as used to specify dependencies is not. `^10.0` is a perfectly valid way to say "at least 10.0 and less than 11".

In semver3 you can use `parse('10.0', optional_minor_and_patch=True)` but we don't have semver3 yet (it's not even released yet), so we need to make this up ourself.

### DIFF
--- a/update-requirements
+++ b/update-requirements
@@ -24,7 +24,10 @@ def get_requirement(version):
 
     if version.startswith('^') or version.startswith('~'):
         modifier = version[0]
-        parsed = semver.parse(version[1:])
+        split_version = version[1:].split('.', 2)
+        while(len(split_version) < 3):
+            split_version.append('0')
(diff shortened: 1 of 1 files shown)


update-requirements
```diff
@@ -24,7 +24,10 @@ def get_requirement(version):
 
     if version.startswith('^') or version.startswith('~'):
         modifier = version[0]
-        parsed = semver.parse(version[1:])
+        split_version = version[1:].split('.', 2)
+        while(len(split_version) < 3):
+            split_version.append('0')
```
ekohl, line 29: Is this too smart?
```suggestion
        split_version += (3 - len(split_version)) * ['0']
```

evgeni: Yeah?

ekohl (approved):

PR #8543, opened 2022-09-29
Use python39-psutil with ansible-core 2.13+

evgeni:
In CentOS Stream 8, ansible-core 2.13 is built with Python 3.9 and thus needs the matching psutil package when called by ansible-runner.


evgeni:
Should be picked also to 3.3 (and ideally 3.2, to be nice) once ACK'ed

ekohl (approved):

ekohl:
> Should be picked also to 3.3

7bea260f107f3794e52abe7f301e098f691a136b

> and ideally 3.2, to be nice

For 3.2 (technically EOL now) it doesn't apply cleanly since it's still at version 3.3.1 so I didn't do that.

evgeni:
> For 3.2 (technically EOL now) it doesn't apply cleanly since it's still at version 3.3.1 so I didn't do that.

https://github.com/theforeman/foreman-packaging/pull/8544

PR #8556, opened 2022-10-03
Update rubygem-foreman_ansible to 10.0.1

stejskalleos:


stejskalleos:
@evgeni version with fix for nightly: https://github.com/theforeman/foreman_ansible/pull/563

evgeni:
Should foreman_min_version be bumped to 3.5, as older won't have the new GraphQL gem?

evgeni:
Ah, it's even set in https://github.com/theforeman/foreman_ansible/blob/master/lib/foreman_ansible/register.rb#L4 already

evgeni (approved):

PR #8579, opened 2022-10-10
Allow foreman-ygg-worker to work with newer yggdrasil

adamruzicka:
Yggdrasil past 0.2.1 changed the way how workers are discovered, placing a binary into a directory is no longer sufficient. There has to be an accompanying configuration file that point to said binary.


evgeni:
[test rpm]

adamruzicka:
The built rpm seems to work well

evgeni (approved):

PR #8580, opened 2022-10-11
Add rubygem-webrick package

ekohl:
This package is part of Ruby 2.7's default gems, but in Ruby 3.1 it's no longer there. That's why packages have started to add this dependency.  Since ruby-default-gems on EL8 doesn't have Provides: rubygem(webrick) it isn't satisfied. Rather than hacking around it in every gem that requires it, this starts to package webrick explicitly. In the future it needs to happen anyway.


ehelms (approved):

evgeni:
```console
% koji add-pkg --owner kojiadm foreman-nightly-el8 rubygem-webrick              
Adding 1 packages to tag foreman-nightly-el8
```

evgeni (approved):

PR #8581, opened 2022-10-11
Cherry pick explicit vendorization of sinatra to deb/3.3

ekohl:
After the release of 3.3.1-1 installation of foreman-tasks fails. These picks should allow it to work again.

### DIFF
--- a/plugins/ruby-foreman-tasks/debian/changelog
+++ b/plugins/ruby-foreman-tasks/debian/changelog
@@ -1,4 +1,10 @@
-ruby-foreman-tasks (6.0.2) stable; urgency=low
+ruby-foreman-tasks (6.0.2-2) stable; urgency=low
+
+  * don't vendor sinatra and its gems, this is now provided by foreman
+
+ -- Evgeni Golov <evgeni@debian.org>  Tue, 11 Oct 2022 16:05:03 +0200
+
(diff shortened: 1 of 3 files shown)


ekohl:
```
[2022-10-11T14:08:52.767Z] /usr/bin/foreman-ruby: No such file or directory -- /usr/bin/bundle (LoadError)
```

evgeni:
The problem is that we switched our plugins buildroot to bullseye, and that doesn't install `bundler` by default (which is what we call in `foreman.postinst`) anymore. There have been a series of changes in 3.4+ to make this not necessary (d188386aef78ab95dbb4847117070de857a3e083 88accc17db4fe27d2498281c1409ad54705e2937 87747e3d4f9b9779deef6955790e86caee3e74ec), but I don't feel like we should pick these to 3.3 at this stage.

A valid workaround would be making `foreman` *explicitly* depend on `bundler` and only then try to build the new plugin (read: split the changes into two PRs).

evgeni:
I did that now in https://github.com/theforeman/foreman-packaging/pull/8582, and also added the sinatra change to buster, which you forgot here ;)

ekohl:
Thanks for that. I've prepared a rebase, but waiting on https://ci.theforeman.org/job/foreman-packaging-deb-3.3-release/28/ to complete before I push it.

evgeni (approved):

PR #8583, opened 2022-10-13
clean up ruby and git dependencies

evgeni:

### DIFF
--- a/debian/bullseye/foreman/control
+++ b/debian/bullseye/foreman/control
@@ -3,7 +3,7 @@ Maintainer: Greg Sutcliffe <greg.sutcliffe@gmail.com>
 Section: web
 Priority: extra
 Standards-Version: 3.9.1
-Build-Depends: debhelper (>= 9.20160709), ruby-dev (>= 1:2.7), git | git-core,
+Build-Depends: debhelper (>= 9.20160709), ruby-dev (>= 1:2.7), git,
                zlib1g-dev, libpq-dev, pkg-config, libvirt-dev,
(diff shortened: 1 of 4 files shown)


evgeni:
[test deb]

debian/bullseye/foreman/control
```diff
@@ -14,7 +14,7 @@ Package: foreman
 Architecture: any
 Section: web
 Priority: extra
-Depends: ruby, ruby (>= 1:2.7) | bundler (>= 1.3), ruby-dev, zlib1g-dev,
+Depends: ruby (>= 1:2.7), ruby-dev, zlib1g-dev,
```
ekohl, line 17: ruby only provides `/usr/bin/bundle$(RUBY_VERSION)`. Looking at our packaging I think we consistently use that, but I wonder if anything does use `bundle` at runtime.

ekohl (commented):
On a related note: I saw that a lot of plugins depend on bundler, but that can be dropped too. It build depends on foreman which provides everything we need (including the shared script).

debian/bullseye/foreman/control
evgeni: Correct, but in the current notation, `bundler` is not installed anyways on `ruby >= 2.7` (unless pulled in otherwise).

If things break at runtime, we should have test for that. So :woman_shrugging:

evgeni: But your comment is somewhat right, as we do *not reliably* use `bundle$(RUBY_VERSION`… adding a change for that!

ekohl (approved):

PR #8587, opened 2022-10-14
Update ruby-foreman-tasks foreman version

elkinaguas:
Update the foreman version ruby-foreman-tasks depends on.

When trying to update Foreman from v3.3 to v3.4, ruby-foreman-tasks says it depends on foreman (>= 3.4.0-2) but 3.4.0-1 is to be installed. The current foreman version specified in https://github.com/theforeman/foreman-packaging/blob/deb/3.4/plugins/ruby-foreman-tasks/debian/control is 3.4.0-2.

Foreman community ref: https://community.theforeman.org/t/not-able-to-upgrade-foreman-from-v3-3-to-v3-4/30711


evgeni:
This shouldn't be needed, I commented on the community thread.

PR #8594, opened 2022-10-17
Drop unused PLUGIN_HAS_API definition

evgeni:


ekohl (approved):

evgeni:
(and if you're going to say: "but evgeni, did you just use this almost nonsense change to check if other plugins build after your latest changes?!", the answer is: yes, yes I did!)

ekohl:
I'm glad you added that because I was already guessing it ;)

PR #8599, opened 2022-10-18
Update hammer_cli_foreman_webhooks to 0.0.4

evgeni:


evgeni:
@ofedoren mind double checking me, is 0.0.4 fine for 3.4, or should we keep that at 0.0.3?

ofedoren:
@evgeni, yeap, that should be fine.

PR #8602, opened 2022-10-18
Update rubygem-puppetdb_foreman to 6.0.0

dgoetz:
Cherry picked to supported versions as it breaks installations.

(cherry picked from commit 57b2a5c534ab1446b68163dd6d7ee6bd9edb6a54)


ekohl (commented):
The rewritten spec only works on Foreman 3.4+ since it drops EL7 support. For 3.3 a different pick is needed. IIRC it only broke on 3.4 so I'd also be fine with only updating 3.4 to 6.0.0. Up to you if you want to spend the time to make it work on 3.3.

dgoetz:
Thanks, I did not think about this! I will close the PR for now and have an eye on the community for reports to see if a pick is really needed.

PR #8605, opened 2022-10-19
Bump rubygem-hammer_cli_foreman_tasks to 0.0.18-1

evgeni:


evgeni:
@adamruzicka or was it intentional that 0.0.18 was never packaged?

adamruzicka:
No, I just did a release to rubygems and hoped the automation would kick in and bump the packaging? Shouldn't that have happened?

evgeni:
It should, if the package would have had the `template: hammer_plugin` header, which this PR adds ;)

adamruzicka:
Ah, TIL, thank you

ekohl (approved):

PR #8607, opened 2022-10-19
Regenerate hammer plugins based on latest template

evgeni:


packages/plugins/rubygem-hammer_cli_foreman_openscap/rubygem-hammer_cli_foreman_openscap.spec
```diff
@@ -1,70 +1,74 @@
...
+
+Name: rubygem-%{gem_name}
 Version: 0.1.13
-Release: 1%{?foremandist}%{?dist}
-Group: Applications/System
-License: GPLv3
-URL: https://github.com/theforeman/hammer_cli_foreman_openscap
+Release: 2%{?foremandist}%{?dist}
+Summary: Foreman OpenSCAP commands for Hammer
+License: GPL-3
```
evgeni, line 11: `GPL-3` doesn't look correct?

ekohl: Are you saying it should be `GPL-3.0` upstream to be valid SPDX? Then our [license.json](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/licenses.json) will also understand it.

evgeni: That's what I said and committed to :)

ekohl: Well, GH didn't show your message until I submitted. I blame that.

packages/plugins/rubygem-hammer_cli_foreman_ssh/rubygem-hammer_cli_foreman_ssh.spec
```diff
@@ -1,78 +1,72 @@
...
+%{gem_spec}
+%config(noreplace) %{hammer_confdir}/cli.modules.d/%{plugin_name}.yml
 
 %files doc
 %doc %{gem_docdir}
 %doc %{gem_instdir}/README.md
+%doc %{gem_instdir}/config
 
 %changelog
+* Wed Oct 19 2022 Evgeni Golov 0.0.3-1
```
ekohl, line 67: It only came out in 2015. Don't you think it's too bleeding edge?

ekohl (approved):

ekohl:
Looks like it now requires `net-ssh-multi`:
```
[2022-10-19T10:50:46.972Z]     package: rubygem-hammer_cli_foreman_ssh-0.0.3-1.fm3_5.el8.noarch from downloaded_rpms
[2022-10-19T10:50:46.972Z]       unresolved deps:
[2022-10-19T10:50:46.972Z]         rubygem(net-ssh-multi) >= 1.2.1
```

evgeni:
Always did?

https://github.com/theforeman/hammer-cli-foreman-ssh/blob/52b0de73a46a4de757d2a1e8b20555e664be2dd0/hammer_cli_foreman_ssh.gemspec#L17

ekohl:
Hmm you're right. However, we only package 1.2.0, not 1.2.1.

evgeni:
> Hmm you're right. However, we only package 1.2.0, not 1.2.1.

Nothing as easy as that ;)

packages/plugins/rubygem-net-ssh-multi/rubygem-net-ssh-multi.spec
```diff
@@ -1,91 +1,69 @@
...
+Release: 1%{?dist}
 Summary: Control multiple Net::SSH connections via a single interface
-Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 1.2.0
-Release: 10%{?dist}
-Group: Development/Languages
 License: MIT
 URL: https://github.com/net-ssh/net-ssh-multi
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
-Patch0: rubygem-net-ssh-multi-1.2.0-minitest.patch
```
evgeni, line 17: the patch is part of 1.2.1, so it's OK that it's gone

packages/plugins/rubygem-net-ssh-multi/rubygem-net-ssh-multi.spec
```diff
@@ -1,91 +1,69 @@
...
+# by default, so that we can move it into the buildroot in %%install
+%gem_install
 
 %install
 mkdir -p %{buildroot}%{gem_dir}
 cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
-%check
-%if 0%{?enable_check}
```
evgeni, line 64: we had enable check hardcoded to 0 anyways, so :woman_shrugging:

ekohl (approved):

PR #8608, opened 2022-10-19
drop rubygem-hammer_cli_csv

evgeni:


evgeni:
@ehelms please double-check me, but my memory says we stopped supporting that hammer extension quite a while ago and just never dropped the packaging.

ekohl:
The last update was in 2018 so I think you're right. Perhaps the repo should be archived too to make it clear?

evgeni:
That's a good idea, but that's also for Eric as I don't have the powers in Katello.

ehelms (approved):

ehelms:
Archived.

evgeni:
```
koji untag-build --all katello-nightly-el8  rubygem-hammer_cli_csv
koji block-pkg katello-nightly-el8 rubygem-hammer_cli_csv
```

PR #8610, opened 2022-10-20
Update the bundler version in Gemfile.lock on package update

evgeni:


ehelms (approved):

evgeni:
meh, this seems to have broken updates, investigating

PR #8620, opened 2022-10-24
write to GITHUB_OUTPUT instead of using set-output

evgeni:
set-output is deprecated https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/scripts/list_updatable_packages
+++ b/scripts/list_updatable_packages
@@ -56,8 +56,9 @@ def main() -> None:
 
     if 'GITHUB_ACTION' in os.environ:
         directories = [entry['directory'] for entry in matrix]
-        print(f'::set-output name=directories::{json.dumps(directories)}')
-        print(f'::set-output name=matrix::{json.dumps(matrix)}')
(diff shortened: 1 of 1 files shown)


scripts/list_updatable_packages
```diff
@@ -56,8 +56,9 @@ def main() -> None:
 
     if 'GITHUB_ACTION' in os.environ:
         directories = [entry['directory'] for entry in matrix]
-        print(f'::set-output name=directories::{json.dumps(directories)}')
-        print(f'::set-output name=matrix::{json.dumps(matrix)}')
+        with open(os.environ['GITHUB_OUTPUT'], 'w') as github_output:
```
ekohl, line 59: Should it be `w` or `a` as mode? The example is `echo .. >> $GITHUB_OUTPUT` so that suggests `a` to me, but they failed to provide a programmatic example in their blog and documentation.

For this particular script it may not matter, but I'd like to know what is the right thing to do.

PR #8625, opened 2022-10-25
Update rubygem-pulp_ostree_client to 2.0.0

evgeni:


evgeni:
right, this needs a release with faraday fixes

packages/katello/rubygem-pulp_ostree_client/pulp_ostree_client-2.0.0a6.gem
```diff
@@ -0,0 +1 @@
+../../../.git/annex/objects/vV/50/SHA256E-s57856--4ebb1d6ede08752bc85109fdbf572c9cd52766b083b90627becf3e9ecbb1be57.0a6.gem/SHA256E-s57856--4ebb1d6ede08752bc85109fdbf572c9cd52766b083b90627becf3e9ecbb1be57.0a6.gem
```
ekohl, line 1: Why is there an additional of the gem, but no removal of the old?

evgeni: because the removal somehow accidentally is already in the repo :woman_shrugging:

ehelms:
@evgeni mind doing an update of this to a7?

ehelms:
> There is a non-dev a7 release?

Sorry, you are right there. The annoying part is that our update job thinks we should update to that and it breaks the job:

https://github.com/theforeman/foreman-packaging/actions/runs/3384595641/jobs/5627825968#step:5:11

evgeni:
Yeah, there is no non prerelease of that gem, which confuses all the tools.

evgeni:
let's fix the source in https://github.com/theforeman/foreman-packaging/pull/8700 and then we can see how this one continues

evgeni:
there is finally a release \o/

ekohl (approved):

PR #8637, opened 2022-10-26

ekohl:


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -55,21 +55,21 @@ BuildRequires: rubygem(pg)
...
-BuildRequires: rubygem(pulp_ansible_client) >= 0.13.1
-BuildRequires: rubygem(pulp_ansible_client) < 0.14
-BuildRequires: rubygem(pulp_container_client) >= 2.10.0
-BuildRequires: rubygem(pulp_container_client) < 2.11.0
-BuildRequires: rubygem(pulp_deb_client) >= 2.18.0
-BuildRequires: rubygem(pulp_deb_client) < 2.19.0
-BuildRequires: rubygem(pulp_rpm_client) >= 3.17.0
-BuildRequires: rubygem(pulp_rpm_client) < 3.18.0
+BuildRequires: rubygem(pulpcore_client) >= 3.21.0
+BuildRequires: rubygem(pulpcore_client) < 3.22.0
```
ekohl, line 59: @evgeni any thoughts on dropping the version bounds? or at least the upper bound? Since this is a proper gem we should have automatic dependencies based on the gemspec. AFAIK that's still verified, which would make the build fail, just later than it is now. It'd reduce the churn on keeping this in sync. At least until https://github.com/theforeman/foreman-packaging/pull/8414.

To be clear, I think it should be a separate effort, but it can help us. Especially if we're going to build more nightly/scratch packages.

evgeni: I'd be cool with that

ekohl:
[test rpm]

packages/katello/rubygem-katello/rubygem-katello.spec
ekohl: Cool. Not sure I can get to it soon, but I'll keep it in mind.

ekohl:
Repoclosure fails because it hasn't been mashed yet.

PR #8658, opened 2022-10-27
fix lockfile updating

evgeni:
Fixes: 1bcac66ffbb620e856b0bb482972102d797c37be


evgeni:
nope, that doesn't work either, fuck

PR #8660, opened 2022-10-27
Require sd_notify for systemd integration support from Puma

ehelms:
I debated if this should just be added to Puma itself as a requirement. As it's only required if the service using Puma implement systemd support which is driven by our implementation in foreman rather than Puma itself.

On nearly all standard installation, a foreman-proxy is installed alongside Foreman which results in the `sd_notify` gem being present. However, we have places like our puppet acceptance tests that test Foreman in isolation and lack this support. I observed, for example, that foreman.service started timing out on start on https://github.com/theforeman/puppet-katello/pull/457. I see the same behaviour locally, but not when I install rubygem-sd_notify.

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -4,7 +4,7 @@
 %global dynflow_sidekiq_service_name dynflow-sidekiq@
 %global rake /usr/bin/rake
 
-%global release 15
+%global release 16
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


ekohl (changes requested):
Technically it's only needed for the `foreman-service` package. I think https://github.com/theforeman/foreman/commit/9b0652fed9f86fd429e237ce359422a1dba14a2a shouldn't have dropped the dependency.

ekohl:
Oh, it should be in https://github.com/theforeman/foreman/blob/develop/bundler.d/service.rb now.

ekohl:
https://github.com/theforeman/foreman/pull/9488 does that.

ekohl (approved):
I hijacked your PR to reflect the changes to Foreman itself.

ehelms:
Gonna have to start charging you a fee for taking over my PRs :laughing:

PR #8666, opened 2022-10-31
Refs #35088 - Remove foreman-gce and mark foreman_google as replacement

ekohl:
I think this is the migration path for users, but not really sure.

PR #8667, opened 2022-10-31
Refs #35088 - Replace foreman_gce with foreman_google

stejskalleos:
Foreman: https://github.com/theforeman/foreman/pull/9272
RPM: https://github.com/theforeman/foreman-packaging/pull/8666


ekohl (changes requested):
In https://github.com/theforeman/foreman-packaging/pull/8666 I let the foreman_google plugin replace `foreman-gce` which means `dnf upgrade` actually removes `foreman-gce`. Here for Debian something similar is needed. Otherwise `foreman-gce` is never removed from the actual system and that blocks upgrades (since it requires `foreman = $version`).

Looking at other files it means adding `Replaces: foreman-gce` to the plugin's control file. You'll also need to create changelog entries (use `scripts/changelog.rb` for that).

Could you also use `Refs #35088` so we can correlate the changes?

evgeni:
Replaces on Debian does not imply apt will replace the PKG (like obsoletes does in RPM), it only means it can overwrite files.

To have the same behavior like in rpm, we need an "empty" foreman-gce package that has a Depends on ruby-foreman-google. 
And ruby-foreman-google should have:
```
Breaks: foreman-gce (<<3.5.0)
Replaces: foreman-gce (<<3.5.0)
Provides: foreman-gce
```

See https://wiki.debian.org/PackageTransition (case 5) and https://wiki.debian.org/RenamingPackages

stejskalleos:
@evgeni @ekohl change log updated & updated control files

debian/bullseye/foreman/control
```diff
@@ -227,3 +216,16 @@ Description: package providing dynflow sidekiq services for Foreman
  a flexible systems management web application.
  .
  The foreman package should be installed along with this package.
+
+Package: foreman-google
```
evgeni, line 220: No, the package still needs to be called foreman-gce, as that's what the users have installed right now. And it needs a "Depends" on ruby-foreman-google, which is the plugin package.

The breaks etc only need to be in the plugin package.

evgeni (changes requested):

debian/bullseye/foreman/control
stejskalleos: Reverted back

ekohl (approved):

ekohl:
Waiting for https://ci.theforeman.org/job/foreman-develop-source-release/1811/ to build so there's a new source tarball.

ekohl:
[test deb]

ekohl:
```
Unable to delete '/home/jenkins/workspace/foreman-packaging-deb-pr-test'. Tried 3 times (of a maximum of 3) waiting 0.1 sec between attempts. (Discarded 87112 additional exceptions)
```

[test deb]

PR #8671, opened 2022-10-31
Update rubygem-foreman_scc_manager to 2.0.0

m-bucher:
Introducing new SCC-product selection page.


ekohl (commented):
Thanks! Minor comment inline.

packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
```diff
@@ -8,8 +8,8 @@
 %global katello_min_version 3.7.0
```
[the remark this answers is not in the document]
m-bucher: Makes sense :sweat_smile:

packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
```diff
@@ -80,7 +86,7 @@ cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
 %foreman_bundlerd_file
-%foreman_precompile_plugin -s
+%foreman_precompile_plugin -a -s
```
m-bucher, line 89: Looks like `-a` is only used by this plugin and foreman_snapshot_management. Is this still relevant?

ekohl: The comment for what it does:
https://github.com/theforeman/foreman-packaging/blob/37b3a9642d73d24026d05f9d95ee45f5144185ba/packages/foreman/foreman/foreman.spec#L850-L854
Having said that, I thought we dropped the apipie cache and always generated it on the fly now. 8509b32e57ce91b949d8a83f40f9ff69e3748186 did make that change for Foreman itself, but not for the plugins.

@ehelms @evgeni am I missing something, or is this an oversight?

evgeni: Looks like an oversight. I think we wanted to keep `-a` as an noop (did we?) but the compilation itself can really be dropped. The bits are unused anyways.

m-bucher: OK, dropped it :grin:

evgeni: Or maybe we just dropped all `-a` calls and called it a day?

ekohl (approved):

PR #8672, opened 2022-10-31
Drop apipie:cache generation in snapshot_mgmt spec

m-bucher:
see: https://github.com/theforeman/foreman-packaging/pull/8671#discussion_r1009674514


ekohl (approved):
Thanks!

PR #8674, opened 2022-11-01
Release rubygem-foreman_maintain 1.2.0

ehelms:


packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec
```diff
@@ -19,29 +13,14 @@ URL: https://github.com/theforeman/foreman_maintain
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 Source1: %{gem_name}.logrotate
 
-Requires: %{?scl_prefix_ruby}ruby(release)
-Requires: %{?scl_prefix_ruby}ruby(rubygems)
-Requires: %{?scl_prefix}rubygem(clamp) >= 0.6.2
-Requires: %{?scl_prefix}rubygem(highline)
+Requires: ruby
```
ekohl, line 16: Any chance you could include the generated Ruby bits inside `# start specfile generated dependencies` and `# end specfile generated dependencies` so that the next update will at least update as needed? I can imagine we'd add a minimal Ruby version specified for example and update it over time.

ekohl: The update script only replaces the parts between the `start` and `end` bits, which means `./bump_rpm.sh` should generally do the right thing. I've been very careful with that.

ehelms: Hrmm, OK, let's see if this is better

ekohl (approved):
Thanks! This is what I meant.

PR #8699, opened 2022-11-04
Allow foreman-build to be included per the needs of the build system

ehelms:
The foreman-build RPM provides a macro for plugin versioning that is not desired across all build systems and rebuilds of RPMs. Dropping the requires on foreman-build allows it to be included in buildroots per the configuration rather than a hard requirement of the foreman-plugin package.

The alternative solution:

Configure each build system with the `foremandist` macro as part of the configuration. For Koji that is using (https://docs.pagure.org/koji/setting_rpm_macros/#setting-rpm-macro-values) and configuring it in tool_belt tag configurations. For mock that is setting `rpmmacros` option, which I believe is the same for Copr. 

I think the standard method is still the proposal here -- a dedicated RPM with macros in it that is included at buildroot install time (e.g. `python39-rpm-macros`).

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -4,7 +4,7 @@
 %global dynflow_sidekiq_service_name dynflow-sidekiq@
 %global rake /usr/bin/rake
 
-%global release 3
+%global release 4
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -506,7 +506,6 @@ Meta package to install asset pipeline support.
 Summary: Foreman plugin support
 Group: Development/Libraries
 Requires: %{name} = %{version}-%{release}
-Requires: %{name}-build = %{version}-%{release}
 Requires: rubygem(activerecord-nulldb-adapter)
 
 %description plugin
@@ -992,6 +991,9 @@ exit 0
 %systemd_postun %{name}.socket
 
 %changelog
+* Wed Nov 23 2022 Eric D. Helms <ericdhelms@gmail.com> - 3.6.0-0.4.develop
+- Drop requirement on foreman-build from foreman-plugin
+
 * Mon Nov 21 2022 Quirin Pamp <pamp@atix.de> - 3.6.0-0.3.develop
 - Ensure tmp directory exists in foreman_precompile_plugin
 


ehelms:
We can.

evgeni:
I like the central place, so I'm :+1: on this PR after branching

ekohl:
And something to consider: ATIX also rebuilds, but AFAIK doesn't use Koji. Doing this would also mean you need to modify mock to ensure it's present if needed.

evgeni:
> And something to consider: ATIX also rebuilds, but AFAIK doesn't use Koji. Doing this would also mean you need to modify mock to ensure it's present if needed.

Adding it here would be sufficient, no?
https://github.com/theforeman/foreman-packaging/blob/4d75a03f6fe3a8b12d0839d3da46075657af25b5/mock/el8.cfg#L5

cc @sbernhard -- can you bring this to the attention whoever is responsible for your build infrastructure?

evgeni:
FTR, talked to @quba42 on IRC, they don't use mock, but adjusting for this change is still a one-liner for them, so we're good :)

pcreech:
Small question... How hard would it be to get this included into 3.5 before it goes GA?

evgeni:
> Small question... How hard would it be to get this included into 3.5 before it goes GA?

I wouldn't be opposed to it, and a cursory look at the current `*-build` tags in Koji, no changes should be necessary, as they all already carry `foreman-build` in both the `build` and `srpm-build` groups (I checked foreman, plugins and katello). So once we have merged this and verified it actually works in nightly, I see no problem?

ehelms:
Soooo, are there any actions here or can this get review/merged?

quba42:
I have a question: Where can I find the file actually included in `foreman-build`? It is referenced in the spec file as `%{_sysconfdir}/rpm/macros.%{name}-dist`.

I am leaning towards cherry picking this change to our internal build, because I don't want the `fmX.Y` part in the version, but I first want to check whether omitting `foreman-build` has any other affects than just the extra version string.

In general I have no objections to this change. I do appreciate the heads up about changes that specifically affect "downstream builds". I would also like to keep learning about the nuts and bolts of how the upstream build works, since I am certain there are things I can improve in our downstream build, since my understanding of the upstream build is surface level.

evgeni:
The generation of the file happens here, based on the version of the RPM itself:
https://github.com/theforeman/foreman-packaging/blob/820296743650417df23bb884a79014aad3772d40/packages/foreman/foreman/foreman.spec#L801-L804

And yes, the package contains exactly one file with that one line.

ekohl:
Which can also be checked this way:
```console
$ wget -q https://yum.theforeman.org/releases/3.5/el8/x86_64/foreman-build-3.5.0-0.17.rc1.el8.noarch.rpm
$ rpm -ql foreman-build-3.5.0-0.17.rc1.el8.noarch.rpm
warning: foreman-build-3.5.0-0.17.rc1.el8.noarch.rpm: Header V4 RSA/SHA256 Signature, key ID 8f5ca95b: NOKEY
/etc/rpm/macros.foreman-dist
```

ehelms:
Rebased

ehelms:
@evgeni @ekohl Mind coming back to this one?

evgeni:
I already said I'm good with this approach in https://github.com/theforeman/foreman-packaging/pull/8699#issuecomment-1303552731 ;-)

ekohl:
I don't mind moving forward on this, but I'm too busy with other things to help out.

ehelms:
Thanks y'all, let's take it for test drive.

evgeni:
builds still have the `fm3_6` macro, so things are good? (http://koji.katello.org/koji/taskinfo?taskID=543468)

PR #8732, opened 2022-11-10
Update ruby-foreman-tasks to 7.1.0

adamruzicka:
CPs:
- 3.5

PR #8736, opened 2022-11-10
Update ruby-foreman-remote-execution to 8.1.0

adamruzicka:
CPs:
- 3.5


evgeni:
```
[2022-11-10T12:49:40.924Z]     foreman_remote_execution (= 8.1.0) was resolved to 8.1.0, which depends on

[2022-11-10T12:49:40.924Z]       foreman-tasks (>= 7.1.0) x86_64-linux
```

I guess you want to update the boundaries in `debian/contol` while the 7.1.0 tasks is still building from #8732

evgeni (approved):

PR #8746, opened 2022-11-10
Update rubygem-smart_proxy_ansible to 3.5.0

adamruzicka:
CPs:
- 3.5


evgeni:
I would have expected this to include changes for https://github.com/theforeman/smart_proxy_ansible/pull/66/, but I don't see them?
There is a `/usr/share/gems/gems/smart_proxy_ansible-3.5.0/bin/ansible-runner-environment.sh` inside the package, but no `/usr/libexec/foreman-proxy/ansible-runner-environment` that would point to it.

packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec
```diff
@@ -125,6 +116,7 @@ for i in ansible ansible_galaxy; do
 done
 
 ln -sv %{_root_sysconfdir}/foreman-proxy/ansible.cfg %{buildroot}%{foreman_proxy_dir}/.ansible.cfg
+ln -sv %{gem_instdir}/bin/ansible-runner-environment.sh %{buildroot}%{_libexecdir}/foreman-proxy/ansible-runner-environment
```
evgeni, line 119: I think you need a `mkdir -p %{buildroot}%{_libexecdir}/foreman-proxy/` first

evgeni (approved):

PR #8747, opened 2022-11-10
Update smart_proxy_ansible to 3.5.0

adamruzicka:
CPs:
- 3.5

### DIFF
--- a/plugins/smart_proxy_ansible/ansible.rb
+++ b/plugins/smart_proxy_ansible/ansible.rb
@@ -1,1 +1,1 @@
-gem 'smart_proxy_ansible', '3.4.1'
+gem 'smart_proxy_ansible', '3.5.0'
--- a/plugins/smart_proxy_ansible/changelog
+++ b/plugins/smart_proxy_ansible/changelog
@@ -1,3 +1,9 @@
+ruby-smart-proxy-ansible (3.5.0-1) stable; urgency=low
+
+  * 3.5.0 released
+
+ -- Adam Ruzicka <aruzicka@redhat.com>  Thu, 10 Nov 2022 16:54:44 +0000
+
 ruby-smart-proxy-ansible (3.4.1-1) stable; urgency=low
 
   * 3.4.1 released
--- a/plugins/smart_proxy_ansible/install
+++ b/plugins/smart_proxy_ansible/install
@@ -1,2 +1,3 @@
 debian/ansible.rb usr/share/foreman-proxy/bundler.d
 settings.d/ansible.yml etc/foreman-proxy/settings.d
+bin/ansible-runner-environment usr/libexec/foreman-proxy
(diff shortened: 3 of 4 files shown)


evgeni:
I would have expected this to include changes for https://github.com/theforeman/smart_proxy_ansible/pull/66/, but I don't see them?
There is a `/usr/bin/ansible-runner-environment.sh` inside the package, but I would have expected it to be `/usr/libexec/foreman-proxy/ansible-runner-environment`?

adamruzicka:
Not sure if this is the cleanest way of going about it, but everything seems to end up in the right place

evgeni:
Yeah, the only thing that I don't like is that it *also* ships the file (*twice*) in `/usr/bin`:

```
drwxr-xr-x root/root         0 2022-11-11 12:35 ./usr/bin/
-rwxr-xr-x root/root       118 2022-11-11 12:35 ./usr/bin/ansible-runner-environment
-rwxr-xr-x root/root       118 2022-11-11 12:35 ./usr/bin/ansible-runner-environment.sh
```

You could probably avoid that by doing something like
```
(RM) $(CURDIR)/debian/ruby-smart-proxy-ansible/usr/bin/ansible-runner-environment*
```
in line 23+

adamruzicka:
Huh, I haven't noticed it ends up in /usr/bin anyway. So instead of copying it all over the place, let's do it like in rpm world - letting it live in /usr/bin and placing a symlink into /usr/libexec?

plugins/smart_proxy_ansible/links
```diff
@@ -1 +1,2 @@
 var/lib/foreman-proxy/ansible usr/share/foreman-proxy/.ansible
+usr/bin/ansible-runner-environment.sh usr/libexec/foreman-proxy/ansible-runner-environment
```
evgeni, line 2: I don't think that file should be in `/usr/bin` under that name? it feels confusing to have it names `ansible-runner-<something>` while in reality it is foreman-proxy specific

adamruzicka: Do we want in in usr/bin under some name or should I just move it out of the way?

adamruzicka: Probably not, they might use it for "run ansible as foreman would", but considering they still need to provide a path to the environment file, we probably don't need to lose any sleep over it.

evgeni: Yeah, I think libexec-only is fine. For those who will want "run ansible as foreman" can do that from there.

evgeni:
```
% dpkg-deb -c ruby-smart-proxy-ansible_9999-3.5.0-plugins+scratchbuild+20221111135010_all.deb |grep envi
-rwxr-xr-x root/root       118 2022-11-11 14:50 ./usr/libexec/foreman-proxy/ansible-runner-environment
```

:partying_face:

evgeni (approved):

PR #8749, opened 2022-11-11
allow excluding certain gems from auto-updates

evgeni:
ostree_client is too special

### DIFF
--- a/list_updatable_packages
+++ b/list_updatable_packages
@@ -12,6 +12,8 @@ import requests
 import tempfile
 
 SUPPORTED_TEMPLATES = ('foreman_plugin', 'hammer_plugin', 'smart_proxy_plugin')
+# pulp_ostree_client: only prereleases exist and our tooling can't handle that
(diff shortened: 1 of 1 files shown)


ekohl (commented):
We already have `is_nightly` logic to skip nightlies. Isn't that a better place? So possibly implement something like `is_updateable` on `Spec` that is `self is_nightly or ...`.

evgeni:
Good idea!

evgeni:
Updated!

list_updatable_packages
```diff
@@ -12,6 +12,7 @@ import requests
 import tempfile
 
 SUPPORTED_TEMPLATES = ('foreman_plugin', 'hammer_plugin', 'smart_proxy_plugin')
+EXCLUDED_GEMS = ('pulp_ostree_client',)
```
ekohl, line 15:
```suggestion
# Only prereleases exist and our tooling can't handle that
EXCLUDED_GEMS = ('pulp_ostree_client',)
```

ekohl (approved):

PR #8751, opened 2022-11-11
Fixes #35743 - Introduce foreman-obsolete-packages

ekohl:
Sometimes packages are removed from our repositories and we want them to be cleaned up automatically. In case there is no replacement (like when a Foreman plugin is retired), there is nothing to obsolete it.

On EL7 we could abuse the tfm package, but no such thing exists on EL8. Fedora has fedora-obsolete-packages for this purpose. foreman-obsolete-packages should fulfill the same purpose.

### DIFF
--- a/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
+++ b/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
@@ -1,0 +1,24 @@
+Name: foreman-obsolete-packages
+Version: 1.0
+Release: 1%{?dist}
+License: MIT
+Summary: A package to obsolete retired packages
+URL: https://github.com/theforeman/foreman-packaging
+
+Obsoletes: rubygem-fog-google < 1.19.0-2
+
+%description
+This package exists only to obsolete other packages which need to be removed
+from the distribution for some reason.
+
+%prep
+
+%build
+
+%install
+
+%files
+
+%changelog
+* Fri Nov 11 2022 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> - 1.0-1
+- Initial package
--- a/comps/comps-foreman-el8.xml
+++ b/comps/comps-foreman-el8.xml
@@ -23,6 +23,7 @@
       <packagereq type="default">foreman-installer-katello</packagereq>
       <packagereq type="default">foreman-journald</packagereq>
       <packagereq type="default">foreman-libvirt</packagereq>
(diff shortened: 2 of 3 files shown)


packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
```diff
@@ -0,0 +1,24 @@
+Name: foreman-obsolete-packages
+Version: 3.6
+Release: 1%{?foremandist}%{?dist}
+License: Public Domain
+Summary: A package to obsolete retired packages
+URL: https://github.com/theforeman/foreman-packaging
+
+Obsoletes: rubygem-fog-google < 1.19.0-2
```
ekohl, line 8: @evgeni do we also want to add some Foreman plugins we recently removed, even if there is no DB cleanup?

evgeni: I'd prefer not to *right now*.

Doing so would leave cruft in the DB and with the pkg gone it's hard to say where the cruft originated from.

evgeni (approved):

packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
```diff
@@ -0,0 +1,24 @@
+Name: foreman-obsolete-packages
+Version: 3.6
+Release: 1%{?foremandist}%{?dist}
```
ekohl, line 3: This results in `3.6-1.fm3_6.el8`. Is `foremandist` needed? Or should we simplify version it 1.0 and rely on foremandist?

evgeni: I guess I'd do the later. There is not much in here that would warrant versionlocking with the Foreman version.

Otherwise you gotta add that to the post-branching-bump and get rather needless bumps as the source never changes.

ekohl: Or would we even go to just `1.0-1` without any Foreman version? That's probably fine since it's unlikely we'll obsolete things in stable branches.

evgeni: Also true.

evgeni:
```
% koji add-pkg --owner kojiadm foreman-nightly-el8 foreman-obsolete-packages
Adding 1 packages to tag foreman-nightly-el8
```

PR #8760, opened 2022-11-15
Update Rails stack to 6.1.7

ehelms:

PR #8761, opened 2022-11-15
Update Rails stack to 6.0.6

ehelms:


ehelms:
> https://github.com/theforeman/foreman/blob/3.3-stable/Gemfile#L10 should also be updated? :)

Blast it, I thought we had stopped being that restrictive.

evgeni:
> Although, I thought bundler_ext mean that it did not enforce the version constraints -- guess I'd have to check that.

That'll might be correct for RPM (which you're adjusting here), but we also could update Debian is the same go? :)

ehelms:
> Not to derail the conversation, but I think we should reconsider the use of `bundler_ext`. Nowadays we actually respect the Bundler/gemspec dependencies and patch them if needed. I'm not sure what else it actually does for us.

Sounds like we should file an issue so we can take a more dedicated look.

ehelms:
> That'll might be correct for RPM (which you're adjusting here), but we also could update Debian is the same go? :)

Pfffft, that's fair, having two different styles gets hard to remember sometimes. To @ekohl, not using bundler_ext, if that will work for us, would help keep that alignment to some degree. 

I shall also have to update the Foreman RPM which has it's dependencies locked down due to the Gemfile.

evgeni (approved):
Same comment as in #8760, this doesn't update the `Requires` lines, rendering the packages uninstallable in the end.

evgeni (changes requested):
Err, wrong button, of course

ekohl:
Given 3.3 is EOL, should we close this?

evgeni:
> Given 3.3 is EOL, should we close this?

Yes.

PR #8764, opened 2022-11-15
Fixes #35756 - do not use --disable-system-checks as it is gone

evgeni:
(cherry picked from commit 4c79bbeb487a9f5c7d94539821776d47b6d67654)


ekohl (approved):
Pending your response on https://github.com/theforeman/foreman-packaging/pull/8763#pullrequestreview-1180809013 but given this is already released I'd be ok with a more narrow patch here.

evgeni:
Merging as-is here and doing the better cleanup "there"

PR #8785, opened 2022-11-18
Update rubygem-foreman_maintain to 1.2.1

evgeni:


evgeni:
Going to merge this myself as it is a trivial bump that equally could have been done by automation like it was for debian: https://github.com/theforeman/foreman-packaging/pull/8769

Also, making sure next time it *will* be made by automation: https://github.com/theforeman/foreman-packaging/pull/8786

PR #8789, opened 2022-11-21
Fix foreman-plugin macro for (some) katello on EL8 builds

quba42:
We needed this for our internal Katello 4.3 packaging job on EL8 (but not on EL7).

Without this change, we had the following error (during the `%install` phase):
```
Errno::ENOENT: No such file or directory @ rb_sysopen - /root/rpmbuild/BUILD/katello-4.3.1/usr/share/foreman/tmp/katello_event_daemon.lock
```

Note that the build command we use is essentially just:
```
rpmbuild -bb /root/rpmbuild/SPECS/rubygem-katello.spec
```

I have no idea why this does not affect all builds everywhere equally, but I wanted to discuss, whether this macro change nevertheless just makes sense? I will open this as a Draft PR for now.

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -4,7 +4,7 @@
 %global dynflow_sidekiq_service_name dynflow-sidekiq@
 %global rake /usr/bin/rake
 
-%global release 2
+%global release 3
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -850,6 +850,7 @@ mv db/schema.rb.nulldb db/schema.rb \\
 ln -s %{nodejs_sitelib} node_modules \\
 export GEM_PATH=%%{buildroot}%%{gem_dir}:\${GEM_PATH:+\${GEM_PATH}}\${GEM_PATH:-\`ruby -e "print Gem.path.join(':')"\`} \\
 unlink tmp \\
+mkdir tmp \\
 \\
 rm \`pwd\`/config/initializers/encryption_key.rb \\
 rm \`pwd\`/config/database.yml \\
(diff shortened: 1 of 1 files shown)


evgeni:
I think the change in this PR is *technically* correct, but it's papering over the real issue that Katello tries to access that file even if it shouldn't.

quba42:
Just going to leave a link to the alternative/complementary fix on the Katello side: https://github.com/Katello/katello/pull/10362

evgeni:
The Katello fix was merged, but I guess this one is still a good idea.

@ekohl @ehelms what do you think?

quba42:
Personally I feel like this makes the macro more robust and more internally consistent (the other instance of `unlink` in the macro is also followed by a `mkdir`).

As a result I tend to agree with @evgeni that it makes sense to merge this even if the original issue is now also handled on the Katello side.

packages/foreman/foreman/foreman.spec
```diff
@@ -991,6 +992,9 @@ exit 0
 %systemd_postun %{name}.socket
 
 %changelog
+* Mon Nov 21 2022 Quirin Pamp <pamp@atix.de> - 3.6.0-0.3.develop
+- Fix foreman-plugin macro
```
ekohl, line 996: This could be a bit more descriptive:
```suggestion
- Ensure tmp directory exists in foreman_precompile_plugin
```

ekohl (commented):
I think this may make sense, but it got me wondering: is there normally a guarantee that it exists.

I did find https://github.com/theforeman/foreman/blob/eb15f5eab69325edbae1147a58c53e9ffe9e17d8/config/initializers/secret_token.rb#L6, which looks poorly written since it should use `FileUtils.mkdir_p` instead, but overall it looks like whenever tmp is used, it ensures the directory exists prior to usage.

Having said that: at runtime we do ensure the directory exists so I think this would be more consistent. But I'd appreciate a better commit message explaining what it's really doing and the reasoning.

packages/foreman/foreman/foreman.spec
quba42: I have applied this, and also made the commit message more descriptive.

evgeni:
The build issue is unrelated, and this LGTM!

evgeni:
[test rpm]

evgeni (approved):

PR #8799, opened 2022-11-23
Update foreman-maintain to 1.2.1

evgeni:


evgeni:
Based on the assessment in https://github.com/theforeman/foreman_maintain/pull/661#issuecomment-1324695440, I think it is safe to pick f-m 1.2.z into 3.4

ekohl (approved):

PR #8813, opened 2022-12-02
Update rubygem-foreman_virt_who_configure to 0.5.11

chris1984:
@ekohl here is the new virt-who plugin gem with the translations you added.


ekohl (approved):
Thanks!

PR #8814, opened 2022-12-02
Split the installer into multiple subpackages & add Pulp installer

ekohl:
Needs https://github.com/theforeman/foreman-installer/pull/822.

To test:

```
dnf install https://yum.theforeman.org/releases/latest/el8/x86_64/foreman-release.rpm https://yum.puppet.com/puppet7-release-el-8.noarch.rpm
cat <<EOF > /etc/yum.repos.d/pulp-installer.repo
[pulp-installer]
name=Pulp Installer
baseurl=https://ekohl.nl/pulp-installer
gpgcheck=0
enabled=1
EOF
dnf install foreman-installer-scenario-pulp
. /etc/profile.d/puppet-agent.sh
openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /etc/pki/tls/private/localhost.key -out /etc/pki/tls/certs/localhost.crt
foreman-installer --pulpcore-servername $HOSTNAME --pulpcore-worker-count 1 --pulpcore-content-service-worker-count 1 --pulpcore-api-service-worker-count 1 --pulpcore-django-secret-key eep4quah0iatheehah9gai3Joo2quooThotahreethahgh0AhD --pulpcore-postgresql-db-password hagoa4see6ziefahthie7peigeega9E
```

Note that in my repo I haven't built the latest changes yet so you still need to hack the repo yourself.

### DIFF
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,3 +1,10 @@
+<<<<<<< HEAD
+=======
+%global configdir %{_datadir}/%{name}/config
+%global parser_cache %{_datadir}/%{name}/parser_cache
+%global scenariodir %{_sysconfdir}/%{name}/scenarios.d
+
+>>>>>>> 95b6583a2 (Split the installer into multiple subpackages)
 %global release 1
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
@@ -7,22 +14,16 @@ Epoch:      1
 Version:    3.8.0
 Release:    %{?prerelease:0.}%{release}%{?prerelease:.}%{?prerelease}%{?nightly}%{?dist}
 Summary:    Puppet-based installer for The Foreman
-Group:      Applications/System
 License:    GPLv3+ and ASL 2.0
 URL:        https://theforeman.org
 Source0:    https://downloads.theforeman.org/%{name}/%{name}-%{version}%{?prerelease:-}%{?prerelease}.tar.bz2
 
 BuildArch:  noarch
 
-Requires:   curl
-Requires:   hostname
-Requires:   puppet-agent >= 7.0.0
-Requires:   rubygem(kafo) >= 6.5.0
-Requires:   rubygem(kafo) < 8.0.0
-Requires:   ruby(release)
+Requires: %{name}-scenario-foreman = %{epoch}:%{version}-%{release}
 
 BuildRequires: asciidoc
-BuildRequires: puppet-agent >= 7.0.0
+BuildRequires: puppet-agent >= 6.15.0
 BuildRequires: rubygem(rake)
 BuildRequires: rubygem(kafo) >= 6.5.0
 BuildRequires: rubygem(kafo) < 8.0.0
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -10,55 +11,92 @@ Epoch:      1
...
 
 BuildArch:  noarch
 
-Requires:   curl
-Requires:   hostname
-Requires:   puppet-agent >= 6.15.0
-Requires:   %{?scl_prefix}rubygem(kafo) >= 6.5.0
-Requires:   %{?scl_prefix}rubygem(kafo) < 7.0.0
-Requires:   %{?scl_prefix_ruby}ruby(release)
+Requires: %{name}-scenario-foreman = %{epoch}:%{version}-%{release}
```
ehelms, line 20: I was not expecting this part, that foreman-installer main package would have a requires on a scenario

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -85,61 +123,71 @@ foreman-installer --scenario katello --migrations-only > /dev/null
...
-%{_sysconfdir}/%{name}/scenarios.d/katello.migrations
-%config(noreplace) %attr(600, root, root) %{_sysconfdir}/%{name}/scenarios.d/katello.yaml
-%config(noreplace) %attr(600, root, root) %{_sysconfdir}/%{name}/scenarios.d/katello-answers.yaml
-%config(noreplace) %{_sysconfdir}/%{name}/scenarios.d/katello-migrations-applied
+%files scenario-foreman
+%{configdir}/foreman.migrations
+%config(noreplace) %attr(600, root, root) %{scenariodir}/foreman.yaml
+%config(noreplace) %attr(600, root, root) %{scenariodir}/foreman-answers.yaml
+%config(noreplace) %{scenariodir}/foreman-migrations-applied
+%{parser_cache}/foreman.yaml
```
ehelms, line 163: These 5 lines have a pattern to them seen in other scenarios -- is it worth a macro?

ekohl: I like that idea.

packages/foreman/foreman-installer/foreman-installer.spec
ekohl: I still debated this in my head. I first wrote it here for compatibility or upgrades. Initially when I wrote this PR I really wanted to create a pulp installer and the split part was mostly to have a decent base to work on.

Perhaps we should not have `foreman-installer` as a real package anymore at all, just make all the scenarios `Provide: foreman-installer = %{version}`?

packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -85,61 +123,71 @@ foreman-installer --scenario katello --migrations-only > /dev/null
...
+%{_datadir}/%{name}/config/foreman.hiera
+%{_datadir}/%{name}/hooks
+%{_datadir}/%{name}/modules
+%{_datadir}/%{name}/VERSION
+%dir %{parser_cache}
+%{_mandir}/man8/%{name}.8*
+
+%files katello-common
+%{_datadir}/%{name}/katello-certs
+%{parser_cache}/foreman-proxy-certs.yaml
```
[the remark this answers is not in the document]
ekohl: You may be right. I wasn't sure where we want to use this. Perhaps we shouldn't even name it an installer scenario and just make it `katello-certs` or something, since there's no separate scenario file it doesn't store any answers. It just provides a script to generate the certificates that could in theory also be used in some standalone server (like an offline CA).

ekohl:
Note to self: this is missing `%post` sections to run the migrations. There we should actually make sure they exit with code 0 (by using `|| :`). We may also need to redirect stderr to /dev/null since the logger complains about an unknown terminal if executed non-interactively via Ansible, or fix that in the installer/Kafo itself.

This came up in https://bugzilla.redhat.com/show_bug.cgi?id=2162704

packages/foreman/foreman-installer/foreman-installer.spec
ekohl: I couldn't figure out how to make that work. Room for a future improvement I guess.

packages/foreman/foreman-installer/foreman-installer.spec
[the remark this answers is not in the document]
ekohl: If we consider the downstream scenario then its own package makes sense.

PR #8822, opened 2022-12-06
rebuild katello meta package to include system-checks fix

evgeni:


ekohl (approved):
Should the commit message also use `Fixes #35756`?

evgeni:
I explicitly did not do that, as the commit doesn't contain any actual changes for that bug ;)

PR #8823, opened 2022-12-06
Update rubygem-dynflow to 1.6.8

adamruzicka:
CPs:
- 3.5


packages/foreman/rubygem-dynflow/rubygem-dynflow.spec
```diff
@@ -1,101 +1,60 @@
...
+%exclude %{gem_instdir}/.github
 %exclude %{gem_instdir}/.gitignore
-%exclude %{gem_instdir}/.github/install_dependencies.sh
-%exclude %{gem_instdir}/.github/workflows/ruby.yml
-%exclude %{gem_instdir}/.github/workflows/release.yml
 %exclude %{gem_instdir}/.rubocop.yml
 %exclude %{gem_instdir}/.rubocop_todo.yml
-%exclude %{gem_instdir}/Dockerfile
-%exclude %{gem_instdir}/docker-compose.yml
+%{gem_instdir}/Dockerfile
```
ekohl, line 53: I'd exclude this:
```suggestion
%exclude %{gem_instdir}/Dockerfile
```

packages/foreman/rubygem-dynflow/rubygem-dynflow.spec
```diff
@@ -1,101 +1,60 @@
...
 %license %{gem_instdir}/MIT-LICENSE
 %{gem_instdir}/extras
 %{gem_libdir}
-%{gem_instdir}/web
+%exclude %{gem_instdir}/web
```
ekohl, line 57: I have some vague recollection that this is actually needed for the web console to function. Are you sure this can be excluded?

packages/foreman/rubygem-dynflow/rubygem-dynflow.spec
```diff
@@ -1,101 +1,60 @@
...
-%exclude %{gem_instdir}/.github/install_dependencies.sh
-%exclude %{gem_instdir}/.github/workflows/ruby.yml
-%exclude %{gem_instdir}/.github/workflows/release.yml
 %exclude %{gem_instdir}/.rubocop.yml
 %exclude %{gem_instdir}/.rubocop_todo.yml
-%exclude %{gem_instdir}/Dockerfile
-%exclude %{gem_instdir}/docker-compose.yml
+%{gem_instdir}/Dockerfile
 %license %{gem_instdir}/MIT-LICENSE
 %{gem_instdir}/extras
```
ekohl, line 55: I wonder: should this be in the doc subpackage?

packages/foreman/rubygem-dynflow/rubygem-dynflow.spec
```diff
@@ -105,11 +64,15 @@ cp -pa .%{gem_dir}/* \
 %doc %{gem_instdir}/README.md
 %{gem_instdir}/Rakefile
 %doc %{gem_instdir}/doc
-%{gem_instdir}/dynflow.gemspec
+%doc %{gem_instdir}/docker-compose.yml
```
ekohl, line 67: I'd exclude this

adamruzicka:
Ah, ok, will do

packages/foreman/rubygem-dynflow/rubygem-dynflow.spec
adamruzicka: extras? They contain configuration for statsd exporter and sources of a tool for decoding msgpack in csv dumps. The tool is packaged as a separate package so the sources should probably be excluded here. Statsd exporter config could probably live in doc, I don't really feel strongly about this

packages/foreman/rubygem-dynflow/rubygem-dynflow.spec
```diff
@@ -1,99 +1,58 @@
...
+Requires: ruby >= 2.3.0
+BuildRequires: ruby >= 2.3.0
+BuildRequires: rubygems-devel
 BuildArch: noarch
-Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
 # end specfile generated dependencies
 
-# Optional dependency that we always pull in
-Requires: %{?scl_prefix}rubygem(statsd-instrument)
-Requires: dynflow-utils
```
evgeni, line 42: This seems like something we want to keep, no?

evgeni:
LGTM, but I'll leave the final review to Ewoud who did a more in-depth look yesterday.

ekohl (approved):
Thanks!

ekohl:
And I just realized I merged multiple commits instead of a single one.

PR #8834, opened 2022-12-12
Add isoinfo dependency for image extraction

bastian-src:
Refer to [Smart Proxy PR](https://github.com/theforeman/smart-proxy/pull/837).

Adds dependency `isoinfo` which is used to extract images.


evgeni:
Where is that package from? I don't see it in CentOS?

bastian-src:
ah thanks! I thought it would be only for debian `genisoimage`. Gonna change it accordingly!

bastian-src:
@evgeni this feature adds a folder  `/var/lib/foreman-proxy/tftp/system_images` to store OS iso images. 
Do you think it's okay to implement it with these static paths or is it better to create a variable for `{_sharedstatedir}/%{name}/tftp/system_images`? (I'll adapt the deb PR accordingly)

PR #8835, opened 2022-12-13
Send signal to only main process foreman-proxy on logrotate

adamruzicka:


adamruzicka:
We seem to already do this for debian

ekohl (approved):
We can probably move logrotate.conf to extras in smart-proxy.git, but that doesn't have to be done for this PR.

Could you create a Redmine for this so we can properly track its cherry picks?

PR #8843, opened 2022-12-15
Update rubygem-kubeclient and fog-kubevirt

dgoetz:
This is required for relaxing the dependency on http gem and to fix a vulnerability in kubeclient.

PR #8844, opened 2022-12-15
Bump rubygem-http to 5.1.0-1

laugmanuel:
needs: https://github.com/theforeman/foreman-packaging/pull/8845 and https://github.com/theforeman/foreman-packaging/pull/8847


laugmanuel:
This will break https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/plugins/rubygem-kubeclient/rubygem-kubeclient.spec#L26-L27

Due to https://github.com/elastic/apm-agent-ruby/issues/922 we need `http >= 4.4.1` for `elastic-apm`, which in turn is needed for `foreman_supervisory_authority`.

dgoetz:
Requires https://github.com/theforeman/foreman-packaging/pull/8843

PR #8846, opened 2022-12-15
Add rubygem-ffi-compiler package

laugmanuel:


laugmanuel:
@dgoetz do you know how to overcome the obal lint error? Looks to me like this happens because the gem is new to the repo?

dgoetz:
It has to be added to in package_manifest.yaml in section foreman_plugin_packages to fix the error.

Also for some reason the script changed the order in the comps when executed by you, which should be avoided.

laugmanuel:
I'll reorder the changes and add the package to the manifest.

dgoetz (approved):
Looks good to me, but not sure if anything else is required for a new package, so I hope for a second approval.

ekohl (approved):
I think obal should do that by itself now, but I'll check after merging if CI fails and otherwise fix it.

PR #8847, opened 2022-12-15
Add rubygem-llhttp-ffi package

laugmanuel:
needs https://github.com/theforeman/foreman-packaging/pull/8846


laugmanuel:
Looks like it breaks on a similar part of the build as I mentioned in Foreman community:
```sh
cp: cannot stat './usr/lib64/gems/ruby/llhttp-ffi-0.4.0/llhttp-ffi/*.so': No such file or directory
```
http://koji.katello.org/kojifiles/work/tasks/1736/541736/build.log

Does anyone have a hint on how to fix this? @dgoetz , @ekohl ?

dgoetz:
When looking at the lines before I see one .so being build in /builddir/build/BUILD/llhttp-ffi-0.4.0/usr/share/gems/gems/llhttp-ffi-0.4.0/ext, but when I tried to copy from there I only hit another issue which made me think there is likely something else wrong, but I did not come up with a solution.

laugmanuel:
@ekohl do you have any idea on how to fix this problem or know anyone who does?

packages/plugins/rubygem-llhttp-ffi/rubygem-llhttp-ffi.spec
```diff
@@ -0,0 +1,88 @@
...
+%gem_install
+
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+mkdir -p %{buildroot}%{gem_extdir_mri}/%{gem_name}
+cp -a .%{gem_extdir_mri}/gem.build_complete %{buildroot}%{gem_extdir_mri}/
+cp -a .%{gem_extdir_mri}/%{gem_name}/*.so %{buildroot}%{gem_extdir_mri}/%{gem_name}
```
ekohl, line 56: I'm not exactly sure why it's so inconsistent, but sometimes the .so file is in a directory, sometimes it's not. I think this is one example where it's not. If you grep for `gem_extdir_mri` you'll see there's not a clear preference for either one or the other.

This may work:

```suggestion
cp -a .%{gem_extdir_mri}/{gem.build_complete,*.so} %{buildroot}%{gem_extdir_mri}/
```

dgoetz: What worked as path for me:

```
cp -a %{buildroot}%{gem_instdir}/ext/x86_64-linux/*.so %{buildroot}%{gem_extdir_mri}/%{gem_name}
```

But then the tests fail:

```
+ ruby -e 'require '\''llhttp-ffi'\'''
/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:83:in `require': cannot load such file -- llhttp-ffi (LoadError)
	from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:83:in `require'
	from -e:1:in `<main>'
```

ekohl: https://www.rubydoc.info/gems/llhttp-ffi/0.4.0 suggests the name that you need to require `llhttp`. Our template doesn't really cover that so it needs some additional work.

laugmanuel: I've changed the line as @dgoetz suggested. The build seems to work.

I've also changed the check line to match the resulting gem name (`llhttp`) - however, the check still fails. I'm wondering if the `*.so` file is indeed included and moved to the correct location?

PR #8870, opened 2023-01-02
find out foreman_min_version based on plugin code

evgeni:

### DIFF
--- a/bump_rpm.sh
+++ b/bump_rpm.sh
@@ -90,6 +90,16 @@ if [[ $CURRENT_VERSION != $NEW_VERSION ]] ; then
 		elif [[ -n $TEMPLATE ]] ; then
 			echo "* Updating requirements"
 			gem2rpm -t $ROOT/gem2rpm/$TEMPLATE.spec.erb *.gem | $SCRIPT_DIR/update-requirements specfile - $SPEC_FILE
+			if [[ $TEMPLATE == foreman_plugin ]]; then
+				UNPACKED_GEM_DIR=$(mktemp -d)
+				gem unpack --target "$UNPACKED_GEM_DIR" *.gem
+				PLUGIN_LIB="${UNPACKED_GEM_DIR}/${GEM_NAME}-${NEW_VERSION}/lib"
+				REQUIRES_FOREMAN=$(grep --recursive --no-filename requires_foreman $PLUGIN_LIB |sed -E 's/[^0-9.]//g')
+				if [[ -n $REQUIRES_FOREMAN ]]; then
+					sed -i "/%global foreman_min_version/ s/foreman_min_version.*/foreman_min_version $REQUIRES_FOREMAN/" $SPEC_FILE
+				fi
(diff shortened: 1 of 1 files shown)


bump_rpm.sh
```diff
@@ -90,6 +90,16 @@ if [[ $CURRENT_VERSION != $NEW_VERSION ]] ; then
 		elif [[ -n $TEMPLATE ]] ; then
 			echo "* Updating requirements"
 			gem2rpm -t $ROOT/gem2rpm/$TEMPLATE.spec.erb *.gem | $SCRIPT_DIR/update-requirements specfile - $SPEC_FILE
+			if [[ $TEMPLATE == foreman_plugin ]]; then
+				UNPACKED_GEM_DIR=$(mktemp -d)
+				gem unpack --target "$UNPACKED_GEM_DIR" *.gem
+				PLUGIN_LIB="${UNPACKED_GEM_DIR}/${GEM_NAME}-${NEW_VERSION}/lib"
+				REQUIRES_FOREMAN=$(grep --recursive --no-filename requires_foreman $PLUGIN_LIB |sed -E 's/[^0-9.]//g')
```
ekohl, line 97: I always forget - is `.` something you need to escape in sed or don't?

ekohl (commented):
I was thinking about this too, but wonder about a dependency generator..

You would still get a failing build if you're building assets so BuildRequires without a version would be a bit more expensive (you see it later than you'd like and it's not greppable), but you don't need to update it manually.

Of course, that could be a version 2 of this. This is already an improvement over what we have. My expectation is that for a few plugins we may chase the plugin author to increase the minimum version.

bump_rpm.sh
evgeni: Yes.
Inside a character group (`[…]`) no, outside yes.
What you *need* to escape inside a character group is a `-` as it otherwise denotes a range (as `0-9` above) *unless* it's the last character of the group…

evgeni:
> I was thinking about this too, but wonder about a dependency generator..
> 
> You would still get a failing build if you're building assets so BuildRequires without a version would be a bit more expensive (you see it later than you'd like and it's not greppable), but you don't need to update it manually.
> 
> Of course, that could be a version 2 of this. This is already an improvement over what we have. My expectation is that for a few plugins we may chase the plugin author to increase the minimum version.

I always found writing dependency generators (for RPMs) hard.

So yes, we could do that, but I am kinda not motivated to ;-)

ekohl (approved):
Let's try this out and see. We know this will break if they use non-trivial requires, but I haven't seen those so we can deal with it once we get there.

evgeni:
> Let's try this out and see. We know this will break if they use non-trivial requires, but I haven't seen those so we can deal with it once we get there.

Yeah, all I've seen based on a quick `git grep` across the org is `>= something` and `> something`, the later is an "bug" on it's own (it uses `> 1.6`, which for Ruby means 1.6.1+ or 1.6.0.1+, but for RPM it's already satisfied by 1.6.0)

PR #8878, opened 2023-01-03
Correct auto-requires generation to allow bundled gems in Ruby 2.7

evgeni:

### DIFF
--- a/packages/foreman/rubygem-mail/rubygem-mail.spec
+++ b/packages/foreman/rubygem-mail/rubygem-mail.spec
@@ -3,7 +3,7 @@
 
 Name: rubygem-%{gem_name}
 Version: 2.8.0
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: Mail provides a nice Ruby DSL for making, sending and reading emails
 License: MIT
 URL: https://github.com/mikel/mail
@@ -15,6 +15,9 @@ BuildRequires: ruby >= 2.5
 BuildRequires: rubygems-devel
 BuildArch: noarch
 # end specfile generated dependencies
+
+# Prefer to consume net-smtp/net-imap/net-pop as a default gem
+Requires: ruby-default-gems < 3
 
 %description
 A really Ruby Mail handler.
@@ -30,6 +33,12 @@ Documentation for %{name}.
 
 %prep
 %setup -q -n  %{gem_name}-%{version}
+
+# On EL8 rubygem-net-smtp/imap/pop are bundled into ruby-libs package and
+# auto-generated dependencies will break dependency resolution
+%gemspec_remove_dep -g net-smtp
+%gemspec_remove_dep -g net-imap
+%gemspec_remove_dep -g net-pop
 
 %build
 # Create the gem as gem install only works on a gem file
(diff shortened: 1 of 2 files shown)


evgeni:
this now installs cleanly on an el8 box with `ruby:2.7` enabled.

packages/foreman/rubygem-mail/rubygem-mail.spec
```diff
@@ -16,6 +16,10 @@ BuildRequires: rubygems-devel
 BuildArch: noarch
 # end specfile generated dependencies
 
+# Prefer to consume net-smtp/net-imap/net-pop as a default gem
```
evgeni, line 19: There are no `Provides: bundled(rubygem-net-*)`, so I didn't add any `(Build)Requires` on that like we do for `racc`.

packages/foreman/rubygem-mail/rubygem-mail.spec
```diff
@@ -16,6 +16,10 @@ BuildRequires: rubygems-devel
...
+Requires: ruby-default-gems
```
ekohl, line 20: I think this should have a `< 3` to make sure they really are there.

packages/foreman/rubygem-mail/rubygem-mail.spec
```diff
@@ -16,6 +16,10 @@ BuildRequires: rubygems-devel
...
+BuildRequires: ruby-default-gems
```
ekohl, line 21: We don't run any tests so I'm not sure a build requires is actually needed?

packages/foreman/rubygem-mail/rubygem-mail.spec
evgeni: just mimicking what we have in `nokogiri.spec` ;)

https://github.com/theforeman/foreman-packaging/commit/0145c819e4f043896d9c6cf3e4b6733b6cd07a06

ekohl: That's because at least Ruby 3.1 still has racc in `ruby-default-gems` (and pretty sure 3.2 didn't drop it either).

evgeni: Watch me cursing at Ruby, Packaging, and moving things around. (Updated)

ekohl: It wouldn't be the first cursing in this area. I know I did my share.

evgeni:
```
[2023-01-03T10:30:52.309Z]       unresolved deps:
[2023-01-03T10:30:52.309Z]         ruby-default-gems < 3
```

what?

packages/foreman/rubygem-mail/rubygem-mail.spec
```diff
@@ -31,6 +34,12 @@ Documentation for %{name}.
 %prep
 %setup -q -n  %{gem_name}-%{version}
 
+# On EL8 rubygem-net-smtp/imap/pop are bundled into ruby-libs package and
```
ekohl, line 37: Just leaving it as a note here that if anyone ever gets back to this: technically it's not just EL8, but also anything else that packages Ruby < 3. Not that we support those, so close enough. We need to deal with this prior to EL9 anyway since there is no Ruby 2.7 there.

ekohl (changes requested):
I wonder why repoclosure is failing. I suspect its' modularity where it's picking up a newer version. Proper fix is to use modularity in repoclosure, but for now I'd exclude it here:
https://github.com/theforeman/foreman-packaging/blob/219ad5fabcb897fd0350cc0cd1889b1ac5526569/repoclosure/yum.conf#L54

ekohl (approved):

PR #8879, opened 2023-01-03
introduce a dynamic foreman:Depends instead of hard-coding versions

evgeni:

### DIFF
--- a/debian/bullseye/foreman/changelog
+++ b/debian/bullseye/foreman/changelog
@@ -1,3 +1,9 @@
+foreman (3.6.0-2) stable; urgency=low
+
+  * introduce a dynamic foreman:Depends instead of hard-coding versions
+
+ -- Evgeni Golov <evgeni@debian.org>  Tue, 03 Jan 2023 10:37:56 +0100
+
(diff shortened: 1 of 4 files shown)


evgeni:
This can be later used by something like:

```patch
diff --git plugins/ruby-foreman-tasks/debian/control plugins/ruby-foreman-tasks/debian/control
index 94e804ebb..331e0da9e 100644
--- plugins/ruby-foreman-tasks/debian/control
+++ plugins/ruby-foreman-tasks/debian/control
@@ -10,6 +10,6 @@ Package: ruby-foreman-tasks
 Architecture: all
 Conflicts: ruby-foreman-dynflow
 Replaces: ruby-foreman-dynflow
-Depends: ${misc:Depends}, bundler, foreman (>= 3.5.0~rc1), ruby-dynflow (>= 1.6.0)
+Depends: ${misc:Depends}, bundler, ${foreman:Depends}, ruby-dynflow (>= 1.6.0)
 Description: Tasks management engine for Foreman.
   Tasks management engine for Foreman. Gives you and overview of what's happening/happened in your Foreman instance.
```

ekohl (approved):

PR #8903, opened 2023-01-04
Update rubygem-foreman_azure_rm to 2.2.9

chris1984:
@ekohl here is the updated azure plugin gem


evgeni (approved):

evgeni:
@chris1984 does this need to be picked to any stable branches? if so, would you please open the cherry pick PRs?

PR #8926, opened 2023-01-11
Update rubygem-foreman_scc_manager to 2.2.0 Foreman 3.5

nadjaheitmann:
(cherry picked from commit d2a38afd5f86a545ab53f20378cdfb8d219653f8)


packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
```diff
@@ -113,6 +112,13 @@ cp -a .%{gem_dir}/* \
 %{gem_instdir}/test
 
 %changelog
+* Wed Jan 11 2023 Nadja Heitmann <nadjah@atix.de> 2.2.0-1
+- Update to 2.2.0
+- Fix permission issue for new installations
+
+* Mon Nov 14 2022 Bernhard Suttner <suttner@atix.de> 2.1.0-1
+- Update to 2.1.0
```
ekohl, line 120: This was a different commit. What we always do is cherry pick both commits so you get the correct log/references. You can feed `git cherry-pick -x` multiple commits to make this easier.

ekohl (changes requested):

packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
nadjaheitmann: > This was a different commit. What we always do is cherry pick both commits so you get the correct log/references. You can feed `git cherry-pick -x` multiple commits to make this easier.

Thanks for the hint, changed the PR accordingly! Is it still possible to CP something to 'rpm/3.3 branch'?

ekohl (approved):
Thanks! Notice how this also included the proper BuildRequires update so they're in sync? That's precisely the reason we cherry pick as much as possible.

packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
ekohl: Yes. We still have the pipelines in place even though we officially ended support for 3.3.

nadjaheitmann:
> Thanks! Notice how this also included the proper BuildRequires update so they're in sync? That's precisely the reason we cherry pick as much as possible.

Yeah, thanks, the feature is very handy, indeed - I like it!

PR #8929, opened 2023-01-12
Add a foreman-discovery-image package

ekohl:
While this could be built in Koji, this actually takes an easier approach and downloads the image from downloads.theforeman.org. It should be noted that the paths there aren't easy to consume, since it includes a git hash. Possibly that's a bug in our release pipeline and it may be easier to fix that than to keep the complex part here.

This is now a draft to open up discussion. The benefit of having a package is that we can deal with this in puppet-foreman_proxy. That means we don't have to download a file manually. Satellite also has a package and today the user is told to manually install it after the installer ran, but unified instructions means a single installer run is sufficient.

Obviously a Debian equivalent should be introduced. That's not done right now, but should be fairly trivial. That's another benefit of avoiding a Koji livecd build: all upstream uses the same image. The potential move from Koji to COPR into account is also taken into account.


ekohl:
One consideration: this RPM will be like 500 MB, which is significant. Especially if we start pushing updates. On the Debian side we have a shared repo between Debian and Ubuntu versions and for RPM we now only have EL8, but if we add EL9 that also doubles.

evgeni:
I think if we package it, it should be built on koji, not downloaded from somewhere.

ekohl:
Taking a look at the future, does it mean we'll build it on EL9 once Foreman is built on EL9 or is that a separate setting? Is this something we could achieve? And if we move to COPR, is it also supported?

evgeni:
I don't think copr can do live image builds like koji can.

@ehelms ?

xsuchy:
> I don't think copr can do live image builds like koji can.

No, Copr cannot create a live image. The preferred tool nowadays is ImageBuilder https://www.redhat.com/sysadmin/rhel-image-builder

PR #8934, opened 2023-01-16
Update rubygem-foreman_discovery to 21.0.5

Ron-Lavi:


evgeni (approved):

evgeni:
I guess the same should be done for 3.4, and Debian?

PR #8935, opened 2023-01-18
Fixes #35743 - Introduce foreman-obsolete-packages

evgeni:
Sometimes packages are removed from our repositories and we want them to be cleaned up automatically. In case there is no replacement (like when a Foreman plugin is retired), there is nothing to obsolete it.

On EL7 we could abuse the tfm package, but no such thing exists on EL8. Fedora has fedora-obsolete-packages for this purpose. foreman-obsolete-packages should fulfill the same purpose.

(cherry picked from commit 3ec631787eb63699112b08f5c4f7d40d83975ba9)

### DIFF
--- a/comps/comps-foreman-el8.xml
+++ b/comps/comps-foreman-el8.xml
@@ -23,6 +23,7 @@
       <packagereq type="default">foreman-installer-katello</packagereq>
       <packagereq type="default">foreman-journald</packagereq>
       <packagereq type="default">foreman-libvirt</packagereq>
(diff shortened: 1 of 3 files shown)


evgeni:
```console
% koji add-pkg --owner kojiadm foreman-3.5-el8 foreman-obsolete-packages
Adding 1 packages to tag foreman-3.5-el8
```

ekohl (approved):
CI is still running, but I doubt it'll fail.

PR #8940, opened 2023-01-19
RHEL cleanup

evgeni:
this cleans up RHEL 7 support from several (but not all) packages.

### DIFF
--- a/packages/foreman/foreman-selinux/foreman-selinux.spec
+++ b/packages/foreman/foreman-selinux/foreman-selinux.spec
@@ -21,7 +21,7 @@
 
 %define moduletype apps
 
-%global release 1
+%global release 2
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -54,11 +54,7 @@ Requires(postun):   /usr/sbin/semodule
 Requires(postun):   /sbin/restorecon
 %{?systemd_requires}
 
-%if 0%{?rhel} == 7
-Requires(post):     policycoreutils-python
-%else
 Requires(post):     policycoreutils-python-utils
-%endif
 
 %description
 SELinux policy module for Foreman
@@ -68,12 +64,7 @@ SELinux policy module for Foreman
 
 %build
 # determine distribution name and version
-%if 0%{?rhel} >= 6
 %define distver rhel%{rhel}
-%endif
-%if 0%{?fedora} >= 18
-%define distver fedora%{fedora}
-%endif
 
 # build policy
 for selinuxvariant in %{selinux_variants}; do
@@ -145,11 +136,7 @@ Requires(post):     selinux-policy-targeted
 Requires(postun):   /usr/sbin/semodule
 Requires(postun):   /sbin/restorecon
 
-%if 0%{?rhel} == 7
-Requires(post):     policycoreutils-python
-%else
 Requires(post):     policycoreutils-python-utils
-%endif
 
 %description -n foreman-proxy-selinux
 SELinux policy module for Foreman Proxy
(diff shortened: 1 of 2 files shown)


evgeni:
repoclosure fails because the tooling is stupid, f-release is available for el9, so it tries to also do closure for the others on el9, which obviously fails

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -29,7 +29,6 @@ Source0:  foreman.repo
 Source1:  foreman-plugins.repo
 Source2:  foreman.gpg
 Source5:  foreman-client.repo
-Source6:  qpid-copr.repo
```
ekohl, line 32: I had the impression that this was still used for the client side. Is it not for katello-agent?

If you're dropping it, you should also remove the file from git.

packages/foreman/foreman-selinux/foreman-selinux.spec
```diff
@@ -68,12 +64,7 @@ SELinux policy module for Foreman
 
 %build
 # determine distribution name and version
-%if 0%{?rhel} >= 6
 %define distver rhel%{rhel}
-%endif
-%if 0%{?fedora} >= 18
-%define distver fedora%{fedora}
```
ekohl, line 75: Unrelated, but this reminded me https://github.com/theforeman/foreman-selinux/pull/148 is still open.

packages/foreman/foreman-selinux/foreman-selinux.spec
```diff
@@ -68,12 +64,7 @@ SELinux policy module for Foreman
 
 %build
 # determine distribution name and version
-%if 0%{?rhel} >= 6
 %define distver rhel%{rhel}
```
ekohl, line 67: Not a blocker for this, but it may not be a bad idea to enhance foreman-selinux to determine this automatically.

packages/foreman/foreman-release/foreman-release.spec
evgeni: and this is why we have code review :see_no_evil:

packages/foreman/foreman-selinux/foreman-selinux.spec
evgeni: Technically, since https://github.com/theforeman/foreman-selinux/commit/0c2d1f0972fca43a008d20260f5983bef4d6ed5e, we don't need to define that at all, as we don't have any distro specific macros.

ekohl: When I looked at that I wondered if the shared make rules would use it. However, that appears to use distro in a different way:
```console
$ rg DISTRO /usr/share/selinux/devel
/usr/share/selinux/devel/include/Makefile
56:ifneq ($(DISTRO),)
57:	M4PARAM += -D distro_$(DISTRO)

/usr/share/selinux/devel/include/build.conf
3:DISTRO ?= redhat
```

So I'm leaning to dropping it: https://github.com/theforeman/foreman-selinux/pull/150

ekohl:
In https://github.com/theforeman/foreman-packaging/pull/8814 I also did some cleanup of the installer spec. Perhaps you can pick that in here too?

packages/foreman/foreman-selinux/foreman-selinux.spec
ekohl: @evgeni shall we merge that PR and then update this PR?

evgeni: I think that's a plan!

evgeni:
rebased and updated for the `DISTRO` change!

ekohl (approved):

PR #8942, opened 2023-01-20
Obsolete rubygem-google-api-client < 0.33.2-3

evgeni:
Mark the package as noarch

google-api-client was removed in 6f25ff6adffd1290b7e6b533038ac3d51ae4163b


evgeni:
Theoretically the `rubygem-google-api-client` obsolete needs to happen during the 3.3→3.4 upgrade, but I don't think we want to backport the new package there? (and then we'd need to drop the fog obsolete)

ekohl (approved):

PR #8841, opened 2022-12-14
Update rubygem-fog-vsphere to 3.6.0

chris1984:
@ekohl here is the packaging pr for the new fog-vsphere gem

PR #8965, opened 2023-01-22
Update rubygem-fog-vsphere to 3.6.0

Ron-Lavi:
new version of fog-vsphere is already updated in foreman develop: https://github.com/theforeman/foreman/blob/a2c20eab3c78b7bfb78157477d7f667f0a7448f2/bundler.d/vmware.rb#L2


ekohl:
There already is https://github.com/theforeman/foreman-packaging/pull/8841, which is close. Then https://github.com/theforeman/foreman-packaging/pull/8856 was automatically opened for this.

PR #8969, opened 2023-01-23
Bump Hammer CLI Katello to 1.7.2

chris1984:


chris1984:
@pcreech here is the hammer version bump

PR #8994, opened 2023-01-26
Update foreman and drop rbvmomish

ehelms:
When rubygem-rbvmomi2 was introduced, it also added rbvmomish which generates a conflict on disk when both are installed (https://community.theforeman.org/t/errors-installing-foreman-vmware/32072). I am opting to use the newer version rather than the older version.


ekohl (changes requested):
Actually, we drop and obsolete `rubygem-rbvmomi`. I think they share the same namespace so you'd have more conflicts.

ehelms:
Updated -- what threw me off was that the foreman spec file still required rbvmomi. This change has two commits, one that updates the foreman spec dependencies, and a second that removes rbvmomi.

packages/foreman/foreman/foreman.spec
```diff
@@ -121,6 +121,7 @@ Requires: rubygem(jwt) < 3.0
 Requires: rubygem(graphql) >= 1.13.0
 Requires: rubygem(graphql) < 1.14.0
 Requires: rubygem(graphql-batch)
+Requires: rubygem(rss)
```
evgeni, line 124: This is ruby 3+ only and shouldn't be here (yet)

evgeni:
`rubygem-rbvmomi2` needs to obsolete `rubygem-rbvmomi`

evgeni:
I've took the liberty to implement these changes to get this going.

ekohl (approved):

PR #8996, opened 2023-01-27
Backport foreman-obsolete-packages to 3.4

evgeni:

### DIFF
--- a/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
+++ b/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
@@ -1,0 +1,24 @@
+Name: foreman-obsolete-packages
+Version: 1.0
+Release: 1%{?dist}
+License: MIT
+Summary: A package to obsolete retired packages
+URL: https://github.com/theforeman/foreman-packaging
+
+Obsoletes: rubygem-fog-google < 1.19.0-2
+
+%description
+This package exists only to obsolete other packages which need to be removed
+from the distribution for some reason.
+
+%prep
+
+%build
+
+%install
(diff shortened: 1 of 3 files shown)


packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
```diff
@@ -0,0 +1,33 @@
...
+%prep
+
+%build
+
+%install
+
+%files
+
+%changelog
+* Fri Jan 27 2023 Evgeni Golov - 1.1-0.1
```
ekohl, line 24: This kind of makes me wonder if we need to introduce foremandist in the release. Now the changelog goes back in version numbers, which isn't great. On the other hand, foremandist wouldn't solve that either.

Perhaps we should just drop the changelog and introduce it as version 0.1.

evgeni: I thought about 0.9, but yeah, anything below 1.0 would work.

Really, don't mind any specific version number, as long as we can smoothen the 3.3→3.4 upgrade.

ekohl: Let's go with 0.9 then.

evgeni: Like this?

ekohl (approved):

evgeni:
```console
% koji add-pkg --owner kojiadm foreman-3.4-el8 foreman-obsolete-packages
Adding 1 packages to tag foreman-3.4-el8
```

PR #9000, opened 2023-01-30
Fix DHCP infoblox smart proxy plugin dependencies

timdeluxe:
because it is broken with the 3.5 foreman release, which ships ruby-infoblox in version 3.x.x


timdeluxe:
My mistake, this was not the cherry picked commit

PR #9001, opened 2023-01-30
Fix DHCP infoblox smart proxy plugin dependencies

timdeluxe:
because it is broken with the 3.5 foreman release, which ships ruby-infoblox in version 3.x.x

### DIFF
--- a/plugins/smart_proxy_dhcp_infoblox/control
+++ b/plugins/smart_proxy_dhcp_infoblox/control
@@ -10,6 +10,6 @@ XS-Ruby-Versions: all
 Package: ruby-smart-proxy-dhcp-infoblox
 Architecture: all
 XB-Ruby-Versions: ${ruby:Versions}
-Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter, foreman-proxy (>= 1.17.0~rc1), ruby-infoblox (>= 2.0.4), ruby-infoblox (<< 3)
+Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter, foreman-proxy (>= 1.17.0~rc1), ruby-infoblox (<< 4)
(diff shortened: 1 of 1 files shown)


timdeluxe:
@ekohl Sorry, i fighted with my IDE a bit. Hope it is correct now?

ekohl (approved):
Thanks!

ekohl:
Apologies for not checking, but it looks like you picked the commit from your branch. 5721b737ba417bd30eed5a3d480d13b6f9e1b898 was the commit that was really merged.

Another thing both @evgeni and I missed is that there is no changelog message. That's how we determine if something needs to be built, and why it hasn't.

The best path forward is to open a PR to `deb/develop` with a changelog entry. `./scripts/changelog.rb` should make that easier. Once merged, we can pick it to `deb/3.5` as well.

timdeluxe:
I am sorry. This whole cherry picking is somewhat confusing to me. While do the changelog change.
But first, what do we do now with https://github.com/theforeman/foreman-packaging/commit/5721b737ba417bd30eed5a3d480d13b6f9e1b898? Are we going to revert that somehow (and redo correctly after)?

ekohl:
No, reverting it for a commit message is not worth it. I'd just roll forward with it now: create the changelog message and try to do better in the future. These things happen and I'm sure I've made the same mistake. If GitHub allowed true fast forward merge (rather than rebasing it) it would probably be easier. We also don't do merge commits. Otherwise it wouldn't have been an issue either.

This stuff is always a bit tricky. We have discussed implementing https://github.com/apps/patchback to automate it, but have yet to find the time to actually do so.

timdeluxe:
Okay, understood.
I now raised https://github.com/theforeman/foreman-packaging/pull/9005 and once merged i will try to do the cherry-pick of that better.

PR #9024, opened 2023-02-09
bump foreman vendor and builder to v12

MariaAga:


ekohl (approved):
Let's find out what breaks.

PR #9035, opened 2023-02-12
Update rubygem-foreman_ansible to 10.4.0

nofaralfasi:


nofaralfasi:
@theforeman/packaging do you know why I'm getting this error? SCAP is not related to ansible as far as I know.

evgeni:
It is not, correct. But the repoclosure check runs on the whole repository, and that fails. No need to worry!

evgeni (approved):

nofaralfasi:
Thanks you Evgeni, and **please** enjoy your Sunday :)

PR #9037, opened 2023-02-12
Updated ruby-foreman-ansible to 10.3.0

nofaralfasi:


evgeni:
Given https://github.com/theforeman/foreman-packaging/pull/9036 is merged, this one can be closed

nofaralfasi:
> Given #9036 is merged, this one can be closed?

Yes yes, you are right, there is no need for that. Closing.

PR #9038, opened 2023-02-13
Fixes #36086 - Allow openscap 1.3.7

evgeni:

### DIFF
--- a/packages/plugins/rubygem-openscap/rubygem-openscap.spec
+++ b/packages/plugins/rubygem-openscap/rubygem-openscap.spec
@@ -4,11 +4,11 @@
 
 %global gem_name openscap
 %global min_openscap_version 1.2.9
-%global max_openscap_version 1.3.7
+%global max_openscap_version 1.3.8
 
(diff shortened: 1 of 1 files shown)


evgeni:
Repoclosure complains about the `rubygem-openscap` in the repo, not the updated one.

ekohl (approved):
This doesn't break non-stream, does it?

evgeni:
> This doesn't break non-stream, does it?

Correct, non stream just has an older version, and we have a range defined ;)

lpramuk:
Can't we just say `openscap < 1.4` to have less work ?

ekohl:
That's a valid question. I first assumed this compiled something, but now I question that. It loads it here:
https://github.com/OpenSCAP/ruby-openscap/blob/1aa57405bd277022e91c3763b8b7fb27e0662c70/lib/openscap/openscap.rb#L7

So now I"m questioning it. Why do we even need this?

I recall we had this for EL7 & EL8 compatibility, but I think we can simplify it. I've opened https://github.com/theforeman/foreman-packaging/pull/9051 to see if that works.

PR #9043, opened 2023-02-14
Update rubygem-foreman_ansible to 11.1.0

nofaralfasi:


evgeni:
```
DEBUG util.py:634:  No matching package to install: 'npm(@theforeman/builder) < 11.0.0'
```

we recently bumped builder and vendor to 12.x, so you need to update the deps.

evgeni (approved):

PR #9047, opened 2023-02-15
Set foreman_min_version on gem creation

ekohl:
Like 267e95a2d756ccfe8060aa33e35ea6b88258d99e did for bump_rpm.sh, this sets the foreman_min_version value for add_gem_package.sh and removes it from the TODO in the template. This reduces the manual work needed to add a Foreman plugin package.

I should note I didn't try this out yet.


evgeni (approved):

ekohl:
Made some fixes, now it actually works. Mind having a last look?

evgeni (approved):

PR #9075, opened 2023-02-22
consider hammer dependencies when bumping

evgeni:

### DIFF
--- a/list_updatable_packages
+++ b/list_updatable_packages
@@ -189,6 +189,8 @@ def main() -> None:
         find_gemfile_lock_specs('foreman-develop-source-release', 'foreman'),
         find_gemfile_lock_specs('smart-proxy-develop-source-release', 'foreman'),
         find_gemfile_lock_specs('katello-master-source-release', 'katello'),
+        find_gemfile_lock_specs('hammer-cli-master-source-release', 'foreman'),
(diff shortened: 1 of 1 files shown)


list_updatable_packages
```diff
@@ -189,6 +189,9 @@ def main() -> None:
         find_gemfile_lock_specs('foreman-develop-source-release', 'foreman'),
         find_gemfile_lock_specs('smart-proxy-develop-source-release', 'foreman'),
         find_gemfile_lock_specs('katello-master-source-release', 'katello'),
+        find_gemfile_lock_specs('hammer-cli-master-source-release', 'foreman'),
+        find_gemfile_lock_specs('hammer-cli-foreman-master-source-release', 'foreman'),
+        find_gemfile_lock_specs('foreman-installer-develop-source-release', 'foreman'),
```
ekohl, line 194: I don't have the PR at hand, but :-1: on the installer. They are mostly development dependencies and really there's only Kafo in https://github.com/theforeman/foreman-installer/blob/develop/Gemfile that we care about. Given how static that is and the potential problems from the other development dependencies I'd keep it out.

evgeni: dropped

ekohl (approved):

PR #9076, opened 2023-02-22
don't downgrade packages

evgeni:
we sometimes end up with the automation trying to downgrade packages, which is almost always wrong, so let's try to prevent it.

this *also* prevents conflicts from failing the workflow, which can be seen as a good thing, or a bad thing

### DIFF
--- a/list_updatable_packages
+++ b/list_updatable_packages
@@ -4,6 +4,7 @@ import json
 import os
 import subprocess
 import sys
+from packaging.version import Version
 from itertools import chain
 from pathlib import Path
 from typing import Dict, Generator, Iterable, Mapping, Optional, Tuple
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I wonder if we should silently ignore this. I'm more in the camp of explicit failures. Can't we set new_version to `DOWNGRADE` if we detect it?

evgeni:
It depends…

In this particular case I was prompted by the fact that `foreman-installer` has a *development* dependency on `librarian-puppet` which in turn pulls in `puppet_forge` and that an old `faraday` (see also the convo in `#voxpupli` about a new librarian release).

In other cases it might be useful to have a failure, yes.

ekohl:
Yes, but we shouldn't even attempt to look at `foreman-installer`. See https://github.com/theforeman/foreman-packaging/pull/9075#discussion_r1114110143

evgeni:
Updated to raise errors.

However, as I "fixed" librarian yesterday, my testcase is gone -.-

ekohl (approved):
Thanks!

PR #9084, opened 2023-02-22
Update rubygem-fog-proxmox to 0.15.0

ekohl:
I opened https://github.com/fog/fog-proxmox/pull/83 to remove the files we `%exclude`.


packages/plugins/rubygem-fog-proxmox/rubygem-fog-proxmox.spec
```diff
@@ -1,115 +1,89 @@
...
-%exclude %{gem_instdir}/Gemfile.lock
+%doc %{gem_instdir}/CHANGELOG.md
+%exclude %{gem_instdir}/Gemfile
 %doc %{gem_instdir}/README.md
-%{gem_instdir}/Rakefile
+%exclude %{gem_instdir}/Rakefile
 %doc %{gem_instdir}/docs
-%doc %{gem_instdir}/examples
-%{gem_instdir}/fog-proxmox.gemspec
+%{gem_instdir}/examples
```
dgoetz, line 79: examples was flagged as doc before, is this change intended?

ekohl: No, good catch. Updating.

PR #9095, opened 2023-02-23
Updated ruby-foreman-ansible to 7.1.5

nofaralfasi:


nofaralfasi:
Since we no longer support this version, I think it is better to close this one.

PR #9101, opened 2023-02-23
Update rubygem-foreman_virt_who_configure to 0.5.13

chris1984:

PR #9114, opened 2023-02-28
Update katello to 4.8.0

Odilhao:

### DIFF
--- a/packages/katello/katello-repos/katello-repos.spec
+++ b/packages/katello/katello-repos/katello-repos.spec
@@ -3,12 +3,12 @@
 %define repo_dir %{_sysconfdir}/yum.repos.d
 %define repo_dist %{dist}
 
-%global prereleasesource nightly
+%global prereleasesource rc1
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global release 2
 
 Name:           katello-repos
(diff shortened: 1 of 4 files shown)


ekohl:
This fails because the 4.8 tags do not exist yet in Katello. It should have been branched, but that hasn't happened. This means we need to take additional care when branching and untag any packages that have incorrect versions.

Odilhao:
[test rpm]

Odilhao:
@ekohl this repoclosure failure is expect for Katello?

ekohl:
I suspect you haven't mashed the repositories yet, so it can't do repoclosure. But before you do that, please fix the branching.

If we look at http://koji.katello.org/koji/buildinfo?buildID=74232 then that shouldn't be in the 4.8 tag. Essentially https://github.com/theforeman/foreman-packaging/commit/73513de4cdd08c2cb7d930c3e83d0b3003e93192 and https://github.com/theforeman/foreman-packaging/commit/fdf43f104abee2f177657f3d4466270f078b99ee, though https://github.com/theforeman/foreman-packaging/pull/9101 is open to include that in 4.8.

Odilhao:
Untagged `katello-repos-4.9-0.1.nightly.el8` running one scratch for #9101

PR #9115, opened 2023-02-28
Rpm/3.5 hammer cli 3.5.1

ofedoren:
Supported Versions:

 * 3.5
 
 
I've used `bump_rpm.sh` as always, but I'm not so sure if it's okay that all scl-related information got deleted. I've checked h-c-katello and it seems this information is still needed. What did I wrong? Also, I'm not sending this as a CP since I don't think we need this to be in `develop` branch, since 3.6 already includes all the new stuff. 3.5 needs only one feature, that's why I did a minor release for 3.5 branch.


ekohl (changes requested):
> I've used `bump_rpm.sh` as always, but I'm not so sure if it's okay that all scl-related information got deleted.

SCL stuff is only needed for EL7 support, but in 3.5 we don't care about that anymore.

> Also, I'm not sending this as a CP since I don't think we need this to be in `develop` branch

The apipie-bindings update sould be a cherry pick, which is nicer in knowing that they're in sync.

ofedoren:
Right, thanks, @ekohl, rebased with CP.

ehelms (approved):

PR #9122, opened 2023-03-03
Use Python 3.11 for Ansible in EL8.8

evgeni:
https://kojihub.stream.centos.org/koji/buildinfo?buildID=30909 is the build switching to 3.11, did not land in a stream compose yet

### DIFF
--- a/packages/plugins/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
+++ b/packages/plugins/ansible-collection-theforeman-foreman/ansible-collection-theforeman-foreman.spec
@@ -4,7 +4,7 @@
 %global collection_name foreman
 %global collection_directory %{_datadir}/ansible/collections/ansible_collections/%{collection_namespace}/%{collection_name}
 
-%global release 1
+%global release 2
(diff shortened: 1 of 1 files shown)


evgeni:
I don't understand why repoclosure fails, the relevant Ansible build is not yet in the repos (and if it would, then so would be Python 3.11).
The build itself works just fine on a recent CS8 container, pulling in Python 3.9 bits as expected right now.

ekohl:
We still run repoclosure on EL7 so perhaps it's some bug in the DNF version there?

evgeni:
> We still run repoclosure on EL7 so perhaps it's some bug in the DNF version there?

That's possible.

evgeni:
[test rpm]

evgeni:
progress, now only python3.11-requests is missing

evgeni:
Seems there is still some chaos in the placement of 3.11 rpms: https://issues.redhat.com/browse/CS-1502

evgeni:
[test rpm]

evgeni:
it is finally green!

PR #9131, opened 2023-03-06
compare_deb_repo cherry-picks for 3.4

evgeni:


evgeni:
merging without review, picks only

PR #9157, opened 2023-03-09
rubygem-bigdecimal is a requirement of sequel

sbernhard:


packages/foreman/rubygem-sequel/rubygem-sequel.spec
```diff
@@ -16,6 +16,9 @@ BuildRequires: rubygems-devel
 BuildArch: noarch
 # end specfile generated dependencies
 
+Requires: rubygem(bigdecimal)
+BuildRequires: rubygem(bigdecimal)
```
ekohl, line 20: We don't run any tests in this gem, so no need for a build dependency
```suggestion
```

packages/foreman/rubygem-sequel/rubygem-sequel.spec
```diff
@@ -16,6 +16,9 @@ BuildRequires: rubygems-devel
 BuildArch: noarch
 # end specfile generated dependencies
 
+Requires: rubygem(bigdecimal)
```
ekohl, line 19: Could you link to the upstream PR?
```suggestion
# https://github.com/jeremyevans/sequel/pull/2003
Requires: rubygem(bigdecimal)
```

packages/foreman/rubygem-sequel/rubygem-sequel.spec
```diff
@@ -66,6 +69,10 @@ find %{buildroot}%{gem_instdir}/bin -type f | xargs chmod a+x
 %doc %{gem_instdir}/doc
 
 %changelog
+* Thu Mar 09 2023 Bernhard Suttner <suttner@atix.de> 5.66.0-2
+- bigdecimal is a requirement of sequel
+
+- Update to 5.66.0
```
ekohl, line 75: I think this may be a copy paste error:
```suggestion
```

I can recommend `rpmdev-bumpspec -c '- bigdecimal is a requirement of sequel' rubygem-sequel.spec` if you're doing it manually now.

ekohl (changes requested):

ekohl (approved):

PR #9158, opened 2023-03-09
Add git-core dep on rubygem-git and remove from Foreman plugins

ekohl:
Prior to this the consumers of rubygem-git specified a git dependency, but this makes rubygem-git do the right thing. It also moves to `git-core`, which as the name implies, is a minimal installation. It saves a lot of perl dependencies, but is sufficient for normal operation.

This mirrors what Fedora does.

Moving from git to git-core was raised in https://bugzilla.redhat.com/show_bug.cgi?id=2175822.


ekohl:
foreman_acd fails to build due to the builder dependency.

ehelms:
[test rpm]

packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
```diff
@@ -35,7 +32,6 @@ BuildRequires: rubygem(git)
 
 # start package.json devDependencies BuildRequires
 BuildRequires: npm(@theforeman/builder) >= 10.1.0
-BuildRequires: npm(@theforeman/builder) < 11.0.0
```
ekohl, line 38: This is a manual change, but it allows to build it.

ekohl:
[test rpm]

ekohl:
rubygem-git was already done. Templates is being updated in https://github.com/theforeman/foreman-packaging/pull/9432. Not sure about ACD, but I'm going to close this and whenever ACD gets an update we'll deal with it.

PR #9168, opened 2023-03-10
Place hammer_cli_katello-1.8.0.gem correctly

ekohl:
The source file belongs in the same directory as the spec file.

Fixes: e36b01270638fc53a1fb38ef6374fe95abcab82f


chris1984:
Sorry about that, honest mistake. Usually I am better than that with these prs.

PR #9181, opened 2023-03-13
Add rubygem-optimist as requirement for rubygem-rbvmomi2

Odilhao:
Rubygem(optimist) is already being added as Rich [Dependency](koji.katello.org/koji/rpminfo?rpmID=428852) , this ensures that the optimist is added in the chain of dependencies when it's not added to the compose.

### DIFF
--- a/packages/foreman/rubygem-rbvmomi2/rubygem-rbvmomi2.spec
+++ b/packages/foreman/rubygem-rbvmomi2/rubygem-rbvmomi2.spec
@@ -3,7 +3,7 @@
 
 Name: rubygem-%{gem_name}
 Version: 3.6.0
-Release: 2%{?dist}
+Release: 3%{?dist}
 Summary: Ruby interface to the VMware vSphere API
 License: MIT
 URL: https://github.com/ManageIQ/rbvmomi2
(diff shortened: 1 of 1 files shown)


packages/foreman/rubygem-rbvmomi2/rubygem-rbvmomi2.spec
```diff
@@ -3,14 +3,16 @@
...
+Release: 3%{?dist}
 Summary: Ruby interface to the VMware vSphere API
 License: MIT
 URL: https://github.com/ManageIQ/rbvmomi2
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
 # start specfile generated dependencies
 Requires: ruby >= 2.4.1
+Requires: rubygem(optimist) >= 3.0
+Requires: rubygem(optimist) < 4.0
```
evgeni, line 15: Please put this outside the "specfile generated dependencies" block, as otherwise it will be removed on the next automatic bump of the spec.

Odilhao: Ack, done

evgeni:
(it seems "the other setup" is missing a `gather_backend = "dnf"` in the relevant config)

Odilhao:
gather_backend = "dnf" solved the problem with rich dependencies

PR #9182, opened 2023-03-13
Adds rbvmomi2 requirement to fog-vsphere

Odilhao:
Same logic as #9181 , rbvmomi2 already comes as Rich Dep for fog-vsphere, this ensures that the package comes when it's not added to the comps.


evgeni:
Same comment as in #9181 - this should not be necessary and the setup that this is failing is in broken.

PR #9208, opened 2023-03-21
Add rubygem-fugit

adamruzicka:
Preparation for https://github.com/theforeman/foreman-tasks/pull/710


packages/plugins/rubygem-et-orbi/rubygem-et-orbi.spec
```diff
@@ -0,0 +1,65 @@
...
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+%files
+%dir %{gem_instdir}
+%doc %{gem_instdir}/CREDITS.md
+%license %{gem_instdir}/LICENSE.txt
+%{gem_instdir}/Makefile
```
ekohl, line 51: Should this be excluded?

packages/plugins/rubygem-fugit/rubygem-fugit.spec
```diff
@@ -0,0 +1,69 @@
...
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+# start specfile generated dependencies
+Requires: ruby
+BuildRequires: ruby
+BuildRequires: rubygems-devel
+BuildRequires: rubygem(et-orbi) >= 1.2.7
+BuildRequires: rubygem(et-orbi) < 2.0.0
+BuildRequires: rubygem(raabro) >= 1.4.0
+BuildRequires: rubygem(raabro) < 2.0.0
```
ekohl, line 19: I wonder why this is being generated. I thought we only did that for gems with native extensions, which fugit doesn't appear to have.

It now means the other packages need to exist before this gem can be packaged.

packages/plugins/rubygem-raabro/rubygem-raabro.spec
```diff
@@ -0,0 +1,65 @@
...
+%install
+mkdir -p %{buildroot}%{gem_dir}
+cp -a .%{gem_dir}/* \
+        %{buildroot}%{gem_dir}/
+
+%files
+%dir %{gem_instdir}
+%doc %{gem_instdir}/CREDITS.md
+%license %{gem_instdir}/LICENSE.txt
+%{gem_instdir}/Makefile
```
ekohl, line 51: Should this be excluded?

packages/plugins/rubygem-fugit/rubygem-fugit.spec
ekohl: Ah ok. So what we now have with EL8 is that runtime requirements are automatically added. Build time dependencies are there if we need to actually run the package in steps like `%build` or `%check`, but here we don't so you can leave it out.

evgeni (commented):
Three nitpicks:
* All `.spec` files have a empty line at the end, which IMHO shouldn't be there
* I'd argue `CREDITS.md` should be in the `doc` subpackage as it is similar to `AUTHORS` and `CONTRIBUTORS` files we put there
* The changes to `comps-foreman-el8.xml` seem unrelated (your locale seems to sort those differently than the previous one, even tho I thought we're pinning the locale here: https://github.com/theforeman/foreman-packaging/blob/2b094ae6c1be4a50989ed32afcce20273ee7c86b/comps_doc.sh#L6)

But yeah, much nitpick, not much else to complain about ;)

comps/comps-foreman-plugins-el8.xml
```diff
@@ -231,12 +234,14 @@
...
       <packagereq type="default">rubygem-ed25519-doc</packagereq>
       <packagereq type="default">rubygem-elastic-apm-doc</packagereq>
+      <packagereq type="default">rubygem-et-orbi-doc</packagereq>
       <packagereq type="default">rubygem-faraday-cookie_jar-doc</packagereq>
       <packagereq type="default">rubygem-faraday-detailed_logger-doc</packagereq>
       <packagereq type="default">rubygem-faraday_middleware-doc</packagereq>
       <packagereq type="default">rubygem-ffi-compiler-doc</packagereq>
       <packagereq type="default">rubygem-fog-kubevirt-doc</packagereq>
       <packagereq type="default">rubygem-fog-proxmox-doc</packagereq>
+      <packagereq type="default">rubygem-foreman-tasks-doc</packagereq>
```
evgeni, line 244: huh, missed that one in the previous review.
there is already an entry for `rubygem-foreman-tasks-doc` in L279, so that one seems superfluous?

adamruzicka: Huh, not sure how it got there. Removed

evgeni (approved):

ekohl (approved):

evgeni:
```console
% koji add-pkg --owner kojiadm foreman-plugins-nightly-el8 rubygem-et-orbi rubygem-fugit rubygem-raabro
Adding 3 packages to tag foreman-plugins-nightly-el8
```

PR #9226, opened 2023-03-22
Rails 6.1.7.3 for 3.6

evgeni:


evgeni:
nightly passes with this (https://ci.theforeman.org/job/foreman-nightly-rpm-pipeline/1828/) so I am kinda confident so will 3.6

PR #9240, opened 2023-03-23
Update katello-host-tools to 4.2.3 with cleanups

ekohl:
This takes https://github.com/theforeman/foreman-packaging/pull/9187 and applies a lot of cleanups on top of it.

I've used OBS to verify that it builds SLES 15.4 and it does (with warnings), but fails on older SLES versions. I think it has to do with Python versions, but I'm not going to dive deeper into that.

@pcreech I haven't tested this on EL6 & EL7 yet (relying on CI at least for EL7), but I think this addresses my concerns.


pcreech:
The bigger issue with older sles versions, from what I saw with the build you referenced in the other PR, was that the py_build macros weren't supported.  I went and looked at the other successful katello-host-tools builds, and all of them use the {__python} setup.py build syntax instead.  Since we've been using py_build for a year now, and haven't had any complaints yet, i'm also inclined to address that outside of these updates.

ACK

pcreech (approved):

PR #9249, opened 2023-03-27
add .standard.yml to ignores

evgeni:


ekohl (approved):
I had to look it up, but this comes from https://github.com/testdouble/standard so a development config.

PR #9254, opened 2023-03-28
katello-host-tools updates for building on el9, el6

pcreech:
Fix building on el6 for downstream consumption.
Build agent for el9 as well


pcreech:
The repoclosure should be "fine" as it's expected since katello-agent on el9 isn't shipped in foreman client repos.

ehelms (approved):

PR #9305, opened 2023-04-21
Don't install package and repository upload on DNF installs

evgeni:

### DIFF
--- a/packages/client/katello-host-tools/katello-host-tools.spec
+++ b/packages/client/katello-host-tools/katello-host-tools.spec
@@ -37,7 +37,7 @@
 
 Name: katello-host-tools
 Version: 4.2.3
-Release: 5%{?dist}
+Release: 6%{?dist}
 Summary: A set of commands and yum plugins that support a Katello host
 Group:   Development/Languages
 %if 0%{?suse_version}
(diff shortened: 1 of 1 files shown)


jturel:
Subman started uploading this data itself on el8 but (_someone_ asked for it) still needed a way to force the uploads so  we left the CLI tools. I have a distant memory of adding dnf plugin commands to do the same so (iirc) there is a way to force the upload without these scripts - doubt anyone ever used them

At one point I think some REX or katello-agent package actions (or something) called these commands after performing whatever the action was - this may break those if that's still the case but was probably a hack anyway :smile:

ekohl:
So is this still a valid PR or should it be closed?

ehelms:
@evgeni I lean towards closing this as it has sat stagnant and we do not have a clear answer to if it would break things.

evgeni:
we don't ship the configs on DNF installs:
https://github.com/theforeman/foreman-packaging/blob/02ad412427a7b723c25b6bdfa515d0a74918c1b3/packages/client/katello-host-tools/katello-host-tools.spec#L362-L365

so it can't (and won't) work today (as it's not needed).

ehelms:
> %config(noreplace) %{plugins_confdir}/package_upload.conf 
>  %config(noreplace) %{plugins_confdir}/enabled_repos_upload.conf

Touche. Care to rebase this PR?

evgeni:
rebased

ekohl:
Looks like there's another conflict by now.

evgeni:
:disappointed:

ehelms (approved):

PR #9320, opened 2023-05-04
regenerate ansible, katello and rex specs

evgeni:

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -1,28 +1,24 @@
 # template: foreman_plugin
-%{?scl:%scl_package rubygem-%{gem_name}}
-%{!?scl:%global pkg_name %{name}}
-
+%global gem_name katello
+%global plugin_name katello
 %global foreman_min_version 3.7
 %global foreman_max_version 3.8
-%global plugin_name katello
-%global gem_name katello
 %global prereleasesource pre.master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global mainver 4.9.0
-%global release 1
+%global release 2
 
-Name:    %{?scl_prefix}rubygem-%{gem_name}
-Summary: Content and Subscription Management plugin for Foreman
-
+Name: rubygem-%{gem_name}
 Version: %{mainver}
 Release: %{?prerelease:0.}%{release}%{?prerelease}%{?nightly}%{?dist}
-Group:   Applications/Systems
+Summary: Content and Subscription Management plugin for Foreman
 License: GPLv2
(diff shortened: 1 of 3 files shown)


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -206,7 +189,13 @@ done
 %{gem_instdir}/package.json
 %{gem_instdir}/webpack
 
+%posttrans
+%{foreman_plugin_log}
```
evgeni, line 193: Funny that we forgot to add that one here -.-

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -1,28 +1,24 @@
...
+%global release 2
 
+Name: rubygem-%{gem_name}
 Version: %{mainver}
 Release: %{?prerelease:0.}%{release}%{?prerelease}%{?nightly}%{?dist}
-Group:   Applications/Systems
-License: GPLv2
-URL:     https://theforeman.org/plugins/katello
+Summary: Content and Subscription Management plugin for Foreman
+License: GPL-2.0
```
ekohl, line 15: Isn't `GPLv2` the correct Fedora tag? I'm not sure we can use SPDX codes in EL8 (though Fedora has adopted them).

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -52,8 +48,6 @@ BuildRequires: rubygem(gettext_i18n_rails)
 BuildRequires: rubygem(apipie-rails) >= 0.5.14
 BuildRequires: rubygem(fx) < 1.0
 BuildRequires: rubygem(pg)
-BuildRequires: rubygem(runcible) >= 2.13.0
```
ekohl, line 55: We can finally drop this, let's not forget the follow up PR to remove it.

packages/katello/rubygem-katello/rubygem-katello.spec
evgeni: maybe. https://github.com/theforeman/foreman-packaging/blob/rpm/develop/licenses.json doesn't list the "no later" version, so probably that's why it got reset to the value in https://github.com/Katello/katello/blob/2cad75d2d309afe0061f2f656d80409feff6c180/katello.gemspec#L10

packages/katello/rubygem-katello/rubygem-katello.spec
evgeni: https://github.com/theforeman/foreman-packaging/pull/9321

ekohl (approved):

evgeni:
repoclosure is confused because the packages are in different repos actually, ignoring

PR #9324, opened 2023-05-04
Bump rubygem-puma to 6.2.2-1 and rubygem-puma-status to 1.6-1

evgeni:


ekohl:
Looking into the changes, this makes `wait_for_less_busy_worker` the default and can help scale under load. Would be interesting to see the performance impact.

evgeni:
needs https://github.com/theforeman/foreman/pull/9699 (and a regen of the deps in `foreman.spec` afterwards)

ekohl (approved):

evgeni:
I've incorporated the foreman changes in here

PR #9335, opened 2023-05-05
add ruby-clamp and update kafo and installer to compatible versions

evgeni:
`hammer` is actually still missing, thus still Draft.


ekohl:
@evgeni given we're branching, is it time to get this in?

evgeni:
@ekohl is there a hammer-clie 3.7.0 release that is compatible with the new clamp?

evgeni:
the hammer part is in https://github.com/theforeman/foreman-packaging/pull/9422 (I didn't want to rebuild all here just for hammer)

ekohl (approved):

PR #9338, opened 2023-05-05
kafo: Regen RPM spec based on latest template

evgeni:


ekohl:
I opened https://github.com/theforeman/kafo/pull/354 and want to release a new major version. Then on the update I'd prefer to clean up all the SCL stuff.

ekohl:
https://github.com/theforeman/foreman-packaging/pull/9339 includes a bump to version 7.0.0.

PR #9345, opened 2023-05-09
Update rubygem-foreman_scap_client to 0.5.1

ares:


packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
```diff
@@ -1,62 +1,39 @@
...
 %description doc
 Documentation for %{name}.
 
 %prep
-gem unpack %{SOURCE0}
-
-%setup -q -D -T -n  %{gem_name}-%{version}
-
-gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
+%setup -q -n  %{gem_name}-%{version}
```
evgeni, line 32: the template we use "expects" el8 or newer tooling, but this gem is also built for el7 (and el6 downstream? i don't know) and thus can't use the updated template.

evgeni:
https://github.com/theforeman/foreman-packaging/pull/9346 only does the version bump, without any regeneration

ares:
closing this then, thank you

PR #9347, opened 2023-05-10
Sunsetting foreman_column_view because functionality being integrated in Foreman itself

dgoetz:
According to https://community.theforeman.org/t/sunsetting-foreman-column-view/33326

I see no need for backporting.

Requires: https://github.com/theforeman/puppet-foreman/pull/1119


evgeni (approved):

evgeni:
```console
% koji untag-build --all foreman-plugins-nightly-el8 rubygem-foreman_column_view
% koji block-pkg foreman-plugins-nightly-el8 rubygem-foreman_column_view
```

PR #9349, opened 2023-05-10
Update rubygem-apipie-rails to 1.0.0

evgeni:


ekohl (approved):
:+1: whenever the Foreman side is also done.

evgeni:
Foreman side: https://github.com/theforeman/foreman/pull/9704

PR #9382, opened 2023-05-16
Update rubygem-foreman_ansible to 12.0.0

nofaralfasi:


ekohl (approved):
Repoclosure failed on the old foreman_ansible and this should fix it.

PR #9397, opened 2023-05-17
Fix for https://projects.theforeman.org/issues/36405

korekhov:

### DIFF
--- a/plugins/ruby-foreman-templates/debian/changelog
+++ b/plugins/ruby-foreman-templates/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-templates (9.3.0-3) stable; urgency=low
+
+  * 9.3.0-3 released
+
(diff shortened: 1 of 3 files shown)


ekohl:
ok to test

plugins/ruby-foreman-templates/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-foreman-templates (9.3.0-3) stable; urgency=low
+
+  * 9.3.0-3 released
```
ekohl, line 3: You can change this to something like "update bundled dependencies for security fixes"

ekohl (commented):
Thanks for this. Could you squash it into a single commit? If you start the commit message with `Fixes #36405 - ` then our automation should pick it up.

korekhov:
Let me start over and create a clean PR.

PR #9400, opened 2023-05-18
Fixes #36405

korekhov:


evgeni:
ok to test

evgeni:
I've also opened https://github.com/theforeman/foreman-packaging/pull/9402 to update it in the RPM packages. Thanks!

evgeni (approved):

PR #9410, opened 2023-05-20
allow bumping puppet-agent-* deps

evgeni:

### DIFF
--- a/list_updatable_packages
+++ b/list_updatable_packages
@@ -98,12 +98,13 @@ def find_gemfile_lock_specs(ci_job: str, subdir: str) -> Generator[Tuple[Spec, s
     base_path = Path('packages') / subdir
 
     for gem in parse_gemfile_lock(fetch_gemfile_lock(url)):
-        package_name = f'rubygem-{gem["name"]}'
-        path = base_path / package_name / f'{package_name}.spec'
-        if path.is_file():
-            spec = Spec(path)
-            if spec.is_updateable:
-                yield spec, gem['version']
+        for package_prefix in ['rubygem', 'puppet-agent']:
+            package_name = f'{package_prefix}-{gem["name"]}'
+            path = base_path / package_name / f'{package_name}.spec'
+            if path.is_file():
+                spec = Spec(path)
+                if spec.is_updateable:
+                    yield spec, gem['version']
 
 
 def find_packaged_specs() -> Generator[Tuple[Spec, None], None, None]:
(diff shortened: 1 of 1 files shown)


list_updatable_packages
```diff
@@ -191,6 +192,7 @@ def main() -> None:
     specs = chain(
         find_packaged_specs(),
         find_gemfile_lock_specs('foreman-develop-source-release', 'foreman'),
+        find_gemfile_lock_specs('foreman-installer-develop-source-release', 'foreman'),
```
evgeni, line 195: @ekohl I recall you had said in the past that this is not worth it, but I think with the puppet-agent gems, it actually is?

ekohl: The problem is that we have many gems that are actually irrelevant for packaging. Then some things end up pulling in older gems, for example because of the old rubocop version we have (which we should update).

You can also see that we don't even pin `puppet-strings` in our `Gemfile` (https://github.com/theforeman/foreman-installer/blob/c6422b45f1bb6e1af8e9792857411b077fe306ac/Gemfile#L12) so if that's the only thing then I wonder if there's a better way.

What we can also consider is a more advanced solution where we can inherit things.

Breaking it down we have:
* Pure runtime: needed to run the installer.
  * puppet-agent (can be AIO)
  * Kafo (+ deps)
* Build time:
  * puppet-strings (to build the parser cache)
  * librarian-puppet (to download the bundled modules)
* Development / testing (irrelevant, so not specified)

Build time is actually split into two. librarian-puppet is needed to build the source tarball while puppet-strings is needed to build the cache. The reason is that it only caches items for classes included in the answer file(s). Technically we could generate one huge cache file for all the classes, but that makes the cache bigger and will slow down runtime.

So that makes me think about building some common file + Gemfile with just the ones we use in packaging (kafo, puppet-strings) and generate a lockfile for that. Then also include the common bits in the generic Gemfile.

Any thoughts on that?

evgeni: Right now, the script finds:

```
packages/foreman/puppet-agent-oauth 1.1.0
packages/foreman/puppet-agent-yard 0.9.34
packages/foreman/puppet-agent-puppet-strings 4.0.0
packages/foreman/puppet-agent-rgen 0.9.1
```

The alternative would be to add something like `template: puppet` to the specs and allow the script to always bump that to "latest", but I am not sure I like that.

I think having a `Gemfile.lock` with `--without test:development` would be the right thing (and would also apply to the other projects, actually, it's just that there its less of an issue).

ekohl: I'm not sure `--without` really ignores those dependencies. For example, if you have a git repo listed in your development group bundler will still clone it to determine the dependencies. Even if you specified `--without development`. That's why I suggested separate files. But there may be different ways (like looking at ENV vars).

evgeni: Yeah, I shouldn't have said `--without` but "a lockfile built in a way that it would not list the unwanted deps" (= what you suggested, without implying an implementation)

PR #9431, opened 2023-05-24
Update rubygem-foreman_templates to 9.4.0

ofedoren:
Supported Versions:

 * Nightly
 * 3.7


ofedoren:
Thanks, @ekohl, aaand since we're at it, since what version (3.5?) we should not use scl macros anymore? Just to ensure "my" packages are up to date and can be installed on el8 (since I presume we don't support el7 anymore).

ekohl:
It is indeed when we dropped EL7 to we stopped caring about EL7 support, which we dropped in Foreman 3.4.

ofedoren:
> Perhaps it's better to drop all the SCL macros while we're at it. I think the easiest way is to change `bump_rpm.sh` to force a regeneration. We already do that here:
> 
> https://github.com/theforeman/foreman-packaging/blob/dcec57b122e2074db760566e122fbf8c53bd4e29/bump_rpm.sh#L83
> 
> 
> I'd suggest to essentially make that `if true` locally just to proceed faster. Then manually add the git requirement bits.

Well, that didn't work well, it removed a lot of stuff instead of updating.

I've regenerated the spec by "re-adding" it via `./add_gem_package` and now it seems OK? I'm not sure it's even legal though...

packages/plugins/rubygem-foreman_templates/rubygem-foreman_templates.spec
```diff
@@ -1,49 +1,36 @@
...
+
+Name: rubygem-%{gem_name}
+Version: 9.4.0
+Release: 1%{?foremandist}%{?dist}
+Summary: Template-syncing engine for Foreman
+License: GPLv3
+URL: https://github.com/theforeman/foreman_templates
+Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+
 # start specfile generated dependencies
```
ekohl, line 14: It's better to explicitly require 1.18.0+ since that contains the dependency on `git-core`:
```suggestion
# Duplicates the autogenerated one so we can ensure git-core is pulled in
Requires: rubygem(git) >= 1.18.0

# start specfile generated dependencies
```

packages/plugins/rubygem-foreman_templates/rubygem-foreman_templates.spec
```diff
@@ -100,11 +73,10 @@ cp -pa .%{gem_dir}/* \
 %{gem_instdir}/app
 %{gem_instdir}/config
 %{gem_instdir}/db
-
 %{gem_libdir}
-%exclude %{gem_instdir}/webpack
+%{gem_instdir}/locale
```
ekohl, line 77: Yay, translations :)

ekohl (commented):
Yes, that's also one that works pretty well. If the end result is correct, I consider it legal.

packages/plugins/rubygem-foreman_templates/rubygem-foreman_templates.spec
ofedoren: Right, updated.

PR #9437, opened 2023-05-25
Bump Katello related packages to their 4.10 versions

wbclark:


ekohl:
This depends on https://ci.theforeman.org/job/katello-master-source-release/2200/ but I wonder why it's been running for 2 days. That isn't right. I'm cancelling and kick off a fresh build.

ekohl:
`bump_rpm.sh` doesn't work with nightly packages for which it needs to retrieve the source via Jenkins, which pretty much all these packages are.

pcreech:
[test rpm]

ekohl (approved):
Repoclosure failure is from the old RPM.

PR #9440, opened 2023-05-25
Remove SCL macros for rubygem-foreman_webhooks 3.1.0

ofedoren:
@ekohl, is this cherry-pickable even though it implies second(?) version of 3.1.0?..

### DIFF
--- a/packages/plugins/rubygem-foreman_webhooks/rubygem-foreman_webhooks.spec
+++ b/packages/plugins/rubygem-foreman_webhooks/rubygem-foreman_webhooks.spec
@@ -1,19 +1,15 @@
 # template: foreman_plugin
-%{?scl:%scl_package rubygem-%{gem_name}}
-%{!?scl:%global pkg_name %{name}}
-
 %global gem_name foreman_webhooks
 %global plugin_name webhooks
 %global foreman_min_version 3.3
 
-Summary:    Plugin for Foreman that allows to configure Webhooks
-Name:       %{?scl_prefix}rubygem-%{gem_name}
-Version:    3.1.0
-Release:    1%{?foremandist}%{?dist}
-Group:      Applications/Systems
-License:    GPLv3
-URL:        https://github.com/theforeman/foreman_webhooks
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-foreman_webhooks/rubygem-foreman_webhooks.spec
```diff
@@ -89,26 +69,27 @@ cp -pa .%{gem_dir}/* \
...
 %{gem_libdir}
-%exclude %{gem_cache}
 %exclude %{gem_instdir}/package.json
 %exclude %{gem_instdir}/webpack
-%exclude %{gem_instdir}/Rakefile
+%exclude %{gem_cache}
 %{gem_spec}
 %{foreman_bundlerd_plugin}
-%{foreman_assets_plugin}
-%{foreman_assets_foreman}
```
ofedoren, line 99: Also, same question: is this supposed to be removed?..

evgeni: No, not at all.

It fails to build exactly because of these:
```
   /usr/share/foreman/public/assets/foreman_webhooks
   /usr/share/gems/gems/foreman_webhooks-3.1.0/public/assets/foreman_webhooks/foreman_webhooks.json
```

Wonder what made the template think that's not required

evgeni: Ah, your plugin doesn't have any asset pipeline assets (only webpack), so these paths are almost empty, but they exist...

Probably best to just put it back (and make the template include it `if has_assets || has_weback`)

evgeni:
> is this cherry-pickable even though it implies second(?) version of 3.1.0?..

I am not Ewoud, but yes, absolutely pickable into 3.7

evgeni (approved):

ekohl:
It is indeed pickable. That's exactly what the release field is for: indicating changes in packaging, but not the original version.

PR #9442, opened 2023-05-26
include foreman_assets_* in files also when webpack only

evgeni:
for both, sprockets and webpack based, assets we use plugin:assets:precompile, which generates the content under foreman_assets_* regardless whether sprockets assets are present or not.

for plugins that only ship webpack based assets, this means that the template will generate a file list that is insufficient and rpmbuild will complain about unpackaged files. not sure those files are really needed as we don't have sprockets in that case, but its easier to just include them for now.


ekohl (commented):
Somehow foreman_azure_rm was building without it (prior to https://github.com/theforeman/foreman-packaging/pull/9634). Did that never work then?

evgeni:
well, it was never compiling assets, to it was not needed?

ehelms (approved):

PR #9443, opened 2023-05-26
Remove SCL macros for hammer-cli-foreman develop

ofedoren:
ASMR (Another SCL Macros Removal).

I'm not sure if I did it right (did it manually this time due to different output from `./add_gem_package`), it was based on the spec we have for hammer core.

Also, I'm now curious how it did work for few Foreman releases, because SCL stuff is here for quite a long time. I'm also afraid now to do the same stuff for all hammer plugins :melting_face:

### DIFF
--- a/packages/foreman/rubygem-hammer_cli_foreman/rubygem-hammer_cli_foreman.spec
+++ b/packages/foreman/rubygem-hammer_cli_foreman/rubygem-hammer_cli_foreman.spec
@@ -1,18 +1,14 @@
 # template: hammer_plugin
-%{?scl:%scl_package rubygem-%{gem_name}}
-%{!?scl:%global pkg_name %{name}}
-
 %global gem_name hammer_cli_foreman
 %global plugin_name foreman
 
-%global release 1
+%global release 2
 %global prereleasesource pre.develop
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 
-%{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
-%global hammer_confdir %{_root_sysconfdir}/hammer
+%global hammer_confdir %{_sysconfdir}/hammer
 
-Name: %{?scl_prefix}rubygem-%{gem_name}
+Name: rubygem-%{gem_name}
 Version: 3.8.0
 Release: %{?prerelease:0.}%{release}%{?prerelease}%{?nightly}%{?dist}
 Summary: Foreman commands for Hammer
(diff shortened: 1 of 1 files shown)


packages/foreman/rubygem-hammer_cli_foreman/rubygem-hammer_cli_foreman.spec
```diff
@@ -22,56 +19,40 @@ URL: https://github.com/theforeman/hammer-cli-foreman
...
-BuildRequires: %{?scl_prefix_ruby}ruby(release)
-BuildRequires: %{?scl_prefix_ruby}ruby
-BuildRequires: %{?scl_prefix_ruby}rubygems-devel
+Requires: ruby
+Requires: ruby(rubygems)
+Requires: rubygem(hammer_cli) >= 3.7.0
+Requires: rubygem(apipie-bindings) >= 0.6.0
+Requires: rubygem(rest-client) >= 1.8.0
+Requires: rubygem(rest-client) < 3.0.0
+Requires: rubygem(jwt) >= 2.2.1
```
ekohl, line 28: Note to self: perhaps these are autogenerated now and can be skipped. If so, the template should be updated.

packages/foreman/rubygem-hammer_cli_foreman/rubygem-hammer_cli_foreman.spec
```diff
@@ -12,7 +9,7 @@
 %{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
 %global hammer_confdir %{_root_sysconfdir}/hammer
```
ekohl, line 10:
```suggestion
%global hammer_confdir %{_sysconfdir}/hammer
```

ekohl (commented):
Normally the SCL macros are unset on the distro. However, on Fedora I've seen some things start to fail somehow. I suspect packit sees something similar. Since we don't care about SCL anymore, it's better to remove it because it also cleans up the spec file.

ofedoren:
Thanks, updated. I also changed the changelog entry, because build failed with:
```
fatal: [rubygem-hammer_cli_foreman]: FAILED! => changed=false 

[2023-05-26T16:39:21.436Z]   changelog_epoch_version_release: 3.8.0-0.2.pre.develop

[2023-05-26T16:39:21.436Z]   msg: changelog entry missing for 3.8.0-0.1.pre.develop

[2023-05-26T16:39:21.436Z]   specfile_epoch_version_release: 3.8.0-0.1.pre.develop
```

Not sure if that's right, so just a heads up.

ehelms (approved):

PR #9485, opened 2023-06-14
Release ansible-collection-theforeman-foreman 3.11.0

evgeni:

### DIFF
--- a/plugins/ansible-collection-theforeman-foreman/debian/rules
+++ b/plugins/ansible-collection-theforeman-foreman/debian/rules
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 
 %:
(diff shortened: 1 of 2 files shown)


plugins/ansible-collection-theforeman-foreman/debian/rules
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 
 %:
-	dh $@
+	dh $@ --buildsystem=none
```
evgeni, line 4: we now ship the Makefile in the collection, which makes `dh` try to use the `Makefile` buildsystem, and that breaks all the things.

ehelms (approved):

PR #9506, opened 2023-06-22
Bump rubygem-foreman_rh_cloud to 8.0.47-1

ShimShtein:


pcreech:
@ShimShtein Looks like this can use a rebase, 8.0.46 was merged.

evgeni:
I've just merged https://github.com/theforeman/foreman-packaging/pull/9514 which is the same diff, but rebased.

PR #9519, opened 2023-06-27
Regenerate scc_manager spec based on latest template

evgeni:

### DIFF
--- a/packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
+++ b/packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
@@ -1,23 +1,16 @@
 # template: foreman_plugin
-%{?scl:%scl_package rubygem-%{gem_name}}
-%{!?scl:%global pkg_name %{name}}
-
 %global gem_name foreman_scc_manager
 %global plugin_name scc_manager
-%global foreman_min_version 3.4
-%global katello_min_version 3.16.0
+%global foreman_min_version 2.1
 
-Name: %{?scl_prefix}rubygem-%{gem_name}
+Name: rubygem-%{gem_name}
 Version: 2.3.0
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Summary: Suse Customer Center plugin for Foreman
-Group: Applications/Systems
 License: GPLv3
 URL: https://www.orcharhino.com/
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
-Requires: %{?scl_prefix}rubygem(katello) >= %{katello_min_version}
-BuildRequires: %{?scl_prefix}rubygem(katello) >= %{katello_min_version}
 # start specfile generated dependencies
 Requires: foreman >= %{foreman_min_version}
 Requires: ruby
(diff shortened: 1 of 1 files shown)


packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
```diff
@@ -1,23 +1,16 @@
 # template: foreman_plugin
-%{?scl:%scl_package rubygem-%{gem_name}}
-%{!?scl:%global pkg_name %{name}}
-
 %global gem_name foreman_scc_manager
 %global plugin_name scc_manager
-%global foreman_min_version 3.4
-%global katello_min_version 3.16.0
```
evgeni, line 8: While this was removed here, there is still a `BuildRequires: rubygem(katello) >= 3.16` (and later an automatically generated `Requires: rubygem(katello) >= 3.16`) in the spec coming via https://github.com/ATIX-AG/foreman_scc_manager/blob/a84d26d6aeb11b03b52c290c515b64ef6794066b/foreman_scc_manager.gemspec#L26

(this was "recently" introduced in https://github.com/ATIX-AG/foreman_scc_manager/commit/414226edd690717ef590a166a055bb6e510f1552 and I guess the RPM spec predates that change by lot)

packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
```diff
@@ -1,23 +1,16 @@
...
+%global foreman_min_version 2.1
```
evgeni, line 4: This comes from https://github.com/ATIX-AG/foreman_scc_manager/blob/a84d26d6aeb11b03b52c290c515b64ef6794066b/lib/foreman_scc_manager/engine.rb#L25
The old RPM spec said 3.4, who is right? I don't know :)

ekohl: I think the Katello >= 3.16 version implies a much newer version than 2.1 anyway.

ekohl (approved):

packages/katello/rubygem-foreman_scc_manager/rubygem-foreman_scc_manager.spec
evgeni: Nope, that's exactly the one…

https://github.com/theforeman/forklift/blob/c60a2d17d9dbdb3a33c7190a4beb0f1ab06ed6e4/vagrant/config/versions.yaml#L22-L31

nadjaheitmann: 3.4 should be correct, also taking into account the configs in package.json file. We should update the gemspec, eventually.

evgeni: Then please update `engine.rb` (and the gemspec) and on next run it will regenerate that line correctly. :)

evgeni: (given 3.4 is old enough, I see no problem in merging as is, we won't be backporting that regeneration anyway, and if we decide to, not to older than 3.5 ;))

PR #9520, opened 2023-06-27
Regenerate RPM spec based on latest template

stejskalleos:
Regenerate `spec` file


stejskalleos:
For https://github.com/theforeman/foreman_leapp/pull/128

evgeni (approved):

PR #9532, opened 2023-06-29
Regenerate RPM spec based on latest template

nadjaheitmann:


packages/plugins/rubygem-foreman_snapshot_management/rubygem-foreman_snapshot_management.spec
```diff
@@ -94,8 +78,6 @@ cp -pa .%{gem_dir}/* \
 %exclude %{gem_cache}
 %{gem_spec}
 %{foreman_bundlerd_plugin}
-%{foreman_assets_plugin}
-%{foreman_assets_foreman}
```
evgeni, line 98: Those two lines need to remain, it's a bug in our template :/

https://github.com/theforeman/foreman-packaging/pull/9442

packages/plugins/rubygem-foreman_snapshot_management/rubygem-foreman_snapshot_management.spec
```diff
@@ -31,6 +27,7 @@ Provides: foreman-plugin-%{plugin_name} = %{version}
 BuildRequires: npm(@babel/core) >= 7.7.0
 BuildRequires: npm(@babel/core) < 8.0.0
 BuildRequires: npm(@theforeman/builder) >= 4.0.0
+BuildRequires: npm(@theforeman/builder) < 5.0.0
```
evgeni, line 30: This needs to be removed (preferably also in your `package.json`, so it won't come back)

nadjaheitmann: We have an open PR for this. We should merge it, eventually.

evgeni (approved):

nadjaheitmann:
Thanks @evgeni

PR #9533, opened 2023-06-29
Regenerate foreman_acd RPM spec based on latest template

nadjaheitmann:

### DIFF
--- a/packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
+++ b/packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
@@ -1,41 +1,33 @@
 # template: foreman_plugin
-%{?scl:%scl_package rubygem-%{gem_name}}
-%{!?scl:%global pkg_name %{name}}
-%{!?_root_localstatedir:%global _root_localstatedir %{_localstatedir}}
-
 %global gem_name foreman_acd
 %global plugin_name acd
 %global foreman_min_version 2.1
 
-Name: %{?scl_prefix}rubygem-%{gem_name}
+Name: rubygem-%{gem_name}
 Version: 0.9.4
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Summary: Foreman plugin to provide application centric deployment and self service portal
-Group: Applications/Systems
 License: GPLv3
 URL: https://www.orcharhino.com
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
-Requires: git
-BuildRequires: git
-
 # start specfile generated dependencies
 Requires: foreman >= %{foreman_min_version}
-Requires: ruby
+Requires: ruby >= 2.5
 BuildRequires: foreman-assets >= %{foreman_min_version}
 BuildRequires: foreman-plugin >= %{foreman_min_version}
-BuildRequires: ruby
+Requires: ruby >= 2.5
+BuildRequires: ruby >= 2.5
 BuildRequires: rubygems-devel
 BuildArch: noarch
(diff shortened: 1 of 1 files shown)


nadjaheitmann:
cc @sbernhard

packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
```diff
@@ -1,35 +1,28 @@
 # template: foreman_plugin
-%{?scl:%scl_package rubygem-%{gem_name}}
-%{!?scl:%global pkg_name %{name}}
-%{!?_root_localstatedir:%global _root_localstatedir %{_localstatedir}}
```
evgeni, line 4: this variable is used a few times in the spec, but now that you're not doing any funny SCL things, you can just replace it with the "original" `_localstatedir`

packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
```diff
@@ -1,35 +1,28 @@
...
-BuildRequires: ruby
+Requires: ruby >= 2.5
+BuildRequires: ruby >= 2.5
 BuildRequires: rubygems-devel
 BuildArch: noarch
 Provides: foreman-plugin-%{plugin_name} = %{version}
-BuildRequires: rubygem(foreman_remote_execution) >= 7.0
-BuildRequires: rubygem(foreman-tasks) >= 7.0
+BuildRequires: rubygem(foreman_remote_execution) >= 3.3.0
+BuildRequires: rubygem(foreman-tasks) >= 0.10
```
evgeni, line 25: I guess some bumping inside the gemspec is in order? :)

nadjaheitmann: :see_no_evil:

packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
nadjaheitmann: @sbernhard Can you have a look, please :)

evgeni: just do a `sed -i 's/_root_localstatedir/_localstatedir/g' packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec` ;)

evgeni:
there is a `BuildRequires: npm(@theforeman/builder) < 11.0.0` in the spec that prevents it from building right now.

nadjaheitmann:
> there is a `BuildRequires: npm(@theforeman/builder) < 11.0.0` in the spec that prevents it from building right now

:tired_face: Let me remove this...

packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
nadjaheitmann: :woman_shrugging:

packages/plugins/rubygem-foreman_acd/rubygem-foreman_acd.spec
nadjaheitmann: Funny thing is that is says

>   s.add_dependency 'foreman_remote_execution', '>= 8.0'
     s.add_dependency 'foreman-tasks', '>= 7.0'

in https://github.com/ATIX-AG/foreman_acd/blob/master/foreman_acd.gemspec

evgeni: But not in https://github.com/ATIX-AG/foreman_acd/blob/v0.9.4/foreman_acd.gemspec, which is what we package here.

evgeni (approved):

PR #9535, opened 2023-06-30
Refs #36559 - Implement sysusers.d

ekohl:
Depends on https://github.com/theforeman/foreman/pull/9766 and I also raised some further questions about this in https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/EXUGSSEMY2MVVGOFZIK4NBMQB5PWWWNG/ because I'm uncertain this is correct. Another annoying thing is that the `%sysusers_create_compat` macro is not in EL8. `epel-rpm-macros` provides it but that would pull in EPEL.

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -1,5 +1,4 @@
 %global homedir %{_datadir}/%{name}
-%global confdir extras/packaging/rpm/sources
 %global foreman_rake %{_sbindir}/%{name}-rake
 %global dynflow_sidekiq_service_name dynflow-sidekiq@
 %global rake /usr/bin/rake
@@ -35,10 +34,11 @@ Requires: /etc/cron.d
 Requires: gawk
 Requires: /usr/sbin/sendmail
 
-Requires(pre):  shadow-utils
-Requires(post): systemd-sysv
-Requires(post): systemd-units
-Requires(preun): systemd-units
+BuildRequires:  systemd-rpm-macros
+%systemd_requires
+%if 0%{rhel} >= 9
+%{?sysusers_requires_compat}
+%end
 
 # Subpackages
 Requires: %{name}-debug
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman/foreman.spec
```diff
@@ -1,5 +1,4 @@
 %global homedir %{_datadir}/%{name}
-%global confdir extras/packaging/rpm/sources
```
ekohl, line 2: This looks unused

packages/foreman/foreman/foreman.spec
```diff
@@ -35,10 +34,11 @@ Requires: /etc/cron.d
...
 
-Requires(pre):  shadow-utils
-Requires(post): systemd-sysv
-Requires(post): systemd-units
-Requires(preun): systemd-units
+BuildRequires:  systemd-rpm-macros
+%systemd_requires
+%if 0%{rhel} >= 9
+%{?sysusers_requires_compat}
+%end
```
ekohl, line 41: Is the conditional around it redundant given it uses `%?` already?

packages/foreman/foreman/foreman.spec
```diff
@@ -939,12 +940,14 @@ rm -rf %{buildroot}
...
 %{_tmpfilesdir}/%{name}.conf
+%{_sysusersdir}/%{name}.conf
 
 %pre
-# Add the "foreman" user and group
-getent group %{name} >/dev/null || groupadd -r %{name}
-getent passwd %{name} >/dev/null || \
-useradd -r -g %{name} -d %{homedir} -s /sbin/nologin -c "Foreman" %{name}
+%if 0%{rhel} < 9
+%sysusers_create_compat extras/systemd/%{name}.sysusers
```
ekohl, line 947: The docs do not mention if this is valid. They only mention `%{SOURCE3}`.

packages/foreman/foreman/foreman.spec
```diff
@@ -939,12 +940,14 @@ rm -rf %{buildroot}
...
+%else
+%sysusers_create_package %{name} extras/systemd/%{name}.sysusers
```
ekohl, line 949: `%sysusers_create_compat` isn't in EL8 and we don't want to depend on `epel-rpm-macros` (I think). Same here that I'm not sure if the last path is valid given it isn't documented.

PR #9591, opened 2023-07-21
Updating change-hostname to suggest using global registration

Griffin-Sullivan:

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -140,18 +140,15 @@ module KatelloUtilities
       else
         # the following multi-line string isn't indented because the indents are taken literally.
 %(
-  You will have to install the new bootstrap rpm and reregister all clients and #{@plural_proxy} with subscription-manager
-  (update organization and environment arguments appropriately):
+  You will have to reregister all clients and #{@plural_proxy}.
+  The global registration method requires an activation key and root privileges on the client server.
 
-  yum remove -y katello-ca-consumer*
-  rpm -Uvh http://#{@new_hostname}/pub/katello-ca-consumer-latest.noarch.rpm
-  subscription-manager register --org="Default_Organization" --environment="Library" --force
+  For each client and #{@plural_proxy}, go to Hosts > Register Host in the UI.
+  Select the correct activation key you want to use and any other parameters you wish to include.
+  Once you are done selecting your parameters, click Generate and copy the curl command.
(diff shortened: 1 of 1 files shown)


packages/katello/katello/hostname-change.rb
```diff
@@ -140,18 +140,15 @@ def next_steps_message
       else
         # the following multi-line string isn't indented because the indents are taken literally.
 %(
-  You will have to install the new bootstrap rpm and reregister all clients and #{@plural_proxy} with subscription-manager
-  (update organization and environment arguments appropriately):
+  You will have to reregister all clients and #{@plural_proxy} with the Satellite server.
```
ekohl, line 143: We don't use Satellite in upstream, but I'm not sure if we even handle branding at all.

packages/katello/katello/hostname-change.rb
```diff
@@ -140,18 +140,15 @@ def next_steps_message
...
 
-  yum remove -y katello-ca-consumer*
-  rpm -Uvh http://#{@new_hostname}/pub/katello-ca-consumer-latest.noarch.rpm
-  subscription-manager register --org="Default_Organization" --environment="Library" --force
-
-  Then reattach subscriptions to the client(s) and run:
-
-  subscription-manager refresh
-  yum repolist
+  For each client and #{@plural_proxy}, go to Hosts > Register Host in the Satellite UI.
```
ekohl, line 146: I think this is a poor experience. Users can't script this and if you have 100s or 1000s of hosts it's not going to be acceptable.

IMHO we should have a remote execution job that updates the registration instead. Then we can tell users it can be called via Hammer as well.

ekohl (changes requested):
@ehelms this is the second PR I see around this topic. Is there any RFC that expands on the `katello-ca-consumer` deprecation and its replacements? It feels very ad hoc and not well thought out.

packages/katello/katello/hostname-change.rb
Griffin-Sullivan: Haha good point I should avoid saying Satellite in this.

packages/katello/katello/hostname-change.rb
Griffin-Sullivan: I'm not the expert here, so I would like to see if there's already other ways to achieve what you're talking about. If not, then we should spend time creating a solution like you described.

Griffin-Sullivan:
@ares I was told your opinion would be helpful here as well

PR #9596, opened 2023-07-25
Script to generate a stage repository based on a Copr repository

ehelms:
This is an early draft of a script that will reposync a Copr repository to a folder, copy it into the right structure and add modularity to be used for stage testing. Related to proposal https://github.com/theforeman/foreman-infra/issues/1795#issuecomment-1638048498

With this design I am considering the following (if anyone has input):

 1. This stand-alone script that lives here, or to build this into Obal given some of the knowledge is contained in the `package_manifest,yml`
 2. Whether I can re-use the `yum.conf` in `repoclosure` directory instead of a dedicated reposync configuration. The `yum.conf`has issues with the cache options that are specified and permissions on the filesystem with reposync.

### DIFF
--- a/build_stage_repository.py
+++ b/build_stage_repository.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+
+from subprocess import check_output, STDOUT, CalledProcessError
+import os
+import yaml
+import glob
+import shutil
+import time
+import hashlib
+import sys
+
+
+def move_rpms_from_copr_to_stage(collection, version, src_folder, dest_folder, source=False):
+    if source:
+        print(f"Moving {collection} Source RPMs from Copr directory to stage repository")
+    else:
+        print(f"Moving {collection} RPMs from Copr directory to stage repository")
+
+    if not os.path.exists(dest_folder):
+        os.mkdir(dest_folder)
+
+    repo_folder = f"{src_folder}/el8-{collection}-{version}"
+
+    if source:
+        files = glob.glob(repo_folder + f"/**/*.src.rpm")
+    else:
+        files = glob.glob(repo_folder + f"/**/*.rpm")
+
+    for file in files:
+        file_name = os.path.basename(file)
+        shutil.move(file, os.path.join(dest_folder, file_name))
+
+    shutil.rmtree(src_folder)
+
+
+def modulemd_yaml(collection):
+    return f"modulemd/modulemd-{collection}-el8.yaml"
+
+
+def generate_modulemd_version(version):
+    if version == 'nightly':
+        modulemd_version_prefix = '9999'
+    else:
+        major, minor = version.split('.')
+        modulemd_version_prefix = int(major)*100 + int(minor)
+
+    modulemd_version_string = time.strftime(f"{modulemd_version_prefix}%Y%m%d%H%M%S", time.gmtime())
+
+    return int(modulemd_version_string)
+
+
+def generate_modulemd_context(collection, version):
+    context_string = f"{collection}-{version}"
+    digest = hashlib.sha256(context_string.encode()).hexdigest()
+    return digest[:8]
+
+
+def create_modulemd(collection, version, stage_dir):
+    print("Adding modulemd to stage repository")
+    cmd = [
+        'rpm',
+        '--nosignature',
+        '--query',
+        '--package',
+        f"{stage_dir}/*.rpm",
+        "--queryformat=%{name}-%{epochnum}:%{version}-%{release}.%{arch}\n"
+    ]
+    output = check_output(cmd)
+
+    with open(modulemd_yaml(collection), 'r') as file:
+        modules = yaml.safe_load(file)
+
+    modules['data']['artifacts'] = {'rpms': output.splitlines()}
+    modules['data']['version'] = generate_modulemd_version(version)
+    modules['data']['context'] = generate_modulemd_context(collection, version)
+    modules_yaml = os.path.join(stage_dir, 'repodata', 'modules.yaml')
+
+    with open(modules_yaml, 'w') as modules_file:
+        yaml.dump(modules, modules_file, default_flow_style=False, explicit_start=True, explicit_end=True)
+
+    check_output(['modifyrepo_c', '--mdtype=modules', modules_yaml, f"{stage_dir}/repodata"])
+
+
+def create_repository(repo_dir):
+    check_output(['createrepo', repo_dir])
+
+
+def sync_copr_repository(collection, version, target_dir, source=False):
+    if source:
+        print(f"Syncing {collection} {version} Source RPM repository from Copr")
+    else:
+        print(f"Syncing {collection} {version} RPM repository from Copr")
+
+    cmd = [
+        'dnf',
+        'reposync',
+        '--newest-only',
+        '--repo',
+        f"el8-{collection}-{version}",
+        '--config',
+        'reposync_config.conf',
+        '--download-path',
+        target_dir
+    ]
+
+    if source:
+        cmd.extend([
+            '--source',
+        ])
+    else:
+        cmd.extend([
+            '--exclude',
+            '*.src',
+        ])
+
+    check_output(cmd)
+
+
+def main():
+    try:
+        collection = sys.argv[1]
+        version = sys.argv[2]
+        operating_system = sys.argv[3]
+    except IndexError:
+        raise SystemExit(f"Usage: {sys.argv[0]} collection version os")
+
+    base_dir = 'tmp'
+    rpm_sync_dir = f"{base_dir}/rpms"
+    srpm_sync_dir = f"{base_dir}/srpms"
+
+    stage_dir = f"{base_dir}/{collection}/{version}/{operating_system}/"
+    rpm_dir = f"{stage_dir}/x86_64"
+    srpm_dir = f"{stage_dir}/source"
+
+    if not os.path.exists(rpm_sync_dir):
+        os.makedirs(rpm_sync_dir)
+
+    if not os.path.exists(srpm_sync_dir):
+        os.makedirs(srpm_sync_dir)
+
+    if not os.path.exists(rpm_dir):
+        os.makedirs(rpm_dir)
+
+    if not os.path.exists(srpm_dir):
+        os.makedirs(srpm_dir)
+
+    sync_copr_repository(collection, version, rpm_sync_dir)
+    sync_copr_repository(collection, version, srpm_sync_dir, source=True)
+
+    move_rpms_from_copr_to_stage(collection, version, srpm_sync_dir, srpm_dir, source=True)
+    move_rpms_from_copr_to_stage(collection, version, rpm_sync_dir, rpm_dir)
+
+    create_repository(rpm_dir)
+    create_repository(srpm_dir)
+
+    if collection in ['foreman', 'katello'] and operating_system == 'el8':
+        create_modulemd(collection, version, rpm_dir)
+
+
+if __name__ == '__main__':
+    main()
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ downloaded_rpms
 
 SRPMs/
 mock_builds/
+tmp/
+copr_build_info/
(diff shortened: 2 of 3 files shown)


build_stage_repository.py
```diff
@@ -0,0 +1,121 @@
...
+    src_folder = "{}/el8-{}-{}".format(src_folder, collection, version)
+
+    files = glob.glob(src_folder + '/**/*.rpm')
+
+    for file in files:
+        file_name = os.path.basename(file)
+        shutil.move(file, os.path.join(dest_folder, file_name))
+
+def modulemd_yaml(collection):
+    return "modulemd/modulemd-{}-el8.yaml".format(collection)
```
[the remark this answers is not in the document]
ehelms: Is there a preferred method? I can't keep up :smile:

ehelms:
@ekohl Any thoughts on the design question? stand-alone script here or Obal functionality that drives off package_manifest?

build_stage_repository.py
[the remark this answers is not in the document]
ekohl: If you care about Python 2 you can't use f-strings. Otherwise I'd say it's (almost) always preferred since it's faster and shorter.

ehelms:
In terms of workflow, we have two options.

Option 1:

Clone packaging (Jenkins node) -> Reposync from Copr (Jenkins node) -> Filter repository and add modularity (Jenkins node) -> SCP repository to `stagingyum.theforeman.org` (Jenkins node to web02)


Option 2:

SSH to web02 (Jenkins node) -> Clone/Update packaging repo (on web02) -> Reposync from Copr (on web02) -> Filter and add modularity (on web02)


With each workflow the considerations should be around storage requirements, computational output and parallelization. The storage requirements:

 1. foreman-packaging repository (~700-800 MB)
 2. Copy of Copr repository (200 - 300 MB per)
 3. Filtering + modularity (~10MB?) this operation is a move and not a copy

We will also need to have a workflow that can support independent release streams, which means independent copies of foreman-packaging. Cloning of `foreman-packaging` can add `0.5 to 2 minutes` to any Jenkins pipeline depending on network speeds due to it's size. Having the tooling to generate the repository live in the packaging repository inherently forces us to need the packaging repository.

evgeni:
* If we do a shallow clone (`git clone --branch=rpm/develop --depth=1 https://github.com/theforeman/foreman-packaging`), it only will consume ~140MB and is a tad faster (something like 5 seconds on my laptop, but I don't think it's representative).
* I would prefer to keep `web02` as stupid as possible, so I am in favor not to do any actions (besides storing) there.
* I thought copr can do comps for us, just not modularity (the way we want it)?

My crazy idea (if we can have copr do filtering, and only need to inject modular metadata): we don't have to host the rpms at all our-self. Just the metadata, and then point to copr for the actual content, that would save us from having *another* copy of the rpms lying around for the moment of staging.

ekohl:
> My crazy idea (if we can have copr do filtering, and only need to inject modular metadata): we don't have to host the rpms at all our-self. Just the metadata, and then point to copr for the actual content, that would save us from having _another_ copy of the rpms lying around for the moment of staging.

AFAIK COPR deletes old versions. When Fedora X goes EOL the buildroots are dropped after some time. I like that Foreman has a complete archive of all releases.

If we're throwing crazy ideas out there: should we use Pulp to host the repositories?

evgeni:
> > My crazy idea (if we can have copr do filtering, and only need to inject modular metadata): we don't have to host the rpms at all our-self. Just the metadata, and then point to copr for the actual content, that would save us from having _another_ copy of the rpms lying around for the moment of staging.
> 
> AFAIK COPR deletes old versions. When Fedora X goes EOL the buildroots are dropped after some time. I like that Foreman has a complete archive of all releases.

Sure. We'd still have `yum.theforeman.org` for the real thing.

I was thinking not to host another copy for the staging part (which today resides on koji).

> If we're throwing crazy ideas out there: should we use Pulp to host the repositories?

I for one like static shit ;-)

ekohl:
> I was thinking not to host another copy for the staging part (which today resides on koji).

In https://github.com/theforeman/foreman-infra/issues/1795#issuecomment-1638048498 / https://github.com/fedora-copr/copr/issues/2782 it was explained that we can't solely rely on copr for that. It's a massive downside because it also means we can't easily fork our repos for scratch builds and have working repoclosure (which we don't have today). Also something we could use in our Packit builds if that was supported. However, I wouldn't count that as a regression.

ehelms:
> In [theforeman/foreman-infra#1795 (comment)](https://github.com/theforeman/foreman-infra/issues/1795#issuecomment-1638048498) / [fedora-copr/copr#2782](https://github.com/fedora-copr/copr/issues/2782) it was explained that we can't solely rely on copr for that. It's a massive downside because it also means we can't easily fork our repos for scratch builds and have working repoclosure (which we don't have today). Also something we could use in our Packit builds if that was supported. However, I wouldn't count that as a regression.

From my understanding, and I think it could work (I need to try it out more), is that we can download the metadata only to Jenkins, update the metadata such that the metadata sees the RPMs we need as having effective remote storage in Copr and add modularity. Then copy the metadata to `web02` and proceed with testing. Then we are only storing minimal extra on `web02` and relying on Copr for our staging storage. This reduces bandwidth time to download and storage cost.

We would eventually still need to download the RPMs once we are ready to promote to production but in theory this means we get to optimize for the stage generation and testing (which happens more than push to production).

ehelms:
After extensive testing, the idea of just downloading and updating the metadata to reduce storage and testing time costs failed due to how RPM metadata works. Given that, here is my proposed workflow:

 1. Clone packaging to Jenkins
 2. Download copy of Copr repository to Jenkins
 3. Update repository with modularity
 4. SCP repository to web01 hosted at stagingyum.theforeman.org
 5. Run installation tests
 6. Copy / merge staging repository to production one within web01
 7. Createrepo

This would allow parallelization and keep web01 "dumb". If we let web01 be "smarter" we could do the download on web01 allowing us to optimize such that it only downloads packages not already downloaded and deletes any not specified by the repository metadata. This would require the script (and thus foreman-packaging) to be cloned on web01 and to support parallelization we would only reference the script from the `rpm/develop` branch.

build_stage_repository.py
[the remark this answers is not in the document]
ehelms: As I need this to run on our Jenkins nodes, which are EL7 right now I'll keep this implementation.

ehelms:
This is ready -- tested and used to generate

https://stagingyum.theforeman.org/

build_stage_repository.py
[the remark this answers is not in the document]
ekohl: Note the shebang is `/usr/bin/env python3`, so it is using Python 3, even on EL7.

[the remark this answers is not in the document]
ehelms: Touche, forgot I did that :)

ehelms:
Updated to use f-strings

ehelms:
I am ready for some final review, this script is a key part of the pipelines and I can update (https://github.com/theforeman/jenkins-jobs/pull/351/files#diff-9e291d4cc6508af1581029e464793ffa8059e64e762f338539815a5be6956fd7R73) to not point at my fork.

build_stage_repository.py
```diff
@@ -0,0 +1,182 @@
...
+import time
+import hashlib
+import sys
+
+
+def execute(command):
+    try:
+        return check_output(command, universal_newlines=True, stderr=STDOUT)
+    except CalledProcessError as error:
+        print(error.output)
```
ekohl, line 17: Should this be printed to stderr?

build_stage_repository.py
```diff
@@ -0,0 +1,182 @@
...
+        '--package',
+        f"{stage_dir}/*rpm",
+        "--queryformat=%{name}-%{epochnum}:%{version}-%{release}.%{arch}\n"
+    ]
+    output = execute(cmd)
+
+    with open(modulemd_yaml(collection), 'r') as file:
+        modules = yaml.safe_load(file)
+
+    modules['data']['artifacts'] = {'rpms': output.splitlines()}
```
ekohl, line 80: I did some type checking and `output` can be `False` and `False.splitlines()` will fail.

build_stage_repository.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def create_modulemd(collection, version, stage_dir):
+    print("Adding modulemd to stage repository")
+    cmd = [
+        'rpm',
+        '--nosignature',
+        '--query',
+        '--package',
+        f"{stage_dir}/*rpm",
```
ekohl, line 72:
```suggestion
        f"{stage_dir}/*.rpm",
```

build_stage_repository.py
```diff
@@ -0,0 +1,182 @@
...
+import hashlib
+import sys
+
+
+def execute(command):
+    try:
+        return check_output(command, universal_newlines=True, stderr=STDOUT)
+    except CalledProcessError as error:
+        print(error.output)
+        return False
```
ekohl, line 18: For type checking, would it be better if the exception was actually just raised? Right now I don't know if you want to differentiate between success but no output and False.

build_stage_repository.py
```diff
@@ -0,0 +1,182 @@
...
+            '--download_path',
+            target_dir
+        ])
+
+        if source:
+            cmd.extend([
+                '--source',
+            ])
+
+    execute(cmd)
```
ekohl, line 137: Shouldn't it halt if this fails? Now it just prints an error on stdout and continues.

build_stage_repository.py
```diff
@@ -0,0 +1,182 @@
...
+    cmd = [
+        'reposync',
+        '--newest-only',
+        '--repo',
+        f"el8-{collection}-{version}",
+        '--config',
+        'reposync_config.conf'
+    ]
+
+    if execute(['reposync', '--version']):
```
ekohl, line 110: Here is a vague one. If this fails or doesn't return any output it uses one code path and otherwise the other? Is that for EL7 vs EL8? Perhaps that deserves some code comment.

ekohl (commented):
FYI: in f373bcbad8aa50f728ef60d2bb02fb10bb11539a I took a stab at rewriting it to use pathlib and add data types.

build_stage_repository.py
ehelms: It could be? I just want to see the error when there is one.

build_stage_repository.py
ehelms: Correct -- it's for EL7 vs. EL8. I can add a comment.

ehelms:
Re-factored to clean it up based on suggestions and add more clarity around reposync version. I did not tackle the types or use of Pathlib. I don't see the value in it at the moment for this script, and that may be partly my more classical Python knowledge.

build_stage_repository.py
```diff
@@ -0,0 +1,183 @@
...
+    cmd = [
+        'reposync',
+        '--newest-only',
+        '--repo',
+        f"el8-{collection}-{version}",
+        '--config',
+        'reposync_config.conf'
+    ]
+
+    if is_el7_reposync():
```
ekohl, line 112: Just a thought: we have `dnf` installed even on our EL7 nodes so if you call `dnf reposync` instead of `reposync`, do you get the single modern variant?

ehelms: Oooo nice, I've tested and updated based on this!

build_stage_repository.py
```diff
@@ -0,0 +1,170 @@
...
+        yaml.dump(modules, modules_file, default_flow_style=False, explicit_start=True, explicit_end=True)
+
+    check_output(['modifyrepo_c', '--mdtype=modules', modules_yaml, f"{stage_dir}/repodata"])
+
+
+def create_repository(repo_dir):
+    check_output(['createrepo', repo_dir])
+
+
+def is_el7_reposync():
```
ekohl, line 88: This is now unused

ehelms: Bah, thanks, I thought I dropped it.

ehelms:
One downside of this script living here is that it's harder to use also in pulpcore-packaging.

ehelms:
Moved to https://github.com/theforeman/theforeman-rel-eng/pull/261

PR #9600, opened 2023-07-26
Regenerate rubygem-smart_proxy_remote_execution_ssh spec 

adamruzicka:


adamruzicka:
What was wrong here?
[test rpm]

dgoetz:
It is not bumping the version as it is exactly the same version and release as already published. This is also the latest version, so if you want to change anything you need to bump at least the release to 2. 
As you are not only cleaning up SCL things, but also removing some files from the packages, I think this would be the way to go, but I am not 100% sure if this is really what you want.

adamruzicka:
Oh yeah, it's been a while since I opened it and totally forgot what I was doing here. Thank you

evgeni (approved):

PR #9612, opened 2023-08-01
Bump rubygem-unicode-display_width to 2.4.2-1

ehelms:


ehelms:
This never auto-updated due to a conflict:

```
Found conflict for rubygem-unicode-display_width: 1.8.0 (smart-proxy) != 2.4.2 (hammer-cli)
```

However, that conflict arises not because of production dependencies. The rubocop in smart-proxy is so old that in testing environments it requires an older version of unicode-display_width (could be fixed by https://github.com/theforeman/smart-proxy/pull/860). And thus when the Gemfile.lock is analyzed it sees a conflict.

evgeni (approved):

PR #9617, opened 2023-08-02
Include nodejs packages from Foreman in list of updateable packages

ehelms:
Adding Katello will depend upon https://github.com/theforeman/jenkins-jobs/pull/347

PR #9618, opened 2023-08-02
Handle updating nodejs packages

ehelms:
Requires https://github.com/theforeman/foreman-packaging/pull/9617


.github/workflows/bump_packages.yml
```diff
@@ -43,16 +43,28 @@ jobs:
         sudo curl --create-dirs -o /usr/local/bin/spectool https://pagure.io/rpmdevtools/raw/26a8abc746fba9c0b32eb899b96c92841a37855a/f/spectool.in
         echo 'echo "$GIT_AUTHOR_NAME <$GIT_AUTHOR_EMAIL>"' | sudo tee /usr/local/bin/rpmdev-packager
         sudo chmod +x /usr/local/bin/spectool /usr/local/bin/rpmdev-packager
+    - name: Install dependencies for nodejs
+      if: ${{ matrix.npm_name != null }}
+      run: |
+        sudo curl -fsSL https://deb.nodesource.com/setup_16.x | sudo -E bash - &&\
+        sudo apt-get install -y nodejs
+        sudo npm install npm2rpm
```
ekohl, line 51: Wouldn't it be better to use https://github.com/actions/setup-node
```suggestion
    - name: Install nodejs
      if: ${{ matrix.npm_name != null }}
      uses: actions/setup-node@v3
      with:
        node-version: 16
    - name: Install npm2rpm
      if: ${{ matrix.npm_name != null }}
      run: npm install npm2rpm
```

.github/workflows/bump_packages.yml
```diff
@@ -43,16 +43,28 @@ jobs:
...
     - name: Checkout RPM
       uses: actions/checkout@v3
       with:
         ref: rpm/develop
     - name: Initialize git annex
       run: git annex init
     - name: Bump ${{ matrix.package_name }} to ${{ matrix.new_version }}
+      if: ${{ matrix.gem_name != null }}
       run: ./bump_rpm.sh "${{ matrix.directory }}" "${{ matrix.new_version }}"
```
ekohl, line 60: Would it make sense to enhance `bump_rpm.sh` to recognize NPM packages? We can enhance the spec files with a comment so it can derive the strategy from the actual spec file, just like we do with gems and their template.

ehelms:
This now uses bump_rpm for Nodejs so the only thing needing added is the setup of NodeJS.

ekohl (approved):

PR #9620, opened 2023-08-02
Drop SCL from nodejs-node-gyp spec

ehelms:


ehelms:
[test rpm-copr]

ehelms:
> This looks sane to me? Not that I know much about it anyway.
> 
> The only thing that looked "odd" is that we now have `el8` in the name of some files, but they are applied unconditionally (so could loose the `el8` suffix). But that's not really blocking functionality.

I was hedging my bets that EL9 would end up with it's own specialness, but I can rename and hope for the best.

ehelms:
[test rpm-copr]

ehelms:
[test rpm-copr]

ehelms:
[test rpm-copr]

PR #9621, opened 2023-08-02
Handle bumping nodejs RPMs

ehelms:


bump_rpm.sh
```diff
@@ -33,13 +33,12 @@ ensure_program rpmspec rpm-build
...
 
-cd $1
-
 ROOT=$(git rev-parse --show-toplevel)
 PACKAGE_NAME=$(basename $1)
 SPEC_FILE=*.spec
-GEM_NAME=$(awk '/^%global\s+gem_name/ { print $3 }' $SPEC_FILE)
-CURRENT_VERSION=$(rpmspec --srpm -q --queryformat="%{version}" $SPEC_FILE)
+GEM_NAME=$(awk '/^%global\s+gem_name/ { print $3 }' $1/$SPEC_FILE)
+NPM_NAME=$(awk '/^%global\s+npm_name/ { print $3 }' $1/$SPEC_FILE)
```
ekohl, line 40: Should these live below the `if [[ $PACKAGE_NAME == *rubygem-* ]]` and nodejs equivalent bits? It feels like they're no longer top level items.

ekohl (approved):
I think this is correct. Small detail inline, but not too important.

PR #9639, opened 2023-08-10
Foreman Puppet - require theforeman/builder >= 12.0.1

stejskalleos:


stejskalleos:
Closing the issue, too late I realized I have to change it first in [package.json](https://github.com/theforeman/foreman_puppet/blob/master/package.json)

PR #9687, opened 2023-08-16
Update rubygem-foreman_remote_execution to 10.0.6

adamruzicka:


packages/plugins/rubygem-foreman_remote_execution/foreman_remote_execution-10.0.6.gem
```diff
@@ -0,0 +1 @@
+../../../.git/annex/objects/F6/wJ/SHA256E-s538112--e08198b8fa5de4cb61d5a47da4738009258b3cfafd0374c2f9eb319d64c43612.6.gem/SHA256E-s538112--e08198b8fa5de4cb61d5a47da4738009258b3cfafd0374c2f9eb319d64c43612.6.gem
```
evgeni, line 1: you didn't delete the 10.0.5 gem from git, this confuses annex.

adamruzicka: Yeah, sorry about that, fixed

evgeni (approved):

evgeni:
@ehelms copr failed with:
```
No matching package to install: 'rubygem(deface)'
No matching package to install: 'rubygem(foreman-tasks) >= 7.1.0'
Not all dependencies satisfied
Error: Some packages could not be found.
```

I think it's missing a repo dependency on "itself" (well, the "fork" repo doesn't contain f-tasks, while the "real" copr-plugins does)

ehelms:
Ohhh good find, https://github.com/theforeman/foreman-packaging/pull/9696 to address that

PR #9696, opened 2023-08-16
Add self-reference as an external repository in Copr

ehelms:
This is really only needed for the scratch repositories that get created, however, I don't see it harming the standard repositories. Therefore instead of adding additional variables and functionality this specifies the repository as it's own external repo.

I kinda made this more robust to changes in the process due to how much repeating myself was happening.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -42,46 +42,59 @@ copr_projects:
       - 'python39:3.9'
       - 'nodejs:12'
       - 'postgresql:12'
-    rhel_9_chroot: rhel-9-x86_64
-    rhel_8_chroot: rhel-8-x86_64
-    rhel_7_chroot: rhel-7-x86_64
+    rhel_9: '9'
+    rhel_8: '8'
+    rhel_7: '7'
+    root_repo_url: https://download.copr.fedorainfracloud.org/results/@theforeman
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -42,46 +42,59 @@ copr_projects:
       - 'python39:3.9'
       - 'nodejs:12'
       - 'postgresql:12'
-    rhel_9_chroot: rhel-9-x86_64
-    rhel_8_chroot: rhel-8-x86_64
-    rhel_7_chroot: rhel-7-x86_64
+    rhel_9: '9'
+    rhel_8: '8'
+    rhel_7: '7'
```
evgeni, line 47: what's the benefit of having those numbers in variables?

ehelms: Mostly to prevent copy/paste errors in the future when we add RHEL 9 for example, or other client architectures. This shows up in so many places it could be easy to overlook.

ehelms: However, I am open to not doing it this way.

evgeni: I find it weird to read, but on the other hand, it doesn't hurt, so :shrug:

evgeni (approved):

PR #9705, opened 2023-08-24
Fixes #18539 - Own the public/{assets,webpack} directories

ekohl:
By owning the directories we can also ensure they are properly cleaned up on upgrade.

This is now a draft to see if this works. If it does, we should also verify if this is a problem with other plugins and apply the same. Probably with macros.

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -6,7 +6,7 @@
 %global prereleasesource pre.master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global mainver 4.15.0
-%global release 1
+%global release 2
 
 Name: rubygem-%{gem_name}
 Version: %{mainver}
@@ -143,8 +143,11 @@ done
 %{gem_instdir}/engines
 %{gem_libdir}
 %{gem_instdir}/locale
+%dir %{gem_instdir}/public
+%dir %{gem_instdir}/public/assets
 %{gem_instdir}/public/assets/bastion
 %{gem_instdir}/public/assets/bastion_katello
+%dir %{gem_instdir}/public/webpack
 %exclude %{gem_instdir}/vendor
 %exclude %{gem_cache}
 %{gem_spec}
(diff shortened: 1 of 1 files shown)


ekohl:
Confirmed this works:

Before with 3.11 (because katello nightly doesn't work):
```console
# dnf install https://yum.theforeman.org/releases/3.11/el9/x86_64/foreman-release.rpm https://yum.theforeman.org/katello/4.13/katello/el9/x86_64/katello-repos-latest.rpm
# dnf install rubygem-katello
# dnf remove rubygem-katello
# ls /usr/share/gems/gems/katello*
public
```

After with nightly and the built RPM:
```console
# dnf install https://yum.theforeman.org/katello/nightly/katello/el9/x86_64/katello-repos-latest.rpm https://yum.theforeman.org/releases/nightly/el9/x86_64/foreman-release.rpm
# dnf install https://download.copr.fedorainfracloud.org/results/@theforeman/katello-nightly-staging-scratch-0fbf0e16-ef37-5063-9658-c3bc6540da26/rhel-9-x86_64/07953152-rubygem-katello/rubygem-katello-4.15.0-0.2.pre.master.el9.noarch.rpm
# dnf remove rubygem-katello
# ls /usr/share/gems/gems/katello*
ls: cannot access '/usr/share/gems/gems/katello*': No such file or directory
```

ekohl:
@evgeni before I copy all the plugins, mind taking a look? I don't like that this copies some logic from `%foreman_assets_plugin` and `%foreman_webpack_plugin` but other than introducing a new macro I don't know a better way.

ekohl:
> How about _extending_ those macros?

I also debated introducing a new macro like `foreman_plugin_files` that takes arguments like `foreman_precompile_plugin` with which files to include.

Now that I read it, perhaps `foreman_precompile_plugin` could write out a file that you can pass to `%files` to include.

evgeni:
It could, but it also sounds more complicated than extending the existing macros?

ekohl:
I took a stab at it in https://github.com/theforeman/foreman-packaging/pull/11190 but it really should be easier to build RPMs locally.

ekohl:
https://github.com/theforeman/foreman-packaging/pull/11233

PR #9773, opened 2023-09-06
Update to pulp_deb 3.0

quba42:
Requires: https://github.com/Katello/katello/pull/10719

Also requires: https://github.com/theforeman/pulpcore-packaging/pull/715

There are two options for how to proceed:

1. Wait for a new pulpcore version repository >3.28 to be added to pulpcore-packaging.
2. Add pulp_deb 3.0 to the 3.28 pulpcore repo, in this case, we need to:
   - Add this change to both rpm/develop, and rpm/3.8 (contains Katello 4.10 which already uses pulpcore 3.28)
   - Make sure all new RPMs arrive in the repositories at the same time, which must also coincide with a Katello 4.10.* release.


ianballou:
@Odilhao is it fine to update the katello spec and bump pulp-deb-client in the same PR here?

quba42:
> @Odilhao is it fine to update the katello spec and bump pulp-deb-client in the same PR here?

It is two different commits (which I guess might not be allowed/squished during merge), but it also means I can easily split it up into two PRs.

Odilhao:
We need to split this into 2 PRs, one to update pulp_deb_client and another one to bump the dependency on Katello, if we want to get both scratch's green.

quba42:
> We need to split this into 2 PRs, one to update pulp_deb_client and another one to bump the dependency on Katello, if we want to get both scratch's green.

I can do that.

Odilhao (approved):

PR #9790, opened 2023-09-08
Build Foreman Client for REX Pull Provider on SLES

maximiliankolb:
* On EL, the golang package is called "golang"; on openSUSE and SLES, it's called "go".
* Require redhat-rpm-config for go When building the Foreman Client for REX pull mode, you have to compile go code. On EL, you can set a flag that requires the file "redhat-hardened-ld" which is part of the "redhat-rpm-config" package. See https://fedora.pkgs.org/rawhide/fedora-x86_64/redhat-rpm-config-264-1.fc40.noarch.rpm.html
* Do not use hardened linker on SLES Hardened go linker requires "redhat-rpm-config" which is not available on openSUSE and SLES systems.
* DBus package name depends on distribution
* Remove ldflags for compiling go on SLES
* Unset GO111MODULE on SLES
* Change bin path for yggdrasil on SLES

### DIFF
--- a/packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
+++ b/packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
@@ -15,14 +15,18 @@
         go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -compressdwarf=false -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v %{?**};
 %else
 %if ! 0%{?gobuild:1}
+%if 0%{?suse_version}
+%define gobuild(o:) go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -a -v %{?**};
+%else
 %define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '" -a -v %{?**};
+%endif
 %endif
 %endif
 
 Name: foreman_ygg_worker
 Version: 0.2.2
 Summary: Worker service for yggdrasil that can act as pull client for Foreman Remote Execution
-Release: 1%{?dist}
+Release: 2%{?dist}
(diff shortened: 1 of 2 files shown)


maximiliankolb:
Hey @adamruzicka Can you please have a look at my PR?

packages/client/yggdrasil/yggdrasil.spec
```diff
@@ -68,9 +73,16 @@ make PREFIX=%{_prefix} \
 %{_datadir}/bash-completion/completions/*
 %{_mandir}/man1/*
 %{_prefix}/share/pkgconfig/%{name}.pc
+%if 0%{?suse_version}
+%{_bindir}/%{name}
+%else
 %{_libexecdir}/%{name}
+%endif
```
adamruzicka, line 80: where does this difference come from? Won't the systemd service still expect the binary to live in libexec?

adamruzicka (commented):
Maybe a heretical idea, but do we actually need to build it for suse on suse? The result is a static binary which should work anywhere as long as the architecture and os family matches. Even building on different EL versions feels like a waste in a way

packages/client/yggdrasil/yggdrasil.spec
maximiliankolb: I have reverted this change. It is due to different paths on OpenSUSE/SLES for `/usr/lib`.

maximiliankolb:
Rebased to "rpm/develop" and resolved all merged conflicts.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -34,7 +38,12 @@ Url: https://github.com/%{repo_orgname}/%{repo_name}/
 %endif
 ExclusiveArch: %{go_arches}
 
+%if 0%{?suse_version}
+BuildRequires: go
+%else
 BuildRequires: golang
+BuildRequires: redhat-rpm-config
```
ekohl, line 45: This is an unusual one. I think this is expected to be in the buildroot already. Can you explain why you added this?

maximiliankolb: I have added this to make it more explicit where the hardened linker on EL comes from. Note that this is not available on OpenSUSE/SLES.

From my commit message:

> Require redhat-rpm-config for go
>   When building the Foreman Client for REX pull mode, you have to compile
>   go code. On EL, you can set a flag that requires the file
>   "redhat-hardened-ld" which is part of the "redhat-rpm-config" package.
>   See https://fedora.pkgs.org/rawhide/fedora-x86_64/redhat-rpm-config-264-1.fc40.noarch.rpm.html


I am not sure what you mean by "buildroot". If it's always present when building RPMs for/on EL using foreman-packaging, then I can revert this.

ekohl:
https://github.com/theforeman/foreman-packaging/pull/11116 & https://github.com/theforeman/foreman-packaging/pull/11118 were merged so closing this.

PR #9791, opened 2023-09-08
Update list of supported versions in PR template

maximiliankolb:
Assuming the last two versions are supported, then merging after releasing 3.8 should work :v:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,12 +1,8 @@
 <!--
 If your package needs to be released within one or more release streams, and/or distributions, please open PRs to each of those branches respectively. The easiest way to do this is to make the initial commit for the mainline branch (e.g. rpm/develop or deb/develop) and then cherry pick the commit hash onto each subsequent branch.
 
-Supported Versions:
-
- * Nightly
- * 2.0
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,10 +3,9 @@ If your package needs to be released within one or more release streams, and/or
 
 Supported Versions:
 
- * Nightly
- * 2.0
- * 1.24
- * 1.23
+ * develop
+ * 3.8
+ * 3.7
```
ekohl, line 8: We also still support 3.6.
```suggestion
 * 3.7
 * 3.6
```

maximiliankolb: Even after the release of 3.8? As of right now, do you support 3.8 in RC-phase?

ekohl: > Even after the release of 3.8?

No

> As of right now, do you support 3.8 in RC-phase?

For end users: no, but it is a valid PR target.

Then again, given how old this list is and nobody had an issue with it, perhaps we should drop it and instead link to some place where there is a list of supported versions.

ekohl (approved):

PR #9861, opened 2023-10-06
Bump nodejs-webpack dependencies

ehelms:


ehelms:
Same version -- just bumping the release and nodejs modules included.

ehelms:
This currently fails cause I need to generate the tarball on a nodejs 12 machine.

PR #9863, opened 2023-10-06
Install nokogiri in package update job

ehelms:
Based on https://github.com/theforeman/foreman-packaging/actions/runs/6434684686/job/17474394173

```
FINISHED
Updating comps... - 
<internal:/usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb>:85:in `require': cannot load such file -- nokogiri (LoadError)
	from <internal:/usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from ./add_to_comps.rb:2:in `<main>'
Error: Process completed with exit code 1.
```

Thar probably be dragons with this attempt given nokogiri's past.

### DIFF
--- a/.github/workflows/bump_packages.yml
+++ b/.github/workflows/bump_packages.yml
@@ -39,7 +39,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get install -y --no-install-recommends rpm git-annex python3-semver
-        sudo gem install gem2rpm --no-document
+        sudo gem install gem2rpm nokogiri --no-document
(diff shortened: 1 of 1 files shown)


evgeni:
Huh, interesting. This only should be called when it tries to add a new package, not to update one?!

ehelms:
> Huh, interesting. This only should be called when it tries to add a new package, not to update one?!

Fair.. let me look at that.

ehelms:
This looks better https://github.com/theforeman/foreman-packaging/pull/9865

PR #9878, opened 2023-10-11
Fix how add_npm_package determines if an update is needed

ehelms:

### DIFF
--- a/add_npm_package.sh
+++ b/add_npm_package.sh
@@ -18,6 +18,7 @@ else
   PACKAGE_NAME=nodejs-${PACKAGE_MODULE}
 fi
 PACKAGE_DIR=packages/$BASE_DIR/$PACKAGE_NAME
+SPEC_FILE="${PACKAGE_DIR}/${PACKAGE_NAME}.spec"
 
 ROOT=$(git rev-parse --show-toplevel)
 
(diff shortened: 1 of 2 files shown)


add_npm_package.sh
```diff
@@ -138,7 +139,7 @@ if [[ -z $STRATEGY ]] ; then
   echo "Found $DEPENDENCIES dependencies - using $STRATEGY strategy"
 fi
 
-if [ -f "$PACKAGE_DIR"/*.spec ]; then
```
ekohl, line 141: Context: I think this predates from when we properly aligned all spec files to match their directory name. These days we can rely on that.

add_npm_package.sh
```diff
@@ -18,6 +18,7 @@ else
   PACKAGE_NAME=nodejs-${PACKAGE_MODULE}
 fi
 PACKAGE_DIR=packages/$BASE_DIR/$PACKAGE_NAME
+SPEC_FILE="${PACKAGE_NAME}.spec"
```
ekohl, line 21: Should we make this an absolute path?

ekohl (changes requested):
Can you also use the new variable on line 150?

ehelms:
> Can you also use the new variable on line 150?

Done

PR #9886, opened 2023-10-12
Add a Containerfile to represent the current build environment

ehelms:
All our various scripts are in a variety of languages and dependencies, which I lose track of and sometimes need a defined environment in order to perform actions.

Additionally, the NPM ecosystem with our packaging requires that the sources be generated with the right version of NodeJS which is not available locally on a more up to date Fedora for example. This provides an easy to use environment that is configured with that right set of versions.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,21 @@
+FROM centos:stream8
+
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+
+RUN echo "[git-annex]" >> /etc/yum.repos.d/git-annex.repo \
+         && echo "name=git-annex" >> /etc/yum.repos.d/git-annex.repo \
+         && echo "baseurl='https://downloads.kitenet.net/git-annex/linux/current/rpms/'" >> /etc/yum.repos.d/git-annex.repo \
+         && echo "gpgcheck=0" >> /etc/yum.repos.d/git-annex.repo \
+         && echo "enabled=1" >> /etc/yum.repos.d/git-annex.repo
+
+RUN dnf -y module enable nodejs:14 ruby:2.7
+RUN dnf -y install nodejs vim git rpmdevtools git-annex-standalone wget python3 ruby jq ruby-devel make gcc-c++ postgresql-devel libvirt-devel libxml2-devel libcurl-devel systemd-devel
+
+RUN npm install npm2rpm --global
+
+RUN pip3 install packaging requests
+
+RUN mkdir -p /opt/foreman-packaging
+WORKDIR /opt/foreman-packaging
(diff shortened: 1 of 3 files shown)


enter-container
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+podman run --privileged -it -v ./:/opt foreman-packaging
```
ekohl, line 3: I always use `--rm` as well. Also, is it wise to use `/opt` globally? Wouldn't `/opt/foreman-packaging` be better? And with is `--privileged` needed?

evgeni:
How does this compare to the packaging box we have in forklift (no, I have never used it) and should it be retired then?

https://github.com/theforeman/forklift/blob/master/vagrant/boxes.d/03-packaging.yaml

ehelms:
> How does this compare to the packaging box we have in forklift (no, I have never used it) and should it be retired then?
> 
> https://github.com/theforeman/forklift/blob/master/vagrant/boxes.d/03-packaging.yaml

No idea, I never use it. So yes?

Containerfile
```diff
@@ -0,0 +1,21 @@
+FROM centos:stream8
+
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+
+RUN echo "[git-annex]" >> /etc/yum.repos.d/git-annex.repo \
```
evgeni, line 7:
```
curl -o /etc/yum.repos.d/git-annex.repo  https://downloads.kitenet.net/git-annex/linux/current/rpms/git-annex.repo
```

How about that?

Containerfile
```diff
@@ -0,0 +1,21 @@
...
+         && echo "name=git-annex" >> /etc/yum.repos.d/git-annex.repo \
+         && echo "baseurl='https://downloads.kitenet.net/git-annex/linux/current/rpms/'" >> /etc/yum.repos.d/git-annex.repo \
+         && echo "gpgcheck=0" >> /etc/yum.repos.d/git-annex.repo \
+         && echo "enabled=1" >> /etc/yum.repos.d/git-annex.repo
+
+RUN dnf -y module enable nodejs:12 ruby:3.0
```
evgeni, line 13: Why 3.0?

Containerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN dnf -y install nodejs vim git rpmdevtools git-annex-standalone wget python3 ruby jq ruby-devel make gcc-c++ postgresql-devel libvirt-devel libxml2-devel libcurl-devel systemd-devel
+
+RUN npm install npm2rpm --global
+
+RUN pip3 install packaging requests
```
evgeni, line 18: Shouldn't this install obal and the rest come via deps?

ehelms: We can install Obal. This is actually required by one of the scripts in foreman-packaging.

Containerfile
ehelms: Solid question. I will switch this.

evgeni (approved):

PR #9888, opened 2023-10-13
Update rubygem-apipie-dsl to 2.6.0

ofedoren:


packages/foreman/rubygem-apipie-dsl/rubygem-apipie-dsl.spec
```diff
@@ -1,64 +1,43 @@
...
 
-Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 2.5.0
+Name: rubygem-%{gem_name}
+Version: 2.6.0
 Release: 1%{?dist}
 Summary: Ruby DSL documentation tool
-Group: Development/Languages
 License: MIT
 URL: https://github.com/ofedoren/apipie-dsl
```
ekohl, line 9: This is taken from the gem metadata, but that should be updated to https://github.com/Apipie/apipie-dsl.

ofedoren: Ohhh... I forgot to update it in the gem after I've transferred it under Apipie org :/ I'll update here, and there, but I hope we don't need another release just for that?

ekohl: No, only if we regenerate the spec file. We don't do that often. I'll just happen naturally when there's a new release.

ekohl (approved):

PR #9889, opened 2023-10-13
Require fapolicyd rules packages if fapolicyd present

ehelms:

### DIFF
--- a/packages/foreman/foreman-proxy/foreman-proxy.spec
+++ b/packages/foreman/foreman-proxy/foreman-proxy.spec
@@ -1,7 +1,7 @@
 %global homedir %{_datadir}/%{name}
 %global confdir config
 
-%global release 1
+%global release 2
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
@@ -27,6 +27,9 @@ BuildRequires:  ruby(release) >= 2.5
 Requires:       ruby(release) >= 2.5
 Requires:       rubygems
 Requires:       rubygem(bundler_ext)
+
+# Require fapolicyd package if fapolicyd is present
+Requires: (%{name}-fapolicyd if fapolicyd)
 
 Requires:       foreman-debug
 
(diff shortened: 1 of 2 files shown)


packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -28,6 +28,9 @@ Requires:       ruby(release) >= 2.5
 Requires:       rubygems
 Requires:       rubygem(bundler_ext)
 
+# Require fapolicyd package if fapoolicyd is present
+Requires: (%{name}-fapolicyd if fapolicyd)
```
evgeni, line 32: when this is in place and we've sufficiently tested the flow, I think we can drop the `foreman-fapolicyd → foreman-proxy-fapolicyd` require that you added, as the installation would now be "managed" by the packages here

packages/foreman/foreman-proxy/foreman-proxy.spec
```diff
@@ -28,6 +28,9 @@ Requires:       ruby(release) >= 2.5
 Requires:       rubygems
 Requires:       rubygem(bundler_ext)
 
+# Require fapolicyd package if fapoolicyd is present
```
evgeni, line 31:
```suggestion
# Require fapolicyd package if fapolicyd is present
```

fa :hankey: licyd?

ehelms: You said it, not me.

ehelms:
[test rpm]

ehelms:
[test rpm-copr]

evgeni (approved):
ack pending typo fix (twice)

ekohl (approved):

ekohl:
Tests are still running. Merge when green.

packages/foreman/foreman-proxy/foreman-proxy.spec
evgeni: https://github.com/theforeman/foreman-packaging/pull/9901

PR #9900, opened 2023-10-18
Bump yggdrasil to 0.2.3

adamruzicka:
- Patch yggdrasil to pass FOREMAN_REX_WORKDIR env var to the workers
- Bump yggdrasil to 0.2.3


ekohl:
3.6 is now end of life. Do we still want this?

adamruzicka:
No, it is not really necessary

PR #9911, opened 2023-10-24
add foreman_theme_satellite

evgeni:


packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
```diff
@@ -0,0 +1,93 @@
...
+        %{buildroot}%{gem_dir}/
+
+%foreman_bundlerd_file
+%foreman_precompile_plugin -s
+
+%files
+%dir %{gem_instdir}
+%license %{gem_instdir}/LICENSE
+%{gem_instdir}/app
+%exclude %{gem_instdir}/config/metadata.yml
```
ekohl, line 69: Why is this excluded?

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
```diff
@@ -0,0 +1,93 @@
...
+License: GPLv3
+URL: https://github.com/RedHatSatellite/foreman_theme_satellite
+Source0: https://rubygems.org/downloads/%{gem_name}-%{version}.gem
+
+# start specfile generated dependencies
+Requires: foreman >= %{foreman_min_version}
+Requires: ruby >= 2.7
+BuildRequires: foreman-assets >= %{foreman_min_version}
+BuildRequires: foreman-plugin >= %{foreman_min_version}
+Requires: ruby >= 2.7
```
ekohl, line 19: I wonder why ruby is listed twice.

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
evgeni: it's only really needed during devel builds, for prod we read it from the satellite RPM that places it in a different path:
https://github.com/RedHatSatellite/foreman_theme_satellite/blob/develop/lib/foreman_theme_satellite/engine.rb#L132

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
evgeni: solid q, I have no idea why the template did it like this.

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
ekohl: https://github.com/RedHatSatellite/foreman_theme_satellite/pull/9

Though when I completed it I realized it may have been sufficient to only exclude it in the gemspec. Still, I think this is cleaner.

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
evgeni: https://github.com/theforeman/foreman-packaging/blob/0959e5e0625daa62f6693c91ba27961cb240a90f/gem2rpm/foreman_plugin.spec.erb#L46-L48

https://github.com/theforeman/foreman-packaging/blob/0959e5e0625daa62f6693c91ba27961cb240a90f/gem2rpm/foreman_plugin.spec.erb#L56-L58

I think it was added here https://github.com/theforeman/foreman-packaging/pull/8079/files#diff-b2bcfeb556d566a4b0e36677afbe6912487b4aecf0a06416a72242fcd7827d99

ekohl:
Note to self: still failing because there is no gem available yet.

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
ekohl: https://github.com/theforeman/foreman-packaging/pull/9916

evgeni:
> Note to self: still failing because there is no gem available yet.

it is now!

package_manifest.yaml
```diff
@@ -592,6 +592,8 @@ katello_packages:
         - "el8-foreman-{{ foreman_version }}"
         - "el8-foreman-plugins-{{ foreman_version }}"
         - "el8-katello-candlepin-{{ katello_version }}"
+  children:
+    satellite_packages: {}
```
ehelms, line 596: This is the slightly yucky part being stuck under here, but I assume you are doing so to prevent needing another repository since it needs katello.

ehelms (approved):

package_manifest.yaml
evgeni: Well, it doesn't *need* katello, only none-core dep is deface, so I could also put it into plugins.
This was mostly so I don't have to create an own tag to build into.

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
```diff
@@ -0,0 +1,91 @@
...
+# start specfile generated dependencies
+Requires: foreman >= %{foreman_min_version}
+BuildRequires: foreman-assets >= %{foreman_min_version}
+BuildRequires: foreman-plugin >= %{foreman_min_version}
+Requires: ruby >= 2.7
+BuildRequires: ruby >= 2.7
+BuildRequires: rubygems-devel
+BuildArch: noarch
+Provides: foreman-plugin-%{plugin_name} = %{version}
+BuildRequires: rubygem(activesupport)
```
ekohl, line 23: Not for here, but interesting that it explicitly lists it. I'd imagine it can rely on Foreman to pull it in as part of Rails.

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
```diff
@@ -0,0 +1,91 @@
...
+BuildRequires: rubygem(deface)
+# end specfile generated dependencies
+
+# start package.json devDependencies BuildRequires
```
ekohl, line 27: I'd at least expect `@theforeman/builder` to be here. I think that now has a strict dependency on an older version (`^4.12.0`) so perhaps that's also an issue that needs to be addressed.

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
evgeni: Yeah, it's the part of the gemspec I didn't clean up yet ;)

packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
evgeni: let me regen the spec and see what happens

ekohl (approved):

evgeni:
```
% koji add-pkg --owner kojiadm katello-nightly-el8 rubygem-foreman_theme_satellite
Adding 1 packages to tag katello-nightly-el8
```

PR #9915, opened 2023-10-25
Swap to NodeJS 14

ehelms:


ekohl (commented):
Are we keeping track of doing this somewhere? Like also updating NodeJS on Debian and in our CI?

Edit: I see https://github.com/theforeman/foreman-infra/pull/1953 & https://github.com/theforeman/foreman/pull/9875

evgeni (approved):

PR #9917, opened 2023-10-26
Add foreman-build to Copr chroots

ehelms:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -43,6 +43,8 @@ copr_projects:
       - 'python39:3.9'
       - 'nodejs:12'
       - 'postgresql:12'
+    core_buildroot_packages:
+      - 'foreman-build'
     rhel_9: '9'
     rhel_8: '8'
     rhel_7: '7'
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -62,6 +62,8 @@ copr_projects:
             - "https://yum.puppet.com/puppet7/el/{{ rhel_8 }}/x86_64/"
             - "{{ foreman_staging }}/rhel-{{ rhel_8 }}-x86_64"
           comps_file: "{{ inventory_dir }}/comps/comps-foreman-el{{ rhel_8 }}.xml"
+          buildroot_packages:
+            - foreman-build
```
evgeni, line 66: should this be `core_buildroot_packages` like we have `core_modules` so it's one place to update?

evgeni (approved):

package_manifest.yaml
ehelms: I had this feeling...

evgeni: You know me too well…

PR #9939, opened 2023-10-31
introduce a foreman-devel module

evgeni:

### DIFF
--- a/modulemd/modulemd-foreman-el8.yaml
+++ b/modulemd/modulemd-foreman-el8.yaml
@@ -26,3 +26,24 @@ data:
         platform: ['el8']
         ruby: ['2.7']
         postgresql: ['12']
+...
+---
+document: modulemd
+version: 2
+data:
+  name: foreman-devel
+  stream: el8
+  version: 0
+  arch: x86_64
+  summary: Foreman development module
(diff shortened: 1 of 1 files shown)


evgeni:
requires:
- [ ] https://github.com/theforeman/theforeman-rel-eng/pull/288

ekohl:
> (OTOH, nothing uses this `buildrequires` entry anyway)

Because of this I would do this. On a technical level it doesn't have implications, but avoids duplication.

ekohl:
And now I see you added `foreman` as a dependency to `foreman-devel`, so that makes me reconsider. I'd avoid that potential circular dependency.

modulemd/modulemd-foreman-el8.yaml
```diff
@@ -26,3 +26,24 @@ data:
         platform: ['el8']
         ruby: ['2.7']
         postgresql: ['12']
+...
+---
+document: modulemd
+version: 2
+data:
+  name: foreman-devel
```
evgeni, line 34: I picked `foreman-devel` over `foreman-build` as it's useful for non-build things too, but the name really does not matter much.

ekohl (approved):

PR #9955, opened 2023-11-08
Update puppet-foreman_scap_client to 1.0.0

ekohl:
This updates the dependencies to also include puppet-agent since puppet-agent 8.0.0 it no longer Provides: puppet.


evgeni (approved):

ekohl:
Bleh, I also need to bump puppetlabs-stdlib:
```
 Problem: package puppet-foreman_scap_client-1.0.0-1.el8.noarch from @commandline requires puppetlabs-stdlib >= 4.25.0, but none of the providers can be installed
  - conflicting requests
  - nothing provides puppet >= 5.5.10 needed by puppetlabs-stdlib-5.2.0-1.el8.noarch from foreman-plugins
  - nothing provides puppet < 7.0.0 needed by puppetlabs-stdlib-5.2.0-1.el8.noarch from foreman-plugins
```

ekohl:
But that's only for Puppet 8. I can merge this today for Puppet 7.

packages/plugins/puppet-foreman_scap_client/puppet-foreman_scap_client.spec
```diff
@@ -30,24 +28,21 @@ cp -rp . %{buildroot}/%{puppet_foreman_scap_client_dir}/
...
 
 %files
-%doc NEWS COPYING README.md
-%dir %{puppet_modules_dir}/%{puppet_module}
-%{puppet_foreman_scap_client_dir}/COPYING
-%{puppet_foreman_scap_client_dir}/README.md
-%{puppet_foreman_scap_client_dir}/NEWS
+%doc NEWS README.md REFERENCE.md
+%dir %{puppet_foreman_scap_client_dir}
+%license %{puppet_foreman_scap_client_dir}/COPYING
```
ekohl, line 33: Bleh, I just realized this probably should have been `%license COPYING` and then `%exclude` this file so it's properly placed in `/usr/share/licenses`. Still, this isn't worse than it used to be.

PR #9990, opened 2023-11-16
EL9 yolo

evgeni:


package_manifest.yaml
```diff
@@ -65,6 +65,12 @@ copr_projects:
             - "{{ foreman_staging }}/rhel-{{ rhel_8 }}-x86_64"
           comps_file: "{{ inventory_dir }}/comps/comps-foreman-el{{ rhel_8 }}.xml"
           buildroot_packages: "{{ core_buildroot_packages }}"
+        - name: "rhel-{{ rhel_9 }}-x86_64"
+          external_repos:
+            - "https://yum.puppet.com/puppet7/el/{{ rhel_9 }}/x86_64/"
+            #- "{{ foreman_staging }}/rhel-{{ rhel_9 }}-x86_64"
```
evgeni, line 71: there is no staging repo yet for el9

package_manifest.yaml
```diff
@@ -65,6 +65,12 @@ copr_projects:
...
+          comps_file: "{{ inventory_dir }}/comps/comps-foreman-el{{ rhel_9 }}.xml"
+          #buildroot_packages: "{{ core_buildroot_packages }}"
```
evgeni, line 73: there is no `foreman-build` for el9 yet.

package_manifest.yaml
ehelms: What happens if you go ahead and define it here? As soon as you run this to create the EL9 chroot, this staging repository will exist (since that chroot will trigger Copr to create it).

package_manifest.yaml
ehelms: Does this break builds if we define the package to be included and it does not exist?

ehelms:
Something to consider, since we are bootstrapping for the first time in a while, and we know we will bootstrap again for EL10. For Pulp I introduced tiers in the package_manifest to represent the order in which things must get built to make the bootstrapping easier to repeat: 

https://github.com/theforeman/pulpcore-packaging/blob/rpm/3.28/package_manifest.yaml#L36

package_manifest.yaml
evgeni: Yes, mock freaks out it can't install things and aborts.

package_manifest.yaml
evgeni: Within this PR it would fail until someone with the right permissions calls `obal copr-project …` to create it.

But yes, this is the plan, I just wanted to see if things otherwise build OK.

evgeni:
this was a playground, #10015 was the final one we merged

PR #9999, opened 2023-11-16
drop rubygem-unf and rubygem-unf_ext

evgeni:
they were only dependencies of rubygem-domain_name and that dropped the dep in the latest release


evgeni:
should these be added to `foreman-obsolete-packages`?

ekohl (approved):
Hard to say. It's not in EPEL, so it's not like we'll accidentally remove that, but could something else be using this? Given it shouldn't be needed anymore with Ruby 2.6+ and we generally don't support other packaged Ruby apps I'm leaning to including it in obsoletes.

Would be great to get this sorted before branching.

evgeni:
Updated!

ekohl (approved):

ekohl:
I think this is the first packages we drop with COPR. Is there a procedure? https://github.com/theforeman/foreman-packaging/tree/rpm/develop#howto-removing-a-package doesn't mention it.

cc @ehelms

ehelms:
Here ya go, my winning PR to enumerate that -- https://github.com/theforeman/foreman-packaging/pull/10000

evgeni:
```console
% copr-cli delete-package --name rubygem-unf @theforeman/foreman-nightly-staging
Package rubygem-unf was successfully deleted.
% copr-cli delete-package --name rubygem-unf_ext @theforeman/foreman-nightly-staging
Package rubygem-unf_ext was successfully deleted.
% koji untag-build --all foreman-nightly-el8 rubygem-unf
% koji untag-build --all foreman-nightly-el8 rubygem-unf_ext
% koji block-pkg foreman-nightly-el8  rubygem-unf  rubygem-unf_ext 
```

PR #10002, opened 2023-11-20
Fix dependency on racc in gettext

ekohl:
Gettext 3.4.4 added racc as an [explicit dependency](https://github.com/ruby-gettext/gettext/commit/378f59555d5b3fd606bb0c0f98b87f2711d7049d) because Ruby 3.3 no longer bundles racc. Since we're not on Ruby 3.3 yet, we patch it out and replace it with a the bundled dependency. This is similar to what we do in nokogiri.


ehelms:
[test rpm-copr]

ekohl:
[test rpm-copr]

ekohl:
Oops, didn't see your comment was already there. Looked like some internal COPR issue.

PR #10016, opened 2023-11-23
Build Rails for EL9

evgeni:


evgeni:
This is an alternative to #10015, it has the benefit of being a smaller diff at the cost that now the el8 and el9 builds happen sequentially instead of in parallel, which makes it slower and thus not preferred by me.

ekohl:
I prefer #10015, for the reasons you mention.

PR #9962, opened 2023-11-10
Fix foreman_ygg_worker build on CentOS 9-stream

sbernhard:
This fixes a build issue on CentOS 9 stream:

```
+ head -c20 /dev/urandom
+ od -An -tx1
+ tr -d ' \n'
+ go build -buildmode pie -compiler gc '-tags=rpm_crashtraceback libtrust_openssl ' -ldflags ' -linkmode=external -compressdwarf=false -B 0x9155578dd8908ef952f7d130ca0a82e638b73252 -extldflags '\''-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 '\''' -a -v
go: open ../../../go.mod: no such file or directory
```

PR #10019, opened 2023-11-23
Bump foreman_ygg_worker

ekohl:
This is to verify it still builds. Reportedly, golang 1.21 broke the build: https://github.com/theforeman/foreman-packaging/pull/9962


dgoetz:
Looks like it still builds with golang 1.20.10 (https://download.copr.fedorainfracloud.org/results/%40theforeman/client-nightly-staging-scratch-3af8ee5e-150e-5a75-99b6-4231015161b4/rhel-9-x86_64/06684916-foreman_ygg_worker/chroot_scan/var/lib/mock/rhel-9-x86_64-1700751658.929146/root/var/log/dnf.rpm.log).

ekohl:
Oh yes, COPR builds against RHEL, not CentOS Stream. One of those cases where it's actually biting us.

PR #10021, opened 2023-11-23
Update ruby-foreman-tasks to 8.3.3

adamruzicka:

### DIFF
--- a/plugins/ruby-foreman-tasks/debian/changelog
+++ b/plugins/ruby-foreman-tasks/debian/changelog
@@ -1,3 +1,9 @@
+ruby-foreman-tasks (8.3.3-1) stable; urgency=low
+
+  * 8.3.3 released
+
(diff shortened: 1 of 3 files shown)


plugins/ruby-foreman-tasks/debian/changelog
```diff
@@ -1,3 +1,15 @@
+ruby-foreman-tasks (8.3.3-1) stable; urgency=low
+
+  * 8.3.3 released
+
+ -- Adam Ruzicka <aruzicka@redhat.com>  Thu, 23 Nov 2023 16:39:46 +0100
+
+ruby-foreman-tasks (8.3.3-1) stable; urgency=low
```
ekohl, line 7: Somehow this was log entry was duplicated

ekohl (changes requested):

plugins/ruby-foreman-tasks/debian/changelog
adamruzicka: Sorry about that, fixed

ekohl (approved):

PR #10024, opened 2023-11-24
Update foreman, foreman-proxy & katello dependencies

ekohl:
foreman and foreman-proxy use rich dependencies, but katello is gem based and gem2rpm doesn't understand that yet.

PR #10025, opened 2023-11-24
Implement rich dependencies in gem2rpm templates

ekohl:
For most templates this is pretty straight forward, but for foreman_plugin.spec.erb it combines asset reequirements into the spec.extensions block.  That was already using the more modern runtime_requirements.virtualize methods to generate a correct rpm dependency list. This was never noticed because we (thankfully) don't have foreman plugins with native extensions.

### DIFF
--- a/gem2rpm/default.spec.erb
+++ b/gem2rpm/default.spec.erb
@@ -67,9 +67,8 @@ BuildRequires: <%= requirement 'rubygems-devel', req %>
 <% end -%>
 <% unless spec.extensions.empty? -%>
 <% for dep in runtime_dependencies.virtualize -%>
-<% for req in dep.requirement -%>
-BuildRequires: <%= requirement(dep.name, req) %>
-<% end -%>
+<% build_requires = dep.requirement.map { |req| requirement(dep.name, req) } -%>
+BuildRequires: <%= build_requires.size > 1 ? "(#{build_requires.join(' with ')})" : build_requires.first %>
 <% end -%>
 # Compiler is required for build of gem binary extension.
 # https://fedoraproject.org/wiki/Packaging:C_and_C++#BuildRequires_and_Requires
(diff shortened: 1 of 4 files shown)


evgeni (dismissed):
I trust you… ;)

gem2rpm/foreman_plugin.spec.erb
```diff
@@ -56,27 +56,21 @@ BuildRequires: <%= requirement "ruby#{'-devel' unless spec.extensions.empty?}",
...
+<% unless spec.extensions.empty? -%>
 # Compiler is required for build of gem binary extension.
 # https://fedoraproject.org/wiki/Packaging:C_and_C++#BuildRequires_and_Requires
 BuildRequires: gcc
 <% end -%>
+<% end -%>
 <% if spec.extensions.empty? -%>
 BuildArch: noarch
 <% end -%>
 Provides: foreman-plugin-%{plugin_name} = %{version}
```
evgeni, line 73: Given https://github.com/theforeman/foreman-packaging/pull/10024#discussion_r1404559804, I think there is something wrong with the if/end here, but I don't see what.

evgeni (approved):

gem2rpm/foreman_plugin.spec.erb
ekohl: I can move it up for a smaller diff. And I can simplify the conditions.

ekohl:
I've merged the block that requires gcc and declares it as noarch in a simple if/else condition. Hopefully the end blocks are a bit easier to follow now.

evgeni (approved):

gem2rpm/default.spec.erb
```diff
@@ -67,9 +67,8 @@ BuildRequires: <%= requirement 'rubygems-devel', req %>
 <% end -%>
 <% unless spec.extensions.empty? -%>
 <% for dep in runtime_dependencies.virtualize -%>
-<% for req in dep.requirement -%>
-BuildRequires: <%= requirement(dep.name, req) %>
-<% end -%>
+<% build_requires = dep.requirement.map { |req| requirement(dep.name, req) } -%>
+BuildRequires: <%= build_requires.size > 1 ? "(#{build_requires.join(' with ')})" : build_requires.first %>
```
ekohl, line 71: https://github.com/fedora-ruby/gem2rpm/pull/124 would make this:
```suggestion
BuildRequires: <%= dep.to_rich_rpm %>
```

PR #10035, opened 2023-11-27
Build .mo files for katello

ekohl:
I found out there already was a task to create the .mo files. This avoids having to store them in git as part of the release. The goal of this PR is to find out how to get this working and then apply it for every plugin that needs it.


ekohl:
Summary:
* It builds the .mo files in `locale`
* It doesn't ship the .mo files in the RPM
* It doesn't build the .mo files in `engines/bastion/app/assets/javascripts/bastion/i18n/locale`

ekohl:
We wouldn't have to store .mo files in git anymore if we do this. We may still want to include them in the .gem files, but this would work with nightly as well. Could also be part of `gem build`. Playing with the options now.

PR #10039, opened 2023-11-28
Fix copr version in package_manifest for copr katello

zjhuntin:


package_manifest.yaml
```diff
@@ -78,7 +78,7 @@ copr_projects:
           comps_file: "{{ inventory_dir }}/comps/comps-foreman-plugins-el{{ rhel_8 }}.xml"
           buildroot_packages: "{{ core_buildroot_packages }}"
     katello-copr:
-      copr_project_name: "katello-{{ foreman_version }}-staging"
+      copr_project_name: "katello-{{ katello_version }}-staging"
```
Odilhao, line 81: @ehelms this should be Katellote_version and not Foreman-version right?

Odilhao (approved):

package_manifest.yaml
ehelms: Good catch, yes it should.

ehelms (approved):

PR #10053, opened 2023-11-29
Update debian keyring

zjhuntin:


evgeni (approved):
can you also update it in develop please?

PR #10065, opened 2023-12-01
build installer packages on el9

evgeni:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -509,7 +509,7 @@ foreman_nodejs_packages:
 foreman_installer_packages:
   vars:
     copr_projects:
-      - "{{ hostvars['foreman-copr'] }}"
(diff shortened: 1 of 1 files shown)


evgeni:
I didn't test `foreman-installer` itself, as that is the only package that has `BuildRequires`, but the others did build, so I am rather confident :)

evgeni:
yeah, no, idk.

to nicely test the proxy, we'll need the installer to deploy it. and if stuff builds, we have packaging prepped for it to get the ruby3 fixes once they appear (if any).

so :woman_shrugging:

ehelms (approved):
I'm OK with this... as we fix things those fixes will propagate and arrive. If all the packages build (which I do not think is tested here in the PR) that is.

evgeni:
> If all the packages build (which I do not think is tested here in the PR) that is.

not in the PR, no, but I tested them out-of-band :)

PR #10069, opened 2023-12-01
Use Candlepin dedicated repository

ehelms:


packages/katello/katello-repos/katello-repos.spec
```diff
@@ -1,11 +1,12 @@
 %global pulpcore_version 3.39
+%global candlepin_version 4.3
```
ehelms, line 2: I am not sure if I should start with 4.3 here or lean into the nightly version of the repository for `rpm/develop`.

packages/katello/katello-repos/katello.repo
```diff
@@ -10,10 +10,10 @@ gpgcheck=@REPO_GPGCHECK@
 # Candlepin RPMs as supported by Katello - This is a coordinated
 # copy of Candlepin's packages in order to ensure compatibility
 
-[katello-candlepin]
+[candlepin]
 name=Candlepin: an open source entitlement management system.
-baseurl=https://yum.theforeman.org/katello/@REPO_VERSION@/candlepin/@DIST@/$basearch/
-gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-foreman
+baseurl=https://yum.theforeman.org/candlepin/@CANDLEPIN_VERSION@/@DIST@/$basearch/
+gpgkey=https://yum.theforeman.org/candlepin/@CANDLEPIN_VERSION@/GPG-RPM-KEY-candlepin
```
ehelms, line 16: Copying how Pulpcore is setup in this file right now. There remains the question of how this GPG key ends up here in our newer process with Copr.

I am thinking I may need to add a step that copies the public GPG key into the repository structure during `build_stage_repository` action.

ekohl: Including a file is more reliable. Dnf doesn't check expirations and doesn't refresh if the URL didn't change. I ran into that on Fedora with my personal copr

packages/katello/katello-repos/katello-repos.spec
ekohl: I'd be fine with 4.3 so we can pick it and then update to nightly in a separate bump. But no strong feelings here

ehelms:
This is a related change that is needed -- https://github.com/theforeman/theforeman-rel-eng/pull/313

Odilhao (approved):

Odilhao:
Just hit this while trying to release Katello

https://jenkins-foreman.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/foreman-pipeline-katello-rpm-4.11/detail/foreman-pipeline-katello-rpm-4.11/1/pipeline/93

PR #10072, opened 2023-12-01
Rebuild python packages against python 3.11

Odilhao:


package_manifest.yaml
```diff
@@ -40,11 +40,12 @@ copr_projects:
   vars:
     core_modules:
       - 'ruby:2.7'
-      - 'python39:3.9'
       - 'nodejs:14'
       - 'postgresql:12'
     core_buildroot_packages:
       - 'foreman-build'
+      - python3.11-rpm-macros
+      - python3.11
```
ehelms, line 48: As this change has to be done actively, can you split it out to it's own PR to go in first?

evgeni:
Are those obsoletes really required?

Odilhao:
> Are those obsoletes really required?

My tests failed without it, but I was on `crunch` at the time so my test might be faulty, do you want me to remove that?

ekohl (commented):
Interesting that we never had any instructions to enable the Python 3.9 module. I guess we relied either on Pulpcore or things to automatically enable it?

evgeni:
> Interesting that we never had any instructions to enable the Python 3.9 module. I guess we relied either on Pulpcore or things to automatically enable it?

Nah, it gets enabled automatically when you install stuff from it.

(modularity is weird)

evgeni:
[test rpm-copr]

evgeni:
[test rpm-copr]

evgeni (approved):

PR #10094, opened 2023-12-05
Add client nightly staging to repoclosure

ehelms:


evgeni (changes requested):
I think we should get rid of the ones pointing at `koji.katello.org` to avoid confusion

PR #10106, opened 2023-12-07
Update rubygem-foreman_rh_cloud to 9.0.53

chris1984:


evgeni:
can you please bump in develop first?

chris1984:
Here is the develop one https://github.com/theforeman/foreman-packaging/pull/10121

evgeni:
thanks, I've merged that one.
can you make this PR a CP of the develop one, so the contents are identical? ty!

chris1984:
Updated

evgeni (approved):

PR #10115, opened 2023-12-12
Update foreman deps for webpack 5

evgeni:

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -4,7 +4,7 @@
 %global dynflow_sidekiq_service_name dynflow-sidekiq@
 %global rake /usr/bin/rake
 
-%global release 4
+%global release 5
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman/foreman.spec
```diff
@@ -315,35 +311,34 @@ Summary: Foreman asset pipeline support
 Group: Applications/System
 Requires: %{name} = %{version}-%{release}
 Requires: nodejs >= 14
+Requires: /usr/bin/npx
```
ekohl, line 314: Node to self: this is https://github.com/theforeman/foreman/blob/89c610451c9ea22eb426826e405518e04d68972f/lib/tasks/plugin_assets.rake#L55

ekohl (approved):

evgeni:
ain't nobody cares for repoclosure

PR #10116, opened 2023-12-12
[webpack5] bump webpack related packages to their required versions

evgeni:


ehelms:
This looks like it's failing due to the archive being generated on NodeJS 16+. You can use the container I was working on (because of this problem actually) https://github.com/theforeman/foreman-packaging/pull/9886

evgeni:
Interestingly, https://github.com/theforeman/foreman-packaging/pull/10119 passed.
It was generated with the same new node/npm from Fedora, but with older (6.0 vs 6.1) npm2rpm which does not optimize the generated cache tarball.

packages/foreman/nodejs-sass-loader/nodejs-sass-loader.spec
```diff
@@ -4,32 +4,15 @@
...
-Source11: https://registry.npmjs.org/lru-cache/-/lru-cache-6.0.0.tgz
-Source12: https://registry.npmjs.org/neo-async/-/neo-async-2.6.2.tgz
-Source13: https://registry.npmjs.org/punycode/-/punycode-2.3.1.tgz
-Source14: https://registry.npmjs.org/sass-loader/-/sass-loader-10.4.1.tgz
-Source15: https://registry.npmjs.org/schema-utils/-/schema-utils-3.3.0.tgz
-Source16: https://registry.npmjs.org/semver/-/semver-7.5.4.tgz
-Source17: https://registry.npmjs.org/uri-js/-/uri-js-4.4.1.tgz
-Source18: https://registry.npmjs.org/yallist/-/yallist-4.0.0.tgz
-Source19: nodejs-sass-loader-%{version}-registry.npmjs.org.tgz
+Source0: https://registry.npmjs.org/neo-async/-/neo-async-2.6.2.tgz
```
ekohl, line 13: If it's just a single dependency, it's tempting to package it as non-bundled to avoid including the tarball with JSON responses in our repository history.

ekohl (commented):
In the final version of this PR, will you squash things?

evgeni:
we can actually drop url-loader now, nice

ekohl (approved):

PR #10118, opened 2023-12-12
[webpack5] new packages

evgeni:


ekohl:
This includes https://github.com/theforeman/foreman-packaging/pull/10117 now?

evgeni:
it does, because I didn't want to have too many conflicts :)
(and still managed to get some)

evgeni:
I'll never understand why GH can't solve trivial rebases itself…

PR #10120, opened 2023-12-12
Hardcode EL7 dist to 7 in foreman-client-release

ehelms:
With the switch to building against RHEL, this value can end up being el7_9 resulting in pointing at a repository that does not exist.

Example:

```
@ehelms I think https://yum.theforeman.org/client/nightly/el7/x86_64/foreman-client-release.rpm might be leading to the wrong repository:

failure: repodata/repomd.xml from foreman-client: [Errno 256] No more mirrors to try.
https://yum.theforeman.org/client/nightly/el7_9/x86_64/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found
Thanks for the heads up!
```

### DIFF
--- a/packages/foreman/foreman-release/foreman-release.spec
+++ b/packages/foreman/foreman-release/foreman-release.spec
@@ -10,10 +10,13 @@
 
 %else
 %define repo_dir %{_sysconfdir}/yum.repos.d
-%define repo_dist %{dist}
+
+%if 0%{?rhel}
+%define repo_dist el%{rhel}
+%endif
 %endif
 
-%global release 1
+%global release 2
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


ehelms:
This will need to get cherry-picked to 3.9

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -10,10 +10,15 @@
 
 %else
 %define repo_dir %{_sysconfdir}/yum.repos.d
+
+%if 0%{?rhel} == 7
+%define repo_dist 7
```
evgeni, line 15: shouldn't that be `el7`?

ehelms: yessss

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -10,10 +10,15 @@
 
 %else
 %define repo_dir %{_sysconfdir}/yum.repos.d
+
+%if 0%{?rhel} == 7
+%define repo_dist el7
+%else
```
evgeni, line 16: I wonder if instead this should be
```suggestion
%if 0%{?rhel}
%define repo_dist el%{rhel}
%else
```
to avoid similar surprises in 8.y lifetime

evgeni (approved):
I am cool with the fix, but also posted a comment about possibly broadening it.

evgeni (approved):

PR #10136, opened 2023-12-15
build proxy packages on el9

evgeni:


evgeni:
I tested all packages in this group, they build fine.

ehelms (approved):

PR #10144, opened 2023-12-18
Build core Ruby packages on EL9

evgeni:


evgeni:
rkerberos fails because rake-compiler is not built -- once that is present, it builds fine (I tried)

packages/foreman/rubygem-unicode/rubygem-unicode.spec
```diff
@@ -7,13 +7,13 @@
...
 Version: 0.4.4.4
-Release: 4%{?dist}
+Release: 5%{?dist}
 Summary: Unicode normalization library
 Group: Development/Languages
 License: Ruby
 URL: http://www.yoshidam.net/Ruby.html#unicode
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
-Source1: https://www.ruby-lang.org/en/about/license.txt
+Source1: license.txt
```
evgeni, line 16: the remote-URL-but-not-annexed confused the `srpm`module:

```
TASK [build_package : Build SRPM] *****************************************************************************************************************************************************************************************************
fatal: [rubygem-unicode]: FAILED! => changed=false 
  command: rpmbuild -bs --define _topdir /tmp/tmpiskuazz5 --define _sourcedir /tmp/tmpiskuazz5/SOURCES --define _builddir /tmp/tmpiskuazz5/BUILD --define _srcrpmdir /tmp/tmpiskuazz5 --define _rpmdir /tmp/tmpiskuazz5 --undefine dist /tmp/tmpiskuazz5/rubygem-unicode.spec
  msg: Failed to srpm build
  output: |-
    setting SOURCE_DATE_EPOCH=1615420800
    error: Bad file: /tmp/tmpiskuazz5/SOURCES/license.txt: No such file or directory
  
    RPM build errors:
        Bad file: /tmp/tmpiskuazz5/SOURCES/license.txt: No such file or directory
```

This was the quickest fix, but we might want to fix the module too. :eyes: @ehelms

packages/foreman/rubygem-sqlite3/rubygem-sqlite3.spec
```diff
@@ -22,6 +22,7 @@ BuildRequires: %{?scl_prefix_ruby}ruby(release)
 BuildRequires: %{?scl_prefix_ruby}ruby-devel >= 1.8.7
 BuildRequires: %{?scl_prefix_ruby}rubygems-devel >= 1.3.5
 Provides: %{?scl_prefix}rubygem(%{gem_name}) = %{version}
+BuildRequires: gcc
```
ekohl, line 25: If we regenerate the spec I think this should be automatically done. There is an update to 1.6.9 available, so perhaps just do that? Also drops the SCL stuff.

packages/foreman/rubygem-unicode/rubygem-unicode.spec
ekohl: I think this gem isn't even used anymore:
```console
# dnf repoquery --whatrequires rubygem-unicode --quiet
```

Digging further tells me it was dropped from Hammer: https://github.com/theforeman/hammer-cli/pull/376

evgeni: Oh nice, will drop it then

(the bug in `srpm` remains ;-) )

ekohl: I don't think it's a bug: you don't provide the source, or a way to verify the source if it is retrieved. We shouldn't support that.

ehelms: Aye, unicode should be a goner now (https://github.com/theforeman/hammer-cli/commit/327f19163e74fdc1a418c5157c956cd0f6d5fdf2)

evgeni: Well, the file is in git. Setting the `Source0` to an URL just denotes where the file came from.

evgeni: https://github.com/theforeman/foreman-packaging/pull/10145

packages/foreman/rubygem-sqlite3/rubygem-sqlite3.spec
evgeni: sqlite3 1.5.0+ introduces a dep on mini_portile2, which I didn't want to open up as a can-o-worms right now. only updating to 1.4.4 then,

evgeni:
[test rpm-copr]

(released rake-compiler)

evgeni:
that's as good as it gets for now.

evgeni:
https://github.com/theforeman/foreman-packaging/pull/10149

sassc actually also build-requires ffi, so moved it too

PR #10145, opened 2023-12-18
Drop rubygem-unicode

evgeni:


ekohl (approved):
I don't think it needs to be obsoleted, given it's quite generic and could conflict with other packages. Almost makes me think we need a `foreman-obsoletes` and `foreman-obsoletes-strict` where the former obsoletes just what we package (and would never expect to be elsewhere, like foreman plugins) and the latter anything we've ever packaged but no longer do. Even if things like EPEL could also package it.

evgeni:
```
% copr-cli delete-package --name rubygem-unicode @theforeman/foreman-nightly-staging
Something went wrong:
Error: No package with name rubygem-unicode in copr foreman-nightly-staging
```

@ehelms you never *built* that one? cheater.

PR #10151, opened 2023-12-19
add libsass

evgeni:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -246,6 +246,7 @@ ruby_core_packages:
 
 ruby_core_packages_tier1:
   hosts:
+    libsass: {}
     rubygem-activerecord-nulldb-adapter: {}
     rubygem-activerecord-session_store: {}
     rubygem-addressable: {}
--- a/packages/foreman/libsass/libsass.spec
+++ b/packages/foreman/libsass/libsass.spec
@@ -1,0 +1,137 @@
+Name:           libsass
+Version:        3.6.4
+Release:        5%{?dist}
+Summary:        C/C++ port of the Sass CSS precompiler
+
+License:        MIT
+URL:            http://sass-lang.com/libsass
+Source0:        https://github.com/sass/libsass/archive/%{version}/%{name}-%{version}.tar.gz
+
+BuildRequires:  automake
+BuildRequires:  autoconf
+BuildRequires:  libtool
+BuildRequires:  pkgconfig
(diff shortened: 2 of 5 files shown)


evgeni:
this will require changes to `rubygem-sassc`, see https://github.com/theforeman/foreman-packaging/pull/10152

package_manifest.yaml
```diff
@@ -246,6 +246,7 @@ ruby_core_packages:
 
 ruby_core_packages_tier1:
   hosts:
+    libsass: {}
```
evgeni, line 249: feels a bit weird to have it in `ruby_core_packages`, but then again, it's needed for ruby things…

ekohl:
> this will require changes to `rubygem-sassc`, see #10152

Would it also move rubygem-sassc to a tier2 package?

packages/foreman/libsass/libsass.spec
```diff
@@ -0,0 +1,137 @@
...
+%configure --disable-static
+make %{?_smp_mflags}
+
+
+%install
+make install DESTDIR=$RPM_BUILD_ROOT
+find $RPM_BUILD_ROOT -name '*.la' -exec rm -f {} ';'
+
+
+%ldconfig_scriptlets
```
ekohl, line 53: Is this supported in EL8? https://docs.fedoraproject.org/en-US/epel/epel-packaging/#shared_libraries phrasing implies it is, but it's not explicit.

evgeni:
> > this will require changes to `rubygem-sassc`, see #10152
> 
> Would it also move rubygem-sassc to a tier2 package?

It already is (as it depends on ffi)

packages/foreman/libsass/libsass.spec
evgeni: Kinda sure, yeah.

But will double check

ekohl (approved):
I'm assuming you're adding it on both EL 8 & 9 so we're consistent on all platforms, rather than only adding it with .so.1 on EL9 & use the system .so.0 on EL8.

:+1: assuming the macro does work on EL8.

packages/foreman/libsass/libsass.spec
evgeni: Interestingly, on both EL9 and EL8 this ends up generating nothing:

```
[root@15b1f7780590 /]# rpm -q --scripts libsass
[root@15b1f7780590 /]# 
```

PR #10152, opened 2023-12-19
Update libsass.so ABI to 1 for libsass >= 3.5

evgeni:

### DIFF
--- a/packages/foreman/rubygem-sassc/rubygem-sassc.spec
+++ b/packages/foreman/rubygem-sassc/rubygem-sassc.spec
@@ -5,7 +5,7 @@
 
 Name: rubygem-%{gem_name}
 Version: 2.4.0
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: Use libsass with Ruby!
 License: MIT
 URL: https://github.com/sass/sassc-ruby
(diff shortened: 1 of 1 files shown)


packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -19,8 +19,8 @@ BuildRequires: rubygem(ffi) >= 1.9
 BuildRequires: rubygem(ffi) < 2
 # end specfile generated dependencies
 
-BuildRequires: libsass
-Requires: libsass
+BuildRequires: libsass >= 3.5
+Requires: libsass >= 3.5
```
ekohl, line 23: Let's make it depend on `libsass.so.1()(64bit)` instead, just like we did in openscap:
https://github.com/theforeman/foreman-packaging/blob/d7af65f62c072a02aba04b4027d16603c375477f/packages/plugins/rubygem-openscap/rubygem-openscap.spec#L19-L20

```suggestion
Requires: (libsass.so.1()(64bit) if libc.so.6()(64bit))
Requires: (libsass.so.1 if libc.so.6)
```

evgeni: Would that also work for the `BuildRequire`? Or should that be kept?

ekohl: I purposefully avoided that question by only talking about `Requires` and hoped you'd know. I think it should work, but I guess there's one way to find out.

evgeni: tested it in a copr and worked fine, cool

evgeni:
I am going to merge this, assuming your yesterdays review counts as "ack once this change is applied", as otherwise we have a newer libsass in the repo, but no compatible rubygem-sassc thus resulting in failed builds.

evgeni:
(repoclosure won't be happy, but such is life)

PR #10155, opened 2023-12-19
Release 3.9.0

zjhuntin:


Odilhao (approved):

zjhuntin:
[test rpm-copr]

zjhuntin:
[test rpm-copr]

evgeni:
AFAIK the repoclosure issue is that it does not use the *current* foreman-3.9 repo as a lookaside, so any packages built there fail closure.

I am going to call "this is fine" for now, but we should fix that after GA :)

PR #10157, opened 2023-12-19
Update rubygem-foreman_rh_cloud to 9.0.54

chris1984:


chris1984:
@ekohl here is the packaging pr with your translations in it

evgeni (approved):

PR #10159, opened 2023-12-19
Drop koji repoclosure targets

ehelms:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -144,7 +144,7 @@ foreman_packages:
       - "{{ all_tags['foreman-el8'] }}"
     repoclosure_target_repos:
       el8:
-        - "el8-foreman-{{ foreman_version }}"
+        - "el8-foreman-{{ foreman_version }}-staging"
     repoclosure_lookaside_repos:
       el9:
         - el9-baseos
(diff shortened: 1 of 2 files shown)


package_manifest.yaml
```diff
@@ -923,88 +923,11 @@ repoclosures:
...
-          - el8-puppet-7
-          - "el8-foreman-{{ foreman_version }}-staging"
-          - "el8-foreman-plugins-{{ foreman_version }}-staging"
-          - "el8-candlepin-{{ candlepin_version }}-staging"
+          - "el8-candlepin-{{ candlepin_version }}"
     foreman-client-staging-repoclosure-el9:
       repoclosure_target_repos:
         el9:
-          - "el9-foreman-client-{{ foreman_version }}-staging"
+          - "el9-foreman-client-{{ foreman_version }}"
```
evgeni, line 930: I must admit I liked the `-staging` suffix, as it indicated correctly that we're not looking at the prod repos, but staging.

Not going to block on this tho :)

evgeni (approved):

package_manifest.yaml
ehelms: I liked it too... I just assumed that others would want a clean revert back to the way things were with koji gone. I'd be happy to keep the staging nomenclature.

ekohl: :+1: on keeping staging explicitly named.

ehelms:
Updated to use staging nomenclature for repoclosure everywhere.

evgeni (approved):

PR #10165, opened 2023-12-20
Drop koji

ehelms:
Requires https://github.com/theforeman/foreman-packaging/pull/10159 first

This removes all Koji references, and variables related to Koji including ensuring that Copr is the default build system.


add_gem_package.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash -e
 
 usage() {
-	echo "Usage: $0 GEM_NAME [TEMPLATE [PACKAGE_SUBDIR [KOJI_TAG]]]"
+	echo "Usage: $0 GEM_NAME [TEMPLATE [PACKAGE_SUBDIR [REPO]]]"
 	echo "Valid templates: $(ls gem2rpm | sed 's/.spec.erb//' | tr '\n' ' ')"
-	./list_koji_tags
+	echo "Valid repos: foreman-el8 foreman-plugins-el8 katello-el8 foreman-client"
```
evgeni, line 6: is this missing the `el8` from client intentionally?

ehelms: Client has many distros, and the logic appears to ignore that part anyway.

evgeni: Fair, but in L68 you write:
```
if [[ $REPO == "foreman-client-el8" ]] ; then
```

and the `add_gem_to_comps $REPO` in L64 only works for "valid" repos that match to comps files 1:1.

;-)

ehelms: > Fair, but in L68 you write:

I like to call it editing.

repoclosure/yum.conf
```diff
@@ -183,7 +152,3 @@ failovermethod=priority
 [f29-puppet-6]
 name=f29-puppet-6
 baseurl=http://yum.puppetlabs.com/puppet6/fedora/29/x86_64/
-
-[f29-foreman-client-nightly]
```
ekohl, line 187: I've been thinking about dropping the whole of f29.

ekohl (commented):
There's a merge conflict now.

repoclosure/yum.conf
ehelms: > I've been thinking about dropping the whole of f29.

:+1:  I think we should, we are way behind on Fedora releases, and we should drop this and then be intentional about our support.

ehelms:
This has been rebased

add_gem_package.sh
evgeni, line 6: I still think this should read
```suggestion
	echo "Valid repos: foreman-el8 foreman-plugins-el8 katello-el8 foreman-client-el8"
```

because in L64 you call `add_gem_to_comps $REPO` and `foreman-client` doesn't exist as a comps file and in L68 you match on `foreman-client-*` which requires the second dash.

evgeni (changes requested):

evgeni (approved):

PR #10175, opened 2023-12-21
rubygem-sassc: Make the package noarch

evgeni:

### DIFF
--- a/packages/foreman/rubygem-sassc/rubygem-sassc.spec
+++ b/packages/foreman/rubygem-sassc/rubygem-sassc.spec
@@ -1,11 +1,10 @@
 # template: default
 %global gem_name sassc
 %global gem_require_name %{gem_name}
-%define debug_package %{nil}
 
 Name: rubygem-%{gem_name}
 Version: 2.4.0
-Release: 2%{?dist}
+Release: 3%{?dist}
 Summary: Use libsass with Ruby!
 License: MIT
 URL: https://github.com/sass/sassc-ruby
@@ -15,9 +14,11 @@ Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 Requires: ruby >= 2.0.0
 BuildRequires: ruby-devel >= 2.0.0
 BuildRequires: rubygems-devel
-BuildRequires: rubygem(ffi) >= 1.9
-BuildRequires: rubygem(ffi) < 2
+BuildRequires: (rubygem(ffi) >= 1.9 with rubygem(ffi) < 2)
 # end specfile generated dependencies
+# the upstream source builds an embedded libsass
+# but we use the system one, which makes this package noarch
+BuildArch: noarch
 
 BuildRequires: (libsass.so.1()(64bit) if libc.so.6()(64bit))
 BuildRequires: (libsass.so.1 if libc.so.6)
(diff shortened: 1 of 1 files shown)


evgeni:
Currently repoclosure fails with:

```
    Error: Repoclosure ended with unresolved dependencies (1) across 1 packages.
    package: rubygem-sassc-2.4.0-2.el8.x86_64 from el8-foreman-nightly-staging
      unresolved deps (1):
        (libsass.so.1 if libc.so.6)
```

evgeni:
and the other repoclosure issue should be fixed by https://github.com/theforeman/foreman-packaging/pull/10176

evgeni:
But even when running with enough closure fixes, this doesn't work:

```
    package: rubygem-sassc-2.4.0-3.el8.noarch from repo0
      unresolved deps (1):
        (libsass.so.1 if libc.so.6)
```

evgeni:
oooh, is that maybe because there actually is no `libsass.so.1` for i686, while there *is* libopenscap.so.25`?

evgeni:
@ekohl given we do not build a 32bit libsass, I think I'd just drop the 32bit require and ignore it?

packages/foreman/rubygem-sassc/rubygem-sassc.spec
```diff
@@ -15,9 +14,11 @@ Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
...
-BuildRequires: rubygem(ffi) >= 1.9
-BuildRequires: rubygem(ffi) < 2
+BuildRequires: (rubygem(ffi) >= 1.9 with rubygem(ffi) < 2)
 # end specfile generated dependencies
+# the upstream source builds an embedded libsass
+# but we use the system one, which makes this package noarch
+BuildArch: noarch
 
 BuildRequires: (libsass.so.1()(64bit) if libc.so.6()(64bit))
 BuildRequires: (libsass.so.1 if libc.so.6)
```
ekohl, line 24: The idea was that this would be for non-x86_64 architectures as well, but I may be mistaken in that because at least for AARCH64 on Fedora 40 it also provides the 64 bit version.
```console
$ rpm -q --provides glibc-2.38.9000-29.fc40.aarch64.rpm | grep libc.so.6
libc.so.6()(64bit)
```


```suggestion
```

And then also remove the `Requires`.

ekohl (changes requested):

packages/foreman/rubygem-sassc/rubygem-sassc.spec
ekohl: Or just simplify it to `libsass.so.1` since that's always provided. EL 8+ are never built for i686 only so I don't think we'll ever need to worry about i686.

evgeni: One thing came to my mind: today this package is arch dependent (as the generation thinks it has an extension, even tho we patch it out).
If we keep it like this, we can drop the 32 bit things as they will never match arch wise (like you suggest), but we *also* can go ahead and use %isa to print the real arch for the dep on libsass.

ekohl: I'd prefer to make the package `noarch` (since it is). But then you can't use `%isa` because that's not defined on `noarch` packages.

evgeni:
@ekohl updated

evgeni:
Repoclosure requires https://github.com/theforeman/obal/pull/379 to be merge to work properly

ekohl (approved):

ekohl:
[test rpm-copr]

evgeni:
The closure error is for the package in the repo, not the changed one here. Nice

PR #10186, opened 2024-01-03
Correct deps for Ruby 3.0 bundled gems

evgeni:

### DIFF
--- a/packages/foreman/rubygem-mail/rubygem-mail.spec
+++ b/packages/foreman/rubygem-mail/rubygem-mail.spec
@@ -3,7 +3,7 @@
 
 Name: rubygem-%{gem_name}
 Version: 2.8.1
-Release: 1%{?dist}
+Release: 2%{?dist}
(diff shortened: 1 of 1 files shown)


packages/foreman/rubygem-mail/rubygem-mail.spec
```diff
@@ -17,7 +17,9 @@ BuildArch: noarch
 # end specfile generated dependencies
 
 # Prefer to consume net-smtp/net-imap/net-pop as a default gem
```
ekohl, line 19: Perhaps this should be weakened now?
```suggestion
# Allow to consume net-smtp/net-imap/net-pop as a default gem
```

ekohl (approved):

evgeni:
el9 closure fails because it doesn't look at foreman-staging-el9 because that doesn't exist (yet, this PR is part of the "what needs to be done to be able to generate that repo")

ekohl (approved):

PR #10187, opened 2024-01-03
Drop requirement on foreman-debug

evgeni:
This still packages the integration bits, just doesn't force the users to install foreman-debug.


ekohl (commented):
I'd prefer a Redmine issue for this. And then I'd probably do the same for `foreman`.

evgeni:
closure failure expected, it needs to look at "itself" but that doesn't exist for el9 yet

ekohl (approved):

PR #10190, opened 2024-01-03
there are no more 32bit EFI modules on EL9

evgeni:

### DIFF
--- a/packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
+++ b/packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
@@ -1,6 +1,6 @@
 Name: foreman-bootloaders-redhat
 Version: 202102220000
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: Metapackage with Grub2 and Shim TFTP bootloaders
 
 Group: Applications/System
@@ -11,7 +11,9 @@ BuildArch: noarch
 Source0: foreman-generate-bootloaders
 
 Requires: grub2-tools-extra
+%if 0%{?rhel} == 0
 Requires: grub2-efi-ia32-modules
+%endif
 Requires: grub2-efi-x64-modules
 Requires: grub2-pc-modules
 Requires: shim-ia32
@@ -69,6 +71,9 @@ install -Dp -m0755 %{SOURCE0} %{buildroot}%{_bindir}/foreman-generate-bootloader
 
 
 %changelog
+* Wed Jan 03 2024 Evgeni Golov - 202102220000-2
+- Drop requirement on grub2-efi-ia32-modules on EL9+
+
 * Fri Feb 19 2021 Oliver Freyermuth <o.freyermuth@googlemail.com> 202102220000-1
 - Add efinet module only on *-efi platforms.
 


evgeni:
this breaks el9  tftpboot generation, need to look  into it.

packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
```diff
@@ -11,7 +11,9 @@ BuildArch: noarch
 Source0: foreman-generate-bootloaders
 
 Requires: grub2-tools-extra
+%if 0%{?rhel} == 0
 Requires: grub2-efi-ia32-modules
+%endif
 Requires: grub2-efi-x64-modules
 Requires: grub2-pc-modules
 Requires: shim-ia32
```
ekohl, line 19: Is this still available on EL9?

packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
```diff
@@ -11,7 +11,9 @@ BuildArch: noarch
 Source0: foreman-generate-bootloaders
 
 Requires: grub2-tools-extra
+%if 0%{?rhel} == 0
```
ekohl, line 14: Should this be:
```suggestion
%if 0%{?rhel} < 9
```

On Fedora it's still available and there it would evaluate to `0 < 9` which is true. So technically correct, even if it's irrelevant for us.

ekohl (changes requested):
We'd also need to change the script itself to stop generating it for ia32:
https://github.com/theforeman/foreman-packaging/blob/acccb9aa07cd459547bed0fac802838bcbfb0f3f/packages/foreman/foreman-bootloaders-redhat/foreman-generate-bootloaders#L43-L57

Looking at that, I think it's also never been correct for non-x64.

I can take a stab at it.

packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
evgeni: Surprisingly yes

ekohl: Answering my own question: yes it is.

packages/foreman/foreman-bootloaders-redhat/foreman-bootloaders-redhat.spec
```diff
@@ -11,7 +11,9 @@ BuildArch: noarch
 Source0: foreman-generate-bootloaders
 
 Requires: grub2-tools-extra
```
ekohl, line 13: Further testing shows it wouldn't pull in `grub2-tools` needed for `grub2-mkimge`. I'd suggest
```suggestion
Requires: /usr/bin/grub2-mknetdir
Requires: /usr/bin/grub2-mkimage
```

But then later I found out shim-x64 pulls in grub2-tools.

Still, I like this explicit bit.

ekohl (commented):
I think https://github.com/theforeman/foreman-packaging/pull/10194 implements EL9 compatibility.

evgeni:
obsoleted by #10194

PR #10198, opened 2024-01-04
Rebuild against new foreman-js

evgeni:


evgeni:
@zjhuntin @ekohl this should fix the smoker issues in 3.9

ekohl (approved):

ekohl:
Will this mean 3.9 is newer than nightly?

evgeni:
> Will this mean 3.9 is newer than nightly?

Yeah. We can also rebuild nightly (even tho it's not required there) for consistency.

ekohl:
I'd prefer that and make this a cherry pick.

evgeni:
https://github.com/theforeman/foreman-packaging/pull/10203

PR #10200, opened 2024-01-04
add plugins el9 copr config and comps

evgeni:


comps/comps-foreman-plugins-el9.xml
```diff
@@ -0,0 +1,378 @@
...
+      <packagereq type="default">puppetlabs-stdlib</packagereq>
+      <packagereq type="default">python3.11-ansible-runner</packagereq>
+      <packagereq type="default">python3.11-daemon</packagereq>
+      <packagereq type="default">python3.11-docutils</packagereq>
+      <packagereq type="default">python3.11-lockfile</packagereq>
+      <packagereq type="default">python3.11-packaging</packagereq>
+      <packagereq type="default">python3.11-pbr</packagereq>
+      <packagereq type="default">python3.11-pexpect</packagereq>
+      <packagereq type="default">python3.11-ptyprocess</packagereq>
+      <packagereq type="default">python3.11-pyparsing</packagereq>
```
ekohl, line 31: Will we have `python3.11` or just `python3`?

ekohl (commented):
I'm a bit tempted to start with an empty comps for EL9 and add things as we go along.

evgeni:
> I'm a bit tempted to start with an empty comps for EL9 and add things as we go along.

Why? We didn't do that for core, and I'd expect the list of packages be pretty similar to el8?

comps/comps-foreman-plugins-el9.xml
evgeni: Yes.
(I was trying to avoid this particular edge of this whole thing, as I'd prefer to use system python, but pulpcore might force our hand here)

ekohl:
> Why? We didn't do that for core, and I'd expect the list of packages be pretty similar to el8?

Mostly because I was unsure about the Python stuff.

evgeni:
> > Why? We didn't do that for core, and I'd expect the list of packages be pretty similar to el8?
> 
> Mostly because I was unsure about the Python stuff.

I can leave that one out if you prefer?

ehelms (approved):

PR #10203, opened 2024-01-04
Rebuild against new foreman-js

evgeni:


evgeni:
https://github.com/theforeman/foreman-packaging/pull/10198 was acked and is the same commit, so I am going to assume this is acked too

PR #10205, opened 2024-01-04
only configure python 3.11 for plugins not core

evgeni:
this way we can still build foreman_maintain, websockify and other core bits with system python


Odilhao (approved):

ekohl:
This should be picked to 3.9 too, right?

evgeni:
> This should be picked to 3.9 too, right?

Good point, yes!

Can you do that please? I'm away from Yubikey

ekohl:
https://github.com/theforeman/foreman-packaging/pull/10208

PR #10206, opened 2024-01-04
Refactor spec file bumping to rely on rpmdev-bumpspec

ekohl:
This rewrites the custom logic to deal with versioning to rely on rpmdev-bumpspec. It also adds basic bumping to non-gem/npm packages.

### DIFF
--- a/bump_rpm.sh
+++ b/bump_rpm.sh
@@ -29,6 +29,19 @@ ensure_program() {
 	fi
 }
 
+bump_spec() {
+	SPEC_FILE=$1
+	NEW_VERSION=$2
+
+	spectool --list-files $SPEC_FILE | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git rm
+
+	rpmdev-bumpspec --comment "- Update to ${NEW_VERSION}" --new "${NEW_VERSION}" $SPEC_FILE
+	git add $SPEC_FILE
+
+	spectool --get-files $SPEC_FILE
+	spectool --list-files $SPEC_FILE | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git annex add
+}
+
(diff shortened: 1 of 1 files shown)


bump_rpm.sh
```diff
@@ -62,23 +75,7 @@ if [[ $CURRENT_VERSION != "$NEW_VERSION" ]] ; then
 		cd "$1"
 		GEM_NAME=$(awk '/^%global\s+gem_name/ { print $3 }' $SPEC_FILE)
 
-		spectool --list-files $SPEC_FILE | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git rm
```
evgeni, line 65: won't this leave the old gem in the annex?

bump_rpm.sh
```diff
@@ -62,23 +75,7 @@ if [[ $CURRENT_VERSION != "$NEW_VERSION" ]] ; then
...
-			echo "* Resetting release ($RELEASE) in $SPEC_FILE"
-			sed -i "s/^\(Release:\s\+\)${RELEASE}/\11/" $SPEC_FILE
-		fi
-
-		"$SCRIPT_DIR"/add_changelog.sh $SPEC_FILE <<-EOF
-		- Update to $NEW_VERSION
-		EOF
-
-		spectool --get-files $SPEC_FILE
-		spectool --list-files $SPEC_FILE | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git annex add
```
evgeni, line 80: and won't this not add the new source to the annex?

bump_rpm.sh
ekohl: I've moved those to bump_spec, a custom function that wraps rpmdev-bumpspec

evgeni: Ah, mhh, must have been blind (or not caffeinated).

evgeni (approved):

PR #10207, opened 2024-01-04
Update puppet-agent-puppet-strings to 4.1.2 & its dependencies

ekohl:


evgeni (approved):

packages/foreman/puppet-agent-rgen/puppet-agent-rgen.spec
```diff
@@ -3,8 +3,8 @@
 
 Summary: OAuth Core Ruby implementation for Puppet Agent
 Name: puppet-agent-%{gem_name}
-Version: 0.8.2
-Release: 2%{?dist}
+Version: 0.9.0
```
ekohl, line 6: I chose 0.9.0 because 0.10.0 & 0.10.1 are missing from Rubygems (but tagged in git). Also. 0.9.1 is on Rubygems but not in git so I didn't want to use that either. Filed in https://github.com/mthiede/rgen/issues/38.

It does mean we miss https://github.com/mthiede/rgen/commit/9b7f42d3eca69ac862a3b825bb60184ddfcda950 which may be needed for Puppet 8 where Ruby 3.2 will raise deprecation warnings.

evgeni: Aye. Makes sense

evgeni:
Closure fails because the updated deps are also in this PR but closure looks at each individually (as they land in different scratch repos)

PR #10211, opened 2024-01-05
install rpmdev-bumpspec for bumping packages

evgeni:


ekohl:
I checked that `rpmdev-bumpspec` was installed and I thought it was, because spectool is in the same package. Didn't consider this odd case where we download the scripts manually.

PR #10220, opened 2024-01-05
build proxy plugins on el9

evgeni:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -205,6 +205,12 @@ plugin_packages:
         - el8-extras
         - el8-puppet-7
         - "el8-foreman-{{ foreman_version }}-staging"
+      el9:
+        - el9-baseos
+        - el9-appstream
+        - el9-configmanagement-salt
+        - el9-puppet-7
+        - "el9-foreman-{{ foreman_version }}-staging"
   children:
     foreman_plugin_packages: {}
     foreman_proxy_plugin_packages: {}
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -164,6 +164,8 @@ foreman_packages:
     repoclosure_target_repos:
       el8:
         - "el8-foreman-{{ foreman_version }}-staging"
+      el9:
+        - "el9-foreman-{{ foreman_version }}-staging"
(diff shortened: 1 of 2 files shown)


evgeni:
(draft, I am still testing whether things actually build)

evgeni:
- smart_proxy_container_gateway needs https://github.com/Katello/smart_proxy_container_gateway/pull/30

package_manifest.yaml
```diff
@@ -192,6 +194,8 @@ plugin_packages:
     repoclosure_target_repos:
       el8:
         - "el8-foreman-plugins-{{ foreman_version }}-staging"
+      el9:
+        - "el9-foreman-plugins-{{ foreman_version }}-staging"
     repoclosure_lookaside_repos:
```
evgeni, line 199: also need to add shit to el9-looksasides…

ekohl:
> smart_proxy_pulp needs

We already have Ruby 3.0 & 3.1 testing in Jenkins, so I'd argue it's not needed.

evgeni:
> > smart_proxy_pulp needs
> 
> We already have Ruby 3.0 & 3.1 testing in Jenkins, so I'd argue it's not needed.

If you look at the PR, it also adjusts the gemspec to allow 3.0, I can submit that as an own PR though.

package_manifest.yaml
```diff
@@ -870,10 +877,16 @@ ruby_plugin_packages:
...
+  vars:
+    copr_projects:
+      - "{{ hostvars['plugins-copr-with-el9'] }}"
   hosts:
     rubygem-faraday_middleware: {}
+    rubygem-httparty: {}
     rubygem-infoblox: {}
     rubygem-logify: {}
+    rubygem-mqtt: {}
+    rubygem-multi_xml: {}
```
ekohl, line 889: These look correct: they're only mentioned in the plugin comps.

ekohl (approved):

ekohl:
> And there is no Jenkins run in the PR?

Are you confusing it with the `smart_proxy_container_gateway` PR? I was talking about `smart_proxy_pulp`.

PR #10223, opened 2024-01-07
Rpm/develop bulk update hammer plugins

ofedoren:
That's a try for bulk updating some hammer plugins using `./bump_rpm`. Some of the changes I'd expect, but I'm not sure about the others. Mostly about the ones which **still** have scl login in the spec file. Could someone confirm that's not an issue?

This is also compatible with `rpm/3.9`, so I'd like to CP that into the branch.


evgeni:
Oh, we do, those updates are just newer than the last bumper run

evgeni:
@ofedoren the builds fail for those plugins that newly got localizations, as they do not package the `locales` folder (yet)

do you want to fix them up here in the PR, or split the PR into the ones that build and the ones that do not?
(we can also use the bumper to create all the PRs automatically, and then see what fails and what not)

the SCL changes are OK, we should clean it up properly, but it won't block anything for now

ofedoren:
> Interesting, aren't we updating these automatically with the package bumper? seems like an oversight.

I guess we are, but I wanted make sure and have some tracker that these version are pulled into 3.9 as well. It's easier to do it manually in bulk then checking each version and looking for the commits to CP :) As I see, most of the plugins were autoupdated, so this PR is rebased on top of current `rpm/develop` adding only the plugins which need changes for `locale` to be included.

> do you want to fix them up here in the PR

Yeap, I've updated some plugins' spec file to include `locale`.

ofedoren:
@evgeni, actually, now I see that these 3 plugins were auto-updated by the bot, but that PR's need the same changes. So, I'd suggest to close that ones instead?

Since I'd need to go through that merged auto-prs anyway to gather commit hashes for CP, I don't mind any solution.

evgeni:
it runs on a cron on Sunday morning, but can be triggered by anyone (who has merge permissions in this repo) out-of-band too (which is what I did earlier today).

all PRs are merged (also Debian ones ;-)) so feel free to go picking 

and thanks!

ofedoren:
Closing this one then, thanks a lot @evgeni.

PR #10224, opened 2024-01-08
fix insights client annex source

evgeni:


ekohl (approved):
How was this not noticed before? I think I've seen obal fetch the file as needed before (to verify) and if it's not tracked, then it simply passes.

ekohl:
To add to that: 48057a74e9c722163da97965e242b4ab90bd27f6 renamed the file but didn't change the `Source` line. Then 085498ce65d2d0783c5d4ef946bdf4552fda8cef simply bumped.

evgeni:
I think the check in obal doesn't do what it's supposed to do :(

PR #10225, opened 2024-01-08
build more plugin related packages on el9

evgeni:


package_manifest.yaml
```diff
@@ -721,18 +721,10 @@ foreman_plugin_packages:
   children:
     ruby_plugin_packages: {}
   hosts:
-    ansible-collection-theforeman-operations: {}
-    ansiblerole-foreman_scap_client: {}
-    ansiblerole-insights-client: {}
     ansible-runner: {}
-    ansible-collection-theforeman-foreman: {}
-    cjson: {}
     foreman-discovery-image-service: {}
```
ekohl, line 725: This is a weird one. I wonder if the discovery image should be its own category.

package_manifest.yaml
```diff
@@ -721,18 +721,10 @@ foreman_plugin_packages:
   children:
     ruby_plugin_packages: {}
   hosts:
-    ansible-collection-theforeman-operations: {}
-    ansiblerole-foreman_scap_client: {}
-    ansiblerole-insights-client: {}
     ansible-runner: {}
```
ekohl, line 724: Isn't runner also only for the plugin?

packages/plugins/cjson/cjson.spec
```diff
@@ -28,17 +28,15 @@ developing applications that use cJSON.
 
 %build
 %cmake
-make %{?_smp_mflags}
```
ekohl, line 31: Note to self: this reverts 5a4edfdf077e67e68f3995c2d8169011ed064112

packages/plugins/cjson/cjson.spec
```diff
@@ -28,17 +28,15 @@ developing applications that use cJSON.
...
+%cmake_build
 
 %install
-%make_install
+%cmake_install
 rm -f %{buildroot}%{_libdir}/*.{la,a}
 rm -f %{buildroot}%{_libdir}/cmake/cJSON/*.cmake
 
 %check
-%if 0%{?rhel} > 7
```
ekohl, line 39: Note to self: this reverts d67803aaec2ab881c492abb52468c59c06d71510

packages/plugins/cjson/cjson.spec
ekohl, line 1: Any thoughts on syncing in the latest package from Fedora? Right now there's already 1.7.14-5 in EPEL9 with a certain history. While technically this change brings them back in line, I'd pull the latest spec file in, including the whole history.

package_manifest.yaml
evgeni: yes, but I didn't want to build it just yet, so left it out for now

packages/plugins/cjson/cjson.spec
evgeni: as noted in the commit message, yes :)

packages/plugins/cjson/cjson.spec
evgeni: I can do that. Sure.

evgeni: err. we already had 1.7.14-5 from epel9, but with our patches, so pulling that spec in *would* revert the changes, but the version would not increase :(

evgeni: (we could pull more changelog from fedora, but what's the point?)

package_manifest.yaml
evgeni: a clear maybe (I first moved it to the "other" group too, and then was like "nah, this is a special one")

packages/plugins/cjson/cjson.spec
ekohl: I meant to pull in 1.7.15-2 from Fedora. That way, if they update EPEL9 to exactly the version from Fedora we will be in sync.

evgeni: Ah! Yeah, I can do that.

evgeni:
welp, sometimes our linter is too smart.

evgeni:
finally, :green_apple: again

packages/plugins/cjson/cjson.spec
ekohl: I wish GitHub made those more obvious in but GH and multi commits in PRs remains sub-optimal.

ekohl (approved):

PR #10264, opened 2024-01-10
hammer plugin CPs for 3.9

ofedoren:
@evgeni, here you go :) Same plugins except for h-c-f-admin and h-c-f-openscap. They seem to not be available for debs.


evgeni:
[test deb]

seems there is some race condition :/

evgeni:
I've split out a few into https://github.com/theforeman/foreman-packaging/pull/10265, let's see

PR #10272, opened 2024-01-11
Remove %pretrans section of spec

pcreech:
Utilizing %pretrans with bash pulls in a requirement on /bin/sh, which in certain circumstances (install image generation) can be ran in a clean room environment with no /bin/sh available.

### DIFF
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,4 +1,4 @@
-%global release 3
+%global release 4
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


ehelms:
Stepping back from the immediate problem, I believe this was added many moons ago when we combined the katello-installer into the foreman-installer and was there to support converging paradigms. Given the state we are in, this action should no longer be needed as all installations that have undergone upgrade will have performed this migration already.

@evgeni @ekohl Can you see any case where this would still need to exist?

ekohl:
Good question. I wonder if we ever shipped the old katello installer on EL8 but I think it's sage by now to drop it. Will check when I'm not on my phone

ehelms:
> Good question. I wonder if we ever shipped the old katello installer on EL8 but I think it's sage by now to drop it. Will check when I'm not on my phone

We did not ever ship it. This commit is 5-6 years ago.

evgeni:
The commit (https://github.com/theforeman/foreman-packaging/commit/cf7e36afe5f4525779a38ecff095d809a2f144b0) is from 1.21, so it's pretty safe to be dropped by now.

PR #10275, opened 2024-01-12
build amazing_print for bookworm

evgeni:


ekohl:
We can drop it for now and assume the system Puppet works

evgeni:
[test deb]

PR #10278, opened 2024-01-12
Update browserify-sign for CVE-2023-46234

ehelms:
While we are working to upgrade webpack, this should be a simple update that covers updating a dependency in the older version given we are unsure how long webpack update will take.


packages/foreman/nodejs-webpack/nodejs-webpack.spec
```diff
@@ -5,7 +5,7 @@
 
 Name: %{?scl_prefix}nodejs-webpack
 Version: 3.12.0
-Release: 5%{?dist}
+Release: 2%{?dist}
```
evgeni, line 8: this seems to turn back the time? :)

evgeni (approved):

PR #10279, opened 2024-01-12
Update nodejs-node-gyp dependencies

ehelms:


evgeni:
Can be closed, as we dropped node-sass and thus node-gyp?

PR #10290, opened 2024-01-16
build plugin ruby packages on el9

evgeni:


evgeni:
known broken:
- `rubygem-google-protobuf` (it builds, but `require 'google/protobuf'` hangs the Ruby process :shrug:)
- `rubygem-grpc` (`BuildRequires: rubygem-google-protobuf`)

evgeni:
When releasing, I'll avoid all those google/gapic/grpc gems for now and then we'll see how it goes with them in a separate PR.

evgeni:
loading of `google/protobuf` hangs in `pool.add_serialized_file(descriptor_data)` in `lib/google/protobuf/descriptor_pb.rb` and I have no friggin idea why (or how to debug that)

ehelms:
Did you try newer versions of those google/gapi/grpc packages?

evgeni:
> Did you try newer versions of those google/gapi/grpc packages?

Yes, they fail to build totally. But I found a way…

evgeni:
this is now ready, repoclosure will be fixed once this is in and the net-ssh-* things are built

ehelms (approved):

PR #10294, opened 2024-01-17
move katello ruby packages to an own group and build on el9

evgeni:


evgeni:
I have verified they all build and repoclosure passes.

ekohl (approved):

PR #10295, opened 2024-01-17
Update rubygem-foreman_remote_execution to 12.0.4

adamruzicka:
CPs:
- 3.9


packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -1,10 +1,10 @@
 # template: foreman_plugin
 %global gem_name foreman_remote_execution
 %global plugin_name remote_execution
-%global foreman_min_version 3.9
+%global foreman_min_version 3.10
```
evgeni, line 4: you say you want to CP, but this will prevent CPing -- seems odd to bump min Foreman in a .z plugin release?

evgeni (approved):

PR #10321, opened 2024-01-25
Fixes #37106 - add python-setuptools dep to el6 katello-host-tools

ianballou:
Without python-setuptools, katello-enabled-repos-upload fails on EL6.

### DIFF
--- a/packages/client/katello-host-tools/katello-host-tools.spec
+++ b/packages/client/katello-host-tools/katello-host-tools.spec
@@ -37,7 +37,7 @@
 
 Name: katello-host-tools
 Version: 4.2.3
-Release: 4%{?dist}
+Release: 5%{?dist}
 Summary: A set of commands and yum plugins that support a Katello host
 Group:   Development/Languages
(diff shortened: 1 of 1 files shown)


packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -81,6 +81,10 @@ BuildRequires: python2-devel
 %endif
 %endif
 
+%if 0%{?rhel} == 6
+Requires: python-setuptools
```
ekohl, line 85: Wouldn't this also be an issue on EL7? Or are the scripts not used there?

ianballou: I need to test this. My assumption was this would've also been reported for EL7, but then again that katello-enabled-repos-upload script isn't the default any more

ianballou:
I've also added the requirement for EL7 (and EL5 as it happens)

evgeni (approved):

PR #10341, opened 2024-01-26
Rebuild REX with Webpack 5

evgeni:


ekohl (approved):

ekohl:
[test copr-rpm] because it timed out

evgeni:
It won't build, we need to update a few more packages first I'm afraid (going to verify that via debian)

evgeni:
[test rpm-copr]

PR #10344, opened 2024-01-28
correctly find existing katello packages w/o passing base_dir

evgeni:
bump_rpm.sh doesn't pass a base_dir, so let's find the package on our own instead of trying to add a new package to the foreman dir, while it really can be updated in the katello dir


evgeni:
```
sed: can't read /*.spec: No such file or directory
```

:thinking: that didn't work like I've expected it to work…

evgeni:
https://github.com/theforeman/foreman-packaging/pull/10345

PR #10355, opened 2024-01-30
Use the copr_project_user to calculate the Copr repo root

ehelms:
Otherwise you can scare yourself into thinking you changed the main repo when really you modified your fork.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -53,7 +53,7 @@ copr_projects:
     rhel_8: '8'
     rhel_7: '7'
     puppet_baseurl: https://yum.puppet.com/puppet7
-    root_repo_url: https://download.copr.fedorainfracloud.org/results/@theforeman
+    root_repo_url: "https://download.copr.fedorainfracloud.org/results/{{ copr_project_user }}"
     foreman_staging: "{{ root_repo_url }}/foreman-{{ foreman_version }}-staging"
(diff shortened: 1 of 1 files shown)


package_manifest.yaml
```diff
@@ -53,7 +53,7 @@ copr_projects:
     rhel_8: '8'
     rhel_7: '7'
     puppet_baseurl: https://yum.puppet.com/puppet7
-    root_repo_url: https://download.copr.fedorainfracloud.org/results/@theforeman
+    root_repo_url: "https://download.copr.fedorainfracloud.org/results/{{ copr_project_user }}"
```
evgeni, line 56: This implies `copr_project_user` has all the staging repos available.

I was about to say "this will break my scratch flow", but then realized I set `-e copr_scratch_user=evgeni`, which is different from `copr_project_user`.

So I think it's fine, but wanted to highlight it before it gets forgotten :)

evgeni (approved):

package_manifest.yaml
ehelms: Are you saying this should also check and use `copr_scratch_user` if it is set?

evgeni: Nah, only that those two variables are close naming wise, but mean different things.

I explicitly do not want the scratch_user repos to be used.

PR #10356, opened 2024-01-30
Set copr as the default build system

ehelms:


ehelms:
Eventually we should update the default in Obal, however, I would wait until Koji is nearly sunset to do that.

Odilhao (approved):

Odilhao:
We can also cherrypick to `rpm/3.9`

PR #10359, opened 2024-01-31
Use nodejs-sass instead of nodejs-node-sass

evgeni:


evgeni:
needs #10358

evgeni:
[test rpm-copr]

PR #10380, opened 2024-01-31
Remove cssnano and css-minimizer-webpack-plugin

ehelms:


evgeni:
dropping it from foreman.spec in https://github.com/theforeman/foreman-packaging/pull/10382

evgeni (approved):

PR #10387, opened 2024-02-01
Copy gitconfig into build container

ehelms:
This makes running update commands inside the container work.

### DIFF
--- a/enter-container
+++ b/enter-container
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-podman run --rm --privileged -it -v ./:/opt/foreman-packaging foreman-packaging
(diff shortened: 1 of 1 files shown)


ekohl (commented):
This doesn't handle `~/.config/git/config`. Would this also work if the file doesn't exist?

ehelms:
> This doesn't handle `~/.config/git/config`. Would this also work if the file doesn't exist?

Are you asking if we could use that path instead? or also mount it?

ekohl:
Mostly wondering what happens if you attempt to mount a location that doesn't exist

ehelms:
> Mostly wondering what happens if you attempt to mount a location that doesn't exist

It will yell at you:

```
Error: statfs /home/ehelms/.config/git/config: no such file or directory
```

PR #10389, opened 2024-02-01
Use --legacy-peer-deps in bundled NPM packages

ehelms:


evgeni:
What worries me a bit is that https://docs.npmjs.com/cli/v6/using-npm/config doesn't mention legacy-peer-deps at all.
Later versions do, but v6 doesn't (which is technically correct, for 6 those are not legacy).
But as long it works, I guess it's our best bet.

evgeni (approved):
It builds fine on EL8 with NodeJS 14/NPM 6 and EL9 with NodeJS 18 and NPM 9 and allowed me to build Foreman on EL9

ekohl (approved):
This is less bad than I thought. I suspect npm 6 may ignore unknown flags

PR #10392, opened 2024-02-02
build nodejs packages using node 18 on el9

evgeni:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -40,10 +40,12 @@ all:
 
 copr_projects:
   vars:
-    core_modules:
+    core_modules_el8:
       - 'ruby:2.7'
       - 'nodejs:14'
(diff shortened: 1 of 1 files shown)


ekohl (approved):
Does this mean we need a foreman-devel:el9 as well?

evgeni:
> Does this mean we need a foreman-devel:el9 as well?

technically, yeah. but that's unrelated to this PR (we only really need devel for packit and friends), so I was slacking off on this topic

ekohl:
Which is why I approved. Just something that came up in the review

ehelms:
> > Does this mean we need a foreman-devel:el9 as well?
> 
> technically, yeah. but that's unrelated to this PR (we only really need devel for packit and friends), so I was slacking off on this topic

What is meant by this question?

PR #10404, opened 2024-02-08
move all cockpit related files to cockpit subpackage

evgeni:

### DIFF
--- a/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
+++ b/packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
@@ -5,7 +5,7 @@
 
 Name: rubygem-%{gem_name}
 Version: 12.0.5
-Release: 2%{?foremandist}%{?dist}
+Release: 3%{?foremandist}%{?dist}
 Summary: A plugin bringing remote execution to the Foreman, completing the config management functionality with remote management functionality
 License: GPLv3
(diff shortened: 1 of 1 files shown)


evgeni:
this fixes https://github.com/theforeman/foreman-selinux/pull/163#issuecomment-1934018722

packages/plugins/rubygem-foreman_remote_execution/rubygem-foreman_remote_execution.spec
```diff
@@ -129,11 +128,15 @@ install -Dp -m0644 %{buildroot}%{gem_instdir}/extra/cockpit/settings.yml.example
 %config(noreplace) %{_sysconfdir}/foreman/cockpit/cockpit.conf
 %config(noreplace) %{_sysconfdir}/foreman/cockpit/foreman-cockpit-session.yml
 %{_unitdir}/foreman-cockpit.service
+%{gem_instdir}/extra/cockpit
```
ekohl, line 131: Why do we need the extra directory? All files are copied, except `foreman-cockpit-session` which is a symlink. Perhaps we should simply make line 84 an `install` and `%exclude %{gem_instdir}/extra` instead.

evgeni: maybe. I was not planing to re-think the packaging. but I can :)

evgeni: well, let's see how that goes

PR #10406, opened 2024-02-09
Remove nodejs-downshift

ehelms:

PR #10407, opened 2024-02-09
Katello el9 nodejs

ehelms:
Includes https://github.com/theforeman/foreman-packaging/pull/10406 which is the only bundled package but can be dropped

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -619,14 +619,9 @@ katello_packages:
   children:
     satellite_packages: {}
     katello_common_packages: {}
+    katello_nodejs_packages: {}
     katello_ruby_packages: {}
   hosts:
-    nodejs-angular: {}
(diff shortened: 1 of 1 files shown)


evgeni:
rebased, github is stupid sometimes.

package_manifest.yaml
```diff
@@ -592,7 +592,7 @@ foreman_client_packages:
 katello_packages:
   vars:
     copr_projects:
-      - "{{ hostvars['katello-copr'] }}"
+      - "{{ hostvars['katello-copr-with-el9'] }}"
```
evgeni, line 595: mh, this will build *all* of katello on el9, which we do not want yet (katello itself does not define ruby 3 compat yet)?

ehelms: That was a leftover from doing some testing builds of the other packages.

evgeni (approved):

PR #10409, opened 2024-02-12
split katello hammer packages and build on el9

evgeni:


evgeni:
those are actually already built by @ehelms, this PR just corrects the config

ekohl (approved):

PR #10410, opened 2024-02-12
build all of Katello on EL9

evgeni:

### DIFF
--- a/packages/katello/rubygem-katello/rubygem-katello.spec
+++ b/packages/katello/rubygem-katello/rubygem-katello.spec
@@ -6,7 +6,7 @@
 %global prereleasesource pre.master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
 %global mainver 4.12.0
-%global release 2
+%global release 3
 
 Name: rubygem-%{gem_name}
 Version: %{mainver}
@@ -24,10 +24,10 @@ Requires: (katello-selinux if selinux-policy-targeted)
 Requires: foreman >= %{foreman_min_version}
 BuildRequires: foreman-assets >= %{foreman_min_version}
 BuildRequires: foreman-plugin >= %{foreman_min_version}
-Requires: ruby >= 2.5
-Requires: ruby < 3
-BuildRequires: ruby >= 2.5
-BuildRequires: ruby < 3
+Requires: ruby >= 2.7
+Requires: ruby < 3.1
+BuildRequires: ruby >= 2.7
+BuildRequires: ruby < 3.1
 BuildRequires: rubygems-devel > 1.3.1
 BuildRequires: rubygem(rails)
 BuildRequires: rubygem(json)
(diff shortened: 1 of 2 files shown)


packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -170,6 +172,9 @@ done
 %{foreman_plugin_log}
 
 %changelog
+* Mon Feb 12 2024 Evgeni Golov - 4.12.0-0.3.pre.master
+- Correct (Build)Requiremens for EL9
```
ekohl, line 176:
```suggestion
- Correct (Build)Requirements for EL9
```

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -45,6 +45,8 @@ BuildRequires: rubygem(apipie-rails) >= 0.5.14
 BuildRequires: rubygem(fx) < 1.0
 BuildRequires: rubygem(pg)
 BuildRequires: rubygem(anemone)
+BuildRequires: rubygem(webrick)
+Requires: rubygem(webrick)
```
ekohl, line 49: Since this is only to patch anemone, I'd prefer to do it there. Also, this section will get overwritten by our auto-update so that would revert the changes. So if we keep it, it should be outside of the generated section.

evgeni: This was "intentional" so that the changes from https://github.com/Katello/katello/pull/10885 fix it "correctly" (that PR adds `webrick` as otherwise CI fails with the same error)

evgeni: Opened https://github.com/theforeman/foreman-packaging/pull/10411 to fix anemone directly

evgeni:
Repoclosure is related, needs https://github.com/theforeman/foreman-packaging/pull/10412 which needs Candlepin on EL9

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -25,9 +25,9 @@ Requires: foreman >= %{foreman_min_version}
 BuildRequires: foreman-assets >= %{foreman_min_version}
 BuildRequires: foreman-plugin >= %{foreman_min_version}
 Requires: ruby >= 2.5
-Requires: ruby < 3
+Requires: ruby < 3.1
 BuildRequires: ruby >= 2.5
-BuildRequires: ruby < 3
+BuildRequires: ruby < 3.1
```
ekohl, line 30: I suspect we'll raise this to >= 2.7 and < 4.

evgeni: Probably. My goal was "update it to something so it builds, and then when the actual Ruby3 PR is in, we regenerate the deps anyway"

evgeni:
Should be merged together with https://github.com/theforeman/foreman-packaging/pull/10412 as that contains `katello` which `requires: rubygem-katello`

ekohl (approved):
Will merging this without support in katello master also break the nightly package build?

At least 👍 for the changes once the core PR is merged

evgeni:
> Will merging this without support in katello master also break the nightly package build?
> 
> At least 👍 for the changes once the core PR is merged

That's the hackish beauty of it: As you can see in this PR's test results, things build on EL9 with Ruby 3 and NodeJS 18 just fine.
Several parts of the app are broken (and that's what https://github.com/Katello/katello/pull/10885 is fixing), but it's sufficiently OK to build and produce RPM that can then be tested.
So no, merging this will not break anything, but allow us to produce nightlies that can then be further tested.

PR #10412, opened 2024-02-12
bring back katello and katello-selinux packages

evgeni:

### DIFF
--- a/comps/comps-katello-el9.xml
+++ b/comps/comps-katello-el9.xml
@@ -12,11 +12,11 @@
     <packagelist>
       <!-- katello core packages -->
       <packagereq type="default">foreman-proxy-content</packagereq>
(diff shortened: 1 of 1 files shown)


evgeni:
@ehelms this requires `candlepin-selinux` and `candlepin` for EL9. For that we'd either need to switch to nightly CP repo, or rebuild the last 4.3 release for EL9. I don't mind either way (and maybe the 4.3.12 in nightly should move to 4.3, which ships 4.3.1 today?), but something needs to happen :)

evgeni:
4.3 now carries el9, so https://github.com/theforeman/foreman-packaging/pull/10414 enables it. once in, this can go in.

ehelms (approved):

ekohl (approved):

PR #10428, opened 2024-02-15
Adjust requirements for Ansible using Python 3.12

evgeni:


evgeni:
https://kojihub.stream.centos.org/koji/buildinfo?buildID=50584 in EL8 is now using 3.12

Odilhao (approved):
Why Python 3.12, anyway approved

evgeni:
> Why Python 3.12, anyway approved

Don't ask.

PR #10442, opened 2024-02-16
Update rubygem-hammer_cli_katello to 1.11.2

chris1984:


chris1984:
@pcreech can this get a merge?

ianballou (commented):
Just want some confirmation that the SCL related deletions are expected.

chris1984:
> Just want some confirmation that the SCL related deletions are expected.

I just ran the bump rpm script, let's see if that was intended it passed green so that is a plus. Let's see what Patrick says, worst case I can go back and put those how they were.

chris1984:
> > Just want some confirmation that the SCL related deletions are expected.
> 
> We are indeed slowly dropping those in spec files, but I'd leave them here just for consistency. Or drop them in the whole file, but I wouldn't do that in a patch release.

Sounds good, will push up a change to make the spec file change just the version bump

chris1984:
@ekohl updated

ekohl (approved):

PR #10451, opened 2024-02-20
Bump version to 3.11-develop

pcreech:


evgeni:
This was merged with red builds as foreman 3.11 source doesn't exist yet.

PR #10454, opened 2024-02-20
Foreman 3.10 branching updates - DEB edition

pcreech:


ekohl (commented):
Did we generate a new GPG key for Debian? We normally don't do that for every release. AFAIK with Debian we have a single GPG key for deb.theforeman.org.

ekohl (approved):
They match https://deb.theforeman.org/foreman.asc so perhaps we just have it wrong in nightly?

evgeni:
I think both are fine, they are just exports at different times :woman_shrugging:

pcreech:
For the simplicity of not re-hashing this conversation next branching, I'd say we merge it, so it'll become a no-op in the future.

PR #10457, opened 2024-02-21
Bump hammer version to 3.11-develop

ofedoren:
Not sure if I supposed to do that, but I guess it should be like that. Please, correct me or close it if I'm wrong.


evgeni:
https://github.com/theforeman/foreman-packaging/commit/61ae2d4ccd8d1eaac6c5d765af31edae9c6eb0ea already did this :)

PR #10512, opened 2024-02-29
build foreman on jammy

evgeni:


evgeni:
Ah yes, this actually uses `npm` and that hates us for not supporting Node 18. Luckily we have https://github.com/theforeman/foreman/pull/10016 which I just merged.

evgeni:
ok, this builds. draft as I want the node version hack to be not needed anymore

PR #10513, opened 2024-02-29
Use Python 3 for package.json filtering

evgeni:
Also drop old Python dev dependencies now that we don't vendor wesockify anymore

### DIFF
--- a/debian/bullseye/foreman/rules
+++ b/debian/bullseye/foreman/rules
@@ -34,7 +34,7 @@ build:
 	/bin/sed -i '2i  gem "puma-status"' bundler.d/service.rb
 	$(BUNDLE) install
 	$(BUNDLE) package --no-install
-	/usr/bin/python script/filter-package-json.py
+	/usr/bin/python3 script/filter-package-json.py
 	/usr/bin/npm install --no-audit --no-optional --unsafe-perm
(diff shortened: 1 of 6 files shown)


debian/bullseye/foreman/rules
```diff
@@ -34,7 +34,7 @@ build:
 	/bin/sed -i '2i  gem "puma-status"' bundler.d/service.rb
 	$(BUNDLE) install
 	$(BUNDLE) package --no-install
-	/usr/bin/python script/filter-package-json.py
+	/usr/bin/python3 script/filter-package-json.py
```
ekohl, line 37: It has a shebang. Should we rely on that? And perhaps long term also rewrite it to Ruby so we can drop Python as a build dependency?

ekohl (approved):
I wondered why we need it. If it's just a single script, perhaps we can rewrite it to Ruby.

debian/bullseye/foreman/rules
evgeni: Ah, that's a fair point, I think the explicit `/usr/bin/python` was to *override* the shebang… Let's try that.

evgeni: Heh. The file is not executable. So won't work just now. I'll leave it with the explicit python3 and we'll revisit once we rewrite.

PR #10514, opened 2024-02-29
Reimplement repoquery in a more generic way

ekohl:
This approach is based on rpmdist-repoquery where a certain directory structure is used to track repositories. That is then selected, together with a releasever.

Our own repositories are made generic by using DNF_VAR_$var, which makes branching easier.

It also uses true modularity. For that it hacks together a temporary installroot and uses fakeroot to pretend to be root, because dnf insists you need to be root to enable a module. Due to only using real DNF, we can rely on the modulary metadata and dependency resolution to select the correct modules. Again, reducing duplication.

A major motivation for this is the upcoming CentOS Stream 8 EOL where the repositories will be archived. This new repository structure allows easy switching to another Enterprise Linux for repoclosure checks.

This needs a different approach in how we configure repoclosure in obal and the wrapper scripts should probably live in another repository as well.  Another excercise is the EL7 repoclosure check.


dnfwrapper
ekohl, line 1: This is essentially the same as `rpmdistro-repoquery` but without the `repoquery` in the last line. We can talk to the maintainers to provide this. While talking about this I was pointed to [fedrq](https://pypi.org/project/fedrq) as well, which may be nicer.

repoclosure
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+DNF_VAR_foremanver=nightly
+DNF_VAR_katellover=nightly
+DNF_VAR_candlepinver=nightly
+DNF_VAR_puppetver=7
```
ekohl, line 6: It would be nice if these were provided by the package manifest.

ekohl:
Rebased to resolve conflicts. I've also opened https://github.com/theforeman/foreman-packaging/pull/10869 for the first commit.

distros/centos-stream-legacy/centos-stream-legacy.repo
```diff
@@ -0,0 +1 @@
+/usr/share/rpmdistro-repoquery/distros/centos-stream-legacy/centos-stream-legacy.repo
```
evgeni, line 1: This file comes from `rpmdistro-repoquery`, right?

*checks https://src.fedoraproject.org/rpms/rpmdistro-repoquery*

So EPEL9 will be required on the systems, OK.

distros/centos-stream/foreman-staging.repo
```diff
@@ -0,0 +1,33 @@
+[foreman-$foremanver-staging]
+name=Foreman $foremanver staging EL$releasever
+baseurl=https://stagingyum.theforeman.org/foreman/$foremanver/el$releasever/$basearch/
+module_hotfixes=1
```
evgeni, line 4: is this required given you actually enable modules?

repoclosure
```diff
@@ -0,0 +1,30 @@
...
+		fakeroot ./dnfwrapper "$distro" "$releasever" repoclosure --installroot "$INSTALLROOT" "${@:4}"
+
+		rm -rf "$INSTALLROOT"
+	else
+		./dnfwrapper "$distro" "$releasever" repoclosure "${@:4}"
+	fi
+}
+
+repoclosure centos-stream 9 "" --check foreman-${DNF_VAR_foremanver}-staging --enablerepo puppet7
+repoclosure centos-stream-legacy 8 foreman-devel:el8 --check foreman-${DNF_VAR_foremanver}-staging --enablerepo puppet7
```
evgeni, line 30: this hard-codes closure for foreman.
how would I call it for katello?
how for the case where we add a repo on the CLI to perform closure on a copr build?

evgeni:
My biggest concern here is that this lives in the wrong place.

Today we have three places that do repoclosure: foreman-packaging, pulpcore-packaging, candlepin-packaging.
They all use the same pattern: `repoclosure/yum.conf` defines the repos available, `package_manifest.yml` the repos used for each individual target and `obal` has a playbook that stitches that all together to execute it.

We certainly shouldn't have to have copies of `dnfwrapper` and `repoclosure` in each of those to make it work.

As for `fedrq`: it more reads as a "different" repoquery, not repoclosure, which is another beast in itself, imho

distros/centos-stream-legacy/centos-stream-legacy.repo
ekohl: Yes. My idea was that we would push off the actual repo configs to something common, rather than maintaining it ourselves. We can also choose to copy them of course.

distros/centos-stream/foreman-staging.repo
ekohl: Maybe not, probably a leftover from earlier testing.

repoclosure
ekohl: This is the area where I ended the day of learning. I imagine the next step would be to something work this into obal so we can use it for our `*-packaging` repositories. Ideally set the variables from `package_manifest.yaml`.

Still, I think this script gives a good start to see the general idea.

ekohl:
As I wrote in https://github.com/theforeman/foreman-packaging/pull/10514#discussion_r1624217062, I agree. My ideal is that in `package_manifest.yaml` we can specify the operating systems to run on (AlmaLinux 8, 9, CentOS Stream 9) at the top level and repositories to enable (foreman-staging, foreman-plugins-staging, katello, puppet, etc) for each package group

But then we also have COPR that may replace parts, and my brain was full.

ekohl:
Notes to self: jenkins jobs invokes obal in 2 ways.

First of all, there is repoclosure in RPM pipelines:
```groovy
obal(
    action: 'repoclosure',
    packages: "${repo}-repoclosure-${dist}"
)
```

These correspond to:
https://github.com/theforeman/foreman-packaging/blob/85637cd4402e0f45c453fa2abcf7a03c1701d211/package_manifest.yaml#L905-L1019

And then for individual packages when they've been built:
```groovy
obal(
    action: "repoclosure",
    packages: package_name,
    extraVars: [
        'repoclosure_check_repos': [repo['url']],
        'repoclosure_target_dist': repo['dist']
    ]
)
```

These correspond to (among others):
https://github.com/theforeman/foreman-packaging/blob/85637cd4402e0f45c453fa2abcf7a03c1701d211/package_manifest.yaml#L1027-L1050

ekohl:
Let's first start with deduplicating the configs: https://github.com/theforeman/foreman-packaging/pull/10881 reuses settings in the manifest.

ehelms:
My first impression is we make `dnfwrapper` a helper function within Obal's module_utils, and then we adapt the repoclosure module logic to do what the `repoclosure` script in here is doing. And we'll need some way to intelligently decide if we do it the old or the new way.

ekohl:
I was starting to lean to the same thing. Some further notes for myself below.

Essentially one use case calls it with:
```yaml
repoclosure_target_repos:
  el8:
    - "foreman-{{ foreman_version }}-staging"
repoclosure_lookaside_repos:
  el8:
    - $OS_REPOS
    - puppet-7
repoclosure_target_dist: el8
```

Then the other sets `repoclosure_target_dist` and adds `repoclosure_check_repos`:
```
repoclosure_target_repos:
  el8:
    - "foreman-{{ foreman_version }}-staging"
repoclosure_lookaside_repos:
  el8:
    - $OS_REPOS
    - puppet-7
repoclosure_target_dist: el8
repoclosure_check_repos:
  - $url
```

The design I'm starting to lean to (and that's where https://github.com/theforeman/foreman-packaging/pull/10881 tries to advance to) is to always refer to some "`host`" for indirection. Each group can have a basename which gets appended with a `repoclosure_target_dist`. Then for each package we only need to add `repoclosure_check_repos`.

This may require some redesign in obal itself, but I think it'll reduce duplication.

ekohl:
Further notes to self: from https://github.com/theforeman/foreman-packaging/pull/10918#issuecomment-2165081943

We apply repoclosure 1-by-1. I think we should consider an approach where we group things.

PR #10534, opened 2024-03-07
Bump rubygem-smart_proxy_discovery to 1.0.5-10

nofaralfasi:

### DIFF
--- a/packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
+++ b/packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
@@ -1,44 +1,29 @@
 # template: smart_proxy_plugin
-%{?scl:%scl_package rubygem-%{gem_name}}
-%{!?scl:%global pkg_name %{name}}
-
-%{!?_root_datadir:%global _root_datadir %{_datadir}}
-%{!?_root_localstatedir:%global _root_localstatedir %{_localstatedir}}
-%{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
-
 %global gem_name smart_proxy_discovery
 %global plugin_name discovery
 
-%global foreman_proxy_min_version 1.25
-%global foreman_proxy_dir %{_root_datadir}/foreman-proxy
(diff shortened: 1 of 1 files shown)


packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
```diff
@@ -109,6 +81,9 @@ mv %{buildroot}%{gem_instdir}/settings.d/discovery.yml.example \
 %doc %{gem_instdir}/README.md
 
 %changelog
+* Thu Mar 07 2024 nofaralfasi <nalfassi@redhat.com> 1.0.5-1
+- Update to 1.0.5-1
```
nofaralfasi, line 85: @evgeni  Any idea why it didn't jump to `1.0.5-10`?

nofaralfasi: And what about line 5? `%global foreman_proxy_min_version 1.24` is correct? (used to be 1.25)

evgeni: That comes from https://github.com/theforeman/foreman-packaging/blob/7511869cc09cf038e6aaa503a08b83412c32f46a/gem2rpm/smart_proxy_plugin.spec.erb#L15, no idea which version your plugin needs :)

packages/plugins/rubygem-smart_proxy_discovery/rubygem-smart_proxy_discovery.spec
```diff
@@ -109,6 +81,9 @@ mv %{buildroot}%{gem_instdir}/settings.d/discovery.yml.example \
 %doc %{gem_instdir}/README.md
 
 %changelog
+* Thu Mar 07 2024 nofaralfasi <nalfassi@redhat.com> 1.0.5-10
+- Update to 1.0.5-10
```
evgeni, line 85:
```suggestion
- Regenerate spec file based on the latest template
```

(or something along these lines)

same for the commit message, then we're good to go :)

evgeni (approved):

PR #10553, opened 2024-03-12
Fix client specs for suse

m-bucher:
Fix some client spec-file problems on SUSE 15 build.
Also OpenSUSE Build Service checks some stuff like `License` string to be compatible with https://spdx.org/licenses/

### DIFF
--- a/packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
+++ b/packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
@@ -15,14 +15,18 @@
         go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -compressdwarf=false -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v %{?**};
 %else
 %if ! 0%{?gobuild:1}
+%if 0%{?suse_version}
+%define gobuild(o:) go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -a -v %{?**};
+%else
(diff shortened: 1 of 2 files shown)


packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -47,9 +47,13 @@ Worker service for yggdrasil that can act as pull client for Foreman Remote Exec
 mkdir -p _gopath/src
 ln -fs $(pwd)/src _gopath/src/%{name}-%{version}
 ln -fs $(pwd)/vendor _gopath/src/%{name}-%{version}/vendor
+ln -fs $(pwd)/go.mod _gopath/src/%{name}-%{version}/go.mod
 export GOPATH=$(pwd)/_gopath
 pushd _gopath/src/%{name}-%{version}
 %{gobuild}
+%if 0%{?rhel} > 7 || 0%{?fedora} || 0%{?suse_version}
+mv ygg_worker %{name}-%{version}
+%endif
```
m-bucher, line 56: Not sure about this one. I assume this has to do with the different contents of `%{gobuild}`, resulting in differently named binaries :thinking:

packages/client/yggdrasil/yggdrasil.spec
```diff
@@ -2,9 +2,13 @@
 
 Name:    yggdrasil
 Version: 0.2.3
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: Message dispatch agent for cloud-connected systems
+%if 0%{?suse_version}
```
ekohl, line 7: For what it's worth, Red Hat is also moving to SPDX tags. We haven't started the bulk migration, but see https://communityblog.fedoraproject.org/important-changes-to-software-license-information-in-fedora-packages-spdx-and-more/ for more. https://docs.fedoraproject.org/en-US/legal/update-existing-packages/ is the official Fedora way, but RHEL is planning the same. I just don't know for sure where that migration currently is. IIRC we've been asked internally to start changing.

It would be my preference to keep things simple. Can we change all the license tags for the client packages at once?

ekohl:
Thanks @evgeni for finishing this.

PR #10556, opened 2024-03-12
Fixes #37246 - Add back ipmitool dependency

ekohl:
During a rewrite of the spec file this was dropped.

Fixes: 6cc3c98edd01 ("Update rubygem-rubyipmi to 0.11.1")


ekohl:
Related: it looks like it was never present in Debian packages.

evgeni (approved):

PR #10601, opened 2024-03-19
Regenerate hammer_cli_foreman_kubevirt.spec file

girijaasoni:


ekohl (changes requested):
Somehow you lost the changelog and that broke the changelog check.

girijaasoni:
> Somehow you lost the changelog and that broke the changelog check.

@ekohl , I've used the command `./add_gem_package.sh hammer_cli_foreman_kubevirt hammer_plugin plugins` to regenerate this file 
Am I missing something? 
I think it could be due to gem gem2rpm. I had to manually install it , is there a specific compatible version?

evgeni:
obsoleted by Leos' PR

PR #10614, opened 2024-03-23
Only require shim-ia32 bootloader on RHEL 8 or less

ehelms:


ehelms:
[test rpm-copr]

evgeni (approved):

ekohl:
Why is this needed? We could build it with EL9 before

ehelms:
For some reason an EL8 version of this package was put into Centos Stream 9 even though RHEL no longer boots on 32-bit UEFI.

evgeni (approved):

evgeni:
> I also do not think anything requires and installs this package in Foreman ecosystem directly? Which means the requires probably haven't been put to the test.

Well, repoclosure catches things, but otherwise I agree

PR #10677, opened 2024-04-09
Fixes #37334 - Drop keycloak-httpd-client-install from EL9

ekohl:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -55,6 +55,17 @@ copr_projects:
             - "{{ foreman_staging }}/rhel-{{ rhel_9 }}-x86_64"
           comps_file: "{{ inventory_dir }}/comps/comps-foreman-el{{ rhel_9 }}.xml"
           buildroot_packages: "{{ core_buildroot_packages }}"
+    foreman-copr-el8:
+      copr_project_name: "foreman-{{ foreman_version }}-staging"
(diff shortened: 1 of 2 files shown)


package_manifest.yaml
```diff
@@ -55,6 +55,17 @@ copr_projects:
...
+      copr_project_name: "foreman-{{ foreman_version }}-staging"
+      copr_project_fork_from: "{{ 'foreman-nightly-staging' if foreman_version != 'nightly' else False }}"
+      copr_project_chroots:
+        - name: "rhel-{{ rhel_8 }}-x86_64"
+          modules: "{{ core_modules_el8 }}"
+          external_repos:
+            - "{{ puppet_baseurl }}/el/{{ rhel_8 }}/x86_64/"
+            - "{{ foreman_staging }}/rhel-{{ rhel_8 }}-x86_64"
+          comps_file: "{{ inventory_dir }}/comps/comps-foreman-el{{ rhel_8 }}.xml"
+          buildroot_packages: "{{ core_buildroot_packages }}"
```
evgeni, line 68: is that really necessary? can't we just not add it to comps?
having multiple copr definitions is a source for errors ;)

ekohl: I debated both. This way it only needs to build on EL8, not on EL9 so you won't get false positives. On the other hand, it's not a regularly changing package.

I've opened https://github.com/theforeman/foreman-packaging/pull/10719 as the alternative so you can decide what you prefer.

evgeni:
the alternative was merged

PR #10679, opened 2024-04-09
Add rubygem-foreman_cve_scanner package

sbernhard:


comps/comps-foreman-plugins-el8.xml
ekohl, line 1: Could you also add it to EL9?

ekohl (changes requested):

ekohl (approved):

PR #10682, opened 2024-04-10
Cherrypick - Bump nodejs-theforeman-vendor/builder to 13.0.1 to 3.9

chris1984:


evgeni:
This is not a clean pick of 2aab37d352d8cc6d753782644944a9c2701cbd47 as that was ontop of c12470f2b4c5ddb8b42b8176a410befb544403f4 which is missing in this PR. Because of that the old 12.2.0 tarball was not properly removed from git and that confuses the hell out of git-annex.

Btw, please use `git cherry-pick -x` so that it is tracked from which commit a cherry pick is made.

evgeni:
I think we also should pick the `nodejs-theforeman-builder` changes from develop, I have no idea what happens if those two get out of sync.

chris1984:
I got the first builder one but when I try to apply your builder one, it fails, looks like it wants https://github.com/theforeman/foreman-packaging/pull/10389/commits/a514951359aaea9d4c445ed0eccc5a485c130e1a but when I apply that one I get more conflicts.

chris1984:
@theforeman/packaging any help here would be appreciated

packages/foreman/nodejs-theforeman-builder/nodejs-theforeman-builder.spec
```diff
@@ -4,42 +4,42 @@
 %global npm_name @theforeman/builder
 
 Name: %{?scl_prefix}nodejs-theforeman-builder
-Version: 12.2.0
+Version: 12.2.3
```
evgeni, line 7: that's the wrong version

chris1984:
Brought in the builder 13.0.1 cherry-pick, hopefully this looks right?

evgeni (approved):

evgeni:
should be merged whenever the next 3.9.z is being prepared (otherwise it'll break plugin building until then).

PR #10683, opened 2024-04-10
Cherrypick - Bump nodejs-theforeman-vendor/builder to 13.0.1 to 3.10

chris1984:


chris1984:
@theforeman/packaging here is the CP for 3.10

evgeni:
I think we also should pick the `nodejs-theforeman-builder` changes from develop, I have no idea what happens if those two get out of sync.

chris1984:
Brought in the builder cherrypick too, scratch builds passed but failing on repoclosure?

chris1984:
@theforeman/packaging any help here would be appreciated

evgeni (approved):
should be merged whenever the next 3.10.z is being prepared (otherwise it'll break plugin building until then).

PR #10719, opened 2024-04-17
Fixes #37334 - Drop keycloak-httpd-client-install from EL9 (comps only)

ekohl:

This is an alternative to https://github.com/theforeman/foreman-packaging/pull/10677 but only drops it from comps, not the COPR build.


evgeni (approved):

evgeni:
Shall we, retrospectively, also remove it from 3.10-el9?

ekohl:
https://github.com/theforeman/foreman-documentation/pull/2970 for the release note.

PR #10720, opened 2024-04-17
Fixes #37334 - Drop keycloak-httpd-client-install from EL9

evgeni:

(cherry picked from commit 4726d83e32b36ffbeb70034be74a77a414f455cb)

### DIFF
--- a/comps/comps-foreman-el9.xml
+++ b/comps/comps-foreman-el9.xml
@@ -40,7 +40,6 @@
       <packagereq type="default">foreman-telemetry</packagereq>
       <packagereq type="default">foreman-vmware</packagereq>
       <packagereq type="default">katello-certs-tools</packagereq>
-      <packagereq type="default">keycloak-httpd-client-install</packagereq>
       <packagereq type="default">libsass</packagereq>
       <packagereq type="default">libsass-devel</packagereq>
       <packagereq type="default">nodejs-argv-parse</packagereq>


sbernhard:
> fixed in RHEL 9.6.
and in EL8 it will still be broken to use newer keycloak version. Am I right?

ekohl:
Likely yes, though we have documented a workaround where you reconfigure the Keycloak server in a compatible way.

Though thanks for bringing this up. We should drop keycloak-httpd-client-install from `rpm/develop` and `rpm/3.13`. If you want to update it for 3.12 and older for EL8 then I think we can consider that.

ekohl:
> Though thanks for bringing this up. We should drop keycloak-httpd-client-install from `rpm/develop` and `rpm/3.13`.

https://github.com/theforeman/foreman-packaging/pull/11478 drops it.

> If you want to update it for 3.12 and older for EL8 then I think we can consider that.

I'll note it makes the documentation very complex. Version 1.3 changes the default prefix, so if that has a mismatch, we'll have very complicated "if on X, do this, if on Y do that". https://docs.theforeman.org/nightly/Configuring_User_Authentication/index-katello.html#prerequisites-for-configuring-foreman-with-keycloak-authentication_keycloak-quarkus already has a workaround for this, but when 1.3 lands we'll need to update this. Possibly add a workaround for the Wildfly-based instructions.

PR #10724, opened 2024-04-19
Update smart_proxy_salt to 5.1.0

nadjaheitmann:

PR #10726, opened 2024-04-19
Update rubygem-smart_proxy_salt to 5.1.0

nadjaheitmann:


sbernhard (approved):
LGTM. Thank you!

ekohl (commented):
While we're updating this, would you mind updating it to fully get rid of EL7 compatibility? So all SCL and Python 2 support.

ekohl:
Oh, I see that's https://github.com/theforeman/foreman-packaging/pull/10724.

PR #10760, opened 2024-05-07
Add libyaml-dev to Build-Depends for psych

evgeni:

### DIFF
--- a/debian/bookworm/foreman/changelog
+++ b/debian/bookworm/foreman/changelog
@@ -1,3 +1,9 @@
+foreman (3.11.0-3) stable; urgency=low
+
+  * Add libyaml-dev to Build-Depends for psych
+
+ -- Evgeni Golov <evgeni@debian.org>  Tue, 07 May 2024 11:41:27 +0200
+
 foreman (3.11.0-2) stable; urgency=low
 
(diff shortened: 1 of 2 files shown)


evgeni:
Needed for https://github.com/theforeman/foreman/commit/fe77720302a6e81ff621ef071486ec123c9f74f7, bookworm is the only release using 3.1

ekohl (approved):
I thought it'd use the already included package :(

evgeni:
> I thought it'd use the already included package :(

You mean `ruby-psych`?

evgeni:
Well, today nothing pulls that in during build.
Also, I am not sure how I feel about mixing vendored and non-vendored things here.

evgeni:
Or do you mean the "default" gem in `libruby3.1`?
`/usr/lib/ruby/gems/3.1.0/specifications/default/psych-4.0.3.gemspec`

:thinking:

evgeni:
I'll merge this to unblock builds, we can still revisit afterwards.

ekohl:
> Or do you mean the "default" gem in `libruby3.1`?

Yes, that's the one I meant. I suspect bundler wants to get the latest & greatest and upgrades it.

PR #10761, opened 2024-05-07
Add rubygem-foreman_resource_quota package

bastian-src:
Adding an rpm build for the [Foreman Resource Quota](https://github.com/ATIX-AG/foreman_resource_quota) plugin.

I've used the `add_gem_package` script and it changed the order of some entries in `comps/comps-foreman-el8` and `comps/comps-foreman-plugin`. I've removed such changes from my commit, but additional verification during the review would be nice.


packages/plugins/rubygem-foreman_resource_quota/rubygem-foreman_resource_quota.spec
```diff
@@ -0,0 +1,93 @@
+# template: foreman_plugin
+%global gem_name foreman_resource_quota
+%global plugin_name resource_quota
+%global foreman_min_version 3.5.0
```
sbernhard, line 4: current foreman version is 3.10. should we bump it?

sbernhard (approved):
LGTM. Thanks.

packages/plugins/rubygem-foreman_resource_quota/rubygem-foreman_resource_quota.spec
evgeni: This comes from the plugin source. Please update it there if needed.

evgeni (approved):

bastian-src:
Thanks @evgeni!

PR #10763, opened 2024-05-07
Fixes #37419 - Patch puma to fix chunked upload bug

ianballou:
Add a patch to solve https://github.com/puma/puma/issues/3337 , which causes a Katello `podman push` to fail for all but  the smallest container images.

I tested that the patch applies to Puma 6.4.2.


ekohl (approved):
Since there isn't a release, I'm OK with applying this. Do you intend to cherry pick this to stable branches? If so, perhaps a Redmine issue is a good idea.

ianballou:
@ekohl it seems to only affect container push, which is going to be partially available for Foreman 3.11, so I wasn't planning on doing backporting.

I'll add a redmine anyway.

ianballou:
@ekohl I'll need help with merging

PR #10663, opened 2024-04-02
Update katello-host-tools to 4.4.0

ianballou:
Bumping k-h-t version since the move to `dnf needs-restarting` is included.

PR #9435, opened 2023-05-24
Release katello-host-tools 4.3.0

wbclark:

PR #10765, opened 2024-05-10
Bump to katello-host-tools 4.4.0

sbernhard:
Based on
- https://github.com/theforeman/foreman-packaging/pull/10663
- https://github.com/theforeman/foreman-packaging/pull/9435

### DIFF
--- a/packages/client/katello-host-tools/katello-host-tools.spec
+++ b/packages/client/katello-host-tools/katello-host-tools.spec
@@ -3,9 +3,6 @@
 %global zypper_install (0%{?suse_version} > 0)
 %global build_tracer 0%{?rhel} >= 7 || 0%{?fedora} || 0%{?suse_version}
 
-%global build_agent 0
-%global legacy_agent 0
-
 %if 0%{?suse_version}
 %define dist suse%{?suse_version}
 %endif
@@ -36,8 +33,8 @@
 %global katello_libdir %{python_libdir}/katello
 
 Name: katello-host-tools
-Version: 4.2.3
-Release: 6%{?dist}
+Version: 4.4.0
+Release: 1%{?dist}
 Summary: A set of commands and yum plugins that support a Katello host
 Group:   Development/Languages
 %if 0%{?suse_version}
@@ -45,8 +42,8 @@ License: LGPL-2.0
 %else
 License: LGPLv2
 %endif
-URL:     https://github.com/Katello/katello-agent
-Source0: https://codeload.github.com/Katello/katello-host-tools/tar.gz/%{version}#/%{name}-%{version}.tar.gz
+URL:     https://github.com/Katello/katello-host-tools
+Source0: https://codeload.github.com/Katello/%{name}/tar.gz/%{version}#/%{name}-%{version}.tar.gz
 
 %if 0%{?suse_version} && 0%{?suse_version} < 1200
 # this needs to be BuildArch for Suse but
@@ -58,6 +55,7 @@ BuildArch: noarch
 
 Requires: subscription-manager
 Obsoletes: %{name}-fact-plugin < %{version}-%{release}
+Obsoletes: katello-agent < %{version}-%{release}
 
 %if %{dnf_install} || 0%{?suse_version} >= 1500
 Requires: python3-subscription-manager-rhsm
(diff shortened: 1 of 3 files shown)


packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -94,46 +95,6 @@ BuildRequires: python-setuptools
 %description
 A set of commands and yum plugins that support a Katello host including faster package profile uploading and bound repository reporting.  This is required for errata and package applicability reporting.
 
-%if %{build_agent}
-%package -n katello-agent
```
ekohl, line 98: Do we need to Obsolete this package so it gets removed?

sbernhard: If we really want to get rid of it, I can add this. Please decide.

sbernhard: Done. Thanks!

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -58,6 +55,11 @@ BuildArch: noarch
 
 Requires: subscription-manager
 Obsoletes: %{name}-fact-plugin < %{version}-%{release}
+Obsoletes: katello-agent
```
ekohl, line 58: It's a best practice to include a version:
```suggestion
Obsoletes: katello-agent < %{version}-%{release}
```

This means that in the future it could be reintroduced. rpmlint should warn about this.

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -36,16 +33,16 @@
...
 Summary: A set of commands and yum plugins that support a Katello host
 Group:   Development/Languages
 %if 0%{?suse_version}
 License: LGPL-2.0
 %else
 License: LGPLv2
 %endif
-URL:     https://github.com/Katello/katello-agent
+URL:     https://github.com/Katello/katello-host-tools
 Source0: https://codeload.github.com/Katello/katello-host-tools/tar.gz/%{version}#/%{name}-%{version}.tar.gz
```
ekohl, line 46: I think it's a good practice to use variables.
```suggestion
Source0: https://codeload.github.com/Katello/%{name}/tar.gz/%{version}#/%{name}-%{version}.tar.gz
```

ekohl (commented):
I haven't had a look at the generated RPMs yet to see what they actually package.

evgeni:
* `katello-host-tools` contains `/usr/lib/python*/site-packages/katello/tracer/__pycache__`, this seems wrong :) (but not a problem from DNF perspective, as it's a folder)
* `katello-host-tools-tracer.el7` contains `/usr/lib/python2.7/site-packages/katello/tracer/dnf.py` but there is no DNF on el7 (by default). If I read https://github.com/Katello/katello-host-tools/blob/37c7fdfe0b9fdfb794174415098d45b46aac3c71/src/katello/tracer/__init__.py#L9-L21 this will "do the right thing" (and fallback to original tracer), but I wonder if it's a good idea to differ depending on "user did install dnf on el7 or not"
* `katello-host-tools-tracer.el7` has no dependency on `python2-tracer` anymore, so the above will actually not work
* no idea about the suse builds, we don't build them :/

evgeni:
I've opened https://github.com/theforeman/foreman-packaging/pull/10804 so that we at least try building suse 15.5

evgeni:
#10804 was merged, so feel free to rebase and see what happens :)

evgeni:
Yay, repoclosure failures. I'll see if we can fix those.

sbernhard:
> Yay, repoclosure failures. I'll see if we can fix those.

maybe you want to use the build dependencies from https://oss.atix.de/html/sles.html

evgeni:
- [ ] https://github.com/theforeman/jenkins-jobs/pull/466
- [ ] https://github.com/theforeman/foreman-packaging/pull/10813

evgeni:
[test rpm-copr]

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -58,6 +55,11 @@ BuildArch: noarch
 
 Requires: subscription-manager
 Obsoletes: %{name}-fact-plugin < %{version}-%{release}
+Obsoletes: katello-agent < %{version}-%{release}
+
+%if %{dnf_install}
+Requires: python3-psutil
+%endif
```
evgeni, line 62: This should go to the tracer subpackage, as that's the one using `psutil`

evgeni:
Three tiny things:
* the psutil dep should be in the tracer subpackage (that's the only thing I'd expect to be fixed to merge this, the rest is cosmetics)
* the suse and el7 tracer packages end up having `tracer/SystemdDbus.py`, which doesn't hurt but also is unused there
* the k-h-t package on el8 and el9 contain `/usr/lib/python*/site-packages/katello/tracer/` directory (empty), again, doesn't hurt but seems odd.

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -328,25 +236,13 @@ exit 0
...
-%if %{yum_install}
-%exclude %{katello_libdir}/tracer.py*
-%else
-%pycached %exclude %{katello_libdir}/tracer.py
-%endif
-%endif
-
 %if %{dnf_install}
+%pycached %exclude %{katello_libdir}/tracer/*.py
+%exclude %{katello_libdir}/tracer/__pycache__
```
evgeni, line 242: I wonder if moving this block to the tracer subpackage (and dropping the `%exclude`) will fix the "el8/9 has empty tracer dir" complaint I had

PR #10731, opened 2024-04-25
un-SCL-ify hammer-cli-katello

ianballou:
Makes the hammer-cli-katello spec file no longer use ye olde SCL macros from the EL7 days.

PR #10767, opened 2024-05-10
Bump hammer-cli-katello to 1.13

ianballou:
Now that the nightly source pipeline is green, we need to bump this to get the package out.


ianballou:
Merge https://github.com/theforeman/foreman-packaging/pull/10731 instead if it's all good to go. Otherwise, this works in the meantime.

evgeni (approved):

PR #10803, opened 2024-05-17
Add libyaml-dev to Depends for psych

evgeni:


evgeni:
@ekohl let's see if that fixes it finally

ekohl (approved):

PR #10812, opened 2024-05-22
merge back plugins-copr-with-el9 into plugins-copr

evgeni:


evgeni:
@zjhuntin this is what broke the branching for you yesterday

ekohl (approved):

PR #10842, opened 2024-05-28
Rebuild kafo and foreman_maintain

evgeni:
It seems these two were missed during branching.


evgeni:
3.11 has 7.5.0/1.6.7 so just bumping the release version is safe.

ekohl (approved):

PR #10863, opened 2024-05-31
katello-repos: Use Pulpcore 3.49, not nightly

evgeni:


evgeni:
:eyes: @pcreech @zjhuntin @Odilhao

PR #10880, opened 2024-06-03
Check GPG key on Candlepin repository

ehelms:
We are using versioned and signed Candlepin repositories, so there is no reason to mark gpgcheck as 0 in `rpm/develop`. This will then make it easier when branching as no special actions are needed.


ekohl (commented):
Today we have `candlepin_version` set to 4.4 here, but also in forklift (https://github.com/theforeman/forklift/blob/649cb93a27f3d09365f8b9276bfe315822b6c8bb/vagrant/config/versions.yaml#L106). I thought we wanted to track the nightly version instead.

dgoetz:
While on this would be this also a good time to look in the [user request](https://community.theforeman.org/t/could-the-candlepin-gpg-keys-be-published-somewhere-http-reachable/38193) to have the key in the repository online in addition to the package?

ehelms:
> I thought we wanted to track the nightly version instead.

I am more leaning towards nightly is more work than it's worth due to the slower update cycle, and ease of which it is to update Candlepin. And that it saves energy and time to just align on a version of Candlepin.

ekohl:
I think the biggest challenge is that you'd constantly need to change the default branch. We saw with pulpcore-packaging that people forget. Git clones also don't automatically pick that up either. It's kind of unfortunate.

I don't think GitHub support branch aliases.

ehelms:
> I think the biggest challenge is that you'd constantly need to change the default branch. We saw with pulpcore-packaging that people forget. Git clones also don't automatically pick that up either. It's kind of unfortunate.
> 
> I don't think GitHub support branch aliases.

I'm not following how git and branches affect.

evgeni:
https://github.com/theforeman/forklift/pull/1829

ehelms:
Thanks for tackling that! I'll let you do the honors of updating https://community.theforeman.org/t/could-the-candlepin-gpg-keys-be-published-somewhere-http-reachable/38193 with the news

packages/katello/katello-repos/katello.repo
```diff
@@ -15,7 +15,7 @@ name=Candlepin: an open source entitlement management system.
 baseurl=https://yum.theforeman.org/candlepin/@CANDLEPIN_VERSION@/@DIST@/$basearch/
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-candlepin
```
evgeni, line 16:
```suggestion
gpgkey=https://yum.theforeman.org/candlepin/@CANDLEPIN_VERSION@/RPM-GPG-KEY-candlepin
```

packages/katello/katello-repos/katello.repo
```diff
@@ -38,7 +38,7 @@ name=Katello Candlepin source
 baseurl=https://yum.theforeman.org/candlepin/@REPO_VERSION@/@DIST@/source/
```
evgeni, line 38:
```suggestion
baseurl=https://yum.theforeman.org/candlepin/@CANDLEPIN_VERSION@/@DIST@/source/
```

packages/katello/katello-repos/katello.repo
```diff
@@ -38,7 +38,7 @@ name=Katello Candlepin source
...
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-candlepin
```
evgeni, line 39:
```suggestion
gpgkey=https://yum.theforeman.org/candlepin/@CANDLEPIN_VERSION@/RPM-GPG-KEY-candlepin
```

packages/katello/katello-repos/katello.repo
ekohl: I'd still prefer to include the GPG keys in the file so it can validate offline. That's the whole point of this package. I now see we don't do it for Pulp either.

One problem with networked URLs is that yum/dnf never checks expiration. So if the keys expire, you must manually replace them. Now this probably isn't a big deal for us if users stay on supported releases, given how often we use new keys but still something to keep in mind.

It does mean more work in the release process.

evgeni: Validate what offline? The RPMs that you need to be online to fetch?
(If you're mirroring the repo, that package is useless anyway and you can to mirror the key too)

To make that more viable, we'd need to ship separate packages for these things, like CentOS does:
```
# rpm -qa centos*
centos-gpg-keys-9.0-23.el9.noarch
centos-stream-repos-9.0-23.el9.noarch
centos-stream-release-9.0-23.el9.noarch
```

(This is not me volunteering to implement this)

ekohl: I meant to include the GPG key in this file. And I too thought about separate release files for each repo, but wasn't keen on implementing it.

ehelms:
@ekohl I think this should go in before branch

packages/katello/katello-repos/katello.repo
```diff
@@ -35,10 +35,10 @@ gpgcheck=@REPO_GPGCHECK@
 
 [candlepin-source]
 name=Katello Candlepin source
-baseurl=https://yum.theforeman.org/candlepin/@REPO_VERSION@/@DIST@/source/
-gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-candlepin
+baseurl=https://yum.theforeman.org/candlepin/@CANDLEPIN_VERSION@/@DIST@/source/
+gpgkey=https://yum.theforeman.org/candlepin/@CANDLEPIN_VERSION@/RPM-GPG-KEY-candlepin
```
evgeni, line 39: here you updated to the web url, but not above. seems inconsistent.

packages/katello/katello-repos/katello.repo
ekohl: Implemented my preference in https://github.com/theforeman/foreman-packaging/pull/11128.

PR #10899, opened 2024-06-06
pin FFI to a version compatible with Ruby 2.7

evgeni:


evgeni:
[test deb]

evgeni:
[test deb]

PR #10931, opened 2024-06-17
scap client: support building on suse

evgeni:

### DIFF
--- a/packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
+++ b/packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
@@ -3,9 +3,18 @@
 
 %define rubyabi 1.8
 
+%if 0%{?suse_version}
+%define mod_name %{gem_name}
+%define gem_instdir %{gem_base}/gems/%{gem_name}-%{version}
+%define gem_libdir %{gem_instdir}/lib
+%define gem_docdir %{gem_base}/doc/%{gem_name}-%{version}
+%define gem_cache %{gem_base}/cache/%{gem_name}-%{version}.gem
+%define gem_spec %{gem_base}/specifications/%{gem_name}-%{version}.gemspec
+%endif
+
 Name: rubygem-%{gem_name}
 Version: 0.5.3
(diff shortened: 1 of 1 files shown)


packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
```diff
@@ -3,32 +3,49 @@
...
 %if 0%{?el6}
 Requires: ruby(abi)
 Requires: rubygem(json) >= 1.4
 Requires: rubygem(json) < 2.0
 %else
+%if 0%{?suse_version} == 0
 Requires: ruby(release)
 %endif
+%endif
 Requires: openscap-scanner
```
evgeni, line 37: that's `openscap-utils` on SUSE

packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
```diff
@@ -3,32 +3,49 @@
 
 %define rubyabi 1.8
```
ekohl, line 4: Is this still used?

ekohl (approved):
The generated files bit looks ok, so I trust you on this (after fixing repoclosure).

packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
ekohl:
```suggestion
%if 0%{?suse_version}
Requires: openscap-utils
%else
Requires: openscap-scanner
%endif
```

packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
evgeni: on el6, yes, but we don't build for that, so can be cleaned up I guess

ekohl: We still have other el6 macros, so we either clean them all up or leave them all in.

PR #10937, opened 2024-06-18
Add rubygem-spidr package

sjha4:
Opening PR based on commit created by:
```
./add_gem_package.sh spidr foreman_plugin katello
```


sjha4:
Probably the foreman-plugin template in command above added it to foreman specific comps..Will move to katello.

packages/katello/rubygem-spidr/rubygem-spidr.spec
```diff
@@ -0,0 +1,85 @@
+# template: foreman_plugin
```
ekohl, line 1: This isn't a Foreman plugin, but just a regular gem. Please use the `gem` template.

ekohl (changes requested):

packages/katello/rubygem-spidr/rubygem-spidr.spec
sjha4: Screenshot from 2024-06-20 16-28-29
These are the templates that I see as available. I was trying to use Default but that doesn't seem to be accepted anymore..

evgeni: `default` should do

sjha4:
Used:
```
./add_gem_package.sh spidr default katello katello-el8
./add_to_comps.rb comps/comps-katello-el8.xml rubygem-spidr
./add_to_comps.rb comps/comps-katello-el9.xml rubygem-spidr
```

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -45,6 +45,7 @@ BuildRequires: rubygem(apipie-rails) >= 0.5.14
 BuildRequires: rubygem(fx) < 1.0
 BuildRequires: rubygem(pg)
 BuildRequires: rubygem(anemone)
+BuildRequires: rubygem(spidr)
```
ekohl, line 48: You shouldn't add this in the same commit because it breaks things. It's a limitation in our build system, but you can only do this one this PR has been merged.

ekohl (changes requested):

packages/katello/rubygem-spidr/rubygem-spidr.spec
```diff
@@ -0,0 +1,74 @@
...
+        %{buildroot}%{gem_dir}/
+
+%files
+%dir %{gem_instdir}
+%exclude %{gem_instdir}/.editorconfig
+%exclude %{gem_instdir}/.github
+%exclude %{gem_instdir}/.gitignore
+%exclude %{gem_instdir}/.yardopts
+%license %{gem_instdir}/LICENSE.txt
+%{gem_instdir}/gemspec.yml
```
evgeni, line 56:
```suggestion
%exclude %{gem_instdir}/gemspec.yml
```

sjha4:
Seems like the EL8 and 9 copr builds are green but the rpm-copr fails with a repo-closure issue..Not sure how to fix that..

evgeni:
Doesn't look like an error you introduced, more something on CentOS' end. Ignore!

evgeni:
Ah, technically comps and package manifest entries should be sorted. Would you mind fixing that?

sjha4:
Updated 👍🏼

evgeni:
@ekohl any further comments from your side?

ehelms (approved):

PR #10951, opened 2024-06-21
Use $releasever_major in release files

ekohl:
This makes distribution upgrades with leapp smoother. Well, this PR is to test that theory.

### DIFF
--- a/packages/foreman/foreman-release/foreman-release.spec
+++ b/packages/foreman/foreman-release/foreman-release.spec
@@ -12,11 +12,11 @@
 %define repo_dir %{_sysconfdir}/yum.repos.d
 
 %if 0%{?rhel}
-%define repo_dist el%{rhel}
+%define repo_dist el$releasever_major
 %endif
 %endif
 
-%global release 1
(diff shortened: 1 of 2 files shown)


evgeni:
That's a feature of DNF 4.18, EL9 has 4.14… ;)

ekohl:
Of course, having that usable would be too easy. Is it bad that I'm already looking forward to dropping EL 9 support now? ;)

evgeni:
it's the circle of life :)

packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -12,11 +12,11 @@
 %define repo_dir %{_sysconfdir}/yum.repos.d
 
 %if 0%{?rhel}
-%define repo_dist el%{rhel}
+%define repo_dist el$releasever_major
```
ekohl, line 15: I think we should at least align both packages on `%{dist}`: https://github.com/theforeman/foreman-packaging/pull/10952

PR #10952, opened 2024-06-21
Use %{dist} to determine repo_dist in foreman-release

ekohl:
On CentOS this is defined as .elVERSION and Fedora as .fcVERSION, which means there's no more need a conditional rhel/fedora should we choose to support Fedora again. The later code already strips the leading dot.

This mirrors the code that's already in katello-repos.

### DIFF
--- a/packages/foreman/foreman-release/foreman-release.spec
+++ b/packages/foreman/foreman-release/foreman-release.spec
@@ -10,13 +10,10 @@
 
 %else
 %define repo_dir %{_sysconfdir}/yum.repos.d
-
-%if 0%{?rhel}
-%define repo_dist el%{rhel}
-%endif
+%define repo_dist %{dist}
 %endif
 
-%global release 1
+%global release 2
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman-release/foreman-release.spec
```diff
@@ -10,13 +10,10 @@
 
 %else
 %define repo_dir %{_sysconfdir}/yum.repos.d
-
-%if 0%{?rhel}
-%define repo_dist el%{rhel}
-%endif
+%define repo_dist %{dist}
```
evgeni, line 13: That's not the same tho.

```
[root@0f95f2b4985d /]# rpm -E '%dist'
.el7_9
[root@0f95f2b4985d /]# rpm -E '%rhel'
7
```

evgeni: https://github.com/theforeman/foreman-packaging/commit/f515215ed3d45bfd6d0d387bf0d10a08fa1084cf

ekohl: Looking at katello-repos it does look like for EL 8 & EL 9 it is correct, so it's just for EL 7 that it's "broken"?

evgeni: Today, yes.
Once RHEL8 becomes "LTS" I expect it to get there too.

ekohl: They really hate people who maintain repositories, don't they?

PR #10975, opened 2024-07-02
Update ruby-smart-proxy-dns-infoblox to 1.2.0

timdeluxe:
(cherry picked from commit 9230685718c11c33beb7bdd6f80235bd4aaa6216)


evgeni:
3.9 is EOL and we stopped publishing updates there, sorry.

timdeluxe:
> 3.9 is EOL and we stopped publishing updates there, sorry.

That is okay. I tried to find a service life / EOL overview of Foreman, but that is really hard to find. Maybe a hint for improvement? ;-)

evgeni:
The usual rule of thumb is "we support the latest two releases", so when 3.11 got released, 3.9 became EOL automatically.
But I guess it could be a bit cleaner, indeed.

PR #10990, opened 2024-07-08
fix repoclosure for Client due to CentOS 7 EOL

evgeni:
- **use vault.centos.org for CentOS 7 content**
- **use dl.fedoraproject.org for EPEL7**


ekohl (approved):

evgeni:
picked to 3.11 (https://github.com/theforeman/foreman-packaging/pull/10991) and 3.10 (https://github.com/theforeman/foreman-packaging/pull/10992)

PR #11027, opened 2024-07-14
add katello host tools

evgeni:


evgeni:
@sbernhard what are your thoughts about compatibility with the atix packages?

PR #11045, opened 2024-07-19
Add spidr gem to 3.11 needed for el9 - PART 2

parthaa:


parthaa:
Merge https://github.com/theforeman/foreman-packaging/pull/11046 before this

packages/katello/rubygem-katello/rubygem-katello.spec
```diff
@@ -42,7 +42,7 @@ BuildRequires: rubygem(gettext_i18n_rails)
 BuildRequires: rubygem(apipie-rails) >= 0.5.14
 BuildRequires: rubygem(fx) < 1.0
 BuildRequires: rubygem(pg)
-BuildRequires: rubygem(anemone)
+BuildRequires: rubygem(spidr)
```
evgeni, line 45: I think this would only work when we have a 4.13.1 gem that swtiched the dep, otherwise it won't build.

So, maybe, this commit should be part of the "4.13.1" release PR?

zjhuntin (approved):

packages/katello/rubygem-katello/rubygem-katello.spec
ekohl: Yes, you're right about that.

ekohl (changes requested):
This isn't valid until we release 4.13.1.

Odilhao:
I think we can continue here https://github.com/theforeman/foreman-packaging/pull/11061 in the 4.13.1 PR

PR #11047, opened 2024-07-22
dont exit add_npm_package when compgen is empty

MariaAga:
Not sure if this is the best solution, but the script exists right after running `PACKAGE_DIR=$(compgen -G "packages/*/${PACKAGE_NAME}/")`
without any error. 
Only happens when the folder is empty, which is to be expected when someone adds a new npm package.
Any other solutions are welcome, or doc changes if I'm running something wrong?

### DIFF
--- a/add_npm_package.sh
+++ b/add_npm_package.sh
@@ -18,10 +18,7 @@ else
   PACKAGE_NAME=nodejs-${PACKAGE_MODULE}
 fi
 
-PACKAGE_DIR=$(compgen -G "packages/*/${PACKAGE_NAME}/")
-if [[ -z "${PACKAGE_DIR}" ]]; then
-  PACKAGE_DIR=packages/$BASE_DIR/$PACKAGE_NAME
-fi
+PACKAGE_DIR=$(compgen -G "packages/*/${PACKAGE_NAME}/" || echo "packages/$BASE_DIR/$PACKAGE_NAME")
 SPEC_FILE="${PACKAGE_DIR}/${PACKAGE_NAME}.spec"
 
 ROOT=$(git rev-parse --show-toplevel)


ekohl (changes requested):
If anything, I'd disable it locally.

Some testing shows you're right:
```console
$ compgen -G "packages/*/nodejs-test"
$ echo $?
1
```

Now I'm starting to wonder how this ever worked. @evgeni did e539c7b7518b01ae862891d8c29e6163b6b7707b break this?

evgeni:
I think it did, yeah. And given we do not add packages often, we didn't notice.

Would `SPEC_FILE=$(compgen -G "packages/*/${PACKAGE_NAME}/${PACKAGE_NAME}.spec" || '')` work?

MariaAga:
adding `|| '')` did resolve the issue as well, thanks!

add_npm_package.sh
```diff
@@ -18,10 +18,7 @@ else
   PACKAGE_NAME=nodejs-${PACKAGE_MODULE}
 fi
 
-PACKAGE_DIR=$(compgen -G "packages/*/${PACKAGE_NAME}/")
-if [[ -z "${PACKAGE_DIR}" ]]; then
-  PACKAGE_DIR=packages/$BASE_DIR/$PACKAGE_NAME
-fi
+PACKAGE_DIR=$(compgen -G "packages/*/${PACKAGE_NAME}/") || PACKAGE_DIR=packages/$BASE_DIR/$PACKAGE_NAME
```
evgeni, line 21:
```suggestion
PACKAGE_DIR=$(compgen -G "packages/*/${PACKAGE_NAME}/" || echo "packages/$BASE_DIR/$PACKAGE_NAME")
```

Does that work too?

MariaAga: yeah, is it preferred?

ekohl: I generally prefer assigning a variable once if possible.

MariaAga: sure, done

evgeni (approved):

ekohl (approved):

PR #11048, opened 2024-07-22
Add nodejs-js-cookie package

MariaAga:


comps/comps-foreman-el8.xml
ekohl, line 1: Could you also add the same to `comps/comps-foreman-el9.xml`? Looks like our script doesn't account for that.

ekohl (changes requested):

comps/comps-foreman-el8.xml
MariaAga: added

ekohl (approved):
Thanks!

PR #11056, opened 2024-07-24
run packaging automation on Sunday *and* Wednesday

evgeni:
This should allow things released Mo/Tu to end up in nightly earlier

### DIFF
--- a/.github/workflows/bump_packages.yml
+++ b/.github/workflows/bump_packages.yml
@@ -3,7 +3,7 @@ name: Update packages
 on:
   workflow_dispatch:
   schedule:
-    - cron: '4 4 * * 0'
+    - cron: '4 4 * * 0,3'
 
 env:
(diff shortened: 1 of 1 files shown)


nadjaheitmann:
That sounds like a good idea :)

ekohl (approved):
We should really fix some of the features. For example, I see `fog-vsphere` would be a downgrade and because we have too many other (long time) failures we don't see that anymore.

Edit: that is probably because foreman develop itself hasn't passed recently so our lockfile is outdated.

evgeni:
> We should really fix some of the features. For example, I see `fog-vsphere` would be a downgrade and because we have too many other (long time) failures we don't see that anymore.
> 
> Edit: that is probably because foreman develop itself hasn't passed recently so our lockfile is outdated.

Yeah. after-PTO me wanted to look at this :)

PR #11062, opened 2024-07-25
Add foreman_scap_client_bash

ofedoren:


packages/client/foreman_scap_client_bash/foreman_scap_client_bash.spec
```diff
@@ -0,0 +1,45 @@
...
+Release: 1%{?dist}
+Summary: Client script that runs OpenSCAP scan and uploads the result to foreman proxy
+
+License: GPLv3
+URL: https://github.com/theforeman/foreman_scap_client_bash
+Source0: https://github.com/%{repo_orgname}/%{repo_name}/releases/download/v%{version}/%{repo_name}-%{version}.tar.gz
+
+Requires: bash
+Requires: curl
+Requires: openscap-scanner
```
adamruzicka, line 17: According to the old package, it is `openscap-utils` on suse

ofedoren:
Updated .spec file, waiting for https://github.com/theforeman/foreman_scap_client_bash/pull/3 to be merged, so I can release 0.2.0 and add it here.

packages/client/rubygem-foreman_scap_client/rubygem-foreman_scap_client.spec
```diff
@@ -54,6 +54,7 @@ BuildRequires: rubygems-devel
 BuildRequires: ruby
 BuildArch: noarch
 Provides: rubygem(%{gem_name}) = %{version}
+Provides: deprecated()
```
evgeni, line 57: What does this do? :D

ofedoren: https://docs.fedoraproject.org/en-US/packaging-guidelines/deprecating-packages/#_marking_a_package_deprecated

AFAIU, deprecates the package? I mean, this package should be installed instead of rubygem. I guess it's enough to have `Obsoletes` in the new package, but just to be sure I've marked rubygem one as deprecated.

evgeni: TIL

Yeah I don't think it does anything valuable for us. The `Obsoletes` should be sufficient.

ofedoren: Dropped the commit with deprecation.

ofedoren:
:green_apple:

adamruzicka (approved):

packages/client/foreman_scap_client_bash/foreman_scap_client_bash.spec
```diff
@@ -0,0 +1,50 @@
...
+Requires: curl
+%if 0%{?suse_version}
+Requires: openscap-utils
+%else
+Requires: openscap-scanner
+%endif
+Requires: bzip2
+BuildArch: noarch
+
+Obsoletes: rubygem-foreman_scap_client
```
evgeni, line 25: Can we please have an versioned obsolete, so we can come back to Ruby if things go south?
```suggestion
Obsoletes: rubygem-foreman_scap_client < 0.6.2-2
```

evgeni (approved):

packages/client/foreman_scap_client_bash/foreman_scap_client_bash.spec
ofedoren: https://github.com/theforeman/foreman-packaging/pull/11636

PR #11079, opened 2024-08-01
Direct the users to use the new CA refresh script instead of the rpm

ShimShtein:
Update the katello host rename script documentation to the new CA refresh script introduced in https://github.com/theforeman/foreman/pull/10208

### DIFF
--- a/packages/katello/katello/hostname-change.rb
+++ b/packages/katello/katello/hostname-change.rb
@@ -143,8 +143,7 @@ module KatelloUtilities
   You will have to install the new bootstrap rpm and reregister all clients and #{@plural_proxy} with subscription-manager
   (update organization and environment arguments appropriately):
 
-  yum remove -y katello-ca-consumer*
-  rpm -Uvh http://#{@new_hostname}/pub/katello-ca-consumer-latest.noarch.rpm
+  set -o pipefail && curl -sS "http://#{@new_hostname}/unattended/public/foreman-ca-refresh" | bash
   subscription-manager register --org="Default_Organization" --environment="Library" --force
 
   Then reattach subscriptions to the client(s) and run:


packages/katello/katello/hostname-change.rb
```diff
@@ -143,8 +143,7 @@ def next_steps_message
   You will have to install the new bootstrap rpm and reregister all clients and #{@plural_proxy} with subscription-manager
```
sbernhard, line 143: Please fix this comment.
Maybe something like:

You will have to apply the new katello-ca-consumer and reregister all clients and #{@plural_proxy} with subscription-manager
Comment

packages/katello/katello/hostname-change.rb
```diff
@@ -143,8 +143,7 @@ def next_steps_message
...
   (update organization and environment arguments appropriately):
 
-  yum remove -y katello-ca-consumer*
-  rpm -Uvh http://#{@new_hostname}/pub/katello-ca-consumer-latest.noarch.rpm
+  set -o pipefail && curl -sS "http://#{@new_hostname}/unattended/public/foreman-ca-refresh" | bash
```
sbernhard, line 146: Is this what you want?

```suggestion
  curl -sS "http://#{@new_hostname}/unattended/public/foreman-ca-refresh" | bash -o pipefail
```

packages/katello/katello/hostname-change.rb
```diff
@@ -143,8 +143,7 @@ def next_steps_message
...
   subscription-manager register --org="Default_Organization" --environment="Library" --force
 
   Then reattach subscriptions to the client(s) and run:
```
sbernhard, line 149: attach is no longer necessary with SCA?

packages/katello/katello/hostname-change.rb
```diff
@@ -143,8 +143,7 @@ def next_steps_message
   You will have to install the new bootstrap rpm and reregister all clients and #{@plural_proxy} with subscription-manager
   (update organization and environment arguments appropriately):
 
-  yum remove -y katello-ca-consumer*
-  rpm -Uvh http://#{@new_hostname}/pub/katello-ca-consumer-latest.noarch.rpm
+  set -o pipefail && curl -sS "http://#{@new_hostname}/unattended/public/foreman-ca-refresh" | bash
   subscription-manager register --org="Default_Organization" --environment="Library" --force
```
sbernhard, line 147: Is this the default way to register? I think, using a activation key is more common. 

Or should is use the HostRegistration template for the whole process?

packages/katello/katello/hostname-change.rb
ShimShtein: No. When curl fails, we want  to stop bash from running, otherwise the return code of the action would be `0`.

packages/katello/katello/hostname-change.rb
ShimShtein: @ianballou any thoughts on this one? I don't mind doing the change, but open to suggestions about what is needed here.

packages/katello/katello/hostname-change.rb
ShimShtein: Host registration is actually a good idea, but we can't put here the command that initiates it, since each invocation generates a new token.
Maybe we should change this part to something that will say "Go to Foreman UI and generate a registration command" (with proper links of course)?

packages/katello/katello/hostname-change.rb
ehelms: When the hostname changes on the server, we don't need to perform a CA refresh, we need to perform a sub-man configuration update (to point to the new hostname).

packages/katello/katello/hostname-change.rb
ehelms: This seems entirely wrong. Just because the hostname changes does not require re-registration of a host.

ShimShtein: If we change Foreman's host name, we will definitely need to reconfigure rhsm to point to the new server, and sometimes (depends if the cert is self-managed for example) we will need the CA to be updated. I think that simplifying the procedure takes precedence here, and having a single procedure will do us good.

ehelms: None of that requires a host to be re-registered.

packages/katello/katello/hostname-change.rb
ehelms: As I stated above, changing the hostname should not require clients to be re-registered, only re-configured.

packages/katello/katello/hostname-change.rb
ShimShtein: @ehelms I take your point about changing from re-registration to reconfiguration.
Now about hostname change: is there a chance that hostname change also requires change to a certificate? For example if the CA is self-signed?

ehelms: Not the CA certificate since it is not tied to a hostname.

sbernhard: I guess, if its only about changing the subscription-manager configuration, something like this is sufficient: https://github.com/theforeman/foreman/blob/develop/app/views/unattended/provisioning_templates/snippet/subscription_manager_setup.erb#L108 ?

Then we should also get rid of https://github.com/theforeman/foreman-packaging/blob/708b947c6bbfe504536aad5d7d70eda34de1a083/packages/katello/katello/hostname-change.rb#L147 and rewrite https://github.com/theforeman/foreman-packaging/blob/708b947c6bbfe504536aad5d7d70eda34de1a083/packages/katello/katello/hostname-change.rb#L149

PR #11096, opened 2024-08-05
Add rubygem-cgi package

ehelms:


ehelms:
New dependency needed by https://github.com/theforeman/foreman-packaging/pull/9759

evgeni:
isn't that a std gem that is only required in 3.3+?

PR #11097, opened 2024-08-05
Add rubygem-base64 package

ehelms:


evgeni:
isn't that a std gem that is only required in 3.3+?

ehelms:
Good question.... is autorequires doing dumb things? 

```
[2024-08-13T12:22:03.254Z]         Error: Repoclosure ended with unresolved dependencies (1) across 1 packages.
[2024-08-13T12:22:03.254Z]         package: rubygem-jwt-2.8.2-1.el8.noarch from repo0
[2024-08-13T12:22:03.254Z]           unresolved deps (1):
[2024-08-13T12:22:03.254Z]             rubygem(base64)
```

ekohl:
Looking at https://stdgems.org/ then we see that Ruby 3.3 and earlier all list base64 as a default gem. https://docs.ruby-lang.org/en/master/NEWS_md.html does mention it will turn into a bundled gem.

You should use `%gemspec_remove_dep base64` to drop it from the gemspec and manually add a dependency on `ruby-default-gems` as a rich dependency: `(rubygem(base64) or ruby-default-gems < 3.4)`. That way we will be ready when we get to Ruby 3.4.

ehelms:
> Looking at https://stdgems.org/ then we see that Ruby 3.3 and earlier all list base64 as a default gem. https://docs.ruby-lang.org/en/master/NEWS_md.html does mention it will turn into a bundled gem.
> 
> You should use `%gemspec_remove_dep base64` to drop it from the gemspec and manually add a dependency on `ruby-default-gems` as a rich dependency: `(rubygem(base64) or ruby-default-gems < 3.4)`. That way we will be ready when we get to Ruby 3.4.

I am trying to capture this into the docs for future reference -- https://github.com/theforeman/foreman-packaging/pull/11119

PR #11101, opened 2024-08-08
Add a filter parameter for bump_packages

ekohl:
Applying this filter allows calling it from a specific plugin to only update a single package. For example, when a gem is released it can automatically trigger this workflow.

Right now the actual scripts don't implement this, so this is just a proposal. My idea is that we enhance our [release-gem](https://github.com/theforeman/actions/blob/v0/.github/workflows/release-gem.yml) workflow to trigger a packaging bump. That way we don't update all existing PRs but still have nice end-to-end automation for nightly. Stable branches still need additional work.

### DIFF
--- a/.github/workflows/bump_packages.yml
+++ b/.github/workflows/bump_packages.yml
@@ -2,6 +2,12 @@ name: Update packages
 
 on:
   workflow_dispatch:
+    inputs:
+      filter:
+        type: string
+        description: A filter to apply on package names
+        required: false
+        default: ""
   schedule:
     - cron: '4 4 * * 0,3'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/bump_packages.yml
```diff
@@ -2,6 +2,12 @@ name: Update packages
 
 on:
   workflow_dispatch:
+    inputs:
+      filter:
+        type: string
+        description: A filter to apply on package names
```
evgeni, line 8: Package names have different structure between Debian and EL. Should this filter rather be on the "source" name?
So, passing `foreman_maintain` here will update `ruby-foreman-maintain` on Debian and `rubygem-foreman_maintain` on RPM?

ekohl: I think so. In the update scripts we know the gem name so we can match against that

evgeni:
heh, so I ended up forgetting about this, and re-implementing it in https://github.com/theforeman/foreman-packaging/pull/12340 with a couple of differences:
- when this is called via schedule, the input is undefined, so I erred on the side of not relying on the default
- in the corresponding `list_updateable_packages` PRs I used "substring" matches and for debian used the gem name, so it should be fine?

PR #11109, opened 2024-08-12
Add ruby-webrick as deb-pkg dependency for jammy/foreman-proxy

theQuestionmark:

### DIFF
--- a/debian/bookworm/foreman-proxy/control
+++ b/debian/bookworm/foreman-proxy/control
@@ -12,7 +12,7 @@ Package: foreman-proxy
 Section: net
 Priority: extra
 Architecture: all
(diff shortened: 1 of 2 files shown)


ekohl (changes requested):
We also have `bookworm` sources. They're blocked on a missing puppetserver package now, but can you also add the dependency there?

theQuestionmark:
Also added for bookworm... Debian 12 has a slightly higher versioning number for the pkgs as ubuntu 24.04...

ekohl (changes requested):
I just noticed the difference in version numbers and wonder why. In our gemspec we just have `~> 1.0` (https://github.com/theforeman/smart-proxy/blob/d287e3cc01698852c9f6ff85eac246263d01a5b3/smart_proxy.gemspec#L21) so I'd expect to match that unless there's a specific other reason.

theQuestionmark:
Altered the version number to v1.0.0 ... Background to my intention: even tho the initial package versions which where released for jammy and bookwork where v1.7.0 in 2021 as the most minimal version which can be installed through the official repository-servers. Some package version numbers are not reflecting the bundled gem version and are just pkg-release numbers. But I also understand your suggestion to keep the versioning between the gemspec and the version specified as package dependency in sync.

ekohl:
ok to test

evgeni (approved):

PR #11110, opened 2024-08-12
Use epel-7-x86_64 copr target for client repos

ekohl:
This was always a better option because it allows COPR to use its own EPEL mirror, which is faster. Now that EPEL 7 has been archived, this is the only way to build on EL7.

Still needs to address repoclosure if we actually want to maintain EL7 builds, which @evgeni would prefer with his downstream hat on.


ehelms (approved):

ekohl:
Opened https://github.com/theforeman/foreman-packaging/pull/11111 to see if this works.

PR #11116, opened 2024-08-13
Fix foreman_ygg_worker spec for suse

evgeni:
- **Revert "[TEMP] disable opensuse building of client"**
- **Fix foreman_ygg_worker spec for suse**


packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -12,17 +12,21 @@
 
 %if 0%{?rhel} > 7 && ! 0%{?fedora}
 %define gobuild(o:) \
-        go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -compressdwarf=false -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v %{?**};
+        GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -compressdwarf=false -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v %{?**};
```
ekohl, line 15: I was surprised this was all hand defined but I see we don't depend on `go-rpm-macros` (and a quick look confirms that it's not trivial). I vaguely recall that's because people didn't want to touch the difference between EL7 and newer. However, there I see it also turns the modules off so it looks correct.

ekohl (approved):

PR #11124, opened 2024-08-14
Add foreman_opennebula plugin

pyslarvt:


pyslarvt:
Broken dependencies. Don't know how to solve it yet.
```
    foreman_opennebula (= 2.3.0) was resolved to 2.3.0, which depends on
      fog-opennebula (~> 0.0) was resolved to 0.0.5, which depends on
        opennebula was resolved to 6.8.3, which depends on
          nokogiri (< 1.16)
```

ekohl:
I haven't checked, but it sounds like it either ships nokogiri >= 1.16 or no nokogiri at all.

plugins/ruby-foreman-opennebula/debian/control
```diff
@@ -0,0 +1,13 @@
+Source: ruby-foreman-opennebula
+Section: ruby
+Priority: extra
+Maintainer: IaaS team IVI.RU <ops@ivi.ru>
+Build-Depends: debhelper (>= 9), foreman-nulldb (>= 3.7), foreman-assets (>= 3.7), ruby-foreman-deface
```
ekohl, line 5: You don't use deface
```suggestion
Build-Depends: debhelper (>= 9), foreman-nulldb (>= 3.7), foreman-assets (>= 3.7)
```

pyslarvt: Yes. It was added as dependency because of the polyglot gem module. Now I included the previous version of polyglot gem to the package to avoid conflict with deface deb and removed this dependency.

pyslarvt:
I fixed it. Now it works as intended.

plugins/ruby-foreman-opennebula/debian/control
evgeni: deface is used: https://github.com/theforeman/foreman_opennebula/blob/7558b7aa7929e6249419c7bfa5bdf95a1bb056a4/app/overrides/subnets/_fields/opennebula_fields.html.erb.deface

evgeni (approved):

pyslarvt:
I found a bug related to creation of ssh key pair for image based provisioning method. Need some time to fix it.

evgeni:
@pyslarvt 4.0.1 is good to go? then I'd merge this and see that we do RPM too.

pyslarvt:
Yes. I fixed it. Now everything works good.

PR #11137, opened 2024-08-19
Add foreman_opennebula plugin

pyslarvt:


evgeni:
[test rpm]

evgeni:
@pyslarvt so RPM built now too.

Do you know if foreman_opennebula/fog-opennebula also works with `opennebula` 7.0.0 gem? 
I had to hack the 6.8.2 a bit to not break things (https://github.com/theforeman/foreman-packaging/pull/12526) and updating to 7.0 would avoid this hack.

pyslarvt:
> Do you know if foreman_opennebula/fog-opennebula also works with `opennebula` 7.0.0 gem? I had to hack the 6.8.2 a bit to not break things (#12526) and updating to 7.0 would avoid this hack.

I think so but let me check again.

evgeni (approved):

PR #11144, opened 2024-08-20
3.12 build system updates

pcreech:


evgeni (approved):

ekohl:
@evgeni shouldn't we also verify the candlepin and pulp GPG keys? I'd still prefer to get https://github.com/theforeman/foreman-packaging/pull/11128 merged and then fix it up as part of branching.

PR #11148, opened 2024-08-21
Update rubygem-foreman_supervisory_authority to 0.2.0

laugmanuel:


packages/plugins/rubygem-elastic-apm/rubygem-elastic-apm.spec
```diff
@@ -1,101 +1,82 @@
...
 %install
 mkdir -p %{buildroot}%{gem_dir}
-cp -pa .%{gem_dir}/* \
+cp -a .%{gem_dir}/* \
         %{buildroot}%{gem_dir}/
 
 %files
 %dir %{gem_instdir}
-%exclude %{gem_instdir}/.ci
+%{gem_instdir}/.ci
```
evgeni, line 49:
```suggestion
%exclude %{gem_instdir}/.ci
```

packages/plugins/rubygem-elastic-apm/rubygem-elastic-apm.spec
```diff
@@ -1,101 +1,82 @@
...
+%exclude %{gem_instdir}/.github
 %exclude %{gem_instdir}/.gitignore
-%exclude %{gem_instdir}/.hound.yml
+%{gem_instdir}/.pre-commit-config.yaml
```
evgeni, line 52:
```suggestion
%exclude %{gem_instdir}/.pre-commit-config.yaml
```

packages/plugins/rubygem-elastic-apm/rubygem-elastic-apm.spec
```diff
@@ -1,101 +1,82 @@
...
 %exclude %{gem_instdir}/.rubocop.yml
+%doc %{gem_instdir}/CODE_OF_CONDUCT.md
+%{gem_instdir}/Dockerfile
```
evgeni, line 55:
```suggestion
%exclude %{gem_instdir}/Dockerfile
```

packages/plugins/rubygem-elastic-apm/rubygem-elastic-apm.spec
laugmanuel: This is the result of `gem2rpm` with the default template.

packages/plugins/rubygem-elastic-apm/rubygem-elastic-apm.spec
laugmanuel: This is the result of `gem2rpm` with the default template.

packages/plugins/rubygem-elastic-apm/rubygem-elastic-apm.spec
laugmanuel: This is the result of `gem2rpm` with the default template.

packages/plugins/rubygem-elastic-apm/rubygem-elastic-apm.spec
evgeni: We can update that ;)

https://github.com/theforeman/foreman-packaging/blob/770f5380dde1ba4686db47ff2965a37c2570cd28/gem2rpm/default.spec.erb#L15-L42

laugmanuel: I've added the recommanded changes. This is probably because the `elastic-apm` gem isn't build by us and therefore doesn't comply with our expectations :wink:

laugmanuel:
This PR needs some more dependencies, we'll package soon. The version of the plugin **does** work, but causes some additional logging (or `elastic-apm` does).

laugmanuel:
depends on https://github.com/theforeman/foreman-packaging/pull/11157

evgeni (approved):

PR #11119, opened 2024-08-13
Add how to handle Ruby default gems to README

ehelms:

PR #11157, opened 2024-08-22
Add and Update foreman_supervisory_authority dependencies

ochnerd:
Add base64-0.2.0 as dependency for llhttp-ffi
Update http to 5.2.0
Update http-form_data to 2.3.0
Add llhttp-ffi-0.5.0 as dependency for http


dgoetz:
ok to test

ochnerd:
I used the script but my Mac kind of complaint about something. I check that and push again, thanks for testing!

ochnerd:
@dgoetz , should ./add_gem_package.sh add them to package_manifest.yaml?

dgoetz:
Yes, there is a function `add_to_manifest` which maps the repository to the corresponding section of the manifest and the functions `add_gem_to_all_comps` and `add_gem_to_comps` ensure packages are added to matching comps.
The manifest is used as base for obal (and all other tooling) resulting in the linting error.

dgoetz:
Repoclosure now fails with

```
Error: Repoclosure ended with unresolved dependencies (3) across 1 packages.
package: rubygem-http-5.2.0-1.el8.noarch from repo0
  unresolved deps (3):
    (rubygem(base64) >= 0.1 with rubygem(base64) < 1)
    (rubygem(http-form_data) >= 2.2 with rubygem(http-form_data) < 3)
    (rubygem(llhttp-ffi) >= 0.5.0 with rubygem(llhttp-ffi) < 0.6)
```

So with base64 being 0.2.0, llhttp-ffi 0.5.0, and http-form_data 2.3.0, this is a problem of order. Not sure if you need to split up the PR once more or having the packages in single commits in correct order would be enough (this is what I typically do).

ochnerd:
Ahh I see the problem. If I think about the order I will probably need to open multiple PR when I understand correctly the dependencies need to already be in the repository. So Single PRs it is. I would leace this PR Open, include the links to the new PRs and close it after the new PRs have been merged.

ochnerd:
base64 PR https://github.com/theforeman/foreman-packaging/pull/11164 has no dependencies

packages/plugins/rubygem-base64/rubygem-base64.spec
ekohl, line 1: Please don't add base64. https://github.com/theforeman/foreman-packaging/pull/11119 works at documenting how you should deal with it. I left other comments below with more specific examples.

packages/plugins/rubygem-http/rubygem-http.spec
```diff
@@ -38,52 +22,41 @@ method chaining system for building requests, similar to Python's Requests.
...
-%{?scl:scl enable %{scl} - << \EOF}
-gem unpack %{SOURCE0}
-%{?scl:EOF}
-
-%setup -q -D -T -n  %{gem_name}-%{version}
-
-%{?scl:scl enable %{scl} - << \EOF}
-gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
-%{?scl:EOF}
+%setup -q -n  %{gem_name}-%{version}
```
ekohl, line 33: And then drop the base64 dependency here:
```suggestion
%setup -q -n  %{gem_name}-%{version}

%gemspec_remove_dep -g base64 '~> 0.1'
```

packages/plugins/rubygem-http/rubygem-http.spec
```diff
@@ -38,52 +22,41 @@ method chaining system for building requests, similar to Python's Requests.
...
         %{buildroot}%{gem_dir}/
 
 %files
 %dir %{gem_instdir}
-%exclude %{gem_instdir}/.coveralls.yml
+%exclude %{gem_instdir}/.github
 %exclude %{gem_instdir}/.gitignore
 %exclude %{gem_instdir}/.rubocop.yml
-%exclude %{gem_instdir}/.travis.yml
+%{gem_instdir}/.rubocop
```
ekohl, line 53:
```suggestion
%exclude %{gem_instdir}/.rubocop
```

packages/plugins/rubygem-llhttp-ffi/rubygem-llhttp-ffi.spec
```diff
@@ -0,0 +1,92 @@
+# template: default
+#
+# PLEASE UPDATE WITH CARE!
+# The directory structure is very special and
+# does not really align with any other package
```
ekohl, line 5: I don't see what's so different. This looks fairly standard to me for a gem with native extensions.

comps/comps-foreman-el8.xml
ekohl, line 1: I'm not sure why it gets sorted very different. We try to set the locale for consistent sorting but perhaps you don't have that locale installed?
https://github.com/theforeman/foreman-packaging/blob/9aeba42bc76b99db734b9fef3f7c4fa0eb55c3ac/comps_doc.sh#L6

ekohl (changes requested):
I think you may be right we need all dependencies satisfied before repoclosure can work. I thought we grouped them, but now that I take a closer look we don't.

packages/plugins/rubygem-base64/rubygem-base64.spec
ochnerd: Because of the dependency Issues I split this PR into the other 3 (I should close this PR i think): 
https://github.com/theforeman/foreman-packaging/pull/11165
https://github.com/theforeman/foreman-packaging/pull/11166
https://github.com/theforeman/foreman-packaging/pull/11167

ochnerd:
Closed in favor for splitted PR 
https://github.com/theforeman/foreman-packaging/pull/11165
https://github.com/theforeman/foreman-packaging/pull/11166
https://github.com/theforeman/foreman-packaging/pull/11167

PR #11164, opened 2024-08-23
Add rubygem-base64 package

ochnerd:
Add base64 as a dependency for http 5.2.0


dgoetz:
ok to test

packages/plugins/rubygem-base64/rubygem-base64.spec
```diff
@@ -0,0 +1,61 @@
...
+%{gem_libdir}
+%exclude %{gem_cache}
+%{gem_spec}
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/README.md
+
+%changelog
+* Fri Aug 23 2024  0.2.0-1
```
dgoetz, line 59: changelog is missing name and optionally email

comps/comps-foreman-el8.xml
```diff
@@ -405,8 +405,8 @@
       <packagereq type="default">rubygem-google-apis-pubsub_v1-doc</packagereq>
       <packagereq type="default">rubygem-google-apis-sqladmin_v1beta4-doc</packagereq>
       <packagereq type="default">rubygem-google-apis-storage_v1-doc</packagereq>
-      <packagereq type="default">rubygem-googleauth-doc</packagereq>
```
dgoetz, line 408: We know this from the past, Mac does sorting a bit different then Linux, can you keep the order and only add the new package(s)?

dgoetz (changes requested):
Two changes required see review comments, except from this the built packages look good!

packages/plugins/rubygem-base64/rubygem-base64.spec
ochnerd: Added my name

comps/comps-foreman-el8.xml
ochnerd: Reverted sorting and only added my package.

dgoetz (approved):

ekohl:
In https://github.com/theforeman/foreman-packaging/pull/11097 we had the same discussion and we don't need this for now.

ekohl:
I opened the revert PR in https://github.com/theforeman/foreman-packaging/pull/11173.

PR #11165, opened 2024-08-23
Add rubygem-llhttp-ffi package

ochnerd:
Add llhttp-ffi 0.5.0 as dependency for http 5.2.0


comps/comps-foreman-plugins-el8.xml
ekohl, line 1: This needs to be el9 now

ekohl (changes requested):
Apologies for the long delay in reviewing.

comps/comps-foreman-plugins-el8.xml
ochnerd: Done, Thanks for Reviewing!

ekohl:
ok to test

ekohl:
> If we go that route then I think we can go another route:
> 
>     * package `rubygems-llhttp`

https://github.com/theforeman/foreman-packaging/pull/12008
 
>     * patch out the gem requirement in `http`

https://github.com/theforeman/foreman-packaging/pull/12009

>     * drop the now unused  `ffi-compiler` from our packaging

https://github.com/theforeman/foreman-packaging/pull/12010

ochnerd:
@ekohl thanks for the idea. It sounds better than using the ffi version.
I was swamped with the zeitwerk changes in our plugins and forgot about this issue.

PR #11166, opened 2024-08-23
Bump rubygem-http-form_data to 2.3.0-1

ochnerd:
Bump rubygem-http-form_data to 2.3.0-1 as dependency for http 5.2.0

PR #11167, opened 2024-08-23
Bump rubygem-http to 5.2.0-1

ochnerd:
Needs:
base64 PR https://github.com/theforeman/foreman-packaging/pull/11164 has no dependencies
llhttp-ffi PR https://github.com/theforeman/foreman-packaging/pull/11165 has dependencies which can be resolved
http-form_data PR https://github.com/theforeman/foreman-packaging/pull/11166 has no dependencies


ekohl (changes requested):
Please prefer `base64` from default gems on Ruby < 3.3, just like we did in https://github.com/theforeman/foreman-packaging/pull/10490.

ochnerd:
Oh ok, i will add the changes and use the default gems. But https://github.com/theforeman/foreman-packaging/pull/11166 and https://github.com/theforeman/foreman-packaging/pull/11165 need to be resolved first because they are needed.

ekohl:
> Oh ok, i will add the changes and use the default gems. But #11166 and #11165 need to be resolved first because they are needed.

If they aren't build dependencies (and in this case I think they aren't) you can include them in the same PR).

ochnerd:
@ekohl , can really include them in the same PR if it failed like in https://github.com/theforeman/foreman-packaging/pull/11157#issuecomment-2304953287 mentioned?

ekohl:
I think you may be right that we need all dependencies.

For some reason, CI didn't run at all on this PR so it looks like nothing is wrong.

packages/plugins/rubygem-http/rubygem-http.spec
```diff
@@ -1,35 +1,20 @@
...
-Requires: %{?scl_prefix}rubygem(http-form_data) < 3
-Requires: %{?scl_prefix}rubygem(http-cookie) >= 1.0
-Requires: %{?scl_prefix}rubygem(http-cookie) < 2
-Requires: %{?scl_prefix}rubygem(addressable) >= 2.3
-Requires: %{?scl_prefix}rubygem(addressable) < 3
-BuildRequires: %{?scl_prefix_ruby}ruby(release)
-BuildRequires: %{?scl_prefix_ruby}ruby >= 2.2
-BuildRequires: %{?scl_prefix_ruby}rubygems-devel
+Requires: ruby >= 2.6
+Requires: (rubygem(base64) or ruby-default-gems < 3.4)
```
ekohl, line 14: This should be outside of the `specfile generated dependencies` block because it's not generated. It'll get removed on the next update

ekohl (changes requested):

packages/plugins/rubygem-http/rubygem-http.spec
ochnerd: I moved it above the block. Is this ok or should it be somewhere else in the spec?

ekohl:
Obsoleted by https://github.com/theforeman/foreman-packaging/pull/12009

PR #11180, opened 2024-08-28
Add ownership of the git configuration file

ekohl:
This allows users to run git config commands without having to set up permissions. The config file is marked as noreplace to respect the user's content.

A `%ghost` may also work, but I'm not sure.

This came up in https://github.com/theforeman/foreman-documentation/pull/3243 and I think this can simplify the procedure for users.


packages/plugins/rubygem-foreman_templates/rubygem-foreman_templates.spec
```diff
@@ -84,6 +88,7 @@ cp -a .%{gem_dir}/* \
 %{foreman_assets_foreman}
 %{foreman_webpack_plugin}
 %{foreman_webpack_foreman}
+%attr(0600, foreman, foreman) %config(noreplace) %{foreman_dir}/.config/git/config
```
dgoetz, line 91: Would this leave us with %{foreman_dir}/.config/ and %{foreman_dir}/.config/git/ being not owned by this package or any other package? I think so what makes me think we need then to add two %dir directives for them.

adamruzicka: > Would this leave us with %{foreman_dir}/.config/ and %{foreman_dir}/.config/git/ being not owned by this package or any other package?

Looks like it

```
# dnf whatprovides ~foreman/.config/git/config
Last metadata expiration check: 0:08:59 ago on Thu Aug 29 14:23:44 2024.
rubygem-foreman_templates-9.5.1-2.fm3_13.el9.noarch : Template-syncing engine for Foreman
Repo        : @System
Matched from:
Filename    : /usr/share/foreman/.config/git/config

# dnf whatprovides ~foreman/.config/git/
Last metadata expiration check: 0:08:56 ago on Thu Aug 29 14:23:44 2024.
Error: No matches found. If searching for a file, try specifying the full path or using a wildcard prefix ("*/") at the beginning.

# dnf whatprovides /usr/share/foreman/.config
Last metadata expiration check: 0:09:45 ago on Thu Aug 29 14:23:44 2024.
Error: No matches found. If searching for a file, try specifying the full path or using a wildcard prefix ("*/") at the beginning.
```

ekohl: > Would this leave us with %{foreman_dir}/.config/ and %{foreman_dir}/.config/git/ being not owned by this package or any other package? I think so what makes me think we need then to add two %dir directives for them.

I thought about that when writing it, but wasn't quite sure. It reminds that https://github.com/theforeman/foreman-packaging/pull/9705 should be finished up as well.

ekohl:
If/When this is merged, we should submit a docs update to simplify the procedure.

PR #11181, opened 2024-08-29
Bump yggdrasil to 0.4.1

adamruzicka:
let's see what happens


adamruzicka:
Welp, el7 has `meson-0.55.1`, yggdrasil requires `>= 0.58.0` so even if I would sort out those failures, we'd lose el7. Gues that's a no-go

packages/client/yggdrasil/yggdrasil.spec
```diff
@@ -44,46 +44,31 @@ a receiving queue for instructions to be sent to the system via a broker.
...
-BUILDFLAGS="%buildflags" \
-make PREFIX=%{_prefix} \
-     SYSCONFDIR=%{_sysconfdir} \
-     LOCALSTATEDIR=%{_localstatedir} \
-     LIBEXECDIR=%{_libexecdir} \
-     SHORTNAME=%{name} \
-     LONGNAME=%{name} \
-     PKGNAME=%{name} \
-     VERSION=%{version}
+%meson "-Dgobuildflags=%buildflags" %ldflags
```
ekohl, line 47: Is it really correct to pass the ldflags to meson? Looks like all versions don't like that. https://docs.fedoraproject.org/en-US/packaging-guidelines/Meson/ suggests that normally it should be able to figure it out by default.

ekohl (commented):
yggdrasil 0.4.1 is in centos-stream 9 at least. If we don't make any modifications, I'd prefer to rely on that once RHEL 9.5 comes out. As for EL 7 & 8: I wouldn't mind keeping them on older versions.

In our package manifest and comps files we can tell the build system to only build for EL 7, 8 and OpenSUSE.

PR #11190, opened 2024-08-30
Refs #18539 - Introduce foreman_plugin_files in foreman.macros

ekohl:
This allows plugins to use `%files -f foreman_plugin_files` and automatically get the correct files section, based on the `foreman_bundlerd_file` and `foreman_precompile_plugin` macros (and their options).


ekohl:
True, but the idea I have is to and up with less logic in `gem2rpm/foreman_plugin.spec.erb` anyway. Combined with https://github.com/theforeman/foreman-packaging/pull/8414 I'd like to get rid of a lot of templating.

Perhaps it's better to separate that out, but I wanted to see. Friday always has time for a bit of experimentation, right?

packages/foreman/foreman/foreman.spec
```diff
@@ -678,6 +678,7 @@ cat > %{buildroot}%{_sysconfdir}/rpm/macros.%{name}-plugin << EOF
 # -n<plugin_name>   Overrides default of gem_name
 %%%{name}_bundlerd_file(n:) \\
 mkdir -p %%{buildroot}%%{%{name}_bundlerd_dir} \\
+echo %%{%{name}_bundlerd_dir}/%%{-n*}%%{!?-n:%%{gem_name}} >> %%%{name}_plugin_files \\
```
ekohl, line 681: Note to self: this and other cases need to be
```suggestion
echo %%{%{name}_bundlerd_dir}/%%{-n*}%%{!?-n:%%{gem_name}} >> %{name}_plugin_files \\
```

ekohl (commented):
I've become convinced this is getting too complex and we should really take it as a bigger step.

PR #11197, opened 2024-09-02
Use Puppet 8 for repoclosure

ekohl:
Now that Foreman supports this, it's good to run with this.

One big change is that puppet-agent 8.x no longer has a `Provides: puppet` line which otherwise we won't capture.

### DIFF
--- a/repoclosure/yum.conf
+++ b/repoclosure/yum.conf
@@ -28,9 +28,9 @@ name=CentOS Stream 9 - CRB
 metalink=https://mirrors.centos.org/metalink?repo=centos-crb-9-stream&arch=x86_64&protocol=https,http
 enabled=1
 
-[el9-puppet-7]
-name=el9-puppet-7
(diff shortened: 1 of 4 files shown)


repoclosure/yum.conf
```diff
@@ -28,9 +28,9 @@ name=CentOS Stream 9 - CRB
 metalink=https://mirrors.centos.org/metalink?repo=centos-crb-9-stream&arch=x86_64&protocol=https,http
 enabled=1
 
-[el9-puppet-7]
-name=el9-puppet-7
-baseurl=https://yum.puppetlabs.com/puppet7/el/9/x86_64/
+[el9-puppet-8]
```
evgeni, line 31: you also need to update package manifest :)
https://github.com/theforeman/foreman-packaging/blob/rpm/develop/package_manifest.yaml#L8
https://github.com/theforeman/foreman-packaging/blob/rpm/develop/package_manifest.yaml#L865

PR #11198, opened 2024-09-03
add satellite collections to packaging

evgeni:
this is mostly so that we can use packit for those collections


ehelms:
Why do these need packit?

evgeni:
Because I want to be able to test changes via PRT

evgeni:
I talked to Zach and he said that no changes are needed on the downstream side.

ehelms:
> Because I want to be able to test changes via PRT

I was hopeful the tests could just use the collection more directly and avoid all of this setup.

evgeni:
Nope. They assume a prod system, which obtains things via RPM.

PR #11216, opened 2024-09-06
Update ansible-runner to 2.3.6

evgeni:


evgeni:
Two things:
* I took 2.3.6, not 2.4.0 as that has a new buildsystem and I didn't want to have that fun just yet
* Once we drop EL8 we can switch runner back to system python (2.3 needs 3.8+, 2.4 needs 3.9+, which EL9 has) and we don't need the "same as ansible version" constraint anymore -- this also has the benefit that we need to package fewer things, only `daemon` is not available in EL9

ehelms (approved):

PR #11217, opened 2024-09-06
Fix build on amazonlinux

sbernhard:


evgeni:
can you bump revision / add changelog, so the thing actually gets built?

otherwise LGTM from far above

evgeni (approved):

PR #11222, opened 2024-09-10
Add fapolicyd restart to ensure rules get loaded

ehelms:
https://issues.redhat.com/browse/SAT-27834 provides some context and information that some users are running into this issue. This appears to be consistently reproducible now even though nothing appears to have changed with fapolicyd package.

### DIFF
--- a/packages/foreman/foreman-fapolicyd/foreman-fapolicyd.spec
+++ b/packages/foreman/foreman-fapolicyd/foreman-fapolicyd.spec
@@ -1,6 +1,6 @@
 Name:     foreman-fapolicyd
 Version:  1.0.1
-Release:  2%{?dist}
+Release:  3%{?dist}
 Summary:  Foreman fapolicyd rules
 
 Group:    System Environment/Base
@@ -11,6 +11,9 @@ Source0:  https://codeload.github.com/theforeman/%{name}/tar.gz/v%{version}#/%{n
 BuildArch: noarch
 
 Requires: fapolicyd
+Requires: systemd
+
+BuildRequires: systemd-rpm-macros
 
 %description
 Foreman fapolicyd rules
@@ -29,10 +32,13 @@ cp 61-foreman-proxy.rules %{buildroot}%{_sysconfdir}/fapolicyd/rules.d/61-forema
 %post
 
 %{_sbindir}/fagenrules --load
+%{_bindir}/systemctl try-restart fapolicyd.service
 
 %postun
 
 %{_sbindir}/fagenrules --load
+# loading fagenrules leaves fapolicyd.service in a dead state, restart to ensure
+%systemd_postun_with_restart fapolicyd.service
 
 %files
 %attr(0644,root,fapolicyd) %{_sysconfdir}/fapolicyd/rules.d/60-foreman.rules
@@ -49,15 +55,21 @@ Foreman Proxy fapolicyd rules
 %post -n foreman-proxy-fapolicyd
 
 %{_sbindir}/fagenrules --load
+%{_bindir}/systemctl try-restart fapolicyd.service
 
 %postun -n foreman-proxy-fapolicyd
 
 %{_sbindir}/fagenrules --load
+# loading fagenrules leaves fapolicyd.service in a dead state, restart to ensure
+%systemd_postun_with_restart fapolicyd.service
 
 %files -n foreman-proxy-fapolicyd
 %attr(0644,root,fapolicyd) %{_sysconfdir}/fapolicyd/rules.d/61-foreman-proxy.rules
 
 %changelog
+* Tue Sep 10 2024 Eric D. Helms <ericdhelms@gmail.com> - 1.0.1-3
+- Add fapolicyd restart to ensure rules get loaded
+
 * Thu Oct 19 2023 Evgeni Golov - 1.0.1-2
 - Drop the requires from foreman-policyd to foreman-proxy-policyd
 


packages/foreman/foreman-fapolicyd/foreman-fapolicyd.spec
```diff
@@ -28,11 +28,23 @@ cp 61-foreman-proxy.rules %{buildroot}%{_sysconfdir}/fapolicyd/rules.d/61-forema
...
 %post
 
-%{_sbindir}/fagenrules --load
+if %{_bindir}/systemctl is-active --quiet fapolicyd.service; then
+  %{_sbindir}/fagenrules --load
+  # loading fagenrules leaves fapolicyd.service in a dead state, restart to ensure
+  %{_bindir}/systemctl restart fapolicyd.service
+else
+  %{_sbindir}/fagenrules --load
+fi
```
ekohl, line 37: Looks like `%{_sbindir}/fagenrules --load` is still unconditional so we can simplify to:
```suggestion
%{_sbindir}/fagenrules --load
if %{_bindir}/systemctl is-active --quiet fapolicyd.service; then
  # loading fagenrules leaves fapolicyd.service in a dead state, restart to ensure
  %{_bindir}/systemctl restart fapolicyd.service
fi
```

Then for `is-active` and `restart` there's also `try-restart`

```suggestion
%{_sbindir}/fagenrules --load
# loading fagenrules leaves fapolicyd.service in a dead state, restart to ensure
%{_bindir}/systemctl try-restart fapolicyd.service
```

But then we don't benefit from modern helpers that only restart once in a single transaction that we have on modern EL (is that EL10+? At least it's already in Fedora). I have a feeling we should use `%systemd_postun_with_restart fapolicyd.service` instead

ekohl:
Any reason you didn't use `try-restart`?

ehelms:
> Any reason you didn't use `try-restart`?

I'm getting old and forgetful :laughing:

ekohl:
You shouldn't need `is-active` anymore. Quoting the man page:
```
       try-restart PATTERN...
           Stop and then start one or more units specified on the command line if the units are running. This does nothing if units are not running.
```

packages/foreman/foreman-fapolicyd/foreman-fapolicyd.spec
```diff
@@ -29,11 +32,17 @@ cp 61-foreman-proxy.rules %{buildroot}%{_sysconfdir}/fapolicyd/rules.d/61-forema
 %post
 
 %{_sbindir}/fagenrules --load
+%{_bindir}/systemctl try-restart fapolicyd.service
 
 %postun
 
 %{_sbindir}/fagenrules --load
 
+if %{_bindir}/systemctl is-active --quiet fapolicyd.service; then
```
ehelms, line 41: I can't tell if this is still needed or not for the postun? I would assume the macro handles it.

packages/foreman/foreman-fapolicyd/foreman-fapolicyd.spec
```diff
@@ -11,6 +11,9 @@ Source0:  https://codeload.github.com/theforeman/%{name}/tar.gz/v%{version}#/%{n
 BuildArch: noarch
 
 Requires: fapolicyd
+Requires: systemd
+
+BuildRequires: systemd
```
ekohl, line 16: I think this is sufficient
```suggestion
BuildRequires: systemd-rpm-macros
```

packages/foreman/foreman-fapolicyd/foreman-fapolicyd.spec
ekohl: On EL9:
```
%systemd_postun_with_restart() \
%{expand:%%{?__systemd_someargs_%#:%%__systemd_someargs_%# systemd_postun_with_restart}} \
if [ $1 -ge 1 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then \
    # Package upgrade, not uninstall \
    /usr/lib/systemd/systemd-update-helper mark-restart-system-units %{?*} || : \
fi \
%{nil}
```

From `/usr/lib/systemd/systemd-update-helper`
```sh
    mark-restart-system-units)
        [ -d /run/systemd/system ] || exit 0

        for unit in "$@"; do
            systemctl set-property "$unit" Markers=+needs-restart &
        done
        wait
        ;;
```

And I recall that at the end of the RPM transaction it will look for all running services with the `needs-restart` marker. This prevents multiple packages restarting the same service.

On EL8 we have the more classic implementation:
```
%systemd_postun_with_restart() \
if [ $1 -ge 1 ] ; then \
        # Package upgrade, not uninstall \
        systemctl try-restart %{?*} &>/dev/null || : \
fi \
%{nil}
```

So I think it's redundant.

packages/foreman/foreman-fapolicyd/foreman-fapolicyd.spec
```diff
@@ -29,10 +32,13 @@ cp 61-foreman-proxy.rules %{buildroot}%{_sysconfdir}/fapolicyd/rules.d/61-forema
 %post
 
 %{_sbindir}/fagenrules --load
+%{_bindir}/systemctl try-restart fapolicyd.service
 
 %postun
 
 %{_sbindir}/fagenrules --load
+# loading fagenrules leaves fapolicyd.service in a dead state, restart to ensure
+%systemd_postun_with_restart fapolicyd.service
```
evgeni, line 41: If it's "dead", will `try-restart` *actually* restart it? Or is it "dead" in terms of "the service doesn't work but systemd still shows it as active"?

ehelms: Great question - I believe the issue we are seeing, and what this solves is, `fapolicyd` the software is in a "dead" state, and does not represent the current set of rules accurately. And the restart will fix that.

ekohl (approved):

PR #11233, opened 2024-09-11
Refs #18539 - Expand foreman_{assets,webpack}_plugin macros

ekohl:
These macros now include the directories above it so plugins properly own them. That means on removal (and upgrade is a special case of removal) they're cleaned up.

This means the macro now only works in the %files section, but in practice that's the only place it's used.

### DIFF
--- a/packages/foreman/foreman/foreman.spec
+++ b/packages/foreman/foreman/foreman.spec
@@ -4,7 +4,7 @@
 %global dynflow_sidekiq_service_name dynflow-sidekiq@
 %global rake /usr/bin/rake
 
-%global release 1
+%global release 2
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


ekohl:
My thought was to fork the COPR repo that was created for this PR and build a plugin there.

ekohl:
https://copr.fedorainfracloud.org/coprs/ekohl/foreman-packaging-pr-11233/build/8009724/

ekohl:
I'm tempted to say this is a relevant build failure.
```
+ '[' -e /builddir/build/BUILDROOT/rubygem-foreman_remote_execution-14.0.0-1.fm3_13.el8.x86_64
/var/tmp/rpm-tmp.K20k5c: line 77: [: missing `]'


RPM build errors:
error: Bad exit status from /var/tmp/rpm-tmp.K20k5c (%install)
    Bad exit status from /var/tmp/rpm-tmp.K20k5c (%install)
```

evgeni:
Well, damn, that macro is used in the precompile_plugin macro.

🐢

PR #11254, opened 2024-09-16
Update rubygem-kafo to 7.5.0

evgeni:


ekohl (commented):
Should we release 7.5.1 to include `Protect parsing of data types from empty lines`?

evgeni:
> Should we release 7.5.1 to include `Protect parsing of data types from empty lines`?

We certainly can. I was trying to use `--checks-only` and when that failed realized 7.5 wasn't in nightly.

ekohl:
I released 7.5.1 now

PR #11261, opened 2024-09-18
add rex-cockpit for debian

evgeni:

### DIFF
--- a/plugins/ruby-foreman-remote-execution/debian/control
+++ b/plugins/ruby-foreman-remote-execution/debian/control
@@ -14,3 +14,10 @@ Depends: ${misc:Depends}, bundler, foreman (>= 3.8.0~rc1), ruby-foreman-tasks (>
   ruby-foreman-deface, ruby-dynflow (>= 1.0.2), ruby-dynflow (<< 2.0.0)
 Description: Foreman Remote Execution Plugin
   A plugin bringing remote execution to the Foreman, completing the config management functionality with remote management functionality
+
+Package: ruby-foreman-remote-execution-cockpit
+Architecture: all
+Depends: ${misc:Depends}, ruby-foreman-remote-execution (= ${binary:Version}), cockpit-ws
+Description: Foreman Remote Execution Plugin - Cockpit integration
+  This package contains files related to Cockpit, mainly foreman-cockpit service
+  and corresponding configuration files.
--- a/plugins/ruby-foreman-remote-execution/debian/ruby-foreman-remote-execution-cockpit.install
+++ b/plugins/ruby-foreman-remote-execution/debian/ruby-foreman-remote-execution-cockpit.install
@@ -1,0 +1,4 @@
+extra/cockpit/foreman-cockpit-session usr/sbin/
+extra/cockpit/foreman-cockpit.service lib/systemd/system/
+extra/cockpit/cockpit.conf etc/foreman/cockpit/
+extra/cockpit/foreman-cockpit-session.yml etc/foreman/cockpit/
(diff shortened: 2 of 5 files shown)


plugins/ruby-foreman-remote-execution/debian/ruby-foreman-remote-execution-cockpit.install
```diff
@@ -0,0 +1,4 @@
+extra/cockpit/foreman-cockpit-session usr/sbin/
+extra/cockpit/foreman-cockpit.service lib/systemd/system/
```
evgeni, line 2: The service file use `/usr/libexec/cockpit-ws`, while on Debian the executable is in `/usr/lib/cockpit/cockpit-ws`.

@adamruzicka @ekohl what do you think would be the best path here? Just patch the .service file in Debian packaging?

ekohl: I think `libexec` is more correct, but IIRC `libexec` isn't defined in the FHS. I think patching it on Debian is fine.

evgeni: At this point the FHS is pretty much dead, and Debian allows `libexec` in packages just fine. But given we're not packaging cockpit, we can't fix it

ekohl: I still think FHS is a good reference, but indeed it should receive an update that's probably not going to exist. Should we file a Debian bug and link to it when patching?

evgeni: I don't think it's necessary

plugins/ruby-foreman-remote-execution/debian/rules
```diff
@@ -15,3 +15,11 @@ PLUGIN_HAS_WEBPACK = true
 PLUGIN_HAS_PACKAGE_JSON = true
 
 include /usr/share/foreman/foreman-plugin.mk
+
+override_dh_auto_install:
+	cp -r ./usr/share/foreman/vendor/ruby/*/gems/foreman_remote_execution-*/extra ./
```
ekohl, line 20: I'm surprised we even ship `extra` to `/usr/share` I wouldn't expect it to because they're intended to be moved to different locations but still allow us to ship a simple gem. Not sure if you want to fix it here. For example by changing this:
```suggestion
	mv ./usr/share/foreman/vendor/ruby/*/gems/foreman_remote_execution-*/extra ./
```

evgeni:
[test deb]

plugins/ruby-foreman-remote-execution/debian/rules
evgeni: We, uh, don't.
That's `./usr`, not `/usr`.
The local copy of `/usr/share/foreman` that we use to build plugins in.
The deb ships the `.gem` and the wepack things.

evgeni:
[test deb]

plugins/ruby-foreman-remote-execution/debian/control
```diff
@@ -14,3 +14,10 @@ Depends: ${misc:Depends}, bundler, foreman (>= 3.8.0~rc1), ruby-foreman-tasks (>
   ruby-foreman-deface, ruby-dynflow (>= 1.0.2), ruby-dynflow (<< 2.0.0)
 Description: Foreman Remote Execution Plugin
   A plugin bringing remote execution to the Foreman, completing the config management functionality with remote management functionality
+
+Package: ruby-foreman-remote-execution-cockpit
+Architecture: all
+Depends: ${misc:Depends}, ruby-foreman-remote-execution, cockpit-ws
```
ekohl, line 20: Should this depend on the exact version of `ruby-foreman-remote-execution` to know for sure the service files match the service itself?

ekohl (approved):

PR #11272, opened 2024-09-19
Update rubygem-hammer_cli_foreman_rh_cloud to 1.0.3

chris1984:


chris1984:
@pcreech here is the translation pr for 6.16, will cherry-pick it in after this one goes in.

PR #11281, opened 2024-09-20
Update rubygem-foreman_rh_cloud to 10.0.2

chris1984:
We had to pin against katello 4.14 which was not GA yet with this [PR](https://github.com/theforeman/foreman_rh_cloud/pull/896) since we were pulling in an older version of Katello which was causing tests to fail and not having the convert2rhel Katello PR. 

Shim suggested we pin this way since we were still in RC mode for Katello 4.14 and he said this would pull in the pre release gems. 

Let me know if you want me to just remove it out of the specfile so it builds correctly, or we need a new gem with it either pinned in a better way or removed from the gemspec now that Katello 4.14.0 is GA


packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -30,7 +30,7 @@ BuildRequires: ruby < 4
 BuildRequires: rubygems-devel
 BuildRequires: rubygem(foreman_ansible)
 BuildRequires: rubygem(foreman-tasks)
-BuildRequires: rubygem(katello)
+BuildRequires: rubygem(katello) >= 4.14.0.rc1.1
```
evgeni, line 33:
```
No matching package to install: 'rubygem(katello) >= 4.14.0.rc1.1'
```

That's not an RPM version definition, but Ruby, and RPM gets confused by that. (`4.14.0.rc1.1` is "bigger" than `4.14.0` which we have).

chris1984: I am guessing I need to a make a new gem with it pinned correctly in the gem spec?

evgeni: Packaging as such doesn't look inside the gem.
So editing it here to just read `>= 4.14.0` should be perfectly fine.
(And then update it in the gem whenever you do the next release)

evgeni (approved):

PR #11296, opened 2024-09-25
Update foreman_ygg_worker to 0.3.0

adamruzicka:

### DIFF
--- a/packages/client/foreman_ygg_worker/foreman_ygg_worker-0.2.2.tar.gz
+++ b/packages/client/foreman_ygg_worker/foreman_ygg_worker-0.2.2.tar.gz
@@ -1,1 +1,0 @@
-../../../.git/annex/objects/0P/xg/SHA256E-s9760263--6059fd5f2ffb73254bea0026004faf9a7b0a20b8035955c41a43de7de155bb9a.tar.gz/SHA256E-s9760263--6059fd5f2ffb73254bea0026004faf9a7b0a20b8035955c41a43de7de155bb9a.tar.gz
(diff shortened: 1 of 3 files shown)


adamruzicka:
Please hold on until I get a confirmation that we should be creating the group ourselves.

evgeni:
@adamruzicka could you please rebase on the latest changes by Ewoud in rpm/develop?

adamruzicka:
Well, this is tricky. foreman_ygg_worker-0.3.0 is now compatible with yggdrasil 0.2.z and 0.4.z (not sure about 0.3.z, probably too but who knows). However, the way we deploy yggdrasil is not compatible with yggdrasil >=0.4 until https://github.com/theforeman/foreman-packaging/pull/11317 lands

adamruzicka:
[test rpm-copr]

pondrejk (approved):
ack from the functional side, tested together with https://github.com/theforeman/katello-pull-transport-migrate/pull/8

evgeni:
Packaging wise this looks fine.
What I am confused about is the user deployment implications.

* Users who want to deploy a fresh, mqtt connected host, just need `foreman_ygg_worker` and the config gets created by the process that is deploying `f_y_w`, right?
* Users who want to migrate an existing, katello-agent, host, install https://github.com/theforeman/katello-pull-transport-migrate/pull/8 and that does things for them.
* Users who already have a 0.2 `f_y_w`, *those* need https://github.com/theforeman/foreman-packaging/pull/11317?

adamruzicka:
In general yes to all three.

> Users who want to deploy a fresh, mqtt connected host, just need foreman_ygg_worker and the config gets created by the process that is deploying f_y_w, right?

Also yes, with "the process" being tweaked in https://github.com/theforeman/foreman/pull/10340

evgeni:
And you can't do the stuff from https://github.com/theforeman/foreman-packaging/pull/11317 in `%post` of `foreman_ygg_worker` because that'd make the setup incompatible with ygg 0.2?
My brain tries really hard to tell me a new package is (UX wise) a bad idea.

adamruzicka:
I need the post thingy to be run when yggdrasil itself gets updated to handle the service rename, which doesn't necessarily have to happen in the same transaction as when f_y_w is changed, so having it in f_y_w's post could miss this situation. The config file changes could be done at any time, but the service rename is tricky.

evgeni:
> My brain tries really hard to tell me a new package is (UX wise) a bad idea.

My main concern was that people need to install that package somehow, but in most cases (most: https://github.com/theforeman/foreman-packaging/pull/11317/files#r1822445063) this will happen automatically, so it's probably fine.

I still don't *like* this. But I also don't see a good way around it.

evgeni (approved):

PR #11297, opened 2024-09-25
add el9 config for mock

evgeni:


adamruzicka:
Works like a charm. Could el9 be made the default, considering el8 is effectively busted?

adamruzicka (approved):

PR #11317, opened 2024-10-04
Add foreman_ygg_migration package

adamruzicka:
This package should get pulled in when yggdrasil >=0.4.z is installed.

When installed, it checks whether yggdrasild.service is enabled. If not, it does nothing. If enabled, this package tweaks yggdrasil's config file to account for the newly needed config options. Additionally, it disables the yggdrasild service and enables yggdrasil.service.

Our worker is left to be activated through dbus.

### DIFF
--- a/packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
+++ b/packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
@@ -1,0 +1,33 @@
+Name: foreman_ygg_migration
+Version: 0.0.1
+Summary: A helper package to ease transition from yggdrasil 0.2.z to 0.4.z
+Release: 1%{?dist}
+License: MIT
+
+Supplements: yggdrasil >= 0.4.0
+Requires:    yggdrasil >= 0.4.0
+
+%description
+A helper package to ease transition from yggdrasil 0.2.z to 0.4.z
+
+%post
+if systemctl is-enabled yggdrasild >/dev/null 2>/dev/null; then
+	grep -Pq '^server'      /etc/yggdrasil/config.toml || sed -i 's/broker.*=/server =/' /etc/yggdrasil/config.toml
+	grep -Pq '^path-prefix' /etc/yggdrasil/config.toml || echo 'path-prefix = "yggdrasil"' >>/etc/yggdrasil/config.toml
+	grep -Pq '^data-host'   /etc/yggdrasil/config.toml || echo 'data-host = ""' >>/etc/yggdrasil/config.toml
+
+	if grep -q FOREMAN_YGG_WORKER_WORKDIR /etc/systemd/system/yggdrasild.service.d/* 2>/dev/null; then
+		mkdir -p /etc/systemd/system/com.redhat.Yggdrasil1.Worker1.foreman.service.d/
+		cp -r /etc/systemd/system/yggdrasild.service.d/* /etc/systemd/system/com.redhat.Yggdrasil1.Worker1.foreman.service.d/
+		systemctl daemon-reload
+	fi
+
+	systemctl disable --now yggdrasild
+	systemctl enable  --now yggdrasil
+fi
+
+%files
+
+%changelog
+* Thu Oct 03 2024 Adam Ruzicka <aruzicka@redhat.com> - 0.0.1-1
+- Initial packaging
--- a/packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
+++ b/packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
@@ -4,7 +4,9 @@ Summary: A helper package to ease transition from yggdrasil 0.2.z to 0.4.z
 Release: 1%{?dist}
 License: MIT
 
+%if 0%{?rhel} >= 8
 Supplements: yggdrasil >= 0.4.0
+%endif
 Requires:    yggdrasil >= 0.4.0
 
 %description
--- a/packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
+++ b/packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
@@ -7,7 +7,7 @@ License: MIT
 %if 0%{?rhel} >= 8
 Supplements: yggdrasil >= 0.4.0
 %endif
-Requires:    yggdrasil >= 0.4.0
+Conflicts:   yggdrasil  < 0.4.0
 
 %description
 A helper package to ease transition from yggdrasil 0.2.z to 0.4.z
--- a/comps/comps-foreman-client-el8.xml
+++ b/comps/comps-foreman-client-el8.xml
@@ -11,6 +11,7 @@
     <uservisible>true</uservisible>
     <packagelist>
       <packagereq type="default">foreman-client-release</packagereq>
+      <packagereq type="default">foreman_ygg_migration</packagereq>
       <packagereq type="default">foreman_ygg_worker</packagereq>
       <packagereq type="default">katello-host-tools</packagereq>
       <packagereq type="default">katello-host-tools-tracer</packagereq>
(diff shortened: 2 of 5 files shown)


packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
```diff
@@ -0,0 +1,34 @@
...
+
+%post
+if systemctl is-enabled yggdrasild >/dev/null 2>/dev/null; then
+	grep -Pq '^server'    /etc/yggdrasil/config.toml || sed -i 's/broker.*=/server =/' /etc/yggdrasil/config.toml
+	grep -Pq '^prefix'    /etc/yggdrasil/config.toml || echo 'prefix = "yggdrasil"' >>/etc/yggdrasil/config.toml
+	grep -Pq '^data-host' /etc/yggdrasil/config.toml || echo 'data-host = ""' >>/etc/yggdrasil/config.toml
+
+	if grep -q FOREMAN_YGG_WORKER_WORKDIR /etc/systemd/system/yggdrasild.service.d/* 2>/dev/null; then
+		mkdir -p /etc/systemd/system/com.redhat.Yggdrasil1.Worker1.foreman.service.d/
+		cp -r /etc/systemd/system/yggdrasild.service.d/* /etc/systemd/system/com.redhat.Yggdrasil1.Worker1.foreman.service.d/
```
ekohl, line 23: After this, you may need to run `systemctl daemon-reload`.

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
```diff
@@ -0,0 +1,34 @@
...
+	fi
+
+	systemctl disable --now yggdrasild
```
ekohl, line 26: Shouldn't the old service get uninstalled and thus removed?

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
```diff
@@ -0,0 +1,34 @@
...
+%endif
+Conflicts:   yggdrasil  < 0.4.0
+
+%description
+A helper package to ease transition from yggdrasil 0.2.z to 0.4.z
+
+%post
+if systemctl is-enabled yggdrasild >/dev/null 2>/dev/null; then
+	grep -Pq '^server'    /etc/yggdrasil/config.toml || sed -i 's/broker.*=/server =/' /etc/yggdrasil/config.toml
+	grep -Pq '^prefix'    /etc/yggdrasil/config.toml || echo 'prefix = "yggdrasil"' >>/etc/yggdrasil/config.toml
```
ekohl, line 18: This doesn't look idempotent. Perhaps `grep` should be stricter to check if the prefix isn't correct? Same below for `data-host`

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
```diff
@@ -0,0 +1,34 @@
...
+	grep -Pq '^data-host' /etc/yggdrasil/config.toml || echo 'data-host = ""' >>/etc/yggdrasil/config.toml
+
+	if grep -q FOREMAN_YGG_WORKER_WORKDIR /etc/systemd/system/yggdrasild.service.d/* 2>/dev/null; then
+		mkdir -p /etc/systemd/system/com.redhat.Yggdrasil1.Worker1.foreman.service.d/
+		cp -r /etc/systemd/system/yggdrasild.service.d/* /etc/systemd/system/com.redhat.Yggdrasil1.Worker1.foreman.service.d/
+	fi
+
+	systemctl disable --now yggdrasild
+	systemctl enable  --now yggdrasil
```
ekohl, line 27: It may need to restart `yggdrasil` if it's already running.

ekohl (commented):
Just gathering my thoughts on this.

This is because today we ship `packages/client/yggdrasil` but future EL versions will ship this instead. Those future EL versions are 0.4 which has a different architecture, but most importantly a different service name.

I'm worried that `if systemctl is-enabled yggdrasild` could be unreliable to detect this. I don't know exactly what the ordering will be, but perhaps the old service file is already cleaned up and systemd doesn't know about it anymore.

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
adamruzicka: That would mean that someone installed new-enough yggdrasil, started it and then installed the migration package, correct?

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
adamruzicka: I was operating under the assumption that the configuration was generated by foreman. That would mean that the `data-host` and `prefix` either aren't there at all or are generated correctly (after https://github.com/theforeman/foreman/pull/10340 gets in). I could tighten it down, but I'm not sure if that's worth it

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
ekohl: Yes, though I also wonder if that use case would be one to error out on. And you really shouldn't error out on RPM scriptlets. Still not entirely sure what's best here and currently my mind is exploring the various paths a user could take.

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
adamruzicka: From what I've seen, we're missing systemd scriptlets in our yggdrasil spec, so while the old package gets uninstalled and the `yggdrasild` service definition goes away, it seems to stay enabled.

ekohl:
> I tested it and it seemed to work, probably because of us not being diligent enough about service handling in the specfile of yggdrasil we ship.

IIRC with EL9 there were some serious changes how systemd RPM packaging worked. The operations are now much more batched than on EL8 with the use `/usr/lib/systemd/systemd-update-helper`. Perhaps we're benefiting from that?

adamruzicka:
[test rpm-copr]

pondrejk (approved):
ack from the functional side, tested together with https://github.com/theforeman/katello-pull-transport-migrate/pull/8

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
```diff
@@ -0,0 +1,34 @@
+Name: foreman_ygg_migration
+Version: 0.0.1
+Summary: A helper package to ease transition from yggdrasil 0.2.z to 0.4.z
+Release: 1%{?dist}
+License: MIT
+
+%if 0%{?rhel} >= 8
+Supplements: yggdrasil >= 0.4.0
```
evgeni, line 8: People can (and do :cry:) disable weak dependencies.
For such setups it will mean that this package won't be automatically installed once they upgrade to 0.4.

This is not a blocker (and there is no better package relationship to fix this), but something we should keep in mind and document somewhere.

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
```diff
@@ -0,0 +1,34 @@
+Name: foreman_ygg_migration
```
evgeni, line 1: I am debating with myself whether this needs to be an own package, vs a sub-package of `foreman_ygg_worker` (smth like `foreman_ygg_worker_migration` or so).

packages/client/foreman_ygg_migration/foreman_ygg_migration.spec
ehelms: What exactly should we document and where?

adamruzicka: We should document that if people have weak dependencies disabled, they should install this package explicitly. As to where, I'm not sure if we already have a procedure or something where this would fit well or we'll need to write a new one

evgeni: Yeah, I am not aware of a good place for this (maybe at least release notes?)

evgeni (approved):

PR #11332, opened 2024-10-09
Drop EL8 bits from Nightly

Odilhao:


ekohl (changes requested):
We have more files, like `modulemd/modulemd-foreman-el8.yaml` and `modulemd/modulemd-katello-el8.yaml`. Then there's the group of shell scripts that still default to el8 in many places:
* `add_npm_package.sh`
* `remove_from_comps`
* `add_to_comps.rb`
* `add_pypi_package.sh`
* `add_gem_package.sh`

Then we have a bit of a wildcard and that's `rhel8-kickstart-setup`. @evgeni or @ehelms know what we should do with this? It's following the upstream name, but should that be updated to rhel9 now?

ekohl:
I started https://github.com/theforeman/foreman-packaging/pull/11370 to make changing the comps files a bit easier.

Odilhao:
I had the `add_*` scripts modified here to drop EL8, I removed and force pushed again

adamruzicka:
Do we also want to keep mock config for el8 around? I know we've kept the one for el7, but is it still needed?

Odilhao:
> Do we also want to keep mock config for el8 around? I know we've kept the one for el7, but is it still needed?

Since we still have clients on EL8, I'm not sure if we need to drop mock now.

ekohl (approved):
We can deal with the add scripts in my PR. This PR now looks OK, but please fix up https://github.com/theforeman/jenkins-jobs/pull/495 so we can merge this.

Odilhao:
I forgot about https://github.com/theforeman/jenkins-jobs/pull/495 sorry, just pushed the change there.

PR #11340, opened 2024-10-14
add codeowners

evgeni:


evgeni:
the teams listed in this PR are added with "write" permissions

nadjaheitmann:
Thanks @evgeni Looking forward to thin is action! I think this also makes sense for foreman_puppet, foreman_salt, proxmox and foreman_kernel_care - just to name the ones that I can think of.

evgeni:
> I think this also makes sense for foreman_puppet, foreman_salt, proxmox and foreman_kernel_care - just to name the ones that I can think of.

Oh absolutely, just wanted to have something to start with :)

PR #11342, opened 2024-10-14
add codeowners

evgeni:


evgeni:
the teams listed in this PR are added with "write" permissions

PR #11369, opened 2024-10-22
drop focal from packaging

evgeni:


scripts/debcheck
```diff
@@ -20,7 +20,7 @@ PUPPET_VERSION=${PUPPET_VERSION:-6}
 # user repos TBD
 
 case ${DEBIAN_RELEASE} in
-  focal)
+  jammy)
```
adamruzicka, line 23: This came as a bit of a surprise, but I guess it makes sense

adamruzicka (approved):

scripts/debcheck
evgeni: Yeah, I don't think we use that script anywhere right now (but we should!), and it was never updated to support both focal *and* jammy… now it got updated at least a bit :)

PR #11370, opened 2024-10-22
Align add_{gem,npm,pypi}_package.sh on the adding to comps & manifest

ekohl:
This uses the exact same bash functions in all 3 scripts. It also changes the parameters a bit to always derive the comps files from the packaging directory. A major addition is that it now always adds to all the comps files instead of just a single one.

Right now this is untested.


add_gem_package.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 usage() {
-	echo "Usage: $0 GEM_NAME [TEMPLATE [PACKAGE_SUBDIR [REPO]]]"
+	echo "Usage: $0 GEM_NAME [TEMPLATE [PACKAGE_SUBDIR]]"
 	echo "Valid templates: $(ls gem2rpm | sed 's/.spec.erb//' | tr '\n' ' ')"
 	echo "Valid repos: foreman-el8 foreman-plugins-el8 katello-el8 foreman-client-el8"
```
ehelms, line 6: This line needs to go.

ekohl: Argh, I read over it a few times and first wanted to drop the actual usage. Then forgot it.

add_gem_package.sh
```diff
@@ -167,7 +158,7 @@ else
 	UPDATE=false
 fi
 
-if [[ -z $GEM_NAME ]] || [[ -z $TEMPLATE_NAME ]] || [[ $UPDATE != true ]] && [[ -z $REPO ]]; then
+if [[ -z $GEM_NAME ]] || [[ -z $TEMPLATE_NAME ]] || [[ $UPDATE != true ]]; then
```
ehelms, line 161: This breaks now when adding a new package because `UPDATE=false` and then `false != true` is always true.

ekohl:
I've started on a slightly different approach where I first introduce a common script to add to all comps files. Then the last commit only focuses on the package manifest addition. Still untested and a rough version, but pushing so it's at no longer just on my desktop.

The first commit still isn't quite complete and I probably want to do some more refactoring before wrapping this up.

add_pypi_package.sh
```diff
@@ -72,40 +88,20 @@ generate_pypi_package() {
...
-		local section="foreman_plugin_packages"
-	elif [[ $REPO == "katello-el8" ]] ; then
-		local section="katello_packages"
-	else
-		# TODO: client packages
-		local section=""
-	fi
+add_to_package_manifest() {
+  local package="${PACKAGE_NAME}"
+  local section="${MANIFEST_SECTION}"
```
ehelms, line 97: We have a weird mix of two spaces and I think 8 spaces throughout this file, this method now highlights that.

add_spec_to_comps
```diff
@@ -0,0 +1,50 @@
...
+	plugins)
+		COMPS_PREFIX=foreman-plugins-
+		;;
+	*)
+		COMPS_PREFIX=
+		;;
+esac
+
+if [[ -z $COMPS_PREFIX ]] ; then
+	echo "Unable to determine package directory for '$SPEC'" >&2
```
ehelms, line 35: I am running into this error message testing it:

```
 ./add_gem_package.sh bootsnap default foreman
~/workspace/upstream/packaging/foreman-packaging/packages/foreman/rubygem-bootsnap ~/workspace/upstream/packaging/foreman-packaging
add bootsnap-1.18.4.gem 
ok                                
(recording state in git...)
~/workspace/upstream/packaging/foreman-packaging
Unable to determine package directory for 'rubygem-bootsnap.spec'
Guessed '.'
```

add_pypi_package.sh
ekohl: Fixed

add_spec_to_comps
ekohl: I just pushed without addressing this, but having it pushed makes it safer to not lose work.

add_gem_package.sh
```diff
@@ -138,22 +105,31 @@ if [[ -z $TEMPLATE_NAME ]] ; then
...
 	fi
 fi
 
+if [[ -z $MANIFEST_SECTION ]] ; then
+	case $PACKAGE_SUBDIR in
+		client)
+			MANIFEST_SECTION="foreman_client_packages"
+			;;
+		foreman)
+			MANIFEST_SECTION="foreman_core_packages"
```
ekohl, line 122: I copied this from the old method, but I don't think this is a very good group for it. On the other hand, I don't think there is a sane default which is why it's now an optional parameter.

add_spec_to_comps
ekohl: Fixed by passing in the absolute path instead of just the filename. Looks like in some files we have `SPEC_FILE` as a relative value while in others it's the absolute path.

ekohl:
I now think this is ready. You may be able to note that I'd like to merge the various scripts into one if possible, but at the very least make sure they mostly work the same.

ehelms (approved):

PR #11462, opened 2024-11-08
Use Pulpcore 3.63 for Katello 4.15

ianballou:
Katello 4.15 / Foreman 3.13 should use Pulpcore 3.63.


Odilhao (changes requested):
We need to bump the version as well.

Odilhao (approved):

PR #11490, opened 2024-11-19
Update rubygem-angular-rails-templates to 1.2.2

ekohl:


packages/plugins/rubygem-angular-rails-templates/rubygem-angular-rails-templates.spec
```diff
@@ -1,80 +1,65 @@
...
-%{?scl:%scl_package rubygem-%{gem_name}}
-%{!?scl:%global pkg_name %{name}}
-
+# template: default
 %global gem_name angular-rails-templates
 
-Name:      %{?scl_prefix}rubygem-%{gem_name}
-Version:   1.1.0
-Release:   2%{?dist}
-Epoch:     1
```
evgeni, line 12: You lost the epoch when regenerating

evgeni (changes requested):

packages/plugins/rubygem-angular-rails-templates/rubygem-angular-rails-templates.spec
ekohl: Now I'm wondering why 2db9adc930e7bb5a6fc802df0ff3933e410205b3 added an epoch in the first place.

evgeni: I am not sure I wanna know

packages/plugins/rubygem-angular-rails-templates/rubygem-angular-rails-templates.spec
```diff
@@ -1,80 +1,66 @@
...
 
 %description
-Adds your HTML templates into Angular's $templateCache using Rails asset pipeline.
-It removes the need for AJAX calls to retrieve the templates (or for you to manually set them into the DOM).
+Use your angular templates with rails' asset pipeline.
+
 
 %package doc
-BuildArch:  noarch
-Requires:   %{name} = %{epoch}:%{version}-%{release}
```
evgeni, line 42: Another lost epoch

packages/plugins/rubygem-angular-rails-templates/rubygem-angular-rails-templates.spec
evgeni: https://github.com/theforeman/foreman-packaging/pull/532 actually explains it rather nicely.

evgeni (approved):

PR #11493, opened 2024-11-21
Fixes #38034 - Drop EL7 client support

ekohl:
This is a work in progress to remove EL7 client support.

It removes the tooling (and we'll need to drop the chroot manually).

Then katello-host-tools.spec is greatly simplified to drop a lot of conditionals. After that the tracer package is no longer needed and can be dropped.


ekohl:
I have reduced the scope to just dropping EL7, moving cleanup to https://github.com/theforeman/foreman-packaging/pull/11632.

adamruzicka (approved):

adamruzicka:
Thank you @ekohl !

ekohl:
Further work:
* [x] Remove EL7 from https://copr.fedorainfracloud.org/coprs/g/theforeman/client-nightly-staging/
* [ ] https://github.com/theforeman/theforeman-rel-eng/pull/465
* [ ] https://github.com/theforeman/jenkins-jobs/pull/531
* [ ] documentation

PR #11502, opened 2024-11-22
Drop tracer from foreman-client-el{8,9}

ekohl:
Since 626307d7f2dc727a873924d34a651aa157932758 tracer support is built into katello-host-tools for EL8+. This means it's only used on EL7.


evgeni (approved):

ekohl:
I considered complicated stuff with buildroots, but https://github.com/theforeman/foreman-packaging/pull/11493 will drop it altogether so that'll clean it up entirely.

PR #11516, opened 2024-11-25
Update rubygem-foreman_rh_cloud to 11.0.3

chris1984:
Not sure why the release job didn't run and do it for me, but here we are to get a new release into stream builds.


evgeni:
> Not sure why the release job didn't run and do it for me, but here we are to get a new release into stream builds.

You didn't tag 11.0.3 :)

evgeni:
[test rpm-copr]

evgeni (approved):

PR #11517, opened 2024-11-26
foreman: Update GEM and NPM dependencies

evgeni:


evgeni:
> I'm glad to see js-cookie being added there, but can't really speak to the rest

it's auto-generated, so I mostly trust it :D

PR #11562, opened 2024-12-06
Use copr:// links for external repositories

ekohl:
This has the benefit that COPR internally knows where to download it from. We don't need to repeat the specific buildroot name.

My long term goal is to have less configuration for the buildroots so it becomes easier to manage multiple. Ideally we support multiple architectures with minimal configuration.

At this moment it's untested, but I'd like to experiment with this. In fact, we should probably use https://fedora-copr.github.io/posts/runtime-dependencies and enhance obal to manage that.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -27,7 +27,7 @@ copr_projects:
     rhel_9: '9'
     rhel_8: '8'
     puppet_baseurl: "https://yum.voxpupuli.org/openvox{{ puppet_version }}"
-    root_repo_url: "https://download.copr.fedorainfracloud.org/results/{{ copr_project_user }}"
+    root_repo_url: "copr://{{ copr_project_user }}"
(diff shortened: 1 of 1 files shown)


ehelms:
That blog post is unclear to me if it translates runtime dependencies into the buildroot dependencies. Or if it translates buildroot dependencies that same way when using `copr://` -- have you been able to test this idea out on a repo?

ekohl:
No, I started to do that but it's the first time I used obal in a long time and immediately ran into https://github.com/theforeman/obsah/pull/37 again. It's https://www.youtube.com/watch?v=AbSehcT19u0 all over again.

ogajduse (approved):
I invested some time in investigating the `copr://` scheme, and I don't see an issue with using it. 

@ekohl Would you like to create an issue in the obal project for the enhancement you mentioned in the PR description?

ekohl:
> @ekohl Would you like to create an issue in the obal project for the enhancement you mentioned in the PR description?

I can file that but, that is a further optimization.

PR #11588, opened 2024-12-11
don't publish katello-pull-transport-migrate until ygg_worker is built

evgeni:


comps/comps-foreman-client-el10.xml
```diff
@@ -14,7 +14,7 @@
       <packagereq type="default">foreman_ygg_worker</packagereq>
```
ekohl, line 14: Should we drop this too just so we are clear in what we publish and don't publish?

ekohl (approved):
I'd also have accepted removal, but I'm fine with this too.

PR #11590, opened 2024-12-13
Update NodeJS used for package bumps to 18

evgeni:
NodeJS 18 is what we use to build packages, so we also should generate
bumps using that. Otherwise the cache generated doesn't work and the
packages fail to build.

### DIFF
--- a/.github/workflows/bump_packages.yml
+++ b/.github/workflows/bump_packages.yml
@@ -48,7 +48,7 @@ jobs:
       if: ${{ matrix.npm_name != null }}
       uses: actions/setup-node@v4
       with:
-        node-version: 12
+        node-version: 18
     - name: Install npm2rpm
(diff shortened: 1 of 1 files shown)


adamruzicka (commented):
Makes sense. Can we trigger a bump with this manually or do we have to merge it and hope?

evgeni:
> Makes sense. Can we trigger a bump with this manually or do we have to merge it and hope?

I can run a bump based on that branch, yes.

evgeni:
https://github.com/theforeman/foreman-packaging/actions/runs/12314402997

evgeni:
https://github.com/theforeman/foreman-packaging/pull/11260 was regenerated by :point_up: and is now green

ekohl (approved):

PR #11609, opened 2025-01-07
Add rubygem-hammer_cli_foreman_resource_quota package

nadjaheitmann:


package_manifest.yaml
```diff
@@ -640,6 +640,7 @@ foreman_plugin_packages_tier2:
     rubygem-foreman_webhooks: {}
     rubygem-foreman_vmwareannotations: {}
     rubygem-foreman_wreckingball: {}
+    rubygem-hammer_cli_foreman_resource_quota: {}
```
evgeni, line 643: This should be in `hammer_plugin_packages`

evgeni (changes requested):
ACK once the entry in the package manifest is fixed :)

package_manifest.yaml
nadjaheitmann: Thanks @evgeni , changed. I was fiddling around with the add_gem_package script for a while :upside_down_face:

evgeni (approved):

PR #11630, opened 2025-01-16
foreman_ygg_worker on el10

adamruzicka:
Forcing a rebuild to see what it actually fails with. The logs from https://github.com/theforeman/foreman-packaging/pull/11532#discussion_r1871018833 don't exist anymore

### DIFF
--- a/packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
+++ b/packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
@@ -5,35 +5,25 @@
 %global repo_orgname theforeman
 %global repo_name foreman_ygg_worker
 %global yggdrasil_libexecdir %{_libexecdir}/yggdrasil
-%{!?_root_sysconfdir:%global _root_sysconfdir %{_sysconfdir}}
-%global yggdrasil_worker_conf_dir %{_root_sysconfdir}/yggdrasil/workers
+%global yggdrasil_worker_conf_dir %{_sysconfdir}/yggdrasil/workers
 
 %global goipath         github.com/%{repo_orgname}/%{repo_name}
 
-%if 0%{?rhel} > 7 && ! 0%{?fedora}
-%define gobuild(o:) \
-        GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -compressdwarf=false -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v %{?**};
-%else
-%if ! 0%{?gobuild:1}
 %if 0%{?suse_version}
 %define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now '" -a -v %{?**};
-%else
-%define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '" -a -v %{?**};
-%endif
-%endif
 %endif
 
 Name: foreman_ygg_worker
 Version: 0.3.0
 Summary: Worker service for yggdrasil that can act as pull client for Foreman Remote Execution
-Release: 1%{?dist}
+Release: 2%{?dist}
 License: MIT
 
 Source0: https://github.com/%{repo_orgname}/%{repo_name}/releases/download/v%{version}/%{repo_name}-%{version}.tar.gz
 Url: https://github.com/%{repo_orgname}/%{repo_name}/
 
-# EL7 doesn't define go_arches
-%if ! 0%{?go_arches:1}
+# suse leap 15.5 doesn't define go_arches
+%if 0%{?suse_version}
 %define go_arches %{ix86} x86_64 %{arm} aarch64 ppc64le
 %endif
 ExclusiveArch: %{go_arches}
@@ -41,17 +31,19 @@ ExclusiveArch: %{go_arches}
 BuildRequires: systemd-rpm-macros
 %if 0%{?suse_version}
 BuildRequires: go
+BuildRequires: golang-packaging
 %else
 BuildRequires: golang
+%if 0%{?rhel} >= 9 || 0%{?fedora}
+BuildRequires: go-rpm-macros
+%else
+%if 0%{?rhel} == 8
+BuildRequires: go-srpm-macros
+%endif
+%endif
 %endif
 
-# Use rich dependencies if available
-%if 0%{?rhel} >= 8 || 0%{?fedora}
 Requires: (yggdrasil >= 0.2 with yggdrasil < 0.5)
-%else
-Requires: yggdrasil >= 0.2
-Conflicts: yggdrasil >= 0.5
-%endif
 
 %description
 Worker service for yggdrasil that can act as pull client for Foreman Remote Execution.
@@ -66,7 +58,7 @@ ln -fs $(pwd)/vendor _gopath/src/%{name}-%{version}/vendor
 ln -fs $(pwd)/go.mod _gopath/src/%{name}-%{version}/go.mod
 export GOPATH=$(pwd)/_gopath
 pushd _gopath/src/%{name}-%{version}
-%{gobuild}
+%gobuild -o %{name}-%{version}
 strip %{name}-%{version}
 popd
 make data LIBEXECDIR=%{yggdrasil_libexecdir}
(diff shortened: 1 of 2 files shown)


ekohl:
This reminded me that you wanted me to clean up EL7 so I trimmed https://github.com/theforeman/foreman-packaging/pull/11493 down to just that.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
ekohl, line 1: You can also drop `_root_sysconfdir` which was needed because of SCL.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -44,6 +41,9 @@ BuildRequires: go
 %else
 BuildRequires: golang
 %endif
+%if 0%{?rhel} > 8 || 0%{?fedora}
```
ekohl, line 44: I'm not a fan of `>`. Perhaps it's my mind, but I always find this easier so I don't need to do the + 1 in my head
```suggestion
%if 0%{?rhel} >= 9 || 0%{?fedora}
```

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -44,6 +41,9 @@ BuildRequires: go
...
+BuildRequires: go-rpm-macros
+%endif
 
 # Use rich dependencies if available
 %if 0%{?rhel} >= 8 || 0%{?fedora}
```
ekohl, line 49: I debated if this was always available. At first I wrote a comment you can but now I realized there's also suse. We build it on 15 and I think that has them too. Have you tried that?

adamruzicka: Let's see how it goes

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -44,14 +40,11 @@ BuildRequires: go
 %else
 BuildRequires: golang
 %endif
+%if 0%{?rhel} >= 9 || 0%{?fedora}
+BuildRequires: go-rpm-macros
```
ekohl, line 44: EL8 has `go-srpm-macros` and that defines `gobuild`

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -5,19 +5,15 @@
...
-%define gobuild(o:) \
-        GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -compressdwarf=false -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v %{?**};
-%else
 %if ! 0%{?gobuild:1}
 %if 0%{?suse_version}
 %define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now '" -a -v %{?**};
 %else
+%if 0%{?rhel} == 8
 %define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '" -a -v %{?**};
 %endif
```
ekohl, line 18: As noted below, on EL8 it's defined in `go-srpm-macros`.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
adamruzicka: Looks like it is supported on everything we currently build for

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -5,19 +5,15 @@
...
 
 %global goipath         github.com/%{repo_orgname}/%{repo_name}
 
-%if 0%{?rhel} > 7 && ! 0%{?fedora}
-%define gobuild(o:) \
-        GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -compressdwarf=false -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v %{?**};
-%else
 %if ! 0%{?gobuild:1}
 %if 0%{?suse_version}
 %define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -linkmode=external -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now '" -a -v %{?**};
```
ekohl, line 14: https://en.opensuse.org/openSUSE:Packaging_Go#Old_style_golang_applications has an example that requires `golang-packaging` and then they use `gobuild` from that but when I try it out on a SLES image I don't see the package.

adamruzicka: I see it on opensuse/leap:15.6 . Is that good enough or are we targeting sles specifically?

ekohl: We have that for ATIX so someone from @theforeman/atix should probably weigh in.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
ekohl: I think ATIX builds on SLES 12 as well so perhaps that's an issue, but I'll leave it for them to comment on.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -5,34 +5,20 @@
...
 License: MIT
 
 Source0: https://github.com/%{repo_orgname}/%{repo_name}/releases/download/v%{version}/%{repo_name}-%{version}.tar.gz
 Url: https://github.com/%{repo_orgname}/%{repo_name}/
 
-# EL7 doesn't define go_arches
+# suse leap 15.5 doesn't define go_arches
 %if ! 0%{?go_arches:1}
 %define go_arches %{ix86} x86_64 %{arm} aarch64 ppc64le
 %endif
```
ekohl, line 24: `go_arches` is defined if you have `golang-packaging` installed according to what I read:
```suggestion
```

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
adamruzicka: I'm having a hard time figuring out how to do it right for suse. There seems to be a `gobuild` macro, but semantics of its use seem to be different from what we have in EL*. Unless anyone knows how to make this work, I'd leave it alone for now.

ekohl (commented):
This is starting to look cleaner and cleaner.

ekohl:
I see the build fail but I can't spot the error message why it failed.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
adamruzicka: No luck it seems
```
Start: rpmbuild foreman_ygg_worker-0.3.0-2.suse.lp155.src.rpm
error: Building target platforms: x86_64
Building for target x86_64
Architecture is not included: x86_64
```

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -5,53 +5,37 @@
...
 License: MIT
 
 Source0: https://github.com/%{repo_orgname}/%{repo_name}/releases/download/v%{version}/%{repo_name}-%{version}.tar.gz
 Url: https://github.com/%{repo_orgname}/%{repo_name}/
 
-# EL7 doesn't define go_arches
-%if ! 0%{?go_arches:1}
-%define go_arches %{ix86} x86_64 %{arm} aarch64 ppc64le
-%endif
 ExclusiveArch: %{go_arches}
```
ehelms, line 21: Opensuse does not define `go_arches` from what I can tell, they define `go_exclusivearches`:

```
%go_exclusivearch \
ExclusiveArch: aarch64 %ix86 x86_64 %arm ppc64 ppc64le s390x riscv64
```
I would try the following:

```
%if 0%{?suse_version}
%go_exclusivearch
%else
ExclusiveArch: %{go_arches}
%endif

Comment (ekohl) [reply packages/client/foreman_ygg_worker/foreman_ygg_worker.spec:21]: That's not too bad. Overall I think the spec is already a lot cleaner

Comment (adamruzicka) [reply packages/client/foreman_ygg_worker/foreman_ygg_worker.spec:21]: > I would try the following:

I think I tried that, but let's give that one more try

Comment (adamruzicka) [reply packages/client/foreman_ygg_worker/foreman_ygg_worker.spec:25]: Hrm, doesn't look like it
```
Start: buildsrpm
Start: rpmbuild -bs
error: line 23: Unknown tag: %{go_exclusivearch}
Building target platforms: x86_64
Building for target x86_64
Finish: rpmbuild -bs
```
https://download.copr.fedorainfracloud.org/results/@theforeman/client-nightly-staging-scratch-7d545e01-9b9c-5fae-a7e1-414cc8b054c9/opensuse-leap-15.5-x86_64/08549528-foreman_ygg_worker/builder-live.log.gz

ekohl (approved):
This looks good to me. If you could rebase to resolve the merge conflict and squash then I think we can proceed.

ekohl:
Perhaps I'm blind, but I don't see anything useful on the OpenSUSE failed now.

adamruzicka:
Not sure what I did there, but it's green now

ekohl:
Then I'll merge it quickly before it turns red again ;)

PR #11631, opened 2025-01-16
add el10 mock config

evgeni:


mock/el10.cfg
```diff
@@ -0,0 +1,90 @@
...
+config_opts['package_manager'] = 'dnf'
+config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
+config_opts['description'] = 'CentOS Stream 10'
+
+config_opts['bootstrap_image'] = 'quay.io/centos/centos:stream10'
+config_opts['bootstrap_image_ready'] = True
+
+#config_opts['module_setup_commands'] = [
+#  ('enable',  'foreman-devel:el10'),
+#]
```
evgeni, line 18: doesn't exist for EL10 yet

ekohl (commented):
I thought about using more of the approach taken with includes. For example, `/etc/mock/centos-stream+epel-10-x86_64.cfg` is
```ini
config_opts["koji_primary_repo"] = "epel"
include('templates/centos-stream-10.tpl')
include('templates/epel-10.tpl')

config_opts['root'] = 'centos-stream+epel-10-x86_64'
config_opts['description'] = 'CentOS Stream 10 + EPEL'
config_opts['target_arch'] = 'x86_64'
config_opts['legal_host_arches'] = ('x86_64',)
```

It does rely on mock configs being up to date, but I think that's safe to rely on. Looking at `/etc/mock/templates/epel-10.tpl` we can then use `config_opts['dnf.conf'] += """` to add our own repos.

evgeni:
the templates is where I got the stuff from yeah…

ekohl:
Let me phrase it different: I'd like to minimize our mock configs. Perhaps even split it so we can layer them, just like in real life. So a config for client bits (which is probably plain EL10), a config for base foreman + plugins, and katello.

ekohl:
@evgeni what do we do with this?

evgeni:
@ekohl updated to use a template

PR #11634, opened 2025-01-16
Add support for devcontainers

adamruzicka:
It would be nice if this could get us to a state where those using vscode (or any of the other devcontainer-related tooling) could just open the repo, let the tooling crunch for a bit and then be able to do all the packaging actions without having to fiddle with anything.

TODOs:
- [x] Add obal, mock and friends. In general get this to a feature parity with what you get when you run the [rpm_packaging role](https://github.com/theforeman/forklift/blob/master/roles/rpm_packaging/tasks/main.yml) from forklift
- [x] some comments about how to make this work with podman instead of docker

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -13,7 +13,7 @@ RUN echo "[git-annex]" >> /etc/yum.repos.d/git-annex.repo \
 RUN dnf -y module enable nodejs:18
 
 RUN dnf install -y epel-release && \
-    dnf -y install nodejs vim git rpmdevtools git-annex-standalone wget ruby jq ruby-devel make gcc-c++ mock postgresql-devel libxml2-devel libcurl-devel systemd-devel \
+    dnf -y install nodejs vim git rpmdevtools git-annex-standalone wget ruby jq ruby-devel make gcc-c++ mock postgresql-devel libxml2-devel libcurl-devel systemd-devel rpmlint \
     python3 python3-pip python3-ruamel-yaml python3-requests python3-packaging
 
 RUN npm install npm2rpm --global
--- a/Containerfile
+++ b/Containerfile
@@ -17,5 +17,7 @@ RUN dnf -y install nodejs vim git rpmdevtools git-annex-standalone wget ruby jq 
 
 RUN npm install npm2rpm --global
 
+RUN gem install gem2rpm
+
 RUN mkdir -p /opt/foreman-packaging
 WORKDIR /opt/foreman-packaging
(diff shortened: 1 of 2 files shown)


ekohl:
> TODOs:
> 
>     * [ ]  Add obal, mock and friends. In general get this to a feature parity with what you get when you run the [rpm_packaging role](https://github.com/theforeman/forklift/blob/master/roles/rpm_packaging/tasks/main.yml) from forklift

I have https://copr.fedorainfracloud.org/coprs/ekohl/obal/ and think I could publish for EL as well. I used to do that for EL7 but never added EL 8 or 9. Would that make it easier?

Though I see the last tag was over a year ago so we should probably create a new release again.

adamruzicka:
We already have the how and the necessary dependencies captured in forklift, but having a copr repo would still make it a little bit easier, although it is not strictly necessary.

Containerfile
```diff
@@ -10,12 +10,15 @@ RUN echo "[git-annex]" >> /etc/yum.repos.d/git-annex.repo \
...
+RUN dnf install -y epel-release && \
+    dnf -y install nodejs vim git rpmdevtools git-annex-standalone wget ruby jq ruby-devel make gcc-c++ mock postgresql-devel libxml2-devel libcurl-devel systemd-devel \
+    python3 python3-pip python3-ruamel-yaml python3-requests python3-packaging
 
 RUN npm install npm2rpm --global
 
-RUN pip3 install packaging requests ruamel-yaml
+RUN gem install gem2rpm
+
+RUN pip3 install rpmspectool obal
```
ekohl, line 21: Also surprises me `rpmspectool` isn't in EPEL (https://src.fedoraproject.org/rpms/rpmspectool). Almost makes me think people typically use Fedora to develop ;)

ekohl (commented):
I have some vague recollection that GH would also recognize these for codespaces and when Vox Pupuli had a credit card attached to it that would bill. Not that we have one now for Foreman, but I still think we should evaluate https://docs.github.com/en/codespaces/managing-codespaces-for-your-organization/choosing-who-owns-and-pays-for-codespaces-in-your-organization just to be sure.

adamruzicka:
Alright, this should be more or less ready to go

> I have some vague recollection that GH would also recognize these for codespaces

I've heard of that, but have no experience with it and no idea how it works, what the permission model is and who it gets billed to. So yes, definitely something to look into.

ekohl (approved):
I don't use this myself so I'm going to trust you on it and merge it.

adamruzicka:
It might be hit and miss based on what people use as a container runtime and what tooling they use to interact with the devconatiner itself (devcontainer-cli vs vscode vs devpod, maybe there are others), but it is something to get people started

PR #11635, opened 2025-01-16
Switch the base image to almalinux:9

adamruzicka:


Containerfile
```diff
@@ -10,12 +10,10 @@ RUN echo "[git-annex]" >> /etc/yum.repos.d/git-annex.repo \
          && echo "gpgcheck=0" >> /etc/yum.repos.d/git-annex.repo \
          && echo "enabled=1" >> /etc/yum.repos.d/git-annex.repo
 
-RUN dnf -y module enable nodejs:14 ruby:2.7
```
ekohl, line 13: We use the nodejs 18 dnf module on EL9 because default is 16 (and lacks nodejs-packaging)

ehelms (approved):

PR #11636, opened 2025-01-17
Make foreman_scap_client_bash obsolete versioned

ofedoren:
Pinned to unreleased and unplanned version, just to be sure.

### DIFF
--- a/packages/client/foreman_scap_client_bash/foreman_scap_client_bash.spec
+++ b/packages/client/foreman_scap_client_bash/foreman_scap_client_bash.spec
@@ -5,7 +5,7 @@
 
 Name: foreman_scap_client_bash
 Version: 0.2.0
-Release: 1%{?dist}
+Release: 2%{?dist}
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I think you need to increase the release since we already built existing one. CI will also yell at you about it.

ofedoren:
Thanks, @ekohl, updated.

packages/client/foreman_scap_client_bash/foreman_scap_client_bash.spec
```diff
@@ -22,7 +22,7 @@ Requires: openscap-scanner
 Requires: bzip2
 BuildArch: noarch
 
-Obsoletes: rubygem-foreman_scap_client
+Obsoletes: rubygem-foreman_scap_client < 0.6.3-1
```
ekohl, line 25: In Fedora it's custom to only pin to the currently packaged version and increase the release by 1

adamruzicka:
Merged as part of https://github.com/theforeman/foreman-packaging/pull/11637 , thank you @ofedoren !

PR #11637, opened 2025-01-20
Make foreman_scap_client_bash obsolete versioned

adamruzicka:
Includes #11636


ekohl:
I suppose we should also drop the Ruby version from our packaging, but that it's too soon to archive https://github.com/theforeman/foreman_scap_client. It may be good to add a note there that it's deprecated in favor of another repo though.

adamruzicka:
Yes, as well as writing a heads up on discourse.

PR #11649, opened 2025-01-23
Add rubygem-rexml package

adamruzicka:
Companion to https://github.com/theforeman/smart-proxy/pull/910


ekohl (changes requested):
I see rubygem-rexml on EL9 so unless we need a version newer than 3.2.5 I think we can ignore this.

PR #11651, opened 2025-01-24
Add foreman_ygg_migration to el10 client comps

adamruzicka:

### DIFF
--- a/comps/comps-foreman-client-el10.xml
+++ b/comps/comps-foreman-client-el10.xml
@@ -12,6 +12,7 @@
     <packagelist>
       <packagereq type="default">foreman-client-release</packagereq>
       <packagereq type="default">foreman_scap_client_bash</packagereq>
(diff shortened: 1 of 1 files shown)


evgeni:
In https://github.com/theforeman/foreman-packaging/pull/11532/files#diff-4eca98f34c4b4de3ea829186cc0bb660ab272ddd28ccc5f50713e50cf656d0f0 we said we don't need that on EL10 -- did that change?
(I am fine including it, just trying to understand what changed from our last assesment)

adamruzicka:
Back then we assumed that people doing in-place upgrade from 9.6 to 10 would already have the yggdrasil-0.4.z. Now it seems el9 won't get newer yggdrasil, unless you take it from epel.

evgeni (approved):

evgeni:
kicked off a el10 build: https://copr.fedorainfracloud.org/coprs/build/8566640

PR #11652, opened 2025-01-24
Declare rexml as a dependency of foreman-proxy

adamruzicka:


ekohl (approved):
I wish we already had more automatic requires.

PR #11671, opened 2025-01-28
templates: Drop python-dev build-dependency

evgeni:


ekohl (commented):
It was added in 5d0282f12548a56c39ba2767fd441c59557a28da. Is this now all right because we no longer rely on gyp, or was it always wrong to begin with?

evgeni:
It's now wrong as there is no `python-dev` in Jammy (there is `python3-dev`, but oh well), and it seems to build fine without, so I just don't care.

evgeni:
(I could imagine Python is what's needed for building, not Python-dev, and that is installed, even tho I do not know why)

ekohl (approved):

PR #11703, opened 2025-01-31
Add python-setuptools-rust

Odilhao:


Odilhao:
Repoclosure failure is expected

PR #11707, opened 2025-02-03
Add python-pycparser

Odilhao:


Odilhao:
Looks like this is provided by Appstream, closing

PR #11730, opened 2025-02-14
Fix rubygem-font-awesome-sass

adamruzicka:
It fails to build with:
```
%patchN is obsolete, use %patch N (or %patch -P N): %patch0 -p1
RPM Failed to parse spec file
```

### DIFF
--- a/packages/foreman/rubygem-font-awesome-sass/rubygem-font-awesome-sass.spec
+++ b/packages/foreman/rubygem-font-awesome-sass/rubygem-font-awesome-sass.spec
@@ -3,7 +3,7 @@
 
 Name: rubygem-%{gem_name}
 Version: 4.6.2
-Release: 9%{?dist}
+Release: 10%{?dist}
 Summary: Font-Awesome SASS
 License: MIT
 URL: https://github.com/FortAwesome/font-awesome-sass
@@ -31,8 +31,7 @@ BuildArch: noarch
 Documentation for %{name}.
 
 %prep
-%setup -q -n  %{gem_name}-%{version}
-%patch0 -p1
+%autosetup -p1 -n %{gem_name}-%{version}
 
 # Foreman relies on the Rails integration and that pulls in sass
 %gemspec_remove_dep -g sass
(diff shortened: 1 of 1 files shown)


evgeni:
> It fails to build with:
> 
> ```
> %patchN is obsolete, use %patch N (or %patch -P N): %patch0 -p1
> RPM Failed to parse spec file
> ```

This should only be a problem on Fedora tho ;-)

adamruzicka:
Well, EL will catch up with that eventually

evgeni:
```
+ cd font-awesome-sass-4.6.2
+ /usr/bin/chmod -Rf a+rX,u+w,g-w,o-w .
Patch #0 (0001-Update-to-use-either-sass-or-sassc-gem.patch):
+ echo 'Patch #0 (0001-Update-to-use-either-sass-or-sassc-gem.patch):'
+ /usr/bin/patch --no-backup-if-mismatch -p1 --fuzz=0
patching file lib/font-awesome-sass.rb
+ echo 'Patch #0 (0001-Update-to-use-either-sass-or-sassc-gem.patch):'
Patch #0 (0001-Update-to-use-either-sass-or-sassc-gem.patch):
+ /usr/bin/patch --no-backup-if-mismatch -p1 --fuzz=0
patching file lib/font-awesome-sass.rb
Reversed (or previously applied) patch detected!  Assume -R? [n] 
Apply anyway? [n] 
Skipping patch.
2 out of 2 hunks ignored -- saving rejects to file lib/font-awesome-sass.rb.rej
error: Bad exit status from /var/tmp/rpm-tmp.xcbgsr (%prep)
    Bad exit status from /var/tmp/rpm-tmp.xcbgsr (%prep)
```

Why is it patching twice?!

packages/foreman/rubygem-font-awesome-sass/rubygem-font-awesome-sass.spec
```diff
@@ -32,7 +32,7 @@ Documentation for %{name}.
 
 %prep
 %setup -q -n  %{gem_name}-%{version}
-%patch0 -p1
+%patch -P 0 -p1
```
ekohl, line 35: Should this use [autosetup](https://rpm-software-management.github.io/rpm/manual/autosetup.html) instead?
```suggestion
%autosetup -p1-n %{gem_name}-%{version}
```

adamruzicka:
Apparently `%patch 0` and `%patch -P 0` don't do exactly the same thing.

packages/foreman/rubygem-font-awesome-sass/rubygem-font-awesome-sass.spec
adamruzicka: That seems to work as well

ekohl (approved):
@evgeni mind checking my suggestion? :)

evgeni (approved):

PR #11737, opened 2025-02-18
Foreman 3.14 keyring updates

pcreech:


ekohl (approved):
AFAIK we haven't rotated the Debian key in a while so perhaps we also need to do this in nightly.

PR #11745, opened 2025-02-19
Update python version requirements for ansible-collection-redhat-satellite and ansible-collection-theforeman-foreman

Odilhao:

PR #11746, opened 2025-02-19
Update python version requirements for ansible-collection-redhat-satellite and ansible-collection-theforeman-foreman

Odilhao:


Odilhao:
Cherrypick from https://github.com/theforeman/foreman-packaging/pull/11745

PR #11782, opened 2025-02-24
Rewrite spec files with %scl_package

ekohl:
There are some spec files (most notably, hammer_plugin) that still use SCL packaging macros. Those aren't needed anymore and break on modern packaging tools, which you notice on Fedora and Packit.

### DIFF
--- a/bump_rpm.sh
+++ b/bump_rpm.sh
@@ -88,11 +88,14 @@ if [[ $CURRENT_VERSION != "$NEW_VERSION" ]] ; then
 		bump_spec $SPEC_FILE "${NEW_VERSION}"
 
 		TEMPLATE="$(awk '/^# template: / { print $3 }' $SPEC_FILE)"
-		if [[ $TEMPLATE == 'scl' ]] || [[ $TEMPLATE == 'nonscl' ]] || [[ -z $TEMPLATE ]]; then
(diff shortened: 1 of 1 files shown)


bump_rpm.sh
```diff
@@ -88,7 +88,10 @@ if [[ $CURRENT_VERSION != "$NEW_VERSION" ]] ; then
 		bump_spec $SPEC_FILE "${NEW_VERSION}"
 
 		TEMPLATE="$(awk '/^# template: / { print $3 }' $SPEC_FILE)"
-		if [[ $TEMPLATE == 'scl' ]] || [[ $TEMPLATE == 'nonscl' ]] || [[ -z $TEMPLATE ]]; then
+		if [[ $TEMPLATE == 'scl' ]] || [[ $TEMPLATE == 'nonscl' ]] || [[ -z $TEMPLATE ]] || grep --no-messages %scl_package $SPEC_FILE; then
+			if [[ $TEMPLATE == 'scl' ]] || [[ $TEMPLATE == 'nonscl' ]] || [[ -z $TEMPLATE ]] ; then
+				TEMPLATE=default
+			fi
```
evgeni, line 94: I don't think that's needed, the code uses `gem2rpm -t "$ROOT/gem2rpm/default.spec.erb" -o $SPEC_FILE ./*.gem` without the `$TEMPLATE` var anyway

ekohl: Oh, I intended to use `$TEMPLATE` in there sonthat's the bug. The idea was to maintain `hammer_pluging` if set

ekohl: Updated

PR #11793, opened 2025-02-26
Update rubygem-pulp_deb_client to 3.5.1

quba42:
I need the newer bindings in order to [fix a bug](https://github.com/Katello/katello/pull/11365).

I am fine with all of this just making it into nightly repositories for now, which I believe queues these changes for the next regular Katello release (4.17).


quba42:
I think I need some help with this one:

- I believe this change requires also updating here: https://github.com/Katello/katello/blob/master/katello.gemspec#L58
- And also here: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/rubygem-katello/rubygem-katello.spec#L52

I do not remember in what order it all needs to be prepared and then merged.

I am fine with all of this just making it into nightly repositories for now, which I believe queues these changes for the next regular Katello release (4.17).

I do need the newer bindings in order to fix a bug.

Question: What else do I need to prepare in what order?

quba42:
Looks like this kind of PR can also be created via automation, e.g.: https://github.com/theforeman/foreman-packaging/pull/11788

quba42:
Closing in favor of #11943

PR #11815, opened 2025-03-03
Update foreman fog proxmox deb

Manisha15:
Update ruby-foreman-fog-proxmox to 0.18.1 in deb/3.14 to support Rails 7.


plugins/ruby-foreman-fog-proxmox/foreman_fog_proxmox.rb
```diff
@@ -1,2 +1 @@
-gem 'foreman_fog_proxmox', '0.17.1'
-gem 'fog-proxmox', '0.15.0'
```
adamruzicka, line 2: I don't think this should be removed

evgeni: Why not? I did the same in https://github.com/theforeman/foreman-packaging/pull/11801
The plugin will load the gem itself

plugins/ruby-foreman-fog-proxmox/debian/gem.list
```diff
@@ -1,6 +1,6 @@
 # list of gem urls to download via Jenkins, space separated
-GEMS="https://rubygems.org/downloads/foreman_fog_proxmox-0.17.1.gem"
-GEMS="$GEMS https://rubygems.org/downloads/fog-proxmox-0.15.0.gem"
+GEMS="https://rubygems.org/downloads/foreman_fog_proxmox-0.18.1.gem"
+GEMS="$GEMS https://rubygems.org/downloads/fog-proxmox-0.15.1.gem"
 # shouldn't be needed, but is currently specified in the gemspec
 # https://github.com/fog/fog-proxmox/pull/71
 GEMS="$GEMS https://rubygems.org/downloads/ipaddress-0.8.3.gem"
```
ekohl, line 6: This PR was merged so perhaps also time to remove it?
```suggestion
```

plugins/ruby-foreman-fog-proxmox/foreman_fog_proxmox.rb
adamruzicka: At first it looked like accidental removal.

> The plugin will load the gem itself

Would that mean that we never needed this or it was needed in the past but not anymore?

evgeni: I lean towards "never needed".
(tasks also doesn't list any deps: https://github.com/theforeman/foreman-packaging/blob/deb/develop/plugins/ruby-foreman-tasks/foreman-tasks.rb)

evgeni: This has been here since https://github.com/theforeman/foreman-packaging/pull/3251 and I see no reason for it.

ekohl (changes requested):
I now see this is in 3.14 and not a cherry pick of 617040d56f7eb8bacef31066ed6e167615a7c935. Could you please make sure it's just a cherry pick commit?

Manisha15:
> I now see this is in 3.14 and not a cherry pick of [617040d](https://github.com/theforeman/foreman-packaging/commit/617040d56f7eb8bacef31066ed6e167615a7c935). Could you please make sure it's just a cherry pick commit?

Updated the PR with just cherry-pick from develop branch.

ekohl (approved):

PR #11838, opened 2025-03-06
foreman: update NPM dependencies

evgeni:


evgeni:
repoclosure expected \o/

PR #11839, opened 2025-03-07
Add support for OpenVox agent

binford2k:
It *might* be better to start at version 8.11.0 instead. The 7.35.0 version is only best-effort supported. It's only intended to help people migrate and then upgrade.

### DIFF
--- a/plugins/puppet-agent-puppet-strings/debian/control
+++ b/plugins/puppet-agent-puppet-strings/debian/control
@@ -8,7 +8,7 @@ Homepage: https://github.com/puppetlabs/puppet-strings
 
 Package: puppet-agent-puppet-strings
 Architecture: all
-Depends: ${shlibs:Depends}, ${misc:Depends}, puppet-agent,
+Depends: ${shlibs:Depends}, ${misc:Depends}, openvox-agent | puppet-agent,
   puppet-agent-yard (>= 0.9.5), puppet-agent-yard (<< 1.0.0),
   puppet-agent-rgen (>= 0.9.0), puppet-agent-rgen (<< 1.0.0)
 Description: Puppet documentation via YARD
--- a/debian/bookworm/foreman-installer/changelog
+++ b/debian/bookworm/foreman-installer/changelog
@@ -1,3 +1,9 @@
+foreman-installer (5.0.0-2) stable; urgency=low
+
+  * Add support for OpenVox agent
+
+ -- Dirk Goetz <dirk.goetz@netways.de>  Thu, 16 Apr 2026 17:21:00 +0100
+
 foreman-installer (5.0.0-1) stable; urgency=low
 
   * Bump changelog to 5.0.0 to match VERSION
--- a/debian/bookworm/foreman-installer/control
+++ b/debian/bookworm/foreman-installer/control
@@ -15,7 +15,7 @@ Architecture: all
 XB-Ruby-Versions: ${ruby:Versions}
 Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter,
          ruby-kafo (>= 7.7.0), ruby-kafo (<< 8.0.0),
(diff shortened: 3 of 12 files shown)


deubert-it:
Thank you @binford2k for taking care of this.

I'd just like to add to this issue that I'm a former puppet ce user, our test environment has successfully upgraded all agents to openvox-agent 8, the puppetdb to openvox-db 8, the puppetserver to openvox-server 8, however the foreman-installer package is no longer available due to package dependencies.

```
apt-get install foreman-installer
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 foreman-installer : Depends: puppet-agent (>= 7.0.0) but it is not going to be installed or
                              puppet (>= 7.0.0)
E: Unable to correct problems, you have held broken packages.
```

The changes which were supplied in this pull request are exactly what we would need to make use of the foreman-installer again. I'm not sure yet what exactly foreman-installer will do, but the scenario answers.yaml seems to have enough options to support different package names, so should be fully useable with openvox packages.

tyybbi:
Personally I'd welcome this PR as I'm in the same boat with @deubert-it as of typing this.

In any case, cheers for the effort to all involved in bringing a full OpenVox support to Foreman!

ogajduse:
[test deb]

ogajduse (changes requested):
I see this PR as an equivalent to #12533 in the rpm/develop branch.

The patch looks sane to me. However, we still need to bump the `changelog` files in all packages that were changed.

- debian/bookworm/foreman-installer/changelog - bump from 3.18.0-1 to 3.18.0-2                                                                                                                                  
- debian/jammy/foreman-installer/changelog - bump from 3.18.0-1 to 3.18.0-2                                                                                                                                     
- plugins/puppet-agent-oauth/debian/changelog - bump from 0.5.5-1 to 0.5.5-2                                                                                                                                    
- plugins/puppet-agent-puppet-strings/debian/changelog - bump from 3.0.1-2 to 3.0.1-3                                                                                                                           
- plugins/puppet-agent-rgen/debian/changelog - bump from 0.9.0-1 to 0.9.0-2                                                                                                                                     
- plugins/puppet-agent-yard/debian/changelog - bump from 0.9.37-1 to 0.9.37-2

plugins/puppet-agent-puppet-strings/debian/control
ekohl, line 1: Perhaps at this point we should switch over to openvox-strings. We may need to fix Kafo to support that too though.

dgoetz:
In https://community.theforeman.org/t/foreman-installer-3-18-1-dependency-problem-on-debian-12/46189/5 a user is complaining about this. For rpm we already have this support and the installer moved to puppet modules which require >8.0.0, so @binford2k are you willing and able to update the pull request?

dgoetz:
Rebased, updated and changelog added.

@ogajduse Can you re-review please?

@ekohl As debian is lacking this support now while RPM already has it, would it be ok to merge this and postpone the switch to openvox-strings?

ekohl:
> @ekohl As debian is lacking this support now while RPM already has it, would it be ok to merge this and postpone the switch to openvox-strings?

The OpenVox Debian packages should also "provide" `puppet-agent` so I think it shouldn't be a hard requirement. However, I don't object to merging this.

dgoetz:
Rebased

ogajduse (changes requested):
The `puppet-agent-puppet-strings` changelog version need to be changed from `3.0.1-2` to `3.0.1-3`.

dgoetz:
> The `puppet-agent-puppet-strings` changelog version need to be changed from `3.0.1-2` to `3.0.1-3`.

Thanks, I adjusted it.

ogajduse (approved):

PR #11840, opened 2025-03-07
Add OpenVox support to RPM

binford2k:
It might be better to start at version 8.11.0 instead. The 7.35.0 version is only best-effort supported. It's only intended to help people migrate and then upgrade.

### DIFF
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -16,7 +16,7 @@ BuildArch:  noarch
 
 Requires:   curl
 Requires:   hostname
-Requires:   puppet-agent >= 7.0.0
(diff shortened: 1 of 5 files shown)


binford2k:
(yes; this does indeed use Puppet to build a package supporting OpenVox. The build dependency tree was rather sprawling.)

ekohl:
ok to test

ekohl (commented):
You'll need to bump the release on each package to get it to build. I normally use `rpmdev-bumpspec` to do so.

ogajduse:
@binford2k Is this PR still relevant given that #12533 was merged?

ekohl (commented):
There is now a `Provides: puppet-agent` in the `openvox-agent` packages so technically this isn't needed anymore.

ekohl:
Oh, and you're right that this is now redundant because of #12533.

PR #11849, opened 2025-03-17
Fix Requires on base64 and csv on Ruby < 3.4

evgeni:
See https://github.com/theforeman/hammer-cli/commit/88c14e374ba43c7aa93c2f88faae02c41870f933


evgeni:
Repoclosure is failing on the existing nightly rpm. not the changes here.

ekohl (approved):

PR #11850, opened 2025-03-17
Fixes #38292 - Remove oVirt [rpm]

stejskalleos:
* For 3.16
* Tracker: https://projects.theforeman.org/issues/38266


packages/foreman/foreman/foreman.spec
```diff
@@ -267,21 +267,6 @@ Meta package to install requirements for OpenStack compute resource support.
 %files openstack
 %{_datadir}/%{name}/bundler.d/openstack.rb
 
-%package ovirt
```
stejskalleos, line 270: @theforeman/packaging, is this enough to remove the rpms from the host?
When users run `dnf update,` will it automatically remove all obsolete oVirt packages, or do I have to write some kind of migration to take care of this?

ekohl (changes requested):

packages/foreman/foreman/foreman.spec
ekohl: No, you should add an `Obsoletes` statement that tells dnf to take care of it. Here's a good example:
https://github.com/theforeman/foreman-packaging/blob/8258c627534f90d72e7695f5b621f154a9ad50f1/packages/plugins/rubygem-foreman_google/rubygem-foreman_google.spec#L15-L16

I'm debating if it should be on the main foreman package or in foreman-obsolete-packages. Probably the former.

packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
```diff
@@ -33,6 +35,9 @@ from the distribution for some reason.
 %files
 
 %changelog
+* Tue Apr 01 2025 Leos Stejskal - 1.11-1
+- Obsolete Ovirt
```
ekohl, line 39:
```suggestion
- Obsolete oVirt
```

packages/foreman/foreman/foreman.spec
ekohl, line 1: On the main package I'd expect an `Obsoletes: %{name}-ovirt < %{version}-%{release}` to make sure the old package is removed on upgrades.

ekohl (changes requested):

packages/foreman/foreman/foreman.spec
```diff
@@ -24,6 +24,7 @@ BuildArch:  noarch
 
 # Plugin was removed in Foreman 3.3, 3.5 includes DB cleanup
 Obsoletes: rubygem-foreman_docker < 5.0.0-4
+Obsoletes: %{name}-ovirt < %{version}-%{release}
```
ekohl, line 27: Can you also add a comment that this was done in 3.16 so in the future we know roughly how old it was?

ekohl (approved):
Minor nit but otherwise :+1:

stejskalleos:
deb: https://github.com/theforeman/foreman-packaging/pull/12097

stejskalleos:
@theforeman/packaging and it's :green_apple:

ShimShtein:
I don't have a setup to test it, but LGTM code-wise

PR #11856, opened 2025-03-18
Update rubygem-acts_as_list to 1.2.4

evgeni:


evgeni:
cc @nofaralfasi :)

adamruzicka:
Does this still need to be a draft?

PR #11906, opened 2025-04-01
Update rubygem-hammer_cli_foreman_rh_cloud to 1.0.4

chris1984:


packages/katello/rubygem-hammer_cli_foreman_rh_cloud/rubygem-hammer_cli_foreman_rh_cloud.spec
```diff
@@ -1,19 +1,14 @@
...
 
-Name: %{?scl_prefix}rubygem-%{gem_name}
-Version: 1.0.3
+Name: rubygem-%{gem_name}
+Version: 1.0.4
 Release: 1%{?foremandist}%{?dist}
 Summary: Foreman Rh Cloud plugin for Hammer CLI
-Group: Development/Languages
-License: GPLv3
+License: GPLv3+
```
evgeni, line 11: This comes from https://github.com/theforeman/hammer-cli-foreman-rh-cloud/commit/071ca73c065105f5734b049fa56cd2b6b61ef448, but I wonder if it's correct?

chris1984: We updated the license because of this test failure https://github.com/theforeman/hammer-cli-foreman-rh-cloud/actions/runs/14181080477/job/39727118228 so Partha changed it here to make it happy https://github.com/theforeman/hammer-cli-foreman-rh-cloud/pull/10/files#diff-c80b304344995dba0dfebab83429d8b58e7bbd54bc7cf8fb72a6c284e9a63bceR18 we have that same license in hammer-cli-katello

ekohl (approved):

PR #11953, opened 2025-04-21
jammy, bookworm: set ruby-webrick dependency to >= 1.7.0

kenyon:
If the dependency is set to >= 1.0.0, then that allows the ruby-webrick of ruby2.7 to remain installed. This breaks foreman-proxy when upgrading from Ubuntu 20.04 to 22.04, for example, since foreman-proxy uses ruby3.0.

This fixes commit 79c8cf9e05f98c02d35d5dd3c79f0bf5f4de92b1, part of #11109.

### DIFF
--- a/debian/bookworm/foreman-proxy/changelog
+++ b/debian/bookworm/foreman-proxy/changelog
@@ -1,3 +1,9 @@
+foreman-proxy (3.15.0-2) stable; urgency=low
+
+  * Increase lower version bound for ruby-webrick to ensure that the ruby3.0 package is installed.
+
+ -- Kenyon Ralph <kralph@qti.qualcomm.com>  Tue, 22 Apr 2025 16:06:10 -0700
+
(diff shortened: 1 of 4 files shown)


evgeni:
Can you please add changelog entries? Otherwise Jenkins won't build it.

kenyon:
@evgeni like that? commit 29b2fa5549124298038efc3a499de3428e6379ec

ekohl (changes requested):
It needs its own entry. There's a helper script in the repo but on mobile it's a pain to link to

kenyon:
Found the helper script. How does commit 00ae474d5e48bf981e95e9ac7ed49dcde63b006e look?

ekohl:
ok to test

ekohl (approved):

evgeni (approved):

kenyon:
@evgeni thanks for merging. Cherry-picks for 3.14 and 3.13 are #11954 and #11955.

PR #11955, opened 2025-04-21
jammy, bookworm: set ruby-webrick dependency to >= 1.7.0

kenyon:
(Cherry-pick of #11953)

If the dependency is set to >= 1.0.0, then that allows the ruby-webrick of ruby2.7 to remain installed. This breaks foreman-proxy when upgrading from Ubuntu 20.04 to 22.04, for example, since foreman-proxy uses ruby3.0.

This fixes commit 79c8cf9e05f98c02d35d5dd3c79f0bf5f4de92b1, part of #11109.

### DIFF
--- a/debian/bookworm/foreman-proxy/changelog
+++ b/debian/bookworm/foreman-proxy/changelog
@@ -1,3 +1,9 @@
+foreman-proxy (3.13.1-2) stable; urgency=low
+
+  * Increase lower version bound for ruby-webrick to ensure that the ruby3.0 package is installed.
+
+ -- Kenyon Ralph <kralph@qti.qualcomm.com>  Mon, 28 Apr 2025 17:00:12 -0700
+
 foreman-proxy (3.13.1-1) stable; urgency=low
 
   * 3.13.1 released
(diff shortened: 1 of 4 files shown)


ogajduse:
[test deb]

pcreech:
[test deb]

ekohl (changes requested):
@kenyon I merged 3.13.1 and that created merge conflicts. Would you mind rebasing and also updating it to a single commit that references 3725e58656a97f8c0748314cf77c899caef11346 because that's the commit we ended up merging?

Then we can release this as part of 3.13.1

ogajduse:
[test deb]

ekohl (changes requested):
You can't change an existing release because we've already built 3.13.1-1 (it's in https://stagingdeb.theforeman.org/pool/bookworm/theforeman-3.13/f/foreman/ for example) and don't do rebuilds. It needs to be 3.13.1-2. For RPMs we have a check that ensures this, but I think we don't for debs.

kenyon:
@ekohl oh I see, misunderstood that 3.13.1-1 was already built. Updated the changelog to 3.13.1-2.

pcreech:
[test deb]

pcreech:
requested change has been made.  merging

PR #11943, opened 2025-04-14
Pulpcore 3.73 bindings bump

ianballou:
Should merge at the same time as https://github.com/Katello/katello/pull/11346

Other related packaging PRs will be created shortly.

PR #11956, opened 2025-04-22
Reverting https://github.com/theforeman/foreman-packaging/pull/11943 due to faraday-net_http bump issues.

pcreech:
Reverting https://github.com/theforeman/foreman-packaging/pull/11943 due to issues with bumping faraday-net_http.


Odilhao:
We need to delete all the builds from #11943 to avoid repoclosure problems after merging this.

Odilhao (approved):

pcreech:
@Odilhao they were already deleted :D the repoclosure issue here is that after deleting the builds, it wasn't bringing in the older ones.  I went into copr and rebuilt the old builds and repoclosure seems happy again.  I'm gonna run test on this PR to make sure

PR #11962, opened 2025-04-23
Drop obsolete from rubygem-pulp_container_client

Odilhao:
…rs during build with newer version of rpmbuild

### DIFF
--- a/packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
+++ b/packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
@@ -3,13 +3,11 @@
 
 Name: rubygem-%{gem_name}
 Version: 2.24.2
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: Pulp container plugin for Pulp3 API Ruby Gem
 License: GPLv2+
 URL: https://github.com/pulp/pulp_container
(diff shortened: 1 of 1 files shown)


Odilhao:
> Looking at [4aa74cc](https://github.com/theforeman/foreman-packaging/commit/4aa74ccabc1bf926b6b0dfe20b6ba9bb9436fcd5) it was renamed on EL7 so I think we never even published it on EL8, let alone EL9. Perhaps just drop it at this point?

Dropped the obsolete.

ekohl (approved):

packages/katello/rubygem-pulp_container_client/rubygem-pulp_container_client.spec
```diff
@@ -3,13 +3,12 @@
 
 Name: rubygem-%{gem_name}
 Version: 2.24.2
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: Pulp container plugin for Pulp3 API Ruby Gem
 License: GPLv2+
 URL: https://github.com/pulp/pulp_container
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
```
ekohl, line 11: Nit: perhaps drop this empty line too?
```suggestion
```

Odilhao: Done

PR #11971, opened 2025-04-28
foreman: Update GEM dependencies

evgeni:


Odilhao (approved):

ekohl (approved):
Related: mind reviewing https://github.com/theforeman/foreman-packaging/pull/8414 so that in the future these updates become redundant?

PR #11982, opened 2025-05-06
Update rubygem-ldap_fluff to 0.9.0

adamruzicka:
Not sure why the automation runs I kicked off didn't pick this up


evgeni:
idk, https://github.com/theforeman/foreman-packaging/pull/11988 existed and I merged it first, so :woman_shrugging:

PR #12009, opened 2025-05-11
Bump rubygem-http to 5.2.0-1

ekohl:
This is an alternative to https://github.com/theforeman/foreman-packaging/pull/11167 and depends on https://github.com/theforeman/foreman-packaging/pull/12008, per https://github.com/theforeman/foreman-packaging/pull/11165#pullrequestreview-2705992080. It also means ffi-compiler can be dropped.


evgeni (approved):

packages/plugins/rubygem-http/rubygem-http.spec
```diff
@@ -1,89 +1,72 @@
...
+%gemspec_remove_dep -g base64
 
-%{?scl:scl enable %{scl} - << \EOF}
-gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec
-%{?scl:EOF}
+# Replace llhttp-ffi with llhttp because it's more performant and easier to
+# package. We also don't care about non-MRI Ruby implementations.
+# https://github.com/httprb/http/issues/797
+%gemspec_remove_dep -g llhttp-ffi "~> 0.5.0"
+%gemspec_add_dep -g llhttp "~> 0.5.0"
```
evgeni, line 43: the versioning between `lhttp` and `lhttp-ffi` is not in sync, this should be `0.6.0` (as you packaged 0.6.1)

https://github.com/bryanp/llhttp/releases

```suggestion
%gemspec_remove_dep -g llhttp-ffi "~> 0.5.0"
%gemspec_add_dep -g llhttp "~> 0.6.0"
```

evgeni (changes requested):

evgeni (approved):

PR #12010, opened 2025-05-11
Drop rubygem-ffi-compiler

ekohl:
This was added as a dependency for rubygem-llhttp-ffi but packaging rubygem-llhttp is much easier and doesn't need this additional gem.


ekohl:
Now also deleted from https://copr.fedorainfracloud.org/coprs/g/theforeman/plugins-nightly-staging/

PR #12030, opened 2025-05-16
Update rubygem-foreman_dlm to 4.0.0

laugmanuel:


laugmanuel:
will be fixed in https://github.com/theforeman/foreman-packaging/pull/12041

PR #12033, opened 2025-05-16
Update rubygem-foreman_vault to 3.0.0

laugmanuel:


laugmanuel:
see https://github.com/theforeman/foreman-packaging/pull/12039

PR #12061, opened 2025-05-20
Drop unversioned obsoletes from foreman_rh_cloud

Odilhao:

### DIFF
--- a/packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
+++ b/packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
@@ -5,17 +5,11 @@
 
 Name: rubygem-%{gem_name}
 Version: 12.1.0
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Summary: Connects Foreman with Red Hat Cloud services
 License: GPLv3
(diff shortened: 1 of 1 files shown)


packages/katello/rubygem-foreman_rh_cloud/rubygem-foreman_rh_cloud.spec
```diff
@@ -5,18 +5,12 @@
...
 Name: rubygem-%{gem_name}
 Version: 12.1.0
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
 Summary: Connects Foreman with Red Hat Cloud services
 License: GPLv3
 URL: https://github.com/theforeman/foreman_rh_cloud
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
-Obsoletes: %{?scl_prefix}rubygem-redhat_access
```
zjhuntin, line 14: Would it be safer to version the obsolete? I guess i'm unsure if we still need this or if it's an artifact we can get rid of.

evgeni: that name is so old, it can be legally drinking in the US, remove it!

zjhuntin (approved):

evgeni:
[test rpm]

PR #12072, opened 2025-05-21
Remove all unversioned obsoletes from specfiles, and don't ignore in rpmlintrc

zjhuntin:
I do not think that any of these are relevant any longer, and as time as gone by unversioned obsoletes have become less acceptable in rpm spec files.


ekohl (commented):
Overall it looks good. Happy to see a lint override removed. On my phone so didn't check if we have any remaining. I suspect we only lint changed packages but that may be ok

packages/foreman/rubygem-apipie-params/rubygem-apipie-params.spec
```diff
@@ -6,7 +6,7 @@
 Summary: DSL for describing data structures
 Name: %{?scl_prefix}rubygem-%{gem_name}
 Version: 0.0.5
-Release: 5%{?dist}
+Release: 7%{?dist}
```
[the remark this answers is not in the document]
evgeni: looking at the changelog below, it seems the bump was executed twice accidentally

evgeni:
```
% git grep 'Obsoletes:[^<]*$'                  
packages/foreman/rubygem-concurrent-ruby-edge/rubygem-concurrent-ruby-edge.spec:%{?scl:Obsoletes: ruby193-rubygem-%{gem_name}}
packages/foreman/rubygem-concurrent-ruby-edge/rubygem-concurrent-ruby-edge.spec:%{?scl:Obsoletes: ruby193-rubygem-%{gem_name}-doc}
```

I think that's the only one left (and you didn't spot it, as it only adds the obsoletes when building into the SCL, which we don't)

packages/plugins/rubygem-jgrep/rubygem-jgrep.spec
```diff
@@ -74,6 +72,9 @@ cp -a ./%{_bindir}/* %{buildroot}%{_bindir}
 
 
 %changelog
+* Wed May 21 2025 Zach Huntington-Meath <zhunting@redhat.com>
```
evgeni, line 75:
```suggestion
* Wed May 21 2025 Zach Huntington-Meath <zhunting@redhat.com> - 1.3.3-12
```

.rpmlintrc
```diff
@@ -1,3 +1 @@
 addFilter('macro-in-comment')
```
evgeni, line 1: bundler_ext fails with `E: specfile-error warning: Macro expanded in comment on line 30:`, even tho this filter is still here.

should we just fix it? :)

ekohl: I wouldn't mind just fixing it.

zjhuntin:
[test rpm]

packages/foreman/rubygem-apipie-params/rubygem-apipie-params.spec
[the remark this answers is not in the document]
zjhuntin: My bad, fixed that.

zjhuntin:
[test rpm]

PR #12097, opened 2025-06-03
Remove oVirt [deb]

stejskalleos:


ekohl (changes requested):
AFAIK in Debian packaging you add `Replaces`. You can look at `foreman-compute` or `foreman-rackspace` for an example because those are packages we've removed in the past.

Also looking at @evgeni because the `Breaks` confuses me. We don't have it for `foreman-rackspace` but do have it for `foreman-compute`.

stejskalleos:
@theforeman/packaging updated, added foreman-ovirt to `Replaces` section

evgeni (changes requested):
There are no changelog entries, so this was never built.
My memory says you need both, Replaces *and* Breaks so that apt automatically removes the package.

stejskalleos:
added changelog and `breaks`

PR #12112, opened 2025-06-06
Add smart proxy request forwarder plugin

nadjaheitmann:


packages/plugins/rubygem-smart_proxy_request_forwarder/rubygem-smart_proxy_request_forwarder.spec
```diff
@@ -0,0 +1,87 @@
+# template: smart_proxy_plugin
+%global gem_name smart_proxy_request_forwarder
+%global plugin_name request_forwarder
+
+%global foreman_proxy_min_version 1.24
```
maximiliankolb, line 5: Does this indicate that the plugin works with foreman-proxy 1.24? I don't think we should make this promise.

Maybe:

```suggestion
%global foreman_proxy_min_version 3.11
```

packages/plugins/rubygem-smart_proxy_request_forwarder/rubygem-smart_proxy_request_forwarder.spec
```diff
@@ -0,0 +1,87 @@
...
+%{gem_spec}
+
+%files doc
+%doc %{gem_docdir}
+%doc %{gem_instdir}/README.md
+
+%changelog
+* Fri Jun 06 2025 Nadja Heitmann <nadjah@atix.de> 0.0.2-1
+- Add rubygem-smart_proxy_request_forwarder generated by gem2rpm using the smart_proxy_plugin template
+
```
maximiliankolb, line 87:
```suggestion
```

maximiliankolb (approved):
Looks sane to me. Though I cannot check if copr built the package. Neither copr nor Jenkins URLs work for me right now.

Another ACK welcome.

packages/plugins/rubygem-smart_proxy_request_forwarder/rubygem-smart_proxy_request_forwarder.spec
nadjaheitmann: @evgeni Any idea why it is using this foreman_proxy_min_version? Is it the default one? I don't think that we set one in the gemspec file of the plugin itself.

evgeni: yeah, 1.24 is hard-coded in the template: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/gem2rpm/smart_proxy_plugin.spec.erb#L15

set whatever you think is appropriate

package_manifest.yaml
```diff
@@ -643,6 +643,7 @@ foreman_plugin_packages_tier2:
     rubygem-foreman_wreckingball: {}
     rubygem-llhttp: {}
     rubygem-puppetdb_foreman: {}
+    rubygem-smart_proxy_request_forwarder: {}
```
ekohl, line 646: I think `foreman_proxy_plugin_packages` is a better group for it.

ekohl (dismissed):
As noted by @evgeni on Matrix: this is a security hazard. It effectively means everybody can create hosts or change hosts by uploading facts because there is no authentication whatsoever.

This can be easily used to break a Foreman installation by overloading it with false data, but note you can change a host's hostgroup via facts and that can also mean entirely different Puppet classes can be applied. This would be a good way to cause downtime in infrastructure.

I think we shouldn't package this until the security issues have been addressed.

evgeni:
Do I read the [code](https://github.com/theforeman/smart-proxy/blob/develop/lib/sinatra/authorization.rb) right, we'd need to add `do_authorize_any` to https://github.com/ATIX-AG/smart_proxy_request_forwarder/blob/f41bf79ba75abcb84f7dfb652256512bf8219c80/lib/smart_proxy_request_forwarder/api.rb#L6 and that'd do it?

ekohl:
Note that permission wise the Smart Proxy must have a feature that allows uploading facts / reports:

https://github.com/theforeman/foreman/blob/9ff8631bcd055e3bf7c9da49a92919abca572f73/app/controllers/api/v2/config_reports_controller.rb#L11
https://github.com/theforeman/foreman/blob/9ff8631bcd055e3bf7c9da49a92919abca572f73/app/controllers/api/v2/hosts_controller.rb#L19

Architecturally it does get a bit messy.

What I actually had thought about in the past was to include this functionality in the specific modules that upload facts.

For Puppet that would be the Puppet module. Then instead of [a complex report processor](https://github.com/theforeman/puppet-puppetserver_foreman/blob/master/files/report.rb) do this in the Smart Proxy module itself. Then you can use the built in HTTP report processor. Similarly, you could do the same for Salt or Ansible, though the last one might be difficult to get auth right.

nadjaheitmann:
We have merged https://github.com/ATIX-AG/smart_proxy_request_forwarder/pull/5 with also some small bug fixes. I have tested the PR with and without using certs for authentication and it seems to work fine. @evgeni @ekohl if you are aware of anything else that prevents this from being merged, let me know. Otherwise, I will bump a new version.

package_manifest.yaml
nadjaheitmann: Moved it there.

evgeni:
LGTM.

@ekohl are your concerns solved too?

evgeni:
@ekohl any objections? otherwise I'll merge this soon.

nadjaheitmann:
> @ekohl any objections? otherwise I'll merge this soon.

Should we move forward with this?

evgeni (approved):

PR #12128, opened 2025-06-16
update hammer plugins with latest template

evgeni:


ekohl (commented):
I'm missing `packages/foreman/rubygem-hammer_cli_foreman/rubygem-hammer_cli_foreman.spec` and `packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec` (where the latter should just have `hammer-cli` IMHO). Otherwise :+1:

ekohl (approved):

PR #12130, opened 2025-06-16
Add Provides for "hammer-cli"

evgeni:

### DIFF
--- a/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
+++ b/packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
@@ -3,7 +3,7 @@
 
 %global hammer_confdir %{_sysconfdir}/hammer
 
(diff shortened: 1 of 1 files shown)


packages/foreman/rubygem-hammer_cli/rubygem-hammer_cli.spec
```diff
@@ -25,6 +25,8 @@ BuildArch: noarch
 Requires: (rubygem(base64) or ruby-default-gems < 3.4)
 Requires: (rubygem(csv) or ruby-default-gems < 3.4)
 
+Provides: hammer-cli = %{version}%{?prerelease}
```
ekohl, line 28: Do we need to include the release here too?
```suggestion
Provides: hammer-cli = %{version}%{?prerelease}-%{release}
```

ekohl (approved):

PR #12167, opened 2025-06-30
Update rubygem-foreman_expire_hosts to 9.0.1

ochnerd:
(cherry picked from commit 7038c044b300b2fb3600c46169b9c78d3ce683c7)


packages/plugins/rubygem-foreman_expire_hosts/rubygem-foreman_expire_hosts.spec
```diff
@@ -77,6 +77,12 @@ mv %{buildroot}%{gem_instdir}/extra/*.cron %{buildroot}%{_sysconfdir}/cron.d/%{g
 %{foreman_plugin_log}
 
 %changelog
+* Wed Jun 18 2025 David Ochner <ochnerd@yahoo.de> - 9.0.1-1
+- Update to 9.0.1
+
+* Tue Jun 17 2025 David Ochner <ochnerd@yahoo.de> - 9.0.0-1
```
evgeni, line 83: You didn't cherry pick this commit, so the old 8.2.0 gem (that was deleted in it), is still present and breaks the build.

ochnerd: Now i included both cherry-picks

PR #12219, opened 2025-07-09
Add nodejs-scalprum-react-core package

ShimShtein:
Needed by https://github.com/theforeman/foreman/pull/10598


zjhuntin:
I'm unsure what's going on but the job seems to be pulling from [https://mirror.alwyzon.net/centos-stream/9-stream/BaseOS/x86_64/os/repodata/repomd.xml](https://mirror.alwyzon.net/centos-stream/9-stream/BaseOS/x86_64/os/repodata/repomd.xml) to get repo data, but does not exist. Unsure why we're seeing it specifically here :confused:

ShimShtein:
[test rpm]

PR #12249, opened 2025-07-16
Speed up build by not running strip_nondeterminism on node_modules

evgeni:


evgeni:
Before this, `dh_strip_nondeterminism` took like 10 minutes, now 30 seconds.

PR #12252, opened 2025-07-16
Update foreman.spec with scalprum

ShimShtein:
Required by https://github.com/theforeman/foreman/pull/10598


evgeni:
You need to bump the release and add a changelog entry for things to build properly.

evgeni (approved):

PR #12335, opened 2025-08-06
Update rubygem-fog-proxmox to 0.15.2

Manisha15:


Manisha15:
@evgeni The foreman-packaging pipeline doesn't trigger new packaging PR for fog-proxmox on new gem release. Can you please help me to automate it?

evgeni:
It doesn't know how to update non-plugins (and non-foreman dependencies).

But if you want it to bump things regardless, you can add it to https://github.com/theforeman/foreman-packaging/blob/8fe38b95bcc2bd2761e9090d2ca89a90a9e56441/list_updatable_packages#L16

(it will then just bump, ignoring whether foreman_fog_proxmox supports that version or not)

nadjaheitmann (approved):

evgeni:
And you probably want to update Debian too: https://github.com/theforeman/foreman-packaging/blob/be9980c60194a43cadd8378292efd2a56a2cf625/plugins/ruby-foreman-fog-proxmox/debian/gem.list#L3

PR #12339, opened 2025-08-08
[rpm] optionally update only specific packages

evgeni:

### DIFF
--- a/list_updatable_packages
+++ b/list_updatable_packages
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import argparse
 import json
 import os
 import subprocess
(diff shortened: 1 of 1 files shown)


evgeni:
> I'm more concerned with one implementation filtering on name and the other on gem name. But we can start with this and see how well it behaves. Worst case it won't update something and you need another manual run. When you forget it the scheduled run will pick it up.

The reasoning here was: for Debian, we only package gems (not node stuff), but Debian does "name normalization" (so the package name does not always include the source [gem] name), while rpm does gem *and* NPM, but the package names contain the literal source names. the alternative would be to introduce a `Spec.source_name` that returns `gem_name || npm_name`, but it didn't feel like we're gaining anything with that.

PR #12340, opened 2025-08-08
Allow passing a package name to bump pacakges

evgeni:


.github/workflows/bump_packages.yml
```diff
@@ -2,6 +2,11 @@ name: Update packages
 
 on:
   workflow_dispatch:
+    inputs:
+      package:
+        description: 'Limit bumping to this package'
```
ekohl, line 7: Some bikeshedding: should this suggest it's a substring match?

ekohl (commented):
Now that I'm more awake I wonder about the filtering: would it make sense to use `fnmatch` within the scripts for a more flexible filter?

evgeni:
> Now that I'm more awake I wonder about the filtering: would it make sense to use `fnmatch` within the scripts for a more flexible filter?

I *think* substring will suffice in most cases, but we can extend, sure.

.github/workflows/bump_packages.yml
evgeni: like this?

ekohl (approved):
There's a typo in the commit message, but otherwise 👍 to try this.

evgeni:
Maybe I wanted to have a cake? :)

PR #12358, opened 2025-08-12
Bump katello related packages to their 4.19 versions

pavanshekar:
Bump katello related packages to their 4.19 versions

### DIFF
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,10 +5,10 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 2
+%global release 1
 
 Name:       katello
-Version:    4.18.0
+Version:    4.19.0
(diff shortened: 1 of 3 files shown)


ogajduse:
ok to test

packages/katello/katello/katello.spec
```diff
@@ -8,7 +8,7 @@
 %global release 2
```
ogajduse, line 8: The package EVR and the changelog EVR do not match. The following patch should fix that.

```suggestion
%global release 1
```

ogajduse (changes requested):

packages/katello/katello/katello.spec
pavanshekar: I have made the suggested change and pushed the updated code

ogajduse:
[test rpm]

Odilhao:
[test rpm]

evgeni:
[test rpm]

evgeni:
okay, closure only complains about existing packages, let's go

PR #12362, opened 2025-08-13
Update rubygem-foreman_scc_manager to 5.0.1

nadjaheitmann:
(cherry picked from commit 576757c4a84414e4b345bcb7956e90e48a2cc3a6)


evgeni:
```
Module not found: Error: Can't resolve '@patternfly/react-icons/dist/esm/icons/times-icon'
```

seems the build is not compatible with the old js stuff in 3.15?

nadjaheitmann:
> ```
> Module not found: Error: Can't resolve '@patternfly/react-icons/dist/esm/icons/times-icon'
> ```
> 
> seems the build is not compatible with the old js stuff in 3.15?

Oh, this is bad. And we actually did test it on a 3.15-stable Foreman.

evgeni:
I *think* this should do it: https://github.com/ATIX-AG/foreman_scc_manager/pull/154

evgeni:
you still need the 5.0.0 commit, as that removes 4.0.1 from the git-annex index, otherwise things explode

evgeni:
I prefer individual pick commits, even if 5.0.0 never existed in 3.15, but it works without ;)

(read: I won't block, but it would make me happier)

nadjaheitmann:
> I prefer individual pick commits, even if 5.0.0 never existed in 3.15, but it works without ;)
> 
> (read: I won't block, but it would make me happier)

It's important to make people happy :) Thanks @evgeni !

PR #12383, opened 2025-08-19
Update rubygem-foreman_ovirt to 2.0.1

fraenki:
This updates rubygem-foreman_ovirt to 2.0.1 to fix cosmetic issues.
Changelog: https://github.com/markt-de/foreman_ovirt/blob/main/CHANGELOG.md


packages/plugins/rubygem-foreman_ovirt/rubygem-foreman_ovirt.spec
```diff
@@ -4,13 +4,17 @@
...
-Version: 2.0.0
+Version: 2.0.1
 Release: 1%{?foremandist}%{?dist}
 Summary: oVirt as a compute resource for The Foreman
 License: GPLv3
 URL: https://github.com/markt-de/foreman_ovirt
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
+# Foreman 3.16 dropped the oVirt compute provider and this is the replacement
+Obsoletes: foreman-ovirt < 3.16.0
```
evgeni, line 15: This is also present in `foreman`:
https://github.com/theforeman/foreman-packaging/blob/94f7f3017bef91098483aefd757fd3925847a1e4/packages/foreman/foreman/foreman.spec#L28-L29

Wonder what RPM/DNF will do with this?

fraenki: But I could test it, when new RPMs are available.

ekohl: It's probably best to drop it on Foreman itself. Then on the downstream side we should solve it in a different way.

ekohl:
I've created https://github.com/orgs/theforeman/teams/ovirt. Since the repo itself isn't under the Foreman org I can't add that repo, but I have added access to the packaging repository.

@fraenki could you submit a PR https://github.com/theforeman/foreman-plugin-overview to add it to https://theforeman.github.io/foreman-plugin-overview/? If you also submit a PR to https://github.com/theforeman/foreman-packaging/blob/deb/develop/CODEOWNERS and https://github.com/theforeman/foreman-packaging/blob/rpm/develop/CODEOWNERS you should be able to merge your own PRs.

You can also trigger automated packaging bumps after a new release. Using the `gh` CLI: `gh workflow run bump_packages.yml --repo theforeman/foreman-packaging -f package=foreman_ovirt`

fraenki:
@ekohl Done. Thank you!

ekohl (approved):
I think for now this is fine, but I think we need a follow up where we remove it from Foreman itself.

As for https://github.com/theforeman/foreman/blob/develop/db/migrate/20250414121956_migrate_ovirt_resources.rb I wonder if we can enhance that with `unless plugin_present?('ovirt')`.

fraenki:
> As for https://github.com/theforeman/foreman/blob/develop/db/migrate/20250414121956_migrate_ovirt_resources.rb I wonder if we can enhance that with `unless plugin_present?('ovirt')`.

Proposal: https://github.com/theforeman/foreman/pull/10657

PR #12396, opened 2025-08-20
Add oVirt as owner of the oVirt packages (RPM)

fraenki:
FYI @ekohl


dgoetz:
The pattern would not match [rubygem-ovirt-engine-sdk](https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/plugins/rubygem-ovirt-engine-sdk), if I am correct, which I would say you should also be allowed to update as a dependency.

PR #12397, opened 2025-08-20
Bump hammer_cli_katello to 1.19

pavanshekar:
Bump hammer_cli_katello to 1.19


ogajduse:
[ok to test]

ogajduse (dismissed):

evgeni:
> Looking at it further, I have no clue how that build could succeed. rubygems.org throws 403. That suggests that release 1.19.1 does not exist.

That's correct. hammer-cli-katello is a nightly package, so the source is fetched from jenkins (https://ci.theforeman.org/job/hammer-cli-katello-main-source-release/), not rubygems!

PR #12421, opened 2025-09-02
Bump nodejs-babel-preset-env to 7.9.5

evgeni:


nadjaheitmann:
@evgeni How comes you're bumping from version 1.7.x to 7.9.5? :)

evgeni:
Well, it also moves from the `babel-preset-env` package to `@babel/preset-env` (but the pkg name remains, due to the way we handle group-packages). So no idea why the jump is so big, but I think it's rather expected.

And yes, I think you can drop babel-preset-env in favor of `@babel/core` (or do plugins not need that at all? cc @MariaAga )

evgeni:
@MariaAga yeah, the question is: do plugins need to depend on that themself or not? given the above PR is not yet merged, I assume "they do now, but soon won't need to"

nadjaheitmann:
You lost me here. What do I need to do about the babel setup once the Foreman PR is merged? Are the dependencies automatically pulled in by Foreman? 

If you are planning to merge the Foreman PR soon, then I tend to wait until it is merged and then update the foreman_acd dependencies (and save multiple rounds of packaging and updates). Unless you prefer to fix it to have green pipelines or whatever is blocking at the moment @evgeni .

MariaAga:
It doesnt look like its (the foreman pr to remove foreman/build) is planned to be merged this month

nadjaheitmann:
OK, thanks for the info, then I'll just update ACD to have the same dependencies as any other plugin.

nadjaheitmann:
> That would mean to remove acd from nightly, tho. As our pipelines do not allow packages with broken dependencies. (It might work, as it's only a build-dependency, but still risky)

@evgeni Feel free to remove ACD from foreman-packaging if this unblocks the pipeline for this PR :)

nadjaheitmann:
> @evgeni Feel free to remove ACD from foreman-packaging if this unblocks the pipeline for this PR :)

Talking about removing ACD: https://github.com/theforeman/foreman-packaging/pull/12677 That would unblock this PR.

nadjaheitmann:
Talking about bumping versions in nodejs packages... Asking out of curiosity: Why do we check in the binary 'packages/foreman/nodejs-babel-preset-env/nodejs-babel-preset-env-7.9.5-registry.npmjs.org.tgz' while all the dependencies are checked in git-annexed? I see that we do this for all nodejs packages because the 'add_npm_package' script does it. But what is the reason no to git-annex the binary file as well?

nadjaheitmann:
> That file is a npm archive that is used for `bundle` packages. As that archive is not hosted anywhere, and is generated by `npm2rpm` we have to store the binary file here in git.

Thanks for clarifying.

ogajduse (approved):

PR #12455, opened 2025-09-09
RPM: Release Foreman 3.16.0

ogajduse:
Release Foreman 3.16.0


zjhuntin (approved):

pcreech:
Unkown issue with the copr build being done, but copr losing track of the state.  THe underlying copr build looks like a success.  So ack, ok to bypass merge rules.

PR #12501, opened 2025-09-20
Use system libffi

evgeni:
Fixes: 5c254bf99b4308ecd9c5bb3e6b790571626df68e


ekohl (approved):
https://github.com/ffi/ffi?tab=readme-ov-file#requirements mentions you can enforce it by calling `gem install ffi -- --enable-system-libffi` but I wonder if the `%gem_install` macro can do that. https://src.fedoraproject.org/rpms/rubygem-ffi/blob/rawhide/f/rubygem-ffi.spec at least doesn't force it.

evgeni:
The bundled version failed to build on EL10, so that's how I found out XD

PR #12561, opened 2025-10-06
Add python-obsah package

archanaserver:

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -525,6 +525,7 @@ foreman_installer_packages:
     puppet-agent-puppet-strings: {}
     puppet-agent-rgen: {}
     puppet-agent-yard: {}
+    python-obsah: {}
     rubygem-ansi: {}
     rubygem-kafo: {}
     rubygem-kafo_parsers: {}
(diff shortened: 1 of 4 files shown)


package_manifest.yaml
```diff
@@ -127,6 +127,8 @@ plugin_packages:
   children:
     foreman_plugin_packages: {}
     foreman_proxy_plugin_packages: {}
+  hosts:
+    python-obsah: {}
```
evgeni, line 131: This should be part of `foreman_installer_packages`, not `plugin_packages`.

packages/plugins/python-obsah/python-obsah.spec
evgeni, line 1: the file should live in `packages/foreman`, not `packages/plugins`

comps/comps-foreman-plugins-el9.xml
```diff
@@ -30,6 +30,7 @@
       <packagereq type="default">python3.12-lockfile</packagereq>
       <packagereq type="default">python3.12-markupsafe</packagereq>
       <packagereq type="default">python3.12-maturin</packagereq>
+      <packagereq type="default">python3.12-obsah</packagereq>
```
evgeni, line 33: this should be in `comps-foreman-el9.xml`, not in plugins.

evgeni (approved):

PR #12615, opened 2025-10-08
Update rubygem hammer cli foreman remote execution to 0.3.4

adamlazik1:


adamlazik1:
Update to 0.3.3 seems to have been missed last time, so I am adding it here as well.

PR #12638, opened 2025-10-21
Fix some client specs for suse

m-bucher:
Some more fixes for building client-packages for Suse Linux Enterprise Server. Especially when building on OpenSuseBuildService (OBS).

### DIFF
--- a/packages/client/foreman_ygg_worker/foreman_ygg_worker-rpmlintrc
+++ b/packages/client/foreman_ygg_worker/foreman_ygg_worker-rpmlintrc
@@ -1,0 +1,6 @@
+# This seems ok to me
+addFilter('invalid-url')
+
+# Ignore failing suse specific checks
+setBadness("suse-dbus-unauthorized-service", 0)
+setBadness("polkit-unauthorized-privilege", 0)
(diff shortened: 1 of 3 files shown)


packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
```diff
@@ -16,22 +16,20 @@
...
+# set only if go_arches is defined
+%if %{?go_arches:1}%{!?go_arches:0}
 ExclusiveArch: %{go_arches}
+%endif
 
 BuildRequires: systemd-rpm-macros
 %if 0%{?suse_version}
-BuildRequires: go
-BuildRequires: golang-packaging
+BuildRequires: (golang-packaging or go)
```
ogajduse, line 32: What is the benefit of switching to rich dependency syntax here?

packages/client/foreman_ygg_worker/foreman_ygg_worker-rpmlintrc
ogajduse, line 1: Please consider adding comments to each of these rules that were added to explain why they are needed. It's unclear what the suppressed warnings mean in the context of this package.

Sidenote: Here we're introducing a new pattern as it is the first package-scoped rpmlintrc to be added to this project.

ogajduse (commented):
Two questions, generally LGTM.

packages/client/foreman_ygg_worker/foreman_ygg_worker.spec
adamruzicka: Judging by the changelog, I'd guess different suse flavours have one or the other.

ogajduse: Oh, sure, I forgot that SUSE systems might have flavours. :sweat_smile:

PR #12276, opened 2025-07-23
Drop ellipsis with tooltip

evgeni:

PR #12768, opened 2025-11-22
Allow bootstrapping foreman-build

ekohl:
In a fresh environment you need the foreman-build package, but you can't build that without foreman itself. This uses [RPM conditionals](https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html) to only build the foreman-build and foreman-plugin packages.

It moves the macro definition next to the mkdir statement to have fewer if statements.

To use this you need to define the with option at the chroot level. I've opened https://github.com/theforeman/obal/pull/419 for that and working on the end-to-end pieces. Once that's merged you can use:

```
obal copr-project foreman-copr -e bootstrap=true
obal release foreman
obal copr-project foreman-copr
```

This is useful if you want to fully bootstrap it from scratch in your own user. It can also be used to start building a new OS (version) or architecture but then it needs some manual tweaking.

### DIFF
--- a/package_manifest.yaml
+++ b/package_manifest.yaml
@@ -1,6 +1,7 @@
 ---
 all:
   vars:
+    bootstrap: false
     foremandist: .fm3_19
     foreman_version: 'nightly'
     katello_version: 'nightly'
@@ -17,8 +18,7 @@ copr_projects:
   vars:
     core_modules_el9:
       - 'nodejs:22'
-    core_buildroot_packages:
-      - 'foreman-build'
+    core_buildroot_packages: "{{ [] if bootstrap|bool else ['foreman-build']}}"
     plugins_buildroot_packages:
       - 'foreman-build'
       - python3.12-rpm-macros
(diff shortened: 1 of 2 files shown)


ogajduse (approved):
Structurally LGTM

ekohl:
Rebased to clean the rpmlint warning and also changed the endif statements to be compatible with EL9 RPM.

ogajduse:
[test rpm]

ogajduse:
@ekohl Do we need a new obal release before we merge this?

ekohl:
It would be good, but we're currently in the stabilization period so big changes like these should IMHO wait until after branching.

ogajduse:
[test rpm]

ogajduse:
Let's get #12276 in first, then rebase this one as they both update and bump foreman.spec

ogajduse:
@ekohl, it is ready to be rebased. #12276 did not change the foreman.spec in the end.

ogajduse:
@ekohl One more thing... While we don't bootstrap every day, I'd say it is worth dropping at least a quick mention in the README about this bootstrap feature. What do you say?
If you agree, it could be a separate PR.

ekohl:
It would make sense to have a condensed version of https://community.theforeman.org/t/bootstrapping-foreman-rpm-packaging/45290 in the README, but right now I don't have the time to work on that.

PR #12875, opened 2025-12-20
Convert nodejs-buffer to single style package

ekohl:
The current bundle is outdated and only works with NPM < 6. Since there are only 2 dependencies the packaging is updated to a single style package with 2 additional dependencies. This avoids storing a tarball with version dependent files.


ogajduse:
[test rpm]

ogajduse:
Repoclosure failure expected.

ogajduse (approved):

PR #12944, opened 2026-01-25
Regenerate broken nodejs vendor caches

ekohl:
After the update to NodeJS 18 or 22 it also included an update to NPM and these packages no longer rebuild. This regenerates them which allow the packages to be rebruilt again.

This is important if you want to rebuild the whole of Foreman, for example on ARM or EL10.


packages/foreman/nodejs-compression-webpack-plugin/nodejs-compression-webpack-plugin.spec
```diff
@@ -67,7 +65,7 @@ for tgz in %{sources}; do
 done
 %{?scl:end_of_scl}
 
-%setup -T -q -a 14 -D -n %{npm_cache_dir}
+%setup -T -q -a 13 -D -n %{npm_cache_dir}
```
ogajduse, line 68: TIL: `-a <n> — Unpack The nth Sources After Changing Directory`

ogajduse (approved):

ogajduse:
@ekohl Thank you!

PR #12945, opened 2026-01-26
Update rubygem-fog-proxmox to 0.15.3

Manisha15:


nadjaheitmann (approved):

nadjaheitmann:
@evgeni Do you mind merging this? Looks like it is not part of proxmox team.

evgeni:
I am sure someone from @theforeman/packaging (who is not @evgeni) can handle that ;)

evgeni:
And yes, we should extend the `CODEOWNERS` to include this package
https://github.com/theforeman/foreman-packaging/blob/7954ced88b77b3422d25e3e98174928acd7efce8/CODEOWNERS#L59-L60

nadjaheitmann:
> And yes, we should extend the `CODEOWNERS` to include this package
> 
> https://github.com/theforeman/foreman-packaging/blob/7954ced88b77b3422d25e3e98174928acd7efce8/CODEOWNERS#L59-L60

What do you think about this? https://github.com/theforeman/foreman-packaging/pull/12947

PR #12969, opened 2026-02-02
Drop direct cron for foreman_expire_hosts plugin

archanaserver:
Relates to - https://github.com/theforeman/foreman_expire_hosts/pull/86
The plugin now registers with Foreman::Cron framework, running via cron:daily on traditional installations or systemd timers on container deployments.

### DIFF
--- a/packages/plugins/rubygem-foreman_expire_hosts/foreman_expire_hosts.cron.d
+++ b/packages/plugins/rubygem-foreman_expire_hosts/foreman_expire_hosts.cron.d
@@ -1,8 +1,0 @@
-SHELL=/bin/sh
-PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
-
-RAILS_ENV=production
-FOREMAN_HOME=/usr/share/foreman
-
-# Send out notifications about expired hosts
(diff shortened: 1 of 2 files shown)


packages/plugins/rubygem-foreman_expire_hosts/foreman_expire_hosts.cron.d
```diff
@@ -4,5 +4,10 @@ PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 RAILS_ENV=production
 FOREMAN_HOME=/usr/share/foreman
 
+# DEPRECATED: This cron job is now handled by the Foreman::Cron framework
```
archanaserver: So the cron.d file is kept here but commented out for backward compatibility with older Foreman versions that lack Foreman::Cron support, would it work this way, or should I consider dropping it completely?

evgeni: Huh, that file exists both in the gem an in the package? Weird.

As with the gem one: drop it, require foreman >= 3.18

zjhuntin (approved):

PR #12972, opened 2026-02-02
Revert pulp client packages

zjhuntin:
These were updated in error, currently Katello is not ready for these client packages, and they should not have been merged. I am unsure why we did not hit the issue with the repoclosure, but it did not.


Odilhao:
Did we delete all the copr builds already?

zjhuntin:
Not yet, I was going to 
Revert here
Delete COPR builds
make PR to bump + rebuild all the reverted builds

zjhuntin:
Now deleted

Odilhao (approved):

PR #12990, opened 2026-02-05
add google team to codeowners

evgeni:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -46,6 +46,10 @@
 # foreman_templates
 /packages/plugins/*foreman_templates/ @theforeman/packaging @theforeman/foreman_templates
 
+# google
+/packages/foreman/rubygem-google*/ @theforeman/packaging @theforeman/google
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -46,6 +46,11 @@
 # foreman_templates
 /packages/plugins/*foreman_templates/ @theforeman/packaging @theforeman/foreman_templates
 
+# google
+/packages/foreman/rubygem-google*/ @theforeman/packaging @theforeman/google
+/packages/plugins/rubygem-google*/ @theforeman/packaging @theforeman/google
+/packages/plugins/*foreman_google/ @theforeman/packaging @theforeman/google
```
ekohl: Perhaps merge these 2?
```suggestion
/packages/plugins/rubygem-*google*/ @theforeman/packaging @theforeman/google
```

ekohl (approved):

CODEOWNERS
evgeni: I was erring on the side of "the ones are deps, the others are the actual plugins", but not married to the solution :)

PR #12994, opened 2026-02-06
Add rubygem-foreman_opentofu package

m-bucher:
The pickle we are in here, is that we need the `tofu` package installed. At the moment that is not part of EL or foreman repos.

https://opentofu.org/docs/intro/install/rpm/

We probably have to think about that some more :thinking:


Manisha15:
[test rpm]

packages/plugins/rubygem-foreman_opentofu/rubygem-foreman_opentofu.spec
```diff
@@ -0,0 +1,83 @@
...
+Requires: ruby < 4
+BuildRequires: ruby >= 2.7
+BuildRequires: ruby < 4
+BuildRequires: rubygems-devel
+BuildArch: noarch
+Provides: foreman-plugin-%{plugin_name} = %{version}
+# end specfile generated dependencies
+
+BuildRequires: rubygem(deface) < 2.0
+Recommends: tofu
```
m-bucher, line 27: Fixed the Dependency problem by using`Recommends`.
We probably want to have a check in the Plugin that reports, if the tofu-command is missing 😄

zjhuntin (approved):

PR #13000, opened 2026-02-10
Bump foreman-tasks min version to 3.18.0

archanaserver:
Relates to https://github.com/theforeman/foreman-tasks/pull/794
Update rubygem-foreman-tasks to require Foreman >= 3.18.0 for Foreman::Cron framework integration.


evgeni:
This will happen automatically when the new release is packaged, no need to do it manually here.

archanaserver:
> This will happen automatically when the new release is packaged, no need to do it manually here.

we do it manually somewhere?

evgeni:
Debian packaging needs to do it manually. But that too can happen when we update to the latest tasks release

PR #13001, opened 2026-02-10
Fixes #39089 - Migrate Foreman cron.d to use Foreman::Cron framework

archanaserver:
Replace individual task cron entries with consolidated `cron:hourly`, `cron:daily`, `cron:weekly`, and `cron:monthly` entrypoints. This leverages the `Foreman::Cron` framework to manage recurring tasks centrally, allowing plugins to register tasks without modifying cron files.

PR #13002, opened 2026-02-10
Migrate Katello cron to consolidated entrypoint

archanaserver:
Relates to https://github.com/Katello/katello/pull/11633
Replace individual katello task cron entries with `cron:weekly` entrypoint. Update Katello packaging to require Foreman >= 3.18.0 for Foreman::Cron framework support.

### DIFF
--- a/packages/katello/katello/katello.cron
+++ b/packages/katello/katello/katello.cron
@@ -1,8 +1,0 @@
-SHELL=/bin/sh
-PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
-
-# Script for initiating removal of orphaned content
-00 22 * * 0	root	foreman-rake katello:delete_orphaned_content RAILS_ENV=production >/dev/null 2>&1
-
-# Script to refresh all alternate content sources
-00 18 * * 0     root    foreman-rake katello:refresh_alternate_content_sources >/dev/null 2>&1
--- a/packages/katello/katello/katello.spec
+++ b/packages/katello/katello/katello.spec
@@ -5,7 +5,7 @@
 %global confdir common
 %global prereleasesource master
 %global prerelease %{?prereleasesource:.}%{?prereleasesource}
-%global release 1
+%global release 2
 
 Name:       katello
 Version:    4.21.0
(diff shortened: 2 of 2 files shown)


ogajduse:
There is not much to review from a packaging perspective. I'd like to see someone from Katello review this PR.

archanaserver:
@ianballou

packages/katello/katello/katello.cron
```diff
@@ -1,8 +1,5 @@
 SHELL=/bin/sh
 PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
 
-# Script for initiating removal of orphaned content
-00 22 * * 0	root	foreman-rake katello:delete_orphaned_content RAILS_ENV=production >/dev/null 2>&1
-
-# Script to refresh all alternate content sources
-00 18 * * 0     root    foreman-rake katello:refresh_alternate_content_sources >/dev/null 2>&1
+# Run weekly Katello cron jobs
+00 18 * * 0     root    foreman-rake cron:weekly >/dev/null 2>&1
```
evgeni: When #13001 is merged, `cron:weekly` would be executed twice.

I think the correct course of action would be to merge #13001 and drop `katello.cron` completely

evgeni (changes requested):
This also needs a revbump and a changelog in `katello.spec`

packages/katello/katello/katello.cron
archanaserver: I agree with that, we don't need separate Katello cron file anymore, otherwise we are calling `cron:weekly` to run twice.

packages/katello/katello/katello.spec
```diff
@@ -129,6 +123,9 @@ Provides a federation of katello services
 # the files section is empty, but without it no RPM will be generated
 
 %changelog
+* Wed Mar 05 2026 Archana Kumari <akumari@redhat.com> - 4.21.0-0.2.master
```
evgeni:
```suggestion
* Thu Mar 05 2026 Archana Kumari <akumari@redhat.com> - 4.21.0-0.2.master
```

rpmlint is picky

archanaserver: argh! why I thought today is `Wed`, even though I can see it's Thur

evgeni (approved):

packages/katello/katello/katello.spec
ekohl: You can use `rpmdev-bumpspec` to bump it for you and it generates the timestamp. `obal bump-release` can do the same thing.

shubhamsg199 (approved):

PR #13004, opened 2026-02-10
Bump katello repos to 4.21 versions

zjhuntin:


ogajduse (approved):

evgeni:
Don't forget pulpcore this time ;)

ianballou:
We don't know what the new pulpcore will be yet, when the upgrade happens we'll bump it here. Keeping it the same for now is fine.

evgeni:
Well, that will make the RCs uninstallable as we enforce gpg sigs for non nighly katello-repos

Edit, ah, well, this is develop anyway, not 3.18…

PR #13031, opened 2026-02-12
Add foreman_opentofu plugin to deb packaging

m-bucher:


plugins/ruby-foreman-opentofu/debian/copyright
```diff
@@ -0,0 +1,7 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: foreman_opentofu
+Source: https://github.com/ATIX-AG/foreman_opentofu
+
+Files: *
+Copyright: 2026 ATIX AG <info@atix.de>
+License: GPLv3
```
ogajduse: Based on https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-field and the  project gemspec:
https://github.com/ATIX-AG/foreman_opentofu/blob/4d13684b6ac17632396145a4bb2e1eb305525d0d/foreman_opentofu.gemspec#L7

the `License` stanza should be

```suggestion
License: GPL-3
```
And below that, I'd expect the license text. `dependencies/bookworm/dynflow/copyright` holds a good example

https://github.com/theforeman/foreman-packaging/blob/ff40d9920a3b2f152dbc91d818ee7bbd2618b85d/dependencies/bookworm/dynflow/copyright#L32-L44

plugins/ruby-foreman-opentofu/debian/control
```diff
@@ -0,0 +1,23 @@
...
+               foreman-nulldb (>= 3.0.0)
+Standards-Version: 4.6.1
+Homepage: https://github.com/ATIX-AG/foreman_opentofu
+
+Package: ruby-foreman-opentofu
+Architecture: all
+Depends: ${misc:Depends},
+         ${ruby:Depends},
+         ${shlibs:Depends},
+         foreman (>= 3.0.0)
```
ogajduse: Looking at the project gemspec, don't we need  `ruby-foreman-deface (<<2.0.0)` in runtime Depends too?

plugins/ruby-foreman-opentofu/debian/copyright
ogajduse, line 1: At the bottom of this file, I suggest adding another File and License stanza to indicate the license used for packaging work.

Examples:
https://github.com/theforeman/foreman-packaging/blob/ff40d9920a3b2f152dbc91d818ee7bbd2618b85d/dependencies/bookworm/dynflow/copyright#L28-L44

https://github.com/theforeman/foreman-packaging/blob/ff40d9920a3b2f152dbc91d818ee7bbd2618b85d/dependencies/bookworm/hammer_cli/copyright#L23-L26

m-bucher:
@ogajduse thanks for the review. I have applied the changes you requested.

evgeni (changes requested):
this is missing an install file, so the package is empty

see e.g. https://github.com/theforeman/foreman-packaging/blob/deb/develop/plugins/ruby-foreman-ansible/debian/install

plugins/ruby-foreman-opentofu/debian/install
```diff
@@ -0,0 +1,4 @@
+foreman_opentofu.rb          usr/share/foreman/bundler.d
+cache                        usr/share/foreman/vendor
+webpack/foreman_opentofu     var/lib/foreman/public/webpack
```
evgeni: you said there is no webpack? :)

m-bucher: I should open my eyes when doing copy-paste :see_no_evil:

evgeni: https://www.youtube.com/watch?v=UNo2-viKfW8

plugins/ruby-foreman-opentofu/debian/control
```diff
@@ -0,0 +1,24 @@
+Source: ruby-foreman-opentofu
+Section: ruby
+Priority: optional
+Maintainer: ATIX AG <info@atix.de>
+Build-Depends: debhelper-compat (= 13),
+               gem2deb (>= 1),
+               rake,
+               foreman (>= 3.0.0),
```
evgeni, line 8: this is in sync with https://github.com/ATIX-AG/foreman_opentofu/blob/8a5617f9bf235eed80d7223b9336cd0855180824/lib/foreman_opentofu/engine.rb#L18, but kinda hard to believe :)

(no, I won't block on this ;) )

evgeni (approved):

PR #13035, opened 2026-02-12
Update pulpcore_version in katello-repos

zjhuntin:


evgeni (approved):
told ya ;)

packages/katello/katello-repos/katello-repos.spec
evgeni: need to bump it

zjhuntin: fixed

ogajduse (approved):

PR #13041, opened 2026-02-13
WIP: Add foreman ansible director plugin

nadjaheitmann:
https://github.com/ATIX-AG/foreman_ansible_director

This plugin add a novelty in development which is using Typescript for the front end. For this to work, we need to package some additional typescript-related packages. 

We were originally planning to put this plugin into Katello space because we plan to use some Katello-specific features in the future. However, at the moment we moved it to the plugin space. If it is a better idea to move it to Katello from the beginning, we can still change that. 

The Typescript config also requires to be loaded via webpack, so we need to add this to Foreman. Therefore, we have the following PR (already merged):
https://github.com/theforeman/foreman/compare/develop...ATIX-AG:foreman:develop_with_ts

We have also added two packages specifically into the Foreman space rather than the plugin space which allow the loading of TS files: https://github.com/theforeman/foreman-packaging/pull/13080


nadjaheitmann:
@theforeman/packaging I am trying to set up a new nodejs package and use the 'add_npm_package.sh' script. It runs well and I think the PR looks alright. However, the git annex cannot be resolved and it fails to build the package:

>[2026-02-16T11:41:31.373Z] TASK [setup_sources : Search annex for web link] *******************************
[2026-02-16T11:41:31.373Z] fatal: [nodejs-ts-loader]: FAILED! => 
[2026-02-16T11:41:31.373Z]     changed: false
[2026-02-16T11:41:31.373Z]     cmd: git annex whereis /home/jenkins/workspace/foreman-packaging-rpm-copr-pr-test/packages/foreman//nodejs-ts-loader/ansi-styles-4.3.0.tgz
[2026-02-16T11:41:31.373Z]         2>/dev/null | grep -q "web:"
[2026-02-16T11:41:31.373Z]     delta: '0:00:00.194576'
[2026-02-16T11:41:31.373Z]     end: '2026-02-16 11:41:31.309333'
[2026-02-16T11:41:31.373Z]     msg: non-zero return code
[2026-02-16T11:41:31.373Z]     rc: 1
[2026-02-16T11:41:31.373Z]     start: '2026-02-16 11:41:31.114757'
[2026-02-16T11:41:31.373Z]     stderr: ''
[2026-02-16T11:41:31.373Z]     stderr_lines: <omitted>
[2026-02-16T11:41:31.373Z]     stdout: ''
[2026-02-16T11:41:31.373Z]     stdout_lines: <omitted>

Does anyone have an idea how to fix this? Any hint appreciated!

nadjaheitmann:
@ogajduse https://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-rpm-copr-pr-test/detail/foreman-packaging-rpm-copr-pr-test/7055/pipeline/#step-176-log-65 The pipeline keeps failing with git-annex errors. Any idea if I need special permissions for git-annex? It is somehow weird, I never had problems with that in previous PRs. It's just with these bundled nodejs-type packages.

ogajduse:
[test rpm]

nadjaheitmann:
> [test rpm]

This won't work, I need to re-create the spec files :) I started it last week but then I did not get around to finish. Besides, this PR is also somehow blocked by the question raised where the foreman_ansible_director plugin and its smart-proxy plugin should live in the long run because of the Pulp dependencies we have in the smart-proxy part.

ogajduse:
> \[test rpm]

This was a scripted comment, using `gh` CLI, that I posted to every PR with a failed RPM check.

Regarding the npm build failures — the git-annex errors you're seeing are a red herring. obal handles those gracefully (`ignore_errors: true` + `addurl` fallback). The actual failure happens later during `%build` in the COPR mock chroot.

The root cause is that npm 10.x (shipped in RHEL 9 / EL9 mock chroots) deprecated and silently ignores the `--cache-min Infinity` flag used in the generated spec files. Without it, npm tries to reach `registry.npmjs.org` during install, which fails because mock has no network access. You can see this in the [COPR build log](https://download.copr.fedorainfracloud.org/results/@theforeman/plugins-nightly-staging-scratch-f947871e-c6f8-506d-a62b-d06e0071dda6/rhel-9-x86_64/10512785-nodejs-types-react/builder-live.log.gz) — npm emits `ENOTFOUND registry.npmjs.org`.

The fix is in [npm2rpm PR #87](https://github.com/nicholasgasior/npm2rpm/pull/87), which replaces the deprecated flags with `--offline --package-lock false --omit optional --install-strategy shallow`. So when you re-create the spec files, make sure to use a version of npm2rpm that includes that fix.

Regarding the Pulp dependencies in `rubygem-foreman_ansible_director` — I'll leave that decision to @Odilhao. I just want to flag that the current PR packages gem version 0.3.3 which pins `pulp_ansible_client >= 0.24.1, <= 0.25.0` and `pulpcore_client >= 3.73.0, < 3.74.0`, while the latest upstream gem (0.5.0) requires significantly newer versions (`>= 0.28.0` / `>= 3.85.0`). Also worth noting: `smart_proxy_ansible_director` has no Pulp dependencies at all — only `smart_proxy_dynflow`.

PR #13051, opened 2026-02-17
hammer_cli_foreman: Update ruby-hammer-cli to 3.18.0

ogajduse:
Bot misses updates to the `control` file during the automated version bump. This patch fixes the version of ruby-hammer-cli that hammer_cli_foreman depends on.

### DIFF
--- a/dependencies/bookworm/hammer_cli_foreman/changelog
+++ b/dependencies/bookworm/hammer_cli_foreman/changelog
@@ -1,3 +1,9 @@
+ruby-hammer-cli-foreman (3.18.0-2) stable; urgency=low
+
+  * 3.18.0-2 Dependency update ruby-hammer-cli (>= 3.18.0)
+
+ -- Ondřej Gajdušek <ogajduse@redhat.com>  Tue, 17 Feb 2026 16:09:39 +0100
+
 ruby-hammer-cli-foreman (3.18.0-1) stable; urgency=low
 
   * 3.18.0 released
--- a/dependencies/bookworm/hammer_cli_foreman/control
+++ b/dependencies/bookworm/hammer_cli_foreman/control
@@ -12,7 +12,7 @@ Package: ruby-hammer-cli-foreman
 Architecture: all
 XB-Ruby-Versions: ${ruby:Versions}
 Depends: ${shlibs:Depends}, ${misc:Depends}, ruby | ruby-interpreter,
-         ruby-hammer-cli (>= 3.17.0),
+         ruby-hammer-cli (>= 3.18.0),
          ruby-apipie-bindings (>= 0.7.0),
          ruby-jwt (>= 2.2.1),
          ruby-rest-client (>= 1.8.0),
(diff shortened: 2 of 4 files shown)


ogajduse:
[test deb]

evgeni:
But there is no such bump in the gem?
https://github.com/theforeman/hammer-cli-foreman/blob/d8b89dc7c7178c9203c7385778cc231e75e7617c/hammer_cli_foreman.gemspec#L31

ogajduse:
@ofedoren Could you please chime in here? I wonder why we bumped the dependency to 3.17.0 here in deb packaging, but not in the gemspec.

ref: https://github.com/theforeman/foreman-packaging/commit/3f02a297ad4a33c1bf562a3ff27a6590971b4ca2

ofedoren:
> @ofedoren Could you please chime in here? I wonder why we bumped the dependency to 3.17.0 here in deb packaging, but not in the gemspec.
> 
> ref: [3f02a29](https://github.com/theforeman/foreman-packaging/commit/3f02a297ad4a33c1bf562a3ff27a6590971b4ca2)

I'm afraid it's just an oversight. Ideally hammer-cli and hammer-cli-foreman should have the same versions (at least X and Y in X.Y.z), so hammer-cli-foreman should ideally depend on the same hammer-cli, but since there were no major changes and we still have an open pin, it shouldn't be a problem.

ogajduse:
@ofedoren I understand the current situation. But the current situation raises a question in my mind, whether `Manual step: Update deps in the gemspec if neccessary"` should be automated. Is it worthwhile to invest the effort in automating it?

ref: https://github.com/theforeman/hammer-cli-foreman/blob/d8b89dc7c7178c9203c7385778cc231e75e7617c/rel-eng/gem_release.ipynb#L211

pcreech (approved):

ogajduse:
By merging this one, we'll get the dependencies out of sync. Based on Oleh's comment, "Ideally hammer-cli and hammer-cli-foreman should have the same versions", I'm going to merge it.

ofedoren:
> @ofedoren I understand the current situation. But the current situation raises a question in my mind, whether `Manual step: Update deps in the gemspec if neccessary"` should be automated. Is it worthwhile to invest the effort in automating it?
> 
> ref: https://github.com/theforeman/hammer-cli-foreman/blob/d8b89dc7c7178c9203c7385778cc231e75e7617c/rel-eng/gem_release.ipynb#L211

I guess it's up to maintainers? I mean, even the message states "if necessary", for a long time hammer-cli didn't have any changes that are required in hammer-cli-foreman to correctly function, so it just works with few older/latest version, and again, we have an open pin, so h-c-foreman will try to use latest hammer-cli if it's available, we just need to ensure that for h-c-foreman version there is hammer-cli version. We always treated them as one and given that all the other plugins depend on hammer-cli+hammer-cli-foreman combo instead of just one of them, it would be easier to just merge them. I doubt anyone uses hammer-cli itself as a framework :shrug:

PR #13054, opened 2026-02-17
Update rubygem-kafo_wizards to 1.0.0

ogajduse:
This update was generated by `./bump_rpm.sh packages/foreman/rubygem-kafo_wizards 1.0.0`. This patch changes the spec template for this package. This now allows `list_updatable_packages` find an update for future rubygem-kafo_wizards releases.

PR #13055, opened 2026-02-17
Add rubygem-logger package

ogajduse:
This patch adds the rubygem `logger`. This is a dependency of #13054 as it was introduced as a new dependency in https://github.com/theforeman/kafo_wizards/commit/d5f394c1b9426036e1274a994ae9e86b6988f241


package_manifest.yaml
```diff
@@ -156,6 +156,7 @@ foreman_core_packages:
       source_location: https://ci.theforeman.org/job/foreman-develop-source-release
       source_system: jenkins
 
+    rubygem-logger: {}
```
zjhuntin: lets put this right under rubygem-kafo_wizards, I think that way we just know they're related.

ogajduse: I agree that the current placement is far from ideal. Since logger is a general-purpose Ruby logging utility (not Rails-specific, not Hammer-specific), I propose moving it to the `ruby_core_packages_tier1` section.

ogajduse:
logger was a default rubygem until 3.4. But since we're on Ruby 3.0, quoting Evgeni here, we "need to play the old %gemspec_remove_dep game)" in ruby-kafo-wizards
.

PR #13080, opened 2026-02-20
Add typescript loader to foreman

nadjaheitmann:
https://github.com/ATIX-AG/foreman_ansible_director

This plugin add a novelty in Foreman development which is using Typescript for the front end. For this to work, we need to package some additional typescript-related packages.

The related Foreman PR can be found here:
https://github.com/theforeman/foreman/pull/10899

Summary (from RFC):

- Integration into Foreman nightly was done in less than 100 line diff due to Webpack already providing a very wrong base.
- Usage of TypeScript remains optional
- There are no changes required to the way Foreman packages are built and delivered.
- No existing workflow or plugins are broken

The related discourse RFC:
https://community.theforeman.org/t/rfc-opt-in-typescript-support-for-the-foreman-frontend-stack/45928


nadjaheitmann:
@theforeman/packaging Can we get a review here, please?

packages/foreman/nodejs-ts-loader/nodejs-ts-loader.spec
```diff
@@ -0,0 +1,107 @@
+%{?scl:%scl_package nodejs-%{npm_name}}
+%{!?scl:%global pkg_name %{name}}
```
adamruzicka: This was generated by one of the templates, right? SCLs are a thing of the distant past :/

nadjaheitmann: I know, all of the npm package spec files look like this 🙈 The spec file was generated with this script: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/add_npm_package.sh 

I assume it is this line and npm2rpm tool that also generates the spec file: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/add_npm_package.sh#L60 

I have locally checked out the tool and ran the script. There is even --scl' option for npm2rpm if I see it correctly: https://github.com/theforeman/npm2rpm/blob/2acb4a459a75a36197cc13886d7aa08b9cf83302/bin/npm2rpm.js#L28 

But this option is not set in the script.

ekohl: > This was generated by one of the templates, right? SCLs are a thing of the distant past :/

https://github.com/theforeman/npm2rpm/pull/87

nadjaheitmann: I updated my PR using @ekohls PR.

nadjaheitmann:
@theforeman/packaging Any chance we get a review here?

ogajduse (commented):
I don't see a reason to block this PR from merging. However, before I approve it here, I'd like [to discuss first](https://community.theforeman.org/t/rfc-opt-in-typescript-support-for-the-foreman-frontend-stack/45928/14) what work is expected from @theforeman/packaging to be done when/if we merge this.

nadjaheitmann:
> I don't see a reason to block this PR from merging. However, before I approve it here, I'd like [to discuss first](https://community.theforeman.org/t/rfc-opt-in-typescript-support-for-the-foreman-frontend-stack/45928/14) what work is expected from @theforeman/packaging to be done when/if we merge this.

Thanks @ogajduse , good call! As mentioned in the discourse, I assume that the packages need regular updates and I suggest that @theforeman/atix jumps in here to leverage the workload from the packaging team or jump in if something really needs to be fixed. I added a third commit that adds the ATIX team as code owners for those two packages.

ogajduse (approved):
Thank you, @nadjaheitmann, for putting the effort into this. Everything looks good to me. In case the patch was (co-)created with https://github.com/theforeman/npm2rpm/pull/87, should we merge it first, release a new npm2rpm version, and make sure we use the new version in foreman-packaging, before we merge this PR?

nadjaheitmann:
> Thank you, @nadjaheitmann, for putting the effort into this. Everything looks good to me. In case the patch was (co-)created with [theforeman/npm2rpm#87](https://github.com/theforeman/npm2rpm/pull/87), should we merge it first, release a new npm2rpm version, and make sure we use the new version in foreman-packaging, before we merge this PR?

Yes, this repo would profit so much from a new npm2rpm release.

ogajduse:
@ekohl How do you see us proceeding with https://github.com/theforeman/npm2rpm/pull/87? Would you find time to merge the PR and release a new version? Would you like @theforeman/packaging to handle something?

CODEOWNERS
```diff
@@ -17,6 +17,9 @@
 /packages/plugins/rubygem-smart_proxy_request_forwarder/ @theforeman/packaging @theforeman/atix
 /packages/plugins/*_cve_scanner/ @theforeman/packaging @theforeman/atix
 /packages/plugins/*_bootdisk/ @theforeman/packaging @theforeman/bootdisk
+/packages/foreman/nodejs-ts-loader/ @theforeman/packaging @theforeman/atix
+/packages/foreman/nodejs-typescript/ @theforeman/packaging @theforeman/atix
+
```
maximiliankolb, line 22:
```suggestion
```

nadjaheitmann: Fixed. Thanks!

ogajduse:
Thank you, @nadjaheitmann! Let's get it in!

PR #13103, opened 2026-02-25
DEB: Release Foreman 3.17.1

ogajduse:
Release Foreman 3.17.1

### DIFF
--- a/debian/bookworm/foreman-installer/changelog
+++ b/debian/bookworm/foreman-installer/changelog
@@ -1,3 +1,9 @@
+foreman-installer (3.17.1-1) stable; urgency=low
+
+  * 3.17.1 released
+
+ -- Ondřej Gajdušek <ogajduse@redhat.com>  Wed, 25 Feb 2026 09:46:46 +0100
+
 foreman-installer (3.17.0-1) stable; urgency=low
 
   * 3.17.0 released
--- a/debian/bookworm/foreman-proxy/changelog
+++ b/debian/bookworm/foreman-proxy/changelog
@@ -1,3 +1,9 @@
+foreman-proxy (3.17.1-1) stable; urgency=low
+
+  * 3.17.1 released
+
+ -- Ondřej Gajdušek <ogajduse@redhat.com>  Wed, 25 Feb 2026 09:46:46 +0100
+
 foreman-proxy (3.17.0-1) stable; urgency=low
 
   * 3.17.0 released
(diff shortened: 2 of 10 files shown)


evgeni:
Meh, something started to pull in `fiddle` which fails to build now.

I've seen a similar failure in https://github.com/theforeman/foremanctl/actions/runs/22386543313/job/64798310513?pr=383

evgeni:
Ahahahaha. `locale` of all things: https://github.com/ruby-gettext/locale/compare/2.1.4...2.1.5

evgeni:
Pin locale or install libffi even tho we don't need fiddle working, what's the smaller evil?

ogajduse:
> Pin locale or install libffi even tho we don't need fiddle working, what's the smaller evil?

reading from their gemspec... `# This is needed only for Windows.`  This made me cry. 

I'd rather install libffi and just live with an extra dependency rather than having to solve the well-known pain of forgotten pin. What do you say, @evgeni?

evgeni:
Given this is a stable release, I'd pin it here and do the new dependency in develop first.

evgeni:
It's not the first time we do debian-specific pins:
https://github.com/theforeman/foreman-packaging/blob/fbf96731961a2f6745e7bbbf50683939ecc764f8/debian/bookworm/foreman/rules#L25-L35

ogajduse:
oh, okay. That makes sense. On it.

debian/bookworm/foreman/control
```diff
@@ -5,7 +5,7 @@ Priority: extra
 Standards-Version: 3.9.1
 Build-Depends: debhelper (>= 9.20160709), ruby-dev (>= 1:2.7), git,
                zlib1g-dev, libpq-dev, pkg-config, libvirt-dev, libyaml-dev,
-               build-essential, curl, nodejs (>= 8.9), npm,
+               build-essential, curl, nodejs (>= 8.9), npm, libffi-dev,
```
evgeni, line 8: Now, this is the "new dep" patch, not the "pin" patch, but: This will break installation as you also need to add `libffi-dev` to the `Depends` of the `foreman` package.

Sadly, the way our Debian packages are architectured, we store the gem source in the cache, but the build needs to happen twice:
1. at package build time to build assets
2. at package installation time to build the binary that the user will actually use

evgeni (approved):

PR #13104, opened 2026-02-25
Add libffi-dev build dependency for locale gem 2.1.5 compatibility

ogajduse:
Based on discussion in #13103, this patch introduces new build dependency to satisfy locale 2.1.5 gem's dependencies.

### DIFF
--- a/debian/bookworm/foreman/control
+++ b/debian/bookworm/foreman/control
@@ -5,7 +5,7 @@ Priority: extra
 Standards-Version: 3.9.1
 Build-Depends: debhelper (>= 9.20160709), ruby-dev (>= 1:2.7), git,
                zlib1g-dev, libpq-dev, pkg-config, libvirt-dev, libyaml-dev,
-               build-essential, curl, nodejs (>= 22.0), npm,
+               build-essential, curl, nodejs (>= 22.0), npm, libffi-dev,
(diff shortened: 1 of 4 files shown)


debian/bookworm/foreman/control
```diff
@@ -5,7 +5,7 @@ Priority: extra
 Standards-Version: 3.9.1
 Build-Depends: debhelper (>= 9.20160709), ruby-dev (>= 1:2.7), git,
                zlib1g-dev, libpq-dev, pkg-config, libvirt-dev, libyaml-dev,
-               build-essential, curl, nodejs (>= 22.0), npm,
+               build-essential, curl, nodejs (>= 22.0), npm, libffi-dev,
```
evgeni, line 8: This will break installation as you also need to add `libffi-dev` to the `Depends` of the `foreman` package.

Sadly, the way our Debian packages are architectured, we store the gem source in the cache, but the build needs to happen twice:
1. at package build time to build assets
2. at package installation time to build the binary that the user will actually use

evgeni:
[test deb]

debian/bookworm/foreman/control
ogajduse: TIL. So, `libffi-dev` is now in `Build-Depends` as well as `Depends`.

evgeni (approved):

PR #13156, opened 2026-03-04
KubeVirt related backports for 3.18

evgeni:


evgeni:
ok to test

arvind4501:
I don't see any thing that requires specific testing, its a backport and one thing that this will make sure that these packages also land in next snap for 6.19 right?

evgeni:
Yes, and 3.18 GA when it comes

arvind4501 (approved):
Thanks @evgeni !

PR #13173, opened 2026-03-08
Remove rubygem-jgrep

ekohl:
This package is unused since foreman_hook was dropped


ogajduse (approved):

ekohl:
Did you also delete it from copr? I still see https://copr.fedorainfracloud.org/coprs/g/theforeman/plugins-nightly-staging/package/rubygem-jgrep. https://github.com/theforeman/foreman-packaging/tree/rpm/develop?tab=readme-ov-file#howto-removing-a-package has the right copr command.

PR #13181, opened 2026-03-10
DEB: Release Foreman 3.18.0

ogajduse:
Release Foreman 3.18.0

### DIFF
--- a/debian/bookworm/foreman/changelog
+++ b/debian/bookworm/foreman/changelog
@@ -1,3 +1,9 @@
+foreman (3.18.0-1) stable; urgency=low
+
+  * 3.18.0 released
+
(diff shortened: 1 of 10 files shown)


evgeni:
look, you need to fiddle with the locks, again!

evgeni (approved):

debian/bookworm/foreman/changelog
```diff
@@ -1,3 +1,15 @@
+foreman (3.18.0-3) stable; urgency=low
```
evgeni: wait, no, we don't need -3 here. -1 below is enough, merge the changelogs

evgeni (changes requested):

debian/bookworm/foreman/changelog
ogajduse: Good catch. Fixed

evgeni (approved):

PR #13203, opened 2026-03-11
Rocket plugin translation updates for 3.18

stejskalleos:


packages/plugins/rubygem-foreman_bootdisk/rubygem-foreman_bootdisk.spec
```diff
@@ -4,7 +4,7 @@
 %global foreman_min_version 3.15
 
 Name: rubygem-%{gem_name}
-Version: 23.1.2
+Version: 23.2.2
```
evgeni, line 7: There seem to be picks missing between those, which led to the old `.gem` not being deleted.

stejskalleos: I see, I didn't include the in-between versions.

evgeni (approved):

PR #13075, opened 2026-02-19
Bump required foreman version in ruby-foreman-templates control

adamlazik1:

PR #13066, opened 2026-02-18
Update ruby-foreman-templates to 11.0.0

adamlazik1:
(cherry picked from commit fb89565ec83093edf070f4ee487a73d260a0c1b1)

PR #13216, opened 2026-03-11
Update ruby-foreman-templates to 11.0.1

adamlazik1:


adamlazik1:
Note: the first commit is just to prevent merge conflicts, the required foreman version was already updated but the changelog was missing an entry.

plugins/ruby-foreman-templates/debian/changelog
```diff
@@ -1,3 +1,15 @@
+ruby-foreman-templates (11.0.1-1) stable; urgency=low
+
+  * 11.0.1 released
+
+ -- Foreman Packaging Automation <packaging@theforeman.org>  Wed, 11 Mar 2026 21:49:17 +0000
+
+ruby-foreman-templates (11.0.0-2) stable; urgency=low
+
+  * Bump required foreman version
```
adamruzicka: Really, where?

adamlazik1: I can remove the first commit but since there is a merge conflict now there might be more again in the future. This discrepancy was created by https://github.com/theforeman/foreman-packaging/pull/13066 vs https://github.com/theforeman/foreman-packaging/pull/13075. Or I could rename the first commit to something like "Sync changelog with deb/develop"?

adamruzicka: > I can remove the first commit but since there is a merge conflict now there might be more again in the future.

Please do. Since we're adding to the top of the file, I wouldn't expect any more clashes as the "change space" is now in sync between develop and 3.18.

adamlazik1: Alright, dropped.

adamruzicka (approved):

PR #13218, opened 2026-03-11
Update ruby-foreman-tasks to 11.1.1

adamlazik1:


adamruzicka (approved):

plugins/ruby-foreman-tasks/debian/changelog
```diff
@@ -1,3 +1,9 @@
+ruby-foreman-tasks (11.1.1-1) stable; urgency=low
+
+  * 11.1.1 released
+
+ -- alazik <alazik@redhat.com>  Thu, 12 Mar 2026 00:23:44 +0100
```
adamruzicka: The convention is to use full names

adamlazik1: Thanks, fixed.

adamruzicka (approved):

PR #13219, opened 2026-03-11
Update rubygem-foreman-tasks to 11.1.1

adamlazik1:


packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec
```diff
@@ -152,6 +152,9 @@ type foreman-selinux-relabel >/dev/null 2>&1 && foreman-selinux-relabel 2>&1 >/d
 %{foreman_plugin_log}
 
 %changelog
+* Wed Mar 11 2026 alazik <alazik@redhat.com> - 11.1.1-1
```
adamruzicka: The convention is to use full names

adamlazik1: Thanks, fixed.

adamruzicka (approved):

PR #13252, opened 2026-03-17
Require kafo >= 7.7

evgeni:
See https://github.com/theforeman/foreman-installer/pull/1056


evgeni:
repoclosure failure is because https://github.com/theforeman/foreman-packaging/pull/13222 is not yet in a published repo

ogajduse (approved):

PR #13253, opened 2026-03-17
Require puppet/openvox >= 8

evgeni:

### DIFF
--- a/packages/foreman/foreman-installer/foreman-installer.spec
+++ b/packages/foreman/foreman-installer/foreman-installer.spec
@@ -1,4 +1,4 @@
-%global release 1
+%global release 2
 %global prereleasesource develop
 %global prerelease %{?prereleasesource}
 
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman-installer/foreman-installer.spec
```diff
@@ -16,13 +16,13 @@ BuildArch:  noarch
 
 Requires:   curl
 Requires:   hostname
-Requires:   (openvox-agent >= 7.35.0 or puppet-agent >= 7.0.0)
+Requires:   (openvox-agent >= 8.23.1 or puppet-agent >= 8.0.0)
```
evgeni, line 19: 8.23.1 is the one we also use in our module metadata as the baseline

ekohl (approved):

evgeni:
I am going to pretend @ekohl approved this on behalf of @theforeman/packaging :see_no_evil:

PR #13266, opened 2026-03-19
Fixes #39161 - Migrate Foreman DEB packaging to use consolidated cron…

evgeni:
… entrypoints


evgeni:
This is a port of @archanaserver's #13001 to Debian

ogajduse (approved):

PR #13313, opened 2026-03-28
Fixes #39189 - backport CVE-2026-33658 fix for activestorage 7.0.10

jakduch:
## Summary

**Redmine:** https://projects.theforeman.org/issues/39189

Backport of the upstream security fix for CVE-2026-33658 (Active Storage DoS via HTTP Range header) to the rubygem-activestorage 7.0.10 package.

Active Storage proxy controller does not limit the number of byte ranges in an HTTP Range header. A request with thousands of small ranges causes disproportionate CPU usage, resulting in a potential DoS vulnerability.

This adds a patch that limits range requests to a single range by default via `ActiveStorage.streaming_max_ranges = 1`.

## Changes

- Added `CVE-2026-33658-limit-range-requests.patch` based on upstream commit rails/rails@b8a1665
- Updated spec to apply patch via `%autosetup -p1`
- Bumped release to 7.0.10-2

## References

- Upstream commit: https://github.com/rails/rails/commit/b8a1665824a43d71cd6406cf9adcae842ceb1c22
- Advisory: https://github.com/rails/rails/security/advisories/GHSA-p9fm-f462-ggrg

### DIFF
--- a/packages/foreman/rubygem-activestorage/CVE-2026-33658-limit-range-requests.patch
+++ b/packages/foreman/rubygem-activestorage/CVE-2026-33658-limit-range-requests.patch
@@ -1,0 +1,31 @@
+--- a/app/controllers/concerns/active_storage/streaming.rb
++++ b/app/controllers/concerns/active_storage/streaming.rb
+@@ -15,6 +15,7 @@
+       ranges = Rack::Utils.get_byte_ranges(range_header, blob.byte_size)
+
+       return head(:range_not_satisfiable) if ranges.blank? || ranges.all?(&:blank?)
++      return head(:range_not_satisfiable) if ranges.length > ActiveStorage.streaming_max_ranges
+
+       if ranges.length == 1
+         range = ranges.first
+--- a/lib/active_storage.rb
++++ b/lib/active_storage.rb
+@@ -360,6 +360,18 @@
+   mattr_accessor :replace_on_assign_to_many, default: false
+   mattr_accessor :track_variants, default: false
+
++  singleton_class.attr_accessor :checksum_implementation
++  @checksum_implementation = OpenSSL::Digest::MD5
++  begin
++    @checksum_implementation.hexdigest("test")
++  rescue # OpenSSL may have MD5 disabled
++    require "digest/md5"
++    @checksum_implementation = Digest::MD5
++  end
++
++  singleton_class.attr_accessor :streaming_max_ranges
++  @streaming_max_ranges = 1
++
+   mattr_accessor :video_preview_arguments, default: "-y -vframes 1 -f image2"
+
+   mattr_accessor :silence_invalid_content_types_warning, default: false
--- a/packages/foreman/rubygem-activestorage/rubygem-activestorage.spec
+++ b/packages/foreman/rubygem-activestorage/rubygem-activestorage.spec
@@ -3,11 +3,12 @@
 
 Name: rubygem-%{gem_name}
 Version: 7.0.10
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: Local and cloud file storage framework
 License: MIT
 URL: https://rubyonrails.org
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+Patch0: CVE-2026-33658-limit-range-requests.patch
 
 # start specfile generated dependencies
 Requires: ruby >= 2.7.0
@@ -29,7 +30,7 @@ BuildArch: noarch
 Documentation for %{name}.
 
 %prep
-%setup -q -n  %{gem_name}-%{version}
+%autosetup -p1 -n  %{gem_name}-%{version}
 
 %build
 # Create the gem as gem install only works on a gem file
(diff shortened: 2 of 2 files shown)


ogajduse:
ok to test

ogajduse (commented):
The RPM spec file looks good to me. I see the patch you included missed 7 lines from the original PR linked in the PR description. Is there a reason to use only a part of the patch you linked in the PR description?

EDIT: here is the missed part: https://github.com/rails/rails/commit/b8a1665824a43d71cd6406cf9adcae842ceb1c22#diff-0e2e89196785734abe50e5ae4c8f9eb8a9ad6b4f69187b478d4c7855d866844fR365-R372

jakduch:
Thanks for the catch! I had the rest of the upstream fix sitting in my clipboard but somehow managed to not paste it in. Updated the patch to include the full upstream fix with `ranges_valid?`, `streaming_max_ranges` and `streaming_chunk_max_size`.

packages/foreman/rubygem-activestorage/CVE-2026-33658-limit-range-requests.patch
```diff
@@ -0,0 +1,41 @@
...
++++ b/lib/active_storage.rb
+@@ -360,6 +360,12 @@
+   mattr_accessor :replace_on_assign_to_many, default: false
+   mattr_accessor :track_variants, default: false
+ 
++  singleton_class.attr_accessor :streaming_max_ranges
++  @streaming_max_ranges = 1
++
++  singleton_class.attr_accessor :streaming_chunk_max_size
++  @streaming_chunk_max_size = 100.megabytes
```
Odilhao: Why are we adding `@streaming_chunk_max_size = 100.megabytes` here? 

The patch upstream does not provide this part.

packages/foreman/rubygem-activestorage/CVE-2026-33658-limit-range-requests.patch
```diff
@@ -0,0 +1,41 @@
...
++
+     # Stream the blob from storage directly to the response. The disposition can be controlled by setting +disposition+.
+     # The content type and filename is set directly from the +blob+.
+     def send_blob_stream(blob, disposition: nil) # :doc:
+--- a/lib/active_storage.rb
++++ b/lib/active_storage.rb
+@@ -360,6 +360,12 @@
+   mattr_accessor :replace_on_assign_to_many, default: false
+   mattr_accessor :track_variants, default: false
+
```
Odilhao: Why are not bringing https://github.com/rails/rails/commit/b8a1665824a43d71cd6406cf9adcae842ceb1c22#diff-0e2e89196785734abe50e5ae4c8f9eb8a9ad6b4f69187b478d4c7855d866844fR365-R369

```ruby
  singleton_class.attr_accessor :checksum_implementation
  @checksum_implementation = OpenSSL::Digest::MD5
  begin
    @checksum_implementation.hexdigest("test")
  rescue # OpenSSL may have MD5 disabled
    require "digest/md5"
    @checksum_implementation = Digest::MD5
  end
``` 
as well?

Odilhao (changes requested):

ogajduse:
I'm not sure what the author's intention was behind not matching the patch 1:1, but here is the version of the patch I propose:

```patch
--- a/app/controllers/concerns/active_storage/streaming.rb
+++ b/app/controllers/concerns/active_storage/streaming.rb
@@ -15,6 +15,7 @@
       ranges = Rack::Utils.get_byte_ranges(range_header, blob.byte_size)

       return head(:range_not_satisfiable) if ranges.blank? || ranges.all?(&:blank?)
+      return head(:range_not_satisfiable) if ranges.length > ActiveStorage.streaming_max_ranges

       if ranges.length == 1
         range = ranges.first
--- a/lib/active_storage.rb
+++ b/lib/active_storage.rb
@@ -360,6 +360,18 @@
   mattr_accessor :replace_on_assign_to_many, default: false
   mattr_accessor :track_variants, default: false

+  singleton_class.attr_accessor :checksum_implementation
+  @checksum_implementation = OpenSSL::Digest::MD5
+  begin
+    @checksum_implementation.hexdigest("test")
+  rescue # OpenSSL may have MD5 disabled
+    require "digest/md5"
+    @checksum_implementation = Digest::MD5
+  end
+
+  singleton_class.attr_accessor :streaming_max_ranges
+  @streaming_max_ranges = 1
+
   mattr_accessor :video_preview_arguments, default: "-y -vframes 1 -f image2"

   mattr_accessor :silence_invalid_content_types_warning, default: false
```

jakduch:
Updated the patch to match your proposed version. Thanks for the cleaner approach.

Regarding the original patch - my intention was to include the full upstream fix for completeness (the `ranges_valid?` method and `streaming_chunk_max_size` were additional hardening layers from the same commit). But I agree that the minimal fix targeting just the CVE is better for a backport - less risk of unintended side effects.

ogajduse:
[test rpm]

ogajduse:
Closing this PR as Foreman is not affected by CVE-2026-33658.

While Foreman ships `rubygem-activestorage` as a dependency, the ActiveStorage engine is not loaded at runtime. It is commented out in Foreman's application configuration:

https://github.com/theforeman/foreman/blob/develop/config/application.rb#L12

```ruby
# 'active_storage/engine',
```

Since the ActiveStorage engine is not initialized, the vulnerable code path (Range header processing in ActiveStorage's proxy controller) is not reachable during normal operation. This means the CVE fix is not needed at this time.

**Related issues:**
- Foreman issue https://projects.theforeman.org/issues/39189 - closed as not affected
- Future work to enable ActiveStorage is tracked in https://projects.theforeman.org/issues/35330

If/when ActiveStorage is enabled in Foreman, this CVE will need to be reassessed and the fix applied at that time.

PR #13315, opened 2026-03-29
Fixes #39190 - Apply CVE-2026-33176 patch for activesupport 7.0.10

jakduch:
Backport of the upstream fix for CVE-2026-33176 (ActiveSupport DoS via large scientific notation strings in number helpers).

ActiveSupport number helpers accept strings with scientific notation (e.g. 1e10000), which BigDecimal expands into huge decimal representations causing excessive memory/CPU consumption.

This patch modifies NumberConverter#valid_bigdecimal to reject strings containing scientific notation characters (d/e), matching the upstream fix in Rails 7.2+.

Upstream commit: rails/rails@ebd6be18
Advisory: GHSA-2j26-frm8-cmj9
Redmine: https://projects.theforeman.org/issues/39190

### DIFF
--- a/packages/foreman/rubygem-activesupport/rubygem-activesupport.spec
+++ b/packages/foreman/rubygem-activesupport/rubygem-activesupport.spec
@@ -3,11 +3,13 @@
 
 Name: rubygem-%{gem_name}
 Version: 7.0.10
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: A toolkit of support libraries and Ruby core extensions extracted from the Rails framework
 License: MIT
 URL: https://rubyonrails.org
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+
+Patch0: cve-2026-33176-number-converter-dos.patch
 
 # start specfile generated dependencies
 Requires: ruby >= 2.7.0
@@ -39,7 +41,7 @@ BuildArch: noarch
 Documentation for %{name}.
 
 %prep
-%setup -q -n  %{gem_name}-%{version}
+%autosetup -p1 -n %{gem_name}-%{version}
 
 %gemspec_remove_dep -g base64
 %gemspec_remove_dep -g drb
@@ -48,6 +50,7 @@ Documentation for %{name}.
 %gemspec_remove_dep -g logger ">= 1.4.2"
 %gemspec_remove_dep -g securerandom ">= 0.3"
 %gemspec_remove_dep -g benchmark ">= 0.3"
+
 
 %build
 # Create the gem as gem install only works on a gem file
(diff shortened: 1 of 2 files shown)


evgeni:
What's so special about CVE-2026-33176 that we need to backport the fix for it?

packages/foreman/rubygem-activestorage/rubygem-activestorage.spec
```diff
@@ -3,11 +3,12 @@
 
 Name: rubygem-%{gem_name}
 Version: 7.0.10
-Release: 1%{?dist}
+Release: 2%{?dist}
 Summary: Local and cloud file storage framework
 License: MIT
 URL: https://rubyonrails.org
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
+Patch0: CVE-2026-33658-limit-range-requests.patch
```
evgeni: This is the wrong patch.

jakduch:
Sorry about the extra commit - will rebase to drop the activestorage change and keep only the activesupport patch.

Re: why this CVE - it is a DoS vulnerability where an attacker can send scientific notation strings (e.g. `1e10000`) to any endpoint that uses ActiveSupport number helpers, causing BigDecimal to allocate excessive memory. Since Foreman is often publicly exposed, this seemed worth patching. But happy to close this if there is a different plan for addressing it (e.g. upgrading Rails).

ogajduse:
ok to test

jakduch:
Fixed the patch - the original one was based on an older activesupport where `valid_bigdecimal` used a simple `BigDecimal(number)` call. In 7.0.10 the method uses a `case/when` block with `BigDecimal(number, exception: false)`. Updated the patch to match the actual 7.0.10 source, consistent with the upstream fix in 7.2.3.1.

packages/foreman/rubygem-activesupport/rubygem-activesupport.spec
```diff
@@ -49,6 +51,8 @@ Documentation for %{name}.
 %gemspec_remove_dep -g securerandom ">= 0.3"
 %gemspec_remove_dep -g benchmark ">= 0.3"
 
+%patch0 -p1
+
```
ogajduse, line 54: Based on https://rpm.org/docs/latest/manual/spec.html, `%patch0` macro is deprecated.
> %patch1 (removed since rpm >= 4.20)

Could you consider using the `%autosetup` macro instead?


```suggestion
%prep
%autosetup -p1 -n %{gem_name}-%{version}

%gemspec_remove_dep -g base64
%gemspec_remove_dep -g drb
%gemspec_remove_dep -g mutex_m
%gemspec_remove_dep -g bigdecimal
%gemspec_remove_dep -g logger ">= 1.4.2"
%gemspec_remove_dep -g securerandom ">= 0.3"
%gemspec_remove_dep -g benchmark ">= 0.3"

```

packages/foreman/rubygem-activestorage/rubygem-activestorage.spec
jakduch: Resolved - the extra activestorage commit was dropped during rebase.

packages/foreman/rubygem-activesupport/rubygem-activesupport.spec
jakduch: Good call, switched to `%autosetup -p1`. Thanks!

ogajduse (approved):
Thank you, @jakduch!

PR #13332, opened 2026-04-02
foreman_opentofu directories and selinux

sbernhard:


m-bucher:
Integrated into #13343

PR #12157, opened 2025-06-26
Update reqs for Remove @theforeman/vendor

ShimShtein:
Second part of removing foreman vendor.

This part is updating the foreman.spec file. Should be merged after https://github.com/theforeman/foreman-packaging/pull/12145

PR #13339, opened 2026-04-07
Bump nodejs-uuid to 13.0.0

jakduch:
## Summary
Update nodejs-uuid from 3.4.0 to 13.0.0 (RFC 9562, adds v6 and v7 UUIDs).

Required by theforeman/foreman#10944 which migrates foreman core from uuid v1 to v7.

### Changes
- nodejs-uuid.spec: 3.4.0 to 13.0.0, updated installed files (dist/ + dist-node/ instead of top-level v1.js..v5.js, removed bin/uuid CLI)
- Source tarball: uuid-3.4.0.tgz to uuid-13.0.0.tgz

### Question for maintainers
After theforeman/foreman#10944 merges, foreman.spec will need its Requires/BuildRequires updated from npm(uuid) >= 3.3.2 < 4.0.0 to >= 13.0.0 < 14.0.0. Should that be a separate PR here (like #12157 was for the vendor removal), or is there a preferred way to coordinate this?

Packages that bundle their own uuid 3.x (patternfly-react, patternfly-react-extensions) are unaffected.


dgoetz:
Ok to test

dgoetz:
Test "rpm" complains about `bogus date in %changelog: Mon Apr 07 2026` as this should be `Tue Apr 07 2026` and `no-%build-section` which is fine as there is no build happening and also the template does not provide this empty section.

jakduch:
Fixed - the changelog date is now `Tue Apr 07 2026` and I also corrected a typo in my name. :D

CI should pass now, thanks for catching that!

PR #13390, opened 2026-04-22
Update pulp bindings for Pulp 3.105

sjha4:
Katello gemspec update PR: https://github.com/Katello/katello/pull/11673/


dgoetz:
ok to test

qcjames53 (approved):
These binding versions all match the correct Y streams and follow the latest Z streams. The changes look good to me.

dgoetz (approved):
Also from packaging everything looks good and tests are green. Should this be merged now?

sjha4:
~~We should be good to merge..The other PRs are also good to go..~~ Let's get the katello updates in first so we don't get ahead on packaging.. https://github.com/Katello/katello/pull/11673

qcjames53:
VCR PR: https://github.com/Katello/katello/pull/11715

zjhuntin:
@sjha4 can we also get a rebase of this, looks like we have a slight conflict. I'll keep an eye on the VCR PR as well.

ianballou:
@zjhuntin the rebase here seems to be set. VCR PR should be all set too.

packages/katello/rubygem-pulpcore_client/pulpcore_client-3.105.4.gem
zjhuntin, line 1: One fix so we can get a green test: need run `git rm pulpcore_client-3.85.17.gem` and then I think we'll be good to go 😄

ianballou: I do wonder if this will permanently be a red build because Katello will complain about not seeing the version requirements it expects, since the Katello requirements get updated in a separate PR.

zjhuntin: It may fail on repoclosure, but this would have prevented being able to build the RPM at all.

sjha4: Updated PR to rm pulpcore 3.85.17

ianballou: > It may fail on repoclosure, but this would have prevented being able to build the RPM at all.

Indeed, just a side note.

PR #13416, opened 2026-04-29
force all packages to be format 1.0

evgeni:
our builder got updated to debian trixie and dpkg is now (rightfully) complaining our packages are not valid 3.0 format.

they never were 3.0, but dpkg accepted it in the past


evgeni:
I intentionally did not update the changelog, to skip builds of 133 packages for now. This was tested in #13406 and worked fine there. This will propagate as we update our packages.

bastelfreak (approved):

adamruzicka (approved):

nadjaheitmann (approved):

nmburgan (approved):

PR #13417, opened 2026-04-30
install which as rubyipmi dependency

arvind4501:
rubyipmi has runtime dependency on `which`  and wothout it fails at https://github.com/logicminds/rubyipmi/blob/main/lib/rubyipmi.rb#L155 with
Error details for No such file or directory - which: Errno::ENOENT: No such file or directory - which


ogajduse (approved):
Makes sense to me.  Thanks for the patch.

PR #13451, opened 2026-05-12
Disable EL8 Foreman/Katello mock repos

ogajduse:
## Summary
Foreman and Katello are no longer built for EL8. This PR removes the foreman, foreman-plugins, and katello repository configurations from the EL8 mock config.

## Changes
- Removed `[foreman]`, `[foreman-plugins]`, and `[katello]` repo sections from `mock/el8.cfg`
- Kept all base EL8 repos since foreman-client is still built for EL8
- No changes to `repoclosure/yum.conf` as it only contains foreman-client repos for EL8

### DIFF
--- a/mock/el8.cfg
+++ b/mock/el8.cfg
@@ -73,18 +73,6 @@ gpgcheck=1
 enabled=0
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
 
-[foreman]
-name=foreman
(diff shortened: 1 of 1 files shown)


evgeni (changes requested):
we have a VCS, we can delete lines and restore them later if needed ;)

zjhuntin (approved):

ogajduse:
> we have a VCS, we can delete lines and restore them later if needed ;)

and I often forget about it :)

evgeni (approved):

PR #13505, opened 2026-05-21
Use systemd service and timer to upload katello tracer

sbernhard:
Instead of using cron, use systemd service and timer to upload the katello tracer. This is modern solution which is also supported by SUSE Micro distribution (which does not deliver a cron daemon)

### DIFF
--- a/packages/client/katello-host-tools/katello-host-tools.spec
+++ b/packages/client/katello-host-tools/katello-host-tools.spec
@@ -4,6 +4,16 @@
 %global build_tracer 0%{?rhel} >= 7 || 0%{?fedora} || 0%{?suse_version}
 
 %if 0%{?suse_version}
+%define pkg_systemd_post()    %service_add_post %{*}
+%define pkg_systemd_preun()   %service_del_preun %{*}
+%define pkg_systemd_postun()  %service_del_postun %{*}
+%else
+%define pkg_systemd_post()    %systemd_post %{*}
+%define pkg_systemd_preun()   %systemd_preun %{*}
+%define pkg_systemd_postun()  %systemd_postun %{*}
+%endif
+
+%if 0%{?suse_version}
 %define dist suse%{?suse_version}
 %endif
 
@@ -33,8 +43,8 @@
 %global katello_libdir %{python_libdir}/katello
 
 Name: katello-host-tools
-Version: 4.5.0
-Release: 2%{?dist}
+Version: 4.6.0
+Release: 1%{?dist}
 Summary: A set of commands and yum plugins that support a Katello host
 Group:   Development/Languages
 %if 0%{?suse_version}
@@ -54,6 +64,8 @@ BuildArch: noarch
 %endif
 
 Requires: subscription-manager
+Requires: systemd
+BuildRequires: systemd-rpm-macros systemd
 Obsoletes: %{name}-fact-plugin < %{version}-%{release}
 Obsoletes: katello-agent < %{version}-%{release}
 
@@ -98,11 +110,6 @@ BuildArch:  noarch
 Summary:    Adds Tracer functionality to a client managed by katello-host-tools
 Group:      Development/Languages
 Requires: %{name} = %{version}-%{release}
-%if 0%{?suse_version}
-Requires: cronie
-%else
-Requires: crontabs
-%endif
 %if %{yum_install}
 Requires: python2-tracer >= 0.6.12
 %endif
@@ -208,9 +215,12 @@ mkdir -p %{buildroot}%{_sbindir}
 cp extra/katello-tracer-upload-dnf %{buildroot}%{_sbindir}/katello-tracer-upload
 %endif
 
-# crontab
-mkdir -p %{buildroot}%{_sysconfdir}/cron.d/
-cp extra/katello-tracer-upload.cron %{buildroot}%{_sysconfdir}/cron.d/katello-tracer-upload
+# tracer systemd service and timer
+install -Dp -m0644 extra/katello-tracer-upload.service.in %{buildroot}%{_unitdir}/katello-tracer-upload.service
+install -Dp -m0644 extra/katello-tracer-upload.timer %{buildroot}%{_unitdir}/katello-tracer-upload.timer
+
+# replace @SBIN_PATH@ in the service file with the actual path to the executable
+sed -i "s|@SBIN_PATH@|%{_sbindir}|" %{buildroot}%{_unitdir}/katello-tracer-upload.service
 %endif
 
 %clean
@@ -221,6 +231,25 @@ rm -rf %{buildroot}
 katello-package-upload 2> /dev/null
 katello-enabled-repos-upload 2> /dev/null
 exit 0
+%endif
+
+%if %{build_tracer}
+%post tracer
+%pkg_systemd_post katello-tracer-upload.timer
+
+if [ "$1" -eq 1 ]; then
+    systemctl enable --now katello-tracer-upload.timer >/dev/null 2>&1 || :
+fi
+
+%preun tracer
+%pkg_systemd_preun katello-tracer-upload.timer
+
+if [ "$1" -eq 0 ]; then
+    systemctl disable --now katello-tracer-upload.timer >/dev/null 2>&1 || :
+fi
+
+%postun tracer
+%pkg_systemd_postun katello-tracer-upload.timer
 %endif
 
 %files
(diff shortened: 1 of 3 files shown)


sbernhard:
Requires https://github.com/Katello/katello-host-tools/pull/171 to be merged first.

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -208,9 +204,12 @@ mkdir -p %{buildroot}%{_sbindir}
 cp extra/katello-tracer-upload-dnf %{buildroot}%{_sbindir}/katello-tracer-upload
 %endif
 
-# crontab
-mkdir -p %{buildroot}%{_sysconfdir}/cron.d/
-cp extra/katello-tracer-upload.cron %{buildroot}%{_sysconfdir}/cron.d/katello-tracer-upload
+# tracer systemd service and timer
+install -Dp -m0644 extra/katello-tracer-upload.service.in %{buildroot}%{_unitdir}/katello-tracer-upload.service
+install -Dp -m0644 extra/katello-tracer-upload.timer %{buildroot}%{_unitdir}/katello-tracer-upload.timer
```
adamruzicka, line 220: With cron, you just drop a file to a location and you're done. With systemd timers, you drop the unit definitions, but you then have to enable them. How would that be handled?

sbernhard: I'm not sure as I'm not yet able to fully test this but I have expected that the %systemd_post macro does enable the service. Need to verify this.

dgoetz: Yes, but you need also the timer installed and enabled so `katello-tracer-upload.{service,timer}` would be the shortest option. The macros install the unit with the preset, so enabled/disabled depends on this.

ogajduse:
[test rpm]

packages/client/katello-host-tools/katello-host-tools.spec
sbernhard: AFAIK, only the timer need to activated. I adapted the spec file accordingly. but actually, I can only fully test it if https://github.com/Katello/katello-host-tools/pull/171 is merged.
See https://documentation.suse.com/en-us/sle-micro/6.0/html/Micro-systemd-working-with-timers/index.html and

ianballou:
katello-host-tools 4.6.0 is ready for packaging https://github.com/Katello/katello-host-tools/releases/tag/4.6.0

sbernhard:
Packaging done @ianballou

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -223,6 +233,25 @@ katello-enabled-repos-upload 2> /dev/null
...
+if [ "$1" -eq 1 ]; then
+    systemctl enable --now katello-tracer-upload.timer >/dev/null 2>&1 || :
+fi
+
+%preun tracer
+%pkg_systemd_preun katello-tracer-upload.timer
+
+if [ "$1" -eq 0 ]; then
+    systemctl disable --now katello-tracer-upload.timer >/dev/null 2>&1 || :
+fi
```
ogajduse, line 249: `%systemd_preun` already expands to `systemctl disable --now` when `$1 == 0` (via `systemd-update-helper remove-system-units`). The manual call duplicates the same operation. Safe to remove the `if` block:

  ```spec
  %preun tracer
  %pkg_systemd_preun katello-tracer-upload.timer
  ```

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -223,6 +233,25 @@ katello-enabled-repos-upload 2> /dev/null
...
+
+%postun tracer
+%pkg_systemd_postun katello-tracer-upload.timer
```
ogajduse, line 252: Fedora convention is to pass both units so both get proper lifecycle handling:

```spec
%post tracer
%pkg_systemd_post katello-tracer-upload.service katello-tracer-upload.timer

%preun tracer
%pkg_systemd_preun katello-tracer-upload.service katello-tracer-upload.timer

%postun tracer
%pkg_systemd_postun katello-tracer-upload.service katello-tracer-upload.timer
  ```

For reference, logrotate does: `%systemd_post logrotate.{service,timer}`
https://src.fedoraproject.org/rpms/logrotate/blob/rawhide/f/logrotate.spec#_91

File: packages/client/katello-host-tools/katello-host-tools.spec [hunk 4]
```diff
@@ -3,6 +3,16 @@
...
 
+%if 0%{?suse_version}
+%define pkg_systemd_post()    %service_add_post %{*}
+%define pkg_systemd_preun()   %service_del_preun %{*}
+%define pkg_systemd_postun()  %service_del_postun %{*}
+%else
+%define pkg_systemd_post()    %systemd_post %{*}
+%define pkg_systemd_preun()   %systemd_preun %{*}
+%define pkg_systemd_postun()  %systemd_postun %{*}
+%endif
```
ogajduse, line 14:
```suggestion
%if 0%{?suse_version}
%define pkg_systemd_post()    %service_add_post %{*}
%define pkg_systemd_preun()   %service_del_preun %{*}
%define pkg_systemd_postun()  %service_del_postun %{*}
%define pkg_systemd_postun_with_restart()  %service_del_postun %{*}
%else
%define pkg_systemd_post()    %systemd_post %{*}
%define pkg_systemd_preun()   %systemd_preun %{*}
%define pkg_systemd_postun()  %systemd_postun %{*}
%define pkg_systemd_postun_with_restart()  %systemd_postun_with_restart %{*}
%endif
```


On Fedora, `%systemd_postun` expands to literally nothing (verified from `/usr/lib/rpm/macros.d/macros.systemd`). This means on package upgrade (`$1 >= 1`), the timer won't be restarted to pick up changes.

```
# From /usr/lib/rpm/macros.d/macros.systemd on Fedora 43:

%systemd_postun() \
%{expand:%%{?__systemd_someargs_%#:%%__systemd_someargs_%# systemd_postun}} \
%{nil}

%systemd_postun_with_restart() \
%{expand:%%{?__systemd_someargs_%#:%%__systemd_someargs_%# systemd_postun_with_restart}} \
if [ $1 -ge 1 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then \
    # Package upgrade, not uninstall \
    /usr/lib/systemd/systemd-update-helper mark-restart-system-units %{?*} || : \
fi \
%{nil}
```

Compare with plocate which uses `%systemd_postun_with_restart` — that calls `systemd-update-helper mark-restart-system-units`, which sets a `Markers=+needs-restart` flag on the unit. The actual restart is then performed by a `transfiletriggerin` from the `systemd` package, which runs `systemd-update-helper system-reload-restart` (daemon-reload + `systemctl reload-or-restart --marked`) whenever files in `/usr/lib/systemd/system/` are added or updated.

For the SUSE compat wrapper this would need a corresponding `%service_del_postun_with_restart` or equivalent — worth checking if SUSE has one, or whether a manual `systemctl try-restart` is needed there.

Two real-world examples from Fedora rawhide:

plocate ([src](https://src.fedoraproject.org/rpms/plocate/blob/rawhide/f/plocate.spec)):
```spec
%postun
%systemd_postun_with_restart plocate-updatedb.service plocate-updatedb.timer
```

util-linux / fstrim ([src](https://src.fedoraproject.org/rpms/util-linux/blob/rawhide/f/util-linux.spec)):
```spec
%postun
%systemd_postun_with_restart fstrim.timer
%systemd_postun fstrim.service
```

Note util-linux uses `_with_restart` on the timer but plain `%systemd_postun` (no-op) on the service — they want the timer restarted on upgrade but don't care about restarting the one-shot service itself. Reasonable pattern for katello-tracer-upload too since it's also a one-shot triggered by timer.

packages/client/katello-host-tools/katello-host-tools.spec
```diff
@@ -223,6 +233,25 @@ katello-enabled-repos-upload 2> /dev/null
...
+%preun tracer
+%pkg_systemd_preun katello-tracer-upload.timer
+
+if [ "$1" -eq 0 ]; then
+    systemctl disable --now katello-tracer-upload.timer >/dev/null 2>&1 || :
+fi
+
+%postun tracer
+%pkg_systemd_postun katello-tracer-upload.timer
+%endif
```
ogajduse, line 253: Same as for the above postun macro


```suggestion
%postun tracer
%pkg_systemd_postun_with_restart katello-tracer-upload.timer
%pkg_systemd_postun katello-tracer-upload.service
%endif
```

SUSE `%service_del_postun` may already do restart (their macros may behave differently from Fedora). If so, the SUSE wrapper for `_with_restart` mapping to plain `%service_del_postun` is correct. But needs someone with SUSE access to verify.

ogajduse (commented):
I checked how Fedora packages (logrotate, plocate, fstrim) handle systemd timers in their specs, and have a few notes.

PR #13549, opened 2026-05-28
Update rubygem-safemode to 2.0.0

ofedoren:
Is required by https://github.com/theforeman/foreman/pull/11007

Also makes https://github.com/theforeman/foreman-packaging/pull/13515 obsolete

This includes 4 commits:
 - Update safemode to 2.0
 - Make Foreman use that version: https://github.com/theforeman/foreman-packaging/pull/13553
 - Since safemode 2.0 requires Prism gem, add it: https://github.com/theforeman/foreman-packaging/pull/13552
 - Drop ruby_parser because it is replaced with Prism: https://github.com/theforeman/foreman-packaging/pull/13554

### DIFF
--- a/packages/foreman/rubygem-safemode/rubygem-safemode.spec
+++ b/packages/foreman/rubygem-safemode/rubygem-safemode.spec
@@ -2,7 +2,7 @@
 %global gem_name safemode
 
 Name: rubygem-%{gem_name}
-Version: 1.5.0
+Version: 2.0.0
 Release: 1%{?dist}
 Summary: A library for safe evaluation of Ruby code based on ParseTree/RubyParser and Ruby2Ruby
 License: MIT
@@ -10,10 +10,10 @@ URL: https://github.com/svenfuchs/safemode
 Source0: https://rubygems.org/gems/%{gem_name}-%{version}.gem
 
 # start specfile generated dependencies
-Requires: ruby >= 2.7
-Requires: ruby < 3.2
-BuildRequires: ruby >= 2.7
-BuildRequires: ruby < 3.2
+Requires: ruby >= 3.0
+Requires: ruby < 3.4
+BuildRequires: ruby >= 3.0
+BuildRequires: ruby < 3.4
 BuildRequires: rubygems-devel
 BuildArch: noarch
 # end specfile generated dependencies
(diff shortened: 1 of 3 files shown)


ofedoren:
RPM failure suggests we split this PR into PRs per commit, but I'd like to see what @theforeman/packaging thinks first

adamruzicka:
Not tfm/packaging, but from what I remember that suggestion should be legit. Repoclosure can't deal really deal with too many interconnected things changing at once.

dgoetz:
It could already be enough to have the commits in the correct order (currently change to Foreman comes before the bump of safemode), but if this does not work you will need to split the PR.

ofedoren:
I've rebased the commits in a different (hopefully right) order, let's see...

adamruzicka:
No luck :/

ofedoren:
Well, since re-ordering didn't help, I've extracted the other 3 commits into separate PRs, hopefully it helps to push this forward

adamruzicka:
So now we have this which has red ci because foreman requires safemode<2 and https://github.com/theforeman/foreman-packaging/pull/13553 which has red ci because there's no safemode>=2 yet. @theforeman/packaging is there a trick to this or will we have to merge one of those, despite the red ci?

We could probably start with the foreman pr, briefly allowing `safemode >= 1.4, <3`, then bump safemode, then raise the lower bound? But it feels a bit complicated for what it is.

dgoetz:
I would say merging the update of safemode to 2.0.0 after prism was merged with an expected repoclosure error caused by Foreman being pinned to <2 is fine (currently it complains about prism), as long as we merge the Foreman PR quickly after, so I would wait for tests being green there and not do this Friday afternoon if something goes wrong.

packages/foreman/rubygem-safemode/rubygem-safemode.spec
```diff
@@ -2,18 +2,18 @@
 %global gem_name safemode
 
 Name: rubygem-%{gem_name}
-Version: 1.5.0
+Version: 2.0.0
 Release: 1%{?dist}
 Summary: A library for safe evaluation of Ruby code based on ParseTree/RubyParser and Ruby2Ruby
```
ogajduse:
```suggestion
Summary: A library for safe evaluation of Ruby code based on Prism and Ruby2Ruby
```

ogajduse (approved):
ACK with one nit

PR #13554, opened 2026-05-29
Remove rubygem-ruby_parser package

ofedoren:
Extracted from https://github.com/theforeman/foreman-packaging/pull/13549

(cherry picked from commit 56db46a694b8d318e450424614bcaa3e471852f8)


adamruzicka:
Huh, I thought safemode depended on this?

ofedoren:
> Huh, I thought safemode depended on this?

Currently, yes. Safemode 2.0 requires Prism instead of ruby_parser, but it seems Prism still needs ruby_parser: https://github.com/ruby/prism/blob/main/Gemfile#L11. Not sure though if we build RPMs for dependencies of dependencies.

In any case, I've created this as a clean up, but if it will appear that we still need this, I'll close it.

UPD: Although, the ruby_parser dep in Prism is in Gemfile, not in the gemspec, so probably we could drop it...

adamruzicka:
> Safemode 2.0 requires Prism instead of ruby_parser

I was under the impression that safemode <2 requires ruby_parser (which it seems to do?) so I expected the ci to fail here

ofedoren:
> I was under the impression that safemode <2 requires ruby_parser (which it seems to do?) so I expected the ci to fail here

Sorry, misunderstood that. Well, both ruby2ruby (< 2.6) and safemode (< 2.0) required ruby_parser before it switched to Prism. It seems current ruby2ruby we ship here pulls ruby_parser anyway?..

I'm not that into the packaging, so not sure about what exactly happens here :/

ofedoren:
Maybe it doesn't fail since we don't actually depend on RPM based version of ruby_parser? I mean both safemode (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/rubygem-safemode/rubygem-safemode.spec#L12-L18) and ruby2ruby (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/rubygem-ruby2ruby/rubygem-ruby2ruby.spec#L12-L17) don't require ruby_parser explicitly.

adamruzicka:
In any case, this looks good, but I'd still leave it to go in last of the bunch.

ogajduse (approved):

ogajduse:
```
$ copr-cli delete-package @theforeman/foreman-nightly-staging --name rubygem-ruby_parser
Package rubygem-ruby_parser was successfully deleted.
```

ogajduse:
PR to obsolete rubygem-ruby_parser:
https://github.com/theforeman/foreman-packaging/pull/13566/

PR #13566, opened 2026-06-01
Obsolete rubygem-ruby_parser

ogajduse:
Add rubygem-ruby_parser and rubygem-ruby_parser-doc to foreman-obsolete-packages.

rubygem-ruby_parser was a runtime dependency of rubygem-safemode 1.x (auto-generated from gemspec via %gem_install), shipped on rpm/3.19. safemode 2.0.0 dropped the ruby_parser dependency, and the package was removed in 67973840f.

### DIFF
--- a/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
+++ b/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
@@ -1,5 +1,5 @@
 Name: foreman-obsolete-packages
-Version: 1.12
+Version: 1.13
 Release: 1%{?dist}
 License: MIT
 Summary: A package to obsolete retired packages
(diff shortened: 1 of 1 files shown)


packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
```diff
@@ -23,6 +23,8 @@ Obsoletes: rubygem-fog-ovirt < 2.0.3-1
 Obsoletes: rubygem-ovirt-engine-sdk < 4.6.0-1
 Obsoletes: rubygem-hammer_cli_foreman_admin < 1.2.2
 Obsoletes: rubygem-hammer_cli_foreman_admin-doc < 1.2.2
+Obsoletes: rubygem-ruby_parser < 3.21.2
+Obsoletes: rubygem-ruby_parser-doc < 3.21.2
```
adamruzicka: _personal preference:_ For majority of the obsolete packages, we deprecated the next release rather than next patch version. I'd prefer to stick with that

```suggestion
Obsoletes: rubygem-ruby_parser < 3.21.1-4
Obsoletes: rubygem-ruby_parser-doc < 3.21.1-4
```

adamruzicka (commented):
One suggestion, otherwise lgtm, it should do the trick anyway.

packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
ogajduse: Good point. Updated.

adamruzicka (approved):

ogajduse:
Closing as we should not obsolete ruby_parser yet.
ref: https://github.com/theforeman/foreman-packaging/pull/13579

ogajduse:
ruby_parser is removed. Foreman runs without it. We can obsolete it now.

ofedoren (approved):

Odilhao (approved):

PR #13568, opened 2026-06-01
Update rubygem-foreman_cve_scanner to 0.6.0

sbernhard:


nadjaheitmann:
The foreman packaging automation can do this for you ;) https://github.com/theforeman/foreman-packaging/pull/13570#event-26225290534

nadjaheitmann:
But it also fails :)

sbernhard:
Ah, thanks @nadjaheitmann. I would prefer this PR instead of the auto-generated one.

nadjaheitmann:
> Ah, thanks @nadjaheitmann. I would prefer this PR instead of the auto-generated one.

Then you'll need to rebase it. Pipeline is now green for the other one.

PR #13605, opened 2026-06-10
add amazing_print to trixie

evgeni:


evgeni:
[test deb]

dependencies/trixie/amazing_print/changelog
```diff
@@ -0,0 +1,29 @@
+ruby-amazing-print (1.7.2-1) stable; urgency=low
+
+  * 1.7.2 released
```
ogajduse, line 3: After we merge, we could upgrade to 2.0.0. But Foreman would need to relax first.
bundler.d/console.rb:
```ruby
gem 'amazing_print', '~> 1.1', :require => 'ap'
```

dependencies/trixie/amazing_print/control
```diff
@@ -0,0 +1,25 @@
+Source: ruby-amazing-print
+Section: ruby
+Priority: optional
+Maintainer: The Foreman Project <packaging@theforeman.org>
+Build-Depends: debhelper (>= 11),
```
ogajduse, line 5: Minor: Trixie ships debhelper 13.
https://packages.debian.org/trixie/debhelper

ogajduse (approved):

dependencies/trixie/amazing_print/changelog
evgeni: Once foreman allows that, we can look into it, yeah

dependencies/trixie/amazing_print/control
evgeni: Correct, but this package is dh compat `11` (newer debhelper can behave like old debhelper, so that packages build correctly, even if dh changes)

Lookin into upgrading to 13 is out of scope for now

ogajduse: I see, it's defined in `dependencies/trixie/amazing_print/compat`

evgeni: Exactly

You can also do this by depending on `debhelper-compat (= 11)` instead, which would do the very same thing: it's provided by any debhelper package 11 or newer and it will behave in a stable manner, no matter which new changes dh will receive

PR #13606, opened 2026-06-10
add amazing_print to noble

evgeni:


ogajduse (approved):
Comments from https://github.com/theforeman/foreman-packaging/pull/13605#pullrequestreview-4485796793 apply here.

PR #13625, opened 2026-06-15
Automatically use system openssl configuration on rpm builds

adamlazik1:
Requires https://github.com/theforeman/smart-proxy/pull/947

### DIFF
--- a/packages/foreman/foreman-proxy/foreman-proxy-use-crypto-policies.patch
+++ b/packages/foreman/foreman-proxy/foreman-proxy-use-crypto-policies.patch
@@ -1,0 +1,25 @@
+--- a/lib/launcher.rb
++++ b/lib/launcher.rb
+@@ -5,7 +5,6 @@
+ require 'proxy/log_buffer/trace_decorator'
+ require 'sd_notify'
+
+-CRYPTO_POLICIES_CONFIG = '/etc/crypto-policies/back-ends/opensslcnf.config'.freeze
+ TLS_MIN_VERSION_MAP = {
+   '1.0' => OpenSSL::SSL::TLS1_VERSION,
+   '1.1' => OpenSSL::SSL::TLS1_1_VERSION,
+@@ -117,13 +116,7 @@
+       return nil if configured&.empty?
+       return configured unless configured.nil?
+
+-      if File.exist?(CRYPTO_POLICIES_CONFIG)
+-        logger.info "Crypto-policies detected, using PROFILE=SYSTEM for TLS ciphers."
+-        'PROFILE=SYSTEM'
+-      else
+-        logger.debug "No crypto-policies detected, using HIGH cipher string as default."
+-        'HIGH'
+-      end
++      'PROFILE=SYSTEM'
+     end
+
+     def validate_tls_ciphers!(ciphers)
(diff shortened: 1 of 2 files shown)


evgeni:
Why are we adding a feature in the main repo, just to patch it in packaging?

adamlazik1:
> Why are we adding a feature in the main repo, just to patch it in packaging?

To differentiate between deb in rpm builds I believe.

evgeni:
Differentiate how?
https://github.com/theforeman/smart-proxy/pull/947 already has code to fall back to non-crypto-policy if the system has no support for that (like Debian)

adamlazik1:
I believe that the argument was that an edge case can happen in which in the opensslcnf.config file is missing, but we should assume that crypto-policies are always applied on an rpm system.

evgeni:
Then check for the existence of /etc/crypto-policies/ directly?

Patching at the packaging level is automatically buying you future build problems (including Packit problems), which I really would recommend against.

packages/foreman/foreman-proxy/foreman-proxy-use-crypto-policies.patch
```diff
@@ -0,0 +1,25 @@
...
+       return configured unless configured.nil?
+
+-      if File.exist?(CRYPTO_POLICIES_CONFIG)
+-        logger.info "Crypto-policies detected, using PROFILE=SYSTEM for TLS ciphers."
+-        'PROFILE=SYSTEM'
+-      else
+-        logger.debug "No crypto-policies detected, using HIGH cipher string as default."
+-        'HIGH'
+-      end
++      'PROFILE=SYSTEM'
```
lhellebr, line 22: On upstream, this either logs crypto-policies are used, or not.
On Satellite, this deletes any logging - there will be no information about crypto-policies.

adamruzicka: > On Satellite, this deletes any logging - there will be no information about crypto-policies.

Would it make sense to log something that has no alternatives?

lhellebr: I think it won't leave the user wondering. It's probably nice to see "yes, if I set crypto policies, they will be taken into account".

adamlazik1: Based on what @evgeni said in the comments I am reevaluating whether we should have this patch at all. If a RHEL system has crypto-policies enabled but no config file, that's a broken system and probably not something we should deal with. Opinions?

adamlazik1: @ekohl also said that the patch might not be necessary in here https://github.com/theforeman/smart-proxy/pull/947#discussion_r3454286766

lhellebr: I thought the reason for this patch was to avoid autodetection during startup every time. If it isn't a problem (and I don't think it is), I also think this patch doesn't need to exist at all. A system either has crypto-policies or it doesn't, and we don't really care whether it's rpm based or something else based.

If a system is broken in a way that it does have crypto policies and doesn't have the file, it's nice to have this exception caught and communicated somehow, but for this case you don't need to explain things, just say that the file doesn't exist.

evgeni: Thank you!

PR #13643, opened 2026-06-22
Update foreman-obsolete-packages to 1.15

ogajduse:
## Summary
- Add `Obsoletes` for `rubygem-stomp` and `rubygem-stomp-doc`
- `rubygem-stomp` was removed from packaging but never obsoleted
- Users who had it installed via Katello comps (3.17/3.18) keep the stale RPM after upgrading to 3.19+
- Verified on a real system: 3.17 → 3.18 → 3.19 → nightly, `rubygem-stomp` persists as orphan

PR #13644, opened 2026-06-22
Obsolete rubygem-stomp in foreman-obsolete-packages

ogajduse:
## Summary
- Add `Obsoletes` for `rubygem-stomp` and `rubygem-stomp-doc` (bump release to 1.11-2)
- `rubygem-stomp` was removed from packaging but never obsoleted
- Users who had it installed via Katello comps (3.17/3.18) keep the stale RPM after upgrading to 3.19
- Verified on a real system: 3.17 → 3.18 → 3.19 → nightly, `rubygem-stomp` persists as orphan
- Uses release bump (1.11-2) instead of version bump to avoid collision with 1.12 on rpm/develop which has different content
- Counterpart of https://github.com/theforeman/foreman-packaging/pull/13643 for rpm/develop

### DIFF
--- a/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
+++ b/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec
@@ -1,6 +1,6 @@
 Name: foreman-obsolete-packages
 Version: 1.11
-Release: 1%{?dist}
+Release: 2%{?dist}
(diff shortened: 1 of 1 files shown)


ogajduse:
Package removed in https://github.com/theforeman/foreman-packaging/pull/13091

zjhuntin (approved):

PR #12872, opened 2025-12-19
Build packages with EL10

ekohl:
This includes:

* https://github.com/theforeman/foreman-packaging/pull/12768
* https://github.com/theforeman/foreman-packaging/pull/12770
* https://github.com/theforeman/foreman-packaging/pull/11562
* https://github.com/theforeman/foreman-packaging/pull/12871
* https://github.com/theforeman/foreman-packaging/pull/12873
* https://github.com/theforeman/foreman-packaging/pull/12874
* https://github.com/theforeman/foreman-packaging/pull/12875
* https://github.com/theforeman/foreman-packaging/pull/12880
* https://github.com/theforeman/foreman-packaging/pull/12881
* https://github.com/theforeman/foreman-packaging/pull/12882
* https://github.com/theforeman/foreman-packaging/pull/12944
* https://github.com/theforeman/foreman-packaging/pull/13053
* https://github.com/theforeman/foreman-packaging/pull/13113
* https://github.com/theforeman/foreman-packaging/pull/13117
* https://github.com/theforeman/foreman-packaging/pull/13146
* https://github.com/theforeman/foreman-packaging/pull/13147
* https://github.com/theforeman/foreman-packaging/pull/13161
* https://github.com/theforeman/foreman-packaging/pull/13173
* https://github.com/theforeman/foreman-packaging/pull/13174
* https://github.com/theforeman/foreman-packaging/pull/13702

Currently known failing packages when building `ruby_core_packages_tier1`:

* `rubygem-safemode` incompatible with Ruby >= 3.2. This is maintained by Foreman and needs work (https://github.com/theforeman/safemode/issues/54). Currently hacked around
* ~~`rubygem-sqlite3` fails to compile on EL10. We're on 1.4.4 which is 3 years old. Might be sufficient to upgrade to 3.x~~

Currently known failing packages when building `ruby_core_packages_tier2`:

* ~~`rubygem-rkerberos` fails to compile on EL10 (https://github.com/domcleal/rkerberos/issues/27). This gem is unmaintained so needs attention. Currently patched out of `foreman-proxy` to allow skipping it.~~

What does work:

* `rails_core_packages`
* `hammer_core_packages`
* `other_core_packages`

What doesn't work:

* ~~`foreman_nodejs_packages` has various issues and I haven't tried to figure out why. Some need their cache rebuilt for NPM 6, others complain about missing files~~
* ~~`foreman` needs at least `rubygem-safemode` and nodejs packages~~
[...]

PR #13702, opened 2026-07-06
Make it easy to rewrite NodeJS packages

ekohl:
When npm2rpm is updated or if you want to update the vendored dependencies of a package then you want to easily regenerate them.

To regenerate all packages, you can use:

    awk '/%global npm_name/ { print $3 }' packages/*/nodejs-*/*.spec | xargs -n 1 -I{} ./add_npm_package.sh {} current
    git diff --cached  --name-only  | grep spec | xargs -n 1 sed -i '0,/Update to/ s/Update to .\+/Regenerate spec file/'

Ideally the last line wouldn't be needed and you can actually pass that in, but that's left as an exercise for the reader.

Split off from https://github.com/theforeman/foreman-packaging/pull/12872.


ekohl:
@zjhuntin @Odilhao @ogajduse mind reviewing this?

add_npm_package.sh
```diff
@@ -49,6 +49,9 @@ ensure_program() {
 }
 
 generate_npm_package() {
+  local release
+  release="${1:-}"
```
ogajduse: I'd suggest the default to be `1`. `npm2rpm -r` defaults to `1` only when the flag is absent. 

```suggestion
  release="${1:-1}"
```

add_npm_package.sh
```diff
@@ -159,13 +164,30 @@ fi
 
 if [[ $UPDATE == true ]] ; then
   EXISTING_VERSION=$(rpmspec --query --srpm --queryformat '%{VERSION}' "$SPEC_FILE")
-  if [[ $REWRITE_ON_SAME_VERSION == true ]] || [[ $VERSION != $EXISTING_VERSION ]]; then
+  if [[ $VERSION == "$EXISTING_VERSION" ]] ; then
+    echo "$PACKAGE_NAME is already at version $VERSION"
+    if [[ $REWRITE_ON_SAME_VERSION == true ]] ; then
+      if [[ -z $STRATEGY ]] ; then
+        if grep -q registry.npmjs.org.tgz "$SPEC_FILE" ; then
```
ogajduse: This no longer matches anything, as the cache files were removed from the project using npm2rpm 8.0.0. Running this would rewrite all bundle packages as `single`.
```suggestion
        if grep -q 'bundled(npm(' "$SPEC_FILE" ; then
```

add_npm_package.sh
```diff
@@ -159,13 +164,30 @@ fi
...
+          STRATEGY="bundle"
+        else
+          STRATEGY="single"
+        fi
+      fi
+      release=$(rpmspec --query --queryformat "%{release}" "$SPEC_FILE")
+      generate_npm_package "$(( ${release%.*} + 1 ))"
```
ogajduse: `rpmspec --query without --srpm` prints one line per binary package. For a spec with a `%package doc` we'd get two lines.
I propose adding `--srpm` as we have it in `add_changelog.sh`

https://github.com/ekohl/foreman-packaging/blob/87867a1c735c8eaf1f339af5c659d8191420d105/add_changelog.sh#L12

I'd also suggest `--undefine=dist` so we get a clean release without dist.

```suggestion
      release=$(rpmspec --query --srpm --undefine dist --queryformat "%{release}" "$SPEC_FILE")
      generate_npm_package "$(( release + 1 ))"
```

ogajduse (commented):
On the rewrite flow: `generate_npm_package` stages `git rm -r $PACKAGE_DIR` before running npm2rpm, which needs the registry twice (`npm install --package-lock-only, then spectool | wget`). If that fails, the package is staged as a pure deletion and `xargs` doesn't stop, it just returns 123 at the end. Across 43 packages, one failure is two lines in `git diff --cached` that look exactly like the other ~1000, and committing the staging area drops a package plus its annexed tarballs.

Not asking for per-package commits, but the description should say the exit code needs checking and that a partial run must be reset before committing. Generating into a temp dir and swapping on success would fix it properly, though that's beyond this PR's scope.

add_npm_package.sh
ekohl: Oh that's ugly behavior on the npm2rpm side. I'll update it here.

ekohl:
> Not asking for per-package commits, but the description should say the exit code needs checking and that a partial run must be reset before committing. Generating into a temp dir and swapping on success would fix it properly, though that's beyond this PR's scope.

I intentionally kept the committing out of the loop. You can do that on a higher level. For example, by using `for` instead of `xargs` in the bash control loop.

add_npm_package.sh
ekohl: In some Ruby packages we also added a comment that has the strategy. That may be a nice idea long term, but today wouldn't work. For now I'll follow your suggestion.

ogajduse (approved):
Thank you, @ekohl!

PR #13717, opened 2026-07-13
Add foreman-rake wrapper to foremanctl

ShimShtein:


packages/foreman/foremanctl/foreman-rake
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
evgeni, line 1: This file should not live in packaging, but in `foremanctl.git`, as we also want to deploy it on Debian systems with the same content

ShimShtein: Makes sense, but how do I connect the two repos?
Not sure how to implement it technically.

ekohl: This is similar to how we implement `foreman-certificate-check` in `foremanctl`.

More background why we design it this way: https://github.com/theforeman/foremanctl/issues/630.

ShimShtein: also: If we do that, we will create a circular reference for testing it:
this PR will require the foremanctl PR, but the tests in the foremanctl PR will rely on this PR to install the wrapper in the proper folder. 
Of course there is an option to test the script from a local location instead of the `/usr/sbin` folder, but then it would be less of an integration test.

ekohl:
This needs to be implemented in `foremanctl`.

PR #13749, opened 2026-07-20
Remove orphaned packages with no reverse dependencies

zjhuntin:
## Summary
- Remove 11 rubygem packages that have zero consumers at both RPM and Ruby gem level
- 6x google-apis-* (dns_v1, iamcredentials_v1, monitoring_v3, pubsub_v1, sqladmin_v1beta4, storage_v1): remnants of fog-google which was already removed and obsoleted. foreman_google uses google-apis-compute_v1 directly, not fog-google
- rubygem-uglifier: replaced by terser for JavaScript minification
- rubygem-colorize: not required by any gem in the repository
- rubygem-apipie-params: not required by apipie-rails, apipie-bindings, or hammer_cli_foreman
- rubygem-declarative-option: remnant of google-api-client (also removed and obsoleted)
- rubygem-method_source: dependency of pry (debugger), which is not shipped
- Bump foreman-obsolete-packages to 1.16 with proper Obsoletes entries for clean upgrade paths

## Context
Part of EL10 gap analysis work. These packages were identified by cross-referencing all 412 gems in the repo against rubygems.org dependency data — none are required by any other shipped gem.

## Test plan
- [ ] Verify `package_manifest.yaml` is valid YAML
- [ ] Verify no remaining spec files reference removed packages
- [ ] Verify foreman-obsolete-packages.spec builds cleanly (rpmlint passes)
- [ ] Confirm Obsoletes versions are higher than the removed package versions


Odilhao (approved):
Reviewed against the same logic used for the pulpcore-packaging EL10 cleanup (theforeman/pulpcore-packaging#2765/#2766/#2768), adapted for Ruby.

## Removal list — verified correct

Cross-checked all 11 removed gems against rubygems.org's reverse-dependency API, intersected against the ~400 gems actually shipped in this repo's `package_manifest.yaml`:

- 10 of 11 (both google-apis-* gems, apipie-params, colorize, declarative-option, method_source) have **zero** active-repo consumers at all, gem-level or RPM-level.
- `uglifier` does show up as a reverse-dependency of 4 active gems (angular-rails-templates, bootstrap-sass, sprockets, sprockets-rails) — but checked each one's actual gemspec, and in all 4 cases `uglifier` is listed only under `development` dependencies (test/build tooling), never `runtime`. Fedora's ruby dependency generator only turns runtime deps into RPM `Requires`, so this doesn't produce any auto-generated dependency — matches the pattern that broke the Python repo (auto-generated deps from unconditional runtime metadata), and confirms this one's clean.
- Grepped the whole tree for `rubygem(<name>)` and `rubygem-<name>` Requires — no hits outside the packages being removed themselves.

## Obsoletes — correct, and a cleaner pattern than what we used upstream

These bump the **Version** field by one patch level (e.g. `Obsoletes: rubygem-uglifier < 4.2.1` for a last-shipped 4.2.0-1) rather than bumping the Release number. Verified with `rpmdev-vercmp` that this sidesteps the off-by-one bug we hit on the pulpcore-packaging side entirely — comparing at the Version level always beats a `.elN`-suffixed release, so there's no need to know the exact last-shipped Release:

```
$ rpmdev-vercmp 4.2.0-1.el9 4.2.1
4.2.0-1.el9 < 4.2.1
```

All 11 `-doc` Obsoletes pairs correctly correspond to specs that actually build a `%package doc` subpackage.

## Manifest / comps

`package_manifest.yaml` and `comps/comps-foreman-el9.xml` are both cleanly scrubbed of all 11 packages, and the package directories are gone from disk. No leftovers found.

Approved. CI still running but PR shows mergeable.

PR #13756, opened 2026-07-23
Re-add rubygem-colorize and rubygem-method_source

zjhuntin:
## Summary
- Re-adds `rubygem-colorize` (0.8.1) and `rubygem-method_source` (1.1.0), which were incorrectly removed as orphaned packages
- `rubygem-puma-status` depends on `colorize` and `rubygem-railties` depends on `method_source`, causing repoclosure failures on EL9 nightly
- Converts `rubygem-colorize` from SCL to default template
- Removes the corresponding Obsoletes from `foreman-obsolete-packages` (bumped to 1.18)


packages/foreman/rubygem-method_source/rubygem-method_source.spec
```diff
@@ -0,0 +1,92 @@
+# template: default
+%global gem_name method_source
+
+Name: rubygem-%{gem_name}
+Version: 1.1.0
+Release: 1%{?dist}
```
ogajduse: I'd bump the release here, same as you did for rubygem-colorize. And add a changelog entry.

ogajduse (changes requested):

ogajduse (approved):

zjhuntin:
Force adding because the repoclosure is going to fail without this.

PR #13823, opened 2026-07-28
Rebuild rubygem-grpc for EL10

zjhuntin:
## EL10 Rebuild — plugins-tier2

Bump release for EL10 rebuild. CI will scratch build these for the `rhel-10-x86_64` chroot.

### Packages (1)

- [ ] rubygem-grpc

Tracked in [el10_rebuild](https://github.com/theforeman/el10_rebuild).


zjhuntin:
[test rpm]

ogajduse (dismissed):

ogajduse:
The chain in rpm-develop:
```
rubygem-foreman_google (3.0.8)
  └─ rubygem(google-cloud-compute) = 0.5.0     ← only explicit BuildRequires
       └─ google-cloud-compute-v1 1.7.1
            └─ gapic-common 0.12.0             → Requires: rubygem(grpc) >= 1.36, < 2
                 └─ googleapis-common-protos 1.3.12 → Requires: rubygem(grpc) >= 1.27, < 2
```

Odilhao:
Reviewed — CI is green on both `rhel-9-x86_64` and `rhel-10-x86_64` scratch builds, and the patch content looks correct (const-correctness fix for `grpc_compression_algorithm_name`, thread-callback signature fixes for `rb_thread_create`/`rb_thread_call_without_gvl` prototype mismatches with wrapper functions where the same callback is reused as both the "without GVL" func and a `ubf`, and removal of the now-unneeded `-Wl,-wrap,memcpy` glibc<2.17 workaround). Release bump and changelog itemization look right.

A couple of nits, non-blocking:
- The patch header (`From: downstream`) has no reference for provenance — worth a short note on where these fixes came from for future maintainers rebasing this delta.
- `-Wno-register` is applied build-wide via `CXXFLAGS` rather than scoped to just the vendored abseil-cpp sources that need it. Fine for now, but if grpc gets bumped later this blanket suppression could mask genuinely new `register` usage in first-party code.

One question on approach: all three fixes in `grpc-gcc14-compat.patch` (pointer types, thread signatures, `-Wl,-wrap,memcpy`) are specifically GCC 14 / glibc 2.39 issues that shouldn't be triggered on EL9's older toolchain. Right now the patch applies unconditionally for both chroots. Was that intentional (to keep NVR/spec content identical across el9/el10 for parity), or should this be scoped to EL10 only (e.g. `%if 0%{?rhel} >= 10`) so the el9 build stays byte-for-byte what it was before this rebuild? Not blocking since el9 CI is green either way, just want to confirm the intended convention here.

packages/plugins/rubygem-grpc/grpc-gcc14-compat.patch
```diff
@@ -0,0 +1,107 @@
...
+     return;
+   }
+-  rb_thread_call_without_gvl(grpc_rb_event_unblocking_func, NULL, NULL, NULL);
++  rb_thread_call_without_gvl(grpc_rb_event_unblocking_func_wrapper, NULL, NULL, NULL);
+   rb_funcall(g_event_thread, rb_intern("join"), 0);
+   g_event_thread = Qnil;
+ }
+diff -up a/src/ruby/ext/grpc/extconf.rb b/src/ruby/ext/grpc/extconf.rb
+--- a/src/ruby/ext/grpc/extconf.rb
++++ b/src/ruby/ext/grpc/extconf.rb
```
pcreech: This modification took place in this commit:  https://github.com/grpc/grpc/commit/34be0d84a99a6b313cef48b4c80e14a0d115d864

We should evaluate if we are ok with the way the patch handles it, or if we want to bring that full diff.

packages/plugins/rubygem-grpc/grpc-gcc14-compat.patch
```diff
@@ -0,0 +1,107 @@
...
+ VALUE grpc_rb_compression_options_algorithm_value_to_name_internal(
+     grpc_compression_algorithm internal_value) {
+-  char* algorithm_name = NULL;
++  const char* algorithm_name = NULL;
+ 
+   if (!grpc_compression_algorithm_name(internal_value, &algorithm_name)) {
+     rb_raise(rb_eArgError, "Failed to convert algorithm value to name");
+diff -up a/src/ruby/ext/grpc/rb_channel.c b/src/ruby/ext/grpc/rb_channel.c
+--- a/src/ruby/ext/grpc/rb_channel.c
++++ b/src/ruby/ext/grpc/rb_channel.c
```
pcreech: The changes from here to the next comment address issues with clang 16 compatibility.  here's the patch that addresses it in grpc:

https://github.com/grpc/grpc/commit/c637a9f82bc43f7380b04baa96df243f157939d4

They took a slightly different approach of instead of calling the void function in the wrapper, the wrapper now handles   the logic, while having the primary function just call the wrapper function.  I would defer to the parent commit for this one unless we have a compelling need.

packages/plugins/rubygem-grpc/grpc-gcc14-compat.patch
```diff
@@ -0,0 +1,107 @@
...
+functions are also used as ubf (3rd arg) which expects void(*)(void*).
+
+diff -up a/src/ruby/ext/grpc/rb_compression_options.c b/src/ruby/ext/grpc/rb_compression_options.c
+--- a/src/ruby/ext/grpc/rb_compression_options.c
++++ b/src/ruby/ext/grpc/rb_compression_options.c
+@@ -296,7 +296,7 @@ VALUE grpc_rb_compression_options_level_
+  * Fails if the enum value is invalid. */
+ VALUE grpc_rb_compression_options_algorithm_value_to_name_internal(
+     grpc_compression_algorithm internal_value) {
+-  char* algorithm_name = NULL;
```
pcreech: This section is 100% the straight issue with gcc14, and fixes it as it was fixed in grpc's codebase.

packages/plugins/rubygem-grpc/rubygem-grpc.spec
```diff
@@ -42,8 +47,12 @@ Documentation for %{name}.
 
 %prep
 %setup -q -n  %{gem_name}-%{version}
+%patch -P0 -p1
```
Odilhao: We have `autosetup -p` to apply patchs w/o the need of using `%patch`

ogajduse:
Looks good to me. I've raised https://github.com/zjhuntin/foreman-packaging/pull/1 to tidy up the patch.

ogajduse (approved):

PR #13850, opened 2026-07-29
Update nodejs-webpack to 5.109.2 (lockfile approach)

zjhuntin:
## Summary
- Updates nodejs-webpack from 5.108.4 to 5.109.2
- Uses the lockfile-based offline install from theforeman/npm2rpm#100 instead of the cache tarball approach
- Replaces the 6 MB binary `*-registry.npmjs.org.tgz` blob with a 32 KB reviewable `*-package-lock.json` text file
- Updates `add_npm_package.sh` to handle the lockfile in a separate commit

Depends on theforeman/npm2rpm#100


Odilhao (approved):
From one tar.gz with 5.5mb to one file with 800 lines is a huge improvement on how we can audit the packages.

PR #13860, opened 2026-07-30
Add the lockfile instead of the npm cache tarball

ogajduse:
Depends on theforeman/npm2rpm#100, which replaces the generated npm cache tarball with a committed `package-lock.json`.

Without this the bundle branch aborts: the `*-registry.npmjs.org.tgz` glob matches nothing, `git add` errors on the pathspec, and the script runs under `bash -e`. The lockfile also needs adding as a plain file — the annex step only globs `*.tgz`.

Draft until npm2rpm#100 lands.

### DIFF
--- a/add_npm_package.sh
+++ b/add_npm_package.sh
@@ -75,8 +75,8 @@ generate_npm_package() {
   echo "FINISHED"
 
   if [ "$STRATEGY" = "bundle" ]; then
-    echo -e "Adding npmjs cache binary... - "
-    git add $PACKAGE_DIR/*-registry.npmjs.org.tgz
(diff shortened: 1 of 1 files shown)


ogajduse:
Closing — @zjhuntin already landed this in 986121500, and our changes are functionally identical: same `git add $PACKAGE_DIR/*-package-lock.json`, same placement. The only difference left is the echo string, and "Adding package lockfile" is clearer than my "Adding lockfile", so there's nothing here worth rebasing for.

Thanks for getting it in ahead of the npm2rpm merge — `bash -e` plus the unmatched `*-registry.npmjs.org.tgz` glob would have broken the first bundle regeneration.

PR #13923, opened 2026-08-03
Update Candlepin to 5.0 in katello-repos

zjhuntin:
## Summary
- Update candlepin_version from 4.8 to 5.0
- Replace GPG signing key with candlepin 5.0 key


packages/katello/katello-repos/candlepin-5.0.gpg
Odilhao, line 1: Should we keep both gpgs? For 4.8 and 5.0? Since 4.8 will be installed on EL9.

Odilhao (approved):

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -1,12 +1,12 @@
 %global pulpcore_version nightly
-%global candlepin_version 4.8
+%global candlepin_version 5.0
```
Odilhao, line 3: Can we all a `if` section here for EL9 to keep on 4.8 and EL10 to go to 5.0?

Odilhao: We are good to move with only 5.0

packages/katello/katello-repos/katello-repos.spec
```diff
@@ -73,6 +73,9 @@ rm -rf %{buildroot}
 %{_sysconfdir}/pki/rpm-gpg/RPM-GPG-KEY-candlepin
 
 %changelog
+* Sun Aug 03 2026 Zach Huntington-Meath <zhunting@redhat.com> - 5.0-0.4.nightly
```
Odilhao: Date needs to be fixed

Odilhao (changes requested):

Odilhao (approved):

PR #13956, opened 2026-08-11
sp-rex_ssh: Require sshpass for password authentication

adamruzicka:


evgeni:
A changelog entry would be cool :)

ogajduse:
> A changelog entry would be cool :)

TIL `dch` exists

evgeni:
> > A changelog entry would be cool :)
> 
> TIL `dch` exists

Yeah, but I don't think it plays very well with out packaging. But we have https://github.com/theforeman/foreman-packaging/blob/deb/develop/scripts/changelog.rb

evgeni (approved):

adamlazik1 (approved):

PR #13985, opened 2026-08-14
Release foremanctl 3.0.0

arvind4501:


packages/foreman/foremanctl/foremanctl-3.0.0.tar.gz.save
evgeni, line 1: nonono :)

packages/foreman/foremanctl/foremanctl.spec
```diff
@@ -1,7 +1,7 @@
 %global __brp_mangle_shebangs_exclude_from ^%{_datadir}/%{name}/collections/.*$
 
 Name:      foremanctl
-Version:   2.3.0
+Version:   3.0.0
 Release:   4%{?dist}
```
evgeni:
```suggestion
Release:   1%{?dist}
```

arvind4501: obal did this change, but i suspected it as wrong and removed it manually, my bad

evgeni (approved):

PR #13674, opened 2026-06-29
Fix find_gemfile_lock_specs to find packages across all subdirectories

ehelms:
Dependencies don't always live in a single subdirectory, so searching only within a hardcoded subdir caused critical updates to be missed.

PR #9470, opened 2023-06-06
Report conflicting projects in updateable list

ehelms:
Example output:

Before:
```
Found conflict for rubygem-concurrent-ruby: 1.1.10 != 1.2.2
Found conflict for rubygem-tilt: 2.2.0 != 2.1.0
Found conflict for rubygem-fast_gettext: 1.8.0 != 2.3.0
Found conflict for rubygem-unicode-display_width: 1.8.0 != 2.4.2
Found conflict for rubygem-version_gem: 1.1.3 != 1.1.2
```

After:
```
Found conflict for rubygem-concurrent-ruby: 1.1.10 (foreman) != 1.2.2 (smart-proxy)
Found conflict for rubygem-tilt: 2.2.0 (foreman) != 2.1.0 (smart-proxy)
Found conflict for rubygem-fast_gettext: 1.8.0 (foreman) != 2.3.0 (hammer-cli)
Found conflict for rubygem-unicode-display_width: 1.8.0 (smart-proxy) != 2.4.2 (hammer-cli)
Found conflict for rubygem-version_gem: 1.1.3 (foreman) != 1.1.2 (hammer-cli)
```

PR #14002, opened 2026-08-19
Fix false CONFLICT for leaf packages also pinned by a lockfile

ogajduse:
`list_updatable_packages` was permanently flagging some packages `CONFLICT` even when every version source actually agreed — e.g. `rubygem-foreman-tasks` and `rubygem-foreman_remote_execution`, both stuck failing every run: https://github.com/theforeman/foreman-packaging/actions/runs/32013508243/job/95338069508

```
Run ./bump_rpm.sh "packages/plugins/rubygem-foreman_remote_execution" "CONFLICT"
rubygem-foreman_remote_execution: requested with a conflict.
This means multiple packages requested different versions
##[error]Process completed with exit code 1.
```

**Cause:** `merge_specs()` treats "supported leaf" packages (tracked via `find_packaged_specs()`, which yields no explicit version — just "check latest upstream") as a source equal to any real `Gemfile.lock` pin. #9470 restructured the comparison for better conflict messages, but the rewrite dropped the original `not in (None, new_version)` tolerance that let a leaf's placeholder be safely overwritten by a real pin — leaving the docstring ("will take the more specific constraint") describing behavior the code no longer had. This stayed unnoticed until #13674 widened the glob that discovers Gemfile.lock-pinned specs, which is what started catching these plugin packages.

**Fix:** restore the original comparison (`not in (None, new_version)`), so a leaf placeholder yields to a real pin instead of being compared against it. Also propagate the actual conflicting versions/projects into the matrix entry, and have `bump_rpm.sh` print that detail on a genuine `CONFLICT` (via the new `DETAIL` env var, wired up in the companion workflow PR below), so the failing job is self-explanatory instead of requiring a cross-reference against the separate `Gather RPMs` job's log. For a genuine conflict (e.g. two projects pinning different versions of a shared dependency), the job now prints:

```
rubygem-concurrent-ruby: requested with a conflict.
This means multiple packages requested different versions:
  1.1.10 (foreman) != 1.3.8 (smart-proxy)
```

instead of just the generic message shown above.

Companion workflow change (to actually wire the new detail through in CI): #14003

### DIFF
--- a/bump_rpm.sh
+++ b/bump_rpm.sh
@@ -71,7 +71,12 @@ else
 
 	if [[ $NEW_VERSION == "CONFLICT" ]] ; then
 		echo "${PACKAGE_NAME}: requested with a conflict."
-		echo "This means multiple packages requested different versions"
+		echo "This means multiple packages requested different versions:"
+		if [[ -n $DETAIL ]] ; then
+			echo "  ${DETAIL}"
+		else
+			echo "  (no details available)"
+		fi
 		exit 1
 	fi
 fi
--- a/list_updatable_packages
+++ b/list_updatable_packages
@@ -209,10 +209,11 @@ def merge_specs(specs: Iterable[Tuple[Spec, Optional[str], Optional[str]]]) -> M
     result: Dict[Spec, Dict[Optional[str], Optional[str]]] = {}
     for spec, new_version, project in specs:
         existing_spec = result.get(spec)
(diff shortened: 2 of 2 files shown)


zjhuntin (approved):
Worked locally for me, tested against foreman_remote_execution and foreman-tasks and worked well. Thanks!

PR #14033, opened 2026-08-27
Fix rubygem-foreman_theme_satellite to use Satellite assets for el10sat builds

zjhuntin:
The `downstream_build` condition was missing `el10sat`, causing el10sat builds to use the upstream gem's Foreman-branded assets instead of the Satellite-branded assets from `foreman_theme_satellite_assets`.

This resulted in the Foreman helmet logo appearing on the login screen instead of the Red Hat Satellite logo.

**Root cause:** The spec file only checked for `.el8sat` and `.el9sat` when deciding whether to replace the upstream assets with Satellite-branded ones.

**Fix:** Add `.el10sat` to the downstream_build condition.

**Impact:** After this fix and a rebuild, el10sat packages will show the correct Red Hat Satellite logo on the login page.

Fixes: Login page shows Foreman logo instead of Satellite logo on el10sat

### DIFF
--- a/packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
+++ b/packages/satellite/rubygem-foreman_theme_satellite/rubygem-foreman_theme_satellite.spec
@@ -3,11 +3,11 @@
 %global plugin_name theme_satellite
 %global foreman_min_version 3.15.0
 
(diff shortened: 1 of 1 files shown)


Odilhao (approved):

Odilhao (approved):

ogajduse (approved):
:pog:

PR #14045, opened 2026-09-01
Cherry pick smart_proxy_openbolt unknown and foreman_openbolt 1.3.0 to rpm/3.17

nmburgan:


packages/plugins/rubygem-smart_proxy_openbolt/rubygem-smart_proxy_openbolt.spec
```diff
@@ -10,7 +10,7 @@
 
 Name: rubygem-%{gem_name}
 Version: 1.2.0
-Release: 1%{?foremandist}%{?dist}
+Release: 2%{?foremandist}%{?dist}
```
bastelfreak, line 13: Thisooks odd. Do we need to rebuild the package on 3.17?

nmburgan: Yeah, I think this picked up a change that happened with 1.2.1 but never got backported or something.

nmburgan: Also I need to update my backport script to not do smart_proxy_openbolt if we're only releasing foreman_openbolt. I thought I had a gate for putting up the actual PRs but apparently not.

PR #14050, opened 2026-09-01
Cherry pick foreman_openbolt 1.3.0 to rpm/3.19

nmburgan:


evgeni:
pick every commit on its own (in one PR), then the deletion of the old annex files is correct.

nmburgan:
Sorry, not sure I understand. Seems like we only have one commit here?

evgeni:
There are three changelog entries, each was an own commit.

nmburgan:
Oh, I see what you're saying. I'll fix that up.

PR #14088, opened 2026-09-07
Document git worktree

lzap:
Thought this would be useful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/rpm-develop*
+/deb-develop*
--- a/README.md
+++ b/README.md
@@ -26,3 +26,15 @@ Contributing
 
 It's generally best to contribute to `rpm/develop` or `deb/develop` unless something
 is specifically broken for an older release. Please fork and send a PR.
+
+Setup
+=====
+
+You can use `git worktree` to maintain two working directories with a shared `.git`:
+
+```bash
+git clone https://github.com/theforeman/foreman-packaging
+cd foreman-packaging
+git worktree add rpm-develop upstream/rpm/develop
+git worktree add deb-develop upstream/deb/develop
+```


ogajduse (commented):
Nice addition. A couple of thoughts on the naming, based on how `theforeman-rel-eng`'s `bump_packaging` script sets up worktrees when `GIT_USE_WORKTREES=true`.

That script expects a `master` checkout with flavor worktrees named `${flavor}-${version}` alongside it, e.g., `rpm-develop`, `deb-develop,`, and then `rpm-5.0`, `deb-5.0` for release branches.

Would it make sense to use that naming instead of `foreman-packaging-rpm`/`foreman-packaging-deb`? It scales to release branches for free, and it avoids a footgun for anyone who later starts using `bump_packaging`'s worktree mode.

Something like:
```bash
git clone https://github.com/theforeman/foreman-packaging -b master master
cd master
git worktree add ../rpm-develop upstream/rpm/develop
git worktree add ../deb-develop upstream/deb/develop
``

Not a blocker; I'm flagging this because it is an established pattern in the release tooling.

**Comment by lzap** [comment]
Hmm how about this slight modification:

```
git clone https://github.com/theforeman/foreman-packaging
cd foreman-packaging
git worktree add rpm-develop upstream/rpm/develop
git worktree add deb-develop upstream/deb/develop
```

Assuming that `/rpm-develop*` (as well as DEB) are in `.gitignore`.

Edit: Pushed such change.

adamruzicka:
> Hmm how about this slight modification:

That feels like a footgun waiting to go off. I'd prefer the worktrees to stay out of the original checkout

lzap:
> That feels like a footgun waiting to go off. I'd prefer the worktrees to stay out of the original checkout

Your choice, what I suggest is a README note only not mandatory thing. I reconfigured my checkout this way, let's see if I shoot myself.

Let me convert this to draft and see if this works.

PR #14103, opened 2026-09-11
Add postgres to recommends since backup needs it

sjha4:
foremanctl backup needs the pg_dump to create DB backups. We have a task to ensure this is installed before backup runs.
https://github.com/theforeman/foremanctl/blob/124a92b8e2272decfa9f48fc6ac4be4428e46c47/src/playbooks/backup/backup.yaml#L16

### DIFF
--- a/packages/foreman/foremanctl/foremanctl.spec
+++ b/packages/foreman/foremanctl/foremanctl.spec
@@ -2,7 +2,7 @@
 
 Name:      foremanctl
 Version:   3.2.0
-Release:   1%{?dist}
+Release:   2%{?dist}
 Summary:   Install Foreman using containers
 
 License:   GPL-2-only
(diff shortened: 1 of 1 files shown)


evgeni:
ok to test

evgeni (approved):

vsedmik (approved):

vijaysawant (approved):

evgeni:
[test rpm]

packages/foreman/foremanctl/foremanctl.spec
```diff
@@ -20,6 +20,7 @@ Recommends:  python3-libsemanage
 Recommends:  python3-psycopg2
 Recommends:  python3-requests
 Recommends:  python3-requests-oauthlib
+Recommends:  postgresql
```
Gauravtalreja1, line 23: I wonder if we really need this dependancy, especially when user has deployment with external db?

evgeni: It's the client bits, the server is in `postgresql-server`

And we use `pg_dump` e.g.

